PR #439, opened 2016-07-03
tests: fix leak in '[un]linkapps' integration test

UniqMartin:

-----

Prior to the fix, every run of the test suite would leave behind a pair of empty directories in `$TMPDIR`. (A temporary home directory was created but only its child `Applications` was wiped when done.)

Noticed while inspecting my local temporary directory. We might want to tweak the temporary directory used during our test suite run so that the file leak detection includes it, too.

cc @eirinikos because tests. :wink:


MikeMcQuaid:
👍

PR #459, opened 2016-07-05
doctor: check for beta XQuartz releases

mistydemeo:

-----

Beta versions of XQuartz have address sanitization enabled, which breaks some software at runtime, including wine. This is likely to result in more confusing error reports if we don't add a warning about it.

Closes #2481.

### DIFF
--- a/Library/Homebrew/extend/os/mac/diagnostic.rb
+++ b/Library/Homebrew/extend/os/mac/diagnostic.rb
@@ -312,6 +312,18 @@ module Homebrew
             https://xquartz.macosforge.org
         EOS
       end
+
+      def check_for_beta_xquartz
+        return unless MacOS::XQuartz.version.include? "beta"
+
+        <<-EOS.undent
+        The following beta release of XQuartz is installed: #{MacOS::XQuartz.version}
+
+        XQuartz beta releases include address sanitization, and do not work with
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
👍

Library/Homebrew/diagnostic.rb
```diff
@@ -1121,6 +1121,18 @@ def check_for_external_cmd_name_conflict
...
+      def check_for_beta_xquartz
+        return unless MacOS::XQuartz.version.include? "beta"
+
+        <<-EOS.undent
+        The following beta release of XQuartz is installed: #{MacOS::XQuartz.version}
+
+        XQuartz beta releases include address sanitization, and do not work with
+        all software; notably, wine will not work with beta releases of XQuartz.
+        We recommend only installing stable releases of XQuartz.
+        EOS
```
UniqMartin: Would you mind indenting the contents between `<<-EOS.undent` and `EOS` by two spaces? That's the style we use in all other messages in this file.

UniqMartin:
This check is macOS-specific, thus it should most likely go into `extend/os/mac/diagnostic.rb` instead of the generic parts contained in `diagnostic.rb`.

> Closes #2481.

This probably needs prefixing with `Homebrew/homebrew-core` for the cross-reference and the auto-close to work as expected.

PR #274, opened 2016-05-25
[WIP] Allow HEAD-formula upgrades

vladshablinsky:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

# Description

This is a GSoC PR for "Better support HEAD formulae" project.

- [x] Implement and override `VCSDownloadStrategy#last_commit` to determine unique identifier for the last commit in the repository.
- [x] Tests for GetDownloadStrategy
- [x] Implement `HeadVersion`
- [x] Add `HeadVersion` tests
- [x] Make `install/reinstall` work for HEAD versions
- [x] Midterm (June 20 - 27)
- [x] Implement `Formula#outdated_versions` (to be reviewed)
- [x] Establish `upgrade/outdated` behaviour  (to be reviewed)
- [ ] Update `print_outdated_json`
- [x] Allow to suppress download strategies output
- [x] Add outdated tests for HEAD versions
- [x] Use GitHub API to fetch last commit hash for github repositories (to be reviewed) 
- [ ] Fix edge cases (WIP)

cc @xu-cheng

PR #460, opened 2016-07-05
Implement `VCSDownloadStrategy#last_commit`

vladshablinsky:

-----

# Description

* Add `VCSDownloadStrategy#last_commit`
* Add `GitDownloadStrategyTests`

This PR is a part of https://github.com/Homebrew/brew/pull/274, so merged commits will be removed from https://github.com/Homebrew/brew/pull/274.

cc @MikeMcQuaid @xu-cheng

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -592,7 +592,7 @@ class GitDownloadStrategy < VCSDownloadStrategy
   end
 
   def last_commit
-    Utils.popen_read("git", "--git-dir", git_dir ,"rev-parse", "HEAD").chomp
+    Utils.popen_read("git", "--git-dir", git_dir ,"rev-parse", "--short", "HEAD").chomp
   end
 
   private
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -153,6 +153,12 @@ class VCSDownloadStrategy < AbstractDownloadStrategy
     version.head?
   end
 
+  # Return last commit's unique identifier for the repository.
+  # Return most recent modified timestamp unless overridden.
+  def last_commit
+    source_modified_time.to_i.to_s
+  end
+
   private
 
   def cache_tag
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
👍 :shipit:

UniqMartin:
:+1:

xu-cheng:
Thanks everyone.

UniqMartin:
FYI, I just fixed a minor problem with this in 32f7e738ecfd84e1d8a063f9af332fdc320dfb7d. It felt innocent enough that I committed it directly. Noticed this because I have a non-standard `core.abbrev` setting and the test suite was failing:

```
  1) Failure:
GitDownloadStrategyTests#test_last_commit [/opt/homebrew/Library/Homebrew/test/test_download_strategies.rb:122]:
--- expected
+++ actual
@@ -1 +1 @@
-"c50c79b"
+"c50c79b9"
```

vladshablinsky:
@UniqMartinI have read git documentation and now understand that I was wrong. Thanks for the fix!
I thought `git --git-dir rev-parse --short HEAD` would print `k` first symbols of revision string where `k` is the minimum length more or equal to 6 so that no other revision starts with that string.

So I would probably correct it this way if I wouldn't realise git works pretty different from my expectations:

```ruby
assert "c50c79b9573c9482b9178a1882695fbbcd36fe7d".start_with?(@strategy.last_commit)
```
Thanks again!

UniqMartin:
No worries! I also overlooked this detail when reviewing this PR. Your snippet, that would have changed the related test instead of the implementation, is also a perfectly valid fix for the test failure.

The reason I decided to make the change in the implementation is because I was of the opinion that these shortened commit hashes will eventually be used for the improved `HEAD` installation support, i.e. they will be used to create a “version number” like `HEAD-<last_commit>`. (Please correct me if I'm wrong about this.) If that's the case, it would be odd that this generated “version number” will be different across different systems with different Git configurations. (Of course there's also the small chance that the 7-character hash becomes non-unique as more commits are added to the repository, which would also change the output of `git rev-parse --short` even if the `=7` is added.)

vladshablinsky:
> Please correct me if I'm wrong about this.

Everything is right. I'm OK with your fix, it makes sense for me.

PR #464, opened 2016-07-05
Refactor accessing meta data of Git repositories

UniqMartin:

-----

Introduce `GitRepositoryExtension` that can be mixed-in into a `Pathname` instance that points at a Git repository (the path is the repository root), providing access to various Git-related meta data.

The goal is to avoid unnecessary code duplication across the `Homebrew.git_*` methods (they query the `brew` repository) and `Tap#git_*` methods (they query the repository associated with the given tap).

### DIFF
--- a/Library/Homebrew/global.rb
+++ b/Library/Homebrew/global.rb
@@ -1,6 +1,7 @@
 require "extend/module"
 require "extend/fileutils"
 require "extend/pathname"
+require "extend/git_repository"
 require "extend/ARGV"
 require "extend/string"
 require "extend/enumerable"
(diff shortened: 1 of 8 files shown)


Library/Homebrew/global.rb
```diff
@@ -19,6 +20,8 @@
 
 require "config"
 
+HOMEBREW_REPOSITORY.extend(GitRepositoryExtension)
```
MikeMcQuaid: I'm normally 👎 on more Pathname mixins but given it's on a single instance: 👍

MikeMcQuaid:
Nice work 👍

Library/Homebrew/global.rb
UniqMartin: I agree and thus intentionally limited this to the very few `Pathname` instances that actually need this. That all methods are prefixed with `git` also helps to avoid possible conflicts.

PR #470, opened 2016-07-07
Run linkage checks in `brew audit`

xu-cheng:
So remaining question is should we keep `brew linkage --test` to check undeclared deps.

### DIFF
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -1,3 +1,5 @@
+require "os/mac/linkage_checker"
+
 module FormulaCellarChecks
   def check_shadowed_headers
     return if ["libtool", "subversion", "berkeley-db"].any? do |formula_name|
@@ -56,10 +58,33 @@ module FormulaCellarChecks
     EOS
   end
 
+  def check_linkage
+    return unless formula.prefix.directory?
+    keg = Keg.new(formula.prefix)
+    checker = LinkageChecker.new(keg, formula)
+
+    if checker.broken_dylibs?
+      audit_check_output <<-EOS.undent
+        The installation was broken.
+        Broken dylib links found:
+          #{checker.broken_dylibs.to_a * "\n          "}
+      EOS
+    end
+
+    if checker.undeclared_deps?
+      audit_check_output <<-EOS.undent
+        Formulae are required to declare all linked dependencies.
+        Please add all linked dependencies to the formula with:
+          #{checker.undeclared_deps.map { |d| "depends_on \"#{d}\" => :linked"} * "\n          "}
+      EOS
+    end
+  end
+
   def audit_installed
     generic_audit_installed
     audit_check_output(check_shadowed_headers)
     audit_check_output(check_openssl_links)
     audit_check_output(check_python_framework_links(formula.lib))
+    check_linkage
   end
 end
--- a/Library/Homebrew/dependable.rb
+++ b/Library/Homebrew/dependable.rb
@@ -1,7 +1,7 @@
 require "options"
 
 module Dependable
-  RESERVED_TAGS = [:build, :optional, :recommended, :run]
+  RESERVED_TAGS = [:build, :optional, :recommended, :run, :linked]
 
   def build?
     tags.include? :build
--- a/Library/Homebrew/dev-cmd/linkage.rb
+++ b/Library/Homebrew/dev-cmd/linkage.rb
@@ -12,125 +12,20 @@
 #  --reverse   - For each dylib the keg references, print the dylib followed by the
 #                binaries which link to it.
 
-require "set"
-require "keg"
-require "formula"
+require "os/mac/linkage_checker"
 
 module Homebrew
   def linkage
-    found_broken_dylibs = false
     ARGV.kegs.each do |keg|
       ohai "Checking #{keg.name} linkage" if ARGV.kegs.size > 1
       result = LinkageChecker.new(keg)
       if ARGV.include?("--test")
         result.display_test_output
+        Homebrew.failed = true if result.broken_dylibs?
       elsif ARGV.include?("--reverse")
         result.display_reverse_output
       else
         result.display_normal_output
-      end
-      found_broken_dylibs = true unless result.broken_dylibs.empty?
-    end
-    if ARGV.include?("--test") && found_broken_dylibs
-      exit 1
-    end
-  end
-
-  class LinkageChecker
-    attr_reader :keg
-    attr_reader :broken_dylibs
-
-    def initialize(keg)
-      @keg = keg
-      @brewed_dylibs = Hash.new { |h, k| h[k] = Set.new }
-      @system_dylibs = Set.new
-      @broken_dylibs = Set.new
-      @variable_dylibs = Set.new
-      @reverse_links = Hash.new { |h, k| h[k] = Set.new }
-      check_dylibs
-    end
-
-    def check_dylibs
-      @keg.find do |file|
-        next if file.symlink? || file.directory?
-        next unless file.dylib? || file.mach_o_executable? || file.mach_o_bundle?
-        file.dynamically_linked_libraries.each do |dylib|
-          @reverse_links[dylib] << file
-          if dylib.start_with? "@"
-            @variable_dylibs << dylib
-          else
-            begin
-              owner = Keg.for Pathname.new(dylib)
-            rescue NotAKegError
-              @system_dylibs << dylib
-            rescue Errno::ENOENT
-              @broken_dylibs << dylib
-            else
-              @brewed_dylibs[owner.name] << dylib
-            end
-          end
-        end
-      end
-
-      begin
-        f = Formulary.from_rack(keg.rack)
-        f.build = Tab.for_keg(keg)
-        filter_out = proc do |dep|
-          dep.build? || (dep.optional? && !dep.option_names.any? { |n| f.build.with?(n) })
-        end
-        declared_deps = f.deps.reject { |dep| filter_out.call(dep) }.map(&:name) +
-                        f.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
-        @undeclared_deps = @brewed_dylibs.keys - declared_deps.map { |dep| dep.split("/").last }
-        @undeclared_deps -= [f.name]
-      rescue FormulaUnavailableError
-        opoo "Formula unavailable: #{keg.name}"
-        @undeclared_deps = []
-      end
-    end
-
-    def display_normal_output
-      display_items "System libraries", @system_dylibs
-      display_items "Homebrew libraries", @brewed_dylibs
-      display_items "Variable-referenced libraries", @variable_dylibs
-      display_items "Missing libraries", @broken_dylibs
-      display_items "Possible undeclared dependencies", @undeclared_deps
-    end
-
-    def display_reverse_output
-      return if @reverse_links.empty?
-      sorted = @reverse_links.sort
-      sorted.each do |dylib, files|
-        puts dylib
-        files.each do |f|
-          unprefixed = f.to_s.strip_prefix "#{@keg.to_s}/"
-          puts "  #{unprefixed}"
-        end
-        puts unless dylib == sorted.last[0]
-      end
-    end
-
-    def display_test_output
-      display_items "Missing libraries", @broken_dylibs
-      puts "No broken dylib links" if @broken_dylibs.empty?
-    end
-
-    private
-
-    # Display a list of things.
-    # Things may either be an array, or a hash of (label -> array)
-    def display_items(label, things)
-      return if things.empty?
-      puts "#{label}:"
-      if things.is_a? Hash
-        things.sort.each do |list_label, list|
-          list.sort.each do |item|
-            puts "  #{item} (#{list_label})"
-          end
-        end
-      else
-        things.sort.each do |item|
-          puts "  #{item}"
-        end
       end
     end
   end
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -3,12 +3,13 @@ require "keg"
 require "formula"
 
 class LinkageChecker
-  attr_reader :keg
+  attr_reader :keg, :formula
   attr_reader :brewed_dylibs, :system_dylibs, :broken_dylibs, :variable_dylibs
   attr_reader :undeclared_deps, :reverse_links
 
-  def initialize(keg)
+  def initialize(keg, formula = nil)
     @keg = keg
+    @formula = formula || resolve_formula(keg)
     @brewed_dylibs = Hash.new { |h, k| h[k] = Set.new }
     @system_dylibs = Set.new
     @broken_dylibs = Set.new
@@ -34,25 +35,43 @@ class LinkageChecker
           rescue Errno::ENOENT
             @broken_dylibs << dylib
           else
-            @brewed_dylibs[owner.name] << dylib
+            tap = Tab.for_keg(owner).tap
+            f = if tap.nil? || tap.core_tap?
+              owner.name
+            else
+              "#{tap}/#{owner.name}"
+            end
+            @brewed_dylibs[f] << dylib
           end
         end
       end
     end
 
-    begin
-      f = Formulary.from_rack(keg.rack)
-      f.build = Tab.for_keg(keg)
+    @undeclared_deps = check_undeclared_deps if formula
+  end
+
+  def check_undeclared_deps
       filter_out = proc do |dep|
-        dep.build? || (dep.optional? && !dep.option_names.any? { |n| f.build.with?(n) })
+        next true if dep.build?
+        dep.optional? && !dep.option_names.any? { |n| formula.build.with?(n) }
       end
-      declared_deps = f.deps.reject { |dep| filter_out.call(dep) }.map(&:name) +
-                      f.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
-      @undeclared_deps = @brewed_dylibs.keys - declared_deps.map { |dep| dep.split("/").last }
-      @undeclared_deps -= [f.name]
-    rescue FormulaUnavailableError
-      opoo "Formula unavailable: #{keg.name}"
-    end
+      declared_deps = formula.deps.reject { |dep| filter_out.call(dep) }.map(&:name)
+      declared_requirement_deps = formula.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
+      declared_dep_names = (declared_deps + declared_requirement_deps).map { |dep| dep.split("/").last }
+      undeclared_deps = @brewed_dylibs.keys.select do |full_name|
+        name = full_name.split("/").last
+        next false if name == formula.name
+        !declared_dep_names.include?(name)
+      end
+      undeclared_deps.sort do |a,b|
+        if a.include?("/") && !b.include?("/")
+          1
+        elsif !a.include?("/") && b.include?("/")
+          -1
+        else
+          a <=> b
+        end
+      end
   end
 
   def display_normal_output
@@ -79,8 +98,6 @@ class LinkageChecker
   def display_test_output
     display_items "Missing libraries", @broken_dylibs
     puts "No broken dylib links" if @broken_dylibs.empty?
-    display_items "Possible undeclared dependencies", @undeclared_deps
-    puts "No undeclared dependencies" if @undeclared_deps.empty?
   end
 
   def broken_dylibs?
@@ -109,5 +126,13 @@ class LinkageChecker
         puts "  #{item}"
       end
     end
+  end
+
+  def resolve_formula(keg)
+    f = Formulary.from_rack(keg.rack)
+    f.build = Tab.for_keg(keg)
+    f
+  rescue FormulaUnavailableError
+    opoo "Formula unavailable: #{keg.name}"
   end
 end
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -1,0 +1,113 @@
+require "set"
+require "keg"
+require "formula"
+
+class LinkageChecker
+  attr_reader :keg
+  attr_reader :brewed_dylibs, :system_dylibs, :broken_dylibs, :variable_dylibs
+  attr_reader :undeclared_deps, :reverse_links
+
+  def initialize(keg)
+    @keg = keg
+    @brewed_dylibs = Hash.new { |h, k| h[k] = Set.new }
+    @system_dylibs = Set.new
+    @broken_dylibs = Set.new
+    @variable_dylibs = Set.new
+    @undeclared_deps = []
+    @reverse_links = Hash.new { |h, k| h[k] = Set.new }
+    check_dylibs
+  end
+
+  def check_dylibs
+    @keg.find do |file|
+      next if file.symlink? || file.directory?
+      next unless file.dylib? || file.mach_o_executable? || file.mach_o_bundle?
+      file.dynamically_linked_libraries.each do |dylib|
+        @reverse_links[dylib] << file
+        if dylib.start_with? "@"
+          @variable_dylibs << dylib
+        else
+          begin
+            owner = Keg.for Pathname.new(dylib)
+          rescue NotAKegError
+            @system_dylibs << dylib
+          rescue Errno::ENOENT
+            @broken_dylibs << dylib
+          else
+            @brewed_dylibs[owner.name] << dylib
+          end
+        end
+      end
+    end
+
+    begin
+      f = Formulary.from_rack(keg.rack)
+      f.build = Tab.for_keg(keg)
+      filter_out = proc do |dep|
+        dep.build? || (dep.optional? && !dep.option_names.any? { |n| f.build.with?(n) })
+      end
+      declared_deps = f.deps.reject { |dep| filter_out.call(dep) }.map(&:name) +
+                      f.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
+      @undeclared_deps = @brewed_dylibs.keys - declared_deps.map { |dep| dep.split("/").last }
+      @undeclared_deps -= [f.name]
+    rescue FormulaUnavailableError
+      opoo "Formula unavailable: #{keg.name}"
+    end
+  end
+
+  def display_normal_output
+    display_items "System libraries", @system_dylibs
+    display_items "Homebrew libraries", @brewed_dylibs
+    display_items "Variable-referenced libraries", @variable_dylibs
+    display_items "Missing libraries", @broken_dylibs
+    display_items "Possible undeclared dependencies", @undeclared_deps
+  end
+
+  def display_reverse_output
+    return if @reverse_links.empty?
+    sorted = @reverse_links.sort
+    sorted.each do |dylib, files|
+      puts dylib
+      files.each do |f|
+        unprefixed = f.to_s.strip_prefix "#{@keg}/"
+        puts "  #{unprefixed}"
+      end
+      puts unless dylib == sorted.last[0]
+    end
+  end
+
+  def display_test_output
+    display_items "Missing libraries", @broken_dylibs
+    puts "No broken dylib links" if @broken_dylibs.empty?
+    display_items "Possible undeclared dependencies", @undeclared_deps
+    puts "No undeclared dependencies" if @undeclared_deps.empty?
+  end
+
+  def broken_dylibs?
+    !@broken_dylibs.empty?
+  end
+
+  def undeclared_deps?
+    !@undeclared_deps.empty?
+  end
+
+  private
+
+  # Display a list of things.
+  # Things may either be an array, or a hash of (label -> array)
+  def display_items(label, things)
+    return if things.empty?
+    puts "#{label}:"
+    if things.is_a? Hash
+      things.sort.each do |list_label, list|
+        list.sort.each do |item|
+          puts "  #{item} (#{list_label})"
+        end
+      end
+    else
+      things.sort.each do |item|
+        puts "  #{item}"
+      end
+    end
+  end
+end


ilovezfs:
Maybe change it to `linkage --broken` and `linkage --undeclared`?

ilovezfs:
I think there needs to be a way to suppress the undeclared linkages from causing audit failure on mere formula revision (not version) bumps, (but possibly flipped back on if a dependency is being added), so I think this needs to be exposed with some kind of command line option to either explicitly opt in or explicitly opt out of the undeclared check.

Also, should we fail audit if something is declared as a dependency but it's NOT linked && NOT marked `=> :run? I think we're going to end up starting to accumulate `depends_on`s that are no longer actually used if that's not being checked, too.

MikeMcQuaid:
> I think there needs to be a way to suppress the undeclared linkages from causing audit failure on mere formula revision

I disagree; you can merge a PR even if CI is red if you know better that it does.

ilovezfs:
>you can merge a PR even if CI is red

But that violates my religion :(

MikeMcQuaid:
@ilovezfs All fall short of the glory of ~~@God~~ @BrewTestBot

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -208,6 +210,31 @@ def check_elisp_root(share, name)
...
+    if checker.broken_dylibs?
+      s += <<-EOS.undent
+      Broken dylib links found:
+        #{checker.broken_dylibs.to_a * "\n        "}
+      EOS
+    end
+
+    if checker.undeclared_deps?
+      s += <<-EOS.undent
+      Undelcared dependencies found:
```
ilovezfs: s/Undelc/Undecl/

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -208,6 +210,31 @@ def check_elisp_root(share, name)
...
+        #{checker.undeclared_deps * "\n        "}
+
+      Formula is required to delcare all direct dependencies.
```
ilovezfs: s/delc/decl/

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -208,6 +210,31 @@ def check_elisp_root(share, name)
...
+      Please check whether these dependencies are intended.
```
ilovezfs: Not sure "intended" is the right word here.

ilovezfs:
This is causing a regression
```
Error: /usr/local/Cellar/amazon-ecs-cli/0.3.1 is not a directory
```

where it doesn't let you run audit at all unless the formula is installed now.
```
iMac-TMP:local joe$ brew uninstall amazon-ecs-cli
Error: No such keg: /usr/local/Cellar/amazon-ecs-cli
iMac-TMP:local joe$ git reset --hard origin/master
HEAD is now at 86b1df9 Autoupdate tweaks (#442)
iMac-TMP:local joe$ brew audit --strict amazon-ecs-cli
iMac-TMP:local joe$ brew pull https://github.com/Homebrew/brew/pull/470
==> Fetching patch 
Patch: https://github.com/Homebrew/brew/pull/470.patch
==> Applying patch
Applying: move LinkageChecker to standalone file
Applying: formula_cellar_checks: add check_linkage
==> Patch closes issue #470
==> Patch changed:
 Library/Homebrew/dev-cmd/linkage.rb       | 114 +----------------------------
 Library/Homebrew/formula_cellar_checks.rb |  28 +++++++
 Library/Homebrew/linkage_checker.rb       | 118 ++++++++++++++++++++++++++++++
 3 files changed, 147 insertions(+), 113 deletions(-)
iMac-TMP:local joe$ brew audit --strict amazon-ecs-cli
Error: /usr/local/Cellar/amazon-ecs-cli/0.3.1 is not a directory

```

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -208,6 +210,31 @@ def check_elisp_root(share, name)
     EOS
   end
 
+  def check_linkage
+    keg = Keg.new(formula.prefix)
```
UniqMartin: The failure for when a formula is not installed should be fixable with `return unless formula.prefix.directory?` prior to this line, just like in the `check_openssl_links` method.

Library/Homebrew/formula_cellar_checks.rb
UniqMartin: My suggestion would be something along the lines of: “Please check wether these dependencies are needed or whether some library was automatically (and unintentionally) picked up by the build system of the formula, something that can usually be influenced with `--without-THING` or some equivalent option.” (As a comprehensible explanation might become even longer, it might make sense to provide just a link and to come up with a dedicated page in our documentation.)

Furthermore, `brew linkage --reverse <formula>` will show what binary of the formula is linking to what, so this might help with the diagnosis of this message and could be suggested as well.

Library/Homebrew/linkage_checker.rb
```diff
@@ -3,12 +3,13 @@
 require "formula"
 
 class LinkageChecker
-  attr_reader :keg
+  attr_reader :keg, :formula
   attr_reader :brewed_dylibs, :system_dylibs, :broken_dylibs, :variable_dylibs
   attr_reader :undeclared_deps, :reverse_links
 
-  def initialize(keg)
+  def initialize(keg, formula=nil)
```
UniqMartin: The `=` lacks surrounding spaces. Would be nice to run RuboCop on this file to see if it flags any other style issues.

Library/Homebrew/linkage_checker.rb
```diff
@@ -3,12 +3,13 @@
...
     @keg = keg
+    @formula = formula || resolve_formula
```
UniqMartin: I think I'd find it easier to understand if `resolve_formula` would accept a `keg` argument instead of operating on the `keg` attribute, but I don't feel strongly about this.

Library/Homebrew/linkage_checker.rb
```diff
@@ -40,18 +41,14 @@ def check_dylibs
...
       end
-      declared_deps = f.deps.reject { |dep| filter_out.call(dep) }.map(&:name) +
-                      f.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
+      declared_deps = formula.deps.reject { |dep| filter_out.call(dep) }.map(&:name) +
+                      formula.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
       @undeclared_deps = @brewed_dylibs.keys - declared_deps.map { |dep| dep.split("/").last }
-      @undeclared_deps -= [f.name]
-    rescue FormulaUnavailableError
-      opoo "Formula unavailable: #{keg.name}"
+      @undeclared_deps -= [formula.name]
```
UniqMartin: Since you have to make changes to this entire code block: I think it would be better to use an early exit with `return unless formula` or better yet relocate the calculations done here into a separate (private) method, e.g.:

```ruby
@undeclared_deps = find_undeclared_deps if formula
```

UniqMartin:
For the background on my comments, please see <https://github.com/Homebrew/brew/commit/080ddd8804be14f4b18f9558b58270456ff313c2#commitcomment-18165984>.

> So remaining question is should we keep `brew linkage --test` to check undeclared deps.

I think we should revert `--test` to its previous meaning of only checking broken linkage (and possibly rename it to `--linkage` as suggested by @ilovezfs). Undeclared dependencies should not cause a failure unless explicitly requested. To summarize:

- `brew linkage <formula>` should print all the traditional information and always exit with success. This is basically verbose linkage information on a per-formula basis and for human consumption only.

- `brew linkage <formula> --reverse` should also print its usual thing and always exit with success. This is even more verbose linkage information on a per-file basis and also for humans only.

- `brew linkage <formula> [--linkage] [--undeclared-deps]` (where at least one of the options is given), should run either of the tests or both and exit with failure if any of them report failure. Maybe a clearer interface would be:

  ```
  brew linkage <formula> --test[=<tests>]
  ```

  where `<tests>` accepts a comma-separated list of checks to perform (currently `linkage` and `undeclared-deps`) and defaults to `linkage`.

> Also, should we fail audit if something is declared as a dependency but it's NOT linked && NOT marked `=> :run` (or it goes without saying `=> :build`)? I think we're going to end up starting to accumulate `depends_on`s that are no longer actually used if that's not being checked, too.

Unfortunately the relationship between dependencies and dynamic linkage is not an equivalence. Dynamic linkage implies dependency, but not the other way round. There are so many different reasons why a formula may need a certain dependency on run time beyond just dynamic linkage. I don't think expanding our use of `:run` for this is going to scale.

The problem with possibly no longer needed `depends_on` lines is already a reality. I think it's out of scope for this PR to address this and I fear it's a challenging task to try to tackle.

ilovezfs:
> I think it's out of scope for this PR to address this 

It's not out of the scope insofar as we're talking about rigorously "dragging in" all linked but undeclared dependencies. The undeclared dependencies exist only because a declared dependency caused them to be installed, or one of the dependencies of a declared dependencies (potentially recursively) caused the them to be installed. Then the undeclared dependency got linked during the build, but that linkage was not (until now) formalized with an actual declaration. However, when/if upstream later removes the original direct dependency that caused the undeclared dependency to occur, there is nothing this PR does to unwind the formalized declaration of the originally undeclared yet linked dependency. And indeed it may end up continuing to be linked at that point ONLY because it is declared or it may end up NOT LINKED AT ALL and yet still declared ("dangling dependency"). Neither of those cases is acceptable and yet we will guarantee they occur all over the place because so many build systems opportunistically link against what's available (one of the major points of superenv hiding, or at least making a best effort to hide, things we didn't ask it to expose) and because the dependencies we're adding solely because they were undeclared are in no way designated as such and are not going to be pointed out by upstream as no longer being needed because upstream never explicitly said they were needed in the first place, so there would be no reason to call them out when later the originating dependency is dropped. As a result, the only way these dependencies will ever get removed is if we go in and reassess every single dependency in the formula to determine whether or not it is actually required directly, actually required indirectly, only opportunistically linked but not actually needed (except perhaps sometimes optionally), or not linked at all.

We already have this problem, but deliberately combining all-things-that-are-linked with all-things-upstream-says-are-required under the same umbrella of "depends_on" with no distinction whatsoever in the DSL between them is going to cause this problem grow at faster rater, and cause it to accumulate over time.

And no I don't think :run actually solves any of this, but the point would be to at least call out things that are not marked build and are not linked as part of audit, which addresses the dangling half but not the opportunistic half. Marking with :run would be the way to "override" that audit error, but of course :dangling-but-that's-fine would work too.

Moreover, the opportunistic half of the problem is the worse of the two anyway because it's much harder to see even if you're looking for it. This is why undeclared dependencies are not actually bugs but features unless they're distinguished from the actual intended direct dependencies in some way in the DSL itself.

Library/Homebrew/linkage_checker.rb
```diff
@@ -0,0 +1,121 @@
...
+    @variable_dylibs = Set.new
+    @undeclared_deps = []
+    @reverse_links = Hash.new { |h, k| h[k] = Set.new }
+    check_dylibs
+  end
+
+  def check_dylibs
+    @keg.find do |file|
+      next if file.symlink? || file.directory?
+      next unless file.dylib? || file.mach_o_executable? || file.mach_o_bundle?
```
MikeMcQuaid: This code should be in `os/mac` as it's OS X specific.

MikeMcQuaid: (Possibly the whole file and the checks)

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -208,6 +210,32 @@ def check_elisp_root(share, name)
...
 
+  def check_linkage
+    return unless formula.prefix.directory?
+    keg = Keg.new(formula.prefix)
+    checker = LinkageChecker.new(keg, formula)
+    s = ""
+
+    if checker.broken_dylibs?
+      s += <<-EOS.undent
+      Broken dylib links found:
```
MikeMcQuaid: What's the actionable outcome? Tell people what they need to do here.

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -208,6 +210,32 @@ def check_elisp_root(share, name)
...
+      EOS
+    end
+
+    if checker.undeclared_deps?
+      s += <<-EOS.undent
+      Undeclared dependencies found:
+        #{checker.undeclared_deps * "\n        "}
+
+      Formula is required to declare all direct dependencies.
+      Please check whether these dependencies are needed.
```
MikeMcQuaid: What's the actionable outcome? Tell people what they need to do here i.e. what lines in the formula need added/removed/changed.

Library/Homebrew/linkage_checker.rb
```diff
@@ -0,0 +1,121 @@
...
+
+class LinkageChecker
+  attr_reader :keg, :formula
+  attr_reader :brewed_dylibs, :system_dylibs, :broken_dylibs, :variable_dylibs
+  attr_reader :undeclared_deps, :reverse_links
+
+  def initialize(keg, formula = nil)
+    @keg = keg
+    @formula = formula || resolve_formula(keg)
+    @brewed_dylibs = Hash.new { |h, k| h[k] = Set.new }
```
MikeMcQuaid: Not clear what a `brewed` dylib is? Installed?

Library/Homebrew/linkage_checker.rb
```diff
@@ -0,0 +1,121 @@
...
+    @system_dylibs = Set.new
+    @broken_dylibs = Set.new
+    @variable_dylibs = Set.new
```
MikeMcQuaid: Also not clear what a `variable` one is.

Library/Homebrew/linkage_checker.rb
```diff
@@ -0,0 +1,121 @@
...
+        end
+      end
+    end
+
+    @undeclared_deps = check_undeclared_deps if formula
+  end
+
+  def check_undeclared_deps
+      filter_out = proc do |dep|
+        dep.build? || (dep.optional? && !dep.option_names.any? { |n| formula.build.with?(n) })
```
MikeMcQuaid: Can you split this logic into multiple `next` statements so you don't need to process the binary logic in your head? Thanks.

Library/Homebrew/linkage_checker.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+      declared_deps = formula.deps.reject { |dep| filter_out.call(dep) }.map(&:name) +
+                      formula.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
```
MikeMcQuaid: Split the two `reject`s into separate variables and combine them.

Library/Homebrew/linkage_checker.rb
```diff
@@ -0,0 +1,121 @@
...
+      undeclared_deps = @brewed_dylibs.keys - declared_deps.map { |dep| dep.split("/").last }
```
MikeMcQuaid: Split the `declared_deps.map` into a separate variable.

Library/Homebrew/linkage_checker.rb
```diff
@@ -0,0 +1,121 @@
...
+    display_items "Possible undeclared dependencies", @undeclared_deps
+    puts "No undeclared dependencies" if @undeclared_deps.empty?
+  end
+
+  def broken_dylibs?
+    !@broken_dylibs.empty?
+  end
+
+  def undeclared_deps?
+    !@undeclared_deps.empty?
```
MikeMcQuaid: `@undeclared_deps.any?`

MikeMcQuaid:
> Marking with :run would be the way to "override" that audit error, but of course :dangling-but-that's-fine would work too.

I think having `:linked` or something may be what we want here but, for changes like this, we really want/need a way to be able to run it over all our packages. I too 😬 a bit at our approaches sometimes of "add a new rule, make the new poor soul to bump the package fix it"; it's pretty unfriendly to first-time contributors.

ilovezfs:
@UniqMartin 

>(and possibly rename it to --linkage as suggested by @ilovezfs)

I think I said `--broken` and `--undeclared`. `brew linkage --linkage` sounds slightly redundant.

UniqMartin:
> I think I said `--broken` and `--undeclared`. `brew linkage --linkage` sounds slightly redundant.

You're right. Sounds silly. Sorry for the misquote!

ilovezfs:
>You're right. Sounds silly. Sorry for the misquote!

No problem. It made me laugh.

Library/Homebrew/formula_cellar_checks.rb
UniqMartin: I made a few suggestions in a (now hidden by default) comment: https://github.com/Homebrew/brew/pull/470#discussion_r70000278

Library/Homebrew/linkage_checker.rb
UniqMartin: These are `dylib`s installed by Homebrew, i.e. from an installed formula. Should be relatively obvious from where this is populated further down. Not sure if there's really a better name.

Do you realize that this and the following comments are about code that isn't actually new and has just moved from one file to another? (I believe you prefer not to modify code while moving.)

Library/Homebrew/linkage_checker.rb
UniqMartin: One prefixed with `@rpath/`, `@loader_path/`, or `@executable_path/`. Those are resolved by the dynamic linker and aren't immediately accessible to static analysis (what is done here in this class). FWIW, `dyld(1)` uses the same terminology.

Library/Homebrew/linkage_checker.rb
UniqMartin: I object strongly and I'm the one who insisted on `!empty?` instead of `any?`. While the latter is more convenient to type and is more succinct, they are not equivalent and have different semantics<sup>(1)</sup> and performance characteristics<sup>(2)</sup>. (I realize that `any?` is used a lot in Homebrew code where it should have been `empty?`, but I'd personally prefer not to proliferate the use of `any?` in new code.)

<sup>(1)</sup> `[nil].any?` vs. `![nil].empty?`
<sup>(2)</sup> Querying the size of the container (usually in constant time) vs. iterating over all members of a container and executing a block on every single element (linear time; though it might be doing short-circuit evaluation, but that's not obvious from the documentation).

xu-cheng:
Only more comments before shipping this?

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -56,10 +58,41 @@ def check_python_framework_links(lib)
...
+    end
+
+    if checker.undeclared_deps?
+      s += <<-EOS.undent
+        Undeclared dependencies found:
+          #{checker.undeclared_deps * "\n          "}
+
+        Formula is required to declare all direct dependencies.
+        Please check whether these dependencies are needed.
+        If so, declare them in the formula using `depends_on`.
```
ilovezfs: This seems too tentative and conditional to me. If I "check whether these dependencies are needed" and decide "nope, they're not," the audit will nonetheless continue to fail until I either add the depends_on or take some affirmative steps to prevent the linkage from occurring.

xu-cheng: That's right. A formula is required to declare all of its directly deps. 
If you don't declared the needed deps, the bottle can be broken as they won't be installed by users.

ilovezfs:
I'd still like to see there be some standard way of designating things that are added solely because "audit says it's undeclared but linked" before we start enforcing this. Even if we punt and make that a standard Ruby comment, that's better than nothing.

Undeclared dependencies are a feature not a bug, but we're killing them off without preserving the information about what was required directly and what got brought in due to something else. (The only case where they're *actually* a bug is if brew install doesn't install them unless they're declared and yet they're truly necessary, but that is explicitly never the case here since they're linked but undeclared precisely because something else caused them to be installed. If they were undeclared, unlinked, and yet necessary, audit would not report them.)

If the thing that was originally required directly no longer is, an undeclared dependency will naturally drop away. A declared dependency will not.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
MikeMcQuaid: `Formulae are required to declare all linked dependencies. Please add all missing dependencies to the formula with` and then a list of `depends_on "foo"` etc.

MikeMcQuaid:
> If the thing that was originally required directly no longer is, an undeclared dependency will naturally drop away. A declared dependency will not.

I think this is still a problem. Perhaps a `depends_on "foo" => :linked` or similar (which we could suggest in the `audit` text will throw an `audit` failure if it _isn't_ linked against in future?

ilovezfs:
@MikeMcQuaid even better it would also say *exactly what formula(e) brought it in originally* (e.g., `depends_on "foo", :linked => ["needy_thing1", "needy_thing2"]` since the most problematic cases are opportunistic linking to formulae that are not actually required anymore (or even "worse" weren't ever required in the first place (!) but got linked simply because they were found by configure) and yet now is still getting linked anyway, even after the original thing that brought it in has been deliberately removed as a depends_on of the formula, due to opportunistic linking/discovery in configure, and not due to some explicit option we passed to configure asking for it to enable that option.

MikeMcQuaid:
> @MikeMcQuaid even better it would also say exactly what formula(e) brought it in originally (e.g., depends_on "foo", :linked => ["needy_thing1", "needy_thing2"] since the most problematic cases are opportunistic linkages to formulae that are not actually required anymore (or even "worse" weren't ever required in the first place (!) but got linked simply because they were found by configure) and yet now are still getting linked anyway, even after the original thing(s) that brought them in have been deliberately removed as depends_ons of the formula, due to opportunistic linking/discovery in configure, and not due to some explicit option we pass to configure asking for configure to enable them.

That's a neat idea but I think it's a feature request to address opportunistic linking better through superenv (once we have these dependencies tagged we can start filtering out more with superenv) and the opportunistic linking will likely be revealed by future `brew audit` runs.

ilovezfs:
@MikeMcQuaid The point is to make it easier for someone removing a dependency to know which `:linked` to reassess as part of the removal procedures, but, yes, we can always require people to do more work in the future and have them sit there and check every single one of the `:linked` even though we could have left a proper message in a bottle (or in this case in a formula) so-to-speak (pun intended). Hopefully the list of `:linked` is short enough that it's as feature-requesty as you suggest.

ilovezfs:
The other reason that we ought to say what's bring it in, is that will allow audit to puke up an error if in the future the `needy_thing1` is no longer a `depends_on` and yet the `:linked` is still sitting there. A `:linked` that doesn't say whose entourage it belonged to leaves audit in the dark.

ilovezfs:
Another thought about this PR. Should we phase this in and start out by always doing the check and printing the undeclared, but only make the presence of undeclared an audit *failure* if `--strict` is passed? That might soften the blow to CI a bit.

MikeMcQuaid:
I'm still not sure I understand by what you mean by "what's bringing it in"; if it links directly (not recursively) to something else I'm not sure that's machine identifiable so it seems fine to omit.

> Another thought about this PR. Should we phase this in and start out by always doing the check and printing the undeclared, but only make the presence of undeclared an audit failure if --strict is passed? That might soften the blow to CI a bit.

I'm 👍 on keeping it as a failure for now.

ilovezfs:
>I'm still not sure I understand by what you mean by "what's bringing it in"; if it links directly (not recursively) to something else I'm not sure that's machine identifiable so it seems fine to omit.

"foo brings in bar" means, for instance, all of the following are true:
a) `my_formula` has a `depends_on "foo"` and does not have a `depends_on "bar"`
b) `brew audit my_formula` says that `bar` is linked but undeclared
c) the reason that `bar` got installed during `brew install --build-bottle my_formula` is that `foo` has a `depends_on "bar"`

MikeMcQuaid:
> c) the reason that bar got installed during brew install --build-bottle my_formula is that foo has a depends_on "bar"

This is the bit I don't think we can necessarily identify automatically through code but if you can: go for it (but I don't think it needs to block this PR).

ilovezfs:
>This is the bit I don't think we can necessarily identify automatically through code

It's the existing`depends_on` (one or more) in `my_formula` for which it is true that `brew deps "foo"` already lists `bar`.

ilovezfs:
Example:
```
iMac-TMP:Formula joe$ brew linkage file-roller
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/atk/lib/libatk-1.0.0.dylib (atk)
  /usr/local/opt/cairo/lib/libcairo-gobject.2.dylib (cairo)
  /usr/local/opt/cairo/lib/libcairo.2.dylib (cairo)
  /usr/local/opt/gdk-pixbuf/lib/libgdk_pixbuf-2.0.0.dylib (gdk-pixbuf)
  /usr/local/opt/gettext/lib/libintl.8.dylib (gettext)
  /usr/local/opt/glib/lib/libgio-2.0.0.dylib (glib)
  /usr/local/opt/glib/lib/libglib-2.0.0.dylib (glib)
  /usr/local/opt/glib/lib/libgobject-2.0.0.dylib (glib)
  /usr/local/opt/glib/lib/libgthread-2.0.0.dylib (glib)
  /usr/local/opt/gtk+3/lib/libgdk-3.0.dylib (gtk+3)
  /usr/local/opt/gtk+3/lib/libgtk-3.0.dylib (gtk+3)
  /usr/local/opt/json-glib/lib/libjson-glib-1.0.0.dylib (json-glib)
  /usr/local/opt/libarchive/lib/libarchive.13.dylib (libarchive)
  /usr/local/opt/libmagic/lib/libmagic.1.dylib (libmagic)
  /usr/local/opt/pango/lib/libpango-1.0.0.dylib (pango)
  /usr/local/opt/pango/lib/libpangocairo-1.0.0.dylib (pango)
iMac-TMP:Formula joe$ perl -p -i -e 's/depends_on "atk"//' file-roller.rb 
iMac-TMP:Formula joe$ brew linkage file-roller
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /usr/local/opt/atk/lib/libatk-1.0.0.dylib (atk)
  /usr/local/opt/cairo/lib/libcairo-gobject.2.dylib (cairo)
  /usr/local/opt/cairo/lib/libcairo.2.dylib (cairo)
  /usr/local/opt/gdk-pixbuf/lib/libgdk_pixbuf-2.0.0.dylib (gdk-pixbuf)
  /usr/local/opt/gettext/lib/libintl.8.dylib (gettext)
  /usr/local/opt/glib/lib/libgio-2.0.0.dylib (glib)
  /usr/local/opt/glib/lib/libglib-2.0.0.dylib (glib)
  /usr/local/opt/glib/lib/libgobject-2.0.0.dylib (glib)
  /usr/local/opt/glib/lib/libgthread-2.0.0.dylib (glib)
  /usr/local/opt/gtk+3/lib/libgdk-3.0.dylib (gtk+3)
  /usr/local/opt/gtk+3/lib/libgtk-3.0.dylib (gtk+3)
  /usr/local/opt/json-glib/lib/libjson-glib-1.0.0.dylib (json-glib)
  /usr/local/opt/libarchive/lib/libarchive.13.dylib (libarchive)
  /usr/local/opt/libmagic/lib/libmagic.1.dylib (libmagic)
  /usr/local/opt/pango/lib/libpango-1.0.0.dylib (pango)
  /usr/local/opt/pango/lib/libpangocairo-1.0.0.dylib (pango)
Possible undeclared dependencies:
  atk
iMac-TMP:Formula joe$ brew deps -1 file-roller | while read f; do brew deps $f | grep -q atk && echo $f ; done
gtk+3
```

ilovezfs:
So for file-roller you'd have
```
  depends_on "atk", :linked => "gtk+3"
  depends_on "cairo", :linked => ["gdk-pixbuf", "gtk+3", "json-glib", "pango"]
  depends_on "gdk-pixbuf", :linked => "gtk+3"
  depends_on "gettext", :linked => ["glib", "gtk+3", "json-glib", "pango"]
  depends_on "glib", :linked => ["gtk+3", "json-glib", "pango"]
  depends_on "pango", :linked => "gtk+3"
```

MikeMcQuaid:
Ok, I see. If you can make the code do that: 👍 (and throw a similar audit if there's trailing linked values that shouldn't be).

ilovezfs:
Yeah, I'll think about it, sure.

Also, it's probably even more useful if deps that are themselves listed as `:linked` be excluded from the lists, in which case the above reduces to:
```
depends_on "atk", :linked => "gtk+3"
depends_on "cairo", :linked => ["gtk+3", "json-glib"]
depends_on "gdk-pixbuf", :linked => "gtk+3"
depends_on "gettext", :linked => ["gtk+3", "json-glib"]
depends_on "glib", :linked => ["gtk+3", "json-glib"]
depends_on "pango", :linked => "gtk+3"
```
which I think makes it pretty clear what's actually going on with these previously-undeclared ones and "who is bringing them in?"

I think it might be useful to have a simple `:linked` sans right hand side lists be part of this PR so that we're at least not "throwing away" *that* information, or at minimum do as we've already done in file-roller here: https://github.com/Homebrew/homebrew-core/blob/7a9338a751f458ef3fff3759ab8d5b91c7d84a2c/Formula/file-roller.rb#L25-L31 with the `# Add linked-library dependencies` comment that you'd suggested at the time. If we do that, then the `:linked` could be added at a later time if/when deemed desirable if @xu-cheng's trigger finger is itchy to ship this right away.

MikeMcQuaid:
> I think it might be useful to have a simple :linked sans right hand side lists be part of this PR

Agreed 👍

xu-cheng:
> I think it might be useful to have a simple :linked sans right hand side lists be part of this PR

I won't have time to do this in this PR. And it would be a too large change for a single PR. So I think we can either ship this with `# Add linked-library dependencies` comment or wait for someone else to implement it.

MikeMcQuaid:
> I won't have time to do this in this PR.

It's just adjusting the output text to have `depends_on ... => :linked` and verifying that doesn't break anything to use that as a `depends_on` tag. It's OK if you don't have time now but I do think it should be changed before this included.

xu-cheng:
@MikeMcQuaid you may not realize that `depends_on "pango", :linked => "gtk+3"` is in fact in compatible with our current DSL. We current on support `depends_on <a value of string/symbol>` or `depends_on <a hash>` not `depends_on <a value>, <a hash>`

MikeMcQuaid:
> @MikeMcQuaid you may not realize that depends_on "pango", :linked => "gtk+3" is in fact incompatible with our current DSL. We currently support depends_on <a value of string/symbol> or depends_on <a hash> not depends_on <a value>, <a hash>

I hadn't checked either way but thanks 👍. In that case I think just `depends_on "pango" => :linked` is fine for now and we can investigate a better format later. In that case it should just be a matter of updating the text that's output by the `audit` to tell people what `depends_on` to add.

ilovezfs:
might be contributor friendly to literally print the blob of formula text we expect inserted

ilovezfs:
Does this seem redundant to anyone else?
```
iMac-TMP:Formula joe$ brew audit file-roller
file-roller:
  * Undeclared dependencies found:
      glib
      pango
      atk
      cairo
      gdk-pixbuf
      gettext
    
    Formulae are required to declare all linked dependencies.
    Please check whether these dependencies are indeed needed.
    If so, please add all missing dependencies to the formula with:
      depends_on "glib" => :linked
      depends_on "pango" => :linked
      depends_on "atk" => :linked
      depends_on "cairo" => :linked
      depends_on "gdk-pixbuf" => :linked
      depends_on "gettext" => :linked
Error: 1 problem in 1 formula
```

ilovezfs:
Also that order seems rather random. Should it be alphabetized or something or is there meaning to the order? Maybe other taps last, tool, perhaps using the same sort that brew list uses?

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -56,10 +58,42 @@ def check_python_framework_links(lib)
...
+    end
+
+    if checker.undeclared_deps?
+      s += <<-EOS.undent
+        Undeclared dependencies found:
+          #{checker.undeclared_deps * "\n          "}
+
+        Formulae are required to declare all linked dependencies.
+        Please check whether these dependencies are indeed needed.
+        If so, please add all missing dependencies to the formula with:
```
ilovezfs: s/Please check whether these dependencies are indeed needed.  If so, please add all missing dependencies to the formula with:/Please add all missing dependencies to the formula with/

xu-cheng: But there are cases that they are in fact accidentally to be picked up. In those case, they should be eliminated by altering the install method.

MikeMcQuaid: In that case the text should elaborate on how to identify if they are needed.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -56,10 +58,42 @@ def check_python_framework_links(lib)
...
+
+        It is likely the installation was broken.
+        Please check formula and make sure it is built from source.
+      EOS
+    end
+
+    if checker.undeclared_deps?
+      s += <<-EOS.undent
+        Undeclared dependencies found:
+          #{checker.undeclared_deps * "\n          "}
```
MikeMcQuaid: Agreed that this is redundant given the `depends_on` below

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
ilovezfs: There is also a problem here in that the brew audit output for contributors locally for these may differ significantly from the brew audit output that would be printed by CI. This is due to opportunistic linking locally if they have additional things installed that aren't installed during bottling. The only valid set will be the set printed by CI, but the PR template tells people to run brew audit --strict locally, which is going to lead to extra junk being thrown in.

ilovezfs: If contributors are going to pre-add these in their PRs (rather than waiting for CI to tell them what to add), the procedure for that should be for the contributor to run brew test-bot in a totally empty prefix, unless we want to be somehow (magically) policing this ourselves (which means us doing the test-bot runs locally since the CI results won't be available since the contributor already over-added, which means brew audit prints nothing and the extra junk makes it to master).

xu-cheng: > In that case the text should elaborate on how to identify if they are needed

There is no determined way to do it without any knowledge of the formula itself.

> There is also a problem here in that the brew audit output for contributors locally for these may differ significantly from the brew audit output that would be printed by CI. ...

This is actually what this PR tries to achieve that it is to expose such problem instead of offering a solution. Because the latter one is a case by case thing.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
xu-cheng: May I suggest to list in here rather than `depends_on` below(we could offer one example for the first formula). Because authors are required to check whether they are intended.

MikeMcQuaid: 👎 we should make it easy for authors to copy-paste these things.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
MikeMcQuaid: > There is no determined way to do it without any knowledge of the formula itself.

We need to provide some guidance otherwise the `audit` is not useful for our contributors.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
xu-cheng: Then I would prefer to keep both.

ilovezfs: for what purpose? parsing the text output?

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
ilovezfs: I think maybe we should limit this to being printed by audit only if it's being called from test-bot to avoid garbage.

MikeMcQuaid: @ilovezfs I think it's fine for now; we can reevaluate it if it encourages bad behaviour.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
MikeMcQuaid: 👎 to the duplication.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
ilovezfs: Which makes me think maybe this all still belongs in `brew linkage --test` :)

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
xu-cheng: > for what purpose? parsing the text output
For reminding that they should be checked instead of just adding them.

MikeMcQuaid: That's already explicitly stated by the test below.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
MikeMcQuaid: Explain how you should check if they are needed or remove the text saying people should check and instead rely on our current code review process (my preferred option).

xu-cheng: How? All just tell me what should I write in here?

MikeMcQuaid: Change the text to:

`Formulae are required to declare all linked dependencies. Please add all linked dependencies to the formula with:` and remove the `Undeclared dependencies found:` above.

ilovezfs: >we can reevaluate it if it encourages bad behaviour.

It's not bad behavior (unless you mean the software not the people). The contributors who are "following the rules" will run brew audit locally, and it will expressly tell them to add garbage. That exact same command run during CI would not have done so.

MikeMcQuaid: @ilovezfs Yeh, I meant the software. "Undesirable results" would be better. Let's see if they add garbage before assuming they will 😀.

ilovezfs: The obvious solution is to have CI run `brew audit --linkage` and NOT to put that option in the PR template.

ilovezfs: >@ilovezfs Yeh, I meant the software. "Undesirable results" would be better. Let's see if they add garbage before assuming they will :grinning:.

I was sort of hoping you were accusing the contributors of being naughty, since that's actually much funnier :)

ilovezfs:
This now looks much better to me:
```
iMac-TMP:local joe$ brew audit file-roller
file-roller:
  * Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "atk" => :linked
      depends_on "cairo" => :linked
      depends_on "gdk-pixbuf" => :linked
      depends_on "gettext" => :linked
      depends_on "glib" => :linked
      depends_on "pango" => :linked
Error: 1 problem in 1 formula
```

However, this looks like a bit of a "mess":
```
iMac-TMP:local joe$ brew uninstall glib
Uninstalling /usr/local/Cellar/glib/2.48.1... (425 files, 22.3M)
iMac-TMP:local joe$ brew audit file-roller
file-roller:
  * Broken dylib links found:
      /usr/local/opt/glib/lib/libgthread-2.0.0.dylib
      /usr/local/opt/glib/lib/libgio-2.0.0.dylib
      /usr/local/opt/glib/lib/libgobject-2.0.0.dylib
      /usr/local/opt/glib/lib/libglib-2.0.0.dylib
    
    It is likely the installation was broken.
    Please check formula and make sure it is built from source.
    Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "atk" => :linked
      depends_on "cairo" => :linked
      depends_on "gdk-pixbuf" => :linked
      depends_on "gettext" => :linked
      depends_on "pango" => :linked
Error: 1 problem in 1 formula
iMac-TMP:local joe$
```
And is that really only "one problem"? Seems to imply that adding the `depends_on`s is going to fix everything, but the glib issue is entirely separate and won't be fixed by that.

ilovezfs:
```
iMac-TMP:local joe$ brew audit file-roller
file-roller:
  * Broken dylib links found:
      /usr/local/opt/glib/lib/libgthread-2.0.0.dylib
      /usr/local/opt/glib/lib/libgio-2.0.0.dylib
      /usr/local/opt/glib/lib/libgobject-2.0.0.dylib
      /usr/local/opt/glib/lib/libglib-2.0.0.dylib
    
    It is likely the installation was broken.
    Please check formula and make sure it is built from source.
    
    Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "atk" => :linked
      depends_on "cairo" => :linked
      depends_on "gdk-pixbuf" => :linked
      depends_on "gettext" => :linked
      depends_on "pango" => :linked
Error: 1 problem in 1 formula

```
Getting better. Maybe it can be a separate `*` for `Undeclared dependencies` and register 2 problems?

xu-cheng:
> Maybe it can be a separate * for Undeclared dependencies and register 2 problems?

It would mean calling `audit_check_output` inside `check_linkage` which would be not fit with rest of code style. OR we split the check into two methods, but that would cause LinkerChecker run twice.

ilovezfs:
>It would mean calling audit_check_output inside check_linkage which would be not fit with rest of code style. 

I can live with that if it means it's easier for people to read the output ;)

ilovezfs:
```
iMac-TMP:Formula joe$ brew audit file-roller
iMac-TMP:Formula joe$ brew audit --strict file-roller
file-roller:
  * `desc` (line 3) should be put before `homepage` (line 2)
  * Broken dylib links found:
      /usr/local/opt/glib/lib/libgthread-2.0.0.dylib
      /usr/local/opt/glib/lib/libgio-2.0.0.dylib
      /usr/local/opt/glib/lib/libgobject-2.0.0.dylib
      /usr/local/opt/glib/lib/libglib-2.0.0.dylib
    
    It is likely the installation was broken.
    Please check formula and make sure it is built from source.
  * Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "atk" => :linked
      depends_on "cairo" => :linked
      depends_on "gdk-pixbuf" => :linked
      depends_on "gettext" => :linked
      depends_on "pango" => :linked
Error: 3 problems in 1 formula
```
This looks reasonable to me.

Only complaint is that the blank line above "It is likely the installation was broken" looks a little weird. Also I guess that text should say "is broken" not "was broken."

ilovezfs:
Also, I'm not sure what

```
Please check formula and make sure it is built from source.
```

means.

xu-cheng:
> Also, I'm not sure what
Please check formula and make sure it is built from source.
means.

Because other than deleting one of dependencies' installation, I think most common cause is the formula fetches binaries from upstream instead of building from source.

ilovezfs:
Do you mean bottles?

xu-cheng:
No, I mean formulae in `homebrew/bintary`

JCount:
If audit checks are run on a configured formula, rather than a bottle, the dependencies for the enabled options are identified by the linkage checker as not being declared in the formula, throwing an error.

For example:
mkvtoolnix 9.3.1 --with-gettext --with-lzo --with-qt5 ;
brew audit mkvtoolnix ;
gives https://gist.github.com/JCount/076ca82c969653f89f91933c1081b821

@ilovezfs @tdsmith both felt that this specific behavior was likely undesirable.

xu-cheng:
@JCount should be fixed now.

JCount:
Thanks!

DomT4:
I'm increasingly considering this to be a mistake outside of `--strict`.

I'm frankly fed up at the amount of my time this check is taking up & with the greatest respect for the work they do I'm frustrated that maintainers responsible for implementing this check are barely present around `homebrew/core` to deal with dancing around it like a slightly drunk turtle. It's easier to ignore it being a pain from a distance, I'd imagine.

IMO, this still isn't particularly ready to be something exposed to _every_ contributor. Xu's changes here made it better, but it's still frustrating, unclear & sometimes plain wrong.

UniqMartin:
FWIW, I think the second part of this check (namely the one that ultimately suggests to add lines like `depends_on "dog" => :linked`) should be completely reverted and formulae that had such lines added should be stripped of them (hopefully fairly painless and easily done).

There have been major concerns with this check and while it solves a problem, it does so in a way that had no consensus at all. (The only concern addressed was to make the output more actionable—something that IMO made things worse as there's no “one size fits all” solution for those *possibly* undeclared dependencies.) Yes, I realize I was initially in part a supporter of this check, but re-reading the following discussion made me realize that the pain far outweighs the benefits. Sorry I failed (due to lack of time) to chime in with a revised opinion later on!

MikeMcQuaid:
> There have been major concerns with this check and while it solves a problem, it does so in a way that had no consensus at all.

While I think we still need to tweak the way this is handled: I think not solving problems unless there is a consensus is a bit of a dangerous path to go down because it implies we let users suffer because we can't agree. I'd like to humbly propose that in this case (and in any others) that if we want it reverted you (not @UniqMartin specifically but anyone) are prepared to do the work yourself to work on an alternative (e.g. make PRs rather than issue comments). Perhaps that involves fixing `uses --recursive` but we need _something_ to handle this.

MikeMcQuaid:
Again the "you" below does not refer to any individual as it applies to me plenty of the time too:

On the flip side, there's a reason we say `There are no obligations, but we'd appreciate your continuing help in keeping on top of contributions.` when we invite people to be core contributors. It's pretty noticeable that there's folks who are maintainers but not following homebrew-core (or checking the backlog periodically) and this means this work falls back to other maintainers who are (who may prefer to be working on Homebrew/brew but have to do disproportionate work on it). If you're shipping something that affects formula contributors: you should be around to watch how they react to your changes and tweak things after the fact. This does not mean that you need to be available in the next minute/hour/day but if you're not on vacation and not noticing comments related to your changes a week later: you are not really following due diligence.

DomT4:
I can handle `:linked` _in theory_ but I think our practical implementation leaves a fair bit to be desired for now and it should be nested under `--strict` in the audit.

I can get over my gripes with `:linked` as an idea, which primarily revolve around it making formulae that link into heavy dependency trees _(Cough, GTK/GTK+3)_ a mess & us currently having no way to track when a dependency is listed as `:linked` but no link is present.

However, on a practical level right now I think it's unsuitable to be exposed to regular contributors & I'm tired of dealing with the fallout. I am grateful to maintainers for putting in the work on this PR & related ones; I'd like to spend more time with you than I do in the other place in an ideal world, I'm not ungrateful for the work here at all, I just think we need to refine the system before we expose it this widely.

MikeMcQuaid:
> However, on a practical level right now I think it's unsuitable to be exposed to regular contributors & I'm tired of dealing with the fallout.

As one of the main people dealing with that (and @ilovezfs who has also 👍d above): I think that's good enough to just make this `--strict` for now.

UniqMartin:
(Yes, I realize that most of what I'm replying to wasn't necessarily addressing me.)

> While I think we still need to tweak the way this is handled: I think not solving problems unless there is a consensus is a bit of a dangerous path to go down because it implies we let users suffer because we can't agree.

The current solution trades a possibly improved experience for our users with apparently major grief for maintainers (those active in homebrew/core) and contributors. I'm not sure that's the right balance in this particular case, not to mention that this adds `:linked` to our formula API despite it only being a marker for a workaround that we have no truly satisfying solution to yet. I'm not trying to block change/progress here; my subjective feeling is just that the change as-is didn't have enough support from those participating in the discussion in this PR.

(I guess I made a mistake by using the word “consensus” that represents a more fundamental level of agreement. I didn't mean to imply we always need a consensus for a change like this.)

> I'd like to humbly propose that in this case (and in any others) that if we want it reverted you (not @UniqMartin specifically but anyone) are prepared to do the work yourself to work on an alternative (e.g. make PRs rather than issue comments). Perhaps that involves fixing `uses --recursive` but we need _something_ to handle this.

I agree that expecting a PR that tweaks the behavior, e.g. only enable the check for `--strict`, instead of idle words in comment is preferable. But working on an alternative is a completely different thing and feels a bit unfair, as it sounds like someone gets to “break” something and someone who's annoyed/affected by the outfall of that change is forced to investigate and implement a better solution. Or am I missing the point?

> If you're shipping something that affects formula contributors: you should be around to watch how they react to your changes and tweak things after the fact.

Fully agreed.

> > However, on a practical level right now I think it's unsuitable to be exposed to regular contributors & I'm tired of dealing with the fallout.
>
> As one of the main people dealing with that (and @ilovezfs who has also 👍d above): I think that's good enough to just make this `--strict` for now.

I was close to submitting a PR for this when I realized: That would mainly address the problem of CI builds marked as failed due to this check. We're still asking contributors in our PR template to run `brew audit --strict --online`, so they would still be exposed to this undeclared dependency check and its output, if they accurately follow our PR check list. Is that tolerable? Should we hide the check behind a dedicated option like `--undeclared-deps`? Any other ideas?

ilovezfs:
Without getting into the politics of how decisions are made, work is distributed, or any of the rest of that, my current belief is that this change, including the `:linked` decorator and anywhere it has been used, should be rolled back entirely not simply confined to `--strict`. It was both theoretically and practically more sound as a non-fatal, informational output, only exposed via `brew linkage`, which requires `HOMEBREW_DEVELOPER` be set, and involved neither with `brew audit` (strict and non-strict alike), nor with `brew linkage --test`.

>dancing around it like a slightly drunk turtle

I still haven't stopped laughing about that.

MikeMcQuaid:
> I'm not trying to block change/progress here; my subjective feeling is just that the change as-is didn't have enough support from those participating in the discussion in this PR.

Yep, that's fair.

> (I guess I made a mistake by using the word “consensus” that represents a more fundamental level of agreement. I didn't mean to imply we always need a consensus for a change like this.)

👍 

> I agree that expecting a PR that tweaks the behavior, e.g. only enable the check for --strict, instead of idle words in comment is preferable. But working on an alternative is a completely different thing and feels a bit unfair, as it sounds like someone gets to “break” something and someone who's annoyed/affected by the outfall of that change is forced to investigate and implement a better solution. Or am I missing the point?

I think we agree here; I guess I mean even just a PR to revert the original commit (or delete the added code) is preferable to discussion on a merged PR.

> I was close to submitting a PR for this when I realized: That would mainly address the problem of CI builds marked as failed due to this check. We're still asking contributors in our PR template to run brew audit --strict --online, so they would still be exposed to this undeclared dependency check and its output, if they accurately follow our PR check list. Is that tolerable? Should we hide the check behind a dedicated option like --undeclared-deps? Any other ideas?

I think we can't/shouldn't rely on people necessarily running things that our CI doesn't. That said, I'd happily put this behind another `audit` flag that CI always runs (but I think from @DomT4's comments that is the main problem, here). I guess I'm hesitant to turn off something annoying that does catch/fix a problem when we don't have another solution to the problem or someone else 🙌 to work on it.

> It was both theoretically and practically more sound as a non-fatal, informational output, only exposed via brew linkage, which requires HOMEBREW_DEVELOPER be set, and involved neither with brew audit (strict and non-strict alike), nor with brew linkage --test.

I guess the main argument here is: does it help to have things that flag problems if users/maintainers never see them because CI is never red as a result?

It feels like the best solution is a fatal check that allows direct or indirect linkage (not just direct like we have currently) and for `brew uses  --recursive` to get fixed up (which, I believe, it will be just by us using Ruby 2).

UniqMartin:
> I think we agree here; I guess I mean even just a PR to revert the original commit (or delete the added code) is preferable to discussion on a merged PR.

Yes, I agree. Thanks for picking this up before I managed to set aside some time for that!

> It feels like the best solution is a fatal check that allows direct or indirect linkage (not just direct like we have currently) and […]

Agreed. If we could teach that check to take indirect dependencies into account when checking for apparently undeclared dependencies due to detected linkage, it would be more useful.

> […] and for `brew uses --recursive` to get fixed up (which, I believe, it will be just by us using Ruby 2).

Sadly, only using Ruby 2 will lessen the pain and make the performance of that more tolerable, but a proper fix and a more appropriate performance for this kind of operation probably involves rewriting significant parts of our dependency resolution code.

xu-cheng:
I'm deep sorry that such feature causes more trouble then it solves. :+1: on partially revert it. 

I think for long term, if we want to move bottle block outside formula to a separate file with many other meta information (like binaries list), we could consider to put linkage information there automatically by the bot.

PR #471, opened 2016-07-07
linkage: don't conflate broken and undeclared

ilovezfs:
See discussion in https://github.com/Homebrew/brew/commit/080ddd8804be14f4b18f9558b58270456ff313c2

Cf. https://github.com/Homebrew/brew/pull/470

### DIFF
--- a/Library/Homebrew/dev-cmd/linkage.rb
+++ b/Library/Homebrew/dev-cmd/linkage.rb
@@ -7,9 +7,8 @@
 # formulae.
 #
 # Options:
-#  --test      - testing version: only display broken libraries and possibly
-#                undeclared dependencies; exit non-zero if any breakage was
-#                found (i.e. any of these yield a nonempty result).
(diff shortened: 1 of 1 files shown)


ilovezfs:
It would be nice to have CI working again between now and whenever #470 is ready to ship.

UniqMartin:
I have some thoughts on the topic and while I read all the comments so far, I haven't found the time yet to comment myself. However, I agree with the assessment that `brew linkage --test` currently does too much and conflates two rather orthogonal checks (I didn't realize this when I reviewed and :+1:ed the PR that changed that).

:+1: on the revert until we have properly discussed and ironed out the desired semantics of those checks. The revert will allow us to have this discussion without the pressure to fix this ASAP.

DomT4:
I'm 👍 on the temporary revert as well, so we don't sprint into another change here without thorough discussion.

PR #478, opened 2016-07-09
Implement HeadVersions

vladshablinsky:

-----

# Description

This is a bunch of commits taken from https://github.com/Homebrew/brew/pull/274 to be reviewed and merged.

* Introduce `HeadVersion` subclass of `Version` to deal with `HEAD` versions.
* Use revisions for `HEAD` versions
* Use new format for `HEAD` kegs: `HEAD-<commit_hash>_<revision>`.
* Add tests

Any feedback appreciated.

cc @xu-cheng

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1326,7 +1326,7 @@ class Formula
       }
     end
 
-    hsh["installed"] = hsh["installed"].sort_by { |i| Version.new(i["version"]) }
+    hsh["installed"] = hsh["installed"].sort_by { |i| Version.create(i["version"]) }
 
     hsh
   end
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -287,6 +287,14 @@ class Formula
     active_spec.version
   end
 
+  def update_head_version
+    return unless head?
+    return unless head.downloader.is_a?(VCSDownloadStrategy)
+    return unless head.downloader.cached_location.exist?
+
+    head.version.update_commit(head.downloader.last_commit)
+  end
+
   # The {PkgVersion} for this formula with {version} and {#revision} information.
   def pkg_version
     PkgVersion.new(version, revision)
@@ -405,11 +413,23 @@ class Formula
     Pathname.new("#{HOMEBREW_LIBRARY}/LinkedKegs/#{name}")
   end
 
+  def latest_head_prefix
+    head_versions = installed_prefixes.map do |pn|
+      pn_pkgversion = PkgVersion.parse(pn.basename.to_s)
+      pn_pkgversion if pn_pkgversion.head?
+    end.compact
+
+    latest_head_version = head_versions.max_by do |pn_pkgversion|
+      [Tab.for_keg(prefix(pn_pkgversion)).source_modified_time, pn_pkgversion.revision]
+    end
+    prefix(latest_head_version) if latest_head_version
+  end
+
   # The latest prefix for this formula. Checks for {#head}, then {#devel}
   # and then {#stable}'s {#prefix}
   # @private
   def installed_prefix
-    if head && (head_prefix = prefix(PkgVersion.new(head.version, revision))).directory?
+    if head && (head_prefix = latest_head_prefix) && head_prefix.directory?
       head_prefix
     elsif devel && (devel_prefix = prefix(PkgVersion.new(devel.version, revision))).directory?
       devel_prefix
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1518,7 +1518,12 @@ class Formula
   def eligible_kegs_for_cleanup
     eligible_for_cleanup = []
     if installed?
-      eligible_kegs = installed_kegs.select { |k| pkg_version > k.version }
+      eligible_kegs = if head? && (head_prefix = latest_head_prefix)
+        installed_kegs - [Keg.new(head_prefix)]
+      else
+        installed_kegs.select { |k| pkg_version > k.version }
+      end
+
       if eligible_kegs.any?
         eligible_kegs.each do |keg|
           if keg.linked?
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -29,6 +29,7 @@ class SoftwareSpec
   def_delegators :@resource, :cached_download, :clear_cache
   def_delegators :@resource, :checksum, :mirrors, :specs, :using
   def_delegators :@resource, :version, :mirror, *Checksum::TYPES
+  def_delegators :@resource, :downloader
 
   def initialize
     @resource = Resource.new
@@ -203,7 +204,7 @@ end
 class HeadSoftwareSpec < SoftwareSpec
   def initialize
     super
-    @resource.version = Version.new("HEAD")
+    @resource.version = HeadVersion.new("HEAD")
   end
 
   def verify_download_integrity(_fn)
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -204,7 +204,7 @@ end
 class HeadSoftwareSpec < SoftwareSpec
   def initialize
     super
-    @resource.version = HeadVersion.new("HEAD")
+    @resource.version = Version.create("HEAD")
   end
 
   def verify_download_integrity(_fn)
--- a/Library/Homebrew/test/test_pkg_version.rb
+++ b/Library/Homebrew/test/test_pkg_version.rb
@@ -7,12 +7,12 @@ class PkgVersionTests < Homebrew::TestCase
   end
 
   def test_parse
-    assert_equal PkgVersion.new(Version.new("1.0"), 1), PkgVersion.parse("1.0_1")
-    assert_equal PkgVersion.new(Version.new("1.0"), 1), PkgVersion.parse("1.0_1")
-    assert_equal PkgVersion.new(Version.new("1.0"), 0), PkgVersion.parse("1.0")
-    assert_equal PkgVersion.new(Version.new("1.0"), 0), PkgVersion.parse("1.0_0")
-    assert_equal PkgVersion.new(Version.new("2.1.4"), 0), PkgVersion.parse("2.1.4_0")
-    assert_equal PkgVersion.new(Version.new("1.0.1e"), 1), PkgVersion.parse("1.0.1e_1")
+    assert_equal PkgVersion.new(Version.create("1.0"), 1), PkgVersion.parse("1.0_1")
+    assert_equal PkgVersion.new(Version.create("1.0"), 1), PkgVersion.parse("1.0_1")
+    assert_equal PkgVersion.new(Version.create("1.0"), 0), PkgVersion.parse("1.0")
+    assert_equal PkgVersion.new(Version.create("1.0"), 0), PkgVersion.parse("1.0_0")
+    assert_equal PkgVersion.new(Version.create("2.1.4"), 0), PkgVersion.parse("2.1.4_0")
+    assert_equal PkgVersion.new(Version.create("1.0.1e"), 1), PkgVersion.parse("1.0.1e_1")
   end
 
   def test_comparison
@@ -24,26 +24,26 @@ class PkgVersionTests < Homebrew::TestCase
     assert_operator v("HEAD"), :>, v("1.0")
     assert_operator v("1.0"), :<, v("HEAD")
 
-    v = PkgVersion.new(Version.new("1.0"), 0)
+    v = PkgVersion.new(Version.create("1.0"), 0)
     assert_nil v <=> Object.new
     assert_raises(ArgumentError) { v > Object.new }
-    assert_raises(ArgumentError) { v > Version.new("1.0") }
+    assert_raises(ArgumentError) { v > Version.create("1.0") }
   end
 
   def test_to_s
-    assert_equal "1.0", PkgVersion.new(Version.new("1.0"), 0).to_s
-    assert_equal "1.0_1", PkgVersion.new(Version.new("1.0"), 1).to_s
-    assert_equal "1.0", PkgVersion.new(Version.new("1.0"), 0).to_s
-    assert_equal "1.0", PkgVersion.new(Version.new("1.0"), 0).to_s
+    assert_equal "1.0", PkgVersion.new(Version.create("1.0"), 0).to_s
+    assert_equal "1.0_1", PkgVersion.new(Version.create("1.0"), 1).to_s
+    assert_equal "1.0", PkgVersion.new(Version.create("1.0"), 0).to_s
+    assert_equal "1.0", PkgVersion.new(Version.create("1.0"), 0).to_s
     assert_equal "HEAD_1", PkgVersion.new(Version.create("HEAD"), 1).to_s
     assert_equal "HEAD-ffffff_1", PkgVersion.new(Version.create("HEAD-ffffff"), 1).to_s
   end
 
   def test_hash
-    p1 = PkgVersion.new(Version.new("1.0"), 1)
-    p2 = PkgVersion.new(Version.new("1.0"), 1)
-    p3 = PkgVersion.new(Version.new("1.1"), 1)
-    p4 = PkgVersion.new(Version.new("1.0"), 0)
+    p1 = PkgVersion.new(Version.create("1.0"), 1)
+    p2 = PkgVersion.new(Version.create("1.0"), 1)
+    p3 = PkgVersion.new(Version.create("1.1"), 1)
+    p4 = PkgVersion.new(Version.create("1.0"), 0)
     assert_equal p1.hash, p2.hash
     refute_equal p1.hash, p3.hash
     refute_equal p1.hash, p4.hash
--- a/Library/Homebrew/test/test_pkg_version.rb
+++ b/Library/Homebrew/test/test_pkg_version.rb
@@ -35,7 +35,8 @@ class PkgVersionTests < Homebrew::TestCase
     assert_equal "1.0_1", PkgVersion.new(Version.new("1.0"), 1).to_s
     assert_equal "1.0", PkgVersion.new(Version.new("1.0"), 0).to_s
     assert_equal "1.0", PkgVersion.new(Version.new("1.0"), 0).to_s
-    assert_equal "HEAD", PkgVersion.new(Version.new("HEAD"), 1).to_s
+    assert_equal "HEAD_1", PkgVersion.new(Version.create("HEAD"), 1).to_s
+    assert_equal "HEAD-ffffff_1", PkgVersion.new(Version.create("HEAD-ffffff"), 1).to_s
   end
 
   def test_hash
--- a/Library/Homebrew/test/test_versions.rb
+++ b/Library/Homebrew/test/test_versions.rb
@@ -55,7 +55,11 @@ class VersionComparisonTests < Homebrew::TestCase
 
   def test_HEAD
     assert_operator version("HEAD"), :>, version("1.2.3")
+    assert_operator version("HEAD-abcdef"), :>, version("1.2.3")
     assert_operator version("1.2.3"), :<, version("HEAD")
+    assert_operator version("1.2.3"), :<, version("HEAD-fedcba")
+    assert_operator version("HEAD-abcdef"), :==, version("HEAD-fedcba")
+    assert_operator version("HEAD"), :==, version("HEAD-fedcba")
   end
 
   def test_comparing_alpha_versions
@@ -154,6 +158,12 @@ class VersionParsingTests < Homebrew::TestCase
   def test_no_version
     assert_version_nil "http://example.com/blah.tar"
     assert_version_nil "foo"
+  end
+
+  def test_create
+    v = Version.create("1.20")
+    refute_predicate v, :head?
+    assert_equal "1.20", v.to_str
   end
 
   def test_version_all_dots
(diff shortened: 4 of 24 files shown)


Library/Homebrew/test/test_versions.rb
```diff
@@ -54,8 +54,10 @@ def test_comparing_regular_versions
   end
 
   def test_HEAD
-    assert_operator version("HEAD"), :>, version("1.2.3")
-    assert_operator version("1.2.3"), :<, version("HEAD")
+    assert_operator head_version("HEAD"), :>, version("1.2.3")
+    assert_operator head_version("HEAD-abcdef"), :>, version("1.2.3")
```
MikeMcQuaid: Would be good to have comparisons between different HEAD versions here too.

MikeMcQuaid:
Could you also create a homebrew-core PR using this (if any formula changes are needed) so we can see what that looks like? Thanks!

vladshablinsky:
> Could you also create a homebrew-core PR using this (if any formula changes are needed) so we can see what that looks like? 

I'm not sure what you mean. To see what this looks like all you need is install formula which `head.downloader` is `VCSDownloadStrategy` with `--HEAD` flag.
Examples:
`ack` for `git`
`disktype` for `cvs`
`sdl_rtf` for `mercurial`


`HEAD`-versions upgrades are not touched in this PR.

Library/Homebrew/test/test_versions.rb
vladshablinsky: @MikeMcQuaid  :+1: Helped me to notice that I've forgotten to rewrite `Version` comparator a little, but have been relying on the behaviour when the comparator is rewritten. It doesn't change anything too much if it all worked without that comparator, but anyway.

I suppose that `HEAD` version equals to any other `HEAD` version. If we want to get latest `HEAD` prefix, we can do it on keg level, like I do that in `Formula#latest_head_prefix`.

MikeMcQuaid: > I suppose that HEAD version equals to any other HEAD version. If we want to get latest HEAD prefix, we can do it on keg level, like I do that in Formula#latest_head_prefix.

I think we should consider making the version comparison take into account `latest_head_prefix` (if it doesn't already) so that we can e.g. `brew upgrade` between HEAD versions. It might also not be a bad idea to mandate a `version` if e.g. `git describe --tags` doesn't return anything so we can compare between HEAD versions and normal versions (but obviously that doesn't need to be part of this PR).

vladshablinsky: The idea with `HEAD` versions is that if there is some upstream version that differs from one of installed `HEAD`s, then all of installed `HEAD`s are outdated. So we can compare `HEAD`-versions, but not using `HeadVersion#<=>` because to compare `HEAD`-versions inside `HeadVersion` we need more than just their versions strings, we need add `downloader` or `formula` to `HeadVersion`. 

> It might also not be a bad idea to mandate a version if e.g. git describe --tags doesn't return anything so we can compare between HEAD versions and normal versions (but obviously that doesn't need to be part of this PR).

In the original PR there is some logic added to the `Tab` which can allow us to compare `HEAD` versions with normal versions: tabs store latest `stable` and `devel` versions at the moment the keg is installed, so if `HEAD` keg's  `stable` is less than current stable version, then `HEAD`-keg is outdated.

MikeMcQuaid: > The idea with HEAD versions is that if there is some upstream version that differs from one of installed HEADs, then all of installed HEADs are outdated.

Cool 👍 



> In the original PR there is some logic added to the Tab which can allow us to compare HEAD versions with normal versions: tabs store latest stable and devel versions at the moment the keg is installed, so if HEAD keg's stable is less than current stable version, then HEAD-keg is outdated.

👍

Library/Homebrew/formula.rb
```diff
@@ -405,11 +414,21 @@ def linked_keg
...
   end
 
+  def latest_head_prefix
+    head_prefixes = installed_prefixes.select do |p|
+      p.basename.to_s.start_with?("HEAD")
+    end
+
+    head_prefixes.sort_by do |p|
+      Tab.for_keg(Keg.new(p)).source_modified_time
+    end.last
```
UniqMartin: You can use `max_by { … }` instead of `sort_by { … }.last` if you're not actually interested in the sorted array. (Sorting the array just to discard it afterwards is more expensive and makes the purpose of the code less obvious.)

Library/Homebrew/formula.rb
```diff
@@ -405,11 +414,21 @@ def linked_keg
     Pathname.new("#{HOMEBREW_LIBRARY}/LinkedKegs/#{name}")
   end
 
+  def latest_head_prefix
+    head_prefixes = installed_prefixes.select do |p|
+      p.basename.to_s.start_with?("HEAD")
+    end
+
+    head_prefixes.sort_by do |p|
+      Tab.for_keg(Keg.new(p)).source_modified_time
```
UniqMartin: Two remarks:

1. You can drop the intermediate `Keg.new` as `Tab.for_keg` also directly operates on `Pathname` instances. (This will also skip checks in `Keg.new`, but they shouldn't be necessary here.)

2. Please use `pn` instead of `p` when enumerating `Pathname` instances (here and in the code above). I'm not sure we've been consistent with this in the existing code, but `pn` makes it a bit clearer that it's a `Pathname` object.

Library/Homebrew/formula.rb
```diff
@@ -287,6 +287,15 @@ def version
...
   end
 
+  def update_head_version
+    return unless head?
+    head_resource = head.get_resource
+    if head_resource.downloader.is_a?(VCSDownloadStrategy) && \
+        head_resource.downloader.cached_location.exist?
+      head_resource.version.update_commit(head_resource.downloader.last_commit)
+    end
+  end
```
UniqMartin: I think this can be made a little bit more readable with additional early returns like this:

```ruby
def update_head_version
  return unless head?

  downloader = head.get_resource.downloader
  return unless downloader.is_a?(VCSDownloadStrategy)
  return unless downloader.cached_location.exist?

  head.get_resource.version.update_commit(downloader.last_commit)
end
```

Library/Homebrew/software_spec.rb
```diff
@@ -107,6 +108,10 @@ def resource(name, klass = Resource, &block)
     end
   end
 
+  def get_resource
+    @resource
+  end
```
UniqMartin: Attribute readers in Ruby don't normally have a `get_` prefix. Is there any reason we cannot name this `resource` and replace the entire method with a simple `attr_reader :resource`.

Library/Homebrew/version.rb
```diff
@@ -179,6 +179,18 @@ def self.detect(url, specs)
...
+    if val.respond_to?(:to_str)
+      if val.start_with?("HEAD")
+        HeadVersion.new(val)
+      else
+        Version.new(val)
+      end
+    else
+      raise TypeError, "Version value must be a string; got a #{val.class} (#{val})"
+    end
+  end
```
UniqMartin: This is purely a style issue, but here's a suggestion that is easier to read in my opinion:

```ruby
def self.create(val)
  unless val.respond_to?(:to_str)
    raise TypeError, "Version value must be a string; got a #{val.class} (#{val})"
  end

  if val.start_with?("HEAD")
    HeadVersion.new(val)
  else
    Version.new(val)
  end
end
```

An even more radical version would be to replace the final `if` statement with (but I'm not sure this is considered good style and would only adopt it if other maintainers agree):

```ruby
klass = val.start_with?("HEAD") ? HeadVersion : Version
klass.new(val)
```

Library/Homebrew/version.rb
```diff
@@ -379,3 +392,29 @@ def self._parse(spec)
...
+class HeadVersion < Version
+  attr_reader :commit
+
+  def initialize(val)
+    if val.respond_to?(:to_str)
+      @version = val.to_str
+    else
+      raise TypeError, "Version value must be a string; got a #{val.class} (#{val})"
+    end
+    @commit = @version[/HEAD-(.+)/, 1]
```
UniqMartin: Regular expressions are greedy by default, thus I can't think of a situation where this wouldn't already do the right thing. Nonetheless I'd prefer to be more explicit: `/^HEAD-(.+)$/`.

Library/Homebrew/version.rb
```diff
@@ -379,3 +392,29 @@ def self._parse(spec)
...
+  end
```
UniqMartin: This is purely a style issue, but here's a suggestion that is easier to read in my opinion (similarly to another such comment I made earlier):

```ruby
def initialize(val)
  unless val.respond_to?(:to_str)
    raise TypeError, "Version value must be a string; got a #{val.class} (#{val})"
  end

  @version = val.to_str
  @commit = @version[/HEAD-(.+)/, 1]
end
```

Library/Homebrew/test/testing_env.rb
```diff
@@ -26,6 +26,10 @@ def version(v)
       Version.new(v)
     end
 
+    def head_version(v)
+      HeadVersion.new(v)
+    end
```
UniqMartin: Can we avoid adding this method and instead modify the `version` method directly above to use `Version.create`?

Library/Homebrew/test/test_pkg_version.rb
```diff
@@ -35,7 +35,8 @@ def test_to_s
     assert_equal "1.0_1", PkgVersion.new(Version.new("1.0"), 1).to_s
     assert_equal "1.0", PkgVersion.new(Version.new("1.0"), 0).to_s
     assert_equal "1.0", PkgVersion.new(Version.new("1.0"), 0).to_s
-    assert_equal "HEAD", PkgVersion.new(Version.new("HEAD"), 1).to_s
+    assert_equal "HEAD_1", PkgVersion.new(Version.new("HEAD"), 1).to_s
+    assert_equal "HEAD-ffffff_1", PkgVersion.new(Version.new("HEAD-ffffff"), 1).to_s
```
UniqMartin: These new cases look like they should be using `Version.create`, not `Version.new`. Or am I misunderstanding the purpose of this test?

vladshablinsky: Absolutely! Thanks!

vladshablinsky: `HeadVersion.new` is also OK though.

UniqMartin:
I have left quite a few code comments, though nothing that looks like a major issue to me.

While looking at the diff, I noticed that `Version.new` is sometimes used where it probably shouldn't be. Now that we have two different version classes and `Version.create` to hide this implementation detail, I wonder whether we should substitute **all** instances of `Version.new` in the Homebrew code base with `Version.create` (except for the implementation of `Version.create` itself).

I personally think we should. How do others feel about this? My gut feeling is that if we leave things in a mixed state where we sometimes use `Version.new`, sometimes `HeadVersion.new`, and sometimes `Version.create`, we're bound to make mistakes and will be running into hard to debug issues soon.

Library/Homebrew/test/test_pkg_version.rb
UniqMartin: Yes, it is. But see my other code comments and my non-code comment https://github.com/Homebrew/brew/pull/478#issuecomment-231606435 below (in case you agree with that) for where I'm coming from and why I'm suggesting `Version.create`.

Library/Homebrew/software_spec.rb
vladshablinsky: Yes, there is a reason for that. `SoftwareSpec#resourse` already exists.
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/software_spec.rb#L99-L108

vladshablinsky: I've grepped where I use `get_resource`, it seems I don't need it at all anymore. Maybe at the first place there was more sense in it, but now I don't need to get `resource`, all I need is `downloader` and `version` and I can get them using delegators already defined in `SoftwareSpec`.

```ruby
  def_delegators :@resource, :version, :mirror, *Checksum::TYPES
  def_delegators :@resource, :downloader
```

vladshablinsky:
@UniqMartin I was able to substitute everything as you suggested this way (I'm sure there is a way to do it easier):

```bash
find . -type d \( -path ./test/vendor -o -path ./os \) -prune -o -name '*.rb' -exec  sed -i '' 's/HeadVersion.new/Version.create/g' {} ';'
find . -type d \( -path ./test/vendor -o -path ./os \) -prune -o -name '*.rb' -exec  sed -i '' 's/Version.new/Version.create/g' {} ';'
find . -type d \( -path ./test/vendor -o -path ./os \) -prune -o -name '*.rb' -exec  sed -i '' 's/PkgVersion.create/PkgVersion.new/g' {} ';'
find . -type d \( -path ./test/vendor -o -path ./os \) -prune -o -name '*.rb' -exec  sed -i '' 's/MacOS::Version.create/MacOS::Version.new/g' {} ';'
find . -type d \( -path ./test/vendor -o -path ./os \) -prune -o -name '*.rb' -exec  sed -i '' 's/OS::Mac::Version.create/OS::Mac::Version.new/g' {} ';'
```

Basically I substituted `Version.new` and `HeadVersion.new` with `Version.create` excluding `OS::Mac::Version.new`, stuff related to `os` and `Version.create` itself.

Running `brew tests` gave me only one failure, which was really my fault, because I called `start_with?` method on `val` but not on `val.to_str` inside `Version` constructor. The thing is that we don't need val to be a string, we only need it to respond to `to_str`.

```
  1) Failure:
VersionTests#test_accepts_objects_responding_to_to_str [/usr/local/Library/Homebrew/version.rb:187]:
unexpected invocation: #<Mock:0x7f96f42dfa18>.start_with?('HEAD')
satisfied expectations:
- allowed any number of times, not yet invoked: #<Mock:0x7f96f42dfa18>.to_str(any_parameters)


892 runs, 2210 assertions, 1 failures, 0 errors, 3 skips
```

Library/Homebrew/version.rb
```diff
@@ -379,3 +392,29 @@ def self._parse(spec)
     return m.captures.first unless m.nil?
   end
 end
+
+class HeadVersion < Version
+  attr_reader :commit
+
+  def initialize(val)
+    unless val.respond_to?(:to_str)
```
dunn: Since `HeadVersion` new is only used by `Version.create`, which already does the type-checking, is this needed?

vladshablinsky: I'm not sure. It seems that it's not needed, but it ensures that if someone calls `HeadVersion.new`, it'll fail if `val` doesn't respond to `to_str`.

How about this?
```ruby
def initialize(val)
  super
  @commit = @version[/^HEAD-(.+)$/, 1]
end
```

Also we can add  `private_class_method :new` to `HeadVersion` and don't do the `to_str` check, but again, I'm not sure this is a good idea.

dunn: >  it ensures that if someone calls HeadVersion.new, it'll fail if val doesn't respond to to_str.

Good point.

If you think it's a better idea to leave it as-is, that's fine.

Library/Homebrew/software_spec.rb
UniqMartin: Sorry for the noise! I completely overlooked the existing `resource` method, but given how widely used it is, my comment feels a bit silly in retrospect. Thanks for checking nonetheless and if we don't need it anymore, that's even better.

However, should we find out we still need it, I think it could be called `url_resource` or something similar (to make it clear that it is not some random resource but the one associated with the primary `url` statement of the formula).

vladshablinsky: > Thanks for checking nonetheless and if we don't need it anymore, that's even better.

Initially I wanted to call it `resource`, but eventually couldn't. :)

Library/Homebrew/version.rb
UniqMartin: Using the base class initializer is of course a good thing. :+1: I'm not overly bothered with the duplicate check, but I think we could fairly safely remove it in the constructors, as we no longer expect to call `[Head]Version.new` directly.

UniqMartin:
> I was able to substitute everything as you suggested this way (I'm sure there is a way to do it easier):

I was thinking about some Perl-style regular expressions and some negative lookbehind to exclude unwanted prefixes, but didn't get very far, to be honest. I think your approach is fine and as long as you checked every changed line before adding them to your commit, things should be just fine. (I reviewed the relevant commit and haven't noticed any wrong substitutions.)

I'm fairly happy with this PR as-is. How do other maintainers feel about this?

(You could maybe squash the “version: call start_with? on String object” commit into the commit that added the modified line, but otherwise I'm happy with the commit structure, too.)

xu-cheng:
@vladshablinsky One last thing before we ship this, we need to update the cleanup logic to clean up old HEAD installation. I'm thinking:
```diff
diff --git a/Library/Homebrew/formula.rb b/Library/Homebrew/formula.rb
index fa67ce9..d922384 100644
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1517,7 +1517,11 @@ class Formula
   def eligible_kegs_for_cleanup
     eligible_for_cleanup = []
     if installed?
-      eligible_kegs = installed_kegs.select { |k| pkg_version > k.version }
+      eligible_kegs = if pkg_version.head?
+        installed_kegs - [Keg.new(latest_head_prefix)]
+      else
+        installed_kegs.select { |k| pkg_version > k.version }
+      end
       if eligible_kegs.any?
         eligible_kegs.each do |keg|
           if keg.linked?
```

With `latest_head_prefix` from #274.

Library/Homebrew/formula.rb
```diff
@@ -405,11 +413,19 @@ def linked_keg
     Pathname.new("#{HOMEBREW_LIBRARY}/LinkedKegs/#{name}")
   end
 
+  def latest_head_prefix
+    head_prefixes = installed_prefixes.select do |pn|
+      pn.basename.to_s.start_with?("HEAD")
```
xu-cheng: We probably need to check revision as well

vladshablinsky: As far as I understood it helps us to detect if `HEAD` is outdated, so we don't use it's prefix in `installed_prefix` if `head?`.
Is it true that once revision is assigned to a formula it cannot decrease or be removed even if new version is bumped?

xu-cheng: The revision will be reset to zero once there is a stable version bump.

My previous thought is that if there is a revision bump during same stable version, we may have multi HEAD installations with the same source_modified_time. Therefore, we need to check revision number. Considering we will have check `head_prefix_up_to_date?`, I think we should be fine even the revision is reset to zero after stable version bump.

To make this method more readable, how about:
```ruby
def latest_head_prefix
  head_vesrions = {}
  installed_prefxies.each do |pn|
    pn_pkgversion = PkgVersion.parse(pn.basename)
    next unless pn_pkgversion.head?
    head_versions[pn_pkgversion] = Tab.for_keg(pn).source_modified_time
  end

  latest_source_modified_time = head_versions.values.max
  latest_head_version = head_versions.select do |v, t|
    v.revision == revision && t == latest_source_modfied_time 
  end.first
  prefix(latest_head_version) if latest_head_version
end
```

xu-cheng:
@vladshablinsky What do you think of cleanup I mentioned above? We need address that as well before shipping this PR.

vladshablinsky:
> What do you think of cleanup I mentioned above? We need address that as well before shipping this PR.

Working on it.

Also, we need to make `PkgVersion#revision` public to add
```
 pn_pkgversion = PkgVersion.parse(pn.basename)
pn_pkgversion.head? && pn_pkgversion.revision == revision
```
is it OK?

xu-cheng:
> we need to make PkgVersion#revision public

:+1:

Library/Homebrew/formula.rb
vladshablinsky: I finally got it. Yes, it seems to work. Please, correct me if I'm wrong: `head_prefix_up_to_date?(prefix)` for this logic is called in two cases: 

1. When it definitely returns false (revision haven't been skipped since that prefix had been installed, which implies that stable haven't been changed) 
2. When we skip revision to `0` and it increments until it's the same as `prefix` has, when they are the same, `head_prefix_up_to_date` can be called, but it'll return false.

vladshablinsky: @xu-cheng but we'll need different cleanup behaviour, because we don't know the reason why `latest_head_prefix` returns `nil` and whether it's because no head prefixes exist or it's because all of them are outdated. So we need either to check if there is at least one `HEAD` installed or make `latest_head_prefix` like this:

```ruby
def latest_head_prefix
  head_vesrions = Hash.new { |hash, key| hash[key] = [] }
  installed_prefxies.each do |pn|
    pn_pkgversion = PkgVersion.parse(pn.basename)
    next unless pn_pkgversion.head?
    head_versions[Tab.for_keg(pn).source_modified_time] << pn_pkgversion
  end

  latest_source_modified_time = head_versions.keys.max
  latest_head_version = head_versions[latest_source_modified_time].max_by do |pkg_version|
    pkg_version.revision
  end
  prefix(latest_head_version.first) if latest_head_version
end
```

xu-cheng: How about:
```ruby
def latest_head_prefix
  head_versions = installed_prefixes.map do |pn|
    pn_pkgversion = PkgVersion.parse(pn.basename)
    pn_pkgversion if pn_pkgversion.head?
  end.compact

  latest_head_version = head_versions.max_by  do | pn_pkgversion|
    [Tab.for_keg(prefix(pn_pkgversion)).source_modified_time,  pn_pkgversion.revision]
  end
  prefix(latest_head_version) if latest_head_version
end
```

xu-cheng: Or, we may consider to split it into two methods: `latest_head_version` and `latest_head_prefix`. Because we still need to check revision number inside `head_version_up_to_date?`

vladshablinsky: Then, for cleanup we can make this:

```ruby
eligible_kegs = if head? && (head_prefix = latest_head_prefix)
  if head_prefix_up_to_date?(head_prefx)
    installed_kegs - [Keg.new(head_prefix)]
  else
    installed_kegs
  end
else
  installed_kegs.select { |k| pkg_version > k.version }
end
```

xu-cheng: I don't think you need to check `head_prefix_up_to_date` inside `eligible_kegs`. Noted, if `head_prefix_up_to_date` returns false, `installed?`(one line above `eligible_kegs =`) will return false as well.

xu-cheng: FYI, we can ignore `head_prefix_up_to_date` for now in this PR.  It should be part of #274, after we merge this PR.

vladshablinsky: I also noticed this. I think we need to add "Tab stores stable/devel/HEAD" logic to this PR, because it's getting difficult to keep in mind both this branch and the other PR branch in mind. We began to discuss `head_prefix_up_to_date?` here and there is no such method at all on this branch. Or we should carefully think what we merge in this PR, which seems to be more difficult.

xu-cheng: I think for this PR, we just assume head prefix is always up to dated(just like homebrew has always assumed before this PR). Therefore, you don't need to consider `head_prefix_up_to_date` at all in this PR. As result, brew install/reinstall/cleanup should work fine. The only thing missing is brew upgrade, which should be targeted in the other PR.

vladshablinsky:
It seems everything is fixed according to the comments now. I also added/changed testes where necessary.

While making last changes I thought that it probably would be handy to have `setup_tab_for_prefix` in `testing_env`.

xu-cheng:
Good work @vladshablinsky Thanks everyone.

Issue #87, opened 2016-04-14
`brew update` reports spurious change for files moved within tap repo


apjanke:
If a formula file in a tap is moved between the allowable formula directories (`/`, `/Formula`, and `/HomebrewFormula`), it will be reported as changed by the next `brew update` that picks it up.

The reported change will look something like this.

```
$ brew update
Updated 1 tap (homebrew/versions).
==> New Formulae
homebrew/versions/allegro4
homebrew/versions/android-ndk-r10d
homebrew/versions/android-ndk-r10e
homebrew/versions/ansible14
homebrew/versions/ansible18
```

This blocks rearranging the standard taps to move formula files from the root dir to `Formula`, which is our current preferred organization.

Blocks homebrew/homebrew-boneyard#108.

xu-cheng:
I think a more serious blocker is the `FormulaVersions` won't work after the formula directory change. As the result, we may get wrong revision number for bottles.

mikemcquaid:
> I think a more serious blocker is the FormulaVersions won't work after the formula directory change. As the result, we may get wrong revision number for bottles.

Yup, we'll need to make sure all these sort of things work after renaming (they can be made to)

UniqMartin:
The `update-report`-related issue of reporting spurious changes has been addressed in #480.

MikeMcQuaid:
> The update-report-related issue of reporting spurious changes has been addressed in #480.

Sounds sufficient to me for now at least.

apjanke:
I tested this out with a dummy repo, moving a bunch of formulae from the root dir to `Formula/`, and it worked – no spurious messages. Looks like it's safe to reorganize taps now.

PR #480, opened 2016-07-09
update: Don't report formulae that are moved within a tap but not renamed

jawshooah:

-----

It's misleading to report a formula as "new" when it's only been moved from one directory to another within the same tap.

Ref: [caskroom/homebrew-cask#22669 (comment)](https://github.com/caskroom/homebrew-cask/pull/22669#issuecomment-231457315)

### DIFF
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -209,8 +209,12 @@ class Reporter
         end
         @report[:M] << tap.formula_file_to_name(src)
       when /^R\d{0,3}/
-        @report[:D] << tap.formula_file_to_name(src) if tap.formula_file?(src)
-        @report[:A] << tap.formula_file_to_name(dst) if tap.formula_file?(dst)
+        src_full_name = tap.formula_file_to_name(src)
+        dst_full_name = tap.formula_file_to_name(dst)
+        # Don't report formulae that are moved within a tap but not renamed
+        next if src_full_name == dst_full_name
+        @report[:D] << src_full_name if tap.formula_file?(src)
+        @report[:A] << dst_full_name if tap.formula_file?(dst)
       end
     end
 
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -209,12 +209,13 @@ class Reporter
         end
         @report[:M] << tap.formula_file_to_name(src)
       when /^R\d{0,3}/
+        next unless tap.formula_file?(dst)
         src_full_name = tap.formula_file_to_name(src)
         dst_full_name = tap.formula_file_to_name(dst)
         # Don't report formulae that are moved within a tap but not renamed
         next if src_full_name == dst_full_name
-        @report[:D] << src_full_name if tap.formula_file?(src)
-        @report[:A] << dst_full_name if tap.formula_file?(dst)
+        @report[:D] << src_full_name
+        @report[:A] << dst_full_name
       end
     end
 
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -209,7 +209,6 @@ class Reporter
         end
         @report[:M] << tap.formula_file_to_name(src)
       when /^R\d{0,3}/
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cmd/update-report.rb
```diff
@@ -209,8 +209,12 @@ def report
...
         @report[:M] << tap.formula_file_to_name(src)
       when /^R\d{0,3}/
-        @report[:D] << tap.formula_file_to_name(src) if tap.formula_file?(src)
-        @report[:A] << tap.formula_file_to_name(dst) if tap.formula_file?(dst)
+        src_full_name = tap.formula_file_to_name(src)
+        dst_full_name = tap.formula_file_to_name(dst)
+        # Don't report formulae that are moved within a tap but not renamed
+        next if src_full_name == dst_full_name
+        @report[:D] << src_full_name if tap.formula_file?(src)
+        @report[:A] << dst_full_name if tap.formula_file?(dst)
```
UniqMartin: There's a very subtle problem with these two lines. If formulae happen to be moved within a tap *and* a formula gets renamed before or afterwards (thus `src_full_name != dst_full_name`) and the update covers both operations, then the report will look something like this (I experimented a bit with my local copy of the `homebrew/gui` tap):

```
Updated 1 tap (homebrew/gui).
==> New Formulae
homebrew/gui/xchat2
```

But it should have looked something like this:

```
Updated 1 tap (homebrew/gui).
==> Renamed Formulae
homebrew/gui/xchat -> homebrew/gui/xchat2
```

The reason is that `if tap.formula_file?(src)` will always be `false` as it checks against the current `Tap#formula_dir`. It's admittedly a rare corner case, but this should solve it:

```ruby
if tap.formula_file?(dst)
  @report[:D] << src_full_name
  @report[:A] << dst_full_name
end
```

UniqMartin:
Thanks for taking a look and for this PR! I commented on a corner case, but otherwise it's looking good.

Unfortunately, other parts of Homebrew also assume that `Tap#formula_dir` never changes during the lifetime of a tap and thus break if formulae are relocated within a tap. See issue #87 for details. Do you maybe see yourself tackling these as well? (Would be awesome, but I certainly don't expect you to.)

Library/Homebrew/cmd/update-report.rb
jawshooah: Nice catch! Fixed in 2648506

UniqMartin: Thanks for finding an even more elegant solution and particularly for adding a nice test case!

Library/Homebrew/cmd/update-report.rb
```diff
@@ -209,8 +209,13 @@ def report
         end
         @report[:M] << tap.formula_file_to_name(src)
       when /^R\d{0,3}/
-        @report[:D] << tap.formula_file_to_name(src) if tap.formula_file?(src)
-        @report[:A] << tap.formula_file_to_name(dst) if tap.formula_file?(dst)
+        next unless tap.formula_file?(dst)
```
UniqMartin: I just realized this check can be dropped completely, as there's a check preceding the `case` statement that already makes sure that at least one of the modified paths is a formula file:

```ruby
next unless paths.any? { |p| tap.formula_file?(p) }
```

jawshooah: Got it, removed in 7195878.

UniqMartin:
Merged in 4abdeb7b5eb06acb633a9ac12d7c681d736a5286. Thank you for this contribution to Homebrew, @jawshooah! :tada:

Sorry for the long delay! I'd be still interested in your answer to the following question:

> Unfortunately, other parts of Homebrew also assume that `Tap#formula_dir` never changes during the lifetime of a tap and thus break if formulae are relocated within a tap. See issue #87 for details. Do you maybe see yourself tackling these as well?

jawshooah:
@UniqMartin No worries! Sorry for the delay in getting back to you. I haven't looked too deeply into `FormulaVersions` but it seems like a proper fix might require a deep dive into Git plumbing commands. I'll see about giving it a shot later this evening, but I can't make any promises.

PR #481, opened 2016-07-10
update: mimic doctor permissions advice

### LINKED ISSUE #476: "Error: /usr/local must be writable!" (Sierra 10.12 Beta (16A239j) )
Just an FYI, and to get some advice on a better way of solving this! When running `brew doctor` I get this error: 

`Warning: /usr/local is not writable.`

And when running 'brew update', I get this:

`Error: /usr/local must be writable!`

I "solved" the problem by running `sudo chown -R `whoami` /usr/local`. That seems to have cleared the issue up.

Thanks!
Ian

DomT4:

-----

We might as well be consistent about it, since it's not automatically obvious to people what the next step is after we've pointed out the directory isn't writable.

Fixes #476.

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -288,12 +288,21 @@ EOS
   # check permissions
   if [[ "$HOMEBREW_PREFIX" = "/usr/local" && ! -w /usr/local ]]
   then
-    odie "/usr/local must be writable!"
+    odie <<-EOS
+/usr/local is not writable. You should change the ownership
+and permissions of /usr/local back to your user account:
+  sudo chown -R \$(whoami) /usr/local
+EOS
   fi
 
   if [[ ! -w "$HOMEBREW_REPOSITORY" ]]
   then
-    odie "$HOMEBREW_REPOSITORY must be writable!"
+    odie <<-EOS
+$HOMEBREW_REPOSITORY is not writable. You should change the
+ownership and permissions of $HOMEBREW_REPOSITORY back to your
+user account:
+  sudo chown -R \$(whoami) $HOMEBREW_REPOSITORY
+EOS
   fi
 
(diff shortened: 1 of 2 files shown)


UniqMartin:
:+1:

Library/Homebrew/cmd/update.sh
```diff
@@ -288,12 +288,22 @@ EOS
   # check permissions
   if [[ "$HOMEBREW_PREFIX" = "/usr/local" && ! -w /usr/local ]]
   then
-    odie "/usr/local must be writable!"
+    odie <<-EOS
+/usr/local is not writable. You should probably change the
```
MikeMcQuaid: Let's drop the probably; anyone who knows enough to disagree won't blindly follow the instruction.

MikeMcQuaid:
One nit otherwise 👍

Library/Homebrew/cmd/update.sh
DomT4: Do you want it dropped from the doctor output as well, or just here?

MikeMcQuaid: May as well do both 👍

UniqMartin:
One advantage of the otherwise pretty useless previous output in `cmd/update.sh` was that it was shell-agnostic (as issue 528 just made me realize). :wink:

DomT4:
It's perhaps worth noting perhaps that we've historically said IRT to things like `fish` & other Bourne-incompatible shells that users are assumed to be familiar enough to translate bash/etc instructions to the necessary syntax.

I very much like _(and occasionally use)_ `fish`, but I don't think it captures a level of popularity to merit us sprinkling alternative `fish`-compatible syntax everywhere.

UniqMartin:
:+1: The Gist of my comment in that other issue was basically saying (or trying to say) the same.

PR #488, opened 2016-07-10
Go Language Tweaks.

DomT4:

-----

I'm so tired of having to catch this stuff in post or us leaving it there forever because nobody sees or bothers to not ignore the existing `oppo` warning.

Let's punch harder on the empty arrays when `HOMEBREW_DEVELOPER` is set so we can start killing code that isn't necessary. User impact should be limited given the `HOMEBREW_DEVELOPER` couching, mostly means we can catch this with CI more than anything.

You only need to use `require "language/go` if you're using `go_resource`s, which we'll be doing less & less over time as upstream pressure to adopt the vendor system comes to bear. That is the _only_ purpose of the Go language module in Homebrew. Let's insert an `audit` warning about using `require "language/go` unnecessarily.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -676,6 +676,10 @@ class FormulaAuditor
     if text =~ /def plist/ && text !~ /plist_options/
       problem "Please set plist_options when using a formula-defined plist."
     end
+
+    if text =~ %r{require "language/go"} && text !~ /go_resource/
+      problem "require \"language/go\" is unnecessary unless using `go_resource`s"
+    end
   end
 
   def audit_line(line, lineno)
--- a/Library/Homebrew/language/go.rb
+++ b/Library/Homebrew/language/go.rb
@@ -7,7 +7,13 @@ module Language
     # The resource names should be the import name of the package,
     # e.g. `resource "github.com/foo/bar"`
     def self.stage_deps(resources, target)
-      opoo "tried to stage empty resources array" if resources.empty?
+      if resources.empty?
+        if ARGV.homebrew_developer?
+          odie "tried to stage empty Language::Go resources array"
+        else
+          opoo "tried to stage empty Language::Go resources array"
+        end
+      end
       resources.grep(Resource::Go) { |resource| resource.stage(target) }
     end
   end
--- a/Library/Homebrew/test/test_language_go.rb
+++ b/Library/Homebrew/test/test_language_go.rb
@@ -5,7 +5,11 @@ require "language/go"
 
 class LanguageGoTests < Homebrew::TestCase
   def test_stage_deps_empty
-    Language::Go.expects(:opoo).once
+    if ARGV.homebrew_developer?
+      Language::Go.expects(:odie).once
+    else
+      Language::Go.expects(:opoo).once
+    end
     mktmpdir do |path|
       shutup { Language::Go.stage_deps [], path }
     end


MikeMcQuaid:
👍

Library/Homebrew/cmd/audit.rb
```diff
@@ -676,6 +676,10 @@ def audit_text
     if text =~ /def plist/ && text !~ /plist_options/
       problem "Please set plist_options when using a formula-defined plist."
     end
+
+    if text =~ %r{require "language/go"} && text !~ /go_resource/
```
UniqMartin: Feel free to ignore this, but I'm feeling like we shouldn't be using regular expressions where they aren't needed, even if that makes for slightly less uniform code. In this case I'd be advocating for `String#include?`. One issue is that it's easy to forget to escape some special characters in regular expressions (not the case here), another that (in theory) a simple substring search should be significantly cheaper.

Library/Homebrew/language/go.rb
```diff
@@ -7,7 +7,11 @@ module Go
     # The resource names should be the import name of the package,
     # e.g. `resource "github.com/foo/bar"`
     def self.stage_deps(resources, target)
-      opoo "tried to stage empty resources array" if resources.empty?
+      if resources.empty? && ENV["HOMEBREW_DEVELOPER"].nil?
```
UniqMartin: Use `!ARGV.homebrew_developer?` instead of directly querying `ENV`.

Library/Homebrew/language/go.rb
```diff
@@ -7,7 +7,11 @@ module Go
...
+        opoo "tried to stage empty Language::Go resources array"
+      elsif resources.empty?
+        odie "tried to stage empty Language::Go resources array"
+      end
```
UniqMartin: Let's add some nesting to avoid repeating the condition (which makes the code less obvious in my opinion):

```ruby
if resources.empty?
  if ARGV.homebrew_developer?
    odie "tried to stage empty Language::Go resources array"
  else
    opoo "tried to stage empty Language::Go resources array"
  end
end
```

This duplication makes me want to have a `opoo_for_mere_mortals_and_odie_for_developers` method (obviously with a better name than this). Maybe something like this (not available today), would make sense:

```ruby
opoo "tried to stage empty Language::Go resources array",
     :fatal => ARGV.homebrew_developer?
```

Library/Homebrew/test/test_language_go.rb
```diff
@@ -5,7 +5,11 @@
 
 class LanguageGoTests < Homebrew::TestCase
   def test_stage_deps_empty
-    Language::Go.expects(:opoo).once
+    if ENV["HOMEBREW_DEVELOPER"].nil?
```
UniqMartin: Use `!ARGV.homebrew_developer?` instead of directly querying `ENV`.

Library/Homebrew/cmd/audit.rb
DomT4: Heh, it's funny you mention this because we had almost the exact same discussion a fair while ago now. I'll hollowly echo my comments from then which were more or less that the idea sounds good but I'd prefer not to conflate that change with this PR. Less hollowly, I'm happy to push a fresh PR afterwards.

Library/Homebrew/language/go.rb
DomT4: I hate nested `if` arguments & find them a pain to visually parse quickly, so that was deliberately avoided, but I'm happy to adopt it if I'm in the minority there.

UniqMartin:
Definitely a desirable change, though I have some comments about the implementation (some ignorable nit picking, but also more substantial comments that I'd like to see addressed).

Library/Homebrew/test/test_language_go.rb
DomT4: Forgot that existed, to be honest 👍.

Library/Homebrew/cmd/audit.rb
UniqMartin: I know. That's why I included parts like “even if that makes for slightly less uniform code”. But really, feel free to ignore this. I just felt the urge to waste some time and express my feelings (again). :wink:

Library/Homebrew/language/go.rb
UniqMartin: Well, I hate having to parse what is basically the same condition twice, If I could have parsed it only once and then skipped the whole block. 😛

DomT4:
Changes pushed.

UniqMartin:
:+1:

Library/Homebrew/cmd/audit.rb
DomT4: It's 3am, so I could be in idiot mode. I probably am, to be fair. But `text.include?` is vomiting some errors for me in audit:
```ruby
if text.include?("def plist") && !text.include?("plist_options")
  problem "Please set plist_options when using a formula-defined plist."
end
```
```
~> brew audit a2ps
Error: undefined method `include?' for #<FormulaText:0x007ff6a21512d8>
Please report this bug:
/usr/local/Library/Homebrew/cmd/audit.rb:682:in `audit_text'
/usr/local/Library/Homebrew/cmd/audit.rb:1006:in `audit'
/usr/local/Library/Homebrew/cmd/audit.rb:69:in `block in audit'
/usr/local/Library/Homebrew/cmd/audit.rb:65:in `each'
/usr/local/Library/Homebrew/cmd/audit.rb:65:in `audit'
/usr/local/Library/brew.rb:87:in `<main>'
```

It works outside of `audit` though if I drop into `brew irb`:
```ruby
text = "blob"
text.include?("blob") && !text.include?("blos")
=> true
```

In which way am I missing something obvious here?

UniqMartin: Well, `text` is not a `String` instance (like in your `irb` example), but a `FormulaText` instance as you can learn from the backtrace. Unfortunately, the latter doesn't exactly behave like a plain string (though it implements `=~` and thus gives a good first impression), so my suggestion to use `include?` is somewhat moot …

PR #351, opened 2016-06-13
config: expose Git version and path

UniqMartin:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes?
- [x] Have you successfully run `brew tests` with your changes locally?

-----

A working and somewhat up-to-date Git is central to a working Homebrew installation. This exposes the current Git version and which `git` binary will actually be called by the `Library/ENV/scm/git` wrapper that is used for invoking Git in all of Homebrew. The lookup logic is a bit involved, so this adds a level of transparency difficult to achieve by asking users for their `PATH` and various other information from their environment. The lookup logic is (copied from #330, first match wins):

1. If `HOMEBREW_GIT` is set in the environment, try that.
2. If `GIT` is set in the environment, try that.
3. If the `git` formula is installed, try that.
4. Iterate over all `git` binaries in `PATH`, but exclude `/usr/bin/git` because it might be a non-functional popup stub if neither Xcode nor Command Line Tools are installed.
5. Try locating a `git` binary using `xcrun -find git`.
6. Try using `git` from Xcode if it is installed in `/Applications/Xcode.app`.
7. Try `/usr/bin/git` if all else failed and we're sure it's not a popup stub.
8. Give up.

This PR doesn't change (and doesn't intend to change) the lookup logic as suggested in https://github.com/Homebrew/brew/issues/330#issuecomment-225430965. That's something for a separate PR, though the `brew config` output added here and the supporting code may help make such a PR in the future.

(This would have tremendously helped with debugging the issue in #330 experienced by two users.)

PR #492, opened 2016-07-11
config: show system ruby for all system

xu-cheng:


Library/Homebrew/system_config.rb
```diff
@@ -136,6 +149,7 @@ def dump_verbose_config(f = $stdout)
       f.puts "Perl: #{describe_perl}"
       f.puts "Python: #{describe_python}"
       f.puts "Ruby: #{describe_ruby}"
+      f.puts "System Ruby: #{describe_system_ruby}"
```
MikeMcQuaid: This isn't the system Ruby, though, it's whatever Ruby is being used to execute Homebrew (which isn't necessarily the system one on non-OS X)

xu-cheng: I think system Ruby means the ruby Homebrew is using rather than OS system ruby.

MikeMcQuaid: Maybe name it `Homebrew Ruby:` instead?

xu-cheng: :+1:

Library/Homebrew/system_config.rb
```diff
@@ -85,6 +85,19 @@ def describe_ruby
       end
     end
 
+    def describe_system_ruby
+      s = ""
+      case RUBY_VERSION
+      when /^1\.[89]/, /^2\.0/
```
MikeMcQuaid: Some duplication here with the os/mac version above.

xu-cheng: What do you suggest to remove duplication? A new method like `describe_homebrew_ruby_version`?

Library/Homebrew/system_config.rb
xu-cheng: Also, I think it may be a good idea to rename `Git` to `Homebrew Git` as they are in the same category, i.e., tools used by Homebrew itself.

Library/Homebrew/system_config.rb
MikeMcQuaid: We've used the `generic_` pattern before.

Library/Homebrew/system_config.rb
MikeMcQuaid: The main difference is that Git uses the first one in the PATH and the Homebrew Ruby is differentiating from the Ruby above it by the fact it may not be the first in the PATH.

Library/Homebrew/system_config.rb
xu-cheng: I'm not sure how to use such pattern.

Library/Homebrew/system_config.rb
xu-cheng: OK

Library/Homebrew/system_config.rb
MikeMcQuaid: Look at e.g. `dump_generic_verbose_config` or `grep` for `def generic_` and you'll see how they are used. In short: define a generic method in the main class, call it from the `os` class.

xu-cheng: But currently for `describe_system_ruby`, the generic version and mac version only shares a partial logic. So I'm not sure how to make it work. Unless we want some thing like:
```ruby
def describe_system_ruby
  s = generic_describe_system_ruby
  if RUBY_PATH.to_s =~ %r{^/System/Library/Frameworks/Ruby.framework/Versions/[12]\.[089]/usr/bin/ruby}
    s.gsub! " => #{RUBY_PATH}", ""
  end
  s
end
```

Which I don't think it gets better.

MikeMcQuaid: You could pull just the `case` into a function

xu-cheng:
Updated.

MikeMcQuaid:
👍

Library/Homebrew/system_config.rb
UniqMartin: In case of Git it's only using the first in PATH to a very coarse approximation. Other than that, the selection logic is a bit more involved: See top comment of #351 for details.

Library/Homebrew/extend/os/mac/system_config.rb
```diff
@@ -28,14 +28,8 @@ def describe_xquartz
...
-    def describe_system_ruby
-      s = ""
-      case RUBY_VERSION
-      when /^1\.[89]/, /^2\.0/
-        s << "#{RUBY_VERSION}-p#{RUBY_PATCHLEVEL}"
-      else
-        s << RUBY_VERSION
-      end
+    def describe_homebrew_ruby
+      s = describe_homebrew_ruby_version
```
lembacon: @xu-cheng `describe_homebrew_ruby_version` will possibly return a frozen string. The following lines may conditionally modify `s`. Adding a `.dup` would be great. :-)

xu-cheng: Fixed in https://github.com/Homebrew/brew/commit/99e9a22b851c027f286247ad37fb81d468c201e7

Thanks @lembacon

PR #495, opened 2016-07-12
audit: update ruby dependency advice

DomT4:

-----

Should be non-controversial but here's a PR to give people time to object if they wish to.

Simple enough change to reflect that we'll accept `:ruby` as a dependency but not `"ruby"` and a brief explanation as to why & how to use the new(ish) syntax.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -363,10 +363,16 @@ class FormulaAuditor
           problem "Don't use git as a dependency"
         when "mercurial"
           problem "Use `depends_on :hg` instead of `depends_on 'mercurial'`"
-        when "ruby"
-          problem "Don't use ruby as a dependency. We allow non-Homebrew ruby installations."
         when "gfortran"
           problem "Use `depends_on :fortran` instead of `depends_on 'gfortran'`"
+        when "ruby"
+          problem <<-EOS.undent
+            Don't use "ruby" as a dependency. If this formula requires a
+            minimum Ruby version not provided by the system you should
+            use the RubyRequirement:
+              depends_on :ruby => "1.8"
+            where "1.8" is the minimum version of Ruby required.
+          EOS
         when "open-mpi", "mpich"
           problem <<-EOS.undent
             There are multiple conflicting ways to install MPI. Use an MPIRequirement:


Library/Homebrew/cmd/audit.rb
```diff
@@ -363,10 +363,16 @@ def audit_deps
...
-          problem "Don't use ruby as a dependency. We allow non-Homebrew ruby installations."
         when "gfortran"
           problem "Use `depends_on :fortran` instead of `depends_on 'gfortran'`"
+        when "ruby"
+          problem <<-EOS.undent
+            Don't use "ruby" as a dependency. If this formula requires a
+            minimum Ruby version not provided by the system you should
+            use the RubyRequirement:
+              depends_on :ruby => "1.8"
+            Where "1.8" is the minimum version of Ruby required.
```
MikeMcQuaid: Drop the capital `w`

MikeMcQuaid:
👍 really clear messaging

Library/Homebrew/cmd/audit.rb
```diff
@@ -363,10 +363,16 @@ def audit_deps
           problem "Don't use git as a dependency"
         when "mercurial"
           problem "Use `depends_on :hg` instead of `depends_on 'mercurial'`"
-        when "ruby"
-          problem "Don't use ruby as a dependency. We allow non-Homebrew ruby installations."
         when "gfortran"
           problem "Use `depends_on :fortran` instead of `depends_on 'gfortran'`"
+        when "ruby"
```
UniqMartin: Why the move? It's not liked you sorted things alphabetically. :wink:

Library/Homebrew/cmd/audit.rb
```diff
@@ -363,10 +363,16 @@ def audit_deps
...
+          problem <<-EOS.undent
+            Don't use "ruby" as a dependency. If this formula requires a
+            minimum Ruby version not provided by the system you should
+            use the RubyRequirement:
```
UniqMartin: I wonder how many people not familiar with our implementation will be confused by this part of the sentence (and fail to recognized that `:ruby` and `RubyRequirement` are related).

UniqMartin:
:+1: Much better message and actually reflects the reality of our Ruby handling.

Library/Homebrew/cmd/audit.rb
DomT4: I wanted to blob the two bigger `EOS` elements together so we didn't have a short one, a bigger one, a short one, a bigger one, and so on. Anal, I'm aware 😄.

Library/Homebrew/cmd/audit.rb
DomT4: will do 👍

Library/Homebrew/cmd/audit.rb
DomT4: Some, probably. I wasn't overly keen on including the explicit `RubyRequirement` mention but we've been doing so for `open-mpi` for a while & nobody's publicly gone _"Huh?"_.

Library/Homebrew/cmd/audit.rb
UniqMartin: And completely irrational, if you'd ask me. (Or just an argument that doesn't work given how my brain is wired.) :wink: But I can deal with that.

Library/Homebrew/cmd/audit.rb
UniqMartin: I don't have a good alternative I can suggest. Just wanted to see if you had thought about this and maybe had a clever idea how to rephrase that. I guess we can easily deal with the few cases that will be confused by that, but most will probably get the idea.

Library/Homebrew/cmd/audit.rb
DomT4: Hey, I've never claimed to be rational 😉.

Library/Homebrew/cmd/audit.rb
DomT4: I can't wrap my head around an easy way to necessarily explain `:blah` consistently relates to special Requirement magic sadly. I'm not sure it's an easy concept to grasp if you're unfamiliar with Homebrew's codebase. We've seen enough examples of people doing things like:

```ruby
depends_on "ruby" => "1.8"
```
Before being corrected.

PR #496, opened 2016-07-12
ENV/scm/git: rewrite in Bash

xu-cheng:


Library/ENV/scm/git
```diff
@@ -1,71 +1,113 @@
...
-  exec path, *ARGV unless path == "/usr/bin/#{F}"
-end
-
-popup_stub = false
-if executable? "/usr/bin/xcode-select"
+  fi
+  exec "$@"
+}
+
+F="${0##*/}"
```
MikeMcQuaid: I know it's not your issue but a nicer variable name would be great here.

Library/ENV/scm/git
```diff
@@ -1,71 +1,113 @@
...
+D="$(quiet_cd "${0%/*}/" && pwd -P)"
```
MikeMcQuaid: Again: I know it's not your issue but a nicer variable name would be great here.

MikeMcQuaid:
Wow, nice turnaround on this 👍. Looks good to me. Does `shellcheck` seem OK? If so and if you've tested it: :shipit:

Library/ENV/scm/git
xu-cheng: Would you mind to suggest a better name as naming is hard 😂

Library/ENV/scm/git
MikeMcQuaid: `FILENAME`?

Library/ENV/scm/git
MikeMcQuaid: `SELF_DIRECTORY`?

Library/ENV/scm/git
xu-cheng: How about `SHIMS_FILE` and SHIMS_DIR` or `SCM_FILE` and `SCM_DIR`? `FILENAME` itself is too vague.

MikeMcQuaid: `SCM_FILE` 👍

PR #398, opened 2016-06-24
tests: add cmd/link, cmd/unlink integration test

eirinikos:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

This change adds test coverage for `brew link` and `brew unlink`, though [these lines of `cmd/link.rb`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/link.rb#L53-L55) still lack coverage (I haven't yet figured out how the `LinkError` gets raised).

[These lines](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/link.rb#L53-L55) also lack coverage, though somehow they *were* tested when I tried using a combination of `test/testball.rb`+ `setup_test_formula "baz"` (instead of only `setup_test_formula "testball"`). Since there are three other `ensure` blocks that do the following, I was hoping that the `cmd`s could be refactored into their own method):
```
cmd("uninstall", "--force", testball)
cmd("cleanup", "--force", "--prune=all")
```

But I wanted to test a `keg_only` formula and I'm not sure if there's a way to modify the existing `test/testball` file to make it such.

PR #500, opened 2016-07-12
tests: add cmd/test integration test

eirinikos:

-----

This change adds test coverage for `brew test foo`. It also incorporates @UniqMartin's [suggestion](https://github.com/Homebrew/brew/pull/398/files#r70382703) to use `/^testball/` in the `case` statement. 🎈 

[These lines](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/test.rb#L65-L85) remain untested, though. Is it feasible to test a formula within the test suite? It looks like using Minitest assertions is not an option, because when doing so, I receive this error:

``undefined method `assertions' for #<Formulary::FormulaNamespace[...]::TestballCopy:[...]>``

### DIFF
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -773,4 +773,31 @@ class IntegrationCommandTests < Homebrew::TestCase
     assert_match "testball does not have a version \"0.3\"",
       cmd_fail("switch", "testball", "0.3")
   end
+
+  def test_test_formula
+    assert_match "This command requires a formula argument", cmd_fail("test")
+    assert_match "Testing requires the latest version of testball",
+      cmd_fail("test", testball)
+
+    cmd("install", testball)
+    assert_match "testball defines no test", cmd_fail("test", testball)
+
+    setup_test_formula "testball_copy", <<-EOS.undent
+      head "https://github.com/example/testball2.git"
+
+      devel do
+        url "file://#{File.expand_path("..", __FILE__)}/tarballs/testball-0.1.tbz"
+        sha256 "#{TESTBALL_SHA256}"
+      end
+
+      keg_only "just because"
+
+      test do
+      end
+    EOS
+
+    cmd("install", "testball_copy")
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Great work. I'm happy as-is but @UniqMartin may have other review.

> These lines remain untested, though. Is it feasible to test a formula within the test suite? It looks like using Minitest assertions is not an option, because when doing so, I receive this error:
> 
> undefined method `assertions' for #<Formulary::FormulaNamespace[...]::TestballCopy:[...]>

I also saw this before. Any thoughts on that @uniqmartin?

UniqMartin:
Sorry, will provide a review as soon as possible (probably later tonight). Meanwhile, this PR could be rebased on top of an up-to-date `master` branch as the `/^testball/`-related changes already went in when #398 was merged and there's presently a merge conflict.

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:
Looks like one of the tests is failing (but I'm not sure why): https://bot.brew.sh/job/Brew%20Pull%20Requests/1190/version=yosemite/testReport/junit/brew-test-bot/yosemite/tests___no_compat/

All passing for you locally, I assume?

UniqMartin:
This failure is related to some of the work to make HEAD installations upgradable, thus @vladshablinsky might have some input. I suspect that the method that is trying to access `HeadVersion#commit` should properly check whether it is actually dealing with a HEAD version, particularly in cases where the installed formula version doesn't really match what is claimed in the argument list (`--HEAD` in this case).

Independently of this, there's also a problem in the test code added here in that it first installs a stable version of a formula and then tries to run `brew test --HEAD` which only really makes sense if the HEAD version was installed previously.

The failure can be observed with a simple `brew tests --only=integration_cmds/test_formula` if this branch is rebased/merged on top of an up-to-date `master`.

vladshablinsky:
Thanks for finding a bug @eirinikos and @UniqMartin, seems to be fixed in (https://github.com/Homebrew/brew/issues/644).

MikeMcQuaid:
Thanks again @eirinikos! Tests were still failing but just for `brew cask list`.

PR #501, opened 2016-07-12
tests: add cmd/irb integration test

eirinikos:

-----

This change adds testing for `brew irb`. Several lines remain untested, though. I haven't yet figured out a way to quit the `IRB` shell from within the test... 😞

### DIFF
--- a/Library/Homebrew/test/lib/config.rb
+++ b/Library/Homebrew/test/lib/config.rb
@@ -2,10 +2,9 @@ require "tmpdir"
 require "pathname"
 
 HOMEBREW_BREW_FILE = Pathname.new(ENV["HOMEBREW_BREW_FILE"])
-HOMEBREW_TEMP = Pathname.new(ENV["HOMEBREW_TEMP"] || Dir.tmpdir)
 
 TEST_TMPDIR = ENV.fetch("HOMEBREW_TEST_TMPDIR") { |k|
-  dir = Dir.mktmpdir("homebrew_tests", HOMEBREW_TEMP)
+  dir = Dir.mktmpdir("homebrew-tests-", ENV["HOMEBREW_TEMP"] || "/tmp")
   at_exit { FileUtils.remove_entry(dir) }
   ENV[k] = dir
 }
@@ -24,6 +23,7 @@ HOMEBREW_CACHE_FORMULA = HOMEBREW_PREFIX.parent+"formula_cache"
 HOMEBREW_LOCK_DIR      = HOMEBREW_PREFIX.parent+"locks"
 HOMEBREW_CELLAR        = HOMEBREW_PREFIX.parent+"cellar"
 HOMEBREW_LOGS          = HOMEBREW_PREFIX.parent+"logs"
+HOMEBREW_TEMP          = HOMEBREW_PREFIX.parent+"temp"
 
 TESTBALL_SHA1 = "be478fd8a80fe7f29196d6400326ac91dad68c37"
 TESTBALL_SHA256 = "91e3f7930c98d7ccfb288e115ed52d06b0e5bc16fec7dce8bdda86530027067b"
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -18,6 +18,7 @@ class IntegrationCommandTests < Homebrew::TestCase
       HOMEBREW_CACHE.children,
       HOMEBREW_LOCK_DIR.children,
       HOMEBREW_LOGS.children,
+      HOMEBREW_TEMP.children,
       HOMEBREW_PREFIX/"bin",
       HOMEBREW_PREFIX/"share",
       HOMEBREW_PREFIX/"opt",
@@ -688,5 +689,21 @@ class IntegrationCommandTests < Homebrew::TestCase
 
     cmd("install", "testball")
     assert_match "Would remove", cmd("unlink", "--dry-run", "testball")
+  end
+
+  def test_irb
+    assert_match "'v8'.f # => instance of the v8 formula",
+      cmd("irb", "--examples")
+
+    setup_test_formula "testball"
+
+    irb_test = HOMEBREW_TEMP/"irb-test.rb"
+    irb_test.write <<-EOS.undent
+      "testball".f
+      :testball.f
+      exit
+    EOS
+
+    assert_match "Interactive Homebrew Shell", cmd("irb", irb_test)
   end
 end
--- a/Library/Homebrew/test/testing_env.rb
+++ b/Library/Homebrew/test/testing_env.rb
@@ -7,7 +7,7 @@ require "formulary"
 
 # Test environment setup
 (HOMEBREW_LIBRARY/"Taps/homebrew/homebrew-core/Formula").mkpath
-%w[cache formula_cache locks cellar logs].each { |d| HOMEBREW_PREFIX.parent.join(d).mkpath }
+%w[cache formula_cache locks cellar logs temp].each { |d| HOMEBREW_PREFIX.parent.join(d).mkpath }
 
 # Test fixtures and files can be found relative to this path
 TEST_DIRECTORY = File.dirname(File.expand_path(__FILE__))


UniqMartin:
> Several lines remain untested, though. I haven't yet figured out a way to quit the IRB shell from within the test...

Should be doable if you invoke `brew irb` (which with very minor modifications behaves just like plain `irb`—check out the man page of the latter) and provide a Ruby script file as an argument, e.g. this works for me locally:

```console
$ cat irb-test.rb
puts "Hello World!"
exit
$ brew irb irb-test.rb
irb-test.rb(main):001:0> puts "Hello World!"
Hello World!
=> nil
irb-test.rb(main):002:0> exit
```

eirinikos:
Wow! Thanks @UniqMartin, I didn't know that `irb` takes Ruby script files as arguments.

I've written `irb-test.rb` to `HOMEBREW_CACHE` to take advantage of the existing `teardown` method. Does that make sense? Or would it be better to have the file live elsewhere (whether or not it's a path that's included in `teardown`)?

UniqMartin:
> Wow! Thanks @UniqMartin, I didn't know that `irb` takes Ruby script files as arguments.

Funny thing is, I didn't know either. The only thing I knew was that it is possible to send commands to `irb` via a pipe or standard input, e.g. `echo exit | irb` or `irb < irb-test.rb`. Fortunately, `irb` has proper documentation and `brew irb` is just a very shallow wrapper for it.

> I've written `irb-test.rb` to `HOMEBREW_CACHE` to take advantage of the existing `teardown` method. Does that make sense? Or would it be better to have the file live elsewhere (whether or not it's a path that's included in `teardown`)?

It's okay-ish to use `HOMEBREW_CACHE` for that, but this case makes me want to have a dedicated scratch space that will also be cleaned up on teardown and can be used for stuff, that doesn't really belong in any of the predefined `HOMEBREW_*` directories.

eirinikos:
> It's okay-ish to use `HOMEBREW_CACHE` for that, but this case makes me want to have a dedicated scratch space that will also be cleaned up on teardown and can be used for stuff, that doesn't really belong in any of the predefined `HOMEBREW_*` directories.

Makes sense. I've been meaning to follow-up with you about an [idea](https://github.com/Homebrew/brew/pull/439#issue-163567527) you'd mentioned in the last week or two -- nesting `HOMEBREW_TEMP` within `TEST_TMPDIR`. I mentioned this to @MikeMcQuaid and he said that nesting `HOMEBREW_TEMP` sounded like a good idea.

Maybe this `irb-test.rb` file could go in there (once we have changed `test/lib/config.rb` to reflect the nesting)?

I'm guessing that the definition for `HOMEBREW_TEMP` would become something like  `HOMEBREW_PREFIX.parent + "temp"`.

MikeMcQuaid:
> Makes sense. I've been meaning to follow-up with you about an idea you'd mentioned in the last week or two -- nesting HOMEBREW_TEMP within TEST_TMPDIR. I mentioned this to @MikeMcQuaid and he said that nesting HOMEBREW_TEMP sounded like a good idea.
> 
> Maybe this irb-test.rb file could go in there (once we have changed test/lib/config.rb to reflect the nesting)?
> 
> I'm guessing that the definition for HOMEBREW_TEMP would become something like HOMEBREW_PREFIX.parent + "temp".

Was thinking the same thing. All sounds good to me 👍

eirinikos:
Looks like it's not just a simple matter of:
- removing [this line](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/lib/config.rb#L5) in `test/lib/config.rb`
- adding `HOMEBREW_TEMP = HOMEBREW_PREFIX.parent+"temp"` to the end of the list of similar paths
- making a simple modification to [this line](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/lib/config.rb#L8) in the definition for `TEST_TMPDIR` 😞 

It seems like the redefinition of `TEST_TMPDIR` is what eludes me.
My attempts to modify `dir` to forgo `HOMEBREW_TEMP` and then running `brew tests` have led to a multitude of errors saying `No such file or directory` -- a total of 113 errors and 13 failures for the test suite. Looks like it has to do with how `mktmpdir` is defined, but I haven't found a way forward.

Another question -- I've tried leaving `HOMEBREW_TEMP` and `TEST_TMPDIR` intact and simply adding and using an arbitrary path like `HOMEBREW_SCRATCH = HOMEBREW_PREFIX.parent+"scratch"`.

Adding `HOMEBREW_SCRATCH.children` to `paths_to_delete` in the `teardown` method does not prevent a file leak (`HOMEBREW_SCRATCH` itself needs to be added). What explains the fact that a directory like `HOMEBREW_CACHE` or `HOMEBREW_LOGS` can persist through the test suite (while `HOMEBREW_SCRATCH` cannot)?

Thanks!

UniqMartin:
> Looks like it's not just a simple matter of:
> - removing [this line](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/lib/config.rb#L5) in `test/lib/config.rb`
> - adding `HOMEBREW_TEMP = HOMEBREW_PREFIX.parent+"temp"` to the end of the list of similar paths
> - making a simple modification to [this line](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/lib/config.rb#L8) in the definition for `TEST_TMPDIR` 😞 
>
> It seems like the redefinition of `TEST_TMPDIR` is what eludes me.
My attempts to modify `dir` to forgo `HOMEBREW_TEMP` and then running `brew tests` have led to a multitude of errors saying `No such file or directory` -- a total of 113 errors and 13 failures for the test suite. Looks like it has to do with how `mktmpdir` is defined, but I haven't found a way forward.

All of the steps you've taken sound right to me, though of course I'm not 100% sure about the exact modifications you've made. Can you post the exact changes you've made when trying this? You could for example make those changes in a new separate branch, push it to your GitHub clone, and then link to the branch from here.

> Another question -- I've tried leaving `HOMEBREW_TEMP` and `TEST_TMPDIR` intact and simply adding and using an arbitrary path like `HOMEBREW_SCRATCH = HOMEBREW_PREFIX.parent+"scratch"`.
>
> Adding `HOMEBREW_SCRATCH.children` to `paths_to_delete` in the `teardown` method does not prevent a file leak (`HOMEBREW_SCRATCH` itself needs to be added). What explains the fact that a directory like `HOMEBREW_CACHE` or `HOMEBREW_LOGS` can persist through the test suite (while `HOMEBREW_SCRATCH` cannot)?

It might be not immediately clear, but those directories are not mentioned in the leak report because they are created even before the “before snapshot” that is used for comparison when checking for file leaks is created. To make `HOMEBREW_SCRATCH` work the same way, aside from what you already did, it would need to be created [in these lines in `testing_env.rb`](https://github.com/Homebrew/brew/blob/91d32e77adf743510f0873e9195b3fe0dc065efd/Library/Homebrew/test/testing_env.rb#L9-L10).

eirinikos:
> Can you post the exact changes you've made when trying this? You could for example make those changes in a new separate branch, push it to your GitHub clone, and then link to the branch from here.

Yes! Thanks for the suggestion; that's exactly what I've done [here](https://github.com/eirinikos/brew/commit/a54f226b405115572655e63fd3870e06faea3b65). This commit replaces `HOMEBREW_TEMP` in the definition of `TEST_TMPDIR`, but I've had the same results when removing the second argument entirely -- i.e., `Dir.mktmpdir("homebrew_tests")`. It's clear that neither method is the correct solution, which made me dread pushing "bad code" to this branch (something that seemed unavoidable earlier). 😬 

> 
It might be not immediately clear, but those directories are not mentioned in the leak report because they are created even before the “before snapshot” that is used for comparison when checking for file leaks is created. To make `HOMEBREW_SCRATCH` work the same way, aside from what you already did, it would need to be created [in these lines in `testing_env.rb`](https://github.com/Homebrew/brew/blob/91d32e77adf743510f0873e9195b3fe0dc065efd/Library/Homebrew/test/testing_env.rb#L9-L10).

Oh, right! I should have thought to inspect `testing_env.rb` for clues. Thanks again.

UniqMartin:
> Yes! Thanks for the suggestion; that's exactly what I've done [here](https://github.com/eirinikos/brew/commit/a54f226b405115572655e63fd3870e06faea3b65). This commit replaces `HOMEBREW_TEMP` in the definition of `TEST_TMPDIR`, but I've had the same results when removing the second argument entirely -- i.e., `Dir.mktmpdir("homebrew_tests")`. It's clear that neither method is the correct solution, which made me dread pushing "bad code" to this branch (something that seemed unavoidable earlier). 😬 

The problem here is twofold:

1. One problem comes from the fact that the directory you're specifying for `HOMEBREW_TEMP` doesn't exist yet which confuses `Dir.mktmpdir` and friends and causes this alarming number of failures. The solution for this part is the same like for the scratch directory, i.e. create it in `testing_env.rb`. (This implies only deleting its children in `teardown`.)

2. The other issue once the 1st part is addressed, is a bit more involved and is caused by historic limitations. Homebrew is using [UNIX domain sockets](https://en.wikipedia.org/wiki/Unix_domain_socket) internally and because `Dir.tmpdir` defaults to the contents of the `TMPDIR` environment variable (inspect it yourself by executing `echo "$TMPDIR"` in a terminal) and that paths happens to be quite long already, affixing the additional path components crosses this historic limit and causes error messages like “too long unix socket path”. The only real solution is to make the path shorter and this is achieved by using, e.g.,  `/tmp` as is done by Homebrew [during a regular (non-test) run](https://github.com/Homebrew/brew/blob/91d32e77adf743510f0873e9195b3fe0dc065efd/Library/Homebrew/config.rb#L35-L36).

This is basically your commit, but with my modifications on top (in case any of the above prose isn't sufficiently accurate/clear) and without the test added by you:

```diff
diff --git i/Library/Homebrew/test/lib/config.rb w/Library/Homebrew/test/lib/config.rb
index 062bed33..5818aec4 100644
--- i/Library/Homebrew/test/lib/config.rb
+++ w/Library/Homebrew/test/lib/config.rb
@@ -2,10 +2,9 @@ require "tmpdir"
 require "pathname"
 
 HOMEBREW_BREW_FILE = Pathname.new(ENV["HOMEBREW_BREW_FILE"])
-HOMEBREW_TEMP = Pathname.new(ENV["HOMEBREW_TEMP"] || Dir.tmpdir)
 
 TEST_TMPDIR = ENV.fetch("HOMEBREW_TEST_TMPDIR") { |k|
-  dir = Dir.mktmpdir("homebrew_tests", HOMEBREW_TEMP)
+  dir = Dir.mktmpdir("homebrew-tests-", ENV["HOMEBREW_TEMP"] || "/tmp")
   at_exit { FileUtils.remove_entry(dir) }
   ENV[k] = dir
 }
@@ -24,6 +23,7 @@ HOMEBREW_CACHE_FORMULA = HOMEBREW_PREFIX.parent+"formula_cache"
 HOMEBREW_LOCK_DIR      = HOMEBREW_PREFIX.parent+"locks"
 HOMEBREW_CELLAR        = HOMEBREW_PREFIX.parent+"cellar"
 HOMEBREW_LOGS          = HOMEBREW_PREFIX.parent+"logs"
+HOMEBREW_TEMP          = HOMEBREW_PREFIX.parent+"temp"
 
 TESTBALL_SHA1 = "be478fd8a80fe7f29196d6400326ac91dad68c37"
 TESTBALL_SHA256 = "91e3f7930c98d7ccfb288e115ed52d06b0e5bc16fec7dce8bdda86530027067b"
diff --git i/Library/Homebrew/test/test_integration_cmds.rb w/Library/Homebrew/test/test_integration_cmds.rb
index c020deaf..07068019 100644
--- i/Library/Homebrew/test/test_integration_cmds.rb
+++ w/Library/Homebrew/test/test_integration_cmds.rb
@@ -18,6 +18,7 @@ class IntegrationCommandTests < Homebrew::TestCase
       HOMEBREW_CACHE.children,
       HOMEBREW_LOCK_DIR.children,
       HOMEBREW_LOGS.children,
+      HOMEBREW_TEMP.children,
       HOMEBREW_PREFIX/"bin",
       HOMEBREW_PREFIX/"share",
       HOMEBREW_PREFIX/"opt",
diff --git i/Library/Homebrew/test/testing_env.rb w/Library/Homebrew/test/testing_env.rb
index 214aa978..e55f3f14 100644
--- i/Library/Homebrew/test/testing_env.rb
+++ w/Library/Homebrew/test/testing_env.rb
@@ -7,7 +7,7 @@ require "formulary"
 
 # Test environment setup
 (HOMEBREW_LIBRARY/"Taps/homebrew/homebrew-core/Formula").mkpath
-%w[cache formula_cache locks cellar logs].each { |d| HOMEBREW_PREFIX.parent.join(d).mkpath }
+%w[cache formula_cache locks cellar logs temp].each { |d| HOMEBREW_PREFIX.parent.join(d).mkpath }
 
 # Test fixtures and files can be found relative to this path
 TEST_DIRECTORY = File.dirname(File.expand_path(__FILE__))
```

eirinikos:
> One problem comes from the fact that the directory you're specifying for `HOMEBREW_TEMP` doesn't exist yet which confuses `Dir.mktmpdir` and friends and causes this alarming number of failures. The solution for this part is the same like for the scratch directory, i.e. create it in `testing_env.rb`. (This implies only deleting its children in `teardown`.)
> 
> The other issue once the 1st part is addressed, is a bit more involved and is caused by historic limitations. Homebrew is using [UNIX domain sockets](https://en.wikipedia.org/wiki/Unix_domain_socket) internally and because `Dir.tmpdir` defaults to the contents of the `TMPDIR` environment variable (inspect it yourself by executing `echo "$TMPDIR"` in a terminal) and that paths happens to be quite long already, affixing the additional path components crosses this historic limit and causes error messages like “too long unix socket path”. The only real solution is to make the path shorter and this is achieved by using, e.g., `/tmp` as is done by Homebrew [during a regular (non-test) run](https://github.com/Homebrew/brew/blob/91d32e77adf743510f0873e9195b3fe0dc065efd/Library/Homebrew/config.rb#L35-L36).

Thanks @UniqMartin. 👍  Everything makes sense. I haven't worked with `Dir.tmpdir`, `Dir.mktmpdir`, or `$TMPDIR` very often, so this has been a good lesson. At some point I should also do some more reading on Unix domain sockets and the like (beyond what was in the Wikipedia article you linked).

UniqMartin:
I'm happy with the code as-is and I think it is ready to be merged. I have one thing I'd like you to ask, in the hopes of not asking too much. The thing is, we currently have two rather orthogonal changes in this PR and in a single commit:

- Change how the temporary directory is setup and handled during the tests.
- Addition of a test for `brew irb`.

It would be nice to have this reflected in two separate commits. If I'm asking too much, I'm happy to lend a helping hand in the chat or to do the split for you when merging the changes.

To give a practical example to motivate this: Let's assume we have overlooked some subtlety in how the redefinition of the temporary directory affects the test code and it's starting to act up. (Of course I'm always hoping for the best.) The easiest way to deal with such a situation, particularly if it becomes a pressing issue, is to revert the commit with the problematic changes (to bring them back later in revised form). If the `brew irb` changes were part of the same commit, they would also be reverted despite being perfectly fine and not the source of the problem.

UniqMartin:
Thanks one more time @eirinikos. Great work! :tada:

PR #503, opened 2016-07-13
various regex related cleanup

xu-cheng:
See commit messages for reasoning.


MikeMcQuaid:
Nice work. 👍 if you've tested locally and when 💚

PR #504, opened 2016-07-13
CurlDownloadStrategy#_fetch: do not alter variable @url

xu-cheng:
In case of download failure and retry, altering `@url` can cause side
effect like repeatedly applying HOMEBREW_ARTIFACT_DOMAIN.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -330,23 +330,25 @@ class CurlDownloadStrategy < AbstractFileDownloadStrategy
 
   # Private method, can be overridden if needed.
   def _fetch
+    url = @url
+
     if ENV["HOMEBREW_ARTIFACT_DOMAIN"]
-      @url.sub!(%r{^((ht|f)tps?://)?}, ENV["HOMEBREW_ARTIFACT_DOMAIN"].chomp("/") + "/")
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
:+1:

Library/Homebrew/download_strategy.rb
```diff
@@ -330,23 +330,25 @@ def clear_cache
 
   # Private method, can be overridden if needed.
   def _fetch
+    url = @url
+
     if ENV["HOMEBREW_ARTIFACT_DOMAIN"]
-      @url.sub!(%r{^((ht|f)tps?://)?}, ENV["HOMEBREW_ARTIFACT_DOMAIN"].chomp("/") + "/")
-      ohai "Downloading from #{@url}"
+      url.sub!(%r{^((ht|f)tps?://)?}, ENV["HOMEBREW_ARTIFACT_DOMAIN"].chomp("/") + "/")
```
UniqMartin: Isn't this still indirectly modifying `@url` as `url` is referencing the same `String` instance? If so, this should be fixable by doing `url = @url.dup` or (probably better) by using `String#sub` here.

xu-cheng: 28861ac

PR #510, opened 2016-07-14
update: git init for core tap as well

xu-cheng:
A temporary measure before we decide whether vendor git or not.

Related PR https://github.com/Homebrew/install/pull/47

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -22,22 +22,42 @@ git() {
 git_init_if_necessary() {
   if [[ -n "$HOMEBREW_OSX" ]]
   then
-    OFFICIAL_REMOTE="https://github.com/Homebrew/brew.git"
+    BREW_OFFICIAL_REMOTE="https://github.com/Homebrew/brew"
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/update.sh
```diff
@@ -22,22 +22,42 @@ git() {
 git_init_if_necessary() {
   if [[ -n "$HOMEBREW_OSX" ]]
   then
-    OFFICIAL_REMOTE="https://github.com/Homebrew/brew.git"
+    BREW_OFFICIAL_REMOTE="https://github.com/Homebrew/brew"
+    CORE_OFFICIAL_REMOTE="https://github.com/Homebrew/homebrew-core"
   else
```
MikeMcQuaid: Don't need to fix this now but I'd actually check for Linux and otherwise do just `Homebrew/homebrew-core` so we don't assume non-OS X == Linux.

MikeMcQuaid:
👍

Library/Homebrew/cmd/update.sh
xu-cheng: c2ece46

PR #514, opened 2016-07-15
ARGV#resolved_formulae: also load build options

xu-cheng:
This fixes the problem that `brew audit foo` complains that
an optional dependency is undeclared.

### DIFF
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -27,9 +27,11 @@ module HomebrewArgvExtension
     @resolved_formulae ||= (downcased_unique_named - casks).map do |name|
       if name.include?("/")
         f = Formulary.factory(name, spec)
-        if spec(default=nil).nil? && f.any_version_installed?
-          installed_spec = Tab.for_formula(f).spec
-          f.set_active_spec(installed_spec) if f.send(installed_spec)
+        if f.any_version_installed?
+          tab = Tab.for_formula(f)
+          resolved_spec = spec(default=nil) || tab.spec
+          f.set_active_spec(resolved_spec) if f.send(resolved_spec)
+          f.build = tab
         end
         f
       else
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -219,10 +219,18 @@ class Formulary
   # It will auto resolve formula's spec when requested spec is nil
   def self.from_rack(rack, spec = nil)
     kegs = rack.directory? ? rack.subdirs.map { |d| Keg.new(d) } : []
-
     keg = kegs.detect(&:linked?) || kegs.detect(&:optlinked?) || kegs.max_by(&:version)
-    return factory(rack.basename.to_s, spec || :stable) unless keg
 
+    if keg
+      from_keg(keg, spec)
+    else
+      factory(rack.basename.to_s, spec || :stable)
+    end
+  end
+
+  # Return a Formula instance for the given keg.
+  # It will auto resolve formula's spec when requested spec is nil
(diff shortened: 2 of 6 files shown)


MikeMcQuaid:
👍

DomT4:
I don't want to er, annoy you here, but this still fails on `if build.with?` dependencies 🙈.

```ruby
  depends_on "pcre" => :optional
  depends_on "gettext" => :optional
  depends_on "openssl" if build.with? "brewed-openssl"
  depends_on "curl" if build.with? "brewed-curl"
  depends_on "go" => :build if build.with? "persistent-https"
```

```
git:
  * Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "curl" => :linked
      depends_on "openssl" => :linked
Error: 1 problem in 1 formula
```

xu-cheng:
@DomT4 that would be a known issue due to the usage of `if` for deps declaration

DomT4:
So should we ignore it or should those become:
```ruby
depends_on "curl" => :linked if build.with? "brewed-curl"
```
I presume ignore it, but ❓

ilovezfs:
Yeah this is still pretty broken:
```
iMac-TMP:Homebrew joe$ brew audit --strict vifm
vifm:
  * Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "atk" => :linked
      depends_on "cairo" => :linked
      depends_on "gdk-pixbuf" => :linked
      depends_on "gettext" => :linked
      depends_on "glib" => :linked
      depends_on "pango" => :linked
Error: 1 problem in 1 formula
```
triggered by
```
  depends_on "gtk+" => :optional

  if build.with? "gtk+"
    depends_on "atk" => :linked
    depends_on "cairo" => :linked
    depends_on "gdk-pixbuf" => :linked
    depends_on "gettext" => :linked
    depends_on "glib" => :linked
    depends_on "pango" => :linked
  end
```

>@DomT4 that would be a known issue due to the usage of if for deps declaration 

"Known" or not, it would be nice to at least have a plan of some kind for fixing it.

xu-cheng:
@ilovezfs The only solution is to work on `option "with-foo" do ... end` DSL. As far as I'm aware no one is working on that.

xu-cheng:
Although, it's possible to workaround it by rewriting `ARGV` during formula loading. However, that would be a short-term hack and won't solve everything.

ilovezfs:
The fact that you can explicitly follow the audit directions and still not be able to get the audit error to go away is going to be ... confusing I suspect, or people will just add the linked outside of the `if build.with?` which would be even worse. :see_no_evil:

ilovezfs:
Perhaps this particular audit check should be skipped if the local install receipt indicates something other than a default build.

xu-cheng:
@ilovezfs Well, FYI, this long term bug has caused much worse issues than just `audit` result. For example, it affects the some basic function of `FormulaInstaller`.

ilovezfs:
Since CI will spit out the correct answer, I'm not sure why we'd tell people to run audit locally in the PR template if the output is known to be wrong.

MikeMcQuaid:
We can always `grep` the actual formula text if we're in `audit` for this case to try and produce a better output.

ilovezfs:
@MikeMcQuaid This was @UniqMartin's response when I suggested something similar in another scenario https://github.com/Homebrew/homebrew-core/pull/2253#issuecomment-227665702

Not sure if the same objections apply here or not.

MikeMcQuaid:
@ilovezfs We already have audits that do that and as long as the overhead isn't too high I'm cool with it, personally.

PR #516, opened 2016-07-15
respect user's curlrc file

xu-cheng:
Users may have ~/.curlrc file to include options like proxies.
However, since we overwrite HOME environment variable during the build
and test, curl won't be able to find it.

This commit solves this issue by using CURL_HOME environment variable,
which will be pointed to the original HOME path.

From `curl(1)`:
> 1) curl tries to find the "home dir": It first  checks  for  the
     CURL_HOME and then the HOME environment variables. Failing that,
     it uses getpwuid() on Unix-like systems (which returns the  home
     dir  given the current user in your system). On Windows, it then
     checks for the APPDATA variable, or as a last resort the '%USER-
     PROFILE%\Application Data'.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1324,6 +1324,8 @@ class Formula
   # @private
   def run_test
     old_home = ENV["HOME"]
+    old_curl_home = ENV["CURL_HOME"]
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍

PR #518, opened 2016-07-15
brew.sh: make sudo-checker cross-platform

xu-cheng:


Library/brew.sh
```diff
@@ -195,7 +195,7 @@ then
   HOMEBREW_BASH_COMMAND="$HOMEBREW_LIBRARY/Homebrew/dev-cmd/$HOMEBREW_COMMAND.sh"
 fi
 
-if [[ "$(id -u)" = "0" && "$(/usr/bin/stat -f%u "$HOMEBREW_BREW_FILE")" != "0" ]]
+if [[ "$(id -u)" = "0" && "$(ls -nd "$HOMEBREW_BREW_FILE" | /usr/bin/awk '{ print $3 }')" != "0" ]]
```
MikeMcQuaid: Should be able to do this without `awk`: CC @UniqMartin for Bash skills 😁

Library/Homebrew/brew.sh
```diff
@@ -195,7 +195,8 @@ then
   HOMEBREW_BASH_COMMAND="$HOMEBREW_LIBRARY/Homebrew/dev-cmd/$HOMEBREW_COMMAND.sh"
 fi
 
-if [[ "$(id -u)" = "0" && "$(/usr/bin/stat -f%u "$HOMEBREW_BREW_FILE")" != "0" ]]
+brew_file_ls_info=($(ls -nd "$HOMEBREW_BREW_FILE"))
+if [[ "$(id -u)" = "0" && "${brew_file_ls_info[2]}" != "0" ]]
```
UniqMartin: Can we move this check into a shell function? This would allow for early returns (avoiding an `ls` invocations who's output will never be used in the typical case) and local variables (avoiding additional global variables like `brew_file_ls_info`).

For performance reasons it would also be desirable to exchange the inner and outer check, i.e. first check if we've been invoked with a command that we prefer not to run as root, and only if that's the case, check whether we're root and `bin/brew` is owned by root.

Here's a suggestion (untested), that also incorporates my desire to avoid parsing `ls` output:

```sh
check-run-command-as-root() {
  case "$HOMEBREW_COMMAND" in
    analytics|\
    create|\
    install|\
    link|\
    migrate|\
    pin|\
    postinstall|\
    reinstall|\
    switch|\
    tap-pin|\
    tap|\
    update|\
    upgrade|\
    vendor-install)
      # Continue check as we prefer to not run these as root.
      ;;
    *)
      return
      ;;
  esac

  [[ "$(id -u)" = 0 ]] || return

  # Assume BSD user space, but fall back to GNU if this fails.
  local brew_file_uid
  brew_file_uid="$(/usr/bin/stat -f%u "$HOMEBREW_BREW_FILE" 2>/dev/null)"
  : "${brew_file_uid:="$(stat --format=%u "$HOMEBREW_BREW_FILE" 2>/dev/null)"}"
  [[ "$brew_file_uid" != 0 ]] || return

  odie <<EOS
Cowardly refusing to 'sudo brew $HOMEBREW_COMMAND'
You can use brew with sudo, but only if the brew executable is owned by root.
However, this is both not recommended and completely unsupported so do so at
your own risk.
EOS
}

check-run-command-as-root
```

xu-cheng:
Updated.

Library/Homebrew/brew.sh
```diff
@@ -195,19 +195,30 @@ then
   HOMEBREW_BASH_COMMAND="$HOMEBREW_LIBRARY/Homebrew/dev-cmd/$HOMEBREW_COMMAND.sh"
 fi
 
-if [[ "$(id -u)" = "0" && "$(/usr/bin/stat -f%u "$HOMEBREW_BREW_FILE")" != "0" ]]
-then
+check-run-command-as-root() {
   case "$HOMEBREW_COMMAND" in
-    analytics|install|reinstall|postinstall|link|pin|update|upgrade|vendor-install|create|migrate|tap|tap-pin|switch)
+    analytics|install|reinstall|postinstall|link|pin|update|upgrade|\
+    vendor-install|create|migrate|tap|tap-pin|switch)
```
UniqMartin: Would you mind sorting the entries alphabetically like I did in my code snippet? The order currently makes no sense (to me) and this is making it harder to see whether a command is included in this list or not.

Library/Homebrew/brew.sh
```diff
@@ -195,19 +195,30 @@ then
...
+      ;;
+    *)
+      return
+      ;;
+  esac
+
+  [[ "$(id -u)" = 0 ]] || return
+
+  local brew_file_ls_info=($(ls -nd "$HOMEBREW_BREW_FILE"))
```
UniqMartin: I'm still not thrilled with parsing `ls` output, but if I'm the only one concerned about this, then I guess this is a concern that can be ignored.

UniqMartin:
:+1: with two minor comments that may or may not be addressed.

Library/Homebrew/brew.sh
xu-cheng: I keep to use `ls`. The reasons are:
* It would mean shellout only once instead of twice.
* It avoids surprised behaviors like a failed `stat` outputs something we thought it's success.

UniqMartin: Thanks! I can see an argument for using `ls`. Hopefully few people are running `brew` as root (I never do) and even fewer with a `brew` command we don't want them to run as root, so this check will need to be performed fairly infrequently.

I'm glad we were able to rearrange this, so this isn't shelling out every single time to `id`, even if the command being invoked doesn't care at all whether it is invoked as root or not.

PR #524, opened 2016-07-16
Xcode: track/warn prerelease versions

DomT4:

-----

A number of users have switched over to Xcode 8 on 10.11, which is somehow even more of a breaky decision than running Xcode 8 on the pre-release 10.12. Since we warn about the latter, it makes sense to warn about the former.

### DIFF
--- a/Library/Homebrew/extend/os/mac/diagnostic.rb
+++ b/Library/Homebrew/extend/os/mac/diagnostic.rb
@@ -4,6 +4,7 @@ module Homebrew
       def all_development_tools_checks
         %w[
           check_for_unsupported_osx
+          check_for_prerelease_xcode
           check_for_bad_install_name_tool
           check_for_installed_developer_tools
           check_xcode_license_approved
@@ -27,6 +28,21 @@ module Homebrew
         <<-EOS.undent
           You are using OS X #{MacOS.version}.
           #{who} do not provide support for this #{what}.
+          You may encounter build failures or other breakages.
+          Please create pull-requests instead of filing issues.
+        EOS
+      end
+
+      def check_for_prerelease_xcode
+        return if ARGV.homebrew_developer?
+        # Running a pre-release Xcode on a pre-release OS is expected
+        # and likely to cause less problems than a stable Xcode will.
+        return if OS::Mac.prerelease?
+        return unless MacOS::Xcode.installed?
+        return unless MacOS::Xcode.prerelease?
+
+        <<-EOS.undent
+          You are using a pre-release version of Xcode.
           You may encounter build failures or other breakages.
           Please create pull-requests instead of filing issues.
         EOS
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -27,6 +27,11 @@ module OS
         end
       end
 
+      def prerelease?
+        # TODO: bump to version >= "8.1" after Xcode 8.0 is stable.
+        version > "7.3.1"
+      end
+
       def outdated?
         version < latest_version
       end


Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -32,6 +33,22 @@ def check_for_unsupported_osx
...
+        # and likely to cause less problems than stable Xcodes.
+        return if OS::Mac.prerelease?
+        return unless MacOS::Xcode.installed?
+        return unless MacOS::Xcode.prerelease?
+
+        <<-EOS.undent
+          You are using a pre-release version of Xcode.
+          You may encounter build failures or other breakages.
+          Please consider creating pull-requests instead of
+          filing issues.
```
UniqMartin: Let's make this less vague (like in `check_for_unsupported_osx` directly above) and let's keep it on a single line:

```
Please create pull-requests instead of filing issues.
```

(As a side note: Shouldn't “pull requests” be written without the hyphen?)

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -27,6 +27,11 @@ def latest_version
         end
       end
 
+      def prerelease?
+        # TODO: bump version after Xcode 8.0 is stable.
+        version > "7.3.1"
```
UniqMartin: Maybe make that `version >= "8.0"` in case we'll be graced with Xcode 7.4 (though by now it seems unlikely)? What would we bump this to once Xcode 8 becomes stable?

Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -32,6 +33,22 @@ def check_for_unsupported_osx
         EOS
       end
 
+      def check_for_prerelease_xcodes
```
UniqMartin: There's only one Xcode we're checking, thus let's make this `check_for_prerelease_xcode` (without the trailing `s`).

Library/Homebrew/extend/os/mac/diagnostic.rb
DomT4: I basically cribbed it from the directly above check, but I softened the tone a bit because people have pushed back a little on the language.

Library/Homebrew/os/mac/xcode.rb
DomT4: We haven't got above a `x.3.1` since OS X 10.7, it felt safe enough, but my initial writing was `>= "8.0"` and then I disagreed with myself.

Plan for once Xcode 8 is stable is simply `version > "8.0"`. We have to update Xcode code every release anyway, it only adds an extra line to remember.

Library/Homebrew/extend/os/mac/diagnostic.rb
DomT4: Fair 👍

Library/Homebrew/extend/os/mac/diagnostic.rb
UniqMartin: I don't think we need to soften this. We're already saying “please” and it's not like we're expecting something that is unheard of, i.e. “if you're using beta software, you understand that things *will* be broken and people might not be able to help you”. No?

Library/Homebrew/os/mac/xcode.rb
UniqMartin: I think once Xcode 8 is stable, we should bump this to `version >= "8.1"` as otherwise people will come complaining that `brew doctor` is unhappy when they update to 8.0.1 immediately (and before we get a chance to bump the version in here).

And we know already that anything `> "8.0"` and `< "8.1"` isn't going to be a beta release but a bugfix release, right?

MikeMcQuaid: Agreed on stronger wording (and, if that's not consistent with elsewhere, make it so 👍)

DomT4:
Addressed concerns, rebased & repushed.

UniqMartin:
:+1:

PR #529, opened 2016-07-16
ARGV#kegs and ARGV#resolved_formulae tweak

xu-cheng:
See each commit for detail

### DIFF
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -25,7 +25,7 @@ module HomebrewArgvExtension
   def resolved_formulae
     require "formula"
     @resolved_formulae ||= (downcased_unique_named - casks).map do |name|
-      if name.include?("/")
+      if name.include?("/") || File.exist?(name)
         f = Formulary.factory(name, spec)
         if f.any_version_installed?
           tab = Tab.for_formula(f)
@@ -254,7 +254,7 @@ module HomebrewArgvExtension
   def downcased_unique_named
     # Only lowercase names, not paths, bottle filenames or URLs
     @downcased_unique_named ||= named.map do |arg|
-      if arg.include?("/") || arg.end_with?(".tar.gz")
+      if arg.include?("/") || arg.end_with?(".tar.gz") || File.exist?(arg)
         arg
       else
         arg.downcase
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -49,7 +49,7 @@ module HomebrewArgvExtension
     require "keg"
     require "formula"
     @kegs ||= downcased_unique_named.collect do |name|
-      rack = Formulary.to_rack(name)
+      rack = Formulary.to_rack(name.downcase)
 
       dirs = rack.directory? ? rack.subdirs : []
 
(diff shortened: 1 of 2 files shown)


xu-cheng:
I merged this as it's needed since I ship the HeadVersion change. (i.e. if opt link and LinkedKeg is broken, we need `installed_prefix` instead of `prefix` to locate the latest Head installation.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -25,7 +25,7 @@ def formulae
   def resolved_formulae
     require "formula"
     @resolved_formulae ||= (downcased_unique_named - casks).map do |name|
-      if name.include?("/")
+      if name.include?("/") || File.exist?(name)
```
UniqMartin: I'm not sure the added ` File.exist?(name)` makes sense. Isn't `name` just the argument given via the command line, i.e. typically something like `ack`? What I'm trying to say, doesn't this simply check for a file of the same name in the current working directory? What was the original intention of adding this?

UniqMartin: Sorry, after re-reading the commit message I think I understand this better. But I think the check should be a bit stricter, e.g. something like:

```ruby
(File.file?(name) && name.end_with?(".rb"))
```

And could you please explain what issue this solves, as I'm still unsure about this?

xu-cheng: This is for loading formula from file path, e.g. `brew foo bar.rb`

xu-cheng: The issue tried to solve is as following. You have a custom formula `foo.rb`. You should be able to do thing like `brew install/reinstall foo.rb`. And it should be loading the custom formula. 

You are right we probably should check file ext.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -254,7 +254,7 @@ def spec(default = :stable)
   def downcased_unique_named
     # Only lowercase names, not paths, bottle filenames or URLs
     @downcased_unique_named ||= named.map do |arg|
-      if arg.include?("/") || arg.end_with?(".tar.gz")
+      if arg.include?("/") || arg.end_with?(".tar.gz") || File.exist?(arg)
```
UniqMartin: Similar comment here about a bit stricter checking. I don't think it's helpful to accept arbitrary file names that might or might not exist in the current working directory. Maybe it's worth doing the check in a separate method named `formula_file?` as this check is already done in two different locations.

From the comment above this: “not paths, bottle filenames”. – This doesn't seem to be the case, and apparently already for a long time. 😕

Library/Homebrew/extend/ARGV.rb
UniqMartin: Thanks for the explanation! It's going to be interesting (though very unlikely) once there's a file named `foo.rb` and a formula named `foo.rb` (perfectly legal formula name in file `foo.rb.rb`). I have no idea what this will resolve to in this case. 😀

> You are right we probably should check file ext.

:+1:

UniqMartin:
Sorry I hadn't taken the time to provide feedback on this earlier …

Library/Homebrew/extend/ARGV.rb
```diff
@@ -65,10 +65,18 @@ def kegs
           Keg.new(linked_keg_ref.resolved_path)
         elsif dirs.length == 1
           Keg.new(dirs.first)
-        elsif (prefix = (name.include?("/") ? Formulary.factory(name) : Formulary.from_rack(rack)).prefix).directory?
-          Keg.new(prefix)
         else
-          raise MultipleVersionsInstalledError.new(rack.basename)
+          f = if name.include?("/") || File.exist?(name)
```
UniqMartin: Just for completeness' sake: Same comment about stricter check and maybe a `formula_file?` helper method to avoid duplicate and inconsistent checks.

Library/Homebrew/extend/ARGV.rb
xu-cheng: Actually, until this PR, `brew reinstall foo.rb` will complain that formula unfound.

PR #530, opened 2016-07-16
ENV: fix homebrew_extra_pkg_config_paths

ilovezfs:
Switch to HOMEBREW_LIBRARY since HOMEBREW_LIBRARY_PATH already includes
"/Homebrew" and HOMEBREW_LIBRARY is less confusing.

Closes https://github.com/Homebrew/homebrew-core/issues/3038.

### DIFF
--- a/Library/Homebrew/extend/os/mac/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/std.rb
@@ -31,7 +31,7 @@ module Stdenv
   end
 
   def homebrew_extra_pkg_config_paths
-    ["#{HOMEBREW_LIBRARY_PATH}/Homebrew/os/mac/pkgconfig/#{MacOS.version}"]
+    ["#{HOMEBREW_LIBRARY}/Homebrew/os/mac/pkgconfig/#{MacOS.version}"]
   end
 
   # Sets architecture-specific flags for every environment variable
--- a/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
@@ -26,7 +26,7 @@ module Superenv
   # @private
   def homebrew_extra_pkg_config_paths
     paths = \
-      ["#{HOMEBREW_LIBRARY_PATH}/Homebrew/os/mac/pkgconfig/#{MacOS.version}"]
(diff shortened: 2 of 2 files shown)


ilovezfs:
@MikeMcQuaid this fixes a regression introduced by a02be9eea204264ce48827ac365c8b53b986d66a

By the way, this is at least the second time this same thing has happened probably because the variable name is "backwards."

xu-cheng:
We probably should avoid to use `HOMEBREW_LIBRARY_PATH` at all, which is intended to refer to the ruby load path. 

As it would be help for readability and avoid misunderstanding, if we only keep `HOMEBREW_LIBRARY` and remove `HOMEBREW_LIBRARY_PATH`.

DomT4:
👍 to Xu's comment. Let's try to get this shipped out fairly quickly, leaving things like this broken for longer than discovered is extra pain for us & users.

ilovezfs:
PR refreshed.

DomT4:
LGTM. I personally wouldn't bother to wait for CI on this one, but up to you.

ilovezfs:
Ran test-bot locally, and it's fine of course

UniqMartin:
Thanks for the fix! I agree that the similarity between `HOMEBREW_LIBRARY_PATH` and `HOMEBREW_LIBRARY` can cause confusion and thus lead to subtle bugs like these. I'm not sure we can eliminate the former completely, as `HOMEBREW_LIBRARY_PATH` and `HOMEBREW_LIBRARY/"Homebrew"` are very different things during `brew tests`. But we can at least try to find a better name.

Unfortunately, `HOMEBREW_LOAD_PATH` is already taken and is a list of paths during testing, so cannot be used as a substitute for `HOMEBREW_LIBRARY_PATH`. Does `HOMEBREW_CODE_PATH` sound reasonable?

MikeMcQuaid:
Sorry for the pain, folks.

> Thanks for the fix! I agree that the similarity between HOMEBREW_LIBRARY_PATH and HOMEBREW_LIBRARY can cause confusion and thus lead to subtle bugs like these. I'm not sure we can eliminate the former completely, as HOMEBREW_LIBRARY_PATH and HOMEBREW_LIBRARY/"Homebrew" are very different things during brew tests. But we can at least try to find a better name.

Agreed.

PR #532, opened 2016-07-16
Prevent brew audit crash from nil exitstatus

shulima:

-----

While working on some new formulae I realized that any attempts to audit a formula (any formula, not just my new creations) resulted in:
```
$ brew audit --strict --online libhivex
Error: undefined method `>' for nil:NilClass
Please report this bug:
/usr/local/Library/Homebrew/cmd/style.rb:67:in `check_style_impl'
/usr/local/Library/Homebrew/cmd/style.rb:45:in `check_style_json'
/usr/local/Library/Homebrew/cmd/audit.rb:62:in `audit'
```

I've yet to figure out why `$?.exitstatus` is nil (I'm betting my money on ruby version clash, I have several rubies managed with chruby), but regardless it'd probably be prudent to check for nil before doing the comparison.

I'm afraid that adding tests for this behaviour exceeds my capabilites.

### DIFF
--- a/Library/Homebrew/cmd/style.rb
+++ b/Library/Homebrew/cmd/style.rb
@@ -63,8 +63,10 @@ module Homebrew
       !$?.success?
     when :json
       json = Utils.popen_read_text("rubocop", "--format", "json", *args)
-      # exit status of 1 just means violations were found; others are errors
-      raise "Error while running rubocop" if $?.exitstatus > 1
+      # exit status of 1 just means violations were found; other numbers mean execution errors
+      # exitstatus can also be nil if RuboCop process crashes, e.g. due to
+      # native extension problems
+      raise "Error while running RuboCop" if $?.exitstatus.nil? || $?.exitstatus > 1
       RubocopResults.new(Utils::JSON.load(json))
     else
       raise "Invalid output_type for check_style_impl: #{output_type}"


UniqMartin:
According to the documentation, `$?.exitstatus` should only be `nil` if `$?.exited?` is `false` (i.e. the child process is still running) which would be rather surprising at this point. To me this seems to be a more involved bug and I'd like to find out the root cause, if possible.

Does the failure reproducibly happen on every single invocation of `brew audit` or only sporadically? Can you tell me a bit more about your setup so that I can try to reproduce the problem? (I've never encountered this problem before.)

> I'm betting my money on ruby version clash, I have several rubies managed with chruby

Homebrew normally uses the system Ruby for running itself, even if there are other Rubies in PATH, thus this seems unlikely. Can you please provide your `brew config` output? Thanks!

shulima:
Happy to cooperate!

The issue happens with every invocation. I'm positive it's something to do with the fact that I use chruby to switch between homebrew-installed ruby versions [1], because after switching to system ruby (`chruby system`) brew audit works fine. 

I believe `$?.exitstatus` can also be nil if the child was signaled (https://www.ruby-forum.com/topic/4402802), and I've seen segfaults before due to mismatched gems with native extensions (e.g. json).

Here's brew config with my usual 2.2.5:

```
$ brew config
HOMEBREW_VERSION: 0.9.9
HEAD: 008e8b7a1063a1160029d601cffdb57f7156dd1d
Last commit: 60 minutes ago
Core tap HEAD: d352c79b8213c1e8ebbf722c430a258f1e23f641
Core tap last commit: 3 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local
CPU: quad-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 7.3 build 703
Git: 2.9.1 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /usr/local/Cellar/ruby22/2.2.5/bin/ruby
Java: 1.8.0_31, 1.6.0_65-b14-468
OS X: 10.11.5-x86_64
Xcode: 7.3.1
CLT: 7.3.1.0.1.1461711523
X11: 2.7.9 => /opt/X11
```

And here's `brew config` after `chruby system`:
```
$ brew config
HOMEBREW_VERSION: 0.9.9
HEAD: 008e8b7a1063a1160029d601cffdb57f7156dd1d
Last commit: 63 minutes ago
Core tap HEAD: d352c79b8213c1e8ebbf722c430a258f1e23f641
Core tap last commit: 3 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local
CPU: quad-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 7.3 build 703
Git: 2.9.1 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.12/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /usr/local/bin/ruby => /usr/local/Cellar/ruby/2.3.1/bin/ruby
Java: 1.8.0_31, 1.6.0_65-b14-468
OS X: 10.11.5-x86_64
Xcode: 7.3.1
CLT: 7.3.1.0.1.1461711523
X11: 2.7.9 => /opt/X11
```

[1] I've known for a while it wasn't a good idea, but I was too lazy to compile rubies with ruby-install. Until now.

UniqMartin:
Thanks, I'll take a closer look and see if I can reproduce this. If you're happy to experiment, would you mind starting a `brew irb` session and then running the following commands?

```ruby
Homebrew.install_gem_setup_path! "rubocop", "0.41.2"
safe_system "rubocop", "ack".f.path
puts $?.inspect
safe_system "rubocop", "zpython".f.path
puts $?.inspect
```

I'd be interested whether this causes any errors or whether this successfully runs RuboCop.

shulima:
Looks like I was on the right track with the segfault:

https://gist.github.com/shulima/8e26ffcad90ba19cca0654be3c225902

I switched to 2.0.0-p648 and gem-pristine'd rubocop -- that removed the crash and the problem.

Library/Homebrew/cmd/style.rb
```diff
@@ -64,7 +64,7 @@ def check_style_impl(files, output_type, options = {})
     when :json
       json = Utils.popen_read_text("rubocop", "--format", "json", *args)
       # exit status of 1 just means violations were found; others are errors
-      raise "Error while running rubocop" if $?.exitstatus > 1
+      raise "Error while running rubocop" if $?.exitstatus.nil? || $?.exitstatus > 1
```
UniqMartin: Can you extend the comment above this line to explain why `exitstatus` can be `nil`? I think that's non-obvious and thus worth documenting. Maybe something like “exitstatus can be nil if the RuboCop process crashes, e.g. due to problems with native Ruby extensions”?

While you're here, can you also fix the capitalization of RuboCop (👈 that's how it should be spelled) in the message string?

UniqMartin:
> Looks like I was on the right track with the segfault:

Thanks! I'm more confident merging your fix (once you address my code comment) now that we have understood the underlying issue. You've got good instincts! `Utils.open_read_text`, that is used internally to invoke RuboCop and collect the results, unfortunately discards all output to `$stderr`, hence the cause was not immediately obvious.

> I switched to 2.0.0-p648 and gem-pristine'd rubocop -- that removed the crash and the problem.

:+1:

shulima:
Amended as requested.

UniqMartin:
Merged in b51283424321cb996d0a587e9a62a070d21f3342. Thank you for this contribution to Homebrew, @shulima! :tada: You rock!

MikeMcQuaid:
Nice work here 👍

PR #533, opened 2016-07-16
Keg#link: run optlink first

xu-cheng:
This prevents a link conflict during `brew upgrade` causing opt link
stuck to old version. At this point, users will have to run `brew switch`
to fix it.


UniqMartin:
:+1: Thanks for the quick fix and thanks for the explanation in Homebrew/homebrew-core#2987!

PR #534, opened 2016-07-16
Fix HeadVersions

### LINKED ISSUE #531: HEAD installs broken.

```
brew install pond --HEAD -v
...
==> go build -tags nogui -o /usr/local/Cellar/pond/HEAD-72b0708_1/bin/pond ./client
Error: Empty installation
/usr/local/Library/Homebrew/formula_installer.rb:592:in `build'
/usr/local/Library/Homebrew/formula_installer.rb:237:in `install'
/usr/local/Library/Homebrew/cmd/install.rb:269:in `install_formula'
/usr/local/Library/Homebrew/cmd/install.rb:149:in `block in install'
/usr/local/Library/Homebrew/cmd/install.rb:149:in `each'
/usr/local/Library/Homebrew/cmd/install.rb:149:in `install'
```

```
ls -1 /usr/local/Cellar/pond/HEAD-72b0708_1
INSTALL_RECEIPT.json
LICENSE
README.md
bin
```

```
/usr/local/Cellar/pond/HEAD-72b0708_1/bin/pond
>>> Pond...
>>> Pond keeps private keys, messages etc on disk for a limited amount of time and that information can be encrypted with a passphrase.
```

@vladshablinsky

vladshablinsky:

-----

Fixes #531

### DIFF
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -53,7 +53,7 @@ class SoftwareSpec
     @resource.owner = self
     resources.each_value do |r|
       r.owner     = self
-      r.version ||= version
+      r.version ||= (version.head? ? Version.create("HEAD") : version.dup)
     end
     patches.each { |p| p.owner = self }
   end


vladshablinsky:
This is a quick fix for the bug.

cc @xu-cheng @DomT4

vladshablinsky:
I think, this is a better way to fix it.

```diff
diff --git a/Library/Homebrew/build.rb b/Library/Homebrew/build.rb
index 8915e94..f7a4543 100644
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -127,6 +127,7 @@ class Build
         interactive_shell(formula)
       else
         formula.prefix.mkpath
+        formula.head.lock_version! if formula.head
 
         formula.install
 
diff --git a/Library/Homebrew/software_spec.rb b/Library/Homebrew/software_spec.rb
index b16a33c..0a6c848 100644
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -53,7 +53,7 @@ class SoftwareSpec
     @resource.owner = self
     resources.each_value do |r|
       r.owner     = self
-      r.version ||= version.dup
+      r.version ||= version
     end
     patches.each { |p| p.owner = self }
   end
@@ -207,6 +207,10 @@ class HeadSoftwareSpec < SoftwareSpec
     @resource.version = Version.create("HEAD")
   end
 
+  def lock_version!
+    version.lock!
+  end
+
   def verify_download_integrity(_fn)
     nil
   end
diff --git a/Library/Homebrew/version.rb b/Library/Homebrew/version.rb
index 002bd53..cadd2c2 100644
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -401,7 +401,12 @@ class HeadVersion < Version
     @commit = @version[/^HEAD-(.+)$/, 1]
   end
 
+  def lock!
+    @locked = true
+  end
+
   def update_commit(commit)
+    return if @locked
     @commit = commit
     @version = if commit
       "HEAD-#{commit}"
```

Library/Homebrew/software_spec.rb
```diff
@@ -53,7 +53,7 @@ def owner=(owner)
     @resource.owner = self
     resources.each_value do |r|
       r.owner     = self
-      r.version ||= version
+      r.version ||= version.dup
```
xu-cheng: I would suggest `version.head? ? HeadVersion.new("HEAD") : version`

As the commit of formula should have nothing to do with the version of resources.

UniqMartin:
I think the current fix in this PR makes more sense (in my opinion), as I believe every resource should have an independent `Version` instance and they shouldn't be influencing each other.

If this already fixes the issue (and doesn't introduce any new issues we're aware of), let's merge this now and discuss a possible refinement of the fix in a separate PR (where we have more time).

Library/Homebrew/software_spec.rb
UniqMartin: I think the `dup` is cleaner and safer, not to mention that we don't want to instantiate `HeadVersion` directly (only via `Version.create`), but that's just a detail.

UniqMartin: Is there any scenario where it does make sense that resources reference the same `Version` (or `HeadVersion`) instance as the formula itself and where changes to one of them should influence all the others? If not, I'm strongly in favor of `dup`.

xu-cheng: My main thought is we should discard the commit of HEAD-commit from formula when passing to resources  

We could certainly add `dup` to non-stable version as `... : version.dup`.

vladshablinsky: Does the `version` store the commit at the moment we call `version.dup`?

xu-cheng: That's why I think we should only call `dup` for non-Head version, which doesn't have commit.

xu-cheng: Also FYI, the reason that we need discard commit is otherwise we will have to redownload sources every time we install formula. (Resource version will interfere the cache file name)

UniqMartin: I see, makes sense! So based on that, I think this is the best short-term fix:

```ruby
r.version ||= (version.head? ? Version.create("HEAD") : version.dup)
```

vladshablinsky: I mean what if we call `dup` on HEAD version, is there commit at the moment we `call` dup? So, `Version.create` and `version.dup` are almost the same things, however `Version.create` is easier to understand.

vladshablinsky: I see, that @UniqMartin :+1:'d your suggestion, so I'm fixing it that way.

vladshablinsky: > I see, makes sense! So based on that, I think this is the best short-term fix:
> r.version ||= (version.head? ? Version.create("HEAD") : version.dup)

The same as I committed before refreshing the webpage! 👍

vladshablinsky:
Thanks @xu-cheng @UniqMartin and @DomT4 ! And sorry for the bug. 😕

UniqMartin:
Thanks for fixing this so quickly, @vladshablinsky!

DomT4:
This was an impressively quick PR; there's not that many in Homebrew I don't at least see before they get merged 😄. Looks good locally. Thank you @vladshablinsky! Appreciate the quick fix.

PR #537, opened 2016-07-17
remove unnecessary build options manipulation

xu-cheng:
* We already loaded build options in ARGV.resolved_formulae for test and
postinstall
* remove unnecessary `Tab.for_formula`


MikeMcQuaid:
Nice 👍

UniqMartin:
:+1: Great simplification here.

PR #538, opened 2016-07-17
LinkageChecker: simplify logic

xu-cheng:
BuiltOptions#with? already handles the option_names for dependencies.

Also check disabled recommended deps.

### DIFF
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -53,7 +53,8 @@ class LinkageChecker
   def check_undeclared_deps
       filter_out = proc do |dep|
         next true if dep.build?
-        dep.optional? && !dep.option_names.any? { |n| formula.build.with?(n) }
+        next false unless dep.optional? || dep.recommended?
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍

Library/Homebrew/os/mac/linkage_checker.rb
```diff
@@ -53,7 +53,8 @@ def check_dylibs
   def check_undeclared_deps
       filter_out = proc do |dep|
         next true if dep.build?
-        dep.optional? && !dep.option_names.any? { |n| formula.build.with?(n) }
+        next false if !dep.optional? && !dep.recommended?
```
UniqMartin: I think this is a case where `unless` would improve readability, i.e.:

```ruby
next false unless dep.optional? || dep.recommended?
```

UniqMartin:
:+1: apart from one minor suggestion.

UniqMartin:
*Note regarding the commit message/PR title:* It should be “LinkageChecker: simplify logic” (i.e. without a plural-s since “logic” doesn't have a plural).

PR #546, opened 2016-07-18
test-bot: fix coverage on Travis CI, related improvements

UniqMartin:

-----

The primary aim is to fix a regression caused by #505, namely that `brew tests --coverage` runs twice and the second run (with `--generic`) replaces the results of the previous run, leading to a low coverage of just 33% on <https://coveralls.io/github/Homebrew/brew>.

To make this PR less dull, I allowed myself to include an update of the SimpleCov and Coveralls gems and to tweak the coverage settings a bit, now that *all* Ruby code relevant to the coverage report lives in `Library/Homebrew/`. (See commit messages for even more detail.)

This has been extensively tested locally, with Ruby 1.8.7/2.0 and with various combinations of command line options. (*Fun fact:* At the time of testing we were covering exactly 8,000 lines.)

cc @MikeMcQuaid

### DIFF
--- a/Library/Homebrew/dev-cmd/test-bot.rb
+++ b/Library/Homebrew/dev-cmd/test-bot.rb
@@ -659,14 +659,15 @@ module Homebrew
 
       if @tap.nil?
         tests_args = []
+        tests_args_coverage = []
         if RUBY_TWO
           tests_args << "--official-cmd-taps"
-          tests_args << "--coverage" if ENV["TRAVIS"]
+          tests_args_coverage << "--coverage" if ENV["TRAVIS"]
         end
         test "brew", "tests", *tests_args
         test "brew", "tests", "--generic", "--only=integration_cmds",
                               *tests_args
-        test "brew", "tests", "--no-compat"
+        test "brew", "tests", "--no-compat", *tests_args_coverage
         test "brew", "readall", "--syntax"
         # test update from origin/master to current commit.
         test "brew", "update-test"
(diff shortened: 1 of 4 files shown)


Library/Homebrew/dev-cmd/test-bot.rb
```diff
@@ -658,15 +658,17 @@ def homebrew
...
+          tests_args_generic << "--official-cmd-taps"
+          tests_args_coverage << "--coverage" if ENV["TRAVIS"]
         end
-        test "brew", "tests", *tests_args
-        test "brew", "tests", "--generic", "--only=integration_cmds",
-                              *tests_args
-        test "brew", "tests", "--no-compat"
+        test "brew", "tests", *tests_args_classic
+        test "brew", "tests", *tests_args_generic
+        test "brew", "tests", *tests_args_coverage
```
UniqMartin: I guess there are different alternatives here, but generating the coverage report during `brew tests --no-compat` felt like the safest and most desirable option. This would also be consistent with what we have been doing in the past before some refactoring changed this `test-bot` code.

MikeMcQuaid: I think these args are much harder to follow than what there was before, personally. I'd be in favour of inlining all non-Ruby 2 stuff.

MikeMcQuaid:
One comment otherwise 👍

MikeMcQuaid:
And thanks for the fix 🙇

Library/Homebrew/dev-cmd/test-bot.rb
UniqMartin: Fair point. I was personally more concerned with line length and with a more uniform look (and making sure that all invocations have individual variables for their arguments to avoid false sharing of arguments).

UniqMartin:
Updated to address your comment. Is this better now? (The diff is certainly smaller.)

MikeMcQuaid:
👍

PR #555, opened 2016-07-19
Remove legacy Aliases/Formula symlink from Library

UniqMartin:

-----

This has been working flawlessly for me for the past three months, thus basically since we have made the split. Enough time for active projects to catch up with these developments and drop the legacy symlinks that just mask broken behavior in software/scripts that are still relying on those locations. (It also helps to further tidy up the repository outside of `Library/Homebrew/` and avoids symlinks that point nowhere when only this repository is checked out, i.e. when not immediately installing the homebrew/core tap, too.)

### DIFF
--- a/Library/Aliases
+++ b/Library/Aliases
@@ -1,1 +1,0 @@
-Taps/homebrew/homebrew-core/Aliases
--- a/Library/Formula
+++ b/Library/Formula
@@ -1,1 +1,0 @@
-Taps/homebrew/homebrew-core/Formula


MikeMcQuaid:
Vaguely concerned we may break assumptions of other scripts/tools but given 1.0 is approaching it makes sense to formalise our API so 👍

UniqMartin:
Cheers! I'll wait a bit to see if there's input from other maintainers and until a time when I will be able to monitor the issue queue a bit better to be able to react to unexpected grief this might cause despite my general optimism regarding this change.

DomT4:
I somewhat fear we've broken a fair bit already lately, albeit for good intentions & this may end up doing the same for a small element of people. We've got to pull the trigger at some point though, so I don't have an objection to getting it over with.

MikeMcQuaid:
> I somewhat fear we've broken a fair bit already lately, albeit for good intentions & this may end up doing the same for a small element of people. We've got to pull the trigger at some point though, so I don't have an objection to getting it over with.

My view is that we're getting towards a 1.0.0 and at that point we'll just stop doing stuff like this. I'd like all @Homebrew/maintainers to think about what backwards incompatible changes we want to make for that point (as it'll be a very long time before we'll do so again)

ilovezfs:
I actually still find the Library/Formula symlink useful as a shortcut. I'm sure I can find an alternative solution, but I've been finding it a valuable "feature" in its own right.

UniqMartin:
> I actually still find the Library/Formula symlink useful as a shortcut. I'm sure I can find an alternative solution, but I've been finding it a valuable "feature" in its own right.

Would you mind sharing how you use it? Maybe I can suggest some alternative that works without the symlink or compensate for the removal by tweaking some other part of Homebrew.

ilovezfs:
I use it in two ways:
1. with `rcd` which I define as follows:
```
rcd(){
        target="$(mdutil -t "$1")"
        if [ $? -eq 0 ]
        then
                pushd "$target"
        else
                pushd "$1"
        fi
}
```
`realpath` from coreutils could be used instead of `mdutil -t` if you don't have Spotlight enabled.
Usage:
```
bash-3.2$ rcd /usr/local/Library/Formula
bash-3.2$ pwd
/usr/local/Library/Taps/homebrew/homebrew-core/Formula
bash-3.2$ 
```
2. When doing `scp` from `10.0.1.15:/usr/local/Library/Formula` to /Users/joe/Developer/homebrew-core/Formula`:
Usage:
```
bash-3.2$ pwd
/Users/joe
bash-3.2$ h
refs/heads/master
/Users/joe/Developer/homebrew-core
bash-3.2$ pwd
/Users/joe/Developer/homebrew-core
bash-3.2$ scp 10.0.1.15:/usr/local/Library/Formula/gitlab-ci-multi-runner.rb Formula
```

UniqMartin:
> 1) with `rcd` which I define as follows:

Assuming your primary goal is to end up in the Git repository and not necessarily in the `Formula` directory, incorporating `brew --repository homebrew/core` could help. E.g. `tcd` (tap cd):

```sh
tcd() {
  local tap="$1"
  [[ "$tap" = */* ]] || tap="homebrew/$tap"
  pushd "$(brew --repository "$tap")"
}

# Usage:
tcd core
tcd uniqmartin/tools
```

> 2) When doing `scp` from `10.0.1.15:/usr/local/Library/Formula` to `/Users/joe/Developer/homebrew-core/Formula`:

`brew ruby -e 'ARGV[0].f.path'` (requires homebrew/dev-tools) could prove useful as a building block, though having `brew formula FORMULA` similarly to the existing `brew command COMMAND` would probably be useful, too. E.g. `bfp` (brew formula path):

```sh
alias bfp='brew ruby -e "puts ARGV[0].f.path"'

# Usage:
bfp qt5
bfp homebrew/science/octave
```

ilovezfs:
Yeah, those look like good workarounds. The only quibble I have is that they require definition in the environment whereas the symlink is currently guaranteed to be there regardless of environment.

For the latter, I assume you're recommending doing this?
```
scp 10.0.1.15:`bfp qt5` Formula
```

UniqMartin:
> For the latter, I assume you're recommending doing this?
>
> ```
> scp 10.0.1.15:`bfp qt5` Formula
> ```

`bfp` was mainly meant as a demonstration of how this could be done, but yes, if Homebrew is installed into the same prefix on both machines, this should work. If not, this needs some more work. If I had a similar workflow to you, I would probably go one step further and wrap the `scp` invocation in a separate shell function or helper script that accepts a formula name.

ilovezfs:
What is the benefit of killing the symlink? Lengthening the path to the core formulae has always felt like a regression to me, and the symlink has been a semi-adequate redress, but I'm sure your workarounds will suffice if this change is actually necessary.

UniqMartin:
The longer path is the price of the core/formula split and handling homebrew/core like any other tap (with a few exceptions). Killing the symlinks isn't strictly necessary, but as I mentioned in my top comment it removes legacy cruft (otherwise some people/scripts will continue to make wrong assumptions about where formulae actually reside) and avoids symlinks that point outside of the repository and thus appear broken if this repository is checked out in isolation.

ilovezfs:
We should brace ourselves for the amount of breakage this is going to cause for third parties, and the possibly resultant anger and frustration, as well as temporary support burden, but I think it's fine to :ship:.

MikeMcQuaid:
1.0.0 is approaching us and we're pushing a few other similarly implicit API-breaking things. I think we should merge this before then and I'll write up the various big changes we've made.

MikeMcQuaid:
This has been waiting for long enough without additional thoughts and I'd rather we break this a while before 1.0.0 rather than after so: 🚢ing.

tkelman:
> 1.0.0 is approaching us and we're pushing a few other similarly implicit API-breaking things. I think we should merge this before then and I'll write up the various big changes we've made.

Is there a place third-parties can click subscribe or otherwise be informed of these before they happen?

DomT4:
This repo, essentially. Some stuff goes on the mailing list, some stuff goes on Twitter, but ultimately if you need to know day-in-day-out what Homebrew is doing with core code this repo is the one to watch.

It's relatively low traffic as far as large projects ever are low traffic ever since the formulae element was split out.

UniqMartin:
Thanks for shipping this, @MikeMcQuaid! :bow: I was hesitant to merge this in the past few weeks because we've seen a fair amount of breaking changes and/or severe bugs in the recent past and I didn't want to contribute with another breaking change without being able to watch bot and repository activity adequately and react to potential problems quickly …

MikeMcQuaid:
> Is there a place third-parties can click subscribe or otherwise be informed of these before they happen?

The mailing list or our Twitter for huge changes but generally: no. I plan on fixing our release and release announcement process for 1.0.0.

PR #558, opened 2016-07-19
tests: add cmd/analytics integration test

eirinikos:

-----

This change adds test coverage for `brew analytics`. 🎈

### DIFF
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -737,4 +737,23 @@ class IntegrationCommandTests < Homebrew::TestCase
     assert_match "Patch failed to apply",
       cmd_fail("pull", "https://github.com/Homebrew/homebrew-core/pull/1")
   end
+
+  def test_analytics
+    HOMEBREW_REPOSITORY.cd do
+      shutup do
+        system "git", "init"
+      end
+    end
+
+    assert_match "Invalid usage", cmd_fail("analytics", "on", "off")
+    assert_match "Invalid usage", cmd_fail("analytics", "testball")
+    assert_match "Analytics is enabled", cmd("analytics")
+    assert_match "Analytics is disabled",
+      cmd("analytics", "HOMEBREW_NO_ANALYTICS" => "1")
+
+    cmd("analytics", "regenerate-uuid")
+    cmd("analytics", "on")
+    cmd("analytics", "off")
+    assert_match "Analytics is disabled", cmd("analytics")
+  end
 end


MikeMcQuaid:
Nice work 👍. Will let @UniqMartin do his usual 😉

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -706,4 +706,23 @@ def test_irb
...
+    assert_match "Invalid usage", cmd_fail("analytics", "testball")
+    assert_match "Analytics is enabled", cmd("analytics")
+    assert_match "Analytics is disabled",
+      cmd("analytics", "HOMEBREW_NO_ANALYTICS" => "1")
+
+    HOMEBREW_REPOSITORY.cd do
+      shutup do
+        system "git", "init"
+      end
+    end
```
UniqMartin: Mostly out of curiosity: Why is the Git repository initialized in the middle of the test and how does (or doesn't) this influence the commands that are issued before and after this block?

UniqMartin:
Didn't find anything to complain about (I know, I'm getting a bit out of my character :joy:), but I left a question as a code comment nonetheless.

:+1: Looks good to me!

Library/Homebrew/test/test_integration_cmds.rb
eirinikos: Hmmm. I had thought that the Git repository needed to be initialized before running `cmd("analytics", "regenerate-uuid")`, but this `cmd` seems to run smoothly regardless. I would have expected an error and I thought I remembered receiving one last week...

It does need to be initialized before toggling between `brew analytics on` and `brew analytics off`; running those `cmd`s without first initializing the repo does result in this error:

```
error: could not lock config file /tmp/homebrew-tests-20160726-97875-1wboued/prefix/.git/config:
No such file or directory
Error: Failure while executing:
git config --file=/tmp/homebrew-tests-20160726-97875-1wboued/prefix/.git/config
--replace-all homebrew.analyticsdisabled false[true]
```

In any case, I guess there's no solid reason for the Git repository to be initialized in the middle of the test. It seems to be fine to initialize it at the beginning. 😄

eirinikos:
Thank you! I went ahead and moved the Git repository initialization to the beginning of the test (since there wasn't a compelling reason to keep it in the middle). And I rebased my branch to resolve the previously existing merge conflicts.

Library/Homebrew/test/test_integration_cmds.rb
UniqMartin: Thanks for the detailed response! I'm happy with the conclusion you drew from this (moving the repository initialization to the top of the test). :+1:

UniqMartin:
Good work! Thanks again @eirinikos! :tada:

MikeMcQuaid:
Thanks @eirinikos!

PR #559, opened 2016-07-20
tests: add cmd/switch integration test

eirinikos:

-----

This change adds test coverage for `brew switch`.

Since [these lines](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/switch.rb#L20-L33) return exit codes that fall outside the purview of `cmd_fail`, would it be OK to add another helper method to address them?

### DIFF
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -756,4 +756,21 @@ class IntegrationCommandTests < Homebrew::TestCase
     cmd("analytics", "off")
     assert_match "Analytics is disabled", cmd("analytics")
   end
+
+  def test_switch
+    assert_match "Usage: brew switch <name> <version>", cmd_fail("switch")
+    assert_match "testball not found", cmd_fail("switch", "testball", "0.1")
+
+    setup_test_formula "testball", <<-EOS.undent
+      keg_only "just because"
+    EOS
+
+    cmd("install", "testball")
+    testball_rack = HOMEBREW_CELLAR/"testball"
+    FileUtils.cp_r testball_rack/"0.1", testball_rack/"0.2"
+
+    cmd("switch", "testball", "0.2")
+    assert_match "testball does not have a version \"0.3\"",
+      cmd_fail("switch", "testball", "0.3")
+  end
 end


MikeMcQuaid:
> Since these lines return exit codes that fall outside the purview of cmd_fail, would it be OK to add another helper method to address them?

I think it'd be worth just handling all non-zero codes the same.

MikeMcQuaid:
Good work. Again, waiting for @UniqMartin 😀

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -706,4 +706,18 @@ def test_irb
...
+  def test_switch
+    assert_match "Usage: brew switch <name> <version>", cmd_fail("switch")
+
+    setup_test_formula "testball", <<-EOS.undent
+      keg_only "just because"
+    EOS
+
+    cmd("install", "testball")
+    testball_keg = HOMEBREW_CELLAR/"testball"
+    FileUtils.cp_r testball_keg/"0.1", testball_keg/"0.2"
```
UniqMartin: It's a bit of a technicality, but this should be`testball_rack`. Here's why:

- `HOMEBREW_CELLAR` - cellar (home to all installed formulae)
- `HOMEBREW_CELLAR/name` – rack (home to all installed versions of a formula)
- `HOMEBREW_CELLAR/name/pkg_version` - keg (home to a specific version plus revision of a formula)

The `name` and `pkg_verison` placeholders above are references to the corresponding `Formula` instance methods.

UniqMartin:
> > Since these lines return exit codes that fall outside the purview of `cmd_fail`, would it be OK to add another helper method to address them?
>
> I think it'd be worth just handling all non-zero codes the same.

Agreed. I guess we're not yet covering those cases with a non-zero exit status just yet, so they still need to be added to the test, right? Otherwise I'm basically happy with the code as-is.

Library/Homebrew/test/test_integration_cmds.rb
eirinikos: Makes sense! Thank you for pointing this out and tying it all together.

eirinikos:
> > > Since these lines return exit codes that fall outside the purview of `cmd_fail`, would it be OK to add another helper method to address them?
> > 
> > I think it'd be worth just handling all non-zero codes the same.
> 
> Agreed. I guess we're not yet covering those cases with a non-zero exit status just yet, so they still need to be added to the test, right? Otherwise I'm basically happy with the code as-is.

Thanks. Yes, that's right. I've pushed a new commit in which `cmd_fail` is modified in a way that seems clunky to me, but maybe it's acceptable? Looking forward to hearing your thoughts. And I've added a couple of `cmd_fail`s to address the lines in `brew switch` that return exit codes 2 and 3. I'm guessing that we'd want to split the `cmd_fail` changes from the `test_switch` changes, but first I'll wait for feedback. 😺 

I finally realized that `cmd` and `cmd_fail` make their own assertions (even though I'd seen them, I never really registered their existence). And I've seen that simply using `cmd` and `cmd_fail` often seems to render a separate `assert_match` (or other assertion) as unnecessary. Would it still be better (i.e., more informative or helpful) to use these assertions even when they're not strictly necessary?

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -102,8 +102,9 @@ def cmd(*args)
   def cmd_fail(*args)
     output = cmd_output(*args)
     status = $?.exitstatus
-    $stderr.puts "\n#{output}" if status != 1
-    assert_equal 1, status
+    $stderr.puts "\n#{output}" if status < 1 || status > 3
+    assert_operator status, :>, 0
+    assert_operator status, :<, 4
```
UniqMartin: The implicit assumption that is always made on the command line is that any nonzero exit code counts as failure. For this reason, the best way to address the too restrictive assumption previously made here would be to simply consider every nonzero exit code a failure (in the sense that the command failed) and thus a success (in terms of that the failure was expected in the test).

Or to be a bit more specific, this should do (and yes, ideally is done in a separate commit):

```ruby
$stderr.puts "\n#{output}" if status == 0
refute_equal 0, status
```

UniqMartin:
Since most of your PRs end up modifying the end of the `test_integration_cmds.rb` file, every merged PR is going to cause another to have merge conflicts. I guess that's a nice way to have more practice with `git rebase`. 😸 In the future, those kinds of conflicts could be minimized if the test methods were roughly added in alphabetical order instead of appending them to the existing ones. (But there's nothing fundamentally wrong with how you've done things thus far.)

> I've pushed a new commit in which `cmd_fail` is modified in a way that seems clunky to me, but maybe it's acceptable? Looking forward to hearing your thoughts.

I've tried to address this in a separate code comment.

> And I've added a couple of `cmd_fail`s to address the lines in `brew switch` that return exit codes 2 and 3.

Those look basically fine to me.

> I'm guessing that we'd want to split the `cmd_fail` changes from the `test_switch` changes, but first I'll wait for feedback.

Yes, that would be perfect!

> And I've seen that simply using `cmd` and `cmd_fail` often seems to render a separate `assert_match` (or other assertion) as unnecessary. Would it still be beneficial (i.e., more informative or helpful) to use these assertions even when they're not strictly necessary?

This always depends a bit on the context. In some cases the command output itself isn't very useful and checking it additionally doesn't provide any added value. But more often than not the expected failure cannot be trivially inferred from the invocation alone, thus an assert on the error message provides an additional clue as to what the test is checking. (And the additional assert is basically free, as it is significantly less expensive than the `cmd[_fail]` call.)

eirinikos:
> Since most of your PRs end up modifying the end of the `test_integration_cmds.rb` file, every merged PR is going to cause another to have merge conflicts. I guess that's a nice way to have more practice with `git rebase`. 😸 In the future, those kinds of conflicts could be minimized if the test methods were roughly added in alphabetical order instead of appending them to the existing ones. (But there's nothing fundamentally wrong with how you've done things thus far.)

Good idea, thanks! I'll keep this in mind for future reference. In the meantime, yes, it has been good to have more practice with `git rebase` (which is exactly what I've just done for this branch). 😄 

> In some cases the command output itself isn't very useful and checking it additionally doesn't provide any added value. But more often than not the expected failure cannot be trivially inferred from the invocation alone, thus an assert on the error message provides an additional clue as to what the test is checking. (And the additional assert is basically free, as it is significantly less expensive than the `cmd[_fail]` call.)

Thanks for the explanation!

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -756,4 +756,20 @@ def test_analytics
...
+    setup_test_formula "testball", <<-EOS.undent
+      keg_only "just because"
+    EOS
+
+    cmd("install", "testball")
+    testball_rack = HOMEBREW_CELLAR/"testball"
+    FileUtils.cp_r testball_rack/"0.1", testball_rack/"0.2"
+
+    cmd("switch", "testball", "0.2")
+    cmd_fail("switch", "testball", "0.3")
```
UniqMartin: I think this would also benefit from an additional `assert_match` on the command output. If I'm not mistaken, it's checking [this part](https://github.com/Homebrew/brew/blob/39453691ba3208dbd2e6ce1fd341d430b31dff55/Library/Homebrew/cmd/switch.rb#L25-L33) of the `brew switch` code, which isn't entirely obvious from the `cmd_fail` call alone.

eirinikos: Makes sense! 👍

UniqMartin: Thanks! Much better now. 😸

UniqMartin:
@MikeMcQuaid Any final comments? I'm happy with this PR as-is and think its ready to merge.

MikeMcQuaid:
Thanks again @eirinikos 🎉

PR #562, opened 2016-07-20
formula_dir update

AnastasiaSulyagina:

-----
Required for `brew update` to work properly with deleted and moved casks (for context see [#15603](https://github.com/caskroom/homebrew-cask/issues/15603))

cc @MikeMcQuaid

### DIFF
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -299,7 +299,7 @@ class Tap
 
   # path to the directory of all {Formula} files for this {Tap}.
   def formula_dir
-    @formula_dir ||= [path/"Formula", path/"HomebrewFormula", path].detect(&:directory?)
+    @formula_dir ||= [path/"Formula", path/"HomebrewFormula", path/"Casks", path].detect(&:directory?)
   end
 
   # an array of all {Formula} files of this {Tap}.
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -70,6 +70,7 @@ class Tap
   def clear_cache
     @remote = nil
     @formula_dir = nil
+    @cask_dir = nil
     @formula_files = nil
     @alias_dir = nil
     @alias_files = nil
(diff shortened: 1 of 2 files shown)


Library/Homebrew/tap.rb
```diff
@@ -299,7 +299,7 @@ def custom_remote?
 
   # path to the directory of all {Formula} files for this {Tap}.
   def formula_dir
-    @formula_dir ||= [path/"Formula", path/"HomebrewFormula", path].detect(&:directory?)
+    @formula_dir ||= [path/"Formula", path/"HomebrewFormula", path/"Casks", path].detect(&:directory?)
```
MikeMcQuaid: Probably worth adding a new function `cask_dir` and use that instead?

Library/Homebrew/tap.rb
```diff
@@ -302,6 +303,11 @@ def formula_dir
     @formula_dir ||= [path/"Formula", path/"HomebrewFormula", path].detect(&:directory?)
   end
 
+  # path to the directory of all casks for caskroom/cask {Tap}.
+  def cask_dir
+    @cask_dir ||= [path/"Casks", path].detect(&:directory?)
```
xu-cheng: This should just be `path/Casks` as I don't remember cask supporting put cask file in the top directory, right?

Library/Homebrew/tap.rb
```diff
@@ -317,7 +323,7 @@ def formula_files
   def formula_file?(file)
     file = Pathname.new(file) unless file.is_a? Pathname
     file = file.expand_path(path)
-    file.extname == ".rb" && file.parent == formula_dir
+    file.extname == ".rb" && (file.parent == formula_dir || file.parent == cask_dir)
```
xu-cheng: 👎 I suggest you to create a new method `cask_file?`

Library/Homebrew/tap.rb
```diff
@@ -561,6 +567,14 @@ def formula_dir
   end
 
   # @private
+  def cask_dir
+    @cask_dir ||= begin
+      self.class.ensure_installed!
+      super
+    end
+  end
```
xu-cheng: Core tap, i.e.`homebrew/core` doesn't have casks. So you don't need this method.

MikeMcQuaid:
Thanks again @AnastasiaSulyagina! 🎉

PR #566, opened 2016-07-21
Option: also store value

xu-cheng:
This fixes long term problem that `brew install` fails to remember the
value part of args like `--with-foo=bar` passed by users. As result,
such value will not be preserved when invoking `brew reinstall` or
`brew upgrade`.

A real formula example can be `homebrew/science/abyss`.

Since there may have two options sharing the same name but different
values. We will have to use Hash instead of set for `Options` object.
The Hash will be a map between option name to option object. When
merging options under the same name, we choose to keep old option if
possible unless the new option has a value while the old one doesn't.

### DIFF
--- a/Library/Homebrew/options.rb
+++ b/Library/Homebrew/options.rb
@@ -1,11 +1,13 @@
 require "set"
 
 class Option
-  attr_reader :name, :description, :flag
+  attr_reader :name, :value, :description, :flag
 
-  def initialize(name, description = "")
-    @name = name
-    @flag = "--#{name}"
+  def initialize(arg, description = "")
+    arg = arg.match(/^([^=]+=?)(.+)?$/)
+    @name = arg[1]
+    @value = arg[2]
+    @flag = "--#{name}#{value}"
     @description = description
   end
 
@@ -15,16 +17,16 @@ class Option
 
   def <=>(other)
     return unless Option === other
-    name <=> other.name
+    flag <=> other.flag
   end
 
   def ==(other)
-    instance_of?(other.class) && name == other.name
+    instance_of?(other.class) && flag == other.flag
   end
   alias_method :eql?, :==
 
   def hash
-    name.hash
+    flag.hash
   end
 
   def inspect
@@ -57,41 +59,60 @@ end
 class Options
   include Enumerable
 
-  def self.create(array)
-    new array.map { |e| Option.new(e[/^--([^=]+=?)(.+)?$/, 1] || e) }
+  attr_reader :options
+  protected :options
+
+  def self.create(array = [])
+    options = Hash.new
+    array.each do |option|
+      option = Option.new(option.strip_prefix("--")) unless option.is_a?(Option)
+      options[option.name] = option
+    end
+    new options
   end
 
-  def initialize(*args)
-    @options = Set.new(*args)
+  # We store options in a Hash (option name => option object),
+  # because they may have different value under the same name.
+  def initialize(options)
+    @options = options
   end
 
   def each(*args, &block)
-    @options.each(*args, &block)
+    @options.each_value(*args, &block)
   end
 
   def <<(o)
-    @options << o
+    opt = @options[o.name]
+    if opt
+      @options[o.name] = o if opt.value.nil? && !o.value.nil?
+    else
+      @options[o.name] = o
+    end
     self
   end
 
-  def +(o)
-    self.class.new(@options + o)
+  def +(other)
+    new_options = @options.merge(other.options) do |_name, old_opt, new_opt|
+      if old_opt.value.nil? && !new_opt.value.nil?
+        new_opt
+      else
+        old_opt
+      end
+    end
+    self.class.new new_options
+  end
+  alias_method :|, :+
+
+  def -(other)
+    self.class.new Hash[@options.reject { |name, _option| other.options.key?(name) }]
   end
 
-  def -(o)
-    self.class.new(@options - o)
-  end
-
-  def &(o)
-    self.class.new(@options & o)
-  end
-
-  def |(o)
-    self.class.new(@options | o)
+  def &(other)
+    self.class.new Hash[@options.select { |name, _option| other.options.key?(name) }]
   end
 
   def *(arg)
-    @options.to_a * arg
+    @options.values * arg
   end
 
   def empty?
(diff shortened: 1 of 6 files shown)


Library/Homebrew/options.rb
```diff
@@ -1,11 +1,12 @@
...
 class Option
-  attr_reader :name, :description, :flag
+  attr_reader :name, :value, :description, :flag
 
-  def initialize(name, description = "")
-    @name = name
-    @flag = "--#{name}"
+  def initialize(arg, description = "")
+    @name = arg[/^[^=]+=?/]
+    @value = arg[@name.size..-1]
```
MikeMcQuaid: Would be nice to capture both of these with a single regex.

MikeMcQuaid:
One nit otherwise 👍

xu-cheng:
Actually, rethink a bit, I find this solution of this PR is a little dirty and may cause some problem during option merging. I will explore if there is a better way.

xu-cheng:
SO I have rewritten this PR. More tests will be added.

Library/Homebrew/options.rb
```diff
@@ -15,16 +17,16 @@ def to_s
 
   def <=>(other)
     return unless Option === other
-    name <=> other.name
+    (name <=> other.name).nonzero? || (value <=> other.value)
```
MikeMcQuaid: Splitting this into a `return` on the line before would be nice.

Library/Homebrew/options.rb
```diff
@@ -15,16 +17,16 @@ def to_s
...
   end
 
   def ==(other)
-    instance_of?(other.class) && name == other.name
+    instance_of?(other.class) && name == other.name && value == other.value
```
MikeMcQuaid: Splitting this into two other returns would be nice.

Library/Homebrew/options.rb
```diff
@@ -15,16 +17,16 @@ def to_s
...
   end
   alias_method :eql?, :==
 
   def hash
-    name.hash
+    name.hash ^ value.hash
```
MikeMcQuaid: It's not obvious what this does; can you explain?

Library/Homebrew/test/test_options.rb
```diff
@@ -56,7 +56,7 @@ def test_equality
 
 class OptionsTests < Homebrew::TestCase
   def setup
-    @options = Options.new
+    @options = Options.create
```
MikeMcQuaid: Is it possible to avoid changing this?

Library/Homebrew/options.rb
xu-cheng: Fixed by using `flag` directly.

Library/Homebrew/options.rb
xu-cheng: Fixed by using `flag` directly.

Library/Homebrew/options.rb
xu-cheng: Fixed by using `flag` directly.

Library/Homebrew/test/test_options.rb
xu-cheng: The idea is that `Options.new` will become a private implementation detail after this PR, it should not be called outside of `Options`.

MikeMcQuaid: What's the advantage of that? Feels better to move the logic into the current constructor.

xu-cheng: It cannot. The default constructor is used to create object directly for the output of operator like `|`, `&`, etc. While `create` method is used to initialize `Options` from an array of flags. They are different.

MikeMcQuaid: Can you elaborate on why `new` and `create` need to have different behaviour? Will you make `new` private? If not, the API seems pretty confusing.

UniqMartin: I have to say that I was wondering the same and it feels like this is perfectly doable within the regular constructor (so there would be no need to have a `create` method).

xu-cheng: * First of all, although `Options` is used to present an array of flags, in order to handle value part for flags like `--with-foo=bar`, the options are stored in a `Hash` internally.
* `Options.create` is used to initialize Options from an array of flags. It will do some preprocess to generate the hash object we will used to present this array of flags.
* `Options.new` will just take hash object directly to create a new options object. This is useful for operators like `+`, `-`, `&`, etc. (See the logic of these methods for example).

>  Will you make new private?

Yes. But how to tag it as private?

UniqMartin: Why cannot the constructor make the distinction between what is being passed to it and then act accordingly? (Sorry, I haven't found the time to carefully look at the code, but a cursory glance didn't reveal anything that would make this behavior impossible.)

> Yes. But how to tag it as private?

```ruby
private_class_method :new
```

MikeMcQuaid: Agreed 👍

xu-cheng: > Why cannot the constructor make the distinction between what is being passed to it and then act accordingly? (Sorry, I haven't found the time to carefully look at the code, but a cursory glance didn't reveal anything that would make this behavior impossible.)

We certainly can. The only question is will it hurts the readability and usage. As it will look like something like this:
```ruby
def initialize(args)
  case args
  when Array then .... # input is `["--foo=A", "--bar"]`
  when Hash then ....  # input is `{"--foo" => Option("--foo=A"), "--bar" => Option("--bar") }`
  end
end
```
I don't think we should expose this to the outside. As it's worthy to mention that `Options.create` has been there long before this PR.

MikeMcQuaid: That syntax seems better to me. Another option is to have `Option.from_hash` and `Options.from_array` methods.



> As it's worthy to mention that Options.create has been there long before this PR.

Sure and it may be worth removing. My main complaints are that you're have to alter so many tests to use it and stating that it shouldn't be used without using anything like `private` to ensure it is not.

MikeMcQuaid:
@xu-cheng Gentle ping on this.

chaimleib:
I think this PR was closed by accident. #4870 didn't resolve and close this PR; it modified the recipe to avoid our problem. Could someone reopen this ticket?

MikeMcQuaid:
@chaimleib It was closed because the comments weren't addressed for months. Could you try and open a new pull request? [This document](https://github.com/Homebrew/brew/blob/master/docs/How-To-Open-a-Homebrew-Pull-Request-(and-get-it-merged).md#how-to-open-a-homebrew-pull-request-and-get-it-merged) should help and I'm happy to walk you through anything else.

Thanks!

Issue #540, opened 2016-07-17
audit: suggest full formula name only for other taps


ilovezfs:
Since getdp is in homebrew/science the `:linked` below should not be printed as full names.

http://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/5157/version=mavericks/console

```
==> FAILED
Error: 1 problem in 1 formula
homebrew/science/getdp:
  * Formulae are required to declare all linked dependencies.
    Please add all linked dependencies to the formula with:
      depends_on "fftw" => :linked
      depends_on "homebrew/science/hwloc" => :linked
      depends_on "homebrew/science/netcdf" => :linked
      depends_on "homebrew/science/parmetis" => :linked
      depends_on "homebrew/science/scalapack" => :linked
      depends_on "homebrew/science/suite-sparse" => :linked
      depends_on "homebrew/science/sundials" => :linked
      depends_on "homebrew/science/superlu_dist" => :linked
```

CC @xu-cheng

xu-cheng:
we could probably do this:
```diff
diff --git a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
index dd21559..de61fb5 100644
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -72,10 +72,11 @@ module FormulaCellarChecks
     end
 
     if checker.undeclared_deps?
+      undeclared_deps = checker.undeclared_deps.map { |d| d.strip_prefix("#{formula.tap}/") }
       audit_check_output <<-EOS.undent
         Formulae are required to declare all linked dependencies.
         Please add all linked dependencies to the formula with:
-          #{checker.undeclared_deps.map { |d| "depends_on \"#{d}\" => :linked"} * "\n          "}
+          #{undeclared_deps.map { |d| "depends_on \"#{d}\" => :linked"} * "\n          "}
       EOS
     end
   end
```

ilovezfs:
Should work but seems a tad indirect :)

MikeMcQuaid:
I think it's fine as-is but don't feel strongly either way.

ilovezfs:
@xu-cheng I'd ship it with the change you suggested. Contributors are already spontaneously manually intervening and editing out the tap prefixes, which is less than ideal CX (contributor experience).

[...]

PR #572, opened 2016-07-22
check_linkage: print full name for other taps only

ilovezfs:
`depends_on` only requires full name when the dependency is a member of
some other tap. Even though redundant full names work, they are rarely
used in practice, so to avoid confusion, it would be best for
`brew audit` to follow the use-non-full-name-when-possible convention
when it prints undeclared linkage failures.

For example, if the formula being audited is in homebrew/science, print

```
      depends_on "homebrew/dupes/ncurses" => :linked
      depends_on "htslib" => :linked
```

not

```
      depends_on "homebrew/dupes/ncurses" => :linked
      depends_on "homebrew/science/htslib" => :linked
```

Note that this change is irrelevant with respect to core formulae since
their full names are not prefixed with "homebrew/core" anyway.

Thanks to Xu Cheng for contributing the code in this commit.

Closes https://github.com/Homebrew/brew/issues/540.

### DIFF
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -72,10 +72,11 @@ module FormulaCellarChecks
     end
 
     if checker.undeclared_deps?
+      undeclared_deps = checker.undeclared_deps.map { |d| d.strip_prefix("#{formula.tap}/") }
       audit_check_output <<-EOS.undent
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @xu-cheng

xu-cheng:
:+1:

UniqMartin:
:+1:

ilovezfs:
Since I think the underlying changes that this tweaks should be fully reverted, I'm closing this PR, which would be one more commit to have to revert.

PR #576, opened 2016-07-23
audit: only check undeclared deps for standard installations.

xu-cheng:
In fact, we don't really care about undeclared dependencies for optional
installations. Because, this is mainly used to help us to detect breakage
for bottles so we can do a revision bump.

cc @DomT4

### DIFF
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -71,6 +71,9 @@ module FormulaCellarChecks
       EOS
     end
 
+    # only check undeclared deps for standard installations.
+    return unless formula.build.used_options.empty?
+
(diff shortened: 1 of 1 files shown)


DomT4:
This does cut out a lot of the false positives, which is a 👍 for reducing user confusion & me spending less time repeatedly explaining this particular edge case. I remain unconvinced this particular `:linked` check is really a net gain for us at this point, but this at least eases off one edge case.

xu-cheng:
> I remain unconvinced this particular :linked check is really a net gain for us at this point

It is important to have it so we can avoid missing formulae whenever there is a version bump.

MikeMcQuaid:
> It is important to have it so we can avoid missing formulae whenever there is a version bump.

@xu-cheng In which case: you should be helping explain this to contributors who are struggling with this in homebrew/core.

PR #581, opened 2016-07-25
utils/github: fix broken pipe error

### LINKED ISSUE #573: Error: Broken pipe when running brew search
I'm running OSX 10.11.6 with Homebrew. I've just run `brew update`, so it should be up to date.

When I try to search for any packages, I get a `broken pipe` error at the end:
```
bash-3.2$ brew search csvkit
Error: Broken pipe
```
This is the output of `brew doctor`:
```
bash-3.2$ brew doctor
Your system is ready to brew.
```

Any things I should check?

UniqMartin:

-----

Steps to reproduce:

- Unset any possibly set `HOMEBREW_GITHUB_*` environment variables.
- Switch to a Git without `git-credential-osxkeychain` or temporarily rename this binary.

Before:

```console
$ brew search php
Error: Broken pipe
```

After:

```console
$ brew search php
No formula found for "php".
==> Searching pull requests...
Closed pull requests:
n98-magerun: migrate to php (https://github.com/Homebrew/homebrew-core/pull/1534)
Error: GitHub API Error: API rate limit exceeded for A.B.C.D. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
Try again in 25 minutes and 34 seconds, or create a personal access token:
  https://github.com/settings/tokens/new?scopes=&description=Homebrew
and then set the token as: export HOMEBREW_GITHUB_API_TOKEN="your_new_token"
```

Closes #573.

**Update:** Slightly expanded the scope of the PR [as mentioned in a comment below](https://github.com/Homebrew/brew/pull/581#issuecomment-234932442).

### DIFF
--- a/Library/Homebrew/utils/github.rb
+++ b/Library/Homebrew/utils/github.rb
@@ -50,7 +50,7 @@ module GitHub
       if ENV["HOMEBREW_GITHUB_API_TOKEN"]
         ENV["HOMEBREW_GITHUB_API_TOKEN"]
       elsif ENV["HOMEBREW_GITHUB_API_USERNAME"] && ENV["HOMEBREW_GITHUB_API_PASSWORD"]
-        [ENV["HOMEBREW_GITHUB_API_USERNAME"], ENV["HOMEBREW_GITHUB_API_PASSWORD"]]
+        [ENV["HOMEBREW_GITHUB_API_PASSWORD"], ENV["HOMEBREW_GITHUB_API_USERNAME"]]
       else
         github_credentials = api_credentials_from_keychain
         github_username = github_credentials[/username=(.+)/, 1]
--- a/Library/Homebrew/utils/github.rb
+++ b/Library/Homebrew/utils/github.rb
@@ -52,11 +52,7 @@ module GitHub
       elsif ENV["HOMEBREW_GITHUB_API_USERNAME"] && ENV["HOMEBREW_GITHUB_API_PASSWORD"]
         [ENV["HOMEBREW_GITHUB_API_USERNAME"], ENV["HOMEBREW_GITHUB_API_PASSWORD"]]
       else
-        github_credentials = Utils.popen("git credential-osxkeychain get", "w+") do |io|
-          io.puts "protocol=https\nhost=github.com"
-          io.close_write
-          io.read
-        end
+        github_credentials = api_credentials_from_keychain
         github_username = github_credentials[/username=(.+)/, 1]
         github_password = github_credentials[/password=(.+)/, 1]
         if github_username && github_password
@@ -66,6 +62,20 @@ module GitHub
         end
       end
     end
+  end
+
+  def api_credentials_from_keychain
+    Utils.popen(["git", "credential-osxkeychain", "get"], "w+") do |pipe|
+      pipe.write "protocol=https\nhost=github.com\n"
+      pipe.close_write
+      pipe.read
+    end
+  rescue Errno::EPIPE
+    # The above invocation via `Utils.popen` can fail, causing the pipe to be
+    # prematurely closed (before we can write to it) and thus resulting in a
+    # broken pipe error. The root cause is usually a missing or malfunctioning
+    # `git-credential-osxkeychain` helper.
+    ""
   end
 
   def api_credentials_type


xu-cheng:
Can we do:
```ruby
begin
 github_credentials = Utils.popen ...
rescue
end
```

Or
```ruby
 github_credentials = Utils.popen (...) rescue ""
```

ilovezfs:
Would this be solved by the proposal to vendor git?

UniqMartin:
> Can we do:

Yes, if that seems preferable. The current version is optimized for a minimal diff (no need for a dedicated `begin` block and adjustment of the indentation) and since no other part of the code is doing I/O it's essentially the same. (But I can definitely see the benefit of bringing the `rescue` closer to the source of the exception.)

> Or

No. RuboCop objects to `rescue` as a modifier and I agree. It's too important to be stuffed away at the end of the line where it's easier to overlook and the modifier doesn't allow to specify the exceptions that should be handled, making it more likely to unintentionally handle other unrelated exceptions that shouldn't be silently swallowed.

> Would this be solved by the proposal to vendor git?

I think vendoring Git would reduce the likelihood of this happening, but I'm pretty sure there can be other error conditions that lead to an early exit without accepting any input even if the helper is present. (And I think the vendored Git was only meant to be used if the host Git is deemed too old, so it wouldn't actually help in many cases—unless I'm misremembering this part.)

ilovezfs:
It seems odd that we're getting a broken pipe instead of an explicit error? Is this caused by some fragility in Utils.popen?

UniqMartin:
Our version of `popen` redirects standard error to `/dev/null`, thus any error messages will be discarded, but the result of the invocation (the exit code) will still be reflected in `$?` after the block given to `popen` terminates.

The process we're talking to is free to terminate or stop accepting input at any time, thus the broken pipe error can literally happen in the middle of a `puts`/`write` operation and (to my knowledge) there isn't a more elegant/correct way of handling this than rescuing `Errno::EPIPE`.

ilovezfs:
Would this happen if `git credential-osxkeychain get` was executed in a shell instead?

UniqMartin:
This entirely depends on the program that is feeding input to the `git credential-osxkeychain get` process and how it decided to handle or not handle this case. The default signal handler for `SIGPIPE` is to simply kill the program as soon as a `write` fails due to a broken pipe. The other option is to have a custom handler for `SIGPIPE` or to mask this signal. The latter causes `write` to return `EPIPE` which is translated to the exception handled here.

So yes, in some sense this is also happening if executed in a shell, but we don't usually see the broken pipe error as it isn't considered a true error in most circumstances. E.g. in `cat file.txt | head -n 1` it is actually desirable that `head` dies after it has output the first line, so that the pipe is broken and `cat` can learn its pointless to continue reading the input file. (Whether `cat` internally relies on the default `SIGPIPE` behavior or checks for the `EPIPE` return value I don't know.)

Because the Homebrew process is doing a lot more than just feeding input to that helper (unlike in the `cat` example above), it makes sense to actually handle the broken pipe and then move on instead of being killed by the OS.

ilovezfs:
Why it preferable to return `[]` rather than give an explicit, custom error to the users? Won't the empty search output be confusing when they know what they're searching for does indeed exist?

UniqMartin:
> Won't the empty search output be confusing when they know what they're searching for does indeed exist?

It's not the search result that is empty, but the API credentials. Consequently, the code will continue to make unauthenticated GitHub API requests and either return the desired search results or report back with an error that the rate limit has been exceeded. That error also suggests to create and set a token, thus a possibly incomplete search result shouldn't be causing much confusion—certainly less than the broken pipe error. (See my top comment for an example output for the rate limit case.)

> Why it preferable to return `[]` rather than give an explicit, custom error to the users?

Because it is just one of several options and not having the credentials available this way isn't a fatal condition. We can still try and often succeed with an unauthenticated request. Maybe it's an unusual configuration on macOS, but it's still perfectly legal to have a Git installation without this credentials helper and making its absence stop the rest of the processing in its tracks didn't feel appropriate. If we wanted to make the user aware of this issue, I think the best thing we can do is output a warning. (I'm happy to make adjustments to the error reporting if my assessment of the situation isn't as accurate as I believe it to be.)

ilovezfs:
That makes sense. However, at least on macOS, I'm not sure why we'd want to support a git lacking this feature. We could simply require the use of an appropriate git, no?

ilovezfs:
In any case, I think we should consider requiring a git that has this feature, but it seems fine to :ship: this since you were saying you believe this can occur in some other scenario besides lacking the helper too.

UniqMartin:
Updated!

1. Refactored the fix to address feedback on code style/organization and moved some work to a helper method to avoid excessively deep nesting due to another `begin` block.

2. Did some minor refactoring to simplify the extraction code.

3. Fixed wrong element order when using `HOMEBREW_GITHUB_API_{USERNAME,PASSWORD}`. (Currently undocumented and probably not very widely used, if nobody noticed that the password and the user name were swapped.)

I'm happy to move the 2nd and/or 3rd commit to a separate PR, but I think this PR still makes sense with these additional changes on top if reviewed commit by commit.

ilovezfs:
Should this go in a separate method, too?

```
+    credentials = Hash[credentials.scan(/^(username|password)=(.*)$/)]
+    if credentials.key?("username") && credentials.key?("password")
+      credentials.values_at("password", "username")
+    else
+      []
+    end
```

UniqMartin:
How would further splitting this up benefit the code?

ilovezfs:
The rescue still seems to be a bit separated from where @xu-cheng was actually saying he'd prefer it be, no?

Library/Homebrew/utils/github.rb
```diff
@@ -50,24 +50,34 @@ def api_credentials
...
-          [github_password, github_username]
-        else
-          []
-        end
+        api_credentials_from_keychain("https", "github.com")
       end
     end
   end
 
+  def api_credentials_from_keychain(protocol, host)
```
MikeMcQuaid: I wonder if we need these args given that this only ever makes sense for HTTPS and GitHub.

Library/Homebrew/utils/github.rb
```diff
@@ -50,24 +50,34 @@ def api_credentials
...
+    credentials = Utils.popen("git credential-osxkeychain get", "w+") do |pipe|
+      pipe.write "protocol=#{protocol}\nhost=#{host}\n"
+      pipe.close_write
+      pipe.read
+    end
+
+    credentials = Hash[credentials.scan(/^(username|password)=(.*)$/)]
```
MikeMcQuaid: I think the `Hash[]` approach is less obvious in what it's actually doing, personally.

Library/Homebrew/utils/github.rb
```diff
@@ -50,24 +50,34 @@ def api_credentials
       if ENV["HOMEBREW_GITHUB_API_TOKEN"]
         ENV["HOMEBREW_GITHUB_API_TOKEN"]
       elsif ENV["HOMEBREW_GITHUB_API_USERNAME"] && ENV["HOMEBREW_GITHUB_API_PASSWORD"]
-        [ENV["HOMEBREW_GITHUB_API_USERNAME"], ENV["HOMEBREW_GITHUB_API_PASSWORD"]]
+        [ENV["HOMEBREW_GITHUB_API_PASSWORD"], ENV["HOMEBREW_GITHUB_API_USERNAME"]]
```
MikeMcQuaid: Why reverse the order, here?

MikeMcQuaid:
A few comments otherwise 👍 from me.

Library/Homebrew/utils/github.rb
UniqMartin: They made it easier to locally test my fix because I don't have the GitHub credentials in my Keychain, but I do have other credentials accessible by the Git credentials helper. That said, I'll remove the arguments.

Library/Homebrew/utils/github.rb
UniqMartin: I'll drop the refactoring commit and revert to the previous code.

Library/Homebrew/utils/github.rb
UniqMartin: Because this fixes a bug. Please see the commit message of the commit that changes this and the way the return value of `api_credentials` is used elsewhere.

MikeMcQuaid: Cool. Would be good to put stuff like that in the PR body too, I think.

UniqMartin:
Updated to address all comments and to drop the refactoring due to questionable benefit.

Library/Homebrew/utils/github.rb
MikeMcQuaid: Making it easier to locally test your fix is a good enough reason for me 👍

Library/Homebrew/utils/github.rb
UniqMartin: I did in https://github.com/Homebrew/brew/pull/581#issuecomment-234932442 and updated the PR body to reference this comment. What else could/should I have done? (Only noticed the bug after addressing the first feedback.)

MikeMcQuaid: That's more than enough, I missed it, thanks!

MikeMcQuaid:
LGTM 👍

UniqMartin:
> The rescue still seems to be a bit separated from where @xu-cheng was actually saying he'd prefer it be, no?

@ilovezfs Thanks for reminding me of that! Are you happy with the current simplified diff?

ilovezfs:
@UniqMartin this looks great.

UniqMartin:
Thanks everyone for your feedback and your patience while I iterated through some revisions of this PR that were less than ideal! I'm now much more happy with the outcome.

PR #584, opened 2016-07-25
[RFC] Allow HEAD-upgrades

vladshablinsky:

-----

This is a copy of #274, but not filled with old comments and suggestions regarding the project in general, so the maintainers can leave their feedback.

 The further work on GSoC project will be continued here.

### DIFF
--- a/Library/Homebrew/cmd/outdated.rb
+++ b/Library/Homebrew/cmd/outdated.rb
@@ -32,10 +32,19 @@ module Homebrew
 
   def print_outdated(formulae)
     verbose = ($stdout.tty? || ARGV.verbose?) && !ARGV.flag?("--quiet")
+    fetch_head = ARGV.fetch_head?
 
-    formulae.select(&:outdated?).each do |f|
+    outdated_formulae = formulae.select { |f| f.outdated?(:fetch_head => fetch_head) }
+
+    outdated_formulae.each do |f|
       if verbose
-        puts "#{f.full_name} (#{f.outdated_versions*", "} < #{f.pkg_version})"
+        outdated_versions = f.outdated_versions(:fetch_head => fetch_head)
+        current_version = if f.head? && outdated_versions.any? { |v| v.to_s == f.pkg_version.to_s }
+          "latest HEAD"
+        else
+          f.pkg_version.to_s
+        end
+        puts "#{f.full_name} (#{outdated_versions.join(", ")}) < #{current_version}"
       else
         puts f.full_name
       end
@@ -44,11 +53,20 @@ module Homebrew
 
   def print_outdated_json(formulae)
     json = []
-    outdated = formulae.select(&:outdated?).each do |f|
+    fetch_head = ARGV.fetch_head?
+    outdated_formulae = formulae.select { |f| f.outdated?(:fetch_head => fetch_head) }
+
+    outdated = outdated_formulae.each do |f|
+      outdated_versions = f.outdated_versions(:fetch_head => fetch_head)
+      current_version = if f.head? && outdated_versions.any? { |v| v.to_s == f.pkg_version.to_s }
+        "HEAD"
+      else
+        f.pkg_version.to_s
+      end
 
       json << { :name => f.full_name,
-                :installed_versions => f.outdated_versions.collect(&:to_s),
-                :current_version => f.pkg_version.to_s }
+                :installed_versions => outdated_versions.collect(&:to_s),
+                :current_version => current_version }
     end
     puts Utils::JSON.dump(json)
 
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -19,10 +19,15 @@ module Homebrew
     Homebrew.perform_preinstall_checks
 
     if ARGV.named.empty?
-      outdated = Formula.installed.select(&:outdated?)
+      outdated = Formula.installed.select do |f|
+        f.outdated?(:fetch_head => ARGV.fetch_head?)
+      end
+
       exit 0 if outdated.empty?
     else
-      outdated = ARGV.resolved_formulae.select(&:outdated?)
+      outdated = ARGV.resolved_formulae.select do |f|
+        f.outdated?(:fetch_head => ARGV.fetch_head?)
+      end
 
       (ARGV.resolved_formulae - outdated).each do |f|
         versions = f.installed_kegs.map { |keg| keg.version }
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -913,7 +913,7 @@ class MercurialDownloadStrategy < VCSDownloadStrategy
   end
 
   def last_commit
-    Utils.popen_read("hg", "parent", "--template", "{node}", "-R", cached_location.to_s)
+    Utils.popen_read("hg", "parent", "--template", "{node|short}", "-R", cached_location.to_s)
   end
 
   private
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -6,6 +6,7 @@ class AbstractDownloadStrategy
   include FileUtils
 
   attr_reader :meta, :name, :version, :resource
+  attr_reader :shutup
 
   def initialize(name, resource)
     @name = name
@@ -17,6 +18,19 @@ class AbstractDownloadStrategy
 
   # Download and cache the resource as {#cached_location}.
   def fetch
+  end
+
+  # Supress output
+  def shutup!
+    @shutup = true
+  end
+
+  def puts(*args)
+    super(*args) unless shutup
+  end
+
+  def ohai(*args)
+    super(*args) unless shutup
   end
 
   # Unpack {#cached_location} into the current working directory, and possibly
@@ -57,6 +71,14 @@ class AbstractDownloadStrategy
     # 2 as default because commands are eg. svn up, git pull
     args.insert(2, "-q") unless ARGV.verbose?
     args
+  end
+
+  def safe_system(*args)
+    if @shutup
+      quiet_system(*args) || raise(ErrorDuringExecution.new(args.shift, *args))
+    else
+      super(*args)
+    end
   end
 
   def quiet_safe_system(*args)
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -147,6 +147,16 @@ class VCSDownloadStrategy < AbstractDownloadStrategy
     end
   end
 
+  def fetch_last_commit
+    fetch
+    last_commit
+  end
+
+  def commit_outdated?(commit)
+    @last_commit ||= fetch_last_commit
+    commit != @last_commit
+  end
+
   def cached_location
     @clone
   end
@@ -756,6 +766,45 @@ class GitDownloadStrategy < VCSDownloadStrategy
   end
 end
 
+class GitHubGitDownloadStrategy < GitDownloadStrategy
+  def initialize(name, resource)
+    super
+    if @url =~ %r{^https?://github\.com/([^/]+)/([^/]+)\.git$}
+      @user = $1
+      @repo = $2
+    end
+  end
+
+  def github_last_commit
+    return if ENV["HOMEBREW_NO_GITHUB_API"]
+
+    output, _, status = curl_output "-H", "Accept: application/vnd.github.v3.sha", \
+      "-I", "https://api.github.com/repos/#{@user}/#{@repo}/commits/#{@ref}"
+
+    commit = output[/^ETag: \"(\h+)\"/, 1] if status.success?
+    version.update_commit(commit) if commit
+    commit
+  end
+
+  def multiple_short_commits_exist?(commit)
+    return if ENV["HOMEBREW_NO_GITHUB_API"]
+    output, _, status = curl_output "-H", "Accept: application/vnd.github.v3.sha", \
+      "-I", "https://api.github.com/repos/#{@user}/#{@repo}/commits/#{commit}"
+
+    !(status.success? && output && output[/^Status: (200)/, 1] == "200")
+  end
+
+  def commit_outdated?(commit)
+    @last_commit ||= github_last_commit
+    if !@last_commit
+      super
+    else
+      return true unless @last_commit.start_with?(commit)
+      multiple_short_commits_exist?(commit)
+    end
+  end
+end
+
 class CVSDownloadStrategy < VCSDownloadStrategy
   def initialize(name, resource)
     super
@@ -957,6 +1006,8 @@ class DownloadStrategyDetector
 
   def self.detect_from_url(url)
     case url
+    when %r{^https?://github\.com/[^/]+/[^/]+\.git$}
+      GitHubGitDownloadStrategy
     when %r{^https?://.+\.git$}, %r{^git://}
       GitDownloadStrategy
     when %r{^https?://www\.apache\.org/dyn/closer\.cgi}, %r{^https?://www\.apache\.org/dyn/closer\.lua}
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1000,7 +1000,7 @@ class Formula
   end
 
   # @private
-  def outdated_versions
+  def outdated_versions(options = {})
     @outdated_versions ||= begin
       all_versions = []
 
@@ -1009,16 +1009,22 @@ class Formula
       installed_kegs.each do |keg|
         version = keg.version
         all_versions << version
-        return [] if pkg_version <= version
+
+        return [] if pkg_version <= version && !version.head?
       end
 
-      all_versions.sort!
+      head_version = latest_head_version
+      if head_version
+        head_version_outdated?(head_version, options) ? all_versions.sort! : []
+      else
+        all_versions.sort!
+      end
     end
   end
 
   # @private
-  def outdated?
-    !outdated_versions.empty?
+  def outdated?(options = {})
+    !outdated_versions(options).empty?
   rescue Migrator::MigrationNeededError
     true
   end
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -413,16 +413,36 @@ class Formula
     Pathname.new("#{HOMEBREW_LIBRARY}/LinkedKegs/#{name}")
   end
 
-  def latest_head_prefix
+  def latest_head_version
     head_versions = installed_prefixes.map do |pn|
       pn_pkgversion = PkgVersion.parse(pn.basename.to_s)
       pn_pkgversion if pn_pkgversion.head?
     end.compact
 
-    latest_head_version = head_versions.max_by do |pn_pkgversion|
+    head_versions.max_by do |pn_pkgversion|
       [Tab.for_keg(prefix(pn_pkgversion)).source_modified_time, pn_pkgversion.revision]
     end
-    prefix(latest_head_version) if latest_head_version
+  end
+
+  def latest_head_prefix
+    head_version = latest_head_version
+    prefix(head_version) if head_version
+  end
+
+  def head_version_outdated?(version, options={})
+    tab = Tab.for_keg(prefix(version))
+
+    return true if stable && tab.stable_version && tab.stable_version < stable.version
+    return true if devel && tab.devel_version && tab.devel_version < devel.version
+
+    if options[:fetch_head]
+      return false unless head && head.downloader.is_a?(VCSDownloadStrategy)
+      downloader = head.downloader
+      downloader.shutup! unless ARGV.verbose?
+      downloader.commit_outdated?(version.version.commit)
+    else
+      false
+    end
   end
 
   # The latest prefix for this formula. Checks for {#head}, then {#devel}
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -449,8 +449,8 @@ class Formula
   # and then {#stable}'s {#prefix}
   # @private
   def installed_prefix
-    if head && (head_prefix = latest_head_prefix) && head_prefix.directory?
-      head_prefix
+    if head && (head_version = latest_head_version) && !head_version_outdated?(head_version)
+      latest_head_prefix
     elsif devel && (devel_prefix = prefix(PkgVersion.new(devel.version, revision))).directory?
       devel_prefix
     elsif stable && (stable_prefix = prefix(PkgVersion.new(stable.version, revision))).directory?
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1001,24 +1001,28 @@ class Formula
 
   # @private
   def outdated_versions(options = {})
-    @outdated_versions ||= begin
-      all_versions = []
-
+    @outdated_versions ||= Hash.new do |cache, key|
       raise Migrator::MigrationNeededError.new(self) if migration_needed?
+      cache[key] = _outdated_versions(key)
+    end
+    @outdated_versions[options]
+  end
 
-      installed_kegs.each do |keg|
-        version = keg.version
-        all_versions << version
+  def _outdated_versions(options = {})
+    all_versions = []
 
-        return [] if pkg_version <= version && !version.head?
-      end
+    installed_kegs.each do |keg|
+      version = keg.version
+      all_versions << version
 
-      head_version = latest_head_version
-      if head_version
-        head_version_outdated?(head_version, options) ? all_versions.sort! : []
-      else
-        all_versions.sort!
-      end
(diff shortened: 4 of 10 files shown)


vladshablinsky:
# Problem

The question raised in this PR is using long git commit hashes and short git commit hashes.
The format of the HEAD-versions we use is the following: `HEAD-<version>_<revision>`, where version is a commit hash. There are plenty of download strategies, but I refer only to `GitDownloadStrategy` here, because others have either other 'commit' structure or no short-hashes support.

To compare two HEAD version we need to compare commit hashes. And this is what this PR is about. Suppose we only store long hashes. So, to detect if installed HEAD-version is outdated we need to compare upstream hash of the project with hash used in installed HEAD-version prefix. If they are the same, then installed HEAD is up-to-date, otherwise we have outdated HEAD.

Let's now suppose we use short hashes. What we do to detect outdated versions is a) calling `VCSDownloadStrategy#last_commit` which goes to cached location and gets the hash and b) compare returned `last_commit` with HEAD-prefix. If it any time `last_commit` increases its length, then installed HEAD-prefix cannot be equal to it anymore, thus, it's outdated.

More precisely:

```bash
$ ls /usr/local/Cellar/ack
HEAD-a0a0a0a

2.0.0-p481 :001 > Formulary.factory("ack").head.downloader.last_commit
 => "a0a0a0a0"
2.0.0-p481 :002 > ^D
```
That means that `a0a0a0a` is no longer unique within the repo, but it was when we installed `HEAD-a0a0a0a`, therefore HEAD is outdated.

The problem arises when we introduce `GitGitHubDownloadStrategy` to be able to detect outdated versions without fetching a repository. `GitGitHubDownloadStrategy#last_commit` returns full-length hash. The problem is that we can't check installed hash and fetched hash for equality anymore. What we can is to perform `start_with?` check, but it still can't guarantee the version is up-to-date.

Example:

```
$ ls $(brew --cellar)/ack
HEAD-b090608c
```

Let's now suppose we want to check if `HEAD-b09060c` is outdated
using GitHub API, which return full-length prefix and let's
suppose it returns the following hash:

```
b090608c49643099a47247d79ac9ca80d05c2ee0
```

How do we treat it? Is the HEAD version outdated or not? Suppose
it's not since prefixes are the same. After a while we check it
again and GitHub API returns the following hash:

```
b090608c00000000000000000000000000000000
```

So, we still think our installed HEAD is up-to-date, however
that's not true.

# Workaround

* Store long hashes in Tab. Cons: complicate.
* Use short commits and use `==` for `VCSDownloadStrategy` and `start_with?` for `GitHubDownloadStrategy`. Cons: there is a possibility we get outdated version (see the example).
* Store long hashes everywhere. Cons: long hashes in `outdated` outputs and long prefixes.

For now long hashes used everywhere in the code and even if we use short hashes for git we cannot escape from long hashes for other VCSs.

@xu-cheng

UniqMartin:
> - Store long hashes in Tab. Cons: complicate.

I think this would be the cleanest solution. I haven't spent much time thinking about how much this complicates the whole design, so take this with a grain of salt.

> - Use short commits and use `==` for VCSDownloadStrategy and `start_with?` for `GitHubDownloadStrategy`. Cons: there is a possibility we get outdated version (see the example).

I don't think this is an unreasonable strategy. While this won't be perfect, the likelihood of this failing would be close to 0% if the shortened hash isn't overly short, e.g. 16 characters. Even if this wasn't sufficient to uniquely identify all commits, it would only affect a few commits and in extremely big Git repositories with millions if not billions of commits and objects.

> - Store long hashes everywhere. Cons: long hashes in `outdated` outputs and long prefixes.

This could be partially addressed by using a more efficient encoding that uses fewer characters at the expense of no longer having hex digits and thus adding an additional layer of abstraction that might complicate matters. E.g. by re-encoding the full commit hash in Base32, its length could be shortened from 40 to 32 characters. (Just wanted to mention this theoretical possibility. I don't think this is a practical solution due to its complexity.)

xu-cheng:
@vladshablinsky can you give a list of exemple prefix for different VCS strategy? I think there is no much advantage to short git commit if we cannot short commit for other VCS installation.

Library/Homebrew/formula.rb
```diff
@@ -413,24 +413,42 @@ def linked_keg
...
+    tab = Tab.for_keg(prefix(version))
+
+    return true if stable && tab.stable_version && tab.stable_version < stable.version
+    return true if devel && tab.devel_version && tab.devel_version < devel.version
+
+    if options[:fetch_head]
+      return false unless head && head.downloader.is_a?(VCSDownloadStrategy)
+      downloader = head.downloader
+      downloader.shutup! unless ARGV.verbose?
+      version.version.commit != downloader.fetch_last_commit
```
xu-cheng: You probably need to cache the upstream last commit.

xu-cheng:
Also, @vladshablinsky shouldn't be this PR ready to be merged(after review)? If so, let's remove WIP tag in the title.

xu-cheng:
@Homebrew/maintainers This PR should be the final one for @vladshablinsky's GSoC project. The only remaining problem is how we should store VCS commit in the prefix, as mentioned in the above.

All of implementations for better supporting HEAD have been done. Please give your valuable comments. Thanks. 

Finally thanks for @vladshablinsky to finish his project way before schedule.

Library/Homebrew/cmd/outdated.rb
```diff
@@ -28,10 +28,18 @@ def outdated
 
   def print_outdated(formulae)
     verbose = ($stdout.tty? || ARGV.verbose?) && !ARGV.flag?("--quiet")
+    fetch_head = ARGV.fetch_head?
 
-    formulae.select(&:outdated?).each do |f|
+    formulae.select do |f|
+      f.outdated?(:fetch_head => fetch_head)
+    end.each do |f|
```
UniqMartin: I have to admit that I'm not a big fan of chaining multiple multi-line `do` blocks. Would the following be also acceptable to you?

```ruby
outdated_formulae = formulae.select do |f|
  f.outdated?(:fetch_head => fetch_head)
end

outdated_formulae.each do |f|
  …
```

Library/Homebrew/cmd/outdated.rb
```diff
@@ -28,10 +28,18 @@ def outdated
...
       if verbose
-        puts "#{f.full_name} (#{f.outdated_versions*", "} < #{f.pkg_version})"
+        outdated_versions = f.outdated_versions(:fetch_head => fetch_head)
+        if outdated_versions.any? && f.head? && !fetch_head
+          puts "#{f.full_name} (#{outdated_versions*", "}) HEAD update available"
+        else
+          puts "#{f.full_name} (#{outdated_versions*", "}) < #{f.pkg_version}"
```
UniqMartin: Now that you're modifying this anyway: Can you please replace `*", "` with `.join(", ")` here and two lines above? The former is a lot more obscure.

Library/Homebrew/formula.rb
```diff
@@ -980,7 +998,7 @@ def migration_needed?
   end
 
   # @private
-  def outdated_versions
+  def outdated_versions(options={})
```
UniqMartin: Please use spaces around the assignment (i.e. `options = {}`).

Library/Homebrew/formula.rb
```diff
@@ -989,16 +1007,22 @@ def outdated_versions
...
+      else
+        all_versions.sort!
+      end
     end
   end
 
   # @private
-  def outdated?
-    outdated_versions.any?
+  def outdated?(options={})
```
UniqMartin: Please use spaces around the assignment (i.e. `options = {}`).

Library/Homebrew/test/test_formula.rb
```diff
@@ -527,60 +568,67 @@ def setup
   end
 
   def teardown
-    @f.rack.rmtree
+    @f.rack.rmtree if @f.rack.exist?
   end
 
-  def setup_tab_for_prefix(prefix, tap_string=nil)
+  def setup_tab_for_prefix(prefix, options={})
```
UniqMartin: Please use spaces around the assignment (i.e. `options = {}`).

Library/Homebrew/cmd/outdated.rb
UniqMartin: Also this pattern of checking for outdated formulae appears in at least three places in this diff. I wonder if there's a good way to refactor this to avoid repeating this three-line `select` code snippet everywhere.

Library/Homebrew/tab.rb
```diff
@@ -31,7 +31,12 @@ def self.create(formula, compiler, stdlib)
       "source" => {
         "path" => formula.path.to_s,
         "tap" => formula.tap ? formula.tap.name : nil,
-        "spec" => formula.active_spec_sym.to_s
+        "spec" => formula.active_spec_sym.to_s,
+        "versions" => {
+          "stable" => formula.stable ? formula.stable.version.to_s : nil,
+          "devel" => formula.devel ? formula.devel.version.to_s : nil,
+          "head" => formula.head ? formula.head.version.to_s : nil
+        }
```
UniqMartin: I think in addition to this you will also need to populate `attributes["source"]["versions"]` in `Tab.from_file_content`. Otherwise `Tab#versions` will return `nil` for existing tabs that were created before this change and this will cause `Tab#stable_version` and `Tab#devel_version` to fail.

Library/Homebrew/tab.rb
```diff
@@ -231,6 +241,18 @@ def spec
...
+    source["versions"]
+  end
+
+  def stable_version
+    Version.create(versions["stable"]) if versions["stable"]
+  end
+
+  def devel_version
+    Version.create(versions["devel"]) if versions["devel"]
+  end
```
UniqMartin: Is there a reason there's both `Tab#stable_version` and `Tab#devel_version` but no`Tab#head_version`? This feels like a surprising omission to me.

Library/Homebrew/tab.rb
vladshablinsky: :+1: Already implemented it locally. Haven't pushed the changes yet, but I will.

Library/Homebrew/formula.rb
vladshablinsky: Do I cache it in `Formula` or in `VCSDownloadStrategy`?

vladshablinsky:
> shouldn't this PR be ready to be merged(after review)? If so, let's remove WIP tag in the title.

I had three questions left regarding the project:

* 1. Seems like [`curl_output`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils/curl.rb#L22-L34) never fails, so we probably need something like this:
```ruby
  def fetch_last_commit
    return unless head?
    commit = github_last_commit || super
    version.update_commit(commit)
    commit
  end
```

* 2. Do we need to check that last fetched commit not `nil`, maybe raise an Error or something?
* 3. How do I change `print_outdated_json` behaviour if `--fetch-HEAD` flag passed.

Also, what if we changhe `--fetch-HEAD` to `--fetch-head`?

UniqMartin:
> * 1. Seems like [`curl_output`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils/curl.rb#L22-L34) never fails, […]

It does, but you have to check the status yourself after invoking it, e.g.:

```ruby
output, errors, status = curl_output(…)
handle_failure unless status.success?
```

(See [`GitHub.open`](https://github.com/Homebrew/brew/blob/9f6cb8ed040d37d9706ee2b26e2c0a414f9a4660/Library/Homebrew/utils/github.rb#L149-L163) for a practical example where this is used to handle the error case.)

vladshablinsky:
Still can't understand why it [fails on Maverics](http://bot.brew.sh/job/Brew%20Pull%20Requests/1117/version=mavericks/testReport/junit/brew-test-bot/mavericks/tests/). I thought maybe the problem was in parentheses, so I installed ruby 1.8.7 and tries some silly stuff like this:

```ruby
dummy_f = "f"

def dummy_f.outdated_versions(options = {})
  [self, options.keys]
end

def dummy_refute_predicate o1, op, msg = nil
  raise "Raised with msg: #{msg}" if !o1.__send__(op)
end

dummy_refute_predicate dummy_f.outdated_versions(:fetch_head => true), :empty?
```

But it all works good, so, I don't know what is the reason of the fail.

Library/Homebrew/test/test_formula.rb
```diff
@@ -591,23 +639,97 @@ def test_mixed_taps_outdated_version_installed
...
+      shutup do
+        system "git", "init"
+        system "git", "add", "--all"
+        system "git", "commit", "-m", "Initial commit"
+      end
+    end
+
+    refute_predicate f.outdated_versions(:fetch_head => true), :empty?
+
+    tab_a["versions"] = { "stable" => f.version }
```
UniqMartin: The test failure is actually happening here. (You cannot fully trust the line numbers in the test bot output because your PR is being merged on top of an up-to-date `master`, so things might move a little bit.) The problem is that `OpenStruct` in Ruby 1.8.7 doesn't have a `[]=` method. If you change it to the following, it will work:

```ruby
tab_a.versions = { "stable" => f.version }
```

Library/Homebrew/formula.rb
```diff
@@ -980,7 +998,7 @@ def migration_needed?
   end
 
   # @private
-  def outdated_versions
+  def outdated_versions(options = {})
     @outdated_versions ||= begin
```
UniqMartin: I wonder if it is still appropriate to cache the return value in a single instance variable, as the result can change based on the given `options`. I would expect the following sequence to produce different results, but it probably won't because the first one will have already populated the cache, so the options on the second invocation don't have any effect:

```ruby
f.outdated_versions
f.outdated_versions(:fetch_head => true)
```

Library/Homebrew/test/test_formula.rb
vladshablinsky: Thanks!

Library/Homebrew/formula.rb
vladshablinsky: What should I do about it? We can cache outdated versions in two instance variables, but I think we shouldn't give up caching.

UniqMartin: I think a simple solution would be to cache based on the given options, though I assume this will continue to be slightly incorrect, as I think `:fetch_head => true` means “fetch HEAD now (possibly again) and then return a result based on this most recent fetch”.

A cleaner solution would be to think about what doesn't change during the lifetime of a Homebrew process and only cache that. Then refactor the method to compute the final result from the information in the cache and add to that the information that can potentially change on every invocation of this method.

vladshablinsky: I'm not sure but I think that we don't need to fetch remote repository twice when we already have outdated versions with `:fetch_head => true` cached somewhere.

As I can see we only use `Formula#outdated` and `Formula#outdated_versions` in `cmd/upgrade` and `cmd/outdated`. For `outdated` we don't need to reassign anything (at least now). For `upgrade` we  need `outdated` only to understand that it's time to upgrade.

Also, supposing we don't cache anything at all, is it possible that `outdated_versions` returns something different from what it returns on its first call? It seems that no, otherwise we would have the same caching problem we have now. And assuming that our further behaviour depends only on what is returned by `outdated_versions` on its first call, if `outdated_versions(:fetch_head => true)` is not empty on its first call it won't change from what it is on first call. If `outdated_versions(:fetch_head => true)` is empty on its first call, I think it's even worse if it gets not empty somewhere in the middle of the process. So, I think having `outdated_versions` with and without the flag isn't that bad.

Library/Homebrew/cmd/outdated.rb
```diff
@@ -28,10 +28,18 @@ def outdated
...
 
-    formulae.select(&:outdated?).each do |f|
+    outdated_formulae = formulae.select { |f| f.outdated?(:fetch_head => fetch_head) }
+
+    outdated_formulae.each do |f|
       if verbose
-        puts "#{f.full_name} (#{f.outdated_versions*", "} < #{f.pkg_version})"
+        outdated_versions = f.outdated_versions(:fetch_head => fetch_head)
+        if outdated_versions.any? && f.head? && !fetch_head
+          puts "#{f.full_name} (#{outdated_versions.join(", ")}) HEAD update available"
```
MikeMcQuaid: Let's go for `< latest HEAD`

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -18,10 +18,15 @@ def upgrade
...
+      outdated = Formula.installed.select do |f|
+        f.outdated?(:fetch_head => ARGV.fetch_head?)
+      end
+
       exit 0 if outdated.empty?
     elsif ARGV.named.any?
-      outdated = ARGV.resolved_formulae.select(&:outdated?)
+      outdated = ARGV.resolved_formulae.select do |f|
+        f.outdated?(:fetch_head => ARGV.fetch_head?)
+      end
```
MikeMcQuaid: Would be nice to avoid repeating this if possible.

Library/Homebrew/download_strategy.rb
```diff
@@ -756,6 +788,47 @@ def fix_absolute_submodule_gitdir_references!
...
+      @user = $1
+      @repo = $2
+    end
+  end
+
+  def github_last_commit
+    return if ENV["HOMEBREW_NO_GITHUB_API"]
+    args = []
+    args << "https://api.github.com/repos/#{@user}/#{@repo}/commits/#{@ref}"
+    args << "-I" << "-H" << "Accept: application/vnd.github.v3.sha"
```
MikeMcQuaid: Use long `curl` arguments here so it's more obvious what the do.

Library/Homebrew/tab.rb
```diff
@@ -67,6 +72,14 @@ def self.from_file_content(content, path)
       end
     end
 
+    if attributes["source"]["versions"].nil?
+      attributes["source"]["versions"] = {
+        "stable" => nil,
+        "devel" => nil,
+        "head" => nil
```
MikeMcQuaid: Tiny nit: add `,` after last elements in hashes.

MikeMcQuaid:
Great work here @vladshablinsky; made a few comments about nits but nothing major here. Once they are fixed I'm 👍 on @xu-cheng 🚢ing this when he's happy.

Library/Homebrew/formula.rb
```diff
@@ -980,25 +998,45 @@ def migration_needed?
...
+        return @outdated_versions
+      elsif @outdated_versions_head_fetched && options[:fetch_head]
+        return @outdated_versions_head_fetched
       end
 
+      if options[:fetch_head]
+        @outdated_versions_head_fetched ||= _outdated_versions(options)
+      else
+        @outdated_versions ||= _outdated_versions(options)
+      end
```
xu-cheng: Instead of introducing two variable and some complicated logics, you can actually do this:
```ruby
def outdated_versions(options = {})
  @outdated_versions ||= Hash.new do |options|
    @outdated_versions[options] = _outdated_versions(options)
  end
  @outdated_versions[options]
end
```

Library/Homebrew/formula.rb
```diff
@@ -980,25 +998,45 @@ def migration_needed?
...
+      all_versions << version
+
+      return [] if pkg_version <= version && !version.head?
+    end
+
+    head_version = latest_head_version
+    if head_version
+      head_version_outdated?(head_version, options) ? all_versions.sort! : []
+    else
       all_versions.sort!
```
xu-cheng: A nitpick: I would prefer to using `all_version.sort` in this line and the line above. After all, you are using the return value.

bfontaine: Am I missing something or could it be simplified?
```rb
# before
if head_version
  head_version_outdated?(head_version, options) ? all_versions.sort! : []
else
  all_versions.sort!
end

# after
if head_version && !head_version_outdated?(head_version, options)
  []
else
  all_versions.sort!
end
```

vladshablinsky: @bfontaine looks good, thanks. Probably it was similar to your suggestion at some point locally.

Library/Homebrew/cmd/upgrade.rb
vladshablinsky: Any ideas on how we can achieve it? I can think of adding some methods to `ARGV` for that purpose, but that's not the best solution I guess.

vladshablinsky:
@xu-cheng there is some code uncovered in `cmd/outdated`, so maybe some new integration tests should be added, shouldn't they? Other things seems to be covered where it's possible. `GitHubDownloadStrategy` is not covered and it's probably expected, but most of the other code is covered.

Note about CVS:

I spent some time trying to get last revision in the CVS repository, it seems there is no way to do that just running one command and not scanning through all the metadata CVS stores for each file. So, I think it'll be better to just have timestamps as a unique identifier for CSV.

Library/Homebrew/formula.rb
```diff
@@ -413,24 +413,46 @@ def linked_keg
...
+    head_version = latest_head_version
+    prefix(head_version) if head_version
+  end
+
+  def tab_version_outdated?(version)
+    tab = Tab.for_keg(prefix(version))
+
+    return true if stable && tab.stable_version && tab.stable_version < stable.version
+    return true if devel && tab.devel_version && tab.devel_version < devel.version
+  end
```
xu-cheng: I would avoid to introduce this method, as `tab_version_outdated?` should be implementation details that should not be exposed.

Can we merge it into `head_version_outdated?`? The only places I see you use `tab_version_outdated?` is in `outdated?` in which case, you can just use `head_version_outdated?(version, :fetch_head => false)`.

Library/Homebrew/formula.rb
```diff
@@ -413,24 +413,46 @@ def linked_keg
...
+
+  def head_version_outdated?(version, options = {})
+    return true if tab_version_outdated?(version)
+
+    if options[:fetch_head]
+      return false unless head && head.downloader.is_a?(VCSDownloadStrategy)
+      downloader = head.downloader
+      downloader.shutup! unless ARGV.verbose?
+      downloader.commit_outdated?(version.version.commit)
+    end
```
xu-cheng: a nitpick, can we have a `else false end` here. A method with `?` in name should only return true or false, but not nil.

Library/Homebrew/cmd/outdated.rb
```diff
@@ -40,11 +49,20 @@ def print_outdated(formulae)
 
   def print_outdated_json(formulae)
     json = []
-    outdated = formulae.select(&:outdated?).each do |f|
+    fetch_head = ARGV.fetch_head?
+    outdated_formulae = formulae.select { |f| f.outdated?(:fetch_head => fetch_head) }
+
+    outdated = outdated_formulae.each do |f|
+      outdated_versions = f.outdated_versions(:fetch_head => fetch_head)
+      current_version = if f.tab_version_outdated?(f.latest_head_version)
```
xu-cheng: It's bad practice to use `f.latest_head_version`, because it's only referred to one of version and has different semantical meaning with `outdated_versions`.

This can(should) be `if f.head? && outdated_versions.include?(f.pkg_version)`. The idea is that:
(1) You should check whether it's a head formula.
(2) If `f.pkg_version` is the same as one of installed versions, it means that you didn't fetch latest HEAD.

However, in order to make it work, you are required to implement `HeadVersion#==`.

Library/Homebrew/cmd/outdated.rb
```diff
@@ -28,10 +28,19 @@ def outdated
...
+    outdated_formulae.each do |f|
       if verbose
-        puts "#{f.full_name} (#{f.outdated_versions*", "} < #{f.pkg_version})"
+        outdated_versions = f.outdated_versions(:fetch_head => fetch_head)
+        if outdated_versions.any? && f.head? && !fetch_head \
+            || (version = f.latest_head_version) && f.tab_version_outdated?(version)
+          puts "#{f.full_name} (#{outdated_versions.join(", ")}) < latest HEAD"
+        else
+          puts "#{f.full_name} (#{outdated_versions.join(", ")}) < #{f.pkg_version}"
+        end
```
xu-cheng: This part is overcomplicated, you should use the same logic as below `print_outdated_json`, i.e.
```ruby
      current_version = if f.head? && outdated_versions.any? { |v| v.to_s ==  f.pkg_version.to_s }
        "latest HEAD"
      else
        f.pkg_version.to_s
      end
      puts "#{f.full_name} (#{outdated_versions.join(", ")}) < #{current_version}"

```

xu-cheng:
> there is some code uncovered in cmd/outdated, so maybe some new integration tests should be added, shouldn't they? Other things seems to be covered where it's possible. GitHubDownloadStrategy is not covered and it's probably expected, but most of the other code is covered.

It is preferable to have the test as well. But it can be done in follow up PR.

xu-cheng:
Thanks @vladshablinsky. Congratulation on finishing the GSoC project.

UniqMartin:
Great work! :tada:

MikeMcQuaid:
Great work on this project @vladshablinsky!

PR #587, opened 2016-07-25
tests: fix code style issues

eirinikos:

-----

This change fixes many of the existing bugs reported by `brew style` for `Homebrew/test`.

33 bugs remain; many of them are one of the following:
- `Method definitions must not be nested. Use lambda instead.`
  - these method definitions all (or mostly) seem to be enclosed within `formula do` blocks; does it make sense to convert these to lambdas? (I haven't yet figured out a way to do so that passes `brew tests`.)

- `Use snake_case for method names.`
  - examples of these methods include `test_check_DYLD_vars` and `test_create_DATA`; it seems like these acronyms are best kept as-is.

- `Prefer to_s over string interpolation.`
  - these bugs are easy to fix, but many instances of string interpolation seem to be tolerated by `brew style` (e.g., in `test_integration_cmds`; why is that?

I'm sensing that Rubocop's opinions are meant to be guidelines (rather than gospel), but maybe I'm wrong!

🎈

### DIFF
--- a/Library/Homebrew/test/.rubocop.yml
+++ b/Library/Homebrew/test/.rubocop.yml
@@ -1,0 +1,5 @@
+inherit_from: ../../.rubocop.yml
+
+# `formula do` uses nested method definitions
+Style/NestedMethodDefinition:
+  Enabled: false
--- a/Library/Homebrew/test/test_diagnostic.rb
+++ b/Library/Homebrew/test/test_diagnostic.rb
@@ -105,11 +105,11 @@ class DiagnosticChecksTest < Homebrew::TestCase
     sep = File::PATH_SEPARATOR
     # ensure /usr/bin is before HOMEBREW_PREFIX/bin in the PATH
     ENV["PATH"] = "/usr/bin#{sep}#{bin}#{sep}" +
-      ENV["PATH"].gsub(%r{(?:^|#{sep})(?:/usr/bin|#{bin})}, "")
+                  ENV["PATH"].gsub(%r{(?:^|#{sep})(?:/usr/bin|#{bin})}, "")
 
     # ensure there's at least one file with the same name in both /usr/bin/ and
     # HOMEBREW_PREFIX/bin/
-    (bin/"#{File.basename Dir["/usr/bin/*"].first}").mkpath
+    (bin/File.basename(Dir["/usr/bin/*"].first)).mkpath
 
     assert_match "/usr/bin occurs before #{HOMEBREW_PREFIX}/bin",
       @checks.check_user_path_1
@@ -129,7 +129,7 @@ class DiagnosticChecksTest < Homebrew::TestCase
   def test_check_user_path_sbin
     sbin = HOMEBREW_PREFIX/"sbin"
     ENV["PATH"] = "#{HOMEBREW_PREFIX}/bin#{File::PATH_SEPARATOR}" +
-      ENV["PATH"].gsub(%r{(?:^|#{File::PATH_SEPARATOR})#{sbin}}, "")
+                  ENV["PATH"].gsub(/(?:^|#{Regexp.escape(File::PATH_SEPARATOR)})#{Regexp.escape(sbin)}/, "")
     (sbin/"something").mkpath
 
     assert_nil @checks.check_user_path_1
@@ -162,7 +162,7 @@ class DiagnosticChecksTest < Homebrew::TestCase
     end
   end
 
-  def test_check_DYLD_vars
+  def test_check_dyld_vars
     ENV["DYLD_INSERT_LIBRARIES"] = "foo"
     assert_match "Setting DYLD_INSERT_LIBRARIES",
       @checks.check_DYLD_vars
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -194,8 +194,8 @@ class IntegrationCommandTests < Homebrew::TestCase
   end
 
   def test_env
-    assert_match %r{CMAKE_PREFIX_PATH="#{HOMEBREW_PREFIX}[:"]},
-                 cmd("--env")
+    assert_match(/CMAKE_PREFIX_PATH="#{Regexp.escape(HOMEBREW_PREFIX)}[:"]/,
+                 cmd("--env"))
   end
 
   def test_prefix_formula
--- a/Library/Homebrew/test/test_keg.rb
+++ b/Library/Homebrew/test/test_keg.rb
@@ -154,7 +154,7 @@ class LinkTests < Homebrew::TestCase
     refute_predicate HOMEBREW_PREFIX/"lib/foo", :directory?
   end
 
-  def test_unlink_ignores_DS_Store_when_pruning_empty_dirs
+  def test_unlink_ignores_ds_store_when_pruning_empty_dirs
     mkpath HOMEBREW_PREFIX/"lib/foo/bar"
     touch HOMEBREW_PREFIX/"lib/foo/.DS_Store"
     mkpath @keg/"lib/foo/bar"
--- a/Library/Homebrew/test/test_patch.rb
+++ b/Library/Homebrew/test/test_patch.rb
@@ -21,13 +21,13 @@ class PatchTests < Homebrew::TestCase
     assert_equal :p1, patch.strip
   end
 
-  def test_create_DATA
+  def test_create_data
     patch = Patch.create(:p0, :DATA)
     assert_kind_of DATAPatch, patch
     assert_equal :p0, patch.strip
   end
 
-  def test_create_DATA_without_strip
+  def test_create_data_without_strip
     patch = Patch.create(:DATA, nil)
     assert_kind_of DATAPatch, patch
     assert_equal :p1, patch.strip
@@ -112,7 +112,6 @@ end
 class ExternalPatchTests < Homebrew::TestCase
   def setup
     @p = ExternalPatch.new(:p1) { url "file:///my.patch" }
-
   end
 
   def test_url
--- a/Library/Homebrew/test/test_patching.rb
+++ b/Library/Homebrew/test/test_patching.rb
@@ -2,15 +2,15 @@ require "testing_env"
 require "formula"
 
 class PatchingTests < Homebrew::TestCase
-  TESTBALL_URL = "file://#{TEST_DIRECTORY}/tarballs/testball-0.1.tbz"
-  TESTBALL_PATCHES_URL = "file://#{TEST_DIRECTORY}/tarballs/testball-0.1-patches.tgz"
-  PATCH_URL_A = "file://#{TEST_DIRECTORY}/patches/noop-a.diff"
-  PATCH_URL_B = "file://#{TEST_DIRECTORY}/patches/noop-b.diff"
+  TESTBALL_URL = "file://#{TEST_DIRECTORY}/tarballs/testball-0.1.tbz".freeze
+  TESTBALL_PATCHES_URL = "file://#{TEST_DIRECTORY}/tarballs/testball-0.1-patches.tgz".freeze
+  PATCH_URL_A = "file://#{TEST_DIRECTORY}/patches/noop-a.diff".freeze
+  PATCH_URL_B = "file://#{TEST_DIRECTORY}/patches/noop-b.diff".freeze
   PATCH_A_CONTENTS = File.read "#{TEST_DIRECTORY}/patches/noop-a.diff"
   PATCH_B_CONTENTS = File.read "#{TEST_DIRECTORY}/patches/noop-b.diff"
-  APPLY_A = "noop-a.diff"
-  APPLY_B = "noop-b.diff"
-  APPLY_C = "noop-c.diff"
+  APPLY_A = "noop-a.diff".freeze
+  APPLY_B = "noop-b.diff".freeze
+  APPLY_C = "noop-c.diff".freeze
 
   def formula(*args, &block)
     super do
@@ -157,7 +157,7 @@ class PatchingTests < Homebrew::TestCase
         sha256 TESTBALL_PATCHES_SHA256
         apply APPLY_B
       end
-   }
+    }
   end
 
   def test_patch_p0
@@ -200,7 +200,7 @@ class PatchingTests < Homebrew::TestCase
     assert_patched formula { patch :p0, PATCH_B_CONTENTS }
   end
 
-  def test_patch_DATA_constant
+  def test_patch_data_constant
     assert_patched formula("test", Pathname.new(__FILE__).expand_path) {
       def patches
         :DATA
--- a/Library/Homebrew/test/test_pathname.rb
+++ b/Library/Homebrew/test/test_pathname.rb
@@ -46,7 +46,7 @@ class PathnameTests < Homebrew::TestCase
     refute_predicate @dir, :exist?
   end
 
-  def test_rmdir_if_possible_ignore_DS_Store
+  def test_rmdir_if_possible_ignore_ds_store
     mkdir_p @dir
     touch @dir/".DS_Store"
     assert @dir.rmdir_if_possible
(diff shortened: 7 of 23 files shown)


Library/Homebrew/test/test_tap.rb
```diff
@@ -154,7 +154,7 @@ def test_git_variant
 
     assert_equal "e1893a6bd191ba895c71b652ff8376a6114c7fa7", @tap.git_head
     assert_equal "e189", @tap.git_short_head
-    assert_match %r{years ago}, @tap.git_last_commit
+    assert_match /years ago/, @tap.git_last_commit
```
DomT4: Could this just be `"years ago"` & avoid `//` entirely given we're not actually using a regex here or am I missing something?

eirinikos: Makes sense to me!

UniqMartin:
> This change fixes many of the existing bugs reported by `brew style` for `Homebrew/test`.

Sorry if I'm being a bit pedantic here, but “bugs” feels like to strong of a word. There's nothing inherently wrong or broken about the code, it's just that it doesn't fully conform to certain code style guidelines.

Aside from this there's also a minor issue with the commit subject (and PR title). The way I tend to read them is `<area of changed code>: <what was changed>` and in this spirit a more fitting subject would be `tests: fix code style issues` or something similar.

> - `Method definitions must not be nested. Use lambda instead.`
>   - these method definitions all (or mostly) seem to be enclosed within `formula do` blocks; does it make sense to convert these to lambdas? (I haven't yet figured out a way to do so that passes `brew tests`.)

This is one of those cases where refactoring is unlikely to make sense and RuboCop suggests a solution that doesn't apply because it fails to understand the full context of the surrounding code. Depending on whether this affects individual lines or short blocks of code, the best solution is to suppress those complaints by temporarily disabling one or more cops, as [detailed in the RuboCop documentation](http://rubocop.readthedocs.io/en/latest/configuration/#disabling-cops-within-source-code).

> - `Use snake_case for method names.`
>   - examples of these methods include `test_check_DYLD_vars` and `test_create_DATA`; it seems like these acronyms are best kept as-is.

True. Silencing the relevant cop is the best way to address this and to drop the violation count.

> - `Prefer to_s over string interpolation.`
>   - these bugs are easy to fix, but many instances of string interpolation seem to be tolerated by `brew style` (e.g., in `test_integration_cmds`; why is that?

This mostly sounds like a misunderstanding to me, i.e. in which contexts this suggestion applies. `"#{something}"` is an unnecessary use of string interpolation because there's only the variable `something` and no surrounding text or other variables, thus `something` (if it is a string) or `something.to_s` (if it can be converted to a string) is more appropriate. "#{something}\n" instead is a perfectly fine use of string interpolation.

> I'm sensing that Rubocop's opinions are meant to be guidelines (rather than gospel), but maybe I'm wrong!

Indeed, they are guidelines and where possible, we have tweaked the settings because our preferred style doesn't fully align with what RuboCop considers to be the best default. It can sometimes make sense to tweak a rule instead of “fixing” our code, though most suggestions and defaults are good (and just show how old and diverse some of our code base is).

Library/Homebrew/test/test_diagnostic.rb
```diff
@@ -142,7 +142,7 @@ def test_check_user_path_bin
   def test_check_user_path_sbin
     sbin = HOMEBREW_PREFIX/"sbin"
     ENV["PATH"] = "#{HOMEBREW_PREFIX}/bin#{File::PATH_SEPARATOR}" +
-      ENV["PATH"].gsub(%r{(?:^|#{File::PATH_SEPARATOR})#{sbin}}, "")
+                  ENV["PATH"].gsub(/(?:^|File::PATH_SEPARATOR.to_s)sbin.to_s/, "")
```
UniqMartin: The substitution of `#{File::PATH_SEPARATOR}` with `File::PATH_SEPARATOR.to_s` is wrong and the same applies to the `#{sbin}` and `sbin.to_s` pair. A relatively easy way to check this is to enter both regular expressions into `brew irb` (though we'll need to also provide `sbin`):

```console
$ brew irb
irb(main):001:0> sbin = HOMEBREW_PREFIX/"sbin"
=> #<Pathname:/opt/homebrew/sbin>
irb(main):002:0> %r{(?:^|#{File::PATH_SEPARATOR})#{sbin}}
=> /(?:^|:)\/opt\/homebrew\/sbin/
irb(main):003:0> /(?:^|File::PATH_SEPARATOR.to_s)sbin.to_s/
=> /(?:^|File::PATH_SEPARATOR.to_s)sbin.to_s/
```

The correct substitution should be:

```ruby
/(?:^|#{Regexp.escape(File::PATH_SEPARATOR)})#{Regexp.escape(sbin)}/
```

(`Regexp.escape` makes additionally sure to properly handle any characters that we want to be interpreted literally, but that have special meaning when used inside a regular expression.)

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -193,7 +193,7 @@ def test_cellar_formula
   end
 
   def test_env
-    assert_match %r{CMAKE_PREFIX_PATH="#{HOMEBREW_PREFIX}[:"]},
+    assert_match /CMAKE_PREFIX_PATH="#{HOMEBREW_PREFIX}[:"]/,
```
UniqMartin: Nothing wrong here, but now that you're touching this it would good to escape the embedded string, i.e.:

```ruby
/CMAKE_PREFIX_PATH="#{Regexp.escape(HOMEBREW_PREFIX)}[:"]/
```

(This is generally a good idea to avoid misinterpretation of characters with special meaning in regular expressions. And we don't know in advance whether `HOMEBREW_PREFIX` might contain such problematic characters.)

eirinikos:
Sorry -- I accidentally closed this PR when I decided to rename my local and remote branches (without realizing the effect on the PR). 😳  Working on addressing the feedback...

eirinikos:
> Sorry if I'm being a bit pedantic here, but “bugs” feels like to strong of a word. There's nothing inherently wrong or broken about the code, it's just that it doesn't fully conform to certain code style guidelines.
> 
> Aside from this there's also a minor issue with the commit subject (and PR title). The way I tend to read them is `<area of changed code>: <what was changed>` and in this spirit a more fitting subject would be `tests: fix code style issues` or something similar.
> 

No worries, that makes sense to me. I thought I remembered hearing these issues referred to as "bugs" some time ago, but I see that that isn't the most accurate way to refer to them. I've updated the commit and PR title, though the branch still carries the old name. 😬 

Thanks for pointing me to the Rubocop documentation on temporarily disabling cops. I've done this for the **`Method definitions must not be nested`** and **`Use snake_case for method names`** warnings. I followed the example of some [existing calls to `rubocop:disable`](https://github.com/Homebrew/brew/blob/9e7b8c61a99c4d4c69da36f95759e375b7a4e116/Library/Homebrew/compat/hardware.rb#L3-L4) and included a code comment explaining the purpose of disabling Rubocop (one comment per type of cop, per file).

I also removed the string interpolation in [`test_formulary`](https://github.com/Homebrew/brew/pull/587/files#diff-37389a0e65da3f23aad3b3b96636d8e0L134) and [`test_tap`](https://github.com/Homebrew/brew/pull/587/files#diff-3676702e08834bbec8ee89f8f9869981L184) since it was unnecessary in both instances (I hadn't actually examined these offenses before). It may be obvious that I approached this task with haste and impatience, so I've learned my lesson. Thank you @UniqMartin and @DomT4 for taking a closer look here! 😁 

9 Rubocop offenses remain in `Homebrew/test`; I hope it's OK for these to stay on the back-burner for now:
```
== test_dependency_collector.rb ==
C: 10: 40: Avoid the use of the case equality operator ===.
== test_ENV.rb ==
W: 31:  5: Do not suppress exceptions.
W: 31:  5: Avoid rescuing the Exception class. Perhaps you meant to rescue StandardError?
== test_formula.rb ==
W:511: 23: Comparison of something with itself detected.
C:523: 29: Do not use semicolons to terminate expressions.
== test_patching.rb ==
C:126:  9: Avoid multi-line chains of blocks.
C:140:  9: Avoid multi-line chains of blocks.
C:239:  9: Avoid multi-line chains of blocks.
== testing_env.rb ==
C: 45:  7: Replace class var @@log with a class instance var.
```

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -193,7 +193,7 @@ def test_cellar_formula
   end
 
   def test_env
-    assert_match %r{CMAKE_PREFIX_PATH="#{HOMEBREW_PREFIX}[:"]},
+    assert_match /CMAKE_PREFIX_PATH="#{Regexp.escape(HOMEBREW_PREFIX)}[:"]/,
```
MikeMcQuaid: This needs to be wrapped with `()`: http://bot.brew.sh/job/Brew%20Pull%20Requests/1111/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/readall___syntax/

Library/Homebrew/test/test_diagnostic.rb
```diff
@@ -162,7 +162,8 @@ def test_check_for_config_scripts
     end
   end
 
-  def test_check_DYLD_vars
+  # let's disable rubocop (it flags this method as not being snake_case)
```
MikeMcQuaid: Let's just rename the method instead.

Library/Homebrew/test/test_keg.rb
```diff
@@ -154,7 +154,8 @@ def test_unlink_prunes_empty_toplevel_directories
     refute_predicate HOMEBREW_PREFIX/"lib/foo", :directory?
   end
 
-  def test_unlink_ignores_DS_Store_when_pruning_empty_dirs
+  # let's disable rubocop (it flags this method as not being snake_case)
+  def test_unlink_ignores_DS_Store_when_pruning_empty_dirs # rubocop:disable MethodName
```
MikeMcQuaid: Let's rename the method instead.

Library/Homebrew/test/test_patch.rb
```diff
@@ -21,13 +21,14 @@ def test_create_string_without_strip
     assert_equal :p1, patch.strip
   end
 
-  def test_create_DATA
+  # let's disable rubocop (it flags this method as not being snake_case)
+  def test_create_DATA # rubocop:disable MethodName
```
MikeMcQuaid: Let's rename the method instead

Library/Homebrew/test/test_patch.rb
```diff
@@ -21,13 +21,14 @@ def test_create_string_without_strip
...
     patch = Patch.create(:p0, :DATA)
     assert_kind_of DATAPatch, patch
     assert_equal :p0, patch.strip
   end
 
-  def test_create_DATA_without_strip
+  def test_create_DATA_without_strip # rubocop:disable MethodName
```
MikeMcQuaid: Let's rename the method instead.

Library/Homebrew/test/test_patching.rb
```diff
@@ -200,17 +201,18 @@ def test_patch_string_with_strip
     assert_patched formula { patch :p0, PATCH_B_CONTENTS }
   end
 
-  def test_patch_DATA_constant
+  # let's disable rubocop (it flags this method as not being snake_case)
+  def test_patch_DATA_constant # rubocop:disable MethodName
```
MikeMcQuaid: Let's rename the method instead.

Library/Homebrew/test/test_pathname.rb
```diff
@@ -46,7 +46,8 @@ def test_rmdir_if_possible
     refute_predicate @dir, :exist?
   end
 
-  def test_rmdir_if_possible_ignore_DS_Store
+  # let's disable rubocop (it flags this method as not being snake_case)
+  def test_rmdir_if_possible_ignore_DS_Store # rubocop:disable MethodName
```
MikeMcQuaid: Let's rename the method instead

Library/Homebrew/test/test_versions.rb
```diff
@@ -53,7 +53,8 @@ def test_comparing_regular_versions
     assert_operator version("1.2.3"), :<, version("1.2.3-p34")
   end
 
-  def test_HEAD
+  # let's disable rubocop (it flags this method as not being snake_case)
+  def test_HEAD # rubocop:disable MethodName
```
MikeMcQuaid: Let's rename the method instead.

Library/Homebrew/test/test_caveats.rb
```diff
@@ -18,7 +18,8 @@ def test_empty?
     f = formula do
       url "foo-1.0"
 
-      def caveats
+      # let's disable rubocop here since it dislikes nested method definitions
+      def caveats # rubocop:disable NestedMethodDefinition
```
MikeMcQuaid: Instead of all these, add a `.rubocop.yml` file in `Library/Homebrew/test` with:

```
inherit_from: ../../.rubocop.yml

# `formula do` needs uses method definitions
Style/NestedMethodDefinition:
  Enabled: false
```

MikeMcQuaid:
@eirinikos Nice work! Added a few comments and then we can 🚢 this.

Library/Homebrew/test/test_integration_cmds.rb
eirinikos: Oops. Got it. 👍

eirinikos: ...though now Rubocop is saying `Don't use parentheses around a literal`... 😑

MikeMcQuaid: Sorry, needs to be this line and the next

eirinikos: Do you mean like this?
```
assert_match(/CMAKE_PREFIX_PATH="#{Regexp.escape(HOMEBREW_PREFIX)}[:"]/,
             cmd("--env"))
```

MikeMcQuaid: Yep!

eirinikos:
@MikeMcQuaid Thanks!

Library/Homebrew/test/.rubocop.yml
```diff
@@ -0,0 +1,5 @@
+inherit_from: ../../.rubocop.yml
+
+# `formula do` uses nested method definitions
+Style/NestedMethodDefinition:
+  Enabled: false
```
MikeMcQuaid: Add a newline at the end of this file (ideally configure your editor to do this automagically!)

Library/Homebrew/test/test_diagnostic.rb
```diff
@@ -105,11 +105,11 @@ def test_check_user_path_usr_bin_before_homebrew
     sep = File::PATH_SEPARATOR
     # ensure /usr/bin is before HOMEBREW_PREFIX/bin in the PATH
     ENV["PATH"] = "/usr/bin#{sep}#{bin}#{sep}" +
-      ENV["PATH"].gsub(%r{(?:^|#{sep})(?:/usr/bin|#{bin})}, "")
+                  ENV["PATH"].gsub(%r{(?:^|#{sep})(?:/usr/bin|#{bin})}, "")
 
     # ensure there's at least one file with the same name in both /usr/bin/ and
     # HOMEBREW_PREFIX/bin/
-    (bin/"#{File.basename Dir["/usr/bin/*"].first}").mkpath
+    (bin/(File.basename Dir["/usr/bin/*"].first)).mkpath
```
UniqMartin: Minor nit, but it's nicer to put parentheses around the method argument(s) instead of wrapping the method invocation as a whole in parentheses, i.e.:

```ruby
(bin/File.basename(Dir["/usr/bin/*"].first)).mkpath
```

Library/Homebrew/test/.rubocop.yml
eirinikos: Got it! 👍

Library/Homebrew/test/test_diagnostic.rb
eirinikos: Makes sense. 👍

UniqMartin:
Merged in 8ec59253df47b9e597bc6a5ad488c42e035c0351. Thank you for another great contribution to Homebrew, @eirinikos! :tada:

MikeMcQuaid:
Nice work @eirinikos 👏

PR #588, opened 2016-07-26
Cask migration in migrate_tap_migration

AnastasiaSulyagina:

-----
Added logic for casks to be migrated on `brew update`

### DIFF
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -193,7 +193,7 @@ class Reporter
       dst = Pathname.new paths.last
 
       next unless dst.extname == ".rb"
-      next unless paths.any? { |p| tap.formula_file?(p) }
+      next unless paths.any? { |p| tap.formula_file?(p)}
 
       case status
       when "A", "D"
@@ -256,13 +256,20 @@ class Reporter
         next unless (HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
         new_tap = Tap.fetch(new_tap_name)
         new_tap.install unless new_tap.installed?
-        ohai "#{name} has been moved to Homebrew. Installing #{name}...", <<-EOS.undent
+        ohai "#{name} has been moved to Homebrew.", <<-EOS.undent
           To uninstall the cask run:
             brew cask uninstall --force #{name}
         EOS
-        system HOMEBREW_BREW_FILE, "install", "#{new_tap_name}/#{name}"
-        unless Formulary.factory("#{new_tap_name}/#{name}").keg_only?
-          system HOMEBREW_BREW_FILE, "link", "#{new_tap_name}/#{name}", "--force"
+        new_full_name = "#{new_tap_name}/#{name}"
+        next if (HOMEBREW_CELLAR/name.split("/").last).directory?
+        ohai "Installing #{name}..."
+        system HOMEBREW_BREW_FILE, "install", new_full_name
+        begin
+          unless Formulary.factory(new_full_name).keg_only?
+            system HOMEBREW_BREW_FILE, "link", new_full_name, "--force"
+          end
+        rescue Exception => e
+          onoe e if ARGV.homebrew_developer?
         end
         next
       end
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -249,6 +249,20 @@ class Reporter
   def migrate_tap_migration
     report[:D].each do |full_name|
       name = full_name.split("/").last
+      if tap == "caskroom/cask"
+        next unless (dir = HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
+        next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
+        new_tap = Tap.fetch(new_tap_name)
+        new_tap.install unless new_tap.installed?
+        ohai "#{name} has been moved to Homebrew Core. Installing #{name}..."
+        system HOMEBREW_BREW_FILE, "install", name
+        ohai <<-EOS.undent
+            To uninstall the cask run:
+              brew cask uninstall --force #{name}
+        EOS
+        next
+      end
+
       next unless (dir = HOMEBREW_CELLAR/name).exist? # skip if formula is not installed.
       next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
       tabs = dir.subdirs.map { |d| Tab.for_keg(Keg.new(d)) }
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -249,22 +249,25 @@ class Reporter
   def migrate_tap_migration
     report[:D].each do |full_name|
       name = full_name.split("/").last
+      new_tap_name = tap.tap_migrations[name]
+      next if new_tap_name.nil? # skip if not in tap_migrations list.
+
       if tap == "caskroom/cask"
-        next unless (dir = HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
-        next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
+        next unless (HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
         new_tap = Tap.fetch(new_tap_name)
         new_tap.install unless new_tap.installed?
-        ohai "#{name} has been moved to Homebrew Core. Installing #{name}..."
-        system HOMEBREW_BREW_FILE, "install", name
-        ohai <<-EOS.undent
-            To uninstall the cask run:
-              brew cask uninstall --force #{name}
+        ohai "#{name} has been moved to Homebrew. Installing #{name}...", <<-EOS.undent
+          To uninstall the cask run:
+            brew cask uninstall --force #{name}
         EOS
+        system HOMEBREW_BREW_FILE, "install", "#{new_tap_name}/#{name}"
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,6 +249,20 @@ def updated?
...
+        system HOMEBREW_BREW_FILE, "install", name
+        ohai <<-EOS.undent
+            To uninstall the cask run:
+              brew cask uninstall --force #{name}
+        EOS
+        next
+      end
+
       next unless (dir = HOMEBREW_CELLAR/name).exist? # skip if formula is not installed.
       next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
```
MikeMcQuaid: Could maybe move this line above so it's shared between both?

Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,6 +249,20 @@ def updated?
   def migrate_tap_migration
     report[:D].each do |full_name|
       name = full_name.split("/").last
+      if tap == "caskroom/cask"
+        next unless (dir = HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
+        next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
+        new_tap = Tap.fetch(new_tap_name)
+        new_tap.install unless new_tap.installed?
+        ohai "#{name} has been moved to Homebrew Core. Installing #{name}..."
```
MikeMcQuaid: `#{name} has been moved to Homebrew. Installing #{name}..."`

MikeMcQuaid: You may need to run `brew link --force` to overwrite any symlinks that Homebrew Cask has created.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,6 +249,20 @@ def updated?
   def migrate_tap_migration
     report[:D].each do |full_name|
       name = full_name.split("/").last
+      if tap == "caskroom/cask"
+        next unless (dir = HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
```
UniqMartin: The assignment to `dir` seems unnecessary. This variable is not used in the rest of this code block, unless I'm overlooking something.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,6 +249,20 @@ def updated?
...
+        next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
+        new_tap = Tap.fetch(new_tap_name)
+        new_tap.install unless new_tap.installed?
+        ohai "#{name} has been moved to Homebrew Core. Installing #{name}..."
+        system HOMEBREW_BREW_FILE, "install", name
+        ohai <<-EOS.undent
+            To uninstall the cask run:
+              brew cask uninstall --force #{name}
+        EOS
```
UniqMartin: - The indentation inside the `<<-EOS.undent`–`EOS` block is to deep. It needs to be only be indented by two additional spaces (currently four).

- As mentioned in a previous PR, the first argument of `ohai` is more of a title and subject to truncation and thus isn't really suitable for multi-line output like this. We need either to switch to `puts` or provide a relatively short header like in the migration below from Homebrew to Cask.

  Or maybe with some rearrangement this can be combined with the `ohai` above the `system HOMEBREW_BREW_FILE, "install", name` line.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,8 +249,21 @@ def updated?
   def migrate_tap_migration
     report[:D].each do |full_name|
       name = full_name.split("/").last
-      next unless (dir = HOMEBREW_CELLAR/name).exist? # skip if formula is not installed.
       next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
```
UniqMartin: I know you've just moved around existing code, but I think it would be easier to follow (and also avoid an overly long line) if this was split into two lines. Assignments in conditions are sometimes surprising and a careless reader could think that's actually a check for equality. Here's what I'm suggesting:

```ruby
new_tap_name = tap.tap_migrations[name]
next if new_tap_name.nil? # skip if not in tap_migrations list.
```

Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,8 +249,21 @@ def updated?
...
+      if tap == "caskroom/cask"
+        next unless (HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
+        new_tap = Tap.fetch(new_tap_name)
+        new_tap.install unless new_tap.installed?
+        system HOMEBREW_BREW_FILE, "install", name
+        system HOMEBREW_BREW_FILE, "link", name, "--force"
+        ohai "#{name} has been moved to Homebrew. Installing #{name}...", <<-EOS.undent
+          To uninstall the cask run:
+            brew cask uninstall --force #{name}
+        EOS
```
UniqMartin: Thanks for addressing the problem with `ohai`'s arguments, but now this is somewhat awkwardly output *after* the formula has already been installed and linked. Shouldn't this be placed before the preceding two lines?

Library/Homebrew/cmd/update-report.rb
AnastasiaSulyagina: Agree. For me it was quite surprising when I saw this first time.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,8 +249,21 @@ def updated?
...
     report[:D].each do |full_name|
       name = full_name.split("/").last
-      next unless (dir = HOMEBREW_CELLAR/name).exist? # skip if formula is not installed.
       next unless new_tap_name = tap.tap_migrations[name] # skip if formula is not in tap_migrations list.
+      if tap == "caskroom/cask"
+        next unless (HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
+        new_tap = Tap.fetch(new_tap_name)
+        new_tap.install unless new_tap.installed?
+        system HOMEBREW_BREW_FILE, "install", name
+        system HOMEBREW_BREW_FILE, "link", name, "--force"
```
UniqMartin: I'm not sure if this is a problem of practical relevance, but it feels like it could be safer to use `full_name` instead of just `name` in case a Cask is ever migrated to a non-core formula and there is some ambiguity. But I have to admit that I haven't fully though this through, thus feel free to educate me if that's not going to happen or not an issue …

AnastasiaSulyagina: For now casks are only migrated to taps within Homebrew organisation and I don't think that there are some duplicated or ambiguous formulas here. But it will definitely be no harm if I change `name` to `full_name`, I guess.

xu-cheng: Can you check whether the formula is keg only before force link?

AnastasiaSulyagina: Could you explain a bit about "Keg only"?

UniqMartin: > Edit: but it should be not full_name but new_tap_name/name

Yes, you're right. Thanks for noticing this. :+1:

xu-cheng: @AnastasiaSulyagina http://www.rubydoc.info/github/Homebrew/brew/master/Formula#keg_only-class_method

AnastasiaSulyagina:
I thought it should work well with just `new_tap_name/name` instead of `"#{new_tap_name}/#{name}"` but it did not. If it should though could somebody please explain what i'm doing wrong?

UniqMartin:
> I thought it should work well with just `new_tap_name/name` instead of `"#{new_tap_name}/#{name}"` but it did not. If it should though could somebody please explain what i'm doing wrong?

You're not doing anything wrong. `"#{new_tap_name}/#{name}"` is the correct way to do this.

I think you got confused by the `path/"suffix"` pattern, but this only works because `path` is often a `Pathname` instance and thus `Pathname#/` is used to construct a new `Pathname` instance with the given suffix. In this case we're dealing with simple strings and `/` simply doesn't make sense.

AnastasiaSulyagina:
Thanks, yeh, I didn't understand how it works and thought if strings can be merged in one place, why not in another :D

Library/Homebrew/cmd/update-report.rb
```diff
@@ -249,8 +249,25 @@ def updated?
...
+      if tap == "caskroom/cask"
+        next unless (HOMEBREW_REPOSITORY/"Caskroom"/name).exist?
+        new_tap = Tap.fetch(new_tap_name)
+        new_tap.install unless new_tap.installed?
+        ohai "#{name} has been moved to Homebrew. Installing #{name}...", <<-EOS.undent
+          To uninstall the cask run:
+            brew cask uninstall --force #{name}
+        EOS
+        system HOMEBREW_BREW_FILE, "install", "#{new_tap_name}/#{name}"
+        unless Formulary.factory("#{new_tap_name}/#{name}").keg_only?
```
xu-cheng: Loading formulae need to be guarded by `begin ... rescue Execption ... end`. You can see the example codes in this very file at below. 

Also I think it's worth to check whether this formula is installed before to install it.

MikeMcQuaid:
@AnastasiaSulyagina Nice work! Made some tweaks in 2bd5e0e and ebe8b8d you may be interested in and changed the commit message in be5484d. Read http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html for your future commit messages and try to squash your commits into one. Thanks!

PR #592, opened 2016-07-27
allow symlinks into /System in relocatable bottles

tdsmith:
This allows Python virtualenvs to be relocatable.

### DIFF
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -35,7 +35,7 @@ EOS
 MAXIMUM_STRING_MATCHES = 100
 
 module Homebrew
-  def keg_contains(string, keg, ignores)
+  def keg_contain?(string, keg, ignores)
     @put_string_exists_header, @put_filenames = nil
 
     def print_filename(string, filename)
@@ -95,10 +95,10 @@ module Homebrew
       end
     end
 
-    keg_contains_absolute_symlink_starting_with?(string, keg) || result
+    keg_contain_absolute_symlink_starting_with?(string, keg) || result
   end
 
-  def keg_contains_absolute_symlink_starting_with?(string, keg)
+  def keg_contain_absolute_symlink_starting_with?(string, keg)
     absolute_symlinks_start_with_string = []
     keg.find do |pn|
       if pn.symlink? && (link = pn.readlink).absolute?
@@ -237,10 +237,10 @@ module Homebrew
         if ARGV.include? "--skip-relocation"
           skip_relocation = true
         else
-          relocatable = false if keg_contains(prefix_check, keg, ignores)
-          relocatable = false if keg_contains(cellar, keg, ignores)
+          relocatable = false if keg_contain?(prefix_check, keg, ignores)
+          relocatable = false if keg_contain?(cellar, keg, ignores)
           if prefix != prefix_check
-            relocatable = false if keg_contains_absolute_symlink_starting_with?(prefix, keg)
+            relocatable = false if keg_contain_absolute_symlink_starting_with?(prefix, keg)
           end
           skip_relocation = relocatable && !keg.require_install_name_tool?
         end
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -100,16 +100,11 @@ module Homebrew
 
   def keg_contains_absolute_symlink_starting_with?(string, keg)
     absolute_symlinks_start_with_string = []
-    absolute_symlinks_rest = []
     keg.find do |pn|
       if pn.symlink? && (link = pn.readlink).absolute?
         if link.to_s.start_with?(string)
           absolute_symlinks_start_with_string << pn
-        else
-          absolute_symlinks_rest << pn
         end
-
-        result = true
       end
     end
 
@@ -120,16 +115,9 @@ module Homebrew
           puts "  #{pn} -> #{pn.resolved_path}"
         end
       end
-
-      unless absolute_symlinks_rest.empty?
-        opoo "Absolute symlink:"
-        absolute_symlinks_rest.each do |pn|
-          puts "  #{pn} -> #{pn.resolved_path}"
-        end
-      end
     end
 
-    result
+    !absolute_symlinks_start_with_string.empty?
   end
 
   def bottle_output(bottle)
@@ -245,12 +233,15 @@ module Homebrew
           ignores << %r{#{Regexp.escape(HOMEBREW_CELLAR)}/go/[\d\.]+/libexec}
         end
 
+        relocatable = true
         if ARGV.include? "--skip-relocation"
-          relocatable = true
           skip_relocation = true
         else
-          relocatable = !keg_contains(prefix_check, keg, ignores)
-          relocatable = !keg_contains(cellar, keg, ignores) && relocatable
+          relocatable = false if keg_contains(prefix_check, keg, ignores)
+          relocatable = false if keg_contains(cellar, keg, ignores)
+          if prefix != prefix_check
+            relocatable = false if keg_contains_absolute_symlink_starting_with?(prefix, keg)
+          end
           skip_relocation = relocatable && !keg.require_install_name_tool?
         end
         puts if !relocatable && ARGV.verbose?
--- a/Library/Homebrew/cmd/bottle.rb
+++ b/Library/Homebrew/cmd/bottle.rb
@@ -95,6 +95,10 @@ module Homebrew
       end
     end
 
+    keg_contains_absolute_symlink_starting_with?(string, keg) || result
+  end
+
+  def keg_contains_absolute_symlink_starting_with?(string, keg)
     absolute_symlinks_start_with_string = []
     absolute_symlinks_rest = []
     keg.find do |pn|
@@ -110,14 +114,14 @@ module Homebrew
     end
 
     if ARGV.verbose?
-      if absolute_symlinks_start_with_string.any?
+      unless absolute_symlinks_start_with_string.empty?
         opoo "Absolute symlink starting with #{string}:"
         absolute_symlinks_start_with_string.each do |pn|
           puts "  #{pn} -> #{pn.resolved_path}"
         end
       end
 
-      if absolute_symlinks_rest.any?
+      unless absolute_symlinks_rest.empty?
         opoo "Absolute symlink:"
         absolute_symlinks_rest.each do |pn|
           puts "  #{pn} -> #{pn.resolved_path}"


tdsmith:
Without this change:
```
==> Detecting if jrnl-1.9.7.el_capitan.bottle.2.tar.gz is relocatable...
Warning: Absolute symlink:
  /usr/local/Cellar/jrnl/1.9.7/libexec/.Python -> /System/Library/Frameworks/Python.framework/Versions/2.7/Python
  /usr/local/Cellar/jrnl/1.9.7/libexec/include/python2.7 -> /System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7
  /usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/UserDict.py -> /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/UserDict.py
  /usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/_abcoll.py -> /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/_abcoll.py
  /usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/_weakrefset.py -> /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/_weakrefset.py
  /usr/local/Cellar/jrnl/1.9.7/libexec/lib/python2.7/abc.py -> /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/abc.py
```

I think this should always be safe?

tdsmith:
let's block PRs like Homebrew/homebrew-core#3351 and Homebrew/homebrew-core#3352 on this for a second

Library/Homebrew/cmd/bottle.rb
```diff
@@ -102,6 +102,8 @@ def print_filename(string, filename)
         if link.to_s.start_with?(string)
           absolute_symlinks_start_with_string << pn
         else
+          # allow absolute symlinks to system frameworks
+          next if link.to_s.start_with?("/System/")
```
UniqMartin: It feels like this would look more natural if placed before `if link.to_s.start_with?(string)` or is the concern that this could somehow clash with `string` and lead to incorrect results?

UniqMartin:
Minor comment, but this looks a lot like the obvious and right thing to do.

Library/Homebrew/cmd/bottle.rb
tdsmith: Oh, no, I think you're right here

tdsmith:
@ilovezfs and I were chatting and agree that everything SIPpy is probably also safe though we'd want to scope the exemption to only OS X in that case. I kinda think the minimal change is fine for now unless we discover another use case.

UniqMartin:
Great to hear you've also thought about this! I briefly wondered about other system locations, but keeping this minimal until a broader need arises makes a lot of sense.

:+1:

MikeMcQuaid:
We could go further here I think and exclude anything outside `HOMEBREW_PREFIX` and `HOMEBREW_REPOSITORY`.

MikeMcQuaid:
As a bottle that symlinks to e.g. `/usr/X11` or something may or may not be usable on another system but it's still definitely relocatable. This check was added to avoid absolute symlinks to Homebrew paths and not system ones.

tdsmith:
Added some commits that:
- allow bottles to be relocated if they have arbitrary absolute symlinks
- add a new check to disallow absolute symlinks within HOMEBREW_PREFIX (in addition to HOMEBREW_PREFIX/opt and HOMEBREW_CELLAR, which the existing checks cover)

Library/Homebrew/cmd/bottle.rb
```diff
@@ -247,6 +239,9 @@ def bottle_formula(f)
         else
           relocatable = !keg_contains(prefix_check, keg, ignores)
           relocatable = !keg_contains(cellar, keg, ignores) && relocatable
+          unless prefix == prefix_check
```
MikeMcQuaid: `if prefix != prefix_check`

Library/Homebrew/cmd/bottle.rb
```diff
@@ -247,6 +239,9 @@ def bottle_formula(f)
...
+            relocatable = !keg_contains_absolute_symlink_starting_with?(prefix, keg) && relocatable
```
MikeMcQuaid: Can you split this into a few `= true if`s instead? I know that's the case above but it's confusing there too.

MikeMcQuaid:
Looking good. A few final nits from me then 🚢

Library/Homebrew/cmd/bottle.rb
```diff
@@ -233,14 +233,14 @@ def bottle_formula(f)
...
         if ARGV.include? "--skip-relocation"
-          relocatable = true
           skip_relocation = true
         else
-          relocatable = !keg_contains(prefix_check, keg, ignores)
-          relocatable = !keg_contains(cellar, keg, ignores) && relocatable
-          unless prefix == prefix_check
-            relocatable = !keg_contains_absolute_symlink_starting_with?(prefix, keg) && relocatable
+          relocatable = false if keg_contains(prefix_check, keg, ignores)
+          relocatable = false if keg_contains(cellar, keg, ignores)
```
MikeMcQuaid: Might be worth flipping this to be `not_relocatable` or similar and using `||=` so you can avoid running `keg_contains` multiple times when not necessary.

Library/Homebrew/cmd/bottle.rb
```diff
@@ -95,10 +95,10 @@ def print_filename(string, filename)
       end
     end
 
-    keg_contains_absolute_symlink_starting_with?(string, keg) || result
+    keg_contains_absolute_symlink_starting_with(string, keg) || result
   end
 
-  def keg_contains_absolute_symlink_starting_with?(string, keg)
+  def keg_contains_absolute_symlink_starting_with(string, keg)
```
MikeMcQuaid: Why drop the `?`? Does it not return a boolean?

Library/Homebrew/cmd/bottle.rb
```diff
@@ -95,10 +95,10 @@ def print_filename(string, filename)
       end
     end
 
-    keg_contains_absolute_symlink_starting_with?(string, keg) || result
+    keg_contains_absolute_symlink_starting_with(string, keg) || result
```
MikeMcQuaid: `result || keg...` feels slightly more idiomatic and avoids calling the function if unnecessary.

MikeMcQuaid:
Sorry @tdsmith, feel free to push back on any of these changes that aren't really your code.

Library/Homebrew/cmd/bottle.rb
tdsmith: I decided that the current implementation was intentionally avoiding short circuiting so that all issues are always flagged in verbose mode so I followed suit; I think it makes sense to maintain that behavior.

Library/Homebrew/cmd/bottle.rb
tdsmith: It does, but I thought it was better fung shui to match style with the existing `keg_contains`. What do you prefer?

Library/Homebrew/cmd/bottle.rb
MikeMcQuaid: Good point but may be worth only doing that in verbose mode to speed up non-verbose bottling. Feel free to leave as-is, though.

Library/Homebrew/cmd/bottle.rb
MikeMcQuaid: Let's make them both have a trailing `?`.

MikeMcQuaid: (assuming they both only return boolean values)

Library/Homebrew/cmd/bottle.rb
UniqMartin: I'm in favor of retaining the current behavior (always complete check). Would be good to know how expensive this really is before we start optimizing it away …

Library/Homebrew/cmd/bottle.rb
UniqMartin: I think we want to retain this order, too. We want the side effect of output in verbose mode, if there are any offending symlinks.

tdsmith:
Thanks for everyone's comments!

DomT4:
I'm not sure which set of changes is potentially problematic here, both Tim & Mike pushed bottle logic tweaks today, but we're seeing some CI rebellion on bottling.

Yosemite, [alot](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5467/version=yosemite/console):
```
==> brew bottle --verbose --json alot
==> FAILED
==> Determining alot bottle revision...
==> Bottling alot-0.3.7.yosemite.bottle.tar.gz...
Error: No such dylib name: /usr/local/opt/gpgme/lib/libgpgme.11.dylib
```
Same thing with [El Cap here](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5459/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/bottle_alot/). Repeated on Yosemite [here](http://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/5468/version=yosemite/console). PR causing the failures is [here](https://github.com/Homebrew/homebrew-core/pull/1663).

Will poke at it later today if nobody beats me to it, but need to grab some sleep at this point.

MikeMcQuaid:
@DomT4 Couldn't reproduce locally 😥

tdsmith:
Same; I don't think this is failing in a part that I touched. :/ If anything I think @woodruffw's pickier mach-o code might be having a bad interaction with something?

FWIW on `alot` I see:

```
tim@rocketman:~$ brew bottle -vd --json alot
/usr/local/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /usr/local/Library/Taps/homebrew/homebrew-core/Formula/alot.rb
==> Determining alot bottle revision...
==> Bottling alot-0.3.7.el_capitan.bottle.tar.gz...
Changing install name in /usr/local/Cellar/alot/0.3.7/libexec/lib/python2.7/site-packages/gpgme/_gpgme.so
  from /usr/local/opt/gpgme/lib/libgpgme.11.dylib
    to @@HOMEBREW_PREFIX@@/opt/gpgme/lib/libgpgme.11.dylib
...
```

UniqMartin:
@MikeMcQuaid @tdsmith Did any of you do `export HOMEBREW_RUBY_MACHO=1` before trying locally? If it's indeed an unfortunate interaction with `ruby-macho` (as it seems to be given the nature of the failure and the fact that the Mavericks bot is unaffected), that would be necessary to be able to reproduce. I'm looking into it, but the dependencies haven't finished building just yet …

UniqMartin:
> I'm not sure which set of changes is potentially problematic here, both Tim & Mike pushed bottle logic tweaks today, but we're seeing some CI rebellion on bottling.

Before anyone wastes any time hunting down the cause: I was able to reproduce and I'll open an issue with all relevant details shortly. Just let me say that the problem is entirely unrelated to the changes here. The problem is very specific to `alot` and how one of its resources is built.

UniqMartin:
If any of the recent changes can be identified as the trigger for this problem: It would have to be the introduction of the Python `virtualenv` stuff. I suppose the build environment in such a `virtualenv` is sufficiently different from what it was before, that native extensions (or however this is called in Python land) end up being built differently …

DomT4:
To be honest, I forgot `HOMEBREW_RUBY_MACHO` was running on CI, and since it _had_ passed CI before now it seemed likely to be either:
* The move to `virtualenv` in the formula.
* Mike's bottle changes.
* Tim's bottle changes.

Since the failure output was on the bottle, and Tim & Mike had both contributed to this thread & would see the message I thought I'd drop it in here 😅.

woodruffw:
Makes sense for me to drop in as well, since it's almost certainly exposing a flaw in ruby-macho :smile:

UniqMartin:
The problem has been documented in Homebrew/ruby-macho#50. Addressing it will probably require some minor architectural changes, thus a fix won't be available instantly. Any input on how `ruby-macho` should be reacting to similar conditions in the future is very welcome.

PR #600, opened 2016-07-29
Refurbish args in virtualenv

tdsmith:
Instead of adding additional lines to formula.rb:
  https://git.io/vKxxh
just turn on argument refurbishment for any formula that creates a
virtualenv.

cf Homebrew/ruby-macho#50, Homebrew/homebrew-core#1663


tdsmith:
cc @UniqMartin

UniqMartin:
:+1: Thanks for the quick fix!

I tested it with the previously problematic [`alot` formula](https://github.com/Homebrew/homebrew-core/pull/1663) and now everything works like a charm and all binaries are `x86_64`-only (except for the `python` shim in `libexec/bin/`).

MikeMcQuaid:
👍

PR #605, opened 2016-07-30
formula_versions: ignore FormulaMethodDeprecatedError

ilovezfs:
old formula revisions have deprecated elements such as sha1 checksums


ilovezfs:
CC @MikeMcQuaid

xu-cheng:
Actually, this may indicate a more severe problem that we may get revision number wrong.

MikeMcQuaid:
Thanks!

ilovezfs:
No problem :)

Issue #603, opened 2016-07-30
Problems with virtualenv system & dependency formulae.


DomT4:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [ ] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

It seems like things created with the new Python `virtualenv` system can't find dependencies that reside outside the `virtualenv`? That sounds incredibly obvious saying it, but I'm not sure if it was an intentional design decision & how to work around it.

For example, something that needs to `import gobject` typically has a `depends_on "pygobject"` dependency, but if you add that dependency & build something with:

```ruby
    venv = virtualenv_create(libexec)
    %w[
      setuptools attrs cffi chardet cryptography enum34 idna ipaddress
      pyasn1 pyasn1-modules pycparser Pillow pyOpenSSL pyxdg six
      service_identity Twisted zope.interface MarkupSafe Mako
    ].each do |r|
      venv.pip_install resource(r)
    end
    venv.link_scripts(bin) { venv.pip_install buildpath }
```
and then run the software:

```
deluge

[ERROR   ] 02:37:24 ui:171 No module named gobject
Traceback (most recent call last):
  File "/usr/local/Cellar/deluge/1.3.13/libexec/lib/python2.7/site-packages/deluge/ui/ui.py", line 152, in __init__
    from deluge.ui.gtkui.gtkui import GtkUI
  File "/usr/local/Cellar/deluge/1.3.13/libexec/lib/python2.7/site-packages/deluge/ui/gtkui/__init__.py", line 1, in <module>
    from gtkui import start
  File "/usr/local/Cellar/deluge/1.3.13/libexec/lib/python2.7/site-packages/deluge/ui/gtkui/gtkui.py", line 37, in <module>
    import gobject
ImportError: No module named gobject
[ERROR   ] 02:37:24 ui:172 There was an error whilst launching the request UI: gtk
[ERROR   ] 02:37:24 ui:173 Look at the traceback above for more information.
```

This will presumably be problematic for anything that relies on the Python elements of _another formula_ rather than from a resource? Was this intentional? Am I doing something stupid?

DomT4:
[...]

PR #608, opened 2016-07-30
Allow virtualenvs to find Python bindings provided by deps

tdsmith:
cf #603

### DIFF
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -122,6 +122,17 @@ module Language
         ENV.refurbish_args
         venv = Virtualenv.new formula, venv_root, python
         venv.create
+
+        # Find any Python bindings provided by dependencies
+        deps = formula.deps.select do |d|
+          !d.build? && (d.required? || formula.build.with?(d))
+        end
+        xy = Language::Python.major_minor_version python
+        pth_contents = deps.map do |d|
+          "import site; site.addsitedir('#{Formula[d.name].opt_lib}/python#{xy}/site-packages')\n"
+        end
+        (venv_root/"lib/python#{xy}/site-packages/homebrew_deps.pth").write pth_contents.join
+
         venv
       end
 
@@ -223,8 +234,8 @@ module Language
         def do_install(targets)
           targets = [targets] unless targets.is_a? Array
           @formula.system @venv_root/"bin/pip", "install",
-                 "-v", "--no-deps", "--no-binary", ":all:",
-                 *targets
+                          "-v", "--no-deps", "--no-binary", ":all:",
+                          "--ignore-installed", *targets
         end
       end # class Virtualenv
     end # module Virtualenv
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -123,15 +123,18 @@ module Language
         venv = Virtualenv.new formula, venv_root, python
         venv.create
 
-        # Find any Python bindings provided by dependencies
-        deps = formula.deps.select do |d|
-          !d.build? && (d.required? || formula.build.with?(d))
-        end
+        # Find any Python bindings provided by recursive dependencies
+        formula_deps = formula.recursive_dependencies
         xy = Language::Python.major_minor_version python
-        pth_contents = deps.map do |d|
-          "import site; site.addsitedir('#{Formula[d.name].opt_lib}/python#{xy}/site-packages')\n"
+        pth_contents = formula_deps.map do |d|
+          next if d.build?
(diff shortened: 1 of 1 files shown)


Library/Homebrew/language/python.rb
```diff
@@ -122,6 +122,17 @@ def virtualenv_create(venv_root, python = "python", formula = self)
         ENV.refurbish_args
         venv = Virtualenv.new formula, venv_root, python
         venv.create
+
+        # Find any Python bindings provided by dependencies
+        deps = formula.deps.select do |d|
+          !d.build? && (d.required? || formula.build.with?(d))
```
ilovezfs: does this ignore optional & required?

tdsmith: `formula.build.with?(d)` handles optional and recommended.

Library/Homebrew/language/python.rb
```diff
@@ -122,6 +122,17 @@ def virtualenv_create(venv_root, python = "python", formula = self)
...
+        end
+        xy = Language::Python.major_minor_version python
+        pth_contents = deps.map do |d|
+          "import site; site.addsitedir('#{Formula[d.name].opt_lib}/python#{xy}/site-packages')\n"
+        end
+        (venv_root/"lib/python#{xy}/site-packages/homebrew_deps.pth").write pth_contents.join
```
ilovezfs: can this be named anything?

Library/Homebrew/language/python.rb
```diff
@@ -122,6 +122,17 @@ def virtualenv_create(venv_root, python = "python", formula = self)
         ENV.refurbish_args
         venv = Virtualenv.new formula, venv_root, python
         venv.create
+
+        # Find any Python bindings provided by dependencies
+        deps = formula.deps.select do |d|
```
ilovezfs: should this be `recursive_dependencies`?

Library/Homebrew/language/python.rb
tdsmith: Yes.

Library/Homebrew/language/python.rb
tdsmith: It could be; I think either behavior is probably okay.

MikeMcQuaid:
If it fixes it: seems fine to me 👍

DomT4:
Are you going to throw me in the sea if I tell you this doesn't _quite_ work on my example in the Issue because of the way a formula is packaged? 🙈.

```python
[ERROR   ] 02:19:21 ui:171 No module named gtk
Traceback (most recent call last):
  File "/usr/local/Cellar/deluge2/1.3.13/libexec/lib/python2.7/site-packages/deluge/ui/ui.py", line 152, in __init__
    from deluge.ui.gtkui.gtkui import GtkUI
  File "/usr/local/Cellar/deluge2/1.3.13/libexec/lib/python2.7/site-packages/deluge/ui/gtkui/__init__.py", line 1, in <module>
    from gtkui import start
  File "/usr/local/Cellar/deluge2/1.3.13/libexec/lib/python2.7/site-packages/deluge/ui/gtkui/gtkui.py", line 50, in <module>
    reactor = gtk2reactor.install()
  File "/usr/local/Cellar/deluge2/1.3.13/libexec/lib/python2.7/site-packages/twisted/internet/gtk2reactor.py", line 99, in install
    reactor = Gtk2Reactor(useGtk)
  File "/usr/local/Cellar/deluge2/1.3.13/libexec/lib/python2.7/site-packages/twisted/internet/gtk2reactor.py", line 71, in __init__
    import gtk as _gtk
ImportError: No module named gtk
```
The problem here being that our `pygtk` formula's `gtk` module is packaged in:

```
/usr/local/opt/pygtk/lib/python2.7/site-packages/gtk-2.0
```
Rather than:
```
/usr/local/opt/pygtk/lib/python2.7/site-packages
```

Not sure if that's an upstream issue or something problematic to this PR, but thought I'd mention it. Everything else about this is 💯 on solving the other issues encountered.

tdsmith:
> Are you going to throw me in the sea

the North Atlantic, post-haste!

The version that operates on recursive dependencies *should* work here since the relevant file is provided by `pygobject` and pygtk depends_on pygobject; is that not happening?

tdsmith:
Oh: no, it's not happening, because the .pth file pygobject installs is assuming that it's read from the same tree as pygtk, when they're actually in separate cellars. pygobject's .pth file should be rewritten.

DomT4:
> the North Atlantic, post-haste!

Any chance we can go for a warmer sea?

> is that not happening?

I ran `brew pull` on this PR & then `brew install ~/Downloads/deluge2.rb`, and then executing a simple `deluge`, so I guess not?

DomT4:
Thanks Tim ❤️

PR #609, opened 2016-07-31
cleaner: kill unnecessary perl elements

DomT4:

-----

We shouldn't be packaging either `perllocal.pod` or `.packlist` files. Both are only really useful outside of package management. They get automatically generated whenever you install a Perl module.

Debian, Arch, MacPorts & others remove them and we should have been as well really; keeping them causes completely unnecessary conflicts between formulae.

### DIFF
--- a/Library/Homebrew/cleaner.rb
+++ b/Library/Homebrew/cleaner.rb
@@ -1,6 +1,8 @@
 # Cleans a newly installed keg.
 # By default:
 # * removes .la files
+# * removes perllocal.pod files
+# * removes .packlist files
 # * removes empty directories
 # * sets permissions on executables
 # * removes unresolved symlinks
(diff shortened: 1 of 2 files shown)


DomT4:
Going to ninja merge this because I've an incoming PR in `core` that needs it, but raise any objections as desired & I'll fix 'em later today.

Library/Homebrew/cleaner.rb
```diff
@@ -89,6 +91,14 @@ def clean_dir(d)
         next
       elsif path.extname == ".la"
         path.unlink
+      elsif path.basename.to_s == "perllocal.pod"
+        # Both this file & the .packlist one below are completely unnecessary
+        # to package & causes pointless conflict with other formulae. They are
+        # removed by Debian, Arch & MacPorts amongst other packagers as well.
+        # The files are created as part of installing any Perl module.
+        path.unlink
+      elsif path.basename.to_s == ".packlist" # Hidden file, not file extension!
```
UniqMartin: `%w[perllocal.pod .packlist].include?(path.basename.to_s)` could have been used to keep this a single case, as those items seem to be closely related, but the split approach is equally fine and is probably a bit easier to understand code-wise. (Just some random thoughts.)

DomT4: Yeah, I thought about that but I wanted to include the comment about it being a hidden file to clarify I wasn't being an idiot about not using the `extname` further up, and I wasn't sure if that would be confusing clumped in with the same comments about `perllocal.pod`.

PR #344, opened 2016-06-12
Add helper class for Python virtualenvs

tdsmith:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

This PR proposes a new class for interacting with Python virtualenvs which enables a new way to write formulas for Python applications. 

This approach resolves a few problems with the existing approach:
* UX: Installing applications with this method yields [clearer and much less code](https://gist.github.com/tdsmith/cf2ef71adae2718075fef17bd9f6d938) than [the current method](https://github.com/Homebrew/homebrew-core/blob/master/Formula/jrnl.rb#L52-L65). If we trust pip to talk to pypi, we can pass a list of requirements instead of using resources, which means [even less code](https://gist.github.com/f9e83e9ce66fa19349ac711b1c170462).
* Functional: In order that installing an application which uses Python packages does not disturb the global Python environment, we presently install packages and their dependencies into a Cellar-only prefix and write wrapper scripts that set PYTHONPATH to help Python find the installed tools. This generally works, with caveats. One is that namespace packages don't work very well, requiring formula authors to add `touch foo/"__init__.py"` manually. Another is PYTHONPATH propagates to child processes, which makes it difficult for a script installed with python2 to launch another script written in python3, since the Python 2 packages in PYTHONPATH may be incompatible. This has been a problem for e.g. Ansible. Virtualenvs do not have either problem.

This is functionally similar to [pipsi](https://github.com/mitsuhiko/pipsi). I expect this approach to work in general but I'll plan to ask some people in the Python packaging community for review.

Some highlights for review:
* Where is the right place to store the virtualenv package's URL and hash? Should we just package a tarball in the Homebrew repository?
[...]

PR #613, opened 2016-07-31
python: tweak script linking in virtualenv

UniqMartin:
- [x] Have you written new tests for your changes?

----

**Remaining steps/tasks (see below for context):**

- [ ] Agree on the preferred interface for the non-block form.
- [x] Replace minimal implementation with proper one without block/recursion.
- [x] Update documentation in code (YARD).
- [ ] Update documentation in `share/doc/homebrew/`.
- [x] Update tests.

-----

This is an attempt to refine #344, to address https://github.com/Homebrew/brew/pull/344#issuecomment-232577429 by @MikeMcQuaid:

> > `venv.link_scripts(bin) { venv.pip_install "." }`
>
> Don't understand why `venv.link_scripts` is a block rather than a following command.

@ilovezfs expressed similar concerns in https://github.com/Homebrew/brew/pull/344#issuecomment-234242676:

> Also, it does still seem awkward to have `venv.link_scripts` taking the block that does ` venv.pip_install buildpath`. Could the virtualenv accumulate before and after state without needing to resort to nesting?

----

Instead of making the formula author use a slightly awkward block construct and expose this implementation detail (diffing of `@venv_root/"bin"`), offer a more familiar:

```ruby
venv.pip_install buildpath, :link_scripts => bin
```

The default will be not to link any scripts, but this can be of course reversed if that's a useful default behavior (in this case, the opt-out would be `:link_scripts => false`).

*Note:* The current state of the PR is primarily to solicit feedback whether this is a reasonable change to make to the user-facing API of the `virtualenv` support, assuming my proposal doesn't come a bit late (as far as I'm aware this part of the interface hasn't been used yet in formulae).

cc @tdsmith

### DIFF
--- a/Library/Homebrew/test/test_language_python.rb
+++ b/Library/Homebrew/test/test_language_python.rb
@@ -85,7 +85,7 @@ class LanguagePythonTests < Homebrew::TestCase
     bin_before = Dir[bin/"*"]
     FileUtils.touch bin/"kilroy"
     bin_after = Dir[bin/"*"]
-    @venv.expects(:pip_install).with { |*params| params == ["foo"] }
+    @venv.expects(:pip_install).with("foo")
     Dir.expects(:[]).twice.returns(bin_before, bin_after)
 
     @venv.pip_install_and_link "foo"
--- a/Library/Homebrew/test/test_language_python.rb
+++ b/Library/Homebrew/test/test_language_python.rb
@@ -71,18 +71,24 @@ class LanguagePythonTests < Homebrew::TestCase
     @venv.pip_install res
   end
 
-  def test_link_scripts_links_scripts
-    bin = (@dir/"bin")
-    dest = (@dir/"dest")
-    bin.mkpath
-    refute((bin/"kilroy").exist?)
-    refute((dest/"kilroy").exist?)
+  def test_pip_install_links_scripts
+    bin = (@dir/"bin").tap(&:mkpath)
+    dest = @dir/"dest"
+
+    refute_predicate bin/"kilroy", :exist?
+    refute_predicate dest/"kilroy", :exist?
+
     FileUtils.touch bin/"irrelevant"
-    @venv.link_scripts(dest) { FileUtils.touch bin/"kilroy" }
-    assert((bin/"kilroy").exist?)
-    assert((dest/"kilroy").exist?)
-    assert((dest/"kilroy").symlink?)
-    assert_equal((bin/"kilroy").realpath, (dest/"kilroy").realpath)
-    refute((dest/"irrelevant").exist?)
+    @formula.expects(:system).returns(true).with do |*params|
+      FileUtils.touch bin/"kilroy"
+      params.first == @dir/"bin/pip" && params.last == "foo"
+    end
+    @venv.pip_install "foo", :link_scripts => dest
+
+    assert_predicate bin/"kilroy", :exist?
+    assert_predicate dest/"kilroy", :exist?
+    assert_predicate dest/"kilroy", :symlink?
+    assert_equal (bin/"kilroy").realpath, (dest/"kilroy").realpath
+    refute_predicate dest/"irrelevant", :exist?
   end
 end
--- a/Library/Homebrew/test/test_language_python.rb
+++ b/Library/Homebrew/test/test_language_python.rb
@@ -88,7 +88,7 @@ class LanguagePythonTests < Homebrew::TestCase
     assert_predicate bin/"kilroy", :exist?
     assert_predicate dest/"kilroy", :exist?
     assert_predicate dest/"kilroy", :symlink?
-    assert_equal (bin/"kilroy").realpath, (dest/"kilroy").realpath
+    assert_equal((bin/"kilroy").realpath, (dest/"kilroy").realpath)
     refute_predicate dest/"irrelevant", :exist?
   end
 end
--- a/Library/Homebrew/test/test_language_python.rb
+++ b/Library/Homebrew/test/test_language_python.rb
@@ -74,8 +74,9 @@ class LanguagePythonTests < Homebrew::TestCase
   end
 
   def test_pip_install_and_link_links_scripts
-    bin = (@dir/"bin").tap(&:mkpath)
-    dest = @formula.dest
+    bin = @dir/"bin"
+    bin.mkpath
+    dest = @formula.bin
 
     refute_predicate bin/"kilroy", :exist?
     refute_predicate dest/"kilroy", :exist?
--- a/Library/Homebrew/test/test_language_python.rb
+++ b/Library/Homebrew/test/test_language_python.rb
@@ -6,8 +6,10 @@ class LanguagePythonTests < Homebrew::TestCase
   def setup
     @dir = Pathname.new(mktmpdir)
     resource = stub("resource", :stage => true)
+    formula_bin = @dir/"formula_bin"
     @formula = mock("formula") do
       stubs(:resource).returns(resource)
+      stubs(:bin).returns(formula_bin)
     end
     @venv = Language::Python::Virtualenv::Virtualenv.new(@formula, @dir, "python")
   end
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Makes sense to me; can you supply a sample formula diff?

ilovezfs:
any thoughts on having install & install_and_link, and exposing even less of this?

tdsmith:
This was the first way I implemented it. I came to feel that using an options hash was not as nice as composing methods, based on a prejudice against arguments that switch behavior on or off; you can find some comments in that PR where I vacillate between the two options.

> install & install_and_link

This was the second way I implemented it! I think this is probably nicer than an options hash.

UniqMartin:
Pushed a few updates to make the code changes more complete.

> Makes sense to me; can you supply a sample formula diff?

Most formulae will use the single-shot wrapper `virtualenv_install_with_resources` that already exists. The more complicated cases will use `virtualenv_create` and operate on the returned `Virtualenv` instance. For these, the only difference will be to replace:

```
venv.link_scripts(bin) { venv.pip_install buildpath }
```

With:

```ruby
venv.pip_install buildpath, :link_scripts => bin
```

> any thoughts on having install & install_and_link, and exposing even less of this?

I'm fine with that and it would be a trivial change. Basically, I think either of the following read nicer than the block form (though I'm personally most fond of the option hash form):

```ruby
venv.pip_install_and_link buildpath
venv.pip_install buildpath, :link_scripts => true
venv.pip_install buildpath, :link_scripts => bin
```

`Virtualenv#pip_install_and_link` would be basically the current `Virtualenv#link_scripts` but instead of `yield`ing, it would forward its arguments to `Virtualenv#pip_install`.

UniqMartin:
Replaced the functionality of `:link_scripts => bin` with a `#pip_install_and_link` method that uses composition instead and delegates to `#pip_install` for installing the targets.

Library/Homebrew/test/test_language_python.rb
```diff
@@ -71,18 +73,24 @@ def test_pip_install_accepts_resource
...
   end
 
-  def test_link_scripts_links_scripts
-    bin = (@dir/"bin")
-    dest = (@dir/"dest")
-    bin.mkpath
-    refute((bin/"kilroy").exist?)
-    refute((dest/"kilroy").exist?)
+  def test_pip_install_and_link_links_scripts
+    bin = (@dir/"bin").tap(&:mkpath)
```
tdsmith: What does this do?

Library/Homebrew/test/test_language_python.rb
```diff
@@ -71,18 +73,24 @@ def test_pip_install_accepts_resource
...
+    refute_predicate bin/"kilroy", :exist?
+    refute_predicate dest/"kilroy", :exist?
+
     FileUtils.touch bin/"irrelevant"
-    @venv.link_scripts(dest) { FileUtils.touch bin/"kilroy" }
-    assert((bin/"kilroy").exist?)
-    assert((dest/"kilroy").exist?)
-    assert((dest/"kilroy").symlink?)
+    @formula.expects(:system).returns(true).with do |*params|
+      FileUtils.touch bin/"kilroy"
```
tdsmith: The `with` block instructs the mock object about how we expect it should be called; the block should return true if the call is what we expected and shouldn't have any side effects.

Library/Homebrew/test/test_language_python.rb
```diff
@@ -71,18 +73,24 @@ def test_pip_install_accepts_resource
...
+      params.first == @dir/"bin/pip" && params.last == "foo"
+    end
+    @venv.pip_install_and_link "foo"
+
+    assert_predicate bin/"kilroy", :exist?
+    assert_predicate dest/"kilroy", :exist?
+    assert_predicate dest/"kilroy", :symlink?
     assert_equal((bin/"kilroy").realpath, (dest/"kilroy").realpath)
-    refute((dest/"irrelevant").exist?)
+    refute_predicate dest/"irrelevant", :exist?
```
tdsmith: Not related, but why do we prefer this style?

Library/Homebrew/test/test_language_python.rb
UniqMartin: Single-line version of:

```ruby
```

tdsmith: Can we spell that out? I think the reference to `tap` is confusing.

Library/Homebrew/test/test_language_python.rb
UniqMartin: Because it results in a clearer message if the assertion fails. Instead of just saying “assertion failed” (no message is provided) it dumps the `Pathname` instance and the predicate of that instance that was supposed to be true/false.

Library/Homebrew/test/test_language_python.rb
UniqMartin: I was trying to avoid repeating `bin` again and used something every Ruby object has. I used (or maybe abused) `Object#tap`.

UniqMartin: > Can we spell that out?

Yes, we can!

Library/Homebrew/test/test_language_python.rb
```diff
@@ -71,18 +73,24 @@ def test_pip_install_accepts_resource
...
 
-  def test_link_scripts_links_scripts
-    bin = (@dir/"bin")
-    dest = (@dir/"dest")
-    bin.mkpath
-    refute((bin/"kilroy").exist?)
-    refute((dest/"kilroy").exist?)
+  def test_pip_install_and_link_links_scripts
+    bin = (@dir/"bin").tap(&:mkpath)
+    dest = @formula.dest
```
UniqMartin: Indeed. That's a stupid last-minute change I made because I was shuffling variable names back and forth and sadly I failed to run the test suite afterwards.

Library/Homebrew/test/test_language_python.rb
UniqMartin: Can you suggest the correct way how I can make the underlying `pip_install` invocation produce a side effect? Aside from that, I guess in this test I should be doing something like `@venv.expects(:pip_install).yields(nil)` instead …

Library/Homebrew/test/test_language_python.rb
tdsmith: I learned a thing today!

UniqMartin:
I have addressed all comments. I'm not sure how reasonable the stuff is, that I'm doing in the test, but at least I'm not sticking things with side effects into places where they don't belong.

Library/Homebrew/test/test_language_python.rb
```diff
@@ -71,18 +73,27 @@ def test_pip_install_accepts_resource
...
+
     FileUtils.touch bin/"irrelevant"
-    @venv.link_scripts(dest) { FileUtils.touch bin/"kilroy" }
-    assert((bin/"kilroy").exist?)
-    assert((dest/"kilroy").exist?)
-    assert((dest/"kilroy").symlink?)
+    bin_before = Dir[bin/"*"]
+    FileUtils.touch bin/"kilroy"
+    bin_after = Dir[bin/"*"]
+    @venv.expects(:pip_install).with { |*params| params == ["foo"] }
```
tdsmith: You can just say `.with("foo")`: http://gofreerange.com/mocha/docs/Mocha/Expectation.html#with-instance_method

tdsmith:
:+1:

Library/Homebrew/test/test_language_python.rb
UniqMartin: Thanks! I didn't realize this could be simplified like that for simple cases like this one.

UniqMartin:
Apologies that I was hardly around the past few days!

I have to admit that I'm not entirely happy that this was merged without completing the checklist in the top comment (e.g. update documentation in `share/doc/homebrew/`) and with all the noise of the incremental changes preserved in the commit message …

Issue #614, opened 2016-07-31
Cleanup: Calling Formula.md5 is disabled!


rgormley:
- [ X] Ran `brew update` and retried your prior step?
- [ X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [NA ] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

Running brew cleanup reports the following:

```
Error: Calling Formula.md5 is disabled!
Use Formula.sha256 instead.
/Users/rsg/Library/Caches/Homebrew/Formula/sshpass.rb:6:in `<class:Sshpass>'
Please report this bug:
    https://git.io/brew-troubleshooting
/usr/local/Library/Homebrew/utils.rb:153:in `odeprecated'
/usr/local/Library/Homebrew/utils.rb:161:in `odisabled'
/usr/local/Library/Homebrew/compat/md5.rb:3:in `md5'
/Users/rsg/Library/Caches/Homebrew/Formula/sshpass.rb:6:in `<class:Sshpass>'
/Users/rsg/Library/Caches/Homebrew/Formula/sshpass.rb:3:in `load_formula'
/usr/local/Library/Homebrew/formulary.rb:25:in `module_eval'
/usr/local/Library/Homebrew/formulary.rb:25:in `load_formula'
/usr/local/Library/Homebrew/formulary.rb:42:in `load_formula_from_path'
/usr/local/Library/Homebrew/formulary.rb:91:in `load_file'
/usr/local/Library/Homebrew/formulary.rb:82:in `klass'
/usr/local/Library/Homebrew/formulary.rb:78:in `get_formula'
/usr/local/Library/Homebrew/formulary.rb:215:in `factory'
/usr/local/Library/Homebrew/formulary.rb:227:in `from_rack'
/usr/local/Library/Homebrew/cleanup.rb:80:in `block in cleanup_cache'
/usr/local/Library/Homebrew/cleanup.rb:48:in `each'
/usr/local/Library/Homebrew/cleanup.rb:48:in `cleanup_cache'
/usr/local/Library/Homebrew/cleanup.rb:11:in `cleanup'
/usr/local/Library/Homebrew/cmd/cleanup.rb:20:in `cleanup'
/usr/local/Library/Homebrew/brew.rb:87:in `<main>'
```

MikeMcQuaid:
[...]

PR #615, opened 2016-07-31
brew-tap: allow boneyards to include invalid formulae

apjanke:

-----

This PR allows `*boneyard*` taps to be tapped even if their formulae have errors in them.

Now that `brew tap` does a `readall` on taps and balks at adding ones with errors, I can't tap `homebrew/boneyard` any more. The old formulae with `sha1` checksums are considered invalid.

```
$ brew tap homebrew/boneyard
==> Tapping homebrew/boneyard
Cloning into '/usr/local/Library/Taps/homebrew/homebrew-boneyard'...
remote: Counting objects: 1346, done.
remote: Total 1346 (delta 0), reused 0 (delta 0), pack-reused 1346
Receiving objects: 100% (1346/1346), 373.64 KiB | 0 bytes/s, done.
Resolving deltas: 100% (655/655), done.
Checking connectivity... done.
Error: Invalid formula: /usr/local/Library/Taps/homebrew/homebrew-boneyard/attica.rb
Calling Formula.sha1 is deprecated!
Use Formula.sha256 instead.
/usr/local/Library/Taps/homebrew/homebrew-boneyard/attica.rb:6:in `<class:Attica>'
Please report this to the homebrew/boneyard tap!
Error: Invalid formula: /usr/local/Library/Taps/homebrew/homebrew-boneyard/cdf.rb
Calling Formula.sha1 is deprecated!
Use Formula.sha256 instead.
...
Error: Invalid formula: /usr/local/Library/Taps/homebrew/homebrew-boneyard/sparseassembler.rb
Calling Formula.sha1 is deprecated!
Use Formula.sha256 instead.
/usr/local/Library/Taps/homebrew/homebrew-boneyard/sparseassembler.rb:7:in `<class:Sparseassembler>'
Please report this to the homebrew/boneyard tap!
Error: Cannot tap homebrew/boneyard: invalid syntax in tap!
[✘ /usr/local/Library/Taps/homebrew on ⇄ master]
```

The boneyard is where formulae with problems go to die, and we do not maintain boneyarded formulae to our current standards, so it seems like this should be allowed.

Unless someone wants to go through and update the `sha1`s in the boneyard. There's a script to do that, right? I think we should allow it in the mean time. I need to tap the boneyard so I can boneyard a formula.

### DIFF
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -214,7 +214,7 @@ class Tap
 
     begin
       safe_system "git", *args
-      unless Readall.valid_tap?(self, :aliases => true)
+      unless repo.include?("boneyard") || Readall.valid_tap?(self, :aliases => true)
         raise "Cannot tap #{name}: invalid syntax in tap!"
       end
     rescue Interrupt, ErrorDuringExecution, RuntimeError


apjanke:
Hmm. `brew uses` spews warnings when boneyard is tapped. May have to address that, too.

```
$ brew uses icmake
Error: Failed to import: /usr/local/Library/Taps/homebrew/homebrew-boneyard/attica.rb
Calling Formula.sha1 is deprecated!
Use Formula.sha256 instead.
/usr/local/Library/Taps/homebrew/homebrew-boneyard/attica.rb:6:in `<class:Attica>'
Please report this to the homebrew/boneyard tap!
Error: Failed to import: /usr/local/Library/Taps/homebrew/homebrew-boneyard/cdf.rb
Calling Formula.sha1 is deprecated!
Use Formula.sha256 instead.
/usr/local/Library/Taps/homebrew/homebrew-boneyard/cdf.rb:6:in `<class:Cdf>'
Please report this to the homebrew/boneyard tap!
Error: Failed to import: /usr/local/Library/Taps/homebrew/homebrew-boneyard/cgap.rb
Calling Formula.sha1 is deprecated!
Use Formula.sha256 instead.
/usr/local/Library/Taps/homebrew/homebrew-boneyard/cgap.rb:5:in `<class:Cgap>'
Please report this to the homebrew/boneyard tap!
Error: Failed to import: /usr/local/Library/Taps/homebrew/homebrew-boneyard/coremod.rb
Calling Formula.sha1 is deprecated!
Use Formula.sha256 instead.
/usr/local/Library/Taps/homebrew/homebrew-boneyard/coremod.rb:6:in `<class:Coremod>'
```

UniqMartin:
> This PR allows `*boneyard*` taps to be tapped even if their formulae have errors in them.

As you noticed yourself, this might be too narrow. It's annoying to have to manually clone an affected tap repository because `brew tap` refuses to do so. This prevents contributors from installing a tap with the intention to contribute a fix for the open issues. I think a better solution for this would be to offer a `--force` or `--skip-checks` flag for `brew tap`.

> Unless someone wants to go through and update the `sha1`s in the boneyard. There's a script to do that, right? I think we should allow it in the mean time. I need to tap the boneyard so I can boneyard a formula.

That's not really an option as some referenced downloads may no longer be available to recompute the checksum (and sometimes that's precisely the reason why a formula was boneyarded in the first place).

> Hmm. `brew uses` spews warnings when boneyard is tapped. May have to address that, too.

I suspect some other parts of Homebrew will also need some tweaks to their behavior. Unfortunately, I don't have a good idea for how to address this. We can teach `odeprecated` to be a no-op under certain circumstances, but I suspect there will be some resistance to this suggestion and the question would remain under what circumstances this should be the case.

ilovezfs:
There are 26 formulae with sha1 sums in boneyard and 29 total sha1 sums. Perhaps, we should see how many are really unfixable even if it violates our no-maintaining-boneyard-formulae labor laws.

MikeMcQuaid:
It's worth noting that this is a product of having `HOMEBREW_DEVELOPER` set which causes exceptions to be raised: https://github.com/Homebrew/brew/blob/a5a930c48bec7bb7bdc6bbf7e0fe7fe5f564a7fe/Library/Homebrew/utils.rb#L152-L153

Perhaps an alternative would be to tweak the logic as follows:
- Make warnings but don't raise exceptions for `HOMEBREW_DEVELOPER`s
- Make warnings but still don't raise exceptions for `HOMEBREW_DEVELOPER`s if `odisabled`/`options[:die]` is used (so they can fix them)
- Raise exceptions for non-`HOMEBREW_DEVELOPER`s if `odisabled`/`options[:die]` is used

This should allow all these commands to be used as expected. Thoughts?

> There are 26 formulae with sha1 sums in boneyard and 29 total sha1 sums. Perhaps we should see how many are really unfixable even if it violates our no-maintaining-boneyard-formulae labor laws.

I agree. I think it's worth making an exception for the boneyard for the `odisabled` (or soon to be `odisabled`) case.

It's also perhaps worth rethinking the boneyard and/or migration approach altogether; it seems weird that if you have one formula migrated there then `brew install foo` will happily install a formula from the boneyard without any warning or notification. Some thoughts:
- repositories tapped through a migration require fully scoped formulae names e.g. `brew install homebrew/boneyard/foo`
- we somehow only tap formula files as a one-off through a boneyard migration

Relatedly: there's weird behaviour with cached `md5`-using formulae too: https://github.com/Homebrew/brew/issues/614

ilovezfs:
This should fix the boneyard: https://github.com/Homebrew/homebrew-boneyard/pull/137

ilovezfs:
Also needed to remove `fails_with_llvm` from flusspferd (https://github.com/Homebrew/homebrew-boneyard/pull/138), but now the boneyard is tappable again

MikeMcQuaid:
Opened an alternative approach in https://github.com/Homebrew/brew/pull/622.

apjanke:
I think #622 is a better approach. Closing this in favor of that.

PR #616, opened 2016-08-01
keg_relocate: work around `file` bug

DomT4:

-----

There's an old bug in `file` which means it can't read certain files under a non-C locale. This has been fixed upstream for some time, but Apple hasn't picked that fix up & even on OS X El Capitan the `file` is ancient.

This is currently causing a lot of false positives in our bottle code around relocating things like manpages translated into non-English languages, because currently the test does:

```
pn = "/usr/local/Cellar/vim/7.4.2109/share/man/fr/man1/vim.1"
Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text")
```
Which returns `false`. But it isn't returning `false` because the actual result is false, but because `file` panics & fails, which for us equals a `false`. The actual output when accessed is:
```
pn = "/usr/local/Cellar/vim/7.4.2109/share/man/fr/man1/vim.1"
Utils.popen_read("/usr/bin/file", "--brief", pn)
"ERROR: line 22: regexec error 17, (illegal byte sequence)\n"
```
Forcing this check to be done under a "C" locale eliminates this particular false positive for strings we can't relocate & consequently things such as NLS may prove to be more portable in some formulae than is currently the case.

Using `vim` again for the example:
```
pn = "/usr/local/Cellar/vim/7.4.2109/share/man/fr/man1/vim.1"
Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text")
true
```
This reduces the flagged strings from `vim` from 4 issues to 2, the remaining two with the `vim` executable itself which "remembers" the full path to perl, python, ruby, etc during build & vomits that information out when requested by the user. Both the manpages flagged before this change are no longer flagged as unrelocatable.

This won't entirely resolve the NLS problem because some things hardcode in a locale path, which will be stored in the executable, but at the very least it should reduce the number of false positives & may enable relocation where that locale path hasn't been burnt in.

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -64,12 +64,14 @@ class Keg
     text_files = []
     return text_files unless File.exist?("/usr/bin/file")
 
-    path.find do |pn|
-      next if pn.symlink? || pn.directory?
-      next if Metafiles::EXTENSIONS.include? pn.extname
-      # File has known issues with reading files on other locales.
-      # http://bugs.gw.com/view.php?id=292
-      with_custom_locale("C") do
+    # file has known issues with reading files on other locales. Has
+    # been fixed upstream for some time, but a sufficiently new enough
+    # file with that fix is only available in macOS Sierra.
+    # http://bugs.gw.com/view.php?id=292
+    with_custom_locale("C") do
+      path.find do |pn|
+        next if pn.symlink? || pn.directory?
+        next if Metafiles::EXTENSIONS.include? pn.extname
         if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
            pn.text_executable?
           text_files << pn
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -67,9 +67,13 @@ class Keg
     path.find do |pn|
       next if pn.symlink? || pn.directory?
       next if Metafiles::EXTENSIONS.include? pn.extname
-      if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
-         pn.text_executable?
-        text_files << pn
+      # File has known issues with reading files on other locales.
+      # http://bugs.gw.com/view.php?id=292
+      with_custom_locale("C") do
+        if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
+           pn.text_executable?
+          text_files << pn
+        end
       end
     end
 
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -337,6 +337,14 @@ ensure
   ENV["PATH"] = old_path
 end
 
+def with_custom_locale(locale)
+  old_locale = ENV["LC_ALL"]
+  ENV["LC_ALL"] = locale
+  yield
+ensure
+  ENV["LC_ALL"] = old_locale
+end
+
 def run_as_not_developer(&_block)
   old = ENV.delete "HOMEBREW_DEVELOPER"
   yield
--- a/Library/Homebrew/test/test_utils.rb
+++ b/Library/Homebrew/test/test_utils.rb
@@ -78,6 +78,14 @@ class UtilTests < Homebrew::TestCase
     end
   end
 
+  def test_with_custom_locale
+    ENV["LC_ALL"] = "en_US.UTF-8"
+    with_custom_locale("C") do
+      assert_equal "C", ENV["LC_ALL"]
+    end
+    assert_equal "en_US.UTF-8", ENV["LC_ALL"]
+  end
+
   def test_run_as_not_developer
     ENV["HOMEBREW_DEVELOPER"] = "foo"
     run_as_not_developer do


MikeMcQuaid:
Great core PR 👍

:shipit:

DomT4:
Thanks! Will let this sit until this evening to give other maintainers a chance to comment if they wish to, but otherwise I'll merge it then.

apjanke:
:+1:

Library/Homebrew/keg_relocate.rb
```diff
@@ -67,9 +67,13 @@ def text_files
     path.find do |pn|
       next if pn.symlink? || pn.directory?
       next if Metafiles::EXTENSIONS.include? pn.extname
-      if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
-         pn.text_executable?
-        text_files << pn
+      # File has known issues with reading files on other locales.
+      # http://bugs.gw.com/view.php?id=292
+      with_custom_locale("C") do
```
UniqMartin: It's a minor concern, but I'm wondering how much switching back and forth environments for every single file affects performance. I think we could safely move this outside of `path.find do` without adverse effects. The only disadvantage would be that it's a bit more disconnected from the `/usr/bin/file` invocation.

Library/Homebrew/utils.rb
```diff
@@ -337,6 +337,14 @@ def with_system_path
   ENV["PATH"] = old_path
 end
 
+def with_custom_locale(locale)
+  old_locale = ENV["LC_ALL"]
+  ENV["LC_ALL"] = locale
+  yield
+ensure
+  ENV["LC_ALL"] = old_locale
+end
```
UniqMartin: How do you feel about generalizing this method a bit? Making it accept arbitrary changes to the environment would be simple, in fact it would only mean stealing [this code from `testing_env.rb`](https://github.com/Homebrew/brew/blob/7a00f03c92607f65336be0285584e8351ad8a73a/Library/Homebrew/test/testing_env.rb#L128-L136). That way we wouldn't have to come up with a special-purpose method for every case where we need to temporarily alter the environment. On the call site, this would be used as:

```ruby
with_environment("LC_ALL" => "C") { … }
```

Just an idea. Feel free to ignore it, particularly if others also think generalizing this wouldn't be as useful.

UniqMartin:
Have added two code comment of the “nice to have” kind, but generally already :+1: as-is.

Library/Homebrew/keg_relocate.rb
DomT4:
```
vim (Known problematic), before PR:

brew bottle vim  25.07s user 12.28s system 101% cpu 36.910 total
brew bottle vim  24.97s user 11.92s system 100% cpu 36.651 total
brew bottle vim  24.89s user 11.90s system 101% cpu 36.350 total

vim (Known problematic), after PR:

brew bottle vim  24.86s user 11.82s system 100% cpu 36.363 total
brew bottle vim  24.56s user 11.81s system 102% cpu 35.538 total
brew bottle vim  24.59s user 11.65s system 103% cpu 35.085 total


openssl (Known fine), before PR:

brew bottle openssl  17.76s user 7.23s system 102% cpu 24.395 total
brew bottle openssl  17.53s user 7.00s system 103% cpu 23.736 total
brew bottle openssl  17.47s user 6.77s system 103% cpu 23.402 total


openssl (Known fine), after PR:

brew bottle openssl  18.17s user 7.14s system 102% cpu 24.722 total
brew bottle openssl  17.68s user 6.97s system 102% cpu 24.099 total
brew bottle openssl  17.61s user 6.89s system 102% cpu 23.856 total


Martin's proposed rejig:
brew bottle openssl  17.47s user 6.79s system 103% cpu 23.374 total
brew bottle openssl  17.56s user 6.97s system 103% cpu 23.723 total
brew bottle openssl  17.46s user 6.96s system 103% cpu 23.644 total
brew bottle vim  23.81s user 11.49s system 105% cpu 33.582 total
brew bottle vim  23.72s user 11.52s system 106% cpu 33.034 total
brew bottle vim  24.24s user 11.70s system 104% cpu 34.358 total
```

Not a giant different either way, but agree there's a minor performance gain here so will adopt the proposed rejiggle.

Library/Homebrew/utils.rb
DomT4: I lean 👎 on this for now.

I'll reconsider if it gets to the point where we're carrying around a silly amount of custom methods for relatively small cases, but I _personally_ find generic methods make handling a large codebase sometimes unpleasant, particularly around navigating & tying together specific elements efficiently.

But as said, if it gets to a point where this gets silly I'll accept the need. Hopefully that's a reasonable compromise.

tdsmith: Two cents: I wrote the testing code ;) but I think it makes sense in the main codebase and I like that it's more explicit than `with_custom_locale`. I'm happy either way.

DomT4: Let's stick with `with_custom_locale` for now. I kind of want to get it shipped & see whether we see an appreciable difference in bottles being relocatable, and these discussions have tended to get bogged down indefinitely on minor issues like this.

Happy to reconsider later, or even review a follow-up PR to introduce more generic handling throughout the codebase, but yeah, I'm relatively wary of this kind of discussion by now having seen it delay or outright derail core PRs for weeks on end too much over the last ~6 months 😕.

PR #617, opened 2016-08-01
tests: update cmd/analytics integration test

eirinikos:

-----

This change updates `test_analytics` to allow for the possibility that the user has set a value for `ENV["HOMEBREW_NO_ANALYTICS"]`, which was previously overlooked in #558. Thanks to @DomT4 for pointing this out. I also updated the relevant `assert_match` to clarify that `"Analytics is disabled (by HOMEBREW_NO_ANALYTICS)"`.

Is there no way to simply delete `ENV["HOMEBREW_NO_ANALYTICS"]` from within the test? I tried adding `ENV.delete` and `ENV["HOMEBREW_NO_ANALYTICS"] = nil` to the beginning of the test (to no avail).

Thanks.

### DIFF
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -745,16 +745,20 @@ class IntegrationCommandTests < Homebrew::TestCase
       end
     end
 
-    assert_match "Invalid usage", cmd_fail("analytics", "on", "off")
-    assert_match "Invalid usage", cmd_fail("analytics", "testball")
-    assert_match "Analytics is enabled", cmd("analytics")
-    assert_match "Analytics is disabled",
+    assert_match "Analytics is disabled (by HOMEBREW_NO_ANALYTICS)",
       cmd("analytics", "HOMEBREW_NO_ANALYTICS" => "1")
 
-    cmd("analytics", "regenerate-uuid")
-    cmd("analytics", "on")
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> Is there no way to simply delete ENV["HOMEBREW_NO_ANALYTICS"] from within the test? I tried adding ENV.delete and ENV["HOMEBREW_NO_ANALYTICS"] = nil to the beginning of the test (to no avail).

I think your current approach is good/better anyway!

Thanks again @eirinikos!

DomT4:
Thanks for following up on this @eirinikos, Appreciated 😃.

UniqMartin:
> Is there no way to simply delete `ENV["HOMEBREW_NO_ANALYTICS"]` from within the test? I tried adding `ENV.delete` and `ENV["HOMEBREW_NO_ANALYTICS"] = nil` to the beginning of the test (to no avail).

The reason for this is [this line in the `cmd_output` method](https://github.com/Homebrew/brew/blob/a6033529cfdcfc6458aaa83a0bc9c4c141c60f96/Library/Homebrew/test/test_integration_cmds.rb#L70) so all changes to the environment that are supposed to affect the spawned `brew` process need to happen after that (e.g. by passing the environment overrides to `cmd[_fail]` as you did here).

eirinikos:
Thanks @UniqMartin, makes total sense.

PR #623, opened 2016-08-02
bump-formula-pr: format documentation for --help

scpeters:

-----

I've been toying with using the `bump-formula-pr` command, but I couldn't remember the command-line arguments, and they weren't showing up when I pass `--help`, so I formatted the comments based on what I saw in some other commands. Suggestions are welcome of course, but I'm on vacation, so I might not have a chance to respond for at least a week.

I'm pasting the output but it loses some of the formatting:
~~~
brew bump-formula-pr --help
brew bump-formula-pr [--devel] [--dry-run] formula --url=URL --sha256=SHA-256:
    Creates a pull request to update the formula formula
    with new url URL and SHA-256 SHA-256.

    If --devel is passed, bump a devel rather than stable version.

    If --dry-run is passed, print what would be done rather than doing it.

brew bump-formula-pr [--dry-run] [--devel] formula --tag=TAG --revision=REVISION:
    Creates a pull request to update the formula formula
    with new tag TAG and revision REVISION.

    If --devel is passed, bump a devel rather than stable version.

    If --dry-run is passed, print what would be done rather than doing it.
~~~

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -1,18 +1,20 @@
-#:  * `bump-formula-pr` [`--devel`] [`--dry-run`] <formula> `--url=`<URL> `--sha256=`<SHA-256>:
-#:    Creates a pull request to update the formula <formula>
-#:    with new url <URL> and SHA-256 <SHA-256>.
+#:  * `bump-formula-pr` [`--devel`] [`--dry-run`] `--url=`<URL> `--sha256=`<SHA-256> <formula>:
+#:  * `bump-formula-pr` [`--devel`] [`--dry-run`] `--tag=`<TAG> `--revision=`<REVISION> <formula>:
 #:
-#:    If `--devel` is passed, bump a devel rather than stable version.
+#:    Creates a pull request to update the formula with a new url or a new tag.
 #:
-#:    If `--dry-run` is passed, print what would be done rather than doing it.
+#:    If a <URL> is specified, the <SHA-256> checksum of the new download must also be specified.
 #:
-#:  * `bump-formula-pr` [`--dry-run`] [`--devel`] <formula> `--tag=`<TAG> `--revision=`<REVISION>:
-#:    Creates a pull request to update the formula <formula>
-#:    with new tag <TAG> and revision <REVISION>.
+#:    If a <TAG> is specified, the git commit <REVISION> corresponding to that tag must also be specified.
 #:
-#:    If `--devel` is passed, bump a devel rather than stable version.
+#:    If `--devel` is passed, bump the development rather than stable version.
+#:    The development spec must already exist.
 #:
 #:    If `--dry-run` is passed, print what would be done rather than doing it.
+#:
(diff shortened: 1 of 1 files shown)


apjanke:
I kinda prefer the formatting in the old comment block: it's more concise, and options show up in only one place, in a table with options at the beginning of the line so it's easier to scan for them.

apjanke:
Oh, nevermind: I forgot that the formula help text is included in the man page, so uses this more verbose formatting.

MikeMcQuaid:
Looks good to me. @ilovezfs: thoughts?

ilovezfs:
It looks accurate to me. I think we can have multiple synopsis lines without repeating, and formula arguments are usually last. Maybe something like

```
brew bump-formula-pr --help
brew bump-formula-pr [--devel] [--dry-run] --url=URL --sha256=SHA-256 formula
brew bump-formula-pr [--devel] [--dry-run] --tag=TAG --revision=REVISION formula

Creates a pull request to update the formula with a new url or a new tag.

    If a URL is specified, the SHA-256 checksum of the new download must also be specified.

    If a TAG is specified, the git commit REVISION corresponding to that tag must be specified.

    If --devel is passed, bump the development rather than stable version. The development spec must already exist.

    If --dry-run is passed, print what would be done rather than doing it.

Note that brew bump-formula-pr cannot be used to transition a formula from a
url-and-sha256 style specification into a tag-and-revision style specification, nor vice versa.
It must use whichever style specification the preexisting formula already uses. 
```

scpeters:
@ilovezfs: that sounds fine to me. I'll make the change if others concur.

MikeMcQuaid:
I concur.

scpeters:
Updated:

screen shot 2016-08-04 at 2 12 20 pm

UniqMartin:
I think the capitalized placeholders `URL`, `SHA-256`, `TAG`, and `REVISION` look a bit out of place. All existing help texts exclusively use lower-case placeholders. Could you also make sure to hard wrap all non-summary lines so that they stay within an 80-character boundary?

Other than that this is a very welcome documentation improvement. :+1:

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -1,16 +1,23 @@
...
-#   --url:      The new formula URL.
-#   --sha256:   The new formula SHA-256.
-#   --tag:      The new formula's `tag`
-#   --revision: The new formula's `revision`.
+#:  * `bump-formula-pr` [`--devel`] [`--dry-run`] `--url=`<url> `--sha256=`<sha-256> <formula>:
+#:  * `bump-formula-pr` [`--devel`] [`--dry-run`] `--tag=`<tag> `--revision=`<revision> <formula>:
+#:
+#:    Creates a pull request to update the formula with a new url or a new tag.
+#:
+#:    If a <url> is specified, the <SHA-256> checksum of the new download must
```
ilovezfs: this one is still caps

scpeters: thanks for catching that

MikeMcQuaid:
Thanks again @scpeters!

Issue #620, opened 2016-08-02
Handle Versions Better


MikeMcQuaid:
Extracted from https://github.com/Homebrew/brew/issues/60.
- [x] Add code to note in the `INSTALL_RECEIPT` what the name was of the alias used to install (if any) and check that alias in future for upgrades (so e.g. `boost.rb` can be an alias to `boost@1.60.rb` and upgrades to `boost@1.61.rb` when the alias points to that).
- [x] Do not accept new versions in homebrew/versions for any versions above those in core and move old versions back into core.
- [x] Document all this for formula/tap authors (https://github.com/Homebrew/brew/pull/1204)
- [x] Give unversioned formulae that have versioned formulae aliases e.g. `boost@2.0`
- [x] Upgrade immediately to new formulae versions but provide old versions when needed to avoid blocking large upgrades
- [x] At some point in the future if we're on e.g. `boost@3.0` or something (having had `boost@2.1` etc.) and `boost@1.5` no longer compiles we remove any formulae from core that still depend on `boost@1.5` and remove the `boost@1.5` formula itself.

This will allow us to immediately merge e.g. `boost` upgrades and handle the upgrade of dependent formulae on a formula-by-formula basis rather than being blocked on a single formula being upgraded or removed.

DomT4:
I keep trying to get the ball rolling here without much success. Tried to talk people into doing an `llvm38` core formula before giving up & waiting for the core `llvm` formula to be ready to update. Probably my fault for not drawing up a proper timetable outside of my head, but there we go.

> Do not accept new versions in homebrew/versions for any versions above those in core and consider moving old versions back into core.

Would like to get the problem with `brew install/info homebrew/versions/ducks150` resolved on moving formulae before we start reverse migration, but yeah, some stuff will have to come out of Versions in the medium term.

> At some point in the future if we're on e.g. boost30 or something (having had boost21 etc.) we remove any formulae from core that still depend on boost15 and remove the boost15 formula itself.

For the sake of keeping the core actually maintainable I'd like us to follow the same rules we've applied to new Versions submissions, i.e. +/-2 from current release, formulae should be pegged to a specific release branch & not for example `autoconf213`.

MikeMcQuaid:
> For the sake of keeping the core actually maintainable I'd like us to follow the same rules we've applied to new Versions submissions, i.e. +/-2 from current release, formulae should be pegged to a specific release branch & not for example autoconf213.

[...]

PR #625, opened 2016-08-02
dependency: Allow formulas to specify :devel as an option for a dependency

dr-impossible:

-----

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -162,7 +162,13 @@ class FormulaAuditor
     @style_offenses = options[:style_offenses]
     @problems = []
     @text = FormulaText.new(formula.path)
-    @specs = %w[stable devel head].map { |s| formula.send(s) }.compact
+    @specs = %w[stable devel head].map do |spec|
+      if f = formula.send(spec)
+        { :name => spec, :formula => f } 
+      else
+        nil
+      end
+    end.compact
   end
 
   def audit_style
@@ -322,7 +328,7 @@ class FormulaAuditor
     @specs.each do |spec|
       # Check for things we don't like to depend on.
       # We allow non-Homebrew installs whenever possible.
-      spec.deps.each do |dep|
+      spec[:formula].deps.each do |dep|
         begin
           dep_f = dep.to_formula
         rescue TapFormulaUnavailableError
@@ -348,7 +354,7 @@ class FormulaAuditor
         end
 
         dep.options.reject do |opt|
-          next true if opt.name == :devel
+          next true if opt.name == :devel && spec[:name] == "devel"
           next true if dep_f.option_defined?(opt)
           dep_f.requirements.detect do |r|
             if r.recommended?
@@ -358,7 +364,11 @@ class FormulaAuditor
             end
           end
         end.each do |opt|
-          problem "Dependency #{dep} does not define option #{opt.name.inspect}"
+          if opt.name == :devel
+            problem "Don't specify a devel dependency for non-devel builds"
+          else
+            problem "Dependency #{dep} does not define option #{opt.name.inspect}"
+          end
         end
 
         case dep.name
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -348,6 +348,7 @@ class FormulaAuditor
         end
 
         dep.options.reject do |opt|
+          next true if opt.name == :devel
           next true if dep_f.option_defined?(opt)
           dep_f.requirements.detect do |r|
             if r.recommended?
--- a/Library/Homebrew/dependency.rb
+++ b/Library/Homebrew/dependency.rb
@@ -29,7 +29,11 @@ class Dependency
   end
 
   def to_formula
-    formula = Formulary.factory(name)
+    if tags.include?(:devel)
+      formula = Formulary.factory(name, :devel)
+    else
+      formula = Formulary.factory(name)
+    end
     formula.build = BuildOptions.new(options, formula.options)
     formula
   end
--- a/Library/Homebrew/dependency.rb
+++ b/Library/Homebrew/dependency.rb
@@ -29,10 +29,10 @@ class Dependency
   end
 
   def to_formula
-    if tags.include?(:devel)
-      formula = Formulary.factory(name, :devel)
+    formula = if tags.include?(:devel)
+      Formulary.factory(name, :devel)
     else
-      formula = Formulary.factory(name)
+      Formulary.factory(name)
     end
     formula.build = BuildOptions.new(options, formula.options)
     formula
--- a/share/doc/homebrew/Formula-Cookbook.md
+++ b/share/doc/homebrew/Formula-Cookbook.md
@@ -136,10 +136,12 @@ A Symbol (e.g. `:x11`) specifies a [`Requirement`](http://www.rubydoc.info/githu
 
 A Hash (e.g. `=>`) specifies a formula dependency with some additional information. Given a single string key, the value can take several forms:
 
-*   a Symbol (currently one of `:build`, `:optional`, `:recommended`).
+*   a Symbol (currently one of `:build`, `:devel`, `:optional`, `:recommended`).
     - `:build` means that dependency is a build-time only dependency so it can
       be skipped when installing from a bottle or when listing missing
       dependencies using `brew missing`.
+    - `:devel` generates an implicit `devel` option, meaning that the development
+      version of the dependency will be installed.
     - `:optional` generates an implicit `with-foo` option for the formula.
       This means that, given `depends_on "foo" => :optional`, the user must pass `--with-foo` in order to use the dependency.
     - `:recommended` generates an implicit `without-foo` option, meaning that
--- a/Library/Homebrew/test/test_cmd_audit.rb
+++ b/Library/Homebrew/test/test_cmd_audit.rb
@@ -78,6 +78,11 @@ class FormulaAuditorTests < Homebrew::TestCase
     FormulaAuditor.new Formulary.factory(path), options
   end
 
+  def formula_dependency(name, text)
+    path = Pathname.new "#{@dir}/#{name}.rb"
+    path.open("w") { |f| f.write text }
+  end
+
   def test_init_no_problems
     fa = formula_auditor "foo", <<-EOS.undent
       class Foo < Formula
@@ -379,5 +384,61 @@ class FormulaAuditorTests < Homebrew::TestCase
     assert_equal [], fa.problems
   ensure
     ENV["HOMEBREW_NO_GITHUB_API"] = original_value
+  end
+
+  def test_audit_dependency_devel_in_stable
+    bar = formula_dependency "bar", <<-EOS.undent
+      class Bar < Formula
+        homepage "https://github.com/example/bar"
+        url "http://example.com/bar-1.0.tgz"
+
+        devel do
+          url "http://example.com/bar-1.9.tgz"
+        end
+      end
+    EOS
+
+    foo = formula_auditor "foo", <<-EOS.undent, :strict => true, :online => false
+      class Foo < Formula
+        homepage "https://github.com/example/example"
+        url "http://example.com/foo-1.0.tgz"
+
+        depends_on "#{@dir}/bar.rb" => :devel
+      end
+    EOS
+
+    foo.audit_deps
+    assert_equal ["Don't specify a devel dependency for non-devel builds"], foo.problems
+  end
+
+  def test_audit_dependency_devel_in_devel
+    bar = formula_dependency "bar", <<-EOS.undent
+      class Bar < Formula
+        homepage "https://github.com/example/bar"
+        url "http://example.com/bar-1.0.tgz"
+
+        devel do
+          url "http://example.com/bar-1.9.tgz"
+        end
+      end
+    EOS
+
+    foo = formula_auditor "foo", <<-EOS.undent, :strict => true, :online => false
+      class Foo < Formula
+        homepage "https://github.com/example/example"
+        url "http://example.com/foo-1.0.tgz"
+
+        depends_on "#{@dir}/bar.rb"
+
+        devel do
+          url "http://example.com/foo-1.9.tgz"
+
+          depends_on "#{@dir}/bar.rb" => :devel
+        end
+      end
+    EOS
+
+    foo.audit_deps
+    assert_equal [], foo.problems
   end
 end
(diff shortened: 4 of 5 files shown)


dr-impossible:
Currently, it's not possible for a formula to specify the `--devel` option on a dependency. This patch allows that. 

If you define a package with a `devel` block and override the dependencies to also use the `--devel` option, the dependencies fail to build.

Case in point: `gstreamer`. 1.9.1 is the current "devel" version of gstreamer, and all related devel packages in gstreamer require their dependencies to have the same version. For example:

If I add the following to `gst-plugins-base.rb`:
```ruby
  devel do
    url "https://gstreamer.freedesktop.org/src/gst-plugins-base/gst-plugins-base-1.9.1.tar.xz"
    sha256 "8309d1660406274a94bd056f8d805fb35b902208fbb76ff8ca705b37d950fdc9"

    depends_on "gstreamer" => "devel"
  end
```

and the following to `gstreamer.rb`:
```ruby
  devel do
    url "https://gstreamer.freedesktop.org/src/gstreamer/gstreamer-1.9.1.tar.xz"
    sha256 "55304a9e1e8fb5ef82b5b246fef2d9a164eeb0466976a3ad19cfae05b8a94159"
  end
```

and then run the following command:
```bash
brew install gst-plugins-base --devel
```

Homebrew appears to build gstreamer-1.9.1 as a dependency, and is able to run `make` and `make install`, however at some point it selects version 1.8.2 and responds with the error message `Error: empty installation`
Adding the `--debug` option yields the message "Error: You must `brew link gstreamer` before gst-plugins-base can be installed"

DomT4:
Broadly speaking, I think this isn't a good idea, at least until we get `devel` bottles sorted  & shipped. Bottles reduce our maintenance burden significantly, largely because they eliminate the most common breaking factor _(user configuration)_.

A relatively simple change of syntax here somewhat understates how large a change this would be & the impact it would have on maintainers, users & contributors.

There are deeper questions about how to gently degrade if something demands a `:devel` dependency but the formula required doesn't include `devel`. There's also the question of whether it's good UX to allow stable formula to require `:devel` dependencies without user intervention, or whether that's something we'd want to block off or frown upon.

dr-impossible:
I must admit I wasn't aware of the plan to have devel bottles. I'd assumed that using `--devel` came with the similar caveats and pitfalls of using `--head`

I think it's a bit weird to allow a `stable` formula to require a `:devel` dependency. That case would suggest that either the formula isn't really stable, or that the formula for the dependency is outdated.

DomT4:
>  I'd assumed that using --devel came with the similar caveats and pitfalls of using --head

Not as much. We CI test `--devel`, so we know it builds, we know it can be fetched, we know the checksum was correct at time of merge, we know it passes `brew test`, etc. There's no testing on `--HEAD` at all, and likely never will be due to the transient nature of that.

> I think it's a bit weird to allow a stable formula to require a :devel dependency. That case would suggest that either the formula isn't really stable, or that the formula for the dependency is outdated.

It has been requested a few times. The use case is largely where one formula is updated with new releases at a much more rapid cadence than the formula it `depends_on`. I'd lean towards agreeing with you that to me that implies their formula isn't really stable, but that's something that has been requested & we may want to draw a hard line in the sand one way or another if this goes forwards.

dr-impossible:
If you currently override a dependency by adding the `:devel` option within the `devel` block of a formula, `brew audit` will complain but the formula will still build. When you install the formula with `brew install --devel` it appears to download and build the devel version of the dependency, but after the build step, Homebrew checks the prefix path for the stable version of the dependency, which won't be there unless that version was installed as well.

This current behaviour seems a bit broken to me, so I think it would be better for Homebrew to either ignore the `:devel` option on `depends_on` or actually follow it through.

dr-impossible:
For reference, linking to a PR containing an example of the `:devel` option: https://github.com/Homebrew/homebrew-core/pull/3562

dr-impossible:
> There are deeper questions about how to gently degrade if something demands a :devel dependency but the formula required doesn't include devel. 

Based on [this test](https://github.com/Homebrew/brew/blob/028d155e97014076adf4e4dbc7c888fd871de321/Library/Homebrew/test/test_formula_spec_selection.rb#L91), I think if a formula specifies a dependency on `devel` and it's not there, it silently falls back to `stable`:
```ruby
  def test_does_not_raise_for_missing_spec
    f = formula("test", Pathname.new(__FILE__).expand_path, :devel) {
      url "foo-1.0"
      head "foo"
    }

    assert_predicate f, :stable?
  end
```

MikeMcQuaid:
> Based on this test, I think if a formula specifies a dependency on devel and it's not there, it silently falls back to stable:

Yes, it feels like this should 💥 instead.

> I think it's a bit weird to allow a stable formula to require a :devel dependency. That case would suggest that either the formula isn't really stable, or that the formula for the dependency is outdated.

Agreed.

> This current behaviour seems a bit broken to me, so I think it would be better for Homebrew to either ignore the :devel option on depends_on or actually follow it through.

Seems a good enough justification to me.

Library/Homebrew/cmd/audit.rb
```diff
@@ -348,6 +348,7 @@ def audit_deps
         end
 
         dep.options.reject do |opt|
+          next true if opt.name == :devel
```
MikeMcQuaid: Probably want to extend this to if it's `:devel` and the relevant dependency also defines a `devel` spec.

Library/Homebrew/dependency.rb
```diff
@@ -29,7 +29,11 @@ def hash
   end
 
   def to_formula
-    formula = Formulary.factory(name)
+    if tags.include?(:devel)
```
MikeMcQuaid: Can make the `formula = if tags`... instead

MikeMcQuaid:
Tests failing on 10.9: http://bot.brew.sh/job/Brew%20Pull%20Requests/1169/version=mavericks/testReport/junit/brew-test-bot/mavericks/tests/

Library/Homebrew/cmd/audit.rb
dr-impossible: I'll need to dig a bit deeper to figure out how to do this. Whenever I run `brew audit` on a sample formula that uses this, `spec.build.devel?` is always false unless I run `brew audit --devel`, which isn't what we want. Do you happen to know off-hand if there's somewhere else I should look to detect whether the dependency exists within a `devel` block?

ilovezfs:
How is this going to handle the case where the dependency's development version becomes the stable version (or the stable version becomes newer than the old devel version), the devel spec in the dependency is also upgraded to a new development version, and the dependent formula still needs the original version not the new development version? This idea seems like a can of worms.

ilovezfs:
I believe the answer to my question is that any kind of `=> :devel` thing should have to hard-code a specific version and that version should have to appear in the dependency's devel spec.

Also, I think stable specs should probably never be permitted to depend on `=> :devel` and that should cause both install and audit failures.

share/doc/homebrew/Formula-Cookbook.md
```diff
@@ -136,10 +136,12 @@ A Symbol (e.g. `:x11`) specifies a [`Requirement`](http://www.rubydoc.info/githu
 
 A Hash (e.g. `=>`) specifies a formula dependency with some additional information. Given a single string key, the value can take several forms:
 
-*   a Symbol (currently one of `:build`, `:optional`, `:recommended`).
+*   a Symbol (currently one of `:build`, `:devel`, `:optional`, `:recommended`).
     - `:build` means that dependency is a build-time only dependency so it can
       be skipped when installing from a bottle or when listing missing
       dependencies using `brew missing`.
+    - `:devel` generates an implicit `devel` option, meaning that the development
+      version of the dependency will be installed.
```
ilovezfs: I don't see the use case for this. It smells like feature patching. If the software require's a devel dependency, it should be in an explicit development spec of the main software.

ilovezfs:
The situation this PR seeks to address is pretty rare, and I think vendoring or upstream submodules are usually a better answer for Homebrew. Creating inter-formulae dependencies on `devel` specs would mean `test-bot` would need to understand the scenario and test for breakage in the same way that it tests reverse dependency's for broken `brew test` tests and broken dylibs.

It also complicates the version/upgrade logic as well, beyond just the immediate formula's predicament when the devel spec of the dependency change. For example, does this PR mean I can end up needing multiple versions of the same dependency installed simultaneously? If other software I have installed depends on the stable version, does a devel dependency mean that the software just refuses to install and forces me to choose between mutually exclusive formulae versions, or does it proceed to break silently the other package that needs the stable version of the dependency? What happens at brew upgrade time if a devel version of an installed formula goes from needing a stable version of a dependency to needing a devel version of the dependency, but other packages already installed still need the stable version?

What happens if one of the other dependencies of the formula depending on a devel version depends on a stable version of the same software? Does it have to have its own devel spec added that depends on that devel version? What if its of that software is only optional? Does brew reinstall of that dependency with that option get blocked since it will break the reverse dependency's need for the devel version?

And how will all of this play nicely with the idea in https://github.com/Homebrew/brew/issues/620?

ilovezfs:
>This current behaviour seems a bit broken to me, so I think it would be better for Homebrew to either ignore the :devel option on depends_on or actually follow it through.

This is actually because we basically don't do any option verification. This works without a hitch:
```
diff --git a/Formula/godep.rb b/Formula/godep.rb
index b117cf7..637186a 100644
--- a/Formula/godep.rb
+++ b/Formula/godep.rb
@@ -14,7 +14,7 @@ class Godep < Formula
     sha256 "cdbf4b630a90d8cc4d36076a963916fc979311b2acc413865512b78a41c3aed4" => :mavericks
   end
 
-  depends_on "go"
+  depends_on "go" => [:sparkles, "pokemon", :devel, "newest", "with-lots-of-widgets"]
 
   def install
     ENV["GOPATH"] = buildpath
```
So if that behavior concerns you, option verification seems a bit more relevant that fixating on the string "devel" as being special or justifying any particular change to anything.

Regarding this PR, see https://github.com/Homebrew/homebrew-core/issues/3665 for some issues that arise. For instance, if `godep`'s devel spec depended on `go`'s devel spec, none of the formulae that use `go` and/or `godep` would work on 10.12 since they'd `depend on "go"`, not `depend_on "go" => :devel` and `go`'s stable spec doesn't satisfy the maximum OS requirement.

So would we end up having to add `depends_on "go" => :devel` to all of the formulae that use `go` in a devel spec for each one?

But then how would you build a preexisting devel spec of one of those formulae with the stable version of `go` if you were on 10.11 and didn't want to use new devel `go` but *did* want to use the devel version of that particular formula as you had the day before this change was made?

(No we don't support 10.12 yet, but that's irrelevant; this is just to illustrate what happens when you have a max OS requirement in stable but not in devel.)

share/doc/homebrew/Formula-Cookbook.md
MikeMcQuaid: Agreed 👍

MikeMcQuaid:
> I believe the answer to my question is that any kind of => :devel thing should have to hard-code a specific version and that version should have to appear in the dependency's devel spec.

👎 on having specific versions; this wants "version or greater" which we don't really support or want.

There's lots of other questions here that need thought through though.

ilovezfs:
@MikeMcQuaid I think it's relevant here to address the issue raised in https://github.com/Homebrew/homebrew-core/issues/3665#issuecomment-238024925

Is that indeed a bug? Is an up-to-date development version (and also presumably HEAD version) supposed to be able to stand-in for the stable version when fulfilling dependencies and requirements? That seems important to understand in order to be able to assess the meaning of the present PR and how it would impact the ecosystem.

MikeMcQuaid:
@ilovezfs It's not a bug. It used to be `--devel` was applied to everything you were installing at the time but that approach has obvious problems when comparing a dependency that's installed now vs. previously installed.

ilovezfs:
@MikeMcQuaid that's what I suspected. In that case, is it possible to do if MacOS.version == :sierra, stable = Go 1.7rc5 else stable =  "1.6.3"? I think that will fix Homebrew/homebrew-core#3665 without breaking the new approach.

MikeMcQuaid:
@ilovezfs Yeh, could do that. Probably not a bad idea for now at least.

ilovezfs:
I'll simulate it on El Cap using a max OS Yosemite requirement and see if it works.

MikeMcQuaid:
@dr-impossible Any thoughts on addressing these comments (and resolving the conflicts)?

ilovezfs:
@dr-impossible any update here?

jeffheifetz:
I'll comment for @dr-impossible.

> It also complicates the version/upgrade logic as well, beyond just the immediate formula's predicament when the devel spec of the dependency changes. For example, does this PR mean I can end up needing multiple versions of the same dependency installed simultaneously? If other software I have installed depends on the stable version, does a devel dependency mean that the software just refuses to install and forces me to choose between mutually exclusive formulae versions, or does it proceed to break silently the other package that needs the stable version of the dependency? What happens at brew upgrade time if a devel version of an installed formula goes from needing a stable version of a dependency to needing a devel version of the dependency, but other packages already installed still need the stable version?

I believe the project has already taken the approach to never have more than one version of anything linked be it stable, devel or HEAD. But this situation where you have one installed and need another can arise today already and you are forced to choose between mutually exclusive formulae. At least with better dependency listing we can offer better reasoning to the user on which formulae are forcing this situation.

In the example this was drawn from, the installer chooses the wrong version and simply fails.

> What happens if one of the other dependencies of the formula depending on a devel version depends on a stable version of the same software? Does it have to have its own devel spec added that depends on that devel version? What if its use of that software is only optional? Does brew reinstall of that dependency with that option get blocked since it will break the reverse dependency's need for the devel version? Or does the supposedly stable version I'm reinstalling with that option silently start depending on the devel version of the dependency even though stable specs are supposed to only use stable versions of their dependencies?

Again I'd argue that once a user has installed a devel or HEAD formula they can run into issues installing optional formulae. I agree that specifically allowing a devel spec will exacerbate these issues, but it's not as though they don't exist today.


Overall, I think there are a slew of issues that can arise considering the near infinite combinations of packages and dependency trees that can come up. But these issues already can come up today, its simply a question of whether we want to continue to relegate them to the sidelines or not. The problem is exacerbated once we explicitly allow devel dependencies, but hopefully with this better information we can guide users as to how to fix them better.

MikeMcQuaid:
I agree we need to fix this but this PR as-is isn't mergeable, unfortunately and it's been a while since we saw any updates here. Feel free to reopen a new one with conflicts and comments addressed and we can continue discussion there.

jeffheifetz:
I can't seem to find the failing tests anymore, are they secretly cached anywhere?

zmwangx:
No, brew PR builds are kept for 7 days.

MikeMcQuaid:
You should be able to reproduce with `brew tests` locally.

PR #601, opened 2016-07-29
tests: add cmd/migrate integration test

eirinikos:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

This test gives 100% coverage to `brew migrate`. Many thanks to @UniqMartin for explaining how the command works.

It's 50 lines long, though -- sorry. 😬 
It's very likely there's a more efficient way to achieve the same test coverage.

PR #626, opened 2016-08-02
tests: extend cmd/install integration test

eirinikos:

-----

This change adds more test coverage for `brew install`.

There is some overlap between this new `test_install` and `test_migrate` (in #601); this is so that [these lines](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/install.rb#L137-L142) in `cmd/install.rb` can be covered. 😬 

There are ~25 lines that are duplicated between the two tests; perhaps there's a way to shorten them and/or it would be OK to refactor them into their own method?

### DIFF
--- a/Library/Homebrew/test/test_integration_cmds.rb
+++ b/Library/Homebrew/test/test_integration_cmds.rb
@@ -265,9 +265,24 @@ class IntegrationCommandTests < Homebrew::TestCase
   end
 
   def test_install
-    setup_test_formula "testball"
+    setup_test_formula "testball1"
+    assert_match "Specify `--HEAD`", cmd_fail("install", "testball1", "--head")
+    assert_match "No head is defined", cmd_fail("install", "testball1", "--HEAD")
+    assert_match "No devel block", cmd_fail("install", "testball1", "--devel")
+    assert_match "#{HOMEBREW_CELLAR}/testball1/0.1", cmd("install", "testball1")
+    assert_match "testball1-0.1 already installed", cmd("install", "testball1")
+    assert_match "MacRuby is not packaged", cmd_fail("install", "macruby")
+    assert_match "No available formula", cmd_fail("install", "formula")
+    assert_match "This similarly named formula was found",
+      cmd_fail("install", "testball")
 
-    assert_match "#{HOMEBREW_CELLAR}/testball/0.1", cmd("install", "testball")
+    setup_test_formula "testball2"
+    assert_match "These similarly named formulae were found",
+      cmd_fail("install", "testball")
+
+    install_and_rename_coretap_formula "testball1", "testball2"
+    assert_match "testball1 already installed, it's just not migrated",
+      cmd("install", "testball2")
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> would be OK to refactor them into their own method?

Sounds like a good idea!

eirinikos:
> > would be OK to refactor them into their own method?
> 
> Sounds like a good idea!

Cool! I've taken a stab at this. 🔪  I'll split it out into its own PR when everyone has given feedback and is happy with the change.

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -159,6 +159,35 @@ def setup_remote_tap(name)
...
+    end
+
+    cmd("install", old_name)
+    (core_tap.path/"Formula/#{old_name}.rb").unlink
+    formula_renames = core_tap.path/"formula_renames.json"
+    formula_renames.write <<-EOS.undent
+    {
+      "#{old_name}\": \"#{new_name}"
+    }
+    EOS
```
UniqMartin: The lines between `<<-EOS.undent` and `EOS` should be indented by two spaces. The escaped double quotation signs also look a bit wrong here. However, even better than constructing JSON data by hand (where it's easy to make a mistake) would be to just use:

```ruby
formula_renames.write Utils::JSON.dump(old_name => new_name)
```

UniqMartin:
> Cool! I've taken a stab at this. 🔪  I'll split it out into its own PR when everyone has given feedback and is happy with the change.

This looks mostly good, aside from one code comment. I think it's sufficient to split this refactoring off into a separate commit and then extend the `install` test in a second commit. There's no need to create an entirely new PR for this.

MikeMcQuaid:
> I think it's sufficient to split this refactoring off into a separate commit and then extend the install test in a second commit. There's no need to create an entirely new PR for this.

Agreed 👍

Library/Homebrew/test/test_integration_cmds.rb
eirinikos: Thanks! I did make some mistakes in the process of manually constructing the JSON data. Good to know about this `Utils::JSON` method.

eirinikos:
> This looks mostly good, aside from one code comment. I think it's sufficient to split this refactoring off into a separate commit and then extend the `install` test in a second commit. There's no need to create an entirely new PR for this.

Cool, makes sense. 👍

UniqMartin:
This looks great! Merged in c8b0a55c097f13556a27223b15196a3e23e36483 and 8c0b72ac0da44be8cc4655996a336caa46f21694; thanks again @eirinikos! :tada:

> Coverage increased (+0.2%) to 57.559%

And finally we're getting some useful numbers on your progress and the [coverage graph](https://coveralls.io/github/Homebrew/brew) over at Coveralls is useful again.

MikeMcQuaid:
Nice work @eirinikos 😻

PR #627, opened 2016-08-03
formula_installer: option to ignore deps' tabs

ilovezfs:
This allows test-bot to avoid having to explicitly uninstall a formula
just because a contradictory tab is lingering.

### DIFF
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -159,6 +159,10 @@ module HomebrewArgvExtension
     include? "--ignore-dependencies"
   end
 
+  def ignore_tabs?
+    include? "--ignore-tabs"
+  end
+
   def only_deps?
     include? "--only-dependencies"
   end
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -427,7 +427,7 @@ class FormulaInstaller
     end
 
     fi = DependencyInstaller.new(df)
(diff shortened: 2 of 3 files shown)


ilovezfs:
See discussion in https://github.com/Homebrew/homebrew-core/pull/3556.
CC @MikeMcQuaid

Library/Homebrew/extend/ARGV.rb
```diff
@@ -159,6 +159,10 @@ def ignore_deps?
     include? "--ignore-dependencies"
   end
 
+  def ignore_tabs?
+    include? "--ignore-tabs"
```
MikeMcQuaid: Maybe an environment variable instead might be nicer?

Library/Homebrew/formula_installer.rb
```diff
@@ -427,7 +427,7 @@ def install_dependency(dep, inherited_options)
     end
 
     fi = DependencyInstaller.new(df)
-    fi.options           |= tab.used_options
+    fi.options           |= tab.used_options unless ARGV.ignore_tabs?
```
MikeMcQuaid: Can probably/may need to add this to a few places where `tab.used_options` are used.

ilovezfs:
I don't think this is the right solution. Probably a better idea would be to restore the bottle after a reverse-dependency test forces a non-default build, or to not test those reverse-dependencies with test-bot since they're nearly the same as formulae without bottles since they're forcing full rebuilds.

Issue #38, opened 2016-04-06
Make `revision` mean only one thing.


DomT4:
We keep confusing people by using the magical word `revision` to mean different things.

We have:

* Bottle revisions:
```ruby
bottle do
  revision 1
  sha256 "41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3" => :blah
end
```
* Git revisions:
```ruby
url "https://github.com/magic/unicorns.git",
    :tag => "unicorn-1",
    :revision => "eefc6ed865ca20447afc293c986676933dbfb137"
```
* Formulae revisions:
```ruby
url "https://example.org/example.tar.gz"
sha256 "13550350a8681c84c861aac2e5b440161c2b33a3e4f302ac680ca5b686de48de"
revision 1
```

Particularly confusing to new contributors is that we have an audit mechanism that nudges people to remove an unnecessary `revision`, and we see repeated examples of people removing the git commit `revision` rather than the `revision 1`.

We need to move away, long-term, from using the same keyword to represent different things, particularly where those things are used in close proximity on a regular basis. It'd probably be easier to change formulae/bottle meaning of `revision`, IMO, but discuss at will.

apjanke:
:+1: I still get confused about this myself sometimes, and a terminology change - especially w a keyword change to remove context sensitivity in formula files - would probably help.

Might be a bit harder to do the batch change, but maybe keeping 'revision' for formula would work too, with ':revision' for git resources becoming ':commit', since "commit" is in wide informal use that way and associated with Git, and 'bottle_revision' for the bottles. We could prob do a bulk search-and-replace of `:revision => "\w+"` since nothing else uses that symbol AFAIK, and formatting is consistent. I can't easily think of another term to use for formula revisions, but maybe that's just because I'm used to it now.

Though semantically and conversationally, it might be cleaner to change the terms for formula and bottle: since we track our formulae in git, and don't bump the formula revision with every file change, a formula revision may live for several git revisions of the formula file. And all the git revisions of a formula file over its life do not correspond to "revisions" of that formula. (E.g. If I say "what revision of asciidoc are you looking at?" I might be referring to the git revision SHA of `asciidoc.rb` or the `revision 2` inside the file. Can't tell without more context.)

Git is the only one of these three concepts that's from outside Homebrew and has pre-existing terminology, and it uses "revision". So maybe that should take precedence.

UniqMartin:
[...]

PR #636, opened 2016-08-05
virtualenv_install_with_resources: select correct python

ilovezfs:
`virtualenv_install_with_resources` will now attempt to guess the
desired Python based on the active build options and based on the
dependencies and requirements. When the situation is ambiguous (e.g.,
`depends_on :python3` and `build.with? "python"` is true) raise
`FormulaAmbiguousPythonError` unless `:using => "python"` or
`:using => "python3"` has been passed to resolve the ambiguity.

In most cases, this will allow

```
    virtualenv_create(libexec, "python3")
    virtualenv_install_with_resources
```

to be changed to just

```
    virtualenv_install_with_resources
```

### DIFF
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -139,11 +139,33 @@ module Language
         venv
       end
 
+      # Returns true if a formula option for the specified python is currently
+      # active or if the specified python is required by the formula. Valid
+      # inputs are "python", "python3", :python, and :python3. Note that
+      # "with-python", "without-python", "with-python3", and "without-python3"
+      # formula options are handled correctly even if not associated with any
+      # corresponding depends_on statement.
+      # @api private
+      def needs_python?(python)
+        return true if build.with?(python)
+        (requirements.to_a | deps).any? { |r| r.name == python && r.required? }
+      end
+
       # Helper method for the common case of installing a Python application.
       # Creates a virtualenv in `libexec`, installs all `resource`s defined
-      # on the formula, and then installs the formula.
-      def virtualenv_install_with_resources
-        venv = virtualenv_create(libexec)
+      # on the formula, and then installs the formula. An options hash may be
+      # passed (e.g., :using => "python3") to override the default, guessed
+      # formula preference for python or python3, or to resolve an ambiguous
+      # case where it's not clear whether python or python3 should be the
+      # default guess.
+      def virtualenv_install_with_resources(options = {})
+        python = options[:using]
+        if python.nil?
+          wanted = %w[python python3].select { |py| needs_python?(py) }
+          raise FormulaAmbiguousPythonError, self if wanted.size > 1
+          python = wanted.first || "python"
+        end
+        venv = virtualenv_create(libexec, python)
         venv.pip_install resources
         venv.pip_install_and_link buildpath
         venv
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -294,6 +294,17 @@ class FormulaConflictError < RuntimeError
   end
 end
 
+class FormulaAmbiguousPythonError < RuntimeError
+  def initialize(formula)
+    super <<-EOS.undent
+      The version of python to use with the virtualenv in the `#{formula.full_name}` formula
+      cannot be guessed automatically. If the simultaneous use of python and python3
+      is intentional, please add `:using => "python"` or `:using => "python3"` to
+      `virtualenv_install_with_resources` to resolve the ambiguity manually.
+    EOS
+  end
+end
+
 class BuildError < RuntimeError
   attr_reader :formula, :env
 


Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,45 @@ def virtualenv_create(venv_root, python = "python", formula = self)
...
+          end
+      end
+
       # Helper method for the common case of installing a Python application.
       # Creates a virtualenv in `libexec`, installs all `resource`s defined
       # on the formula, and then installs the formula.
-      def virtualenv_install_with_resources
-        venv = virtualenv_create(libexec)
+      def virtualenv_install_with_resources(*args)
+        options = args[-1].is_a?(Hash) ? args.pop : {}
```
UniqMartin: We tend to prefer `args.last` instead of `args[-1]`. However, I wonder why we need this handling at all if an `options = {}` in the method signature will do just fine. It's not like we need the flexibility of mixing an arbitrary number of non-option arguments (currently none) and option arguments right now.

ilovezfs: @UniqMartin https://github.com/Homebrew/brew/blob/master/Library/Homebrew/language/haskell.rb#L75

Just trying to keep some consistency.

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,45 @@ def virtualenv_create(venv_root, python = "python", formula = self)
...
 
+      def needs_python2?
+        requirements.any? do |req|
+          req.is_a?(PythonRequirement) && !req.is_a?(Python3Requirement) &&
+            (req.required? || build.with?(req))
+        end ||
+          deps.any? do |dep|
+            dep.name == "python" && (dep.required? || build.with?(dep))
+          end
+      end
```
UniqMartin: The problem with this method is that absence of any Python 2 requirement or hard-coded dependency doesn't mean the formula doesn't support or want the system Python, thus `python2`. Even checking the stand-alone `--with-python` or `--without-python` options isn't sufficient, unless I'm misreading [`Language::Python.each_python`](https://github.com/Homebrew/brew/blob/38209aadbfe4fd0c6772467c4bc5c63325d53f6c/Library/Homebrew/language/python.rb#L16-L29). That means it should be legal that a formula specifies

```ruby
depends_on :python if MacOS.version <= :snow_leopard
depends_on :python3
```

with the intention of always providing both Python 2 and Python 3 support.

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,45 @@ def virtualenv_create(venv_root, python = "python", formula = self)
         venv
       end
 
+      def needs_python2?
+        requirements.any? do |req|
+          req.is_a?(PythonRequirement) && !req.is_a?(Python3Requirement) &&
```
UniqMartin: Not your fault, but it feels like `PythonRequirement` and `Python3Requirement` having a parent-child relationship instead of being siblings is a design flaw that is exposed by this.

Library/Homebrew/language/python.rb
ilovezfs: That's why it can be manually specified as well. The goal is not to read the formula authors' minds but to provide better default behavior.

Library/Homebrew/language/python.rb
ilovezfs: You read my mind ;)

Library/Homebrew/language/python.rb
UniqMartin: Acknowledged. It's a flexibility that we don't need here and it's unnecessarily bloating the method. `args` is never used beyond this point (contrary to the method you linked).

ilovezfs: Changed.

Library/Homebrew/language/python.rb
UniqMartin: I think the way forward here, which will also massively simplify the logic and avoid having separate methods for Python 2 and Python 3 is to just use `Requirement#name`, i.e. modify this check to be `req.name == "python"` or `req.name == "python3"`.

Library/Homebrew/language/python.rb
UniqMartin: I agree that it does a decent job providing a better default and I'm happy with this. Just wanted to point out that it won't be able to get all cases right automagically.

ilovezfs: Yeah, I agree. I think the main question is whether there are more ambiguous cases that should force an error unless it's specified manually. It seems reasonable to me that in the above scenario
```
depends_on :python if MacOS.version <= :snow_leopard
depends_on :python3
```
that the default would be python3 without forcing it to be specified manually. Ditto in the case of "without-python" and :python3. But I could see someone reasonably wanting those cases to raise unless manually specified.

Library/Homebrew/language/python.rb
ilovezfs: Are you suggesting a single method that accepts a version argument that's a "2" or a "3"?

UniqMartin: I'm suggesting a method that is invoked as either `needs_python?("python")` or `needs_python?("python3")`. This would also help simplify the logic in `virtualenv_install_with_resources`.

ilovezfs: Or maybe you mean a generic `needs_req_or_dep_with_name?(fooreqdepname)`?

UniqMartin: > Or maybe you mean a generic `needs_req_or_dep_with_name?(fooreqdepname)`?

That's what it basically boils down and the method name is certainly very descriptive, but I'm not insisting on generalizing this (unless existing code paths in Homebrew would immediately benefit from such a generalization).

ilovezfs: I guess the same pattern is used for both perl and ruby but they don't have elaborate DSLs (yet) that would benefit at this point.

So one question is whether to also include dependency-less and requirement-less `without-python` and `with-python` options as cases for needs_python?("python") to return true.

UniqMartin: > So one question is whether to also include dependency-less and requirement-less `without-python` and `with-python` options as cases for `needs_python?("python")` to return true.

I think this should be the case. That's a very widely used pattern in our formulae.

ilovezfs: As for the case of `:python` if snow leopard + `:python3` scenario, it occurs in `gdal`, `hashpump`, `pyside`, `shiboken`, `sip`, `uwsgi`, and `z3`.

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,34 @@ def virtualenv_create(venv_root, python = "python", formula = self)
         venv
       end
 
+      def needs_python?(py)
```
ilovezfs: @UniqMartin ok so here's a combo method

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,34 @@ def virtualenv_create(venv_root, python = "python", formula = self)
...
+        requirements.any? do |req|
+          req.name == py && (req.required? || build.with?(req))
+        end ||
+          deps.any? do |dep|
+            dep.name == py && (dep.required? || build.with?(dep))
+          end
```
UniqMartin: I have to say that I dislike the duplication in these `<whatever>.any? do` blocks and even more so that those multi-line blocks are joined via a Boolean operation. Can you try and make this a bit nicer?

Library/Homebrew/language/python.rb
UniqMartin: :+1: Would you mind making the argument name `python` as this is more in line with other methods in this file?

Library/Homebrew/language/python.rb
ilovezfs: I originally had it nicer but `brew style` barked at me :(

xu-cheng: You can try `(requirements | deps).any?`

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,34 @@ def virtualenv_create(venv_root, python = "python", formula = self)
...
+          py3 = needs_python? "python3"
+          if py2 && !py3
+            python = "python"
+          elsif py3 && !py2
+            python = "python3"
+          elsif py2 && py3
+            raise "ambiguous python, please specify"
+          else
+            python = "python"
+          end
```
UniqMartin: I think the following code is more succinct and manages to be equally understandable:

```ruby
wanted = %w[python python3].select { |py| needs_python?(py) }
raise "ambiguous python, please specify" if wanted.size > 1
python = wanted.first || "python"
```

Library/Homebrew/language/python.rb
ilovezfs: looks like it would have to be `(requirements.to_a | deps)` or `(deps.to_a | requirements)` or `(requirements.to_a | deps.to_a)`

ilovezfs: How about
```
      def needs_python?(python)
        return true if build.with?(python)
        (requirements.to_a | deps).any? { |r| r.name == python && r.required? }
      end
```

ilovezfs:
PR refreshed.

Library/Homebrew/language/python.rb
tdsmith: This is also bizarre and unlikely so raising seems good :)

Library/Homebrew/language/python.rb
tdsmith: These are mostly not applications, so they'll never see this code.

tdsmith:
This looks correct to me.

Library/Homebrew/language/python.rb
UniqMartin: I like that a lot more. :+1:

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,22 @@ def virtualenv_create(venv_root, python = "python", formula = self)
...
       # Helper method for the common case of installing a Python application.
       # Creates a virtualenv in `libexec`, installs all `resource`s defined
       # on the formula, and then installs the formula.
-      def virtualenv_install_with_resources
-        venv = virtualenv_create(libexec)
+      def virtualenv_install_with_resources(options = {})
+        python = options[:python]
+        if python.nil?
+          wanted = %w[python python3].select { |py| needs_python?(py) }
+          raise "ambiguous python, please specify" if wanted.size > 1
```
UniqMartin: This exception message is a bit vague. I wonder if you can improve it and make it more actionable, as it seems not entirely unlikely that it will be presented to contributors.

ilovezfs: yeah I intend to make it a "real" exception class, of course :)

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,22 @@ def virtualenv_create(venv_root, python = "python", formula = self)
         venv
       end
 
+      def needs_python?(python)
```
UniqMartin: It would be nice to add documentation to this method (to follow Tim's precedent when he introduced the `virtualenv` code) and as this method is only used internally, either make it private or at least document it as private via `@api private`.

Library/Homebrew/language/python.rb
```diff
@@ -139,11 +139,22 @@ def virtualenv_create(venv_root, python = "python", formula = self)
...
+        return true if build.with?(python)
+        (requirements.to_a | deps).any? { |r| r.name == python && r.required? }
+      end
+
       # Helper method for the common case of installing a Python application.
       # Creates a virtualenv in `libexec`, installs all `resource`s defined
       # on the formula, and then installs the formula.
-      def virtualenv_install_with_resources
-        venv = virtualenv_create(libexec)
+      def virtualenv_install_with_resources(options = {})
```
UniqMartin: The method documentation will need to be updated for the new argument. And I suspect @tdsmith would prefer this to be a plain `python = nil` instead of the option hash, though I can see an argument for the flexibility the latter offers if we decide to add more optional arguments at a later stage.

ilovezfs:
```
virtualenv_install_with_resources :python => "python3"
virtualenv_install_with_resources :python => "python"
```
looks more sane to me than
```
virtualenv_install_with_resources "python3"
virtualenv_install_with_resources "python"
```
or
```
virtualenv_install_with_resources("python3")
virtualenv_install_with_resources("python")
```

```
$ virtualenv --help|grep python
  -p PYTHON_EXE, --python=PYTHON_EXE
                        --python=python2.5 will use the python2.5 interpreter
                        (/usr/local/opt/python/bin/python2.7)
```
So there's a case for calling it `:python_exe` if anyone is concerned about confusion with adding a third `:python` to the code, the other two being
```
depends_on :python
```
and
```
depends_on "numpy" => :python
```
Incidentally, I'm wondering if the `depends_on "numpy" => :python` usage should be entirely deprecated.

UniqMartin: I agree almost universally. I'm not a big fan of `:python_exe` as it smells too much like Windows, but `:python_binary` or something similar would be fine. Maybe to avoid the slightly bogus look of `:python => "python"` at the cost of hard-coding binary names we could make that `:version => 2` or `python_version => 2` or `python_major => 2` (naming is hard!) and leave mapping that to an appropriate Python binary name to the method.

One major downside would be the confusingly different interface of `virtualenv_create` and this method, so it might be prudent to adjust the former if we decide to tweak the latter.

ilovezfs: We could always break out the tried-and-true `:using` :)

UniqMartin: `virtualenv_install_with_resources :using => "python3"` actually reads quite nicely to me!

ilovezfs: @UniqMartin the specter of https://github.com/Homebrew/brew/issues/38 has me worried, but as you know I think that one should be closed.

MikeMcQuaid:
@ilovezfs Gentle ping on this.

MikeMcQuaid:
@ilovezfs What's the state of this; wasn't clear to me.

ilovezfs:
@MikeMcQuaid I'll give it some attention so that it at least has a real exception in there :)

MikeMcQuaid:
@ilovezfs Thanks!

ilovezfs:
@MikeMcQuaid PR refreshed.

MikeMcQuaid:
Seems fine to me but I'd like @tdsmith to take a look when possible.

ilovezfs:
Cool. I squash it down to one commit then.

ilovezfs:
Squashed, and commit message updated.

MikeMcQuaid:
Thanks @ilovezfs!

PR #638, opened 2016-08-05
various: eliminate the usage of `any?`

xu-cheng:
`any?` is not the opposite of `empty?`. Besides the case that
`[false, nil].any?` will return false, `any?`(O(n)) has much worse
performance than `empty?`(O(1)).

cc @UniqMartin

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -628,7 +628,8 @@ class FormulaAuditor
 
     fv = FormulaVersions.new(formula, :max_depth => 10)
     revision_map = fv.revision_map("origin/master")
-    if (revisions = revision_map[formula.version]).any?
+    revisions = revision_map[formula.version]
+    if !revisions.empty?
       problem "revision should not decrease" if formula.revision < revisions.max
     elsif formula.revision != 0
       if formula.stable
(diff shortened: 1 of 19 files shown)


Library/Homebrew/cmd/audit.rb
```diff
@@ -628,7 +628,7 @@ def audit_revision
 
     fv = FormulaVersions.new(formula, :max_depth => 10)
     revision_map = fv.revision_map("origin/master")
-    if (revisions = revision_map[formula.version]).any?
+    if !(revisions = revision_map[formula.version]).empty?
```
UniqMartin: Can you move the assignment before the `if`? Assignments in conditions can be easily misunderstood to be a comparison or missed, particularly when they are hidden in surrounding logic like in this case.

UniqMartin:
:+1: except for a minor nit, but that's unsurprising given I once raised the issue that `any?` might not be a good choice (despite its convenience and brevity) and that `empty?` is to be preferred.

> […], `any?`(O(n)) has much worse performance than `empty?`(O(1)).

In practice that's of course only an issue in pathological cases with long arrays containing mostly `nil`s. Otherwise `any?` will short-circuit the check as soon as it encounters a non-`nil` value.

xu-cheng:
> Otherwise any? will short-circuit the check as soon as it encounters a non-nil value.

😉 just want to add that even with short-circuit the complexity is still the same as O(n).

Library/Homebrew/language/python.rb
```diff
@@ -132,7 +132,7 @@ def virtualenv_create(venv_root, python = "python", formula = self)
           next unless dep_site_packages.exist?
           "import site; site.addsitedir('#{dep_site_packages}')\n"
         end
-        if pth_contents.any?
+        unless pth_contents.empty?
```
tdsmith: This should actually be `any?` because using `next` in `map` may collect `nil` values that we want to disregard.

xu-cheng: I see. I think I will append `.compact` in the expression above, which makes the core more readable.

PR #644, opened 2016-08-05
Fix update commit for non-HEAD kegs with head spec

vladshablinsky:

-----

If we try to call `Formulary.from_keg(f, :head)` on the keg that
is not HEAD-keg itself, we don't need to update commit of
returned formula and should use just HEAD version with nil commit.

Same is true for `ARGV.resolved_formulae`

See https://github.com/Homebrew/brew/pull/500#issuecomment-237886252 and https://github.com/Homebrew/brew/pull/500#issuecomment-237900703

CC @UniqMartin @xu-cheng @MikeMcQuaid @eirinikos

### DIFF
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -34,7 +34,7 @@ module HomebrewArgvExtension
           f.build = tab
           if f.head? && tab.tabfile
             k = Keg.new(tab.tabfile.parent)
-            f.version.update_commit(k.version.version.commit)
+            f.version.update_commit(k.version.version.commit) if k.version.head?
           end
         end
         f
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -246,7 +246,7 @@ class Formulary
       end
     end
     f.build = tab
-    f.version.update_commit(keg.version.version.commit) if f.head?
+    f.version.update_commit(keg.version.version.commit) if f.head? && keg.version.head?
     f
   end
 


UniqMartin:
Thanks for the quick fix! However, I think we've just discovered a more fundamental issue, namely that some code in Homebrew allows the spec to be overridden by a command-line argument like `--HEAD` even if the formula is initialized from an installed keg and the corresponding tab file clearly indicates a different spec for the installed formula. I think we have basically two options to resolve this, both of which diverge from the current behavior:

- Use the spec from the tab file (if it exists) and ignore the spec given via command-line arguments.
- Raise an error if the spec in the tab file (if it exists) and the spec given via command-line arguments don't match.

However, I'm not entirely sure of the full ramifications of this change in behavior, thus any feedback on this (and whether this suggestion even makes sense) is greatly appreciated.

vladshablinsky:
@UniqMartin while I was fixing it was thinking about something similar, not about the solutions, but about the issue itself. But then I thought what `Formulary` was used for. `Formulary` has different methods to generate `Formula` instance, so I thought that `from_keg` is needed to get the formula instance of that keg and spec just sets the spec we want that formula to have. After I realised that it didn't seem that strange to me, because it was better than get formula instance and then set the spec manually. That said, maybe I'm missing something.

xu-cheng:
@UniqMartin actually I think it's prefectly fine to allow overwrite spec in command line. For example, one would anticipate `brew reinstall foo --devel` will just reinstall formulae at new spec with all options preserved.

UniqMartin:
I see. I didn't consider this use case (or other possibly similar use cases). The constellation that revealed the bug addressed in this PR was:

```
brew install foo
brew test foo --devel
brew test foo --HEAD
```

Here, only the stable spec is installed, so it feels weird that the spec can be overridden for `brew test` and, e.g., `brew test --devel` ends up succeeding despite the spec mismatch. Thoughts?

I guess no matter what the “correct” handling of the above scenario is, this PR is probably a sufficient fix for the observed bug and dealing with spec mismatch is a bit out of scope.

UniqMartin:
Thanks again @vladshablinsky for the quick fix! :bow:

PR #648, opened 2016-08-06
audit: placement enforcement tweaks

DomT4:

-----

More style harmonisation.

CC @ilovezfs as you asked for the `include` enforcement.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -219,6 +219,7 @@ class FormulaAuditor
       [/^  keg_only/,                      "keg_only"],
       [/^  option/,                        "option"],
       [/^  depends_on/,                    "depends_on"],
+      [/^  conflicts_with/,                "conflicts_with"],
       [/^  (go_)?resource/,                "resource"],
       [/^  def install/,                   "install method"],
       [/^  def caveats/,                   "caveats method"],
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -203,6 +203,7 @@ class FormulaAuditor
     return unless @strict
 
     component_list = [
+      [/^  include Language::/,            "include method"],
       [/^  desc ["'][\S\ ]+["']/,          "desc"],
       [/^  homepage ["'][\S\ ]+["']/,      "homepage"],
       [/^  url ["'][\S\ ]+["']/,           "url"],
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/audit.rb
```diff
@@ -203,6 +203,7 @@ def audit_file
     return unless @strict
 
     component_list = [
+      [/^  include Language::/,            "include method"],
```
UniqMartin: I'm afraid `include method` is going to be confusing, as `foo method` tends to refer to `def foo` (see below). How do you feel about referring to it as simply `include` or maybe `include directive`?

UniqMartin:
:+1: aside from some terminology nitpicking.

Library/Homebrew/cmd/audit.rb
DomT4: `include directive` sounds 👍 to me. Will change.

ilovezfs:
Shouldn't this also force the blank line following?

UniqMartin:
> ```
> audit: enforce include method placement
> audit: tweak include wording
> ```

Sorry if I'm being obnoxious here, but you are going to squash these two commits to a single commit “audit: enforce include directive placement” before pulling, aren't you?

> Shouldn't this also force the blank line following?

I'm pretty sure we don't have any infrastructure in place to enforce inclusion/omission of blank lines, but I'm sure @DomT4 is happy to make that part of this PR. :trollface:

DomT4:
> Sorry if I'm being obnoxious here, but you are going to squash these two commits to a single commit “audit: enforce include directive placement” before pulling, aren't you?

Thought I already did, to be honest. Apparently not.

> I'm pretty sure we don't have any infrastructure in place to enforce inclusion/omission of blank lines

I can think of no obviously sane way to communicate placement of empty lines to people. We'd be using it already if I could think of a clever way to both implement & communicate it 😅.

UniqMartin:
> I can think of no obviously sane way to communicate placement of empty lines to people.

How about “please add an empty line between include directive (line 2) and desc (line 3)” and “please remove the empty line(s) between desc (line 3) and homepage (line 5)”? Seems entirely doable to me on the communication level. As for the implementation side …

ilovezfs:
I think some apparatus of blank-line virtue would be nice to have but it needn't block this PR despite how much I know @DomT4 is champing at the bit to code it.

PR #649, opened 2016-08-06
Refactor bootstrap and test coverage setup

UniqMartin:

-----

The changes are best viewed commit by commit. The goal is to eliminate some discrepancies between the regular Homebrew bootstrap and the bootstrap performed in the test code for integration command testing. The latter was unnecessarily complex because `brew.rb` still has remnants of the time when it was an executable script with a shebang (and thus needed to initialize itself) which is no longer the case.

Aside from a refactoring, there are two major benefits resulting from this PR:

- `brew.rb` is finally included in the coverage reports. This was previously not possible due to the inflexible bootstrap behavior. (It showed up with zero relevant lines.)
- Coverage initialization and helpers are consolidated in `bootstrap_coverage.rb`, making it much easier to bring much needed performance improvements to `brew tests --coverage`.

cc @eirinikos

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -247,7 +247,11 @@ then
   # Hide shellcheck complaint:
   # shellcheck source=/dev/null
   source "$HOMEBREW_BASH_COMMAND"
-  { update-preinstall; "homebrew-$HOMEBREW_COMMAND" "$@"; exit $?; }
+  {
+    update-preinstall
+    "homebrew-$HOMEBREW_COMMAND" "$@"
+    exit $?
+  }
 else
   # Hide shellcheck complaint:
   # shellcheck source=/dev/null
@@ -256,5 +260,10 @@ else
 
   # Unshift command back into argument list (unless argument list was empty).
   [[ "$HOMEBREW_ARG_COUNT" -gt 0 ]] && set -- "$HOMEBREW_COMMAND" "$@"
-  { update-preinstall; exec "$HOMEBREW_RUBY_PATH" -W0 "$HOMEBREW_LIBRARY/Homebrew/brew.rb" "$@"; }
+  {
+    update-preinstall
+    exec "$HOMEBREW_RUBY_PATH" -W0 \
+      -I "$HOMEBREW_LIBRARY/Homebrew" \
+      -- "$HOMEBREW_LIBRARY/Homebrew/brew.rb" "$@"
+  }
 fi
--- a/Library/Homebrew/config.rb
+++ b/Library/Homebrew/config.rb
@@ -39,4 +39,4 @@ HOMEBREW_TEMP = Pathname.new(ENV.fetch("HOMEBREW_TEMP", "/tmp"))
 HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
 
 # Load path used by standalone scripts to access the Homebrew code base
-HOMEBREW_LOAD_PATH = HOMEBREW_LIBRARY_PATH
+HOMEBREW_LOAD_PATH = HOMEBREW_LIBRARY_PATH.to_s
--- a/Library/Homebrew/config.rb
+++ b/Library/Homebrew/config.rb
@@ -35,10 +35,8 @@ HOMEBREW_LOGS = Pathname.new(ENV["HOMEBREW_LOGS"] || "~/Library/Logs/Homebrew/")
 # Must use /tmp instead of $TMPDIR because long paths break Unix domain sockets
 HOMEBREW_TEMP = Pathname.new(ENV.fetch("HOMEBREW_TEMP", "/tmp"))
 
-unless defined? HOMEBREW_LIBRARY_PATH
-  # Root of the Homebrew code base
-  HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
-end
+# Root of the Homebrew code base
+HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
 
 # Load path used by standalone scripts to access the Homebrew code base
 HOMEBREW_LOAD_PATH = HOMEBREW_LIBRARY_PATH
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -1,8 +1,6 @@
 std_trap = trap("INT") { exit! 130 } # no backtrace thanks
 
 require "pathname"
-HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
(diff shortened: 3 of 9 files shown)


Library/Homebrew/config.rb
```diff
@@ -35,10 +35,8 @@
 # Must use /tmp instead of $TMPDIR because long paths break Unix domain sockets
 HOMEBREW_TEMP = Pathname.new(ENV.fetch("HOMEBREW_TEMP", "/tmp"))
 
-unless defined? HOMEBREW_LIBRARY_PATH
-  # Root of the Homebrew code base
-  HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
-end
+# Root of the Homebrew code base
+HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
```
xu-cheng: May I propose to eliminate `HOMEBREW_LIBRARY_PATH` completely. It's quiet confusing to have it, especially we have another variable called `HOMEBREW_LIBRARY`.

Library/Homebrew/config.rb
```diff
@@ -39,4 +39,4 @@
 HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
 
 # Load path used by standalone scripts to access the Homebrew code base
-HOMEBREW_LOAD_PATH = HOMEBREW_LIBRARY_PATH
+HOMEBREW_LOAD_PATH = HOMEBREW_LIBRARY_PATH.to_s
```
xu-cheng: I think this is bad, as you are altering an immutable value(value whose name are all capitalized). It will cause ruby raising warnings.

xu-cheng:
One suggestion, otherwise :+1:

Library/Homebrew/config.rb
UniqMartin: I'm not sure I understand. Unless I'm missing a piece of the puzzle, I'm simply converting a `Pathname` instance to a string (this should yield a new `String` instance) and assigning it to a new constant, that didn't previously exist. How is that going to cause a warning?

Library/Homebrew/config.rb
UniqMartin: I don't think we can easily eliminate it, but I agree that the name can be confusing (and is known to have caused confusion in the past), thus we should try to find a better name.

We cannot eliminate it as easily because in regular execution `HOMEBREW_LIBRARY_PATH` is simply `HOMEBREW_LIBRARY/"Homebrew"`, thus we might be tempted to replace the former with the latter. However, in the testing code, they point to completely different locations in the file system and there is no obvious relationship between them. Here, `HOMEBREW_LIBRARY_PATH` is still the root of the Homebrew source code, but `HOMEBREW_LIBRARY` points to a temporary location and is not the parent directory of `HOMEBREW_LIBRARY_PATH`.

I'm happy to spend some time on this and to come up with a better name/solution, but I would prefer to do that in a separate follow-up PR. Is that fine by you?

Library/Homebrew/config.rb
xu-cheng: Sorry my bad, I misread the code to `HOMEBREW_LIBRARY_PATH = HOMEBREW_LIBRARY_PATH.to_s`. Please ignore my last comment

Library/Homebrew/brew.sh
```diff
@@ -247,7 +247,11 @@ then
   # Hide shellcheck complaint:
   # shellcheck source=/dev/null
   source "$HOMEBREW_BASH_COMMAND"
-  { update-preinstall; "homebrew-$HOMEBREW_COMMAND" "$@"; exit $?; }
```
MikeMcQuaid: I don't think this quite works the same as the previous version as Bash reads line-by-line and if this file is modified as part of the update process the new version may fail.

MikeMcQuaid:
Nice work, LGTM, :shipit:

MikeMcQuaid:
(after my one nit is addressed)

Library/Homebrew/brew.sh
UniqMartin: You're right, Bash normally reads line-by-line. But that's not what we're relying on here. Instead it's the curly braces (search for “group command” in the Bash man page) that cause the expression to be read as a whole before it gets executed. Whether it extends over several lines or not doesn't matter.

Here are four self-modifying scripts that demonstrate both your point (normally execution line by line) and mine (the different behavior of a group command):

```sh
#!/bin/bash
self="$(<"$0")"
sed -e 's/H[e]llo/Goodbye/' <<<"$self" >"$0"
echo "test-1: Hello World!"
exit 0
```

```sh
#!/bin/bash
self="$(<"$0")"
{
  sed -e 's/H[e]llo/Goodbye/' <<<"$self" >"$0"
  echo "test-2: Hello World!"
  exit 0
}
```

```sh
#!/bin/bash
self="$(<"$0")"
sed -e 's/H[e]llo/Goodbye/' <<<"$self" >"$0"; echo "test-3: Hello World!"; exit 0
```

```sh
#!/bin/bash
self="$(<"$0")"
{ sed -e 's/H[e]llo/Goodbye/' <<<"$self" >"$0"; echo "test-4: Hello World!"; exit 0; }
```

And here's their output:

```console
$ ./test-1
test-1: Goodbye World!
$ ./test-2
test-2: Hello World!
$ ./test-3
test-3: Hello World!
$ ./test-4
test-4: Hello World!
```

UniqMartin: By the way, things will become funny in most cases if the `exit 0` is omitted:

```console
$ ./test-1
test-1: Goodbye World!
$ ./test-2
test-2: Hello World!
./test-2: line 7: syntax error near unexpected token `}'
./test-2: line 7: `}'
$ ./test-3
test-3: Hello World!
./test-3: line 4: unexpected EOF while looking for matching `"'
./test-3: line 5: syntax error: unexpected end of file
$ ./test-4
test-4: Hello World!
./test-4: line 4: syntax error near unexpected token `}'
./test-4: line 4: `}'
```

That's another important thing we rely on: Stop before leaving this group command.

MikeMcQuaid: Gotcha, ignore me then 👍

UniqMartin: No, actually thanks for asking!

1. That's definitely behavior few shell users know about and in many situations they don't need to know about it. But in our case it's important and it's better to be safe.
2. Your comment made me think again about some other code in this PR and I think I identified a problem that has the potential to affect users. I'm investigating this now.

MikeMcQuaid:
@UniqMartin Gentle ping on this.

UniqMartin:
Apologies for the lack of activity here! I'll reopen or resubmit once I find the time.

MikeMcQuaid:
Thanks @UniqMartin!

PR #650, opened 2016-08-06
Update upgrade/outdated documentation

vladshablinsky:

-----

Document new `--fetch-HEAD` flag for `outdated` and `upgrade` commands.

### DIFF
--- a/Library/Homebrew/cmd/outdated.rb
+++ b/Library/Homebrew/cmd/outdated.rb
@@ -1,4 +1,4 @@
-#:  * `outdated` [`--quiet`|`--verbose`|`--json=v1`]:
+#:  * `outdated` [`--quiet`|`--verbose`|`--json=v1`] [`--fetch-HEAD`]:
 #:    Show formulae that have an updated version available.
 #:
 #:    By default, version information is displayed in interactive shells, and
@@ -11,6 +11,10 @@
 #:
 #:    If `--json=`<version> is passed, the output will be in JSON format. The only
 #:    valid version is `v1`.
+#:
+#:    If `--fetch-HEAD` is passed, fetch upstream repository to detect that HEAD
+#:    formula is outdated. Otherwise HEAD-installation is considered outdated if
+#:    new stable or devel version is bumped after that installation.
 
 require "formula"
 require "keg"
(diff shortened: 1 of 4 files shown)


Library/Homebrew/cmd/outdated.rb
```diff
@@ -11,6 +11,10 @@
 #:
 #:    If `--json=`<version> is passed, the output will be in JSON format. The only
 #:    valid version is `v1`.
+#:
+#:    If `--fetch-HEAD` is passed, fetch upstream repository to detect that HEAD
+#:    formula is outdated. Otherwise HEAD-installation is considered outdated if
+#:    new stable or devel version is bumped after that installation.
```
MikeMcQuaid: @vladshablinsky Current text reads a bit weird. How about (assuming it's accurate): 

```
If `--fetch-HEAD` is passed, fetch the upstream repository to detect if the HEAD
installation of the formula is outdated. Otherwise, the repository's HEAD will be
checked for updates when a new stable or devel version has been released.
```

MikeMcQuaid: Also, how long does this operation take on e.g. a standard GitHub repo?

vladshablinsky: > How about (assuming it's accurate)

Looks good.

vladshablinsky: > how long does this operation take on e.g. a standard GitHub repo?

Will test, but shouldn't be too time-consuming for GitHubGitDownloadStrategy. However, for GitDownloadStrategy we need to fetch the repo, so it takes some time. How do I understand whether results are OK or not OK when I do some benchmarks?

MikeMcQuaid: Just post them here 😀. I think we could do it automatically for GitHub, then?

xu-cheng: > I think we could do it automatically for GitHub, then?

I would avoid it otherwise we may get formula tagged as outdated for a popular repo too often whenever we run `brew upgrade/outdated`. We need a flag to control the behaviors, whether you want prompt update or not.

Besides the time is still relative consuming:
```
$ time brew outdated
brew outdated  0.59s user 0.14s system 79% cpu 0.921 total
$ time brew outdated --fetch-HEAD
neovim/neovim/neovim (HEAD-4dc4efc) < HEAD-abb0928dfbf71d5358b56e888db9ed8ff66d8933
brew outdated --fetch-HEAD  0.54s user 0.10s system 13% cpu 4.918 total
```

PR #654, opened 2016-08-07
Fix style inconsistencies in Bash code

UniqMartin:

-----

We have accumulated quite a few style inconsistencies in our Bash code. I fixed all of them following the rules below that I inferred them from what I believe to be our preferred and/or predominant style. (Of course all of those rules are up for debate.)

- **Whitespace**
  - Indent by two spaces.
  - Align single-line items in `case` statements.
  - Put one space between single-line items in `case` and terminating `;;`.
  - Indent items in `case` statements one step (unlike in Ruby code).
  - Use multi-line `if`/`elsif`-`then` and `for`/`while`-`do` statements.
  - Omit space between function name and `()`.
  - Put function “signature” and opening curly brace on same line, e.g. `foo() {`.
  - Put a newline between shebang and shell code.
  - Omit space after redirection operators (`>`, `>>`, `<`), e.g. `>/dev/null`.
- **Redundant Stuff**
  - Omit unneeded trailing `;`.
  - Omit unneeded `-` in `<<-EOS` (we don't tab-indent our here documents).
  - Omit unneeded line continuation (`\`) in arrays.
- **Miscellaneous**
  - Prefer double-quoted strings over single-quoted strings if equally clear. (Continue to use single-quoted string if double-quoted string requires heinous escaping.)
  - Prefer `[[` over `[` and `test`.
  - Prefer `=` over `==` in `[[`.
  - Always quote variable usage and shell substitution unless word-splitting behavior is intended. (For consistency, do this also in assignments where special rules apply.)

### DIFF
--- a/Library/Homebrew/shims/super/ant
+++ b/Library/Homebrew/shims/super/ant
@@ -1,5 +1,7 @@
 #!/bin/bash
+
 export HOMEBREW_CCCFG="O$HOMEBREW_CCCFG"
+
 ant=/usr/bin/ant
-[ -x "$ant" ] || ant="$(${HOMEBREW_BREW_FILE} --prefix ant)/bin/ant"
+[[ -x "$ant" ]] || ant="$("$HOMEBREW_BREW_FILE" --prefix ant)/bin/ant"
 exec "$ant" "$@"
--- a/Library/Homebrew/shims/super/mig
+++ b/Library/Homebrew/shims/super/mig
@@ -1,3 +1,4 @@
 #!/bin/bash
+
 pwd="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-exec xcrun mig -cc $pwd/cc "$@"
+exec xcrun mig -cc "$pwd/cc" "$@"
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -33,7 +33,7 @@ git() {
 }
 
 # Force UTF-8 to avoid encoding issues for users with broken locale settings.
-if [[ "$(locale charmap 2> /dev/null)" != "UTF-8" ]]
+if [[ "$(locale charmap 2>/dev/null)" != "UTF-8" ]]
 then
   export LC_ALL="en_US.UTF-8"
 fi
@@ -48,8 +48,8 @@ else
 fi
 
 case "$*" in
-  --prefix) echo "$HOMEBREW_PREFIX"; exit 0 ;;
-  --cellar) echo "$HOMEBREW_CELLAR"; exit 0 ;;
+  --prefix)            echo "$HOMEBREW_PREFIX"; exit 0 ;;
+  --cellar)            echo "$HOMEBREW_CELLAR"; exit 0 ;;
   --repository|--repo) echo "$HOMEBREW_REPOSITORY"; exit 0 ;;
 esac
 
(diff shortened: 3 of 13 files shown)


xu-cheng:
👍

Library/Homebrew/shims/super/ant
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
+
 export HOMEBREW_CCCFG="O$HOMEBREW_CCCFG"
+
 ant=/usr/bin/ant
-[ -x "$ant" ] || ant="$(${HOMEBREW_BREW_FILE} --prefix ant)/bin/ant"
+[[ -x "$ant" ]] || ant="$("$HOMEBREW_BREW_FILE" --prefix ant)/bin/ant"
```
xu-cheng: Maybe we should use `HOMEBREW_PREFIX` directly.

Library/Homebrew/shims/super/mig
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
+
 pwd="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
xu-cheng: It seems that there are multiple styles for resolving the current directory. I'm wondering should we make them consistence as well?

Library/Homebrew/shims/super/ant
UniqMartin: True. Note that I avoided all changes to functionality, however subtle, and focused on the style issues only, so not addressing this was intentional.

Library/Homebrew/shims/super/mig
UniqMartin: Like in the other comment, I tend to agree. (But as before, not changing functionality and focusing on style changes only was intentional.)

MikeMcQuaid:
👍 but if this fixes all syntax failures in our bash code: please consider adding a check for this to `brew test-bot` or `brew tests` or similar so they don't just break again.

UniqMartin:
> but if this fixes all syntax failures in our bash code: please consider adding a check for this to `brew test-bot` or `brew tests` or similar so they don't just break again.

No, all the changes in this PR are purely about style. If there were any syntax errors, we would have been in more trouble (unless it's in a rarely used `superenv` shim).

However, your comment made me remember `test/test_bash.rb` and I pushed another commit to make sure it checks **all** our Bash code for syntax errors, including `superenv` shims and newly added or relocated code. (I think this test was added when `bin/brew` was still a huge monolithic piece of code and neither `brew.sh` nor `utils{,/*}.sh` existed yet.) Thanks for this! ❤️

MikeMcQuaid:
Sorry, to clarify: I mean if you've fixed all style warnings it'd be great to push an automated `shellcheck` check somewhere. Thanks for that 👍

UniqMartin:
I see. No, we're not quite there yet and most stuff I changed in this PR isn't the kind that would be pointed out by `shellcheck`, though I made sure not to introduce any new warnings. I'll make sure to put this on my TO DO list, but IMO that's definitely something for a follow-up PR.

MikeMcQuaid:
I'd love to see Homebrew/brew "clean" enough at some point that we can have Shellcheck and RuboCop run by CI to check for regressions. I guess we could do some of that automatically if we have CI use a different set of rules and try to fix one set at a time?

UniqMartin:
I fully agree and the idea how to do this is shaping up nicely in my head, thanks also to your input! The only thing that's holding me back from spending more time with this is that I'm currently trying to focus more on GSoC/Outreachy-related stuff and more pressing issues (and yes, this PR doesn't exactly prove my point). All those style-related tasks are definitely quite high on my agenda, if nobody beats me to it, and it would be nice to have some of that before 1.0.

MikeMcQuaid:
@UniqMartin same here.

PR #656, opened 2016-08-07
vendor ruby-macho 0.2.5

woodruffw:

-----

This brings `ruby-macho` 0.2.5 (with a slew of new changes) into `brew`.

Of note:
* The bugs exposed by Homebrew's virtualenv bottling have been fixed (https://github.com/Homebrew/ruby-macho/pull/55, https://github.com/Homebrew/ruby-macho/pull/54)
* `ruby-macho`'s public API is now entirely documented (https://github.com/Homebrew/ruby-macho/pull/46)
* Full support for RPath addition, deletion, and modification is implemented (https://github.com/Homebrew/ruby-macho/pull/35).
* Load commands can now be synthesized and serialized dynamically, giving us a lot more flexibility with respect to future additions to the Mach-O spec (https://github.com/Homebrew/ruby-macho/pull/38).

This PR also updates the bridge code between `ruby-macho` and `brew`.

cc @UniqMartin

### DIFF
--- a/Library/Homebrew/os/mac/ruby_keg.rb
+++ b/Library/Homebrew/os/mac/ruby_keg.rb
@@ -4,7 +4,7 @@ module RubyKeg
   def change_dylib_id(id, file)
     @require_install_name_tool = true
     puts "Changing dylib ID of #{file}\n  from #{file.dylib_id}\n    to #{id}" if ARGV.debug?
-    MachO::Tools.change_dylib_id(file, id)
+    MachO::Tools.change_dylib_id(file, id, :strict => false)
   rescue MachO::MachOError
     onoe <<-EOS.undent
       Failed changing dylib ID of #{file}
@@ -17,7 +17,7 @@ module RubyKeg
   def change_install_name(old, new, file)
     @require_install_name_tool = true
     puts "Changing install name in #{file}\n  from #{old}\n    to #{new}" if ARGV.debug?
-    MachO::Tools.change_install_name(file, old, new)
+    MachO::Tools.change_install_name(file, old, new, :strict => false)
   rescue MachO::MachOError
     onoe <<-EOS.undent
       Failed changing install name in #{file}
--- a/Library/Homebrew/os/mac/ruby_mach.rb
+++ b/Library/Homebrew/os/mac/ruby_mach.rb
@@ -14,7 +14,7 @@ module RubyMachO
       machos = []
       mach_data = []
 
-      if MachO.fat_magic?(macho.magic)
+      if MachO::Utils.fat_magic?(macho.magic)
         machos = macho.machos
       else
         machos << macho
@@ -28,9 +28,8 @@ module RubyMachO
         end
 
         type = case m.filetype
-        when "MH_EXECUTE" then :executable
-        when "MH_DYLIB" then :dylib
-        when "MH_BUNDLE" then :bundle
+        when :dylib, :bundle then m.filetype
+        when :execute then :executable
         else :dunno
         end
 
--- a/Library/Homebrew/vendor/README.md
+++ b/Library/Homebrew/vendor/README.md
@@ -3,7 +3,7 @@ Vendored Dependencies
 
 * [okjson](https://github.com/kr/okjson), version 43.
 
-* [ruby-macho](https://github.com/Homebrew/ruby-macho), version 0.2.4
+* [ruby-macho](https://github.com/Homebrew/ruby-macho), version 0.2.5
 
 ## Licenses:
 
--- a/Library/Homebrew/vendor/macho/macho.rb
+++ b/Library/Homebrew/vendor/macho/macho.rb
@@ -1,4 +1,5 @@
 require "#{File.dirname(__FILE__)}/macho/structure"
+require "#{File.dirname(__FILE__)}/macho/view"
 require "#{File.dirname(__FILE__)}/macho/headers"
 require "#{File.dirname(__FILE__)}/macho/load_commands"
 require "#{File.dirname(__FILE__)}/macho/sections"
@@ -12,5 +13,5 @@ require "#{File.dirname(__FILE__)}/macho/tools"
 # The primary namespace for ruby-macho.
 module MachO
   # release version
-  VERSION = "0.2.4".freeze
+  VERSION = "0.2.5".freeze
 end
--- a/Library/Homebrew/vendor/macho/macho/exceptions.rb
+++ b/Library/Homebrew/vendor/macho/macho/exceptions.rb
@@ -3,6 +3,26 @@ module MachO
   class MachOError < RuntimeError
   end
 
+  # Raised when a Mach-O file modification fails.
+  class ModificationError < MachOError
+  end
+
+  # Raised when a Mach-O file modification fails but can be recovered when
+  # operating on multiple Mach-O slices of a fat binary in non-strict mode.
+  class RecoverableModificationError < ModificationError
+    # @return [Fixnum, nil] The index of the Mach-O slice of a fat binary for
+    #   which modification failed or `nil` if not a fat binary. This is used to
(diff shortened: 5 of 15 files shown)


UniqMartin:
Like last time, I've temporarily tagged the PR with “do not merge” as I want to test drive it a bit more thoroughly on my local machine before merging. (You can expect some more feedback from me tomorrow or the day after. I'm signing off for today.)

MikeMcQuaid:
Nice work! Any thoughts on when this will be ready to replace the current implementation? It's been pretty rock solid so far.

woodruffw:
I think we're getting close to making this the user-facing implementation. In particular, this version contains a completely new implementation of load command modification, one that's a **lot** nicer (and more correct/generic) than the one that I originally wrote.

We're probably going to find small bugs here and there, but I'm much more comfortable exposing this to users than I would have been a few months ago :smile:

UniqMartin:
> Any thoughts on when this will be ready to replace the current implementation?

I'd like it to see some more testing in some corner cases, but that's probably just me being overly cautious. Overall I'm pretty pleased with how this has shaped up.

At this point, the main thing that stops us from delivering this to all users is a hard dependency on Ruby 2.0+. We've worked with this assumption as this simplified some of our work, giving us more time to focus on things that really matter. If we wanted to enable this for everyone, we would need to add a check for a sufficiently new Ruby and otherwise fall back to the `cctools` implementation. The other option would be to ship vendored Ruby 2.0 (if it's ready) and stop worrying about Ruby versions that are more than obsolete nowadays …

woodruffw:
> The other option would be to ship vendored Ruby 2.0 (if it's ready) and stop worrying about Ruby versions that are more than obsolete nowadays …

I am strongly in favor of this option, but falling back to `cctools` seems like reasonable edge behavior for older systems if the vendored ruby 2.x won't be ready in the foreseeable future.

It's also worth noting that users on older systems may be susceptible to the bad `install_name_tool` [behavior we (re)discovered](https://github.com/Homebrew/ruby-macho/issues/41#issuecomment-235475634). Whether or not they get `ruby-macho` immediately, it definitely makes sense to get them away from their system copies of `cctools`.

UniqMartin:
Just a minor thing, but the [`Library/Homebrew/vendor/README.md` file](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/vendor/README.md) also needs an update.

MikeMcQuaid:
Agreed with shipping this to users and waiting on Ruby 2.0. We can probably ship it by default to `HOMEBREW_DEVELOPER`-setting folks first, though.

woodruffw:
> Just a minor thing, but the Library/Homebrew/vendor/README.md file also needs an update.

I knew I forgot something...

UniqMartin:
*Quick status update:* Things are looking pretty good in local testing and the currently running test should finish in 3-4 hours. The only bugs I found so far (two of them) are in the now almost “legacy” `cctools`-based code Mach-O handling. :joy: (Though to be honest, both are relatively minor things and are unlikely to impact users.)

I'm increasingly confident about rolling this out to all users. :tada: Expect a few PRs very soon that will address the discovered bugs and will add a new mode of testing this more thoroughly on the bot that I briefly discussed with Mike (sorry for the tease, more details to follow soon).

UniqMartin:
All my local testing finished successfully, thus I'm inclined to merge this as soon as possible.

However, the stricter behavior we've been previously enforcing (by accident) and that we're now about to loosen to the standards of `install_name_tool` et al. is helping to uncover real problems, like in Homebrew/homebrew-core#3704 for a very recent example (somehow all of them are related to Python 🙈), thus I'm wondering whether we should actually stick with the stricter behavior. An upside is that we bottle less broken stuff, a downside that this is creating (very infrequently) more friction for our contributors. Thoughts?

Also, because I've improved the error reporting to simplify debugging cases like the above in the future, there's a merge conflict that needs to be addressed (but should be trivial).

woodruffw:
> thus I'm wondering whether we should actually stick with the stricter behavior

This is fine by me; now that we actually have the ability to switch between the two programatically, reverting to the more lenient behavior is trivial if the maintenance load becomes untenable. That being said, I'm also 100% fine with having `:strict => false` in this PR - it's your call :smile: 

> An upside is that we bottle less broken stuff, a downside that this is creating (very infrequently) more friction for our contributors.

Shipping bottles that are more correct/less likely to explode is a big :+1: for me. I don't like creating extra trouble for contributors, but it definitely makes sense to encourage the double-checking of build/linkage flags that make build products (needlessly) inconsistent.

> there's a merge conflict that needs to be addressed (but should be trivial).

Got it, I'll resolve it when I'm home.

UniqMartin:
Thanks, @woodruffw! :tada: Let's see how that fares (I'm pretty confident it will do quite well) and unless something comes up over the weekend consider switching `ruby-macho` on for everyone by default (unless they're the minority not using Ruby 2.0 yet). How does that sound?

dunn:
Is there a plan to turn on strict mode later?

UniqMartin:
> Is there a plan to turn on strict mode later?

I don't think we have a firm idea about whether we want to do that or not. Currently, the stricter behavior was blocking a PR from progressing (Homebrew/homebrew-core#3704), so we might need to figure out what to do with such cases to unblock them, as they previously finished just fine and in this sense it would be a bit of a regression.

Any ideas or opinions surrounding this are more than welcome! I definitely would like to revisit this at a later time (but in the not too distant future).

woodruffw:
Do we know of any formula/builds that *intentionally* include inconsistent fat binaries?

If the chance of an inconsistent binary being intentional turns out to be vanishingly small, we should probably retain default to strict mode. Then, when intentional inconsistencies do occur, we could have formula annotate their inconsistency so that `ruby-macho` knows to disable its strictness assumptions. Of course, feel free to ignore that (spontaneous) idea if changing the formula DSL is undesirable.

MikeMcQuaid:
> Do we know of any formula/builds that intentionally include inconsistent fat binaries?

I don't but @DomT4 might.



> If the chance of an inconsistent binary being intentional turns out to be vanishingly small, we should probably retain default to strict mode.

I'd probably prefer to default to the functionality that matches the previous behaviour for end-users. We could adjust it for the bots.

woodruffw:
> I'd probably prefer to default to the functionality that matches the previous behaviour for end-users.

Sounds good to me! Adjusting it for the bots sounds good, although we'll need a way to exempt formulae that are intentionally inconsistent. I'll look into that :smile:

UniqMartin:
> Do we know of any formula/builds that *intentionally* include inconsistent fat binaries?

Given our experience with `ruby-macho` on the bot during the past few weeks/months, all of the cases we've come across so far where fat binaries have been inconsistent has been some build system failure or some other failure.

> If the chance of an inconsistent binary being intentional turns out to be vanishingly small, we should probably retain default to strict mode.

I somewhat agree, but I think this is impractical unless we have a plan for how to deal with the diverging cases (allow opt-out, be able to fix issues if the diverging case turns out to be actually broken). Given that there's no strict requirement for fat binary to be consistent with respect to their linkage and RPATHs, inconsistent fat binaries are perfectly valid, and some (though very few) system fat binaries on macOS are inconsistent, I think we best stick to the previous behavior (compatible with `install_name_tool`) as far as Homebrew is concerned.

MikeMcQuaid:
> I somewhat agree, but I think this is impractical unless we have a plan for how to deal with the diverging cases (allow opt-out, be able to fix issues if the diverging case turns out to be actually broken). Given that there's no strict requirement for fat binary to be consistent with respect to their linkage and RPATHs, inconsistent fat binaries are perfectly valid, and some (though very few) system fat binaries on macOS are inconsistent, I think we best stick to the previous behavior (compatible with install_name_tool) as far as Homebrew is concerned.

Agreed.

PR #660, opened 2016-08-08
Fix style inconsistencies in Bash completion

UniqMartin:
- [ ] Have you written new tests for your changes? 👉 **N/A**

-----

By now we have a pretty clear style for our Bash/shell code, but our Bash completions, that predate all this code, massively diverge from this style, to a point where it's just not fun to edit that file. This starts with the four-space indent, but includes a lot more than that. This is a follow-up to #654 and makes sure to obey all the rules that have been mentioned there.

Like #654 this focuses on style changes and tries to keep functional changes (even seemingly trivial ones) to an absolute minimum. Consequently, a few `shellcheck` warnings remain. They be addressed in a follow-up PR.

*Note:* I made multiple commits, so interested reviewers can pick out some of the finer changes that would be otherwise hidden by the indentation change. I intend to squash them on merge.

### DIFF
--- a/etc/bash_completion.d/brew
+++ b/etc/bash_completion.d/brew
@@ -15,7 +15,7 @@ __brewcomp_words_include() {
 __brewcomp_prev() {
   local idx=$((COMP_CWORD - 1))
   local prv="${COMP_WORDS[idx]}"
-  while [[ $prv == -* ]]; do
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍 as long as it's working for you.

PR #675, opened 2016-08-10
gpg: split out into unique requirement

DomT4:

-----

++++
++++

GPG 1.x has stopped receiving new features, some of which we may well want to take advantage of sooner or later in Homebrew. Upstream has also been attempting to work out for a while how well used it still is which suggests it may "go away" at some point in the future.

Debian is also in the process of migrating GnuPG 1.x to a `gpg1` executable whilst GnuPG 2.1.x assumes the `gpg` executable. There's a detailed video discussion of this from DebConf 2015 [here](http://meetings-archive.debian.net/pub/debian-meetings/2015/debconf15/GnuPG_in_Debian_report.webm).

It's unsafe to assume every `gpg` executable is going to forever equal 1.x and every `gpg2` executable is going to forever equal 2.x. MacGPG2 has been symlinking 2.x as a vanilla `gpg` for a while, for example, and we will be soon as well.

You'll still be able to plonk the `libexec/bin` path of `gpg` in your PATH to access a vanilla `gpg` 1.x executable if you want to, but we're not going to actively keep adding gpg1 support to formulae going forwards. There's really no reason why 99.9% of projects should not or cannot use `gpg2` these days.

This uses detection methods to determine regardless of what the executable is called we're always hitting a 2.0 GnuPG or nothing.

### DIFF
--- a/Library/Homebrew/compat/requirements.rb
+++ b/Library/Homebrew/compat/requirements.rb
@@ -3,7 +3,8 @@ require "requirements"
 XcodeDependency            = XcodeRequirement
 MysqlDependency            = MysqlRequirement
 PostgresqlDependency       = PostgresqlRequirement
-GPGDependency              = GPGRequirement
+GPGDependency              = GPG2Requirement
+GPGRequirement             = GPG2Requirement
 TeXDependency              = TeXRequirement
 MercurialDependency        = MercurialRequirement
 GitDependency              = GitRequirement
--- a/Library/Homebrew/requirements/gpg_requirement.rb
+++ b/Library/Homebrew/requirements/gpg_requirement.rb
@@ -1,0 +1,27 @@
+require "requirement"
+
+class GPGRequirement < Requirement
+  fatal true
+  default_formula "gnupg2"
+
+  satisfy(:build_env => false) { gpg2 || gpg }
+
+  # MacGPG2/GPGTools installs GnuPG 2.0.x as a vanilla `gpg` symlink
+  # pointing to `gpg2`, as do we. Ensure we're actually using a 2.0 `gpg`.
+  # Temporarily, only support 2.0.x rather than the 2.1.x "modern" series.
+  def gpg
+    which_all("gpg").detect do |gpg|
+      gpg_short_version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
+      next unless gpg_short_version
+      Version.create(gpg_short_version.to_s) == Version.create("2.0")
+    end
+  end
+
+  def gpg2
+    which_all("gpg2").detect do |gpg2|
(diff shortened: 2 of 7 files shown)


MikeMcQuaid:
LGTM 👍 🚢

Library/Homebrew/requirements/gpg_requirement.rb
```diff
@@ -0,0 +1,27 @@
...
+  # Temporarily, only support 2.0.x rather than the 2.1.x "modern" series.
+  def gpg
+    which_all("gpg").detect do |gpg|
+      gpg_short_version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
+      next unless gpg_short_version
+      Version.create(gpg_short_version.to_s) == Version.create("2.0")
+    end
+  end
+
+  def gpg2
```
MikeMcQuaid: I'm a bit confused about these methods; they are both looking for `gpg 2.0` but just looking with different names? If so, would be good to make the requirement named `GPG2Requirement` and share code between the two methods (as their only difference as far as I can see is the `which_all` parameter).

DomT4: Yeah, I ended up dropping 1.x support because 1) Upstream is trying to discourage people from using it unless they actively need something extremely portable, 2) Very few things should ever specifically need 1.x, 3) We're renaming the executables in the top-level prefix for `gpg` & I didn't want to start plonking `PATH` env scripting into every formula.

No issue with calling the requirement `GPG2Requirement`. Will do that. IRT the sharing code, 👍 depending on what we do with the discussion in the other PR around this logic needing replicating in two different places currently.

MikeMcQuaid: I think these two methods could share a bunch of logic for now. Once the other PR is merged I think it makes sense for more of this code to live there rather than have non-requirement code calling a requirement.

DomT4:
Now depends on https://github.com/Homebrew/brew/pull/676 to test against properly, but works locally. Rewrote the requirement to use the class logic & added a small requirement test to ensure it finds accurately.

Library/Homebrew/compat/requirements.rb
```diff
@@ -3,7 +3,7 @@
 XcodeDependency            = XcodeRequirement
 MysqlDependency            = MysqlRequirement
 PostgresqlDependency       = PostgresqlRequirement
-GPGDependency              = GPGRequirement
+GPGDependency              = GPG2Requirement
```
MikeMcQuaid: May want to add `GPGRequirement = GPG2Requirement` too.

MikeMcQuaid:
One nit then 👍

DomT4:
Pushed the requested change, and tweaked the `test_gpg` test from the prior PR to fix a pesky `rmtree` warning.

PR #676, opened 2016-08-10
gpg: add initial bare-bones wrapper

DomT4:

-----

++++
++++

This is the first diddy steps of the wrapper, not the whole thing, obviously.

I don't want to dump the whole functionality from other branches yet because A) as mentioned above, merging unique branches together has revealed some not great problems with some of the existing logic not included here, B) this is the absolute minimum required for the core PR to be in a good place to ship, and people are getting anxious with nothing having been shipped yet, including me.

### DIFF
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -20,7 +20,7 @@ class Gpg
   GPG_EXECUTABLE = gpg2 || gpg
 
   def self.available?
-    File.exist?(GPG_EXECUTABLE.to_s) && File.executable?(GPG_EXECUTABLE.to_s)
+    File.executable?(GPG_EXECUTABLE.to_s)
   end
 
   def self.create_test_key(path)
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -1,32 +1,26 @@
 require "utils"
 
 class Gpg
-  # Should ideally be using `GPGRequirement.new.gpg2`, etc to get path here but
-  # calling that directly leads to:
-  # requirement.rb:139:in `which_all': uninitialized constant Requirement::ORIGINAL_PATHS (NameError)
-  # when i.e. including the gpg syntax in wget. Not problematic if not used by formula code.
-  # For now, the path determination blob of code has been semi-modified for here.
-  # Look into this more.
-  def self.gpg
-    which("gpg") do |gpg|
+  def self.find_gpg(executable)
+    which_all(executable).detect do |gpg|
       gpg_short_version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
       next unless gpg_short_version
       Version.create(gpg_short_version.to_s) == Version.create("2.0")
     end
   end
 
+  def self.gpg
+    find_gpg("gpg")
+  end
+
   def self.gpg2
-    which("gpg2") do |gpg2|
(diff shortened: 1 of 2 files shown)


Library/Homebrew/gpg.rb
```diff
@@ -0,0 +1,47 @@
+require "utils"
+
+class Gpg
+  # Should ideally be using `GPGRequirement.new.gpg2`, etc to get path here but
+  # calling that directly leads to:
+  # requirement.rb:139:in `which_all': uninitialized constant Requirement::ORIGINAL_PATHS (NameError)
```
MikeMcQuaid: You could/should probably do the reverse and have the `GPGRequirement` use the code from here. If that still doesn't work, shout and I'll try to get it working.

Library/Homebrew/gpg.rb
```diff
@@ -0,0 +1,47 @@
...
+      gpg2_short_version = Utils.popen_read(gpg2, "--version")[/\d\.\d/, 0]
+      next unless gpg2_short_version
+      Version.create(gpg2_short_version.to_s) == Version.create("2.0")
+    end
+  end
+
+  GPG_EXECUTABLE = gpg2 || gpg
+
+  def self.available?
+    File.exist?(GPG_EXECUTABLE.to_s) && File.executable?(GPG_EXECUTABLE)
```
MikeMcQuaid: I think `File.executable?` may check existence without barfing. May also want to check it's not a directory.

MikeMcQuaid:
A few comments but generally 👍

Library/Homebrew/gpg.rb
DomT4: I thought about that. Wasn't sure if it was acceptable or not, but if it is, sounds reasonable.

I had some issues around `which_all` always returning a `Pathname` even if you forced `to_s` on it but I think I've just worked out why that wasn't working. Will double check with a fuller test locally.

Library/Homebrew/gpg.rb
DomT4: `which_all` (which I've switched back to locally) already does [some checks](https://github.com/Homebrew/brew/blob/604f4b6a0785e0295e548e6d033823303da19061/Library/Homebrew/utils.rb#L430) the file is executable & a file. This is mostly a smoke test so we actually have something for `available?`.

DomT4:
Pushed changes. Obviously it just gets skipped on CI, but the test does pass:

```
GpgTest#test_create_test_key = 0.31 s = .

Finished in 81.197964s, 11.3919 runs/s, 30.0008 assertions/s.

  1) Skipped:
IntegrationCommandTests#test_bundle [/usr/local/Library/Homebrew/test/test_integration_cmds.rb:42]:
HOMEBREW_TEST_OFFICIAL_CMD_TAPS is not set


  2) Skipped:
IntegrationCommandTests#test_services [/usr/local/Library/Homebrew/test/test_integration_cmds.rb:42]:
HOMEBREW_TEST_OFFICIAL_CMD_TAPS is not set


  3) Skipped:
IntegrationCommandTests#test_pull [/usr/local/Library/Homebrew/test/test_integration_cmds.rb:765]:
Requires network connection


  4) Skipped:
IntegrationCommandTests#test_cask [/usr/local/Library/Homebrew/test/test_integration_cmds.rb:42]:
HOMEBREW_TEST_OFFICIAL_CMD_TAPS is not set

925 runs, 2436 assertions, 0 failures, 0 errors, 4 skips
```

MikeMcQuaid:
:shipit: 👍

PR #677, opened 2016-08-10
install_dependencies: Do not truncate dependencies

sjackman:

-----

See Homebrew/legacy-homebrew#48449

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -406,8 +406,9 @@ class FormulaInstaller
   def install_dependencies(deps)
     if deps.empty? && only_deps?
       puts "All dependencies for #{formula.full_name} are satisfied."
-    else
-      oh1 "Installing dependencies for #{formula.full_name}: #{Tty.green}#{deps.map(&:first)*", "}#{Tty.reset}" unless deps.empty?
+    elsif !deps.empty?
+      oh1 "Installing dependencies for #{formula.full_name}: #{Tty.green}#{deps.map(&:first)*", "}#{Tty.reset}",
+        :truncate => false
       deps.each { |dep, options| install_dependency(dep, options) }
     end
 
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -88,8 +88,10 @@ def ohai(title, *sput)
   puts sput
 end
 
-def oh1(title)
-  title = Tty.truncate(title) if $stdout.tty? && !ARGV.verbose?
+def oh1(title, options = {})
+  if $stdout.tty? && !ARGV.verbose? && options.fetch(:truncate, :auto) == :auto
+    title = Tty.truncate(title)
+  end
   puts "#{Tty.green}==>#{Tty.white} #{title}#{Tty.reset}"
 end
 


sjackman:
See @UniqMartin's comment for the reasoning behind this PR: https://github.com/Homebrew/legacy-homebrew/issues/48449#issuecomment-178638401

Library/Homebrew/formula_installer.rb
```diff
@@ -407,7 +407,8 @@ def install_dependencies(deps)
     if deps.empty? && only_deps?
       puts "All dependencies for #{formula.full_name} are satisfied."
     else
-      oh1 "Installing dependencies for #{formula.full_name}: #{Tty.green}#{deps.map(&:first)*", "}#{Tty.reset}" unless deps.empty?
+      oh1 "Installing dependencies for #{formula.full_name}"
```
MikeMcQuaid: Let's not use a newline every time unless needed. Instead, mirror the styling but without the truncation.

UniqMartin: Would it make sense to address this instead by allowing `oh1` to accept options and to optionally disable the truncation, i.e. in cases like these where we want the styling but not the truncation even in non-verbose mode? E.g. something like this:

```ruby
def oh1(title, options = {})
  if $stdout.tty? && !ARGV.verbose? && options.fetch(:allow_truncation, true)
    title = Tty.truncate(title)
  end
  puts "#{Tty.green}==>#{Tty.white} #{title}#{Tty.reset}"
end

oh1 "I shall not be truncated", :allow_truncation => false
```

MikeMcQuaid: I was thinking the same 👍

sjackman: @UniqMartin Would you like to submit a PR with the above code, or shall I add it to this PR?

UniqMartin: @sjackman I'm happy for you to pick it up and make it a part of this PR.

sjackman:
I've updated this PR to add a `:truncate` option to `oh1`.

Library/Homebrew/formula_installer.rb
```diff
@@ -407,7 +407,8 @@ def install_dependencies(deps)
     if deps.empty? && only_deps?
       puts "All dependencies for #{formula.full_name} are satisfied."
     else
-      oh1 "Installing dependencies for #{formula.full_name}: #{Tty.green}#{deps.map(&:first)*", "}#{Tty.reset}" unless deps.empty?
+      oh1 "Installing dependencies for #{formula.full_name}: #{Tty.green}#{deps.map(&:first)*", "}#{Tty.reset}",
+        :truncate => false unless deps.empty?
```
MikeMcQuaid: Could you use the `unless` block form? Makes it a bit easier to follow this. Thanks!

MikeMcQuaid:
One nit otherwise 👍

sjackman:
Fixed the style.

Library/Homebrew/utils.rb
```diff
@@ -88,8 +88,10 @@ def ohai(title, *sput)
   puts sput
 end
 
-def oh1(title)
-  title = Tty.truncate(title) if $stdout.tty? && !ARGV.verbose?
+def oh1(title, options = {})
+  if $stdout.tty? && !ARGV.verbose? && options.fetch(:truncate, true)
+    title = Tty.truncate(title)
+  end
```
UniqMartin: One reason why I suggested to name the option `:allow_truncation` (though that's a bit unwieldy) is that it's less confusing than a plain `:truncate` with a Boolean value. Example:

```ruby
oh1 "This should always be truncated", :truncate => true # default
oh1 "This should never be truncated", :truncate => false
```

I would expect the first line to be always truncated, but instead it means “may truncate”, depending on environmental factors like output to a TTY or the presence of the `--verbose` flag. One idea I have to resolve this is to do the following (though I'm happy to hear about other/better ideas):

```ruby
oh1 "This will sometimes be truncated", :truncate => :auto # default
oh1 "This should never be truncated", :truncate => :never
```

MikeMcQuaid: @UniqMartin Agreed 👍

sjackman: Fixed.

MikeMcQuaid:
Thanks @sjackman!

sjackman:
Thanks for everyone's feedback, and for merging, Mike. I think this patch is a big improvement to usability when the number of packages that will be installed is very large. It gives the user warning about what's coming up, and that they can probably go for :coffee:.

PR #682, opened 2016-08-11
tests: refactor FormulaTextTests in test_cmd_audit, add assertion to test_simple_valid_formula

eirinikos:

-----

In the process of going through `test_cmd_audit.rb` (with the intention of adding some unit tests), I figured I'd try to refactor `formula_text`. I also added an assertion to `test_simple_valid_formula` (wasn't sure if it belonged in the same commit as the refactoring).

Do the assertions' `msg`s need to be corrected? I left them mostly as-is. But even though they're not of critical importance, I was under the impression that they should be worded to address a situation where the assertion fails, e.g.:

`assert ft.has_trailing_newline?, "The formula should have a trailing newline"`

rather than

`assert ft.has_trailing_newline?, "The formula has a trailing newline"`

### DIFF
--- a/Library/Homebrew/test/test_cmd_audit.rb
+++ b/Library/Homebrew/test/test_cmd_audit.rb
@@ -36,6 +36,7 @@ class FormulaTextTests < Homebrew::TestCase
     assert ft =~ /\burl\b/, "The formula should match 'url'"
     assert_nil ft.line_number(/desc/), "The formula should not match 'desc'"
     assert_equal 2, ft.line_number(/\burl\b/)
+    assert ft.include?("Valid"), "The formula should include \"Valid\""
   end
 
   def test_trailing_newline
--- a/Library/Homebrew/test/test_cmd_audit.rb
+++ b/Library/Homebrew/test/test_cmd_audit.rb
@@ -13,53 +13,45 @@ class FormulaTextTests < Homebrew::TestCase
     FileUtils.rm_rf @dir
   end
 
-  def formula_text(name, text)
+  def formula_text(name, body = nil, options = {})
     path = Pathname.new "#{@dir}/#{name}.rb"
-    path.open("w") { |f| f.write text }
+    path.open("w") do |f|
+      f.write <<-EOS.undent
+        class #{Formulary.class_s(name)} < Formula
+          #{body}
+        end
+        #{options[:patch]}
+      EOS
+    end
     FormulaText.new path
   end
 
   def test_simple_valid_formula
-    ft = formula_text "valid", <<-EOS.undent
-      class Valid < Formula
-        url "http://www.example.com/valid-1.0.tar.gz"
-      end
-    EOS
+    ft = formula_text "valid", 'url "http://www.example.com/valid-1.0.tar.gz"'
 
-    refute ft.has_DATA?, "The formula doesn't have DATA"
-    refute ft.has_END?, "The formula doesn't have __END__"
-    assert ft.has_trailing_newline?, "The formula have a trailing newline"
+    refute ft.has_DATA?, "The formula should not have DATA"
+    refute ft.has_END?, "The formula should not have __END__"
+    assert ft.has_trailing_newline?, "The formula should have a trailing newline"
 
-    assert ft =~ /\burl\b/, "The formula match 'url'"
-    assert_nil ft.line_number(/desc/), "The formula doesn't match 'desc'"
+    assert ft =~ /\burl\b/, "The formula should match 'url'"
+    assert_nil ft.line_number(/desc/), "The formula should not match 'desc'"
     assert_equal 2, ft.line_number(/\burl\b/)
   end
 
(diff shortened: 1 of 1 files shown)


Library/Homebrew/test/test_cmd_audit.rb
```diff
@@ -13,53 +13,46 @@ def teardown
     FileUtils.rm_rf @dir
   end
 
-  def formula_text(name, text)
+  def formula_text(name, body = nil, patch = nil)
```
MikeMcQuaid: Could you make an `options = {}` for the `patch`? Will make it a bit more readable.

Library/Homebrew/test/test_cmd_audit.rb
```diff
@@ -13,53 +13,46 @@ def teardown
...
-      class Valid < Formula
-        url "http://www.example.com/valid-1.0.tar.gz"
-      end
-    EOS
+    ft = formula_text "valid", 'url "http://www.example.com/valid-1.0.tar.gz"'
 
     refute ft.has_DATA?, "The formula doesn't have DATA"
     refute ft.has_END?, "The formula doesn't have __END__"
-    assert ft.has_trailing_newline?, "The formula have a trailing newline"
+    assert ft.has_trailing_newline?, "The formula has a trailing newline"
```
MikeMcQuaid: Should read `must have` as its describing the assertion rather than the method or success. Same with values below.

Library/Homebrew/test/test_cmd_audit.rb
```diff
@@ -13,53 +13,46 @@ def teardown
...
 
-    assert ft =~ /\burl\b/, "The formula match 'url'"
+    assert ft =~ /\burl\b/, "The formula matches 'url'"
     assert_nil ft.line_number(/desc/), "The formula doesn't match 'desc'"
     assert_equal 2, ft.line_number(/\burl\b/)
+    assert ft.include?("Valid"), "The formula includes \"Valid\""
```
MikeMcQuaid: `assert_match "Valid", ft`

Library/Homebrew/test/test_cmd_audit.rb
```diff
@@ -13,53 +13,46 @@ def teardown
...
-      class End < Formula
-      end
-      __END__
-      a patch here
-    EOS
-
+    ft = formula_text "end", "", "__END__\na patch here"
     assert ft.has_END?, "The formula has __END__"
-    assert_equal "class End < Formula\nend", ft.without_patch
+    assert_equal "class End < Formula\n  \nend", ft.without_patch
```
MikeMcQuaid: Use `EOS.undent` here rather than embedding newlines.

MikeMcQuaid: (and in other cases you've modified above)

MikeMcQuaid:
A few nits but generally looking good!

Library/Homebrew/test/test_cmd_audit.rb
eirinikos: Do you mean `def formula_text(name, body = nil, options = {})` ?

Library/Homebrew/test/test_cmd_audit.rb
eirinikos: Unfortunately, this raises an error (`TypeError: no implicit conversion of FormulaText into String`). And `ft`'s `@text` can't be accessed, either.

Library/Homebrew/test/test_cmd_audit.rb
eirinikos: Thanks. But I'm having trouble using `EOS.undent ` and getting the test to pass. 😞 

With it written this way:
```
  def test_has_end
    ft = formula_text "end", "", <<-EOS.undent
      __END__
      a patch here
    EOS
    assert ft.has_END?, "The formula must have __END__"
    assert_equal <<-EOS.undent, ft.without_patch
      class End < Formula
      end
    EOS
  end
```

The test returns this discrepancy:
```
 "class End < Formula
-end
-"
+  
+end"
```

I've tried adding newlines here and there, but with no luck so far.

Library/Homebrew/test/test_cmd_audit.rb
MikeMcQuaid: Yep!

Library/Homebrew/test/test_cmd_audit.rb
MikeMcQuaid: Cool, that's fine as-is then, thanks for checking 👍

Library/Homebrew/test/test_cmd_audit.rb
MikeMcQuaid: That's cool, feel free to leave it as it was, thanks for trying 👍

eirinikos:
Thanks!

MikeMcQuaid:
Thanks again @eirinikos!

PR #684, opened 2016-08-11
Don't report .pyc file writes in sandbox logs

### LINKED ISSUE #683: filter .pyc messages out of sandbox log
_From @tdsmith on January 3, 2016 18:51_

Messages of the form `.*Python(\d+) deny file-write-create.*pyc` are just noise and aren't fatal; it would be nice to hide them from the sandbox log summary.

_Copied from original issue: Homebrew/legacy-homebrew#47640_

tdsmith:
These are never fatal and often confusing.

Fixes #683.

### DIFF
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -95,6 +95,10 @@ class Sandbox
       -k Sender sandboxd
     ]
     logs = Utils.popen_read("syslog", *syslog_args)
+
+    # These messages are confusing and non-fatal, so don't report them.
+    logs = logs.gsub(/^.*Python(\d+) deny file-write.*pyc$/, "").strip
+
     unless logs.empty?
       if @logfile
         log = open(@logfile, "w")
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -97,7 +97,7 @@ class Sandbox
     logs = Utils.popen_read("syslog", *syslog_args)
 
     # These messages are confusing and non-fatal, so don't report them.
-    logs = logs.lines.reject { |l| l.match /^.*Python\(\d+\) deny file-write.*pyc$/ }.join
+    logs = logs.lines.reject{ |l| l.match(/^.*Python\(\d+\) deny file-write.*pyc$/) }.join
 
     unless logs.empty?
       if @logfile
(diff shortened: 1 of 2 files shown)


Library/Homebrew/sandbox.rb
```diff
@@ -95,6 +95,10 @@ def exec(*args)
       -k Sender sandboxd
     ]
     logs = Utils.popen_read("syslog", *syslog_args)
+
+    # These messages are confusing and non-fatal, so don't report them.
+    logs = logs.gsub(/^.*Python(\d+) deny file-write.*pyc$/, "").strip
```
MikeMcQuaid: Does this strip the newlines too or need a ` \n`?

Library/Homebrew/test/test_sandbox.rb
```diff
@@ -25,4 +25,21 @@ def test_deny_write
...
+  def test_complains_on_failure
+    Utils.expects(:popen_read => "foo")
+    out, err = capture_io do
+      assert_raises(ErrorDuringExecution) { @sandbox.exec "false" }
+    end
+    assert_match out, "foo"
+  end
+
+  def test_ignores_bogus_python_error
+    bogus_error = "Mar 17 02:55:06 sandboxd[342]: Python(49765) deny file-write-unlink /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/errors.pyc\n"
```
MikeMcQuaid: Maybe worth making it multiline and checking that other lines aren't truncated. I know they aren't from the code above but might be worth making the test a little more robust.

MikeMcQuaid:
Looks good 👍. A few tiny questions/nits but should be good to go soon. Nice work.

Library/Homebrew/sandbox.rb
tdsmith: It'd have to be a multiline regex, it turns out; I don't know why I thought `strip` would fix this. ☕☕☕ Tried an approach with `.lines` instead; let me know how you like it.

Library/Homebrew/test/test_sandbox.rb
tdsmith: While looking at this, discovered that all my `assert_match`es were backwards, oops. Agree that it improves the test~!

MikeMcQuaid:
LGTM 👍, thanks for the changes.

PR #686, opened 2016-08-11
Add completions for --fetch-HEAD flag

vladshablinsky:


vladshablinsky:
@xu-cheng I've added the completions for bash and zsh. Bash completions work just fine, however zsh-completions for flags for `upgrade` works strange locally. Even `brew upgrade` cannot complete `--cleanup`. I guess it's something wrong with my zsh because it's really messy, so, if you don't see anything strange in implementation could we merge it then?

MikeMcQuaid:
ZSH completion looks good to me.

PR #704, opened 2016-08-12
tests: add cmd/audit unit tests

eirinikos:

-----

This change adds unit tests for `brew audit`. I'd like to refactor `formula_auditor`, but I figured I'd save that for a later time.

I'm having trouble getting `test_audit_desc` to pass; I suspect it has to do with iterating over a collection of nested hashes, but for the life of me I can't figure out what the problem is.

### DIFF
--- a/Library/Homebrew/test/test_cmd_audit.rb
+++ b/Library/Homebrew/test/test_cmd_audit.rb
@@ -373,4 +373,98 @@ class FormulaAuditorTests < Homebrew::TestCase
   ensure
     ENV["HOMEBREW_NO_GITHUB_API"] = original_value
   end
+
+  def test_audit_caveats
+    fa = formula_auditor "foo", <<-EOS.undent
+      class Foo < Formula
+        homepage "http://example.com/foo"
+        url "http://example.com/foo-1.0.tgz"
+
+        def caveats
+          "setuid"
+        end
+      end
+    EOS
+
+    fa.audit_caveats
+    assert_equal ["Don't recommend setuid in the caveats, suggest sudo instead."],
+      fa.problems
+  end
+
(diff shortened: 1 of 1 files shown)


Library/Homebrew/test/test_cmd_audit.rb
```diff
@@ -380,4 +380,99 @@ class Foo < Formula
...
+      content = <<-EOS.undent
+        class #{Formulary.class_s(formula[:name])} < Formula
+          url "http://example.com/#{formula[:name]}-1.0.tgz"
+          desc "#{formula[:desc]}"
+        end
+      EOS
+
+      fa = formula_auditor formula[:name], content, :strict => true
+      fa.audit_desc
+      assert fa.problems.empty? # this shouldn't be true, but it is!
```
MikeMcQuaid: It's true for `bar` because the length is 3 * 25 + 3 = 78 which is less than 80. Change it to `* 30` above and it'll work. Additionally, the line below is nicer (and works) as `assert_match formula[:problem], fa.problems.first.to_s`

eirinikos: Wow, thanks! I should have double-checked that arithmetic. 😳 

It looks like `"bar: #{'bar' * 25}".length` actually works out to 80, but as you pointed out, it would need to be more than 80 to trigger the warning.

> Additionally, the line below is nicer (and works) as `assert_match formula[:problem], fa.problems.first.to_s`

👍  I've changed it to `assert_match formula[:problem], fa.problems.first`.
(Forgot that `assert_match` would work here!)

MikeMcQuaid:
Thanks again @eirinikos!

PR #707, opened 2016-08-13
pull: non-core tap support for Homebrew Testing jobs

ilovezfs:
`test-bot --ci-testing` supports a `--tap` option for non-core taps, so
`brew pull` should too when pulling Homebrew Testing jobs.


ilovezfs:
Example usage:
```
brew pull --bottle --tap=homebrew/versions https://bot.brew.sh/job/Homebrew%20Testing/1035/
```

PR #726, opened 2016-08-17
tests: allow brew tests to accept multiple --only args

eirinikos:

-----

This is one of my attempts to change `brew tests` so that it allows multiple `--only` args (i.e., multiple test files and/or test methods). It doesn't work, though! 😞 

I looked for documentation on the arguments that `rake test` accepts; if I understand correctly, then `rake test` does accept options that are used in `Test::Unit` (as mentioned [here](http://rake.rubyforge.org/classes/Rake/TestTask.html)).

That led me to believe that `rake test` would indeed accept multiple test files or test methods as arguments (as detailed [here](http://scottpatten.ca/2007/10/ruby-unit-test-command-line-options.html)).

Unfortunately, I haven't been able to test my changes with `rake test` (or `bundle exec rake test`) itself, since I receive this esoteric error when trying to do so:

```
/usr/local/Library/Homebrew/os/mac.rb:27:in `full_version': undefined method `chomp' for nil:NilClass (NoMethodError)
	from /usr/local/Library/Homebrew/os.rb:19:in `<module:OS>'
	from /usr/local/Library/Homebrew/os.rb:1:in `<top (required)>'
	from /usr/local/Library/Homebrew/extend/os/emoji.rb:1:in `require'
	from /usr/local/Library/Homebrew/extend/os/emoji.rb:1:in `<top (required)>'
	from /usr/local/Library/Homebrew/emoji.rb:24:in `require'
	from /usr/local/Library/Homebrew/emoji.rb:24:in `<top (required)>'
	from /usr/local/Library/Homebrew/utils.rb:2:in `require'
	from /usr/local/Library/Homebrew/utils.rb:2:in `<top (required)>'
	from /usr/local/Library/Homebrew/extend/pathname.rb:4:in `require'
	from /usr/local/Library/Homebrew/extend/pathname.rb:4:in `<top (required)>'
	from /usr/local/Library/Homebrew/global.rb:3:in `require'
	from /usr/local/Library/Homebrew/global.rb:3:in `<top (required)>'
	from /usr/local/Library/Homebrew/test/testing_env.rb:5:in `require'
	from /usr/local/Library/Homebrew/test/testing_env.rb:5:in `<top (required)>'
	from /usr/local/Library/Homebrew/test/test_cmd_audit.rb:1:in `require'
	from /usr/local/Library/Homebrew/test/test_cmd_audit.rb:1:in `<top (required)>'
	from /usr/local/Library/Homebrew/test/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:15:in `require'
	from /usr/local/Library/Homebrew/test/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:15:in `block in <main>'
	from /usr/local/Library/Homebrew/test/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:4:in `select'
	from /usr/local/Library/Homebrew/test/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:4:in `<main>'
```

Thanks!

### DIFF
--- a/Library/Homebrew/cmd/tests.rb
+++ b/Library/Homebrew/cmd/tests.rb
@@ -41,9 +41,14 @@ module Homebrew
       args << "--trace" if ARGV.include? "--trace"
       if ARGV.value("only")
         ENV["HOMEBREW_TESTS_ONLY"] = "1"
-        test_name, test_method = ARGV.value("only").split("/", 2)
-        args << "TEST=test_#{test_name}.rb"
-        args << "TESTOPTS=--name=test_#{test_method}" if test_method
+
+        pairs = ARGV.value("only").split
+        pairs.each do |pair|
+          test_name, test_method = pair.split("/", 2)
+
+          args << "TEST=test_#{test_name}.rb" unless test_name.empty?
+          args << "TESTOPTS=--name=test_#{test_method}" if test_method
+        end
       end
       args += ARGV.named.select { |v| v[/^TEST(OPTS)?=/] }
       system "bundle", "exec", "rake", "test", *args


MikeMcQuaid:
> I looked for documentation on the arguments that rake test accepts; if I understand correctly, then rake test does accept options that are used in Test::Unit (as mentioned here).
> 
> That led me to believe that rake test would indeed accept multiple test files or test methods as arguments (as detailed here).

My reading of http://rake.rubyforge.org/classes/Rake/TestTask.html is that you may only be able to specify a single test, unfortunately. That said, we could move away from using TestTask if it's limiting us in this way.

Library/Homebrew/cmd/tests.rb
```diff
@@ -41,9 +41,14 @@ def tests
       args << "--trace" if ARGV.include? "--trace"
       if ARGV.value("only")
         ENV["HOMEBREW_TESTS_ONLY"] = "1"
-        test_name, test_method = ARGV.value("only").split("/", 2)
-        args << "TEST=test_#{test_name}.rb"
-        args << "TESTOPTS=--name=test_#{test_method}" if test_method
+
+        pairs = ARGV.value("only").split
```
MikeMcQuaid: `ARGV.value` only ever returns a single value. To allow something like `brew tests --only=cmd_audit/audit_desc --only=formulary/to_rack` to work you'd need to declare something like:

```diff
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -107,6 +107,13 @@ module HomebrewArgvExtension
     flag_with_value.strip_prefix(arg_prefix) if flag_with_value
   end
 
+  def values(name)
+    arg_prefix = "--#{name}="
+    flags_with_values = select { |arg| arg.start_with?(arg_prefix) }
+    return [] if flags_with_values.empty?
+    p flags_with_values.map { |arg| arg.strip_prefix(arg_prefix) }
+  end
+
```

and do:

```diff
--- a/Library/Homebrew/cmd/tests.rb
+++ b/Library/Homebrew/cmd/tests.rb
@@ -39,11 +39,9 @@ module Homebrew
 
       args = []
       args << "--trace" if ARGV.include? "--trace"
-      if ARGV.value("only")
+      unless (only = ARGV.values("only")).empty?
         ENV["HOMEBREW_TESTS_ONLY"] = "1"
-
-        pairs = ARGV.value("only").split
-        pairs.each do |pair|
+        only.each do |pair|
           test_name, test_method = pair.split("/", 2)
```

However, even with that the multiple `TEST=` arguments don't seem to do anything 😭

eirinikos: Ah... I'd thought that `ARGV.value("only")` could work for something like `--only=integration_cmds cmd_audit`, but I see that is not the case. 😢

eirinikos:
> My reading of http://rake.rubyforge.org/classes/Rake/TestTask.html is that you may only be able to specify a single test, unfortunately. That said, we could move away from using TestTask if it's limiting us in this way.

Do any alternatives come to mind? I'm guessing that we would move away not just from using TestTask, but from using Rake itself, too?

Library/Homebrew/cmd/tests.rb
UniqMartin: It will work if you make the thing after `--only=` a single argument as seen from the shell, e.g. by using comma (`,`) as a delimiter, or quoting the entire argument that includes the space like `--only="integration_cmds cmd_audit"` or `--only=integration_cmds\ cmd_audit`.

UniqMartin:
Can you elaborate a bit on the motivation for this PR? What is the benefit over running `brew tests` (with the desired options, e.g. different `--only=`) multiple times in a row?

> This is one of my attempts to change `brew tests` so that it allows multiple `--only` args (i.e., multiple test files and/or test methods). It doesn't work, though! 😞

I think from a user interface point of view it might be better (and also involve less changes) to teach `--only=` to accept multiple test files by supplying them in a single argument, delimited by a comma, e.g. `--only=file1,file2` instead of `--only=file1 --only=file2`.

> Unfortunately, I haven't been able to test my changes with `rake test` (or `bundle exec rake test`) itself, since I receive this esoteric error when trying to do so:

We should probably provide a better error message in this case or flat out refuse to run the test suite if invoked from a vanilla `rake test`, but the gist of why you see this esoteric error is that the Homebrew initialization code assumes quite a few environment variables to be set up in advance (in `bin/brew` and `Library/Homebrew/brew.sh`), thus loading any of the Homebrew Ruby code without going through this shell code first hasn't been really supported for quite a while.

MikeMcQuaid:
> Do any alternatives come to mind? I'm guessing that we would move away not just from using TestTask, but from using Rake itself, too?

Thinking about this: the only alternative I can think of is to run `rake` multiple times but then I also thought:

> What is the benefit over running brew tests (with the desired options, e.g. different --only=) multiple times in a row?

which I now think I agree with, unfortunately.

Thanks for the PR anyway @eirinikos and hope it was a useful learning experience but passing on this for now 😭

PR #735, opened 2016-08-17
sandbox tweak

xu-cheng:


Library/Homebrew/sandbox.rb
```diff
@@ -8,11 +8,12 @@ class Sandbox
   ].freeze
 
   def self.available?
-    OS.mac? && File.executable?(SANDBOX_EXEC)
+    OS.mac? && OS::Mac.version >= "10.6" && File.executable?(SANDBOX_EXEC)
```
MikeMcQuaid: Does that file exist and is executable on e.g. 10.5 but just doesn't work? If not I think we can leave it as-is.

Library/Homebrew/sandbox.rb
```diff
@@ -8,11 +8,12 @@ class Sandbox
...
   end
 
   def self.formula?(formula)
     return false unless available?
+    return false if ARGV.no_sandbox?
```
MikeMcQuaid: 👍 to this.

Library/Homebrew/sandbox.rb
xu-cheng: You can see commit message for explanation. Basically it exists but too old to support DSL we use.

MikeMcQuaid: Ok 🆒 👍

PR #740, opened 2016-08-18
BumpFormulaPR learns --remote option

jasonkarns:
When bumping formula that are in taps owned by the developers doing the
bumping, it would be convenient if the submitted PRs could be opened
from the owner repo itself. (For example, I am a contributor to nodenv
tap, and I would like to open PRs using branches from the nodenv org
rather than a fork.)

Presently, bump-formula-pr derives the remote name by running hub-fork.
This actually forks the repo on github. However, if I am already an
owner or contributor of the origin remote, then there is no need to
create a fork. This PR teaches bump-formula-pr to accept the `--remote=`
option, accepting a remote name as the argument.

If provided, hub-fork is not called. Instead, git-push is executed
against the given remote name.

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -142,23 +142,36 @@ module Homebrew
       if ARGV.dry_run?
         ohai "git checkout -b #{branch} origin/master"
         ohai "git commit --no-edit --verbose --message='#{formula.name} #{new_formula_version}#{devel_message}' -- #{formula.path}"
-        ohai "hub fork --no-remote"
-        ohai "hub fork"
-        ohai "hub fork (to read $HUB_REMOTE)"
-        ohai "git push $HUB_REMOTE #{branch}:#{branch}"
+        ohai "git push #{remote} #{branch}:#{branch}"
         ohai "hub pull-request --browse -m '#{formula.name} #{new_formula_version}#{devel_message}'"
       else
         safe_system "git", "checkout", "-b", branch, "origin/master"
         safe_system "git", "commit", "--no-edit", "--verbose",
           "--message=#{formula.name} #{new_formula_version}#{devel_message}",
           "--", formula.path
-        safe_system "hub", "fork", "--no-remote"
-        quiet_system "hub", "fork"
-        remote = Utils.popen_read("hub fork 2>&1")[/fatal: remote (.+) already exists\./, 1]
-        odie "cannot get remote from 'hub'!" if remote.to_s.empty?
         safe_system "git", "push", remote, "#{branch}:#{branch}"
         safe_system "hub", "pull-request", "--browse", "-m",
           "#{formula.name} #{new_formula_version}#{devel_message}\n\nCreated with `brew bump-formula-pr`."
+      end
+    end
+  end
+
+  private
+
+  def remote
+    if ARGV.flag?("--remote") || ARGV.value("remote")
+      ARGV.value("remote").tap do |name|
+        odie "--remote specified but no <remote_url> given" if name.to_s.empty?
+      end
+    elsif ARGV.dry_run?
+      ohai "hub fork --no-remote"
+      ohai "hub fork"
+      ohai "hub fork (to read $HUB_REMOTE)"
+    else
+      safe_system "hub", "fork", "--no-remote"
+      quiet_system "hub", "fork"
+      Utils.popen_read("hub fork 2>&1")[/fatal: remote (.+) already exists\./, 1].tap do |name|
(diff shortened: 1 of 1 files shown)


ilovezfs:
:-1: this seems like something that would make more sense in an external command: https://github.com/Homebrew/brew/blob/028d155e97014076adf4e4dbc7c888fd871de321/share/doc/homebrew/External-Commands.md

MikeMcQuaid:
I don't mind this functionality being there as long as it's not an explicitly passed option e.g. an environment variable. It does make the code more complex so I'd like to see more than just yourself wanting this option first.

Is there a reason you want to not open it on a fork beyond aesthetics?

jasonkarns:
@MikeMcQuaid 

> Is there a reason you want to not open it on a fork beyond aesthetics?

Clutter? The core team on the nodenv repos are the only people submitting formula updates (as it's an automated part of the release for the nodenv plugins themselves) and as owners we all work on origin. Just doesn't make sense to me to have forks of repos I own, I guess.

I guess I'd like to turn the question around :) Is there a reason to use forks when the submission is from a repo owner? 

Side question: what is the criteria for adding predicate methods to ARGV? The conditional itself would be simpler if we could use the same style as `ARGV.dry_run?`


> as long as it's not an explicitly passed option e.g. an environment variable.

I don't understand this. Environment variables are intended as something that lives at a higher level than a command; something to configure across an entire, well, _environment_. The remote option does not meet that high bar. It is a highly localized option that only applies to a given invocation of the command (because it's specific to the tap being bumped). In other words, no one could set the environment variable and use this command regularly. They would need to repeatedly modify or overwrite the variable for individual invocations of the command.

jasonkarns:
@ilovezfs I'm not sure I follow your reasoning. I want to use the existing dev-cmd but just need control over the `git push` bit nestled deep within. Extracting this to an external command would duplicate 99% of the existing dev command only to modify 1%, no?

ilovezfs:
>I'm not sure I follow your reasoning.

I cannot imagine myself ever using this option. Not sure I have much more reasoning for you than that.

MikeMcQuaid:
> I guess I'd like to turn the question around :) Is there a reason to use forks when the submission is from a repo owner?

Yep: the main reason is because `brew tap` is Git based so anyone else will get that branch in their local repo.



> They would need to repeatedly modify or overwrite the variable for individual invocations of the command.

My feeling in your case (correct me if I'm wrong) was that you'd always want it to behave a certain way in which case it seems like it'd be a pain to have to manually remember to type the flag every time.

jasonkarns:
> Yep: the main reason is because brew tap is Git based so anyone else will get that branch in their local repo.

Good point. (I diligently delete the branches whenever they are merged, but that relies on discipline)


> My feeling in your case (correct me if I'm wrong) was that you'd always want it to behave a certain way in which case it seems like it'd be a pain to have to manually remember to type the flag every time.

We already need to to behave 2 different ways: defaulting to forks when bumping nodenv/node-build (which are in homebrew-core) and using our repo when bumping formulas in our tap. But again, the option is still tap-specific. In our use case, nobody types it manually because the whole thing is scripted.

In theory, one _could_ need a different remote name per tap. (plus using fork for homebrew-core) However, in practice there are really only two cases: using fork and using `origin`. Maybe the option shouldn't bother accepting the remote name, but instead just be a flag that triggers using origin? That would simplify the code. Maybe `--push-to-origin` or something?

MikeMcQuaid:
> Maybe the option shouldn't bother accepting the remote name, but instead just be a flag that triggers using origin? That would simplify the code. Maybe --push-to-origin or something?

This makes sense to me 👍

MikeMcQuaid:
and I think you've convinced me that if we can get this code simple enough we should accept this.

jasonkarns:
@MikeMcQuaid see now I'm hesitant because 

> anyone else will get that branch in their local repo.

is now making me think we should just switch to using forks instead of pushing to origin. I may close this issue and use bump-formula-pr unmodified for a bit and see if I still want the origin flag.

jasonkarns:
@jawshooah since you're effectively the only other person that will be bumping nodenv formulae in our nodenv taps, do you have any opinion. (On the idea of moving the homebrew-nodenv taps to a fork model so that any bumps are on our personal forks and PRs opened against origin. This would allow us to use the bump-formula-pr dev-cmd unmodified.)

jawshooah:
@jasonkarns I'm pretty ambivalent on this, as I already create branches on my personal forks for all PRs, even to repos owned by organizations of which I am a part.

jasonkarns:
I think that settles it, then. I'll move to a fork model and use bump-formula-pr as-is.

Thanks everyone!

PR #744, opened 2016-08-18
version: fix parsing of URLs ending with version

ilovezfs:
https://github.com/lihaoyi/Ammonite/releases/download/0.7.4/0.7.4

was parsing as "0.7" not "0.7.4" since ".4" was being treated as a
legitimate file extension. At minimum, let's insist that a valid file
extension include at least one letter to avoid lopping off part of the
correct version string.

CC @DomT4 @UniqMartin

### DIFF
--- a/Library/Homebrew/test/test_versions.rb
+++ b/Library/Homebrew/test/test_versions.rb
@@ -438,8 +438,12 @@ class VersionParsingTests < Homebrew::TestCase
       "https://opam.ocaml.org/archives/easy-format.1.0.2+opam.tar.gz"
   end
 
-  def test_waf_version
+  def test_no_extension_version
     assert_version_detected "1.8.12", "https://waf.io/waf-1.8.12"
+    assert_version_detected "0.7.1", "https://codeload.github.com/gsamokovarov/jump/tar.gz/v0.7.1"
+    assert_version_detected "0.9.1234", "https://my.datomic.com/downloads/free/0.9.1234"
+    assert_version_detected "0.9", "https://my.datomic.com/downloads/free/0.9.1t34"
+    assert_version_detected "1.2.3", "https://my.datomic.com/downloads/free/1.2.3"
   end
 
   def test_dash_separated_version
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -294,6 +294,8 @@ class Version
       spec.basename.to_s
     elsif %r{((?:sourceforge\.net|sf\.net)/.*)/download$}.match(spec_s)
       Pathname.new(spec.dirname).stem
+    elsif %r{\.[^a-zA-Z]+$}.match(spec_s)
+      Pathname.new(spec_s).basename
     else
       spec.stem
     end


ilovezfs:
And here's another one: https://github.com/Homebrew/homebrew-core/blob/master/Formula/datomic.rb#L4-L5

MikeMcQuaid:
👍 but add a test to `test_versions` for these cases.

ilovezfs:
@MikeMcQuaid haha it was already in there!
```
    # URL with no extension
    # e.g. https://waf.io/waf-1.8.12
    # e.g. https://codeload.github.com/gsamokovarov/jump/tar.gz/v0.7.1
    m = /[-v]((?:\d+\.)*\d+)$/.match(spec_s)
```
Looks like the "correct" fix is
```
m = /[-v]?((?:\d+\.)*\d+)$/.match(spec_s)
```

ilovezfs:
@MikeMcQuaid PR refreshed.

MikeMcQuaid:
🚢

ilovezfs:
@MikeMcQuaid any idea what's going on here?
```
$ git push
Counting objects: 7, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (7/7), done.
Writing objects: 100% (7/7), 951 bytes | 0 bytes/s, done.
Total 7 (delta 5), reused 0 (delta 0)
remote: Resolving deltas: 100% (5/5), completed with 5 local objects.
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: Required status check "default" is expected
To https://github.com/Homebrew/brew
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/Homebrew/brew'
```

DomT4:
Oh, Mike banned pushing. You'll need to literally merge it.

ilovezfs:
Huh? Why?

ilovezfs:
Are you saying we have to interact with the brew repo via the web browser to merge anything? Or is there some way this can be done from CLI? Please tell me it's the latter, or I must be in Twilight zone.

DomT4:
Mostly to make sure everything goes through a successful PR (and its various tests) first. There's no way to communicate to GitHub that a PR has passed & is safe to merge via the `brew pull` mechanism.

DomT4:
Possibly my fault. I accidentally nicked his PR by pushing directly 😄.

ilovezfs:
@DomT4 sounds like a slight overreaction :scream:

DomT4:
I loathe merge commits with a sort of unholy passion, and they're the only way to preserve GPG signing of commits from the GitHub interface, so I'm not in love with the block, but I understand the reasoning for it & theoretically it'll prevent breakage from creeping in more than the current system has.

ilovezfs:
@DomT4 `hub merge` command works fine so I've stopped caring now. But I agree merge commits are pretty much awful.

MikeMcQuaid:
`hub merge` or I believe just doing a merge on the command-line. It needs to ensure that the SHA1 that was tested is the one that's merged.

As @DomT4 says: they preserve GPG and they provide a pointer to the PR the commit came from (which `brew pull` loses with multi-commit PRs which most are in Homebrew/brew nowadays).

ilovezfs:
@MikeMcQuaid the biggest problem is timestamps don't get updated unless you manually bump them, so just about everything ends up with dates older than when they actually were committed and the history becomes unintelligible in terms of actual chronological order of commits let alone when they were actually merged but :see_no_evil: as long as it works from CLI.

PR #725, opened 2016-08-16
Move Cask core code to Homebrew

AnastasiaSulyagina:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Moving [Cask](https://github.com/caskroom/homebrew-cask) core code to Homebrew as part of [#14384](https://github.com/caskroom/homebrew-cask/issues/14384).

Change is splitted in 2 commits to be easier to review. Any advice on better splitting or anything else welcome.

cc @MikeMcQuaid @vitorgalvao @jawshooah @adidalal (bring other cask maintainers please)

PR #747, opened 2016-08-18
test-bot: Install dependencies for Cask container tests.

reitermarkus:

-----

In preparation of https://github.com/Homebrew/brew/pull/725, 

- `adobe-air`
- `cabextract`
- `unar` and
- `xz`

have to be installed before running `brew cask-tests`, otherwise some of the container tests will be skipped.

### DIFF
--- a/Library/Homebrew/dev-cmd/test-bot.rb
+++ b/Library/Homebrew/dev-cmd/test-bot.rb
@@ -659,7 +659,13 @@ module Homebrew
         test "brew", "readall", "--syntax"
         if OS.mac? &&
            (HOMEBREW_REPOSITORY/"Library/Homebrew/cask/cmd/brew-cask-tests.rb").exist?
-          run_as_not_developer { test "brew", "tap", "caskroom/cask" }
+          run_as_not_developer do
+            test "brew", "tap", "caskroom/cask"
+            test "brew", "cask", "install", "https://raw.githubusercontent.com/caskroom/homebrew-cask/master/Casks/adobe-air.rb"
+            test "brew", "install", "https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/cabextract.rb"
+            test "brew", "install", "https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/unar.rb"
+            test "brew", "install", "https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/xz.rb"
+          end
           test "brew", "cask-tests"
         end
 


MikeMcQuaid:
> have to be installed before running brew cask-tests, otherwise some of the container tests will be skipped.

Longer-term we need to figure out a way of testing these things that doesn't require additional dependencies being installed on our bots (like `brew tests` does). Eventually we'll want to merge `brew tests` and `brew cask-tests` too.

Unfortunately the `brew install` does an implicit tap of homebrew/core which will be relatively slow. Where are the casks stored in this case? It's not testing a `/usr/local` Homebrew install.

reitermarkus:
> Longer-term we need to figure out a way of testing these things that doesn't require additional dependencies

I completely agree, however, I don't see a way to test the correct detection and extraction of containers without having the tools to extract them installed. I think we can neglect at least `cabextract` and `xz` for now, as there aren't any Casks using them. Also only one using `unar` will be left after https://github.com/caskroom/homebrew-cask/pull/23918 is merge.

MikeMcQuaid:
> I completely agree, however, I don't see a way to test the correct detection and extraction of containers without having the tools to extract them installed. I think we can neglect at least cabextract and xz for now, as there aren't any Casks using them.

Sounds like a good argument to 💀 that code 😉.

Library/Homebrew/dev-cmd/test-bot.rb
```diff
@@ -659,7 +659,13 @@ def homebrew
         test "brew", "readall", "--syntax"
         if OS.mac? &&
            (HOMEBREW_REPOSITORY/"Library/Homebrew/cask/cmd/brew-cask-tests.rb").exist?
-          run_as_not_developer { test "brew", "tap", "caskroom/cask" }
+          run_as_not_developer do
+            test "brew", "tap", "caskroom/cask"
+            test "brew", "cask", "install", "https://raw.githubusercontent.com/caskroom/homebrew-cask/master/Casks/adobe-air.rb"
+            test "brew", "install", "https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/cabextract.rb"
+            test "brew", "install", "https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/unar.rb"
+            test "brew", "install", "https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/xz.rb"
```
MikeMcQuaid: Would it be enough to have these installed systemwide on the build machines? This means they would likely not be kept particularly up-to-date but might be nicer than sticking them here.

reitermarkus: If that's possible, yes, as long as they can extract the containers I don't care if they are up-to-date.

ilovezfs:
:-1:

MikeMcQuaid:
Passing on this because I'd like to get `brew cask-tests` so it doesn't require `caskroom` to be tapped and this is going in the opposite direction.

PR #757, opened 2016-08-19
bump-formula-pr now creates non-tracking branches

### LINKED ISSUE #755: bump-formula-pr fails when attempting to create PR

- [x] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

### Bug reports:

Failed command (bumping nodenv-each formula in nodenv/homebrew-nodenv tap):

```
HOMEBREW_DEVELOPER=true brew bump-formula-pr --dry-run --url=https://github.com/nodenv/nodenv-each/archive/v1.1.0.tar.gz --sha256=b398dabc5569f084b64b1daa377c5fd7ae508a166c5583f9eace00a6aeadf029 nodenv-each
```

Everything succeeded down (and including) the git-push and then:

```
Aborted: head branch is the same as base ("master")
(use `-h <branch>` to specify an explicit pull request head)
Error: Failure while executing: hub pull-request --browse -m nodenv-each\ 1.1.0

Created\ with\ `brew\ bump-formula-pr`.
```

Looking in the tap, the branch was set to use `origin/master` as upstream. The current command that creates the branch is `git checkout -b <branch> origin/master`. On my machine, this results in a local branch `<branch>` tracking `origin/master`. The push still works because the remote is specified explicitly (rather than defaulting to `@{upstream}`), but then the `hub pr` fails because it's trying to open a pr from `origin/master` to `origin/master`.

There are a couple ways to fix this:

1. Pass the `--no-track` option to `git checkout` so that the new branch doesn't set `origin/master` as the upstream
2. Explicitly pass `-h <remote>/<branch>` to `hub pull-request` so that hub uses the proper upstream as head instead of whatever the locally-configured upstream is.

I will shortly be opening a PR that addresses this by implementing option 1 above. Option 2 feels like treating the symptom rather than the cause.

jasonkarns:
Without `--no-track`, some git setups may automatically set
`origin/master` as the tracked upstream for the newly created branch.
This upstream is what hub defaults as --head when opening PRs. By not
allowing git to set `origin/master` as upstream, hub can then use the
proper --head for the PR.

Since hub still needs to know what --head is intended to be, we can set
the branch's upstream when pushing: with the --set-upstream option.

Fixes #755

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -140,15 +140,15 @@ module Homebrew
     formula.path.parent.cd do
       branch = "#{formula.name}-#{new_formula_version}"
       if ARGV.dry_run?
-        ohai "git checkout -b #{branch} origin/master"
+        ohai "git checkout --no-track -b #{branch} origin/master"
         ohai "git commit --no-edit --verbose --message='#{formula.name} #{new_formula_version}#{devel_message}' -- #{formula.path}"
         ohai "hub fork --no-remote"
         ohai "hub fork"
         ohai "hub fork (to read $HUB_REMOTE)"
-        ohai "git push $HUB_REMOTE #{branch}:#{branch}"
+        ohai "git push --set-upstream $HUB_REMOTE #{branch}:#{branch}"
         ohai "hub pull-request --browse -m '#{formula.name} #{new_formula_version}#{devel_message}'"
       else
-        safe_system "git", "checkout", "-b", branch, "origin/master"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Makes sense to me 👍

UniqMartin:
:+1: Agreed, this tweak makes a lot of sense. I think we can merge as soon as the build turns green. (I rescheduled the CI build as it had previously failed for unrelated reasons.)

UniqMartin:
Thanks for your contribution and this tweak to the `bump-formula-pr` command, @jasonkarns! :tada:

PR #758, opened 2016-08-19
Remove Cask developer scripts.

reitermarkus:
Moved them back in https://github.com/caskroom/homebrew-cask/pull/23920 as most of them are meant to be run from within the Tap root.


MikeMcQuaid:
Thanks @reitermarkus!

PR #760, opened 2016-08-19
bump-formula-pr: return to starting branch

jasonkarns:
After branching, bumping, pushing, and pr-ing; return to whatever branch
was originally checked out.

In most cases, I'd imagine users to want to continue receiving tap
updates from master. However, after using bump-formula-pr, the tap in
which the formula was bumped is left on the working branch that was
doing the bumping and pull-request. After opening the PR, we should
return to whatever branch the user originally had checked out – most
likely master. (But git allows us to just say "previous branch" by using
`-`)


UniqMartin:
This makes a lot of sense to me, but would prefer @ilovezfs as the most prominent user of this developer command to chime in and comment on this, too.

ilovezfs:
```
e9da9f0d (ilovezfs         2016-07-06 05:00:45 -0700 218)         safe_system "git", "checkout", "-"
```

UniqMartin:
Thanks for your contribution and this tweak to the `bump-formula-pr` command, @jasonkarns! :tada:

Issue #759, opened 2016-08-19
cask/test is too big


chdiza:
I just updated my local installation, and was surprised to find that the import of the cask infrastructure has dropped some relatively huge files on my machine.  In particular, Library/Homebrew/cask/test/support/binaries is relatively enormous.  This is mostly the fault of "transmission-2.61.dmg".

Surely a smaller, `dmg` file can be used for testing purposes?  Either one holding a tiny app or a dummy app?

That one file is almost twice as big as the entire Library/Homebrew directory was prior to the cask import.

MikeMcQuaid:
We'll accept PRs to remove them.

chdiza:
I know nothing of cask, so I can't do it.  But I hope that someone who knows what they're doing can replace that huge dmg with something smaller.  And preferably the `caffeine` ones with smaller dummies, too.

chdiza:
(As an aside: if `brew` is eventually going to be a cross-platform gem, why is cask infrastructure being pulled in at all?  Non-mac users will have no use for it, but it's tripled the size of HB and---if memory serves---sextupled the size of a shallow clone of the `brew` repo.)

MikeMcQuaid:
> (As an aside: if brew is eventually going to be a cross-platform gem, why is cask infrastructure being pulled in at all? Non-mac users will have no use for it, but it's tripled the size of HB and---if memory serves---sextupled the size of a shallow clone of the brew repo.)

The shallow clone will shrink as we are able to merge more core. It's been integrated because it still seems to have a larger userbase than e.g. Linuxbrew and that code being included is not harming anyone who does not run it.

PR #763, opened 2016-08-20
Use dummy containers for cask tests.

reitermarkus:
Reduces the size of the binaries used for the container tests from 6.3 MB to 86 KB.

-----

Closes https://github.com/Homebrew/brew/issues/759.


MikeMcQuaid:
Nice work 👏

reitermarkus:
> Nice work :clap:

Well, at least now I know how to build an Adobe AIR app, just because I needed to make a dummy one. :joy:

chdiza:
Thanks!

PR #765, opened 2016-08-20
Install Cask’s default tap if untapped.

reitermarkus:
`cask` should auto-tap `caskroom/cask` as before if it is untapped.


MikeMcQuaid:
Loving your work already @reitermarkus!

PR #771, opened 2016-08-20
Forgot renaming `cab-container`.

reitermarkus:


Library/Homebrew/cask/test/test_helper.rb
```diff
@@ -28,15 +28,18 @@ def shutup
...
     ensure
-      $stderr.reopen tmperr
-      $stdout.reopen tmpout
+      $stderr.reopen(err)
+      $stdout.reopen(out)
+      err.close
+      out.close
     end
   end
 end
```
reitermarkus: Not sure what would be the best way to reuse this method.

reitermarkus: I somehow missed that https://github.com/Homebrew/brew/blob/bd1ded93ec46b2ae9c37dbf17dd6d74d3f727440/Library/Homebrew/cask/spec/support/shutup_helper.rb exists.

PR #772, opened 2016-08-21
Refactor `shutup`.

reitermarkus:
Refactored the `shutup` helper method used in tests as this was defined in three different places.


MikeMcQuaid:
Great work. This is the type of refactoring I'm looking forward to.

Just FYI: longer-term we should look at merging `brew tests` and `brew cask-tests` together into a single test run.

PR #773, opened 2016-08-21
Also report Cask coverage.

reitermarkus:
Not actually sure if this is even possible with the current setup, so I might need some help with this.

### DIFF
--- a/Library/Homebrew/test/.simplecov
+++ b/Library/Homebrew/test/.simplecov
@@ -1,18 +1,16 @@
-# vim: filetype=ruby
+#!/usr/bin/env ruby
 
 SimpleCov.start do
-  tests_path = File.dirname(__FILE__)
-
-  minimum_coverage 40 unless ENV["HOMEBREW_TESTS_ONLY"]
-  coverage_dir File.expand_path("#{tests_path}/coverage")
-  root File.expand_path("#{tests_path}/..")
+  coverage_dir File.expand_path("../coverage", File.realpath(__FILE__))
+  root File.expand_path("../..", File.realpath(__FILE__))
 
   # We manage the result cache ourselves and the default of 10 minutes can be
   # too low (particularly on Travis CI), causing results from some integration
   # tests to be dropped. This causes random fluctuations in test coverage.
   merge_timeout 86400
 
-  add_filter "/Homebrew/cask/"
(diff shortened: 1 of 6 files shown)


MikeMcQuaid:
> Not actually sure if this is even possible with the current setup, so I might need some help with this.

I think it's not; pretty sure Coveralls will only take a single test run's worth of coverage so to get this reporting again we'll need to make `brew cask-tests` part of `brew tests` and report a single coverage figure for everything.

reitermarkus:
Nice, it's working now!

MikeMcQuaid:
@reitermarkus Reporting both?

reitermarkus:
Yes, just rebased so this will take a while again to show up.

reitermarkus:
bildschirmfoto 2016-08-26 um 15 09 24

MikeMcQuaid:
Looks like it's covering a bunch of stuff it wasn't before too 🎉

Library/Homebrew/test/.simplecov
```diff
@@ -45,9 +44,9 @@ SimpleCov.start do
...
 
-# Don't use Coveralls outside of CI, as it will override SimpleCov's default
-# formatter causing the `index.html` not to be written once all tests finish.
-if RUBY_VERSION.split(".").first.to_i >= 2 && !ENV["HOMEBREW_INTEGRATION_TEST"] && ENV["CI"]
-  require "coveralls"
-  Coveralls.wear!
-end
+## Don't use Coveralls outside of CI, as it will override SimpleCov's default
+## formatter causing the `index.html` not to be written once all tests finish.
+#if RUBY_VERSION.split(".").first.to_i >= 2 && !ENV["HOMEBREW_INTEGRATION_TEST"] && ENV["CI"]
```
jawshooah: Better to remove this block than leave it commented out, no?

reitermarkus: Should already be removed in the other commit I pushed before.

MikeMcQuaid:
Looks good to me (particularly the numbers 😍) but will let @UniqMartin give this a once over as he's worked on this coverage stuff in the past for us.

reitermarkus:
@UniqMartin, feel free to merge this if you don't have any objections.

PR #786, opened 2016-08-22
gpg: syntax additions

DomT4:

-----

This is split out of yet another in-progress branch, which was split out of another in-progress branch _(yeah, I know I'm making life complicated for myself)_, because I've been having an absolute nightmare with getting the broader fetch/verify mechanism to work consistently for quite some time now 😭, despite trying multiple ways to get that done, so here's _some_ progress to review.

This is obviously a WIP rather than the whole deal, designed to be enough for things like:
```ruby
Formula["wget"].gpg
Formula["wget"].devel.gpg
Formula["wget"].resource("test").gpg
```
to return the correct variables, and for `brew` to be able to load the changed `wget` with `info`, `install`, etc without going _"I don't know what that gpg thing is"_.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -29,9 +29,11 @@ require "migrator"
 # @see https://github.com/styleguide/ruby Ruby Style Guide
 #
 # <pre>class Wget < Formula
+#   desc "Internet file retriever"
 #   homepage "https://www.gnu.org/software/wget/"
-#   url "https://ftp.gnu.org/gnu/wget/wget-1.15.tar.gz"
-#   sha256 "52126be8cf1bddd7536886e74c053ad7d0ed2aa89b4b630f76785bac21695fcd"
+#   url "https://ftpmirror.gnu.org/wget/wget-1.18.tar.xz"
+#   mirror "https://ftp.gnu.org/gnu/wget/wget-1.18.tar.xz"
+#   sha256 "b5b55b75726c04c06fe253daec9329a6f1a3c0c1878e3ea76ebfebc139ea9cc1"
 #
 #   def install
 #     system "./configure", "--prefix=#{prefix}"
@@ -294,6 +296,12 @@ class Formula
   # @see .version
   def version
     active_spec.version
+  end
+
+  # GPG verification for current spec.
+  # @see .gpg
+  def gpg
+    active_spec.gpg
   end
 
   def update_head_version
@@ -1808,6 +1816,17 @@ class Formula
     # <pre>sha256 "2a2ba417eebaadcb4418ee7b12fe2998f26d6e6f7fda7983412ff66a741ab6f7"</pre>
     Checksum::TYPES.each do |type|
       define_method(type) { |val| stable.send(type, val) }
+    end
+
+    # @!attribute [w] gpg
+    # An optional mechanism to verify the download with GnuPG as well as the
+    # standard checksum verification. Upstream must publish signature files
+    # for us to be able to use this mechanism. Those signature files often
+    # have either a ".gpg" or ".sig" file extension.
+    #
+    # <pre>gpg "https://ftpmirror.gnu.org/wget/wget-1.18.tar.xz.sig"</pre>
+    def gpg(val)
+      stable.gpg(val)
     end
 
     # @!attribute [w] bottle
--- a/Library/Homebrew/resource.rb
+++ b/Library/Homebrew/resource.rb
@@ -12,6 +12,7 @@ class Resource
   attr_reader :mirrors, :specs, :using, :source_modified_time
   attr_writer :version
   attr_accessor :download_strategy, :checksum
+  attr_rw :gpg
 
   # Formula name must be set after the DSL, as we have no access to the
   # formula name before initialization of the formula
@@ -37,6 +38,10 @@ class Resource
     def mirrors
       @resource.mirrors
     end
+
+    def gpg
+      @resource.gpg
+    end
   end
 
   def initialize(name = nil, &block)
@@ -46,6 +51,7 @@ class Resource
     @mirrors = []
     @specs = {}
     @checksum = nil
+    @gpg = nil
     @using = nil
     instance_eval(&block) if block_given?
   end
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -3,6 +3,7 @@ require "resource"
 require "checksum"
 require "version"
 require "options"
+require "gpg"
 require "build_options"
 require "dependency_collector"
 require "utils/bottles"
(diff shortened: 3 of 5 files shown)


Library/Homebrew/formula.rb
```diff
@@ -29,9 +29,12 @@
...
 #
 # <pre>class Wget < Formula
+#   desc "Internet file retriever"
 #   homepage "https://www.gnu.org/software/wget/"
-#   url "https://ftp.gnu.org/gnu/wget/wget-1.15.tar.gz"
-#   sha256 "52126be8cf1bddd7536886e74c053ad7d0ed2aa89b4b630f76785bac21695fcd"
+#   url "https://ftpmirror.gnu.org/wget/wget-1.18.tar.xz"
+#   mirror "https://ftp.gnu.org/gnu/wget/wget-1.18.tar.xz"
+#   sha256 "b5b55b75726c04c06fe253daec9329a6f1a3c0c1878e3ea76ebfebc139ea9cc1"
+#   gpg "https://ftpmirror.gnu.org/wget/wget-1.18.tar.xz.sig"
```
MikeMcQuaid: Not sure we need these additions but don't feel strongly. Think `gpg` at least is worth punting on until we've worked through the inevitable early issues.

Library/Homebrew/resource.rb
```diff
@@ -152,6 +157,11 @@ def mirror(val)
     mirrors << val
   end
 
+  def gpg(val = nil)
```
MikeMcQuaid: Can you use `attr_rw` here? Maybe not but figure I'd ask.

Library/Homebrew/software_spec.rb
```diff
@@ -64,6 +66,11 @@ def url(val = nil, specs = {})
     dependency_collector.add(@resource)
   end
 
+  def gpg(val = nil)
```
MikeMcQuaid: Can you use attr_rw here? Maybe not but figure I'd ask.

Also, I think this contradicts the `def_delegators` above.

MikeMcQuaid: (I'd favour just the `def_delegators`)

Library/Homebrew/test/test_gpg.rb
```diff
@@ -7,14 +7,24 @@ def setup
     @dir = Pathname.new(mktmpdir)
   end
 
+  def teardown
+    @dir.rmtree
```
MikeMcQuaid: Favour `FileUtils.rm_rf` which fails more gracefully.

MikeMcQuaid:
A few comments but 👍 to the overall approach. I think it's probably worth punting on merging this until the value is used for something e.g. download verification though.

apjanke:
Approach looks good. Agree on waiting to merge until we have a formula that uses it for something, so we can use that as a "tracer bullet" to exercise the code.

Library/Homebrew/formula.rb
DomT4: Updated it because I thought being API documentation it should reflect the type of formula we'd accept, but I'm happy to drop the rest if you like. Agreed on `gpg` for now though & dropped that.

Library/Homebrew/resource.rb
DomT4: Seems to work `¯\_(ツ)_/¯`

Library/Homebrew/software_spec.rb
DomT4: Dropped in favour of the delegators 👍.

Library/Homebrew/test/test_gpg.rb
DomT4: Done.

DomT4:
Changes pushed. Also added a block to the test & ensured the syntax tests are getting run on CI even if the GPG key creation test isn't.

Library/Homebrew/formula.rb
```diff
@@ -1810,6 +1819,17 @@ def mirror(val)
       define_method(type) { |val| stable.send(type, val) }
     end
 
+    # @!attribute [w] gpg
```
MikeMcQuaid: This is currently `rw` as you're accepting a `nil` val as the reader. Just checking this is intentional.

DomT4: Left over from earlier testing I think. Tweaked it.

Library/Homebrew/formula.rb
```diff
@@ -1810,6 +1818,17 @@ def mirror(val)
...
+    # An optional mechanism to verify the download with GnuPG as well as the
+    # standard checksum verification. Upstream must publish signature files
+    # for us to be able to use this mechanism. Those signature files often
+    # have either a ".gpg" or ".sig" file extension.
+    #
+    # <pre>gpg "https://ftpmirror.gnu.org/wget/wget-1.18.tar.xz.sig"</pre>
+    def gpg(val)
+      stable.gpg(val)
+    end
```
xu-cheng: My two cents is that we need to check the checksum of the signature file. 

So may I suggest a DSL like: `gpg "<url>", :sha256 => "checksum"`

MikeMcQuaid: I disagree with adding another checksum for this 👎

xu-cheng: Also, I'm interesting to know how do you indent to manage the trustiness of public key?

DomT4: A checksum is pretty redundant, IMO. If the signature changes `gpg --verify` breaks. If the tarball/zipball/whatever changes, `gpg --verify` breaks. I can't see an attack method that an extra SHA256 solves.

DomT4: > Also, I'm interesting to know how do you indent to manage the trustiness of public key?

Verification _works_ without having to know a key in advance, but I was planning to improve this, even if not immediately immediately. It took me about two weeks to reconfigure the project after initial feedback & test those changes, so I've had to shove things together less coherently than initially desired or planned 😕.

```ruby
gpg "https://blah.com/blah-1.2.3.sig", :key => "0x872524DB9D74326C"
```
Or such.

xu-cheng: > A checksum is pretty redundant, IMO.

I made a mistake here. You are right.

The real problem is I'm not sure how do you indent to manage the public keys? Because the most comment attack is replacing the source the signature at the same time

xu-cheng: > gpg "https://blah.com/blah-1.2.3.sig", :key => "0x872524DB9D74326C"

This is probably best way other than asking users to verify on their own. Also may I suggest to use full fingerprint instead of any shorter version of it.

DomT4: I'm not planning to declare it safe day one & pour champagne everywhere, to reassure you. Current mindset is that there won't (for now) be a command line flag for it and you'll have to set both `HOMEBREW_DEVELOPER` and `HOMEBREW_ENABLE_GPG` for it to ever come into play. 

That will likely remain the situation for at least a while until I'm happy to call it safe & robust.

> Also may I suggest to use full fingerprint instead of any shorter version of it.

Yup.

MikeMcQuaid:
Passing on this for now but will rereview a future PR with conflicts/problems addressed.

PR #787, opened 2016-08-22
sandbox: allow writing to TMPDIR, TEMP and TMP if set

zmwangx:

-----

These three variables could all affect tempfile creation when set.

### DIFF
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -63,6 +63,9 @@ class Sandbox
     allow_write "^/private/var/folders/[^/]+/[^/]+/[C,T]/", :type => :regex
     allow_write_path HOMEBREW_TEMP
     allow_write_path HOMEBREW_CACHE
+    allow_write_path ENV["TMPDIR"] if ENV["TMPDIR"]
+    allow_write_path ENV["TEMP"] if ENV["TEMP"]
+    allow_write_path ENV["TMP"] if ENV["TMP"]
   end
 
   def allow_write_cellar(formula)


zmwangx:
See https://github.com/Homebrew/homebrew-core/pull/4132 for a real world example of me getting a sandbox violation error when `autoreconf` attempts to write a tempfile due to having set `TMPDIR`, `TEMP` and `TMP`. (`autoreconf` at least recognizes `TMPDIR`, `clang` recognizes `TMP`, and Python for instance recognizes all three.)

apjanke:
Looks good to me.

This code doesn't check the values of `TMP`, `TMPDIR`, or `TEMPDIR` to see if they're sane, so a caller could set them to weird values to obtain access to files they "shouldn't". But I think that's fine, since our use of the sandbox is intended to enforce build hygiene, not to be a security measure against malicious actors, and the env variables are under user control. Going to wait to hear from other maintainers before merging, though.

chdiza:
Please, yes; I've had $TMPDIR set for years and now many of my builds are suddenly breaking.

In the meantime: is there a way to manually disable the sandbox?

MikeMcQuaid:
> But I think that's fine, since our use of the sandbox is intended to enforce build hygiene, not to be a security measure against malicious actors, and the env variables are under user control.

This is not the case, it's primarily intended to stop buildsystems writing to random locations they shouldn't.

Do either of you set HOMEBREW_TEMP? If so, would it work for you to have just that whitelisted?

MikeMcQuaid:
Oh, it already is. May I ask why you don't set that, out of interest?

chdiza:
I don't have HOMEBREW_TEMP set because I've never needed to, because I have TMPDIR set and that has always worked until sometime *very* recently.  I want my TMPDIR to be for everything, not just Homebrew.

But I will see whether leaving my TMPDIR the way it is and setting HOMEBREW_TEMP to point to the same place will work.

MikeMcQuaid:
My vague inkling is that Homebrew should set `TMPDIR`, `TEMP` and `TMP` to `HOMEBREW_TEMP`'s value.

apjanke:
> My vague inkling is that Homebrew should set TMPDIR, TEMP and TMP to HOMEBREW_TEMP's value.

That sounds right to me too.

Library/Homebrew/sandbox.rb
```diff
@@ -63,6 +63,9 @@ def allow_write_temp_and_cache
     allow_write "^/private/var/folders/[^/]+/[^/]+/[C,T]/", :type => :regex
     allow_write_path HOMEBREW_TEMP
     allow_write_path HOMEBREW_CACHE
+    allow_write_path ENV["TMPDIR"] if ENV["TMPDIR"]
+    allow_write_path ENV["TEMP"] if ENV["TEMP"]
+    allow_write_path ENV["TMP"] if ENV["TMP"]
```
xu-cheng: FYI, `allow_write_path` requires the path to be absolute or it will raise error.

zmwangx: `HOMEBREW_TEMP` and `HOMEBREW_CACHE` could also be set to a relative path or even random junk if a user wants to. IMO the user should be responsible for making these paths absolute (if set), because it doesn't make sense to have a relative `TMPDIR`.

zmwangx:
Sorry for replying late; I've been on vacation and didn't have Internet connection for the past day or two.

@MikeMcQuaid 

> May I ask why you don't set that, out of interest?

I used to have it set, but didn't bother anymore after repeatedly hitting `#{HOMEBREW_TEMP} is world-writable but does not have the sticky bit set.` (I know, I should be debugging my cron script which was responsible for setting the sticky bit after mounting, but the benefit offered by `HOMEBREW_TEMP` did not seem to be worth the trouble.)

> My vague inkling is that Homebrew should set TMPDIR, TEMP and TMP to HOMEBREW_TEMP's value.

That's a reasonable idea, so I've opened #800. I'm not closing this just yet though, since as mentioned in #800, it's not a perfect solution. Moreover,

1. `TMPDIR` and such need not be sticky, while `HOMEBREW_TEMP` has to, so people might not be able to point `HOMEBREW_TEMP` to their normal `TMPDIR`;
2. `HOMEBREW_TEMP` kind of fulfills a different and more specific role:

    ```
       HOMEBREW_TEMP
              If  set,  instructs  Homebrew  to  use HOMEBREW_TEMP as the temporary directory for building
              packages. This may be needed if your system temp directory and Homebrew Prefix are  on  dif-
              ferent  volumes, as OS X has trouble moving symlinks across volumes when the target does not
              yet exist.

              This issue typically occurs when using FileVault or custom SSD configurations.
```

@chdiza 

> In the meantime: is there a way to manually disable the sandbox?

`brew install --no-sandbox`.

**Comment by MikeMcQuaid** [comment]
> TMPDIR and such need not be sticky, while HOMEBREW_TEMP has to, so people might not be able to point HOMEBREW_TEMP to their normal TMPDIR;

I forget why this is, can you remember?

> HOMEBREW_TEMP kind of fulfills a different and more specific role

I think that's the typical use-case but we could overload it as it sounds like this is the right outcome if a different reason.

**Comment by zmwangx** [comment]
> > TMPDIR and such need not be sticky, while HOMEBREW_TEMP has to, so people might not be able to point HOMEBREW_TEMP to their normal TMPDIR;
>
> I forget why this is, can you remember?

If `HOMEBREW_TEMP` is not set problems could arise when installing things to keg, but under what circumstances I can't remember. Maybe @DomT4 could shed some light as he authored e4db0adf547d47b71c3feeb98daeedf8acb67274? (The commit links to several issues but none of them seem to explain why the parent directory needs to be sticky.)

> we could overload it

No problem with that.

**Comment by DomT4** [comment]
> Maybe @DomT4 could shed some light as he authored e4db0ad? 

Pretty easy to test;
```
sudo chown -t /private/tmp
brew install v
```
For example.
```
==> Verifying v-1.1.tar.gz checksum
tar xvzf /usr/local/var/homebrew/cache/v-1.1.tar.gz
x v-1.1/
x v-1.1/README
x v-1.1/v
x v-1.1/v.1
Error: parent directory is world writable but not sticky
Please report this bug:
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/tmpdir.rb:92:in `mktmpdir'
/usr/local/Library/Homebrew/utils/fork.rb:6:in `safe_fork'
/usr/local/Library/Homebrew/formula_installer.rb:593:in `build'
/usr/local/Library/Homebrew/formula_installer.rb:246:in `install'
/usr/local/Library/Homebrew/cmd/install.rb:272:in `install_formula'
/usr/local/Library/Homebrew/cmd/install.rb:150:in `block in install'
/usr/local/Library/Homebrew/cmd/install.rb:150:in `each'
/usr/local/Library/Homebrew/cmd/install.rb:150:in `install'
```

The block comes from [Ruby](http://ruby-doc.org/stdlib-2.3.1/libdoc/tmpdir/rdoc/Dir.html) rather than a Homebrew choice, as such.

zmwangx:
> The block comes from Ruby rather than a Homebrew choice, as such.

Yeah, that's it. Pretty annoying.

UniqMartin:
*Just as a side note:* One reason why `HOMEBREW_TEMP` defaults to `/tmp` instead of the contents of `TMPDIR` is that it is used to create UNIX domain sockets for IPC and paths for UNIX domain sockets are severely length-restricted, something that can become a problem with the typically already quite long `TMPDIR` on macOS.

MikeMcQuaid:
Passing on this in favour of https://github.com/Homebrew/brew/pull/800 for now but will reopen if we back out that approach.

PR #790, opened 2016-08-22
Move Cask’s vendored `plist` to `Homebrew/vendor`.

reitermarkus:


MikeMcQuaid:
Could you also add a description to the README: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/vendor/README.md? Thanks!

PR #793, opened 2016-08-23
Merge Cask’s `Pathname` extension with Homebrew’s.

reitermarkus:


Library/Homebrew/extend/pathname.rb
```diff
@@ -216,10 +216,10 @@ def cp_path_sub(pattern, replacement)
...
   # extended to support common double extensions
   def extname(path = to_s)
-    BOTTLE_EXTNAME_RX.match(path)
-    return $1 if $1
-    /(\.(tar|cpio|pax)\.(gz|bz2|lz|xz|Z))$/.match(path)
-    return $1 if $1
+    path =~ BOTTLE_EXTNAME_RX
+    return Regexp.last_match(1) if Regexp.last_match(1)
+    %r{(\.(dmg|tar|cpio|pax)\.(gz|bz2|lz|xz|Z|zip))$} =~ path
+    return Regexp.last_match(1) if Regexp.last_match(1)
```
MikeMcQuaid: I think it's alright to leave this stuff alone (unless Rubocop told you otherwise).

Library/Homebrew/extend/pathname.rb
```diff
@@ -216,10 +216,10 @@ def cp_path_sub(pattern, replacement)
 
   # extended to support common double extensions
   def extname(path = to_s)
-    BOTTLE_EXTNAME_RX.match(path)
-    return $1 if $1
-    /(\.(tar|cpio|pax)\.(gz|bz2|lz|xz|Z))$/.match(path)
-    return $1 if $1
+    path =~ BOTTLE_EXTNAME_RX
+    return Regexp.last_match(1) if Regexp.last_match(1)
+    %r{(\.(dmg|tar|cpio|pax)\.(gz|bz2|lz|xz|Z|zip))$} =~ path
```
MikeMcQuaid: I'd use `/` unless there's a `/` in the regex.

MikeMcQuaid:
A few small comments but nice work. 💀 to duplicated code! 

Something like https://pmd.github.io may be useful in finding more copy-pasted code.

MikeMcQuaid:
> Something like https://pmd.github.io may be useful in finding more copy-pasted code.

Specifically (CC @AnastasiaSulyagina):

```bash
$ brew install pmd
$ cd /usr/local
$ pmd cpd --minimum-tokens 20 --language ruby \
  --files Library/Homebrew --exclude Library/Homebrew/vendor\
   --exclude Library/Homebrew/cask/vendor --exclude Library/Homebrew/test \
   --exclude Library/Homebrew/cask/test --exclude Library/Homebrew/cask/spec
```

Library/Homebrew/extend/pathname.rb
reitermarkus: We'll need to merge our RuboCop configurations, so this will probably be a discussion on its own.

MikeMcQuaid: 👍

UniqMartin: I think both the old and the new code could be improved and made more understandable by dropping reliance on magic global state and doing something like:

```ruby
bottle_ext = path[BOTTLE_EXTNAME_RX, 1]
return bottle_ext if bottle_ext
archive_ext = path[/(\.(tar|cpio|pax)\.(gz|bz2|lz|xz|Z))$/, 1]
return archive_ext if archive_ext
```

UniqMartin: I'm also a bit curious why `zip` is included in that list imported from Cask. I wouldn't expect it to be used in this context, i.e. I've never seen `.tar.zip` or `.dmg.zip` files.

MikeMcQuaid:
:shipit: 🚢

Library/Homebrew/extend/pathname.rb
```diff
@@ -216,10 +216,10 @@ def cp_path_sub(pattern, replacement)
 
   # extended to support common double extensions
   def extname(path = to_s)
-    BOTTLE_EXTNAME_RX.match(path)
-    return $1 if $1
-    /(\.(tar|cpio|pax)\.(gz|bz2|lz|xz|Z))$/.match(path)
-    return $1 if $1
+    bottle_ext = path[BOTTLE_EXTNAME_RX, 1]
+    return bottle_ext if bottle_ext
+    archive_ext = path[/(\.(tar|cpio|pax)\.(gz|bz2|lz|xz|Z))$/, 1]
```
UniqMartin: Did you intentionally drop the `.dmg` and `.zip` extensions again? They were present in the Cask version and in your original version of this PR, but maybe they are indeed no longer needed.

reitermarkus: Yes, dropped them again since I added the `--name` flag to `gunzip`, which takes care of preserving the extension.

Issue #777, opened 2016-08-22
Disambiguate Tty


AnastasiaSulyagina:
Now we have two different Ttys here: [homebrew's](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils.rb#L14) and [cask's](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/lib/hbc/utils/tty.rb). They are quite different so it's probably worth discussing how to merge them. 
(Last time I accidentally applied Homebrew's Tty to Cask it seemed to be not very nice)

cc  @MikeMcQuaid @jawshooah @vitorgalvao @reitermarkus

reitermarkus:
Off the top of my head I can only remember `ohai` being white for some reason instead of bold black.

MikeMcQuaid:
I'm obviously biased but personally I think Cask should seek to use as much Homebrew code as possible and discuss global changes to both Homebrew and Cask when needed. There's going to be a lot of "duplicate but similar" code and I think it's worth being willing to make small changes to both projects now that they are one.

AnastasiaSulyagina:
Theoretically it would be nice to remove cask's Tty since as I understood it's older version of Homebrew's one but I would want to know what in HB's Tty is not ok for cask to fix it so created this issue

MikeMcQuaid:
@AnastasiaSulyagina I think (but @reitermarkus should feel free to contradict me) that it's worth making a PR to make Cask use only Homebrew's and then we can discuss/iterate on that.

reitermarkus:
Well, I like the meta-programming going on in Cask's version. Maybe we could completely rewrite it to have something like `Tty.format("Lorem ipsum.", color: :blue, style: :underline)`.

MikeMcQuaid:
I think the metaprogramming is harder to follow but I'm 👍 on rewriting once we have them combined. Doing both at once is a recipe for 💥

AnastasiaSulyagina:
So first - just remove cask's version?

MikeMcQuaid:
Aye, let's try that.

apjanke:
This might be easier to understand if you view Cask's `Tty` as a pretty complete implementation of ANSI terminal control sequences for color and styling, and Homebrew's `Tty` as an implementation of the specific styling that `brew` uses, along with just the necessary control sequences for those. (And what Homebrew's `Tty` calls `escape` is actually an entire control sequence, bracketed by CSI (which starts with an `Esc` character) and `m`.)

[...]

PR #794, opened 2016-08-23
cask tty removed

AnastasiaSulyagina:

-----
Removed cask's tty. ([#777](https://github.com/Homebrew/brew/issues/777))
HB's tty does not support double arguments, fixed them as I could, feel free to correct.

cc @MikeMcQuaid @jawshooah @vitorgalvao @reitermarkus


Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -87,7 +87,7 @@ def summary
     s = if MacOS.version >= :lion && !ENV["HOMEBREW_NO_EMOJI"]
           (ENV["HOMEBREW_INSTALL_BADGE"] || "\xf0\x9f\x8d\xba") + "  "
         else
-          "#{Hbc::Utils::Tty.blue.bold}==>#{Hbc::Utils::Tty.reset.bold} Success!#{Hbc::Utils::Tty.reset} "
+          "#{Tty.blue}==> Success!#{Tty.reset} "
```
reitermarkus: Only the `==>` should be blue here.

Library/Homebrew/cask/lib/hbc/utils.rb
```diff
@@ -151,14 +150,14 @@ def self.path_occupied?(path)
 
   def self.error_message_with_suggestions
     <<-EOS.undent
-    #{Hbc::Utils::Tty.reset.bold}
+    #
       Most likely, this means you have an outdated version of Homebrew-Cask. Please run:
 
-          #{Hbc::Utils::Tty.green.normal}#{UPDATE_CMD}
+          #{Tty.green}#{UPDATE_CMD}
```
reitermarkus: Missing `reset` here.

Library/Homebrew/cask/lib/hbc/utils.rb
```diff
@@ -151,14 +150,14 @@ def self.path_occupied?(path)
 
   def self.error_message_with_suggestions
     <<-EOS.undent
-    #{Hbc::Utils::Tty.reset.bold}
+    #
```
reitermarkus: #️⃣

MikeMcQuaid:
LGTM other than @reitermarkus's comments.

Library/Homebrew/cask/lib/hbc/installer.rb
AnastasiaSulyagina: `Tty.bold` is private in HB's Tty for some reasons. should I use Tty just on `==>`?

MikeMcQuaid: Let's go for just the `==>` as blue and bold and make the `Success` unbolded.

apjanke: I believe it's private just because it's not used on its own anywhere, and it's a "composing" implementation method that requires a color argument. Only the actually-used "style" methods are public. If a bold-but-unchanged-color style ends up being needed, I'd see no reason you couldn't rename the existing `Tty.bold` to `Tty.bold_and` and make a public `Tty.bold` method.

MikeMcQuaid: It's private basically because we don't want bold being used for styling outside of the existing colour palette.

UniqMartin: There's `Tty.highlight` (slightly more semantic name), that is currently defined as “bold using the foreground color”, so that would work as a replacement. It's used in `pretty_installed` that is used for highlighting installed formulae in update reports and `brew info`. Somewhat confusingly, there's also `Tty.white` with the exact same definition and that's what used in `oh1` and `ohai` for bolding the text after the arrow.

I suggest to use the latter for consistency with `oh1` and `ohai`.

reitermarkus:
The only thing I don't understand is why it's called `Tty.white`, I'm pretty sure `39` is `default`, not `white`.

AnastasiaSulyagina:
Confused me as well when I tried to replace `bold`

UniqMartin:
Probably the same reason why `Tty.em` isn't called `Tty.underline`: History and nobody going to the trouble of cleaning up this part of the code. `39` is indeed the default foreground color, which tends to be white (or some shade of it) when using a dark terminal background.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -87,7 +87,7 @@ def summary
     s = if MacOS.version >= :lion && !ENV["HOMEBREW_NO_EMOJI"]
           (ENV["HOMEBREW_INSTALL_BADGE"] || "\xf0\x9f\x8d\xba") + "  "
         else
-          "#{Hbc::Utils::Tty.blue.bold}==>#{Hbc::Utils::Tty.reset.bold} Success!#{Hbc::Utils::Tty.reset} "
+          "#{Tty.blue}==>#{Tty.reset} #{Tty.white}Success! #{Tty.reset}"
```
UniqMartin: Very minor nit, but I think the trailing space should remain after the final `#{Tty.reset}`, not moved before it.

reitermarkus:
> that tends to be white (or some shade of it) when using a dark terminal background.

Well, considering Terminal's default theme is black on white, this does make even less sense.

UniqMartin:
True, but I believe most maintainers and many seasoned command-line users prefer a dark background. I haven't been around long enough and haven't paid too much attention to defaults, but maybe the default Terminal theme also changed between Homebrew's creation and now?

reitermarkus:
Yes, I think the default changed from Mavericks to Yosemite.

MikeMcQuaid:
I use a white background and have for a while and it looks fine here.

MikeMcQuaid:
I'm happy with this as-is. @reitermarkus any final comments/objections to merging?

reitermarkus:
No, but let's not forget renaming `Tty.white` to something more fitting afterwards.

PR #797, opened 2016-08-23
gpg: troubleshooting fetch

DomT4:
Yanked it out of another branch. This is just the fetch changes that are refusing to work. Everything else is back in the other branch or syntax PR still.

Mostly just trying to work out why after trying to do this a bunch of different custom ways locally, moving to the patch method because we use it already, it works well & seems sane is completely broken for me. My brain has sort of given up on everything around this project by now, so please excuse anything obviously terrible or misunderstood.


MikeMcQuaid:
Closing this as we trouble-shooted it.

PR #800, opened 2016-08-24
config: set TMPDIR, TEMP and TMP to HOMEBREW_TEMP

zmwangx:

-----

`TMPDIR`, `TEMP` and `TMP` (when set) are not whitelisted for writing in `sandbox.rb`, which could result in sandbox violations when programs attempt to write to these locations.

Setting `TMPDIR`, `TEMP` and `TMP` to `HOMEBREW_TEMP` (which defaults to `/tmp` when not set) works around the aforementioned problem and also improves uniformity in the locations of tempfiles created during Homebrew operations.

Caveat: Non-matching `HOMEBREW_TEMP` and `TMPDIR` could lead to undesirable side effects in certain cases, e.g., emacsclient not being able to find an existing server (whose socket lives in `$TMPDIR/emacs$UID/`) when launched through brew edit.

---

See #787 for an alternative approach and discussions.

### DIFF
--- a/Library/Homebrew/config.rb
+++ b/Library/Homebrew/config.rb
@@ -35,6 +35,11 @@ HOMEBREW_LOGS = Pathname.new(ENV["HOMEBREW_LOGS"] || "~/Library/Logs/Homebrew/")
 # Must use /tmp instead of $TMPDIR because long paths break Unix domain sockets
 HOMEBREW_TEMP = Pathname.new(ENV.fetch("HOMEBREW_TEMP", "/tmp"))
 
+# Set common tmpdir environment variables to HOMEBREW_TEMP
+ENV["TMPDIR"] = HOMEBREW_TEMP
+ENV["TEMP"] = HOMEBREW_TEMP
+ENV["TMP"] = HOMEBREW_TEMP
+
 unless defined? HOMEBREW_LIBRARY_PATH
   # Root of the Homebrew code base
   HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent


MikeMcQuaid:
Seems reasonable to me but would probably want another 👍 here. Does this fix the sandbox issues for you?

zmwangx:
Yes, it does. But it's not perfect, as outlined in the caveat above and https://github.com/Homebrew/brew/pull/787#issuecomment-241979474.

UniqMartin:
That caveat sounds like it could be a major issue (or at least an annoyance). Do we know of more software that tends to put their sockets there or is Emacs an isolated case? Is there an easy way to change the socket path? Other than that this looks like a reasonable change to make.

zmwangx:
> Do we know of more software that tends to put their sockets there or is Emacs an isolated case?

I personally don't have many pieces of software that open sockets, and even fewer put their sockets in the global tmpdir. On top of head I can think of tmux, but with a bit of testing, tmux doesn't seem to care about `TMPDIR`, `TEMP` or `TMP`, at least not by default.

Actually, even if there are more software that does this, how many are going to be started inside brew's environment? Emacs is relevant here because `emacsclient` can be used as `VISUAL` or `EDITOR` which is used by `brew edit`.

> Is there an easy way to change the socket path?

One can set `server-socket-dir` to override the default inside one's init file, but I doubt many people know this; it's hardly documented in the GNU Emacs Manual.

Even if `server-socket-dir` is not set, the problem is not as severe as it may sound. `emacsclient` with the `-a ''` option can start a Emacs daemon if one isn't present already, and presumably many people take advantage of this (for instance, my `VISUAL` is a wrapper around `emacsclient -nw -cqta= "$@"`). In that case, not being able to find the existing socket in `TMPDIR` would simply mean another daemon is started with socket inside `HOMEBREW_TEMP`. Having two Emacs daemon might be devastating in the olden "Eight Megabytes And Constantly Swapping" days, but today I doubt anyone will give a damn, except in certain cases, like trying to switch to another frame but can't find it on the server.

---

In summary, I think Emacs is a rather isolated case, and we can document the workarounds somewhere if that's desirable.

MikeMcQuaid:
Let's give this a try but be prepared to back it out if things go 💥.

xu-cheng:
>Caveat: Non-matching HOMEBREW_TEMP and TMPDIR could lead to undesirable side effects in certain cases, e.g., emacsclient not being able to find an existing server (whose socket lives in $TMPDIR/emacs$UID/) when launched through brew edit.

Will it help, if we set these env when running `brew install/postinstall/test` (as in `Formula#run_install/run_postinstall/run_test`) instead of setting them iuncondiation for entire brew process.

zmwangx:
> Will it help, if we set these env when running brew install/postinstall/test (as in Formula#run_install/run_postinstall/run_test) instead of setting them iuncondiation for entire brew process.

That's probably fine.

UniqMartin:
👏 and many thanks @zmwangx for the detailed response!

dunn:
Yeah, this is biting me with `brew edit`.  Can we restrict it to install, postinstall, and test?

DomT4:
Agree with @xu-cheng's suggestion as well 👍.

dunn:
Sorry for not chiming in earlier, I only skimmed the thread and didn't realize how the issue would manifest.

zmwangx:
An attempt at restricting the changes to `install`, `post_install` and `test` is in #817.

Issue #796, opened 2016-08-23
Recent regression on sandboxed installs as root


stouset:
Up until recently, root-owned homebrew worked. There appears now to be an issue where the sandboxed brew install can no longer access TLS certificates when run as the `root` user.

```sh
$ sudo /opt/homebrew/bin/brew install pkg-config
==> Using the sandbox
==> Downloading https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.1.tar.gz

curl: (35) SSL certificate problem: Couldn't understand the server certificate format
Trying a mirror…
```

Compare to non-root-owned:

```sh
$ brew install --sandbox --build-from-source pkg-config
==> Using the sandbox
==> Downloading https://pkgconfig.freedesktop.org/releases/pkg-config-0.29.1.tar.gz
######################################################################## 100.0%
==> ./configure --prefix=/usr/local/Cellar/pkg-config/0.29.1_1 ...
```

MikeMcQuaid:
Thanks for spotting the underlying issue: that this is a `root` problem.

CC @xu-cheng for help with a fix here.

In the mean time: we/I strongly recommend not running Homebrew as root. We don't drop privileges so any random `Makefile` could `sudo rm -rf` any file on your machine. Arguably the sandbox is even more important for root-owned Homebrew so I don't think we should disable it there but I'm wondering whether we should support root-owned Homebrew at all until someone who cares about it implements dropping privileges.

stouset:
I'm trying to bisect to determine the exact location of the problem, but homebrew auto-updating is getting in my way :( It's at least more recent than ed5e2ea3003acc858e0886cb1c4a944f44ba598e

MikeMcQuaid:
It will be almost certainly https://github.com/Homebrew/brew/pull/713

stouset:
Ah, of course. Shouldn't this be fixable by allowing the sandbox to read TLS CAs?

MikeMcQuaid:
Probably. Feel free to beat me or @xu-cheng to a PR.

MikeMcQuaid:
It looks from https://github.com/Homebrew/brew/issues/785 that the sandbox is trying and failing to write to `/private/var/db/mds/system/mds.lock`, if that helps.

MikeMcQuaid:
It looks like the sandbox basically doesn't work as root.

MikeMcQuaid:
Does the build continue once it's been fetched as a non-root user?

stouset:
[...]

PR #805, opened 2016-08-24
Fixes https://github.com/Homebrew/brew/issues/796

xu-cheng:
Curl will request to write /private/var/db/mds/system/mds.lock in order
to access system SSL certs when running Homebrew under root user.

### DIFF
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -154,6 +154,8 @@ class Sandbox
           (literal "/dev/dtracehelper")
           (literal "/dev/null")
           (literal "/dev/zero")
+          ; required by curl for accessing SSL certs when running under root user
+          (literal "/private/var/db/mds/system/mds.lock")
           (regex #"^/dev/fd/[0-9]+$")
           (regex #"^/dev/ttys?[0-9]*$")
           )


UniqMartin:
Do we have some understanding why this lock file needs to be writable? Would be good to have some basic idea of what is going on. Is `mds` in this path referring to “metadata server”, i.e. the background service of the same name related to Spotlight?

xu-cheng:
> Is mds in this path referring to “metadata server”, i.e. the background service of the same name related to Spotlight?

It appears so. Also if you look at the other files, you will see many of them are mentioning security framework, or SSL cert related stuff. However, beyonding these I have zero knowledge on the internal mechanism underneath.

xu-cheng:
More google shows mds may be referred to Module Directory Services http://apple.stackexchange.com/a/167668

MikeMcQuaid:
I think it's worth disabling the sandbox as the root user for now, instead.

xu-cheng:
> I think it's worth disabling the sandbox as the root user for now, instead.

Since sandbox works fine after this fix  and it indeed offers some level of protection(e.g. against `rm -rf /`). I don't see any reason to disable sandbox for root users.

MikeMcQuaid:
@xu-cheng The plan is to disable root access altogether and we don't understand why this is needed.

xu-cheng:
I'm not against disable root access altogether in long term. However in short term, I don't think we should do the opposite to disable sandbox for the time being.

MikeMcQuaid:
@xu-cheng The sandbox has only been enabled for a short period for a single tap and doesn't work. Given we've decided by consensus to not support root usage in the medium term it doesn't make sense to add short-term workarounds to other code.

UniqMartin:
I'm not hyped about temporarily disabling the sandbox for root users until we eventually disallow using Homebrew as root, but given the sandbox has only been enabled very recently (and with limited reach) and that we don't really understand why we need to poke this hole in the sandbox or whether we need to poke more holes that we aren't aware of yet, disabling the sandbox again for the time being feels like the better option. Let's pass on this PR for now.

PR #807, opened 2016-08-25
audit: detect partial component order compliance

ilovezfs:
depends_on "foo"
conflicts_with "bar"
depends_on "baz"

should still detect that "bar" and "baz" are in the wrong order even
though "foo" and "bar" happen to be in the right order.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -72,7 +72,7 @@ module Homebrew
       fa.audit
 
       next if fa.problems.empty?
-
+      fa.problems
       formula_count += 1
       problem_count += fa.problems.size
       problem_lines = fa.problems.map { |p| "* #{p.chomp.gsub("\n", "\n    ")}" }
@@ -121,9 +121,14 @@ class FormulaText
     @text.include? s
   end
 
-  def line_number(regex)
-    index = @lines.index { |line| line =~ regex }
+  def line_number(regex, skip = 0)
+    index = @lines.drop(skip).index { |line| line =~ regex }
     index ? index + 1 : nil
+  end
+
+  def reverse_line_number(regex)
+    index = @lines.reverse.index { |line| line =~ regex }
+    index ? @lines.count - index : nil
   end
 end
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
See https://github.com/Homebrew/homebrew-core/pull/4221 for the motivation for this PR.

MikeMcQuaid:
Some `brew tests` failures here but I agree with the motivation 👍

ilovezfs:
@MikeMcQuaid it passes those tests now.

Library/Homebrew/cmd/audit.rb
```diff
@@ -121,10 +121,15 @@ def include?(s)
...
-  def line_number(regex)
-    index = @lines.index { |line| line =~ regex }
+  def line_number(regex, skip = 0)
+    index = @lines.drop(skip).index { |line| line =~ regex }
     index ? index + 1 : nil
   end
+
+  def reverse_line_number(regex)
+    index = @lines.reverse.index { |line| line =~ regex }
+    index ? @lines.count - index : nil
```
UniqMartin: The following should also do the trick and avoids building and immediately discarding a new array:

```ruby
index = @lines.rindex { |line| line =~ regex }
index ? index + 1 : nil
```

Library/Homebrew/cmd/audit.rb
```diff
@@ -72,7 +72,7 @@ def audit
       fa.audit
 
       next if fa.problems.empty?
-
+      fa.problems
```
UniqMartin: What does this do? This looks like a no-op to me.

Library/Homebrew/cmd/audit.rb
```diff
@@ -121,10 +121,15 @@ def include?(s)
     @text.include? s
   end
 
-  def line_number(regex)
-    index = @lines.index { |line| line =~ regex }
+  def line_number(regex, skip = 0)
+    index = @lines.drop(skip).index { |line| line =~ regex }
```
UniqMartin: The following will avoid constructing an intermediate array with `drop`:

```ruby
index = @lines.each_with_index.find_index { |line, index| index >= skip && line =~ regex }
```

(But if you prefer to stick with your version, it has a bug as `skip` needs to be added to the returned index.)

Library/Homebrew/cmd/audit.rb
ilovezfs: yeah that's dead code. it used to be fa.problems.uniq! which is no longer needed. thanks for spotting it.

Library/Homebrew/cmd/audit.rb
```diff
@@ -226,17 +210,67 @@ def audit_file
...
       [lineno, name]
     end.compact
+    no_problem = true
     present.each_cons(2) do |c1, c2|
-      unless c1[0] < c2[0]
-        problem "`#{c1[1]}` (line #{c1[0]}) should be put before `#{c2[1]}` (line #{c2[0]})"
+      if reverse
+        # scan in the forward direction from the offset
+        audit_components(false, [c1, c2]) if c1[0] > c2[0] # at least one more offense
+      elsif c1[0] > c2[0] && (offset == 0 || previous_pair.nil? || (c1[0] + offset) != previous_before[0] || (c2[0] + offset) != previous_after[0])
```
UniqMartin: I'm afraid I don't understand this condition. Can it be modified so that it is easier to parse and understand?

Library/Homebrew/cmd/audit.rb
ilovezfs: I add offset back in later which is why the code currently works, but absolute line number is probably better.

UniqMartin:
I also agree with the motivation for this PR, but I'm a tiny bit concerned with how invasive the change is. I pointed out a few minor issues, but if you can, I hope you can take another look and maybe simplify things a bit. My gut feeling is that it should be possible to simplify this, but feel free to tell me off until I have supplied evidence in the form of code. (I'm heading out, thus taking a closer look at the code will have to wait until later this evening.)

Library/Homebrew/cmd/audit.rb
ilovezfs: It should be simpler to understand when the offsets don't have to be added back in here, but it's just saying, current found problem != previous pair.

ilovezfs:
@UniqMartin there is no doubt in my mind that some other algorithm would be better than this putrid recursive contraption, but I thought that was too obvious and uncouth to be said aloud :D

MikeMcQuaid:
@UniqMartin If you have any suggestions here that'd be great. While I agree it's more complex I'm not too concerned personally with `audit` complexity. I think this highlights why we need an actual parser like RuboCop to handle these edge cases.

PR #812, opened 2016-08-25
formulary: handle @ formulae.

DomT4:

-----

Useful for https://github.com/Homebrew/brew/issues/620 & https://github.com/Homebrew/homebrew-core/pull/971.

### DIFF
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -83,7 +83,7 @@ module Homebrew
     end
 
     if $stdout.tty?
-      metacharacters = %w[\\ | ( ) [ ] { } ^ $ * + ? .]
+      metacharacters = %w[\\ | ( ) [ ] { } ^ $ * + ?]
       bad_regex = metacharacters.any? do |char|
         ARGV.any? do |arg|
           arg.include?(char) && !arg.start_with?("/")
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -57,6 +57,7 @@ class Formulary
     class_name = name.capitalize
     class_name.gsub!(/[-_.\s]([a-zA-Z0-9])/) { $1.upcase }
     class_name.tr!("+", "x")
+    class_name.gsub!(/\b@\b/, "AT")
     class_name
   end
 
--- a/Library/Homebrew/test/test_formulary.rb
+++ b/Library/Homebrew/test/test_formulary.rb
@@ -10,6 +10,7 @@ class FormularyTest < Homebrew::TestCase
     assert_equal "SLang", Formulary.class_s("s-lang")
     assert_equal "PkgConfig", Formulary.class_s("pkg-config")
     assert_equal "FooBar", Formulary.class_s("foo_bar")
+    assert_equal "OpensslAT11", Formulary.class_s("openssl@1.1")
   end
 end
 


Library/Homebrew/cmd/search.rb
```diff
@@ -86,7 +86,7 @@ def search
       metacharacters = %w[\\ | ( ) [ ] { } ^ $ * + ? .]
       bad_regex = metacharacters.any? do |char|
         ARGV.any? do |arg|
-          arg.include?(char) && !arg.start_with?("/")
+          arg.include?(char) && !arg.start_with?("/") && !arg.include?("@")
```
DomT4: Open to opinions on whether this is sane. It might be a bit broad.

MikeMcQuaid: Not sure what the intent is here?

MikeMcQuaid: Ah because of the `.`? Could consider just removing that.

Library/Homebrew/formulary.rb
```diff
@@ -57,6 +57,7 @@ def self.class_s(name)
     class_name = name.capitalize
     class_name.gsub!(/[-_.\s]([a-zA-Z0-9])/) { $1.upcase }
     class_name.tr!("+", "x")
+    class_name.gsub!("@", "AT")
```
MikeMcQuaid: May want to avoid matching at the beginning or end of the class_name

Library/Homebrew/cmd/search.rb
DomT4: Yeah, it interprets the `.` in something like `openssl@1.1` as a failed desire to regex.

Library/Homebrew/formulary.rb
DomT4: Something like:
```ruby
class_name.gsub!(/\b[@]\b/, "AT")
```
? Or is that too simple/complicated? 🙈

Library/Homebrew/cmd/search.rb
tdsmith: agree that `.` without a modifier isn't a strong regex signal; the current version makes sense 👍

Library/Homebrew/formulary.rb
```diff
@@ -57,6 +57,7 @@ def self.class_s(name)
     class_name = name.capitalize
     class_name.gsub!(/[-_.\s]([a-zA-Z0-9])/) { $1.upcase }
     class_name.tr!("+", "x")
+    class_name.gsub!(/\b[@]\b/, "AT")
```
MikeMcQuaid: Don't think you need the `[]`. Also, maybe can go for `/.@\d` to enforce a digit (for now at least, we can tweak it later). After that 👍 to shipping this.

DomT4: `/.@\d` matches `l@1` in this case, which produces things like:
```
Expected to find class OpenssAT1, but only found: OpensslAT11.
```
I'm not sure how to shake that off in the context of `gsub`.

DomT4:
Will merge this as-is for now, with the redundant `[]` fixed. Feel free to tell me how I was being an idiot around the `l@1` capture upon your return & I'll make any necessary changes 😄.

Library/Homebrew/formulary.rb
UniqMartin: Capture parts of the matched text and use them in the replacement:

```ruby
"openssl@1.1".gsub!(/(.)@(\d)/, "\\1AT\\2")
```

(And a `sub!` should be sufficient since we don't expect to have multiple `@` in a formula name for now.)

DomT4: Thanks! Pushed in the https://github.com/Homebrew/brew/pull/832 PR.

apjanke:
Use lookahead/lookbehind?

```
class_name.gsub!(/(?<=.)@(?=\d)/, "AT")
```

DomT4:
Changed in https://github.com/Homebrew/brew/pull/832 to `(/(.)@(\d)/, "\\1AT\\2")` per Martin's nudge. Don't have a strong preference either way, but if people prefer LA/LB I'm happy to change it again.

PR #817, opened 2016-08-26
Restrict TMPDIR etc. changes to install, post_install and test

zmwangx:

-----

#800 came with a caveat, which despite [my initial confidence](https://github.com/Homebrew/brew/pull/800#issue-172882122), actually [bit others too](https://github.com/Homebrew/brew/pull/800#issuecomment-242592802) (I'm saying "too" because I myself was certainly bitten, but didn't regard it as too big a deal).

In this PR we restrict `TMPDIR`, `TEMP` and `TMP` normalization to `install`, `post_install` and `test` and restore their original values afterwards, thus eliminating the caveat.

There's quite a bit of code duplication here unfortunately. I also have a DRY branch https://github.com/zmwangx/brew/commit/f5cc95c9967793e50a83c675f4e43419fde7e1a6 in case anyone's interested, but it involves a bit of global changes and it seems harder to restore the original values without further global hacks.

CC @xu-cheng (who proposed this) @dunn.

### DIFF
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -103,6 +103,11 @@ class Build
       end
     end
 
+    old_tmpdir = ENV["TMPDIR"]
+    old_temp = ENV["TEMP"]
+    old_tmp = ENV["TMP"]
+    ENV["TMPDIR"] = ENV["TEMP"] = ENV["TMP"] = HOMEBREW_TEMP
+
     formula.extend(Debrew::Formula) if ARGV.debug?
 
     formula.brew do |_formula, staging|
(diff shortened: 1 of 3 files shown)


Library/Homebrew/build.rb
```diff
@@ -103,6 +103,13 @@ def install
       end
     end
 
+    old_tmpdir = ENV["TMPDIR"]
+    old_temp = ENV["TEMP"]
+    old_tmp = ENV["TMP"]
+    ENV["TMPDIR"] = HOMEBREW_TEMP
+    ENV["TEMP"] = HOMEBREW_TEMP
+    ENV["TMP"] = HOMEBREW_TEMP
```
dunn: These can be chained in Ruby, I think? 
```ruby
ENV["TMPDIR"] = ENV["TEMP"] = ENV["TMP"] = HOMEBREW_TEMP
```
Might make the duplication a little less unpleasant.

zmwangx: Yeah, they can be chained, but AFAIK some people detest this style and I've seen style guides (in other languages) that explicitly advise against chaining assignments before. I guess Homebrew maintainers don't mind?

zmwangx: I've chained them up. Let's see if anyone has an objection.

MikeMcQuaid:
Nice work 👍

PR #818, opened 2016-08-26
Remove `exist?` check for `brew-cask-tests.rb`.

reitermarkus:


MikeMcQuaid:
Thanks for the cleanup ❤️

PR #819, opened 2016-08-26
Fix missing `magenta` in Tty.

reitermarkus:
This was overlooked in https://github.com/Homebrew/brew/pull/794.


MikeMcQuaid:
Merging because broken otherwise but would be cool to use the existing colour palette here?

reitermarkus:
What colour do you suggest using instead which would still differentiate `odebug` from `ohai`?

MikeMcQuaid:
Any of the others 😀, I don't have a strong preference. I do wonder whether 1) they need differentiated by colour and 2) it needs coloured at all (for debug text).

reitermarkus:
> 2) it needs coloured at all (for debug text)

I would argue that everything is better with colours. 😄

PR #823, opened 2016-08-26
Refactor Tty.

reitermarkus:
Restored some of the code from Cask's now removed `Tty` module and added a `Formatter` module to help with re-using predefined styles.


Library/Homebrew/cask/lib/hbc/cli/style.rb
```diff
@@ -28,7 +28,8 @@ def install_rubocop
       begin
         Homebrew.install_gem_setup_path! "rubocop-cask", RUBOCOP_CASK_VERSION, "rubocop"
       rescue SystemExit
-        raise Hbc::CaskError, $stderr.string.chomp.sub("#{Tty.red}Error#{Tty.reset}: ", "")
+        require "utils/tty"
+        raise Hbc::CaskError, $stderr.string.chomp.extend(Tty::StringExtension).strip_ansi.gsub("Error: ", "")
```
jawshooah: I think it's much cleaner to either make `strip_ansi` a module method or just monkeypatch `String`. It doesn't make sense to extend each individual string just to call this method.

reitermarkus: Reverted it to being a module method.

Library/Homebrew/utils.rb
```diff
@@ -10,102 +10,28 @@
...
     title = Tty.truncate(title)
   end
-  puts "#{Tty.green}==>#{Tty.white} #{title}#{Tty.reset}"
+  puts "#{Tty.green}==>#{Tty.reset} #{Tty.bold}#{title}#{Tty.reset}"
 end
 
 # Print a warning (do this rarely)
 def opoo(warning)
-  $stderr.puts "#{Tty.yellow}Warning#{Tty.reset}: #{warning}"
+  $stderr.puts "#{Tty.yellow.underline}Warnin#{Tty.reset.yellow}g:#{Tty.reset} #{warning}"
```
chdiza: Looks like the word "Warning" got broken up.

reitermarkus: This is intentional, so the `g` doesn't get underlined.

chdiza: If you don't mind me asking, why is underlining being added here where there was none before?

reitermarkus: It's not being added. Before, [`yellow`](https://github.com/Homebrew/brew/pull/823/files#diff-cef16ceb727f1d5a841e0295b083611fL36) was actually a shortcut to `yellow.underline`.

jawshooah: Wait, why are we underlining everything but the "g"?

reitermarkus: Because the underline would collide with the `g`. This is really a minor detail. 😄

jawshooah: That it is, just a weird one. But it makes sense now 👍

UniqMartin: > Because the underline would collide with the `g`. This is really a minor detail. 😄

I don't understand this. Would you mind explaining it in a little more detail? Thanks!

UniqMartin:
Thanks for taking a stab at this, but I think this goes into the wrong direction and further reinforces a not so good style. My personal opinion is that we should make the styling more semantic rather than hard-coding the style on every use (think of templates rather than direct formatting in word processing). Examples:

- `#{Tty.red.underline}Error#{Tty.reset}` should be `#{Tty.error}Error#{Tty.reset}` and the exact styling (currently red and underlined) an implementation detail of the `Tty.error` method.

  I think it would be nice to also style the following colon in this and similar cases (warning, …), but that's a different debate.

- `#{Tty.underline}URL#{Tty.reset}` (previously `#{Tty.em}URL#{Tty.reset}`) should be `#{Tty.url}URL#{Tty.reset}`.

Not only will this make styling more consistent, but it will also expose when we have too many different styles and the output becomes more colorful and varied than useful. If we can keep the number of `Tty.<predefined_style>` methods limited, this will come almost for free.

> *No description provided.*

I appreciate your work a lot, but I would be a bit happier if you could provide some description of your PRs. Mainly a why and some high-level overview would be nice to have, as those aren't immediately obvious from looking at the diff. This doesn't have to be long, but the length should be non-zero except for small self-explanatory or urgent fixes.

Library/Homebrew/utils.rb
reitermarkus: @UniqMartin, it's just that from a design point-of-view, the underline should not clash with the descenders of lowercase letters.

reitermarkus:
> we should make the styling more semantic

I agree, but speaking of semantics, formatting shouldn't be part of the `Tty` module.

Library/Homebrew/utils.rb
MikeMcQuaid: I think the underline should remain in that case or not be used in words with `g`s

MikeMcQuaid: As this is Homebrew's case: I think it should remain as it was.

MikeMcQuaid: (the previous version looks fine/good to me)

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,45 @@
+module Tty
+  module_function
+
+  def strip_ansi(str)
+    str.gsub(/\033\[\d+(;\d+)*m/, "")
```
MikeMcQuaid: `str` -> `string`

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def width
+    `/usr/bin/tput cols`.strip.to_i
+  end
+
+  def truncate(str)
+    (w = width).zero? ? str.to_s : str.to_s[0, w - 4]
```
MikeMcQuaid: `str` -> `string`, use a longer variable for `w`, use `if` over ternary, use a more self-documenting function for `[]`

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,45 @@
...
+    magenta:       35,
+    cyan:          36,
+    default:       39,
+
+    # Styles
+    reset:          0,
+    bold:           1,
+    italic:         3,
+    underline:      4,
+    strikethrough:  9,
```
MikeMcQuaid: Don't think it's worth aligning these.

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,45 @@
...
+  }.each do |name, code|
```
MikeMcQuaid: I'd consider declaring this as two frozen Hash constants and iterating over that; makes it a big more explicit what's going on.

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,45 @@
...
+    define_singleton_method(name) do
+      @sequence = [*@sequence, code]
+      self
+    end
+  end
+
+  def to_s
+    return "" if [*@sequence].empty? || !$stdout.tty?
```
MikeMcQuaid: Split into two `return`s

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,45 @@
...
+    "\033[#{@sequence.join(";")}m"
+  ensure
+    @sequence = []
```
MikeMcQuaid: This state feels kinda weird to me.

MikeMcQuaid:
> #{Tty.underline}URL#{Tty.reset} (previously #{Tty.em}URL#{Tty.reset}) should be #{Tty.url}URL#{Tty.reset}.

I like this.

> I appreciate your work a lot, but I would be a bit happier if you could provide some description of your PRs.

👍 


> I agree, but speaking of semantics, formatting shouldn't be part of the Tty module.

I think we can consider renaming the module.

The metaprogramming here is fine but I think it's worth being extremely explicit with variable naming and splitting statements onto more lines so it's more obvious to the casual, non-expert Rubyist what's going on here.

Library/Homebrew/cask/lib/hbc/artifact/moved.rb
```diff
@@ -79,10 +79,9 @@ def summarize_artifact(artifact_spec)
     if target.exist?
       target_abv = " (#{target.abv})"
     else
-      warning = "Missing #{self.class.artifact_english_name}"
-      warning = "#{Tty.red}#{warning}#{Tty.reset}: "
+      error = "#{Tty.red}Missing #{self.class.artifact_english_name}:#{Tty.reset} "
```
reitermarkus: Note to self: `Formatter.error`

reitermarkus:
@UniqMartin or @MikeMcQuaid, could someone take a look again?

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  COLOR_CODES = {
+                  red:     31,
+                  green:   32,
+                  yellow:  33,
+                  blue:    34,
+                  magenta: 35,
+                  cyan:    36,
+                  default: 39,
```
MikeMcQuaid: Tiny nit: just indent these at four spaces in and use hash rockets for now (we'll do a mass update 🔜)

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,51 @@
...
+                }
+
+  STYLE_CODES = {
+                  reset:          0,
```
MikeMcQuaid: Similarly here.

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,51 @@
...
+                  underline:      4,
+                  strikethrough:  9,
+                  no_underline:  24,
+                }
+
+  CODES = COLOR_CODES.merge(STYLE_CODES)
+
+  CODES.each do |name, code|
+    define_singleton_method(name) do
+      @sequence = [*@sequence, code]
```
MikeMcQuaid: This is the bit I'm not sure I understand what/why it's doing what it is.

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,51 @@
...
+      self
+    end
+  end
+
+  def to_s
+    return "" if [*@sequence].empty?
+    return "" unless $stdout.tty?
+    "\033[#{@sequence.join(";")}m"
```
MikeMcQuaid: Maybe worth making the `\033` a constant somewhere; used in a couple of places.

MikeMcQuaid:
Looking much better, thanks @reitermarkus.

Library/Homebrew/utils/tty.rb
reitermarkus: Since `@sequence` can be `nil` (initially), we simply use the splat operator to ensure that the end result is always an array which consists of the current `@sequence` followed by the `code`.

Library/Homebrew/utils/tty.rb
reitermarkus: Why *four* spaces, not two?

Library/Homebrew/utils/tty.rb
reitermarkus: The only other place this is used is in the tests. And I'd rather not use constants that are defined inside the to-be-tested class.

Library/Homebrew/utils/tty.rb
MikeMcQuaid: Just because it's in a module hence 2+2.

Library/Homebrew/utils/tty.rb
MikeMcQuaid: Sorry, higher-level: why do we need a `@sequence` variable now and didn't before?

Library/Homebrew/utils/tty.rb
MikeMcQuaid: Used in `strip_ansi` above?

Library/Homebrew/utils/tty.rb
reitermarkus: So we can use chaining like `Tty.red.underline`.

MikeMcQuaid: It feels like this might be done nicer by returning actual instances?

MikeMcQuaid: Or if not: can we split it into more methods to make what's going on more obvious.

Library/Homebrew/utils/tty.rb
reitermarkus: Ah, I missed that. However, how would I do this? The regex needs a literal `'\033\['`, while the `to_s` method needs the escaped `"\033\["`.

Library/Homebrew/utils/tty.rb
reitermarkus: I think returning actual instances would only complicate this. And what do you want to split into more methods here? There's only one `each` loop defining methods, how do you split an `each` loop?

Library/Homebrew/utils/formatter.rb
```diff
@@ -0,0 +1,57 @@
...
+    else
+      "#{Tty.send(color).underline}#{descenders_no_underline(label)}#{Tty.reset}#{message}"
+    end
+  end
+
+  def descenders_no_underline(string)
+    colon_reached = false
+
+    string.chars.map { |letter|
+      unless colon_reached
```
jawshooah: Was this left in by accident? Doesn't the first line of `colorize_text_before_colon` take care of this?

reitermarkus: Yes, good catch.

Library/Homebrew/utils/tty.rb
MikeMcQuaid: Ah I see, fine as-is then 👍

Library/Homebrew/utils/tty.rb
MikeMcQuaid: I don't know, really. I just find the current implementation confusing; it's not obvious that it's for changing and how the `@sequence` stuff works.

Some vague thoughts:
```ruby
@sequence ||= []
@sequence << code
self
```
signals intent a bit more. Even then, I'd be tempted to make a module-level method that is called `append_code` that encapsulates this logic so the `define_singleton_method` just ends up calling it. Similarly with `[*@sequence].empty?` below; having that be a separate module-level method would be good. Finally, using an attribute reader for `sequence` seems better than having raw access given how it's intended to be used.

MikeMcQuaid: I still think it would be good to split this into separate intent-signaling methods. I have my head around the `@sequence` stuff now but it's still a lot more complex to understand from the previous code. My preference is self-documenting code by splitting things into multiple methods that explain clearly what they do through their names. If you'd rather not do that: please add a lot more comments explaining what this is doing, how and why. Thanks!

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -92,7 +93,8 @@ def print_filename(string, filename)
       next unless ARGV.verbose? && !text_matches.empty?
       print_filename string, file
       text_matches.first(MAXIMUM_STRING_MATCHES).each do |match, offset|
-        puts " #{Tty.gray}-->#{Tty.reset} match '#{match}' at offset #{Tty.em}0x#{offset}#{Tty.reset}"
+        puts " #{Tty.gray}-->#{Tty.reset} match '#{match}' at offset #{Tty.underline}0x#{offset}#{Tty.reset}"
+
```
UniqMartin: This looks like a superfluous empty line to me.

Library/Homebrew/formula_installer.rb
```diff
@@ -212,7 +212,7 @@ def install
       opoo "#{formula.full_name}: #{old_flag} was deprecated; using #{new_flag} instead!"
     end
 
-    oh1 "Installing #{Tty.green}#{formula.full_name}#{Tty.reset}" if show_header?
+    oh1 "Installing #{Formatter.success(formula.full_name)}" if show_header?
```
UniqMartin: Using `Formatter.success` in this and several other contexts below where it encloses a formula/dependency name feels somewhat strange to me. Should there be a dedicated formatter for these cases?

Library/Homebrew/test/.simplecov
```diff
@@ -9,11 +9,11 @@ SimpleCov.start do
...
-  add_filter "/Homebrew/cask/spec/"
-  add_filter "/Homebrew/cask/test/"
-  add_filter "/Homebrew/compat/"
-  add_filter "/Homebrew/test/"
-  add_filter "/Homebrew/vendor/"
+  add_filter "/Library/Homebrew/cask/spec/"
+  add_filter "/Library/Homebrew/cask/test/"
+  add_filter "/Library/Homebrew/compat/"
+  add_filter "/Library/Homebrew/test/"
+  add_filter "/Library/Homebrew/vendor/"
```
UniqMartin: Why the change? In any case, this feels misplaced as this is completely unrelated to the changes in this PR (unless I'm missing something).

Library/Homebrew/dev-cmd/test-bot.rb
```diff
@@ -178,19 +178,19 @@ def puts_command
         puts "travis_fold:start:#{@travis_fold_id}"
         puts "travis_time:start:#{@travis_timer_id}"
       end
-      puts "#{Tty.blue}==>#{Tty.white} #{@command.join(" ")}#{Tty.reset}"
+      puts Formatter.headline(@command.join(" "), color: :blue)
```
UniqMartin: I don't feel too strongly about this, but my impression is we want to avoid introducing the new hash syntax until we update our RuboCop configuration and make a sweeping change, thus `:color => :blue` here and similarly in other places.

Library/Homebrew/utils/formatter.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    if label.nil?
+      "#{Tty.send(color)}#{message}#{Tty.reset}"
+    else
+      "#{Tty.send(color).underline}#{descenders_no_underline(label)}#{Tty.reset}#{message}"
+    end
+  end
+
+  def descenders_no_underline(string)
+    string.gsub(/([-+:jgqpy])/, "#{Tty.no_underline}\\1#{Tty.underline}")
```
UniqMartin: Why is this excluding `-`, `+`, and `:` that are clearly characters that stay above the baseline? Think about the `:` in `Formatter.url("http://example.com")`; not underlining this colon looks really weird.

Aside from that, I'm really against this kind of micro-optimization, not to mention that it only looks decent for letters where the descender spans the width of the entire character, e.g. `g` in most fonts, but looks like something is missing for letters like `p`. The problem with the underline is not that it touches descenders but the underline itself that is undesirable from a typographical point of view. Furthermore, this manual treatment will defeat proper treatment by the font and/or the font rendering of my terminal (assuming it did have some clever handling of underlines like a word processor often has).

Finally, check out e.g. the `Security` link on the bottom of every GitHub page. If you hover over it, the whole text is underlined (including the `y`). Like everyone else, we shouldn't bother special-casing this.

Practical suggestions:

- Underline all characters of a URL unconditionally.
- Underline all characters of an error/warning label unconditionally or simply avoid underlining if this feels like it looks too ugly (and yes, underlines are always ugly).

Library/Homebrew/utils/formatter.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def error(string)
+    colorize_text_before_colon(string, :red)
+  end
+
+  def url(string)
+    descenders_no_underline("#{Tty.blue.underline}#{string}#{Tty.reset}")
+  end
+
+  def colorize_text_before_colon(string, color)
```
UniqMartin: Just a nit pick, but this actually colorizes the text before the colon *and* the colon itself (which IMO is good from a stylistic point of view), so the method name isn't 100% spot-on.

Library/Homebrew/utils/tty.rb
```diff
@@ -0,0 +1,52 @@
...
+      self
+    end
+  end
+
+  def to_s
+    return "" if @sequence.nil?
+    return "" unless $stdout.tty?
+    "\033[#{@sequence.join(";")}m"
+  ensure
+    @sequence = nil
```
UniqMartin: Should we optimize for the next invocation and initialize this to an empty array instead? I guess the downside would be that this method would need to check for both `@sequence.nil?` and `@sequence.empty?` in its first line, though I that's probably only going to be the case for `"#{Tty}"` or `Tty.to_s` which would be bogus anyway, right?

Library/Homebrew/formula_installer.rb
reitermarkus: I agree. I only changed this so it would show up in the diff, so we could discuss a better name or format.

Library/Homebrew/test/.simplecov
reitermarkus: Yes, this is unrelated, somehow got in from working on the coverage PR.

Library/Homebrew/utils/formatter.rb
reitermarkus: `colorize_label`, maybe?

Library/Homebrew/utils/formatter.rb
reitermarkus: I can live with your practical suggestions. 👍

Library/Homebrew/utils/tty.rb
reitermarkus: That's exactly why I had a check for `[*@sequence].empty?` here before, which would have included both cases.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -70,6 +70,7 @@ def print_filename(string, filename)
         print_filename(string, file) unless linked_libraries.empty?
         linked_libraries.each do |lib|
           puts " #{Tty.gray}-->#{Tty.reset} links to #{lib}"
+
```
reitermarkus: All these empty lines are so the places where `Tty` is used instead of a `Formatter` show up in the diff. I wasn't sure if these warrant a `Formatter`, or what to name it.

Library/Homebrew/debrew.rb
```diff
@@ -111,7 +111,7 @@ def self.debug(e)
 
     begin
       puts "#{e.backtrace.first}"
-      puts "#{Tty.red}#{e.class.name}#{Tty.reset}: #{e}"
+      puts Formatter.error("#{e.class.name}: #{e}")
```
reitermarkus: I suppose we could do something like `Formatter.error(e, label: e.class.name)`.

UniqMartin: Judging from what I've seen in the code, `Formatter.error([<label>, ]<message>)` would be more practical, as I don't think it makes sense to omit “label” (or more precisely the thing that should be styled). This is how it would be used and what it would yield:

```ruby
Formatter.error("Custom label", "Message") # => "_Custom label:_ Message"
Formatter.error("Error message") # => "_Error:_ Error message"
```

Here, `_<whatever>_` denotes the styled part. Would that work?

Library/Homebrew/dev-cmd/bottle.rb
UniqMartin: I see; this makes sense. I didn't realize this and (if mentioned) overlooked it in previous comments. I think it's important to take our first few steps in the right direction (like with the error message and URL formatting), but we don't necessarily have to eliminate all direct uses of `Tty` in this PR, though we should consider doing so in the not too distant future. But as usual, naming is hard as is identifying the necessary styles while consolidating the probably too many different styles currently in use.

Library/Homebrew/formula_installer.rb
UniqMartin: How about `Formatter.identifier` or `Formatter.name`?

Library/Homebrew/utils/formatter.rb
UniqMartin: Yep, `colorize_label` or possibly a more opaque `style_label` would have worked. (I guess that method name is no longer needed given the changes you pushed since then.)

Library/Homebrew/utils/tty.rb
UniqMartin: I have to agree with Mike that I find the splatting and collecting a bit harder to follow. A possibly easier to read and still succinct alternative for the check could be:

```ruby
return "" if (@sequence || []).empty?
```

Library/Homebrew/brew.rb
```diff
@@ -144,8 +144,8 @@ def require?(path)
   Utils::Analytics.report_exception(e)
   onoe e
   if internal_cmd && defined?(OS::ISSUES_URL)
-    $stderr.puts "#{Tty.white}Please report this bug:"
-    $stderr.puts "    #{Tty.em}#{OS::ISSUES_URL}#{Tty.reset}"
+    $stderr.puts "Please report this bug:"
+    $stderr.puts "    #{Formatter.url(OS::ISSUES_URL)}"
```
UniqMartin: I think I'd like to preserve the boldface in the first line (`#{Tty.bold}Please report this bug:#{Tty.reset}`) and, since we're touching this anyway, adjust the indentation of the second line to two spaces which better aligns with most (if not all) similar user-visible output.

Library/Homebrew/dev-cmd/bottle.rb
UniqMartin: Also, in this specific case I would prefer to swap the underline for something else. `Tty.bold` would be fine by me, but I'm definitely open to alternative suggestions. It's just that the underline seriously and unnecessarily degrades legibility of those hex numbers.

Library/Homebrew/diagnostic.rb
```diff
@@ -477,7 +477,7 @@ def check_user_curlrc
           If you have trouble downloading packages with Homebrew, then maybe this
           is the problem? If the following command doesn't work, then try removing
           your curlrc:
-            curl https://github.com
+            curl #{Formatter.url("https://github.com")}
```
UniqMartin: I would personally not bother highlighting this URL, as it's a command line argument to be copy-pasted together with the `curl` rather than a URL we expect users to check out in isolation (most, if not all, other instances where `Formatter.url` is used). But I don't feel overly strongly about this.

Library/Homebrew/formula_installer.rb
```diff
@@ -421,7 +421,7 @@ def install_dependencies(deps)
     if deps.empty? && only_deps?
       puts "All dependencies for #{formula.full_name} are satisfied."
     elsif !deps.empty?
-      oh1 "Installing dependencies for #{formula.full_name}: #{Tty.green}#{deps.map(&:first)*", "}#{Tty.reset}",
+      oh1 "Installing dependencies for #{formula.full_name}: #{Formatter.success(deps.map(&:first)*", ")}",
```
UniqMartin: IMO, it would be nicer to highlight the individual elements and keep the delimiting comma and space unstyled instead of styling the entire string only after joining the array elements.

Library/Homebrew/utils/formatter.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def success(string, label: nil)
+    label(label, string, :green)
+  end
+
+  def warning(string, label: nil)
+    label(label, string, :yellow)
+  end
+
+  def error(string, label: nil)
```
UniqMartin: Sorry if I'm making you go back and forth a bit, but after giving this some more thought, I think the combined handling of label and message string just complicates things without providing sufficient benefit. I think it would be best to keep the concatenation of label and message string on the call site and let `Formatter.error` et al. just wrap the given string like other `Formatter` methods. I.e. it should be used as follows:

```ruby
Formatter.error("FAILED")
"#{Formatter.error("Error:")} #{message}"
"#{Formatter.error("READ THIS:")} #{Formatter.url("https://example.com/")}"
```

Library/Homebrew/utils/formatter.rb
```diff
@@ -0,0 +1,46 @@
...
+    label(label, string, :red)
+  end
+
+  def url(string)
+    "#{Tty.blue.underline}#{string}#{Tty.reset}"
```
UniqMartin: Would personally prefer to stick with the underline only (no color), but I don't feel very strongly about this.

Library/Homebrew/debrew.rb
UniqMartin: Please ignore my most recent comment; it doesn't make much sense after reviewing the code once more. https://github.com/Homebrew/brew/pull/823#r79279639 is a much saner suggestion.

Library/Homebrew/utils/formatter.rb
reitermarkus: I think this would be a step back, as this gets us to basically start *formatting* on the call side again, which kinda defeats the purpose of having a formatter in the first place.

UniqMartin: I kinda agree. How do you suggest we proceed? The following works for me:

```ruby
Formatter.error(message) # default label "Error"
Formatter.error(message, label: "READ THIS") # custom label
```

But I don't know how to deal with the case where there's neither a label nor a message, but just a string that should be styled like the error label. Should that be a separate method?

reitermarkus: Without a label, the message will be styled the same as the label

`Formatter.error("This string is red.)`

`Formatter.error("This message is not red.", label: "Red Label")`.

Library/Homebrew/utils/tty.rb
reitermarkus: I think we can leave this as-is. I don't even think *optimizing* would be worth it, since you will alway lose the last `[]`.

MikeMcQuaid:
@reitermarkus Can you try to avoid pushing quite as much? Every time you do there's a CI job that takes ~30m to run.

reitermarkus:
@MikeMcQuaid, sorry about that.

MikeMcQuaid:
@reitermarkus No worries, not to know 👍

Library/Homebrew/utils/tty.rb
MikeMcQuaid: I still think I'd really like pretty much all usage of `@sequence` to be in methods explaining what they do. I can take a stab at that if you don't mind me committing on this PR.

reitermarkus: > I can take a stab at that if you don't mind me committing on this PR.

Please do, I don't mind at all.

ilovezfs:
:-1:

Library/Homebrew/dev-cmd/test-bot.rb
MikeMcQuaid: > I don't feel too strongly about this, but my impression is we want to avoid introducing the new hash syntax until we update our RuboCop configuration and make a sweeping change

This was done.

Library/Homebrew/diagnostic.rb
MikeMcQuaid: I'm fine with highlighting it.

Library/Homebrew/utils/formatter.rb
MikeMcQuaid: I like the current version as-is.

Library/Homebrew/utils/formatter.rb
MikeMcQuaid: Actually, no, let's leave it colourless for now.

MikeMcQuaid:
@reitermarkus Sorry it took me so long to get to this. I've updated `utils/tty.rb` in a way that I think signals intent a bit more and should make it more obvious what's going on. If you're 👍 on that and address the other comments (basically now just: don't colour URLs) and fix the merge conflicts (😭) then I'm happy for this to 🚢

reitermarkus:
> fix the merge conflicts (:sob:) 

Totally looking forward to that! :joy:

reitermarkus:
Have looked over this on more time. Merging this now to avoid any more merge conflicts.

ilovezfs:
This seems to have broken CI.
https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/8588/version=el_capitan/console
See https://github.com/Homebrew/brew/issues/1209

PR #825, opened 2016-08-27
brew: execute update before tap

DomT4:

-----

This seems generally like a good idea given that we're making syntax changes to formulae & are going to keep doing so for a little while yet. Taps may have moved over to that syntax, which then causes tap failures if brew isn't up-to-date.

Should fix situations like https://github.com/Homebrew/homebrew-php/issues/3545.


MikeMcQuaid:
Yeh, seems like a good idea. :shipit:

DomT4:
Oh, forgot to tweak the manpage. Will fix that & then merge.

PR #828, opened 2016-08-27
mpi_requirement: update location of open-mpi

ilovezfs:
due to open-mpi migration to homebrew/science

### DIFF
--- a/Library/Homebrew/requirements/mpi_requirement.rb
+++ b/Library/Homebrew/requirements/mpi_requirement.rb
@@ -8,7 +8,7 @@ class MPIRequirement < Requirement
 
   fatal true
 
-  default_formula "open-mpi"
+  default_formula "homebrew/science/open-mpi"
 
   env :userpaths
 


DomT4:
We haven't actually allowed non-core `default_formula` requirements in the past, as a matter of policy. That's always been a deliberate choice. We've maintained previously requirements with a `default_formula` in a non-core tap needed to live in that tap.

I'd probably let this sit for a few days to let the team talk it through, since it means anyone anywhere using the `:mpi` requirement is going to end up with Science tapped, and that's quite a departure from things now.

ilovezfs:
>since it means anyone anywhere using the :mpi requirement is going to end up with Science tapped

I'm less worried about whoever that refers to than I am about the fact that a formula was upgraded in https://github.com/Homebrew/homebrew-core/pull/4154 breaking all of the following and leaving science to scramble to clean up the mess of all of the broken bottles:

```
Josephs-MacBook-Pro:~ joe$ brew uses open-mpi --include-build --include-optional
boost                             homebrew/science/getdp            homebrew/science/paraview       
bsponmpi                          homebrew/science/gmsh             homebrew/science/parmetis       
fftw                              homebrew/science/gromacs          homebrew/science/pastix         
konoha                            homebrew/science/hdf5             homebrew/science/petsc          
homebrew/boneyard/gasnet          homebrew/science/hyphy            homebrew/science/ray            
homebrew/python/h5py              homebrew/science/hypre            homebrew/science/scalapack      
homebrew/python/mpi4py            homebrew/science/lammps           homebrew/science/scotch         
homebrew/science/abinit           homebrew/science/lis              homebrew/science/scotch5        
homebrew/science/abyss            homebrew/science/maker            homebrew/science/slepc          
homebrew/science/alpscore         homebrew/science/meme             homebrew/science/sundials       
homebrew/science/arpack           homebrew/science/mfem             homebrew/science/superlu_dist   
homebrew/science/bless            homebrew/science/mrbayes          homebrew/science/trilinos       
homebrew/science/dealii           homebrew/science/mumps            homebrew/science/zoltan         
homebrew/science/dida             homebrew/science/nest             homebrew/versions/boost159      
homebrew/science/elemental        homebrew/science/neuron           homebrew/versions/boost160      
homebrew/science/exabayes         homebrew/science/nonpareil      
homebrew/science/garli            homebrew/science/p4est          
```

tdsmith:
The migration seems sensible; if we want to move the requirement out of core we could satisfy ourselves that it isn't used anywhere or put a depreciation message on it.

Did we discuss non-core default formulas before? It would be helpful to see those discussions. I don't think we've defined any policy as I understand the word though I'm happy to admit there might be a working consensus. ;) 

Thanks @ilovezfs for starting this discussion; I think this will help improve the robustness of the science tap.

DomT4:
> I don't think we've defined any policy as I understand the word though I'm happy to admit there might be a working consensus

For all intents and purposes this is all I've ever meant by policy. Even the things we write down are reorganised on the fly as necessary, such as SHA1, which is currently due to be hard blocked next month but likely won't be. My usage does meet one of the definitions of the word "policy" though, FWIW 😜.

You yourself discussing `numpy` made a slight virtual scrunched up face about using a non-core formula, way back in https://github.com/Homebrew/legacy-homebrew/pull/36810. I think you may have been the only one to make objection noises to it there, though.

There is precedent for kicking Requirements back to taps unless said requirement is going to be used by core formulae, see https://github.com/Homebrew/legacy-homebrew/pull/46448, sort of reinforced by planning to move the `osxfuse` requirement to `homebrew/fuse` until we decided to permit optional linkage in core formulae.

I agree upgrading `open-mpi` because it passed CI, despite the extreme majority of formulae that use it residing in `science`, is not exactly ideal. I'm unconvinced though that either `konoha` or `lcdf-typetools` can be called Science formulae, and currently they would be migrated in https://github.com/Homebrew/homebrew-core/pull/4274.

I think cross-tap `default_formula` is probably fine as long as we're aware either:

* Everything that uses that requirement by default or `:recommended` needs to live in the same tap as the `default_formula`, regardless of where those formulae would ideally live or:
* We essentially introduce cross-tap dependencies to any `:mpi`-using bottles, and potentially end up accidentally doing cross-tap CI testing every now & again.

Given where these formulae would be heading, probably not a bad idea to CC @Homebrew/science.

ilovezfs:
> I'm unconvinced though that either konoha or lcdf-typetools can be called Science formulae, and currently they would be migrated in Homebrew/homebrew-core#4274.

lcdf-typetools only because my understanding is conflicts_with breaks if the cross-tap is untapped, but actually now there appears to no longer be any conflict, so it can be removed from the PR.

Also, it looks like mpi can be made optional or removed entirely from konoha:
https://github.com/konoha-project/konoha3/blob/93e091e6f0970b0e58b6cbd61026999e434e95cb/src/package-devel/Lib.Mpi/CMakeLists.txt

DomT4:
I'd be fine with `konoha` being boneyarded, FWIW. Only 39 "known" downloads to date, upstream is dead & `make test` hasn't passed in years.

ilovezfs:
```
iMac-TMP:homebrew-core joe$ git diff -- Formula/konoha.rb 
diff --git a/Formula/konoha.rb b/Formula/konoha.rb
index ba93da4..35a2e3c 100644
--- a/Formula/konoha.rb
+++ b/Formula/konoha.rb
@@ -22,7 +22,7 @@ class Konoha < Formula
   deprecated_option "tests" => "with-test"
 
   depends_on "cmake" => :build
-  depends_on :mpi => [:cc, :cxx]
+  depends_on :mpi => [:cc, :cxx, :optional]
   depends_on "pcre"
   depends_on "json-c"
   depends_on "sqlite"
```
did the job if you'd like to keep it on life support.

ilovezfs:
```
98% tests passed, 15 tests failed out of 779
```
just :peach:y

DomT4:
Yeah, it's actually worse in `master` upstream, heh 🙈.

ilovezfs:
I've refreshed https://github.com/Homebrew/homebrew-core/pull/4274 with konoha mpi optional and the lcdf-tools conflicts_with yanked.

DomT4:
Cool 👍. Would like more opinions on the cross-tap `default_formula` thing here, and someone from Science to drop by and say they're okay accepting the formula even if you get hit by a truck & can no longer help maintain them. Those two conditions aside, I have no permanent blocker here.

dpo:
We're happy to maintain those formulae over at Science.

sjackman:
I'd prefer to see `open-mpi` in Homebrew/science.

MikeMcQuaid:
I'm vaguely concerned that things are migrating to science that are dependencies on non-science formulae or used by people outside the broad mission of homebrew-science. I'm thinking of Octave, R and Open-MPI here. 

As for the actual issue at hand: I think we should not use `default_formula` pointing to other taps. The Requirement itself should either be moved to that tap or we should keep the formula in homebrew-core.

ilovezfs:
okie dokie

MikeMcQuaid:
@ilovezfs Let's leave this open for discussion for now. Don't feel strongly enough to shut this down 😉

ilovezfs:
@MikeMcQuaid FWIW I actually agree with you. I just think what happened with the last open-mpi upgrade wasn't exactly OK with me :)

MikeMcQuaid:
@ilovezfs Yeh, we definitely have issues with cross-tap bottling that I'm unsure how we're going to be able to fix. Our versions support will help but I don't see us adding multiple versions for literally everything that a tap can depend on.

ilovezfs:
@MikeMcQuaid next bomb to drop on science: https://github.com/Homebrew/homebrew-core/pull/4341

dpo:
There's already a bug-fix release 2.0.1 of OpenMPI.

MikeMcQuaid:
Let's close this as no-one else chimed in and avoid the migration.

ilovezfs:
:+1:

Issue #738, opened 2016-08-17
Do tapped formulae need to use the full formula name for depends_on statements in that tap?


scpeters:
I maintain the [osrf/simulation](https://github.com/osrf/homebrew-simulation) tap, which has a few formula that depend on formulae from that tap. For example, [osrf/simulation/gazebo6](https://github.com/osrf/homebrew-simulation/blob/master/gazebo6.rb#L25) depends on [osrf/simulation/sdformat3](https://github.com/osrf/homebrew-simulation/blob/master/sdformat3.rb). My question is: does the `depends_on` statement have to use the fully qualified name or is the short name ok?

ie:

~~~
  depends_on "osrf/simulation/sdformat3"
~~~
vs
~~~
  depends_on "sdformat3"
~~~

I've been using the short name without problems for quite a while, but I recently noticed an issue with `test-bot` that I've traced to the use of the full name vs. the short name. I created a pull request to bump the formula revision of `sdformat3` and `gazebo6` ( https://github.com/osrf/homebrew-simulation/pull/129 ), and since gazebo6 depends on sdformat3, test-bot should build sdformat3 first. However, the [Homebrew::Test::formulae function in test-bot.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/test-bot.rb#L763-L782) is given the full names as arguments (`["osrf/simulation/gazebo6", "osrf/simulation/sdformat3"]`) but doesn't match any of those names in the output of `brew deps` since it gives short names, re-order them so that sdformat3 is built first. Instead they are built in alphabetical order.

Sorry if that's a bit long-winded. I could work around this issue by using full names in all my depends_on statements, but I'd prefer if test-bot was intended to handle that case. I'll work on a pull request if so.

scpeters:
To reproduce the error:

~~~
brew test-bot --tap=osrf/simulation --ci-pr https://github.com/osrf/homebrew-simulation/pull/129
~~~

scpeters:
I've made a small modification to test-bot in https://github.com/scpeters/brew/commit/d62fd50f8c6422e1f0a3c55cca1fafd5870f44b0 but it's not perfect. Instead of building `["osrf/simulation/gazebo6", "osrf/simulation/sdformat3"]` in alphabetical order, it builds `sdformat3` and then `["osrf/simulation/gazebo6", "osrf/simulation/sdformat3"]` in alphabetical order. So it builds sdformat3 twice, but at least it doesn't fail this time.

[...]

PR #830, opened 2016-08-28
Add `--full-name` option to `brew deps`

scpeters:

-----

This adds a `--full-name` option to the `brew deps` command so that tapped formulae in the dependencies list will be printed with their full name `owner/tap/formula`. This option is also used when running `brew test-bot` with multiple formulae changed, as described in #738.

For example:

~~~
$ brew deps homebrew/science/ceres-solver
cmake
eigen
gflags
glog
metis
suite-sparse
tbb
~~~

~~~
$ brew deps homebrew/science/ceres-solver --full-name
cmake
eigen
gflags
glog
homebrew/science/metis
homebrew/science/suite-sparse
tbb
~~~

### DIFF
--- a/Library/Homebrew/cmd/deps.rb
+++ b/Library/Homebrew/cmd/deps.rb
@@ -1,4 +1,4 @@
-#:  * `deps` [`--1`] [`-n`] [`--union`] [`--installed`] [`--include-build`] [`--include-optional`] [`--skip-recommended`] <formulae>:
+#:  * `deps` [`--1`] [`-n`] [`--union`] [`--full-name`] [`--installed`] [`--include-build`] [`--include-optional`] [`--skip-recommended`] <formulae>:
 #:    Show dependencies for <formulae>. When given multiple formula arguments,
 #:    show the intersection of dependencies for <formulae>.
 #:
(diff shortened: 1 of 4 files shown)


scpeters:
I didn't include a test case for the new option since `brew deps` doesn't have a test yet, and I wasn't quite up to that task. So the build is marked as a failure since coverage has decreased.

MikeMcQuaid:
Made a small change in https://github.com/Homebrew/brew/pull/830/commits/6f3039fcd3a2bb8ad7de52359a70f60aeac52ce8 but looks good to me and will merge once that's 💚.

MikeMcQuaid:
Actually one last question: can you use `time` to do some crude benchmarking of how the performance differs? Thanks!

scpeters:
Thanks for the improvements in 6f3039f.

For timing tests, I picked a core formula with many dependencies (`ffmpeg`, though most are optional) and repeated it a few times in a bash for loop:

~~~
$ time for i in `seq 10`; do brew deps --include-optional ffmpeg > /dev/null; done

real	0m6.541s
user	0m5.593s
sys	0m1.293s
~~~

With `--full-name`:

~~~
$ time for i in `seq 10`; do brew deps --include-optional ffmpeg --full-name > /dev/null; done

real	0m6.710s
user	0m5.750s
sys	0m1.312s
~~~

It shows about a 2-3% slowdown with `--full-name`.

MikeMcQuaid:
Seems fine, thanks!

scpeters:
thanks!

out of curiosity, how did you push to my fork? I didn't think I had added any external collaborators

PR #833, opened 2016-08-30
xcode: die unless Xcode/CLT up-to-date on prerelease macOS

DomT4:

-----
Not only have several maintainers suggested this is a good idea, including myself repeatedly, but at least a handful of users have said they wouldn't hate the idea so let's follow up better late than never.

Obvious test is obvious:
screen shot 2016-08-30 at 04 02 25

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -223,14 +223,15 @@ module Homebrew
 
   def check_development_tools
     checks = Diagnostic::Checks.new
-    checks.all_development_tools_checks.each do |check|
+    all_development_tools_checks = checks.development_tools_checks +
+                                   checks.fatal_development_tools_checks
+    all_development_tools_checks.each do |check|
       out = checks.send(check)
-      opoo out unless out.nil?
-    end
-    if OS.mac? && MacOS.prerelease?
-      checks.strict_development_tools_checks.each do |strict_check|
-        out = checks.send(strict_check)
-        odie out unless out.nil?
+      next if out.nil?
+      if checks.fatal_development_tools_checks.include?(check)
+        odie out
+      else
+        opoo out
       end
     end
   end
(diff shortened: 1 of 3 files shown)


DomT4:
Huh. That failure did not happen locally.

DomT4:
Oh, I see why. Can I wrap it in a simple `if OS.mac?` or are people going to insist I write a separate macOS compatibility layer for the `install` command? 🙈.

Library/Homebrew/cmd/install.rb
```diff
@@ -227,6 +227,12 @@ def check_development_tools
       out = checks.send(check)
       opoo out unless out.nil?
     end
+    if OS.mac? && MacOS.prerelease?
```
MikeMcQuaid: Pull this logic into `all_development_tools_checks` instead to avoid duplication and using `OS.mac?` here.

MikeMcQuaid:
One comment but otherwise agree with it.

> or are people going to insist I write a separate macOS compatibility layer for the install command? 🙈.

Usually yes but in this case you can just pull into diagnostic instead; I don't think we need a separate set of "strict" checks.

DomT4:
> Pull this logic into all_development_tools_checks instead to avoid duplication and using OS.mac? here.

Can you clarify this a bit?

I didn't lump the Xcode/CLT checks into `all_development_tools_checks` because `all_development_tools_checks` also includes a check as to whether you're running an unsupported version of macOS, so making that section a hard block would stop people running `install` on 10.5-10.8 and 10.12.

Either that or we throw the blob into `all_development_tools_checks` but don't make it fatal, retaining the `oppo` and hope people actually notice it/take it seriously.

I presume I'm misunderstanding how simple you want this, so thought I'd just ask 😅.

MikeMcQuaid:
I missed the `odie` difference 👍 

Just want to avoid have `if OS::Mac.prerelease?` or `if MacOS.` anywhere outside `Library/Homebrew/extend/os/mac` now. It would be good to instead move as much of this logic as possible to a method in `Library/Homebrew/extend/os/mac/diagnostic.rb` which could also do the `odie` and have a no-op in `diagnostic.rb`. e.g. have a `Diagnostic::Checks. check_development_tools` method that's a no-op for non-OS X or consider if this could/should live in `DevelopmentTools` instead.

PR #770, opened 2016-08-20
info: print requirements

ilovezfs:
Closes Homebrew/homebrew-core#4049.

PR #834, opened 2016-08-30
Audit/Testing: various tweaks

DomT4:

-----

Working on replacing myself with CI so I can ride off into the sunset/be more useful around `brew` rather than `core`. Here's some changes.

You probably hate the `test` one, let me know. The change itself has been discussed between @MikeMcQuaid & I for a while, but I'm not promising the implementation will be to everyone's admiration.

### DIFF
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -1253,6 +1253,17 @@ class ResourceAuditor
       end
     end
 
+    # Debian has an abundance of secure mirrors. Let's not pluck the insecure
+    # one out of the grab bag.
+    urls.each do |u|
+      next unless u =~ %r{^http://http\.debian\.net/debian/(.*)}i
+      problem <<-EOS.undent
+        Please use a secure mirror for Debian URLs.
+        We recommend:
+          https://mirrors.ocf.berkeley.edu/debian/#{$1}
+      EOS
+    end
+
     # Check for Google Code download urls, https:// is preferred
     # Intentionally not extending this to SVN repositories due to certificate
     # issues.
--- a/Library/Homebrew/cmd/audit.rb
+++ b/Library/Homebrew/cmd/audit.rb
@@ -1192,6 +1192,7 @@ class ResourceAuditor
            %r{^http://tools\.ietf\.org/},
            %r{^http://launchpad\.net/},
            %r{^http://bitbucket\.org/},
+           %r{^http://anonscm\.debian\.org/},
            %r{^http://cpan\.metacpan\.org/},
            %r{^http://hackage\.haskell\.org/},
            %r{^http://(?:[^/]*\.)?archive\.org},
@@ -1202,6 +1203,8 @@ class ResourceAuditor
         problem "#{p} should be `https://cpan.metacpan.org/#{$1}`"
       when %r{^(http|ftp)://ftp\.gnome\.org/pub/gnome/(.*)}i
         problem "#{p} should be `https://download.gnome.org/#{$2}`"
+      when %r{^git://anonscm\.debian\.org/users/(.*)}i
+        problem "#{p} should be `https://anonscm.debian.org/git/users/#{$1}`"
       end
     end
 


Library/Homebrew/cmd/audit.rb
```diff
@@ -1250,6 +1253,19 @@ def audit_urls
       end
     end
 
+    # Debian has an abundance of secure mirrors. Let's not pluck the insecure
+    # one out of the grab bag.
+    urls.each do |u|
+      next unless u =~ %r{^http://http\.debian\.net/debian/(.*)}i
+      problem <<-EOS.undent
+        Please use a secure mirror for Debian URLs.
+        We recommend either:
```
MikeMcQuaid: Let's just recommend 1️⃣

Library/Homebrew/test.rb
```diff
@@ -17,6 +17,15 @@
   ENV.extend(Stdenv)
   ENV.setup_build_environment
 
+  # This ensures all executables in test do are called with a full path, i.e.
+  # bin/"wget", or Formula["openssl"].opt_bin/"openssl" rather than just "wget".
+  # It's very likely to break tests, so phase it in via CI rather than user
```
MikeMcQuaid: Counterpoint: who is running tests who isn't CI 😀. I think it's worth CI and local/non-developer test behaviour being identical so it allows people to debug failing tests locally (remember the pain from sandbox tests; we enabled sandbox testing for everyone much earlier because of that). 👍 from me to just enabling this everywhere. CC @ilovezfs in case you have strong opinions as the two of you are the ones who will likely feel the fallout from this.

Library/Homebrew/cmd/audit.rb
ilovezfs: berkeley ftw

Library/Homebrew/test.rb
ilovezfs: I've never understood the logic behind the full path thing, but figured there was some esoteric reason for this rule that I had not yet come to fully appreciate.

Library/Homebrew/cmd/audit.rb
DomT4: I've been trying & failing to equal the load a bit so we're not completely whomping a single server. Berkeley has become more or less the primary URL where necessary, though, compared to `mirrorservice` which has remained primarily, well, a mirror.

MikeMcQuaid: Given it's a `mirror` I think it's fine to standardise on one. Regardless, those mirrors aren't hurting for bandwidth.

Library/Homebrew/test.rb
DomT4: `test` runs under `stdenv`, with the idea being that we test what the user has installed rather than our pristine happy environment. Also means there's no automatic shunting of `keg_only` dependencies into the `$PATH`. So for example, with `libressl`:

```
==> openssl dgst -sha256 -out checksum.txt testfile.txt
```
All we'd be doing there is testing the system OpenSSL works, which is nice, but not really what we want.

The start of my PATH as setup by my shell profile is:
```
/Users/xyz/perl5/bin:/usr/local/opt/openssl/bin:/usr/local/opt/curl/bin:/Users/xyz/.opam/system/bin:/usr/local/var/rbenv/shims:/usr/local/bin:/usr/local/sbin
```
Which means if I build a piece of software locally that happens to reside in anything prior to `/usr/local/bin` or `/usr/local/sbin` I'm not really testing the software intended. This can get particularly fun if you're testing stuff on 10.8 where `/usr/local/bin` wasn't yet dominant in the `$PATH`.

DomT4: > Counterpoint: who is running tests who isn't CI 😀.

Well, I'm not a robot 😜.

> 👍 from me to just enabling this everywhere

Happy to do this if ILZ is happy as well.

ilovezfs: Sounds like that concern only applies to things that are keg only.

DomT4: Not really. Also anything that could ever be duplicated in the `$PATH` earlier than `/usr/local/*`, which is well, limited only by user determination to break stuff.

ilovezfs: Yeah, I meant as far as CI goes because I don't really consider `test do` to be a user-facing part of brew.

DomT4: It's developer-facing, though, and developers tend to have fairly dirty `$PATH`s.

ilovezfs: That's not our job to police. I'd rather we left the test environment alone rather than tweaking it like this.

DomT4: I'd rather we have consistent testing. This is something we have to correct contributors on all the time, and I'm pretty sick of telling people to fix it, and I'm pretty sure people are sick of me telling them to fix it.

Whilst technically _"You only absolutely need to do x if x is `keg_only` or x tool you're trying to use from elsewhere in the test is `keg_only`"_ could be correct, that's a hell of a lot more complicated to explain to people & I don't see the upside over this method.

ilovezfs: Or we could stop chasing people around about this altogether and prepend the bin paths for keg only formulae and the sbin paths for all formula.

DomT4: You're basically dancing around proposing tests be run under `superenv`, which has been roundly rejected before.

ilovezfs: So you're agreeing that will work then?

ilovezfs: Also your change has the potential to create false negatives when the brew paths are excised such that the utility called ends up being something else (likely a system binary or a non-brew user binary with the same name), and the something else happens to pass the test. So had you not intervened, the test would have correctly failed with a true positive, but because it fell through to the other binary it "silently" passes with a false negative.

DomT4: I guess to avoid another debate where we chase our tails for many hours before someone finally tires of it, let's fast forward a bit. The existing system of wanting it a specific way but having no enforcement isn't ideal. Are you prepared to do the work to implement an alternative system that you prefer?

ilovezfs: I am perfectly happy with the status quo.

MikeMcQuaid:
:+1: 🚢

ilovezfs:
This can cause false negatives we currently do not get so I'm not sure why you'd want to ship it since it doesn't actually fix any existing bug anyone has pointed to.

DomT4:
Some of your concerns are valid and I'll think about them before shipping that aspect of the changes here. I had stuff to do yesterday so wasn't around.

I'll note though that if you want me to look at your comments on maintaining the core with more significance it would be helpful if you helped maintain more of the core; your own PRs are of a high standard & you throw a lot of work into contributing to those and keeping formulae up-to-date, which is fantastic. That makes you a valuable contributor, but it also somewhat limits how exposed you are to the ways other contributors see/change/use/abuse things.

You've got [issues](https://github.com/Homebrew/homebrew-core/issues/3168) specifically reopened that were then left for dead, tracking [issues](https://github.com/Homebrew/homebrew-core/issues/4029) which aren't [particularly](https://github.com/Homebrew/homebrew-core/issues/2184) complicated & could be resolved relatively swiftly, and there's stuff like https://github.com/Homebrew/brew/pull/770 which was a solid fix improving UX that needed just a little tweaking & you closed it citing _"lack of patience"_.

If you want to take over core stuff completely, please, feel free, but I find it quite grating till then to have my concerns about the status quo dismissed by someone who proposes changes, is asked to help implement them and in reply there's a vaguely non-committal shrug. You're entirely welcome to pick & choose what you work on, everyone does to some extent, but if you want to dismiss my concerns I'd appreciate it if you didn't leave those concerns on my plate forevermore.

Whilst I may be a total PITA to work with, I'm a total PITA who at least _attempts_ to propose or even implement alternatives if I shoot down someone else's fix for an issue that is bothering them.

ilovezfs:
I'm confident you'll realize at some point the above comment is not only gratuitous but also entirely inappropriate for this venue.

DomT4:
I'm going to pretty drastically cut back on the amount of core work I do, so I've dropped the test changes. Someone else can pick it up if they care.

PR #845, opened 2016-09-01
Don't reject -L/usr/local/lib when SDKROOT is defined

### LINKED ISSUE #844: -L/usr/local/lib incorrectly removed by superenv
In https://github.com/Homebrew/homebrew-fuse/issues/62, users report that -L/usr/local/lib is being removed from clang invocations by superenv, at least when Homebrew is not installed to /usr/local/lib.

This is incorrect because
* useful libraries (here, osxfuse) may exist in /usr/local/lib
* /usr/local/lib is removed from the default linker search path when SDKROOT is set

It seems like superenv should not remove -L/usr/local/lib from linker invocations when Homebrew is installed somewhere else.

Todo: figure out where this is happening and the rationale for it

tdsmith:
/usr/local/lib is removed from the default linker search path when
SDKROOT is defined or sysroot is specified.

Homebrew sometimes sets SDKROOT without setting a sysroot as of 45e138f.

This is the simplest possible change, but there are some other options:
* Modify effective_sysroot so that HOMEBREW_SDKROOT gets set identically; maybe @ilovezfs knows why this wasn't done originally?
* Add an additional accessor for SDKROOT
* ?

Fixes #844.

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -279,7 +279,9 @@ class Cmd
   end
 
   def system_library_paths
-    %W[#{sysroot}/usr/lib /usr/local/lib]
+    paths = ["#{sysroot}/usr/lib"]
+    paths << "/usr/local/lib" unless sysroot || ENV["SDKROOT"]
+    paths
   end
 
   def configure?


MikeMcQuaid:
Seems reasonable to me but I don't see any reference to the prefix in this or the referenced commit but this seems to only relate to a non-standard prefix, right? It feels like this is something that should vary depending on the prefix as `/usr/local/lib` is a system prefix now but we want it filtered if that's the HOMEBREW_PREFIX.

tdsmith:
It looks like that case is handled elsewhere; refurbish_args rejects linker search paths that are present in `library_paths` or `system_library_paths` and `HOMEBREW_PREFIX/lib` is added to `library_paths` by superenv.

MikeMcQuaid:
> refurbish_args rejects linker search paths that are present in library_paths or system_library_paths and HOMEBREW_PREFIX/lib is added to library_paths by superenv.

👍

ilovezfs:
I'm still not clear why /usr/local is being hard coded here. By default there's nothing there, so how is it a "special" path?

MikeMcQuaid:
@ilovezfs Probably because it's in the default search path for Clang.

ilovezfs:
Ah ok

Issue #831, opened 2016-08-28
brew info gives error on upgraded pinned formula


koallen:
**This issue is previously opened at `homebrew-core` by my mistake (Homebrew/homebrew-core#4301). Reopening here at `brew`.**

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

### Bug reports:

Previously, I pinned `gcc` since I don't wanna upgrade it. Later `r` is upgraded and that forced a upgrade to `gcc` (from `5.3.0` to `6.1.0`). After that, **I ran `brew cleanup`** to clean up old versions of packages. Just now I was trying to run `brew info gcc`, and I get an error saying `Error: /usr/local/Cellar/gcc/5.3.0 is not a directory`. This is similar to Homebrew/legacy-homebrew#46044.

I have to manually do `brew unpin gcc` and `brew pin gcc` so that I can get the correct info.

Is there some logical error in handling `brew info` with pinned packages?

**PS: The bug is reproduced by @ilovezfs in Homebrew/homebrew-core#4301.**

MikeMcQuaid:
Should be fixed, I believe, by https://github.com/Homebrew/brew/issues/846

PR #846, opened 2016-09-01
formula: pinned keg is not eligible for cleanup

vladshablinsky:
- [WIP] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).

-----

If formula is pinned it shouldn't be eligible for cleanup.
Otherwise we can encounter the following behaviour:

1. Install <formula>
2. Pin <formula>
3. <formula> gets outdated
4. Install <other_formula> that depends on <formula>, which requires
up-to-date installation of <formula>. Thus, <formula> gets updated.
5. Now, if we cleanup <formula> we don't remove pin, so it points to
the keg which doesn't exist any longer. The right behaviour is not
to cleanup pinned kegs, which is done in this commit.

Also see #831 for further info.

=======

How does this look?
Also, I can come up with different behaviour.

E.g. we can _warn and not allow to install formula_ if it depends on outdated formula that is pinned. Or we can repin pinned formula if it eventually gets upgraded by its dependent.

CC @ilovezfs.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1611,6 +1611,8 @@ class Formula
         eligible_kegs.each do |keg|
           if keg.linked?
             opoo "Skipping (old) #{keg} due to it being linked"
+          elsif pinned? && keg == Keg.new(@pin.path.resolved_path)
+            opoo "Skipping (old) #{keg} due to it being pinned"
           else
             eligible_for_cleanup << keg
           end


vladshablinsky:
Will add a test later, when desired behaviour will be chosen.

ilovezfs:
@vladshablinsky this looks good. I do think we should (re-)consider whether the "involuntary" upgrades should be allowed at all when the user has explicitly pinned the formula, though.

vladshablinsky:
> I do think we should (re-)consider whether the "involuntary" upgrades should be allowed at all when the user has explicitly pinned the formula, though.

Agreed. I addressed something similar in the description to the PR. Should I change it this way?

ilovezfs:
@vladshablinsky let's see what @MikeMcQuaid thinks :)

MikeMcQuaid:
Nice work @vladshablinsky!

MikeMcQuaid:
> e.g. we can warn and not allow to install formula if it depends on outdated formula that is pinned. Or we can repin pinned formula if it eventually gets upgraded by its dependent.

Refusing to upgrade either package feels like the best fit here 👍

ilovezfs:
@MikeMcQuaid Agreed. I think that will be the least surprising behavior to users.

Issue #567, opened 2016-07-21
Make `brew install $ALIAS` follow alias changes


MikeMcQuaid:
If `brew install boost` is an alias to `brew install boost123` then if the `boost` alias changes from `boost123` to `boost124` then `brew upgrade` should upgrade between these versions.

This will allow us to use the Homebrew versions-style formula naming (or an alternative like @UniqMartin has suggested e.g. `boost+1.2.3`) combined with these aliases to allow us to have formula pinned to versions if they need to be and have users (and some formulae) follow the latest version of e.g. `boost` if they don't break on newer versions.

When this issue is done we'll have a larger discussion about a new approach to versioning in Homebrew.

scpeters:
Sounds good to me.

penman:
Would this be a reasonable way to do this?

1. Persist the alias used to install a formula in the INSTALL_RECEIPT
2. When upgrading, if an alias is in the INSTALL_RECEIPT and the target of the alias has changed, install the new formulae the alias refers to (and store the alias in its INSTALL_RECEIPT too).

MikeMcQuaid:
@penman Yep, that's the plan! Please feel free to beat me to a PR though 😀

penman:
If somebody had run `brew install boost` and got `boost+1.2.3`, then boost 1.2.4 was released they ran `brew upgrade boost`, they'd get `boost+1.2.4` installed.

But what if they run `brew upgrade boost+1.2.3`? Seems to me that that should ignore that it was originally installed with just `boost` and install the latest version of the `boost+1.2.3` formula, even if `boost` has been re-aliased to `boost+1.2.4`. If that's the case, should that re-write the INSTALL_RECEIPT so that future `brew upgrade`s also stick to `boost+1.2.3`, or leave it alone?

MikeMcQuaid:
> If somebody had run brew install boost and got boost+1.2.3, then boost 1.2.4 was released they ran brew upgrade boost, they'd get boost+1.2.4 installed.

Yep 👍 

> But what if they run brew upgrade boost+1.2.3? Seems to me that that should ignore that it was originally installed with just boost and install the latest version of the boost+1.2.3 formula, even if boost has been re-aliased to boost+1.2.4. If that's the case, should that re-write the INSTALL_RECEIPT so that future brew upgrades also stick to boost+1.2.3, or leave it alone?

[...]

PR #863, opened 2016-09-03
Save aliases in INSTALL_RECEIPT

penman:

-----

This is part of #567 (but not all of it ⏤ I wanted to PR this bit so that I could get feedback since I haven't worked with this codebase before).

This PR adds an `install_name` key to INSTALL_RECEIPT. Usually this will just be set to the name of the formula, but if it was installed as an alias, the `install_name` will be set to the alias instead.

Implementation-wise, I wasn't sure about how best to set the `install_name` on a `Formula` instance (which seemed like the only appropriate place to store it). I thought the best way would probably be a keyword argument, but it could also be a standard optional argument, or an `attr_accessor`. I'm open to suggestions if there's a better way.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -226,6 +226,11 @@ class Formula
 
   public
 
+  # The path that was specified to find/install this formula.
+  def specified_path
+    alias_path || path
+  end
+
   # Is the currently active {SoftwareSpec} a {#stable} build?
   # @private
   def stable?
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -58,6 +58,11 @@ class Formula
   # e.g. `this-formula`
   attr_reader :name
 
+  # The name specified when installing this {Formula}.
+  # Could be the name of the {Formula}, or an alias.
+  # e.g. `another-name-for-this-formula`
+  attr_reader :install_name
+
   # The fully-qualified name of this {Formula}.
   # For core formula it's the same as {#name}.
   # e.g. `homebrew/tap-name/this-formula`
@@ -143,9 +148,10 @@ class Formula
   attr_accessor :build
 
   # @private
-  def initialize(name, path, spec)
+  def initialize(name, path, spec, install_name: name)
     @name = name
     @path = path
+    @install_name = install_name
     @revision = self.class.revision || 0
     @version_scheme = self.class.version_scheme || 0
 
@@ -500,6 +506,21 @@ class Formula
   # @private
   def installed_kegs
     installed_prefixes.map { |dir| Keg.new(dir) }
+  end
+
+  # Formula reference to use to get an identical installation of the formula.
+  #
+  # Usually, the formula's path is a good canonical identifier for the formula.
+  # Just using whatever was passed to `brew install` isn't a good idea, because
+  # if it was e.g. a URL, it could end up being downloaded twice.
+  #
+  # However, if the formula was installed with an alias (i.e. `install_name` is
+  # different from `name`), that should be used instead so that information is
+  # preserved. Aliases are looked up in repositories that have already been
+  # tapped, so we don't have to worry about doing extra HTTP requests, or other
+  # expensive operations, when looking them up again.
+  def install_ref
+    install_name == name ? path : install_name
   end
 
   # The directory where the formula's binaries should be installed.
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -61,7 +61,7 @@ class Formula
   # The name specified when installing this {Formula}.
   # Could be the name of the {Formula}, or an alias.
   # e.g. `another-name-for-this-formula`
-  attr_reader :install_name
+  attr_reader :alias_path
 
   # The fully-qualified name of this {Formula}.
   # For core formula it's the same as {#name}.
@@ -148,10 +148,10 @@ class Formula
   attr_accessor :build
 
   # @private
-  def initialize(name, path, spec, install_name: name)
+  def initialize(name, path, spec, alias_path: nil)
     @name = name
     @path = path
-    @install_name = install_name
+    @alias_path = alias_path
     @revision = self.class.revision || 0
     @version_scheme = self.class.version_scheme || 0
 
@@ -506,21 +506,6 @@ class Formula
   # @private
   def installed_kegs
     installed_prefixes.map { |dir| Keg.new(dir) }
-  end
-
-  # Formula reference to use to get an identical installation of the formula.
-  #
-  # Usually, the formula's path is a good canonical identifier for the formula.
-  # Just using whatever was passed to `brew install` isn't a good idea, because
-  # if it was e.g. a URL, it could end up being downloaded twice.
-  #
-  # However, if the formula was installed with an alias (i.e. `install_name` is
-  # different from `name`), that should be used instead so that information is
-  # preserved. Aliases are looked up in repositories that have already been
-  # tapped, so we don't have to worry about doing extra HTTP requests, or other
-  # expensive operations, when looking them up again.
-  def install_ref
-    install_name == name ? path : install_name
   end
 
   # The directory where the formula's binaries should be installed.
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -589,7 +589,7 @@ class FormulaInstaller
       -I #{HOMEBREW_LOAD_PATH}
       --
       #{HOMEBREW_LIBRARY_PATH}/build.rb
-      #{formula.path}
+      #{formula.install_ref}
     ].concat(build_argv)
 
     Sandbox.print_sandbox_message if Sandbox.formula?(formula)
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -589,7 +589,7 @@ class FormulaInstaller
       -I #{HOMEBREW_LOAD_PATH}
       --
       #{HOMEBREW_LIBRARY_PATH}/build.rb
-      #{formula.alias_path || formula.path}
+      #{formula.specified_path}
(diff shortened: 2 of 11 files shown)


penman:
Looks like that CI failure isn't anything to do with my changes? I ran `brew test-bot` just to make sure and everything passed.

MikeMcQuaid:
> Looks like that CI failure isn't anything to do with my changes? I ran brew test-bot locally just to make sure and everything passed.

Yeh, the Travis build has been flaky this weekend. The `default` is the main one we care about.

Library/Homebrew/formula.rb
```diff
@@ -143,9 +148,10 @@ class Formula
   attr_accessor :build
 
   # @private
-  def initialize(name, path, spec)
+  def initialize(name, path, spec, install_name: name)
```
MikeMcQuaid: We're still mostly using hash rockets until we mass change them at some later point. Were still on Ruby 1.8 until recently!

Library/Homebrew/formula.rb
```diff
@@ -502,6 +508,21 @@ def installed_kegs
...
+  # Just using whatever was passed to `brew install` isn't a good idea, because
+  # if it was e.g. a URL, it could end up being downloaded twice.
+  #
+  # However, if the formula was installed with an alias (i.e. `install_name` is
+  # different from `name`), that should be used instead so that information is
+  # preserved. Aliases are looked up in repositories that have already been
+  # tapped, so we don't have to worry about doing extra HTTP requests, or other
+  # expensive operations, when looking them up again.
+  def install_ref
+    install_name == name ? path : install_name
```
MikeMcQuaid: This will probably need to be a fully-qualified path.

MikeMcQuaid: or alternatively you could return `name` or `install_name` rather than `path` and build the fully-qualified one where it's needed?

Library/Homebrew/formulary.rb
```diff
@@ -68,15 +68,18 @@ class FormulaLoader
...
     def initialize(name, path)
       @name = name
       @path = path.resolved_path
+      @install_name = name
     end
 
     # Gets the formula instance.
     def get_formula(spec)
-      klass.new(name, path, spec)
+      klass.new(name, path, spec, install_name: install_name)
```
MikeMcQuaid: Similarly here wrt hash rockets.

Library/Homebrew/formulary.rb
```diff
@@ -118,6 +121,7 @@ def initialize(alias_path)
       path = alias_path.resolved_path
       name = path.basename(".rb").to_s
       super name, path
+      @install_name = alias_path.basename.to_s
```
MikeMcQuaid: I was wondering if things would 💥 if we used the `@install_name` as the actual value for `name` instead? That means we could avoid having to add new fields. Thoughts?

MikeMcQuaid:
Some comments on this but liking where this is going so far and nice work creating this PR!

Library/Homebrew/formulary.rb
penman: I was afraid to try. :P

I imagine `Formula#name` is used pretty widely across Homebrew. This would, for example, change the install prefix for a formula based on the alias used to install it, which doesn't seem like a great idea to me.

MikeMcQuaid: A good argument 👍

MikeMcQuaid: Could maybe be something like `unresolved_name` or something?

MikeMcQuaid: (naming is hard)

penman: `unresolved_name` would imply to me that it could be a URL, for a file path, which currently it'll never be…

Library/Homebrew/formula.rb
penman: Doesn't look like it's possible to use hash rockets to declare a keyword argument.

Library/Homebrew/formulary.rb
MikeMcQuaid: `alias_name`?

MikeMcQuaid: or, if we went with the full path for above, `alias_path`

Library/Homebrew/formula.rb
penman: (I'll change the calls to `new` to use hash rockets, though.)

MikeMcQuaid: Ah, my mistake, missed this was a kwarg. Fine as-is then 👍

Library/Homebrew/formula.rb
penman: Didn't occur to me that I could return the path to the alias, but it's obvious in hindsight!

Library/Homebrew/formulary.rb
penman: How about we go for `alias_path`, and make it return `nil` rather than the formula path if an alias wasn't used?

Then we could use `formula.alias_path || formula.path` when passing a reference to it, and the meaning would be clearer all round.

MikeMcQuaid: Sounds good 👍.

Library/Homebrew/formula_installer.rb
```diff
@@ -589,7 +589,7 @@ def build
       -I #{HOMEBREW_LOAD_PATH}
       --
       #{HOMEBREW_LIBRARY_PATH}/build.rb
-      #{formula.path}
+      #{formula.alias_path || formula.path}
```
MikeMcQuaid: Think this logic may be worth pulling into a formula method that just does this `||`. `alias_path_or_path` or something with a better name?

Library/Homebrew/tab.rb
```diff
@@ -292,6 +298,7 @@ def to_json
       "time" => time,
       "source_modified_time" => source_modified_time.to_i,
       "HEAD" => self.HEAD,
+      "alias_path" => alias_path && alias_path.to_s,
```
MikeMcQuaid: Probably worth doing `(alias_path.to_s if alias_path)` just to match style below.

MikeMcQuaid:
Nice work today ✨

Library/Homebrew/formula_installer.rb
penman: I thought about this. My reasoning for leaving it as-is is that having a method called `alias_path_or_path` wouldn't really be reducing repetition at all. It just occurred to me that `specified_path` would work as a name, though, so I'll move it into a method.

Library/Homebrew/tab.rb
```diff
@@ -161,6 +166,7 @@ def self.empty
       "time" => nil,
       "source_modified_time" => 0,
       "HEAD" => nil,
+      "alias_path" => nil,
```
MikeMcQuaid: I'd maybe put this in `source` below and perhaps consider overriding `path` there unless it would break existing functionality relying on this in the tab.

Library/Homebrew/tab.rb
```diff
@@ -292,6 +298,7 @@ def to_json
       "time" => time,
       "source_modified_time" => source_modified_time.to_i,
       "HEAD" => self.HEAD,
+      "alias_path" => (alias_path.to_s if alias_path),
```
MikeMcQuaid: Same here.

MikeMcQuaid:
A couple more tiny nits but I ❤️ how this is looking and when you're testing is 🆗 then I'd be happy to merge this.

Library/Homebrew/tab.rb
penman: As far as I can tell nothing uses source.path, so this should be 👍

penman:
I've updated the PR based on your feedback, @MikeMcQuaid.

It also looked to me like putting data into a Tab isn't actually tested anywhere, so I added `test_create` and `test_for_formula` to make sure the `source.path` is set correctly in both places. I could expand this to test all of the Tab attributes if you're interested?

MikeMcQuaid:
@penman This is looking great 👏. How's your testing going? When you consider this ready to 🚢 give me what you've been using for testing and I'll try my own too and, if it all looks good, we 🚢 this thing!

> I could expand this to test all of the Tab attributes if you're interested?

Always interested in more tests! Feel free to make that a future PR, though, to avoid blocking this one.

penman:
@MikeMcQuaid I'm skeptical of that test failure. I'm not sure how this change could break <kbd>cask cleanup</kbd>, and I can't reproduce it locally. Is it something you've seen before?

MikeMcQuaid:
@penman Yeh, ignore it, it's flaky.

jawshooah:
@reitermarkus Looks like [this spec](https://github.com/Homebrew/brew/blob/911edb0ed77859445c3a48f6abe5b9be5807ebf7/Library/Homebrew/cask/spec/cask/cli/cleanup_spec.rb#L56-L74) isn't entirely consistent. Mind taking a look?

penman:
@MikeMcQuaid In that case, I think this is ready.

Mostly, what I've been doing to test is installing <kbd>the_silver_searcher</kbd>, checking the INSTALL_RECEIPT has `source.path` set to the path to the formula, then uninstalling and installing it with <kbd>ag</kbd> (an alias for <kbd>the_silver_searcher</kbd>) and checking `source.path` is set to the path to the alias. I've tried a similar thing with a few other formulae with aliases, and it seems to be working. I've also installed some formulae with no aliases and that's been working too.

Nothing has come up in my general usage of Homebrew, either.

I'm curious to see what you test, though, because I'm sure there are paths I haven't thought of where something might have gone wrong, or maybe there are parts of Homebrew that are generally tricky when things change that I'm not aware of.

MikeMcQuaid:
Looks good to me in local testing. Nice work!

MikeMcQuaid:
Rerunning CI then will merge.

MikeMcQuaid:
@jawshooah @reitermarkus Going to need to remove that spec soon; it's making it pretty hard to get consistently green Homebrew builds.

PR #864, opened 2016-09-03
Skip upgrade of pinned dependency if it's outdated

vladshablinsky:

-----

Description:
====

A fix for #831 was introduced in #846 but the behaviour of install process should have been reconsidered anyway. This PR implements the ideas being discussed in #846.

_Idea:_ Don't upgrade dependency of formula if this dependency is pinned while installing the formula.

```
Let's suppose A depends on B and then do the following:
      * install B with version 0.1.
      * pin B with version 0.1.
      * bump new version of B, e.g. 1.0
      * install A

    B shouldn't get upgraded while we installing A if it's pinned and
    installed with the proper options. This commit implements such
    behaviour.
```

TODO?
======

I'm not sure about two things here:

1. Do I need to add a test for the change. And if we need a test, then where should I put it. Possible variants are `test_integration_cmds`, `test_formula_installer` or `test_dependency`?

2. Do we need to `repin` formula `B` if it's pinned and installation of `A` that depends on `B` requires different options from ones being used in pinned keg of `B`, so we eventually upgrade/reinstall `B`, but don't repin it.

CC @MikeMcQuaid @ilovezfs

### DIFF
--- a/Library/Homebrew/cmd/reinstall.rb
+++ b/Library/Homebrew/cmd/reinstall.rb
@@ -8,7 +8,13 @@ module Homebrew
   def reinstall
     FormulaInstaller.prevent_build_flags unless DevelopmentTools.installed?
 
-    ARGV.resolved_formulae.each { |f| reinstall_formula(f) }
+    ARGV.resolved_formulae.each do |f|
+      if f.pinned?
+        onoe "#{f.full_name} is pinned. You must unpin it to reinstall."
+        next
+      end
+      reinstall_formula(f)
+    end
   end
 
   def reinstall_formula(f)
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -146,11 +146,21 @@ class FormulaInstaller
     raise FormulaInstallationAlreadyAttemptedError, formula if @@attempted.include?(formula)
 
     unless skip_deps_check?
-      unlinked_deps = formula.recursive_dependencies.map(&:to_formula).select do |dep|
+      recursive_deps = formula.recursive_dependencies
+      unlinked_deps = recursive_deps.map(&:to_formula).select do |dep|
         dep.installed? && !dep.keg_only? && !dep.linked_keg.directory?
       end
       raise CannotInstallFormulaError,
         "You must `brew link #{unlinked_deps*" "}` before #{formula.full_name} can be installed" unless unlinked_deps.empty?
+
+      pinned_unsatisfied_deps = recursive_deps.select do |dep|
+        dep.to_formula.pinned? && !dep.satisfied?(inherited_options_for(dep))
+      end
+
+      unless pinned_unsatisfied_deps.empty?
+        raise CannotInstallFormulaError,
+          "You must `brew unpin #{pinned_unsatisfied_deps*" "}` as installing #{formula.full_name} requires the latest version of pinned dependencies"
+      end
     end
   end
 
--- a/Library/Homebrew/test/test_formula.rb
+++ b/Library/Homebrew/test/test_formula.rb
@@ -491,6 +491,31 @@ class FormulaTests < Homebrew::TestCase
     f3.rack.rmtree
   end
 
+  def test_eligible_kegs_for_cleanup_keg_pinned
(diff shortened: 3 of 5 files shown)


MikeMcQuaid:
> Do I need to add a test for the change. And if we need a test, then where should I put it. Possible variants are test_integration_cmds, test_formula_installer or test_dependency?

Ideally yes. Add it to `test_formula_installer` or `test_dependency`, whichever you think is best 👍 

> Do we need to repin formula B if it's pinned and installation of A that depends on B requires different options from ones being used in pinned keg of B, so we eventually upgrade/reinstall B, but don't repin it.

I think if it's pinned we should probably just refuse to do anything with it (reinstall or upgrade) and require the user to manually uninstall/unpin/upgrade to resolve it.

Library/Homebrew/dependency.rb
```diff
@@ -38,8 +38,12 @@ def installed?
     to_formula.installed?
   end
 
+  def pinned?
+    to_formula.pinned?
+  end
+
   def satisfied?(inherited_options)
-    installed? && missing_options(inherited_options).empty?
+    (installed? || pinned?) && missing_options(inherited_options).empty?
```
MikeMcQuaid: This is a bit different to what I think we want. Instead of considering this satisfied it should be considered unsatisfied, refuse to upgrade and print out a message telling the user to `unpin`.

vladshablinsky: Ah, I see, will change it this way then and ask for the comments.

vladshablinsky:
@MikeMcQuaid how does the current implementation look?

Library/Homebrew/formula_installer.rb
```diff
@@ -413,6 +418,8 @@ def install_dependencies(deps)
     elsif !deps.empty?
       oh1 "Installing dependencies for #{formula.full_name}: #{Tty.green}#{deps.map(&:first)*", "}#{Tty.reset}",
         :truncate => false
+
+
```
vladshablinsky: TODO: remove.

Library/Homebrew/cmd/reinstall.rb
```diff
@@ -8,7 +8,10 @@ module Homebrew
   def reinstall
     FormulaInstaller.prevent_build_flags unless DevelopmentTools.installed?
 
-    ARGV.resolved_formulae.each { |f| reinstall_formula(f) }
+    ARGV.resolved_formulae.each do |f|
+      raise "#{f.full_name} is pinned. You must unpin it to reinstall." if f.pinned?
```
MikeMcQuaid: I wonder if it's worth making this an `onoe` and using `next` so it'll print a warning for all of them rather than just the first?

Library/Homebrew/formula_installer.rb
```diff
@@ -146,11 +146,16 @@ def check_install_sanity
...
+      recursive_deps = formula.recursive_dependencies.map(&:to_formula)
+      unlinked_deps = recursive_deps.select do |dep|
         dep.installed? && !dep.keg_only? && !dep.linked_keg.directory?
       end
       raise CannotInstallFormulaError,
         "You must `brew link #{unlinked_deps*" "}` before #{formula.full_name} can be installed" unless unlinked_deps.empty?
+
+      pinned_deps = recursive_deps.select(&:pinned?)
+      raise CannotInstallFormulaError,
+        "You must `brew unpin #{pinned_deps*" "}` before #{formula.full_name} can be installed" unless pinned_deps.empty?
```
MikeMcQuaid: May be worth clarifying a little bit `as installing formula.full_name requires the latest version of pinned_deps`

MikeMcQuaid:
A few small comments but I like the approach 👍

vladshablinsky:
Changed the logic a bit, we don't need to fail if dependency is `statisfied?` even if it's pinned.

ilovezfs:
We don't? Since most things go through `opt` now I think we do indeed need to fail …

vladshablinsky:
@ilovezfs if a pinned dependency is satisfied doesn't it mean that if we unpin it it won't be reinstalled anyway?

MikeMcQuaid:
Thanks again @vladshablinsky!

PR #866, opened 2016-09-04
Catch system libressl in openssl audit as well

tdsmith:
The regex didn't match libcrypto.35.dylib, and it should have.

### DIFF
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -29,13 +29,14 @@ module FormulaCellarChecks
     keg = Keg.new(formula.prefix)
     system_openssl = keg.mach_o_files.select do |obj|
       dlls = obj.dynamically_linked_libraries
-      dlls.any? { |dll| %r{/usr/lib/lib(crypto|ssl)\..*dylib}.match dll }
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Is this because the headers aren't available that we want to prevent this linking? If so: 👍, if not: let's talk more.

DomT4:
We might well want to split this into another check to improve the messaging rather than tacking it on to the existing one. LibreSSL also has `libtls` which OpenSSL does not.

Don't have much to say on policing LibreSSL that I didn't already say in https://github.com/Homebrew/homebrew-core/pull/4442#issuecomment-244549171 so I'll leave it at that 😄.

MikeMcQuaid:
👍

DomT4:
I still think this is potentially worth a unique check or a rewrite of the current message.

If something specifically links to `libtls` substituting a `depends_on "openssl"` is unlikely to make that formula happy, for example. `libtls` isn't something OpenSSL provides at all.

tdsmith:
Going to read "potentially worth" as meaning you wouldn't block this as-is. ;) I see your point but there's probably not good general advice in that case anyway.

DomT4:
I think it merits further thought, but I'm happy to do that thought myself, yeah. This is at least a step.

PR #871, opened 2016-09-04
bump-formula-pr: enhancements

ilovezfs:
set sha256 automatically when possible
formula name guessing
handle explicit version DSL
handle gnu mirrors
add --write option

CC @MikeMcQuaid

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -89,7 +89,16 @@ module Homebrew
     elsif !hash_type
       odie "#{formula}: no tag/revision specified!"
     else
-      odie "#{formula}: no url/#{hash_type} specified!"
+      rsrc = Resource.new { @url = new_url }
+      rsrc.download_strategy = CurlDownloadStrategy
+      rsrc.owner = Resource.new(formula.name)
+      rsrc_path = rsrc.fetch
+      if Utils.popen_read("/usr/bin/tar", "-tf", rsrc_path) =~ /\/.*\./
+        new_hash = rsrc_path.sha256
+      else
+        odie "#{formula}: no url/#{hash_type} specified!" if new_url.include?(".tar")
+        new_hash = rsrc_path.sha256
+      end
     end
 
     if ARGV.dry_run?
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -38,6 +38,7 @@ module Homebrew
       unless contents.errors.empty?
         raise Utils::InreplaceError, path => contents.errors
       end
+      path.atomic_write(contents) if ARGV.include?("--write")
       contents
     else
       Utils::Inreplace.inreplace(path) do |s|
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -100,6 +100,7 @@ module Homebrew
     new_hash = ARGV.value(hash_type)
     new_tag = ARGV.value("tag")
     new_revision = ARGV.value("revision")
+    new_mirror = ARGV.value("mirror")
     new_url_hash = if new_url && new_hash
       true
     elsif new_tag && new_revision
@@ -132,6 +133,10 @@ module Homebrew
       replacement_pairs << [/^  revision \d+\n(\n(  head "))?/m, "\\2"]
     end
 
+    if requested_spec == :stable
+      replacement_pairs << [/(^  mirror .*\n)?/, ""]
+    end
+
     replacement_pairs += if new_url_hash
       [
         [formula_spec.url, new_url],
@@ -145,6 +150,10 @@ module Homebrew
     end
 
     backup_file = File.read(formula.path) unless ARGV.dry_run?
+
+    if new_mirror
+      replacement_pairs << [/^( +)(url \"#{new_url}\"\n)/m, "\\1\\2\\1mirror \"#{new_mirror}\"\n"]
+    end
 
     new_contents = inreplace_pairs(formula.path, replacement_pairs)
 
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -109,7 +109,13 @@ module Homebrew
     elsif !hash_type
       odie "#{formula}: no tag/revision specified!"
     else
-      rsrc = Resource.new { @url = new_url }
+      rsrc_url = if requested_spec != :devel && new_url =~ /.*ftpmirror.gnu.*/
+        new_mirror = new_url.sub "ftpmirror.gnu.org", "ftp.gnu.org/gnu"
+        new_mirror
+      else
+        new_url
+      end
+      rsrc = Resource.new { @url = rsrc_url }
       rsrc.download_strategy = CurlDownloadStrategy
       rsrc.owner = Resource.new(formula.name)
       rsrc_path = rsrc.fetch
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -78,18 +98,34 @@ def bump_formula_pr
...
+    forced_version = ARGV.value("version")
     new_url_hash = if new_url && new_hash
       true
     elsif new_tag && new_revision
       false
     elsif !hash_type
       odie "#{formula}: no tag/revision specified!"
     else
-      odie "#{formula}: no url/#{hash_type} specified!"
+      rsrc_url = if requested_spec != :devel && new_url =~ /.*ftpmirror.gnu.*/
```
MikeMcQuaid: Not sure what `rsrl_url` means, can you use a longer name?

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -38,6 +38,7 @@ def inreplace_pairs(path, replacement_pairs)
       unless contents.errors.empty?
         raise Utils::InreplaceError, path => contents.errors
       end
+      path.atomic_write(contents) if ARGV.include?("--write")
```
MikeMcQuaid: Can you document this?

MikeMcQuaid:
One comment otherwise 👍. You're the main user of this so as long as the docs are kept up to date I'm 🆒 with adding anything and everything in here.

Library/Homebrew/dev-cmd/bump-formula-pr.rb
scpeters: 👍 for documenting this and the other new arguments as well (`--version`, `--mirror`)

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -119,6 +159,23 @@ def bump_formula_pr
...
+      if File.read(formula.path).include?("version \"#{old_formula_version}\"")
+        replacement_pairs << [old_formula_version.to_s, forced_version]
+      else
+        if new_mirror
+          replacement_pairs << [/^( +)(mirror \"#{new_mirror}\"\n)/m, "\\1\\2\\1version \"#{forced_version}\"\n"]
+        else
+          replacement_pairs << [/^( +)(url \"#{new_url}\"\n)/m, "\\1\\2\\1version \"#{forced_version}\"\n"]
+        end
+      end
+    elsif forced_version && forced_version == "0"
```
ilovezfs: yes, see the comment in the commit message: https://github.com/Homebrew/brew/pull/871/commits/54930625add92c4d29ab2798a31c76ea7ef393c8

scpeters: How do you feel about passing an empty string `--version=""` instead of `0` to indicate an obsolete version?

Apologies for the bike-shedding; this is a very nice pull request!

scpeters: I think `--version=` would give an empty string as well

ilovezfs: that could work. one concern with that is `--version=` will be interpreted the same as `--version=""`. It might be better to have something explicit like `--remove-version` but I was trying to avoid proliferating options, though it may be warranted here

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -119,6 +159,23 @@ def bump_formula_pr
...
+      replacement_pairs << [/^  version \"[a-z\d+\.]+\"\n/m, ""]
```
scpeters: this regex doesn't include all version characters. I think adding `-`s, `_`s, and capital letters would cover most of `homebrew/core`

ilovezfs: I suspect .* would be good enough.

scpeters:
patch for my two suggestions:

~~~
diff --git a/Library/Homebrew/dev-cmd/bump-formula-pr.rb b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
index afd18b3..c593aa2 100644
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -163,7 +163,7 @@ module Homebrew
       replacement_pairs << [/^( +)(url \"#{new_url}\"\n)/m, "\\1\\2\\1mirror \"#{new_mirror}\"\n"]
     end
 
-    if forced_version && forced_version != "0"
+    if forced_version && forced_version.length > 0
       if File.read(formula.path).include?("version \"#{old_formula_version}\"")
         replacement_pairs << [old_formula_version.to_s, forced_version]
       else
@@ -173,8 +173,8 @@ module Homebrew
           replacement_pairs << [/^( +)(url \"#{new_url}\"\n)/m, "\\1\\2\\1version \"#{forced_version}\"\n"]
         end
       end
-    elsif forced_version && forced_version == "0"
-      replacement_pairs << [/^  version \"[a-z\d+\.]+\"\n/m, ""]
+    elsif forced_version && forced_version.length == 0
+      replacement_pairs << [/^  version \"[a-zA-Z\d+\._-]+\"\n/m, ""]
     end
     new_contents = inreplace_pairs(formula.path, replacement_pairs)
 
~~~

Library/Homebrew/dev-cmd/bump-formula-pr.rb
scpeters: ah yes, the all-encompassing `.*`

or `\"[^\"]+\"` to force it to match 2 `"`'s

ilovezfs: @scpeters the real trick is fixing this part so it will work with the devel spec too ;)

scpeters:
updated patch for my suggestions:

~~~
diff --git a/Library/Homebrew/dev-cmd/bump-formula-pr.rb b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
index afd18b3..72118be 100644
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -163,7 +163,7 @@ module Homebrew
       replacement_pairs << [/^( +)(url \"#{new_url}\"\n)/m, "\\1\\2\\1mirror \"#{new_mirror}\"\n"]
     end
 
-    if forced_version && forced_version != "0"
+    if forced_version && forced_version.length > 0
       if File.read(formula.path).include?("version \"#{old_formula_version}\"")
         replacement_pairs << [old_formula_version.to_s, forced_version]
       else
@@ -173,8 +173,8 @@ module Homebrew
           replacement_pairs << [/^( +)(url \"#{new_url}\"\n)/m, "\\1\\2\\1version \"#{forced_version}\"\n"]
         end
       end
-    elsif forced_version && forced_version == "0"
-      replacement_pairs << [/^  version \"[a-z\d+\.]+\"\n/m, ""]
+    elsif forced_version && forced_version.length == 0
+      replacement_pairs << [/^  version \"[^\"]+\"\n/m, ""]
     end
     new_contents = inreplace_pairs(formula.path, replacement_pairs)
 
~~~

Library/Homebrew/dev-cmd/bump-formula-pr.rb
scpeters: well then instead of generic matching, we could use the formula accessors. for formula `f`, check if `!f.stable.version.detected_from_url?` (or `!f.devel.version.detected_from_url?`), then get the actual version from `f.stable.version.to_s` (or `f.devel.version.to_s`)

sadly I don't have a patch for that yet ;)

ilovezfs: the problem is that resource blocks can be at the same offset and can also have a version, which can be the exact same value, and can have a url, so we have to guarantee we're actually in a devel block

scpeters: Yeah, I guess that's the issue with using `InReplace`. I can think of two options:

1. Figure out the range of line numbers where the appropriate spec starts and end, and limit the `InReplace` call to those lines (not sure if `InReplace` can do that, but I've done something similar with `sed`).

2. Modify the formula data structure and serialize it to the file. This would probably strip comments and seems like it would be hard to get right.

ilovezfs: It'll be fine with inreplace, just needs an appropriate regex, which shouldn't be too hard, I just haven't looked at it yet. Basically, (devel do)(one url)(any number of mirrors)(version)(anything)(end) -> \1\2\3\5\6

MikeMcQuaid:
I'm up for shipping this pretty much as-is?

ilovezfs:
I have a couple updates locally, which I'll push.

ilovezfs:
@MikeMcQuaid PR refreshed.

MikeMcQuaid:
@ilovezfs :shipit:

PR #874, opened 2016-09-04
doc/Interesting-Taps-&-Branches: rename to Interesting-Taps-&-Forks

zmwangx:

-----

This document lists Tigerbrew and Linuxbrew as interesting forks, not branches (they used to be "Interesting Branches (aka forks)" but that changed in b745546), so "Interesting Taps & Branches" is a more fitting title.

The term "branches" is also presumably confusing to folks with a shallow understanding of GitHub and/or git who may not realize how GitHub forks and git branches are related.

---

By the way, I wonder if it's really a good idea to have an ampersand (`&`) in the title and filename. It used to be part of a wiki page title which is fine, but maybe we could sanitize it a bit when putting it on every user's filesystem?


MikeMcQuaid:
Makes sense, thanks!

PR #876, opened 2016-09-04
include Utils.Shell in Formula

gregory-nisbet:

-----

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -12,6 +12,7 @@ require "pkg_version"
 require "tap"
 require "keg"
 require "migrator"
+require "utils/shell"
 
 # A formula provides instructions and metadata for Homebrew to install a piece
 # of software. Every Homebrew formula is a {Formula}.
@@ -41,6 +42,7 @@ require "migrator"
 class Formula
   include FileUtils
   include Utils::Inreplace
+  include Utils::Shell
   extend Enumerable
 
   # @!method inreplace(paths, before = nil, after = nil)
@@ -1706,6 +1708,12 @@ class Formula
 
       remove_method(:options)
     end
+  end
+
+  public
+
+  def shell_profile(*arguments)
+    Utils::Shell::shell_profile(*arguments)
   end
 
   # The methods below define the formula DSL.
--- a/Library/Homebrew/test/test_formula.rb
+++ b/Library/Homebrew/test/test_formula.rb
@@ -552,6 +552,20 @@ class FormulaTests < Homebrew::TestCase
     end
     assert f_true.pour_bottle?
   end
+
+  def test_shell_profile_exists
+    klass = Class.new(Formula) do
+      def initialize()
+      end
+
+      def my_shell_profile(*arguments)
+        shell_profile(*arguments)
+      end
+    end
+
+    formula = klass.new()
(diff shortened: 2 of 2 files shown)


gregory-nisbet:
This is a patch that makes the `shell_profile` method accessible from the `Formula` class again.

gregory-nisbet:
So I tested this locally and was able to install `nvm` with the two-line change in the PR. Adding a test is proving to be more difficult. Since `shell_profile` is part of the public API for a formula, I wanted to add a test to make sure that it's there.

I've tried everything I can think of to write a test for the presence of the `shell_profile` method.

`Formula.respond_to? :shell_profile` does returns false. Calling `Formula.shell_profile` and checking to make sure it returns a string throws an exception. I tried creating a Dummy class and inheriting from Formula, but apparently you can't have a class local to a method in Ruby.

mistydemeo:
> apparently you can't have a class local to a method in Ruby.

You can make anonymous classes using `Class.new`:

```ruby
klass = Class.new(Formula) do
  # ...
end
```

gregory-nisbet:
Okay I am completely lost on how to write a test for this.

With the patch above installation of `nvm` worked locally and I can refer to `shell_profile` in the body of a class inheriting from Formula.

I'm having trouble understanding exactly how including a module from a class works and "scaling down" the logic so I can understand it. I'm not able to reproduce what's happening inside Homebrew in a toy example.

I have a small toy example of a class `include`-ing a module local to another module in another file. If I use `def shell_profile` then `shell_profile` is accessible from the `my_shell_profile` method and does not need to be qualified with a namespace. However, if I define it as `self.shell_profile` then I cannot use the unqualified method and must explicit qualify it with its module.

```
# util.rb
module Wrapper
    module Utils
        # shell_profile (no error)
        # self.shell_profile (NameError)
        def shell_profile()
            return "4"
        end
    end
end
```

```
# test.rb
require_relative "util"

class Formula
    include Wrapper::Utils

    def initialize()
    end

    def my_shell_profile()
        shell_profile
        # must use Wrapper::Utils::shell_profile if
        # shell_profile was defined using "def self.shell_profile"
    end
end

a = Formula.new()

puts a.my_shell_profile()
```

Here is an excerpt from `utils/shell.rb`

```
 71     # return the shell profile file based on users' preferred shell
 72     def self.shell_profile
 73       SHELL_PROFILE_MAP.fetch(preferred_shell, "~/.bash_profile")
 74     end
```

mistydemeo:
I'm going to go ahead and merge this as-is to stem the bleeding.

gregory-nisbet:
sounds good.

gregory-nisbet:
Don't merge this! It didn't actually fix the issue. It only appeared to because brew executes `git checkout master` in the course of installing a module.

gregory-nisbet:
Okay I think I got it now. Formulas relying on `shell_profile` should work without modification.

I added a test making sure that `shell_profile` is in scope inside the body of a method on a class inheriting from Formula.

Library/Homebrew/formula.rb
```diff
@@ -41,6 +42,7 @@
 class Formula
   include FileUtils
   include Utils::Inreplace
+  include Utils::Shell
```
mistydemeo: You shouldn't need to do this anymore if you're having `#shell_profile` call to the fully-qualified name.

PR #879, opened 2016-09-05
Warn when changing branch on update

### LINKED ISSUE #877: Warn when auto-changing branch?
Homebrew automatically (sometimes unexpectedly) doing some git stuff when updating seems to come up in the issue tracker pretty regularly. In #737, @MikeMcQuaid [mentions]() that this is going to be improved at some point. In the meantime, though, I wonder if it would be worth emitting a warning explaining that Homebrew is checking out `master`, and explaining that setting `HOMEBREW_DEVELOPER` will prevent this behaviour?

Is this a good idea?

penman:

-----

See #877 for motivation behind this.

Fixes #877.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -1,33 +1,18 @@
 HOMEBREW_VERSION="0.9.9"
 
-o() {
+onoe() {
+  if [[ -t 2 ]] # check whether stderr is a tty.
+  then
+    echo -ne "\033[4;31mError\033[0m: " >&2 # highlight Error with underline and red color
+  else
+    echo -n "Error: " >&2
+  fi
   if [[ $# -eq 0 ]]
   then
-    /bin/cat
+    /bin/cat >&2
   else
-    echo "$*"
+    echo "$*" >&2
   fi
-}
-
-highlight() {
-  if [[ -t 1 ]] # check whether stdout is a tty.
-  then
-    echo -ne "\033[4;$1m$2\033[0m"
-  else
-    echo -n "$2"
-  fi
-}
-
-opoo() {
-  highlight 33 Warning >&2 # yellow
-  echo -n ": " >&2
-  o "$@" >&2
-}
-
-onoe() {
-  highlight 31 Error >&2 # red
-  echo -n ": " >&2
-  o "$@" >&2
 }
 
 odie() {
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -1,18 +1,33 @@
 HOMEBREW_VERSION="0.9.9"
 
-onoe() {
-  if [[ -t 2 ]] # check whether stderr is a tty.
-  then
-    echo -ne "\033[4;31mError\033[0m: " >&2 # highlight Error with underline and red color
-  else
-    echo -n "Error: " >&2
-  fi
+o() {
   if [[ $# -eq 0 ]]
   then
-    /bin/cat >&2
+    /bin/cat
   else
-    echo "$*" >&2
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/update.sh
```diff
@@ -242,6 +242,15 @@ EOS
   INITIAL_BRANCH="$(git symbolic-ref --short HEAD 2>/dev/null)"
   if [[ "$INITIAL_BRANCH" != "$UPSTREAM_BRANCH" && -n "$INITIAL_BRANCH" ]]
   then
+
+    if [[ -z "$HOMEBREW_DEVELOPER" ]]
+    then
+      opoo <<EOS
```
MikeMcQuaid: Let's just go for: `echo "Checking out $UPSTREAM_BRANCH in $DIR..."`

and, either here or a sensible place below, do:
```bash
  echo "To checkout $INITIAL_BRANCH in $DIR run:"
  echo "  'cd $DIR && git checkout $INITIAL_BRANCH"
```

 Want to avoid recommending setting `HOMEBREW_DEVELOPER` for now; I've got a solution incoming for this problem.

penman:
Updated the message based on @MikeMcQuaid's feedback.

I've left the Bash `opoo` definition, even though it's no longer used, because it'll probably be useful at some point.

penman:
Using a `Pathname` seems like it might be a mistake. I'll change it back to a string tomorrow and then we won't have to worry as much about errors in distant parts of the codebase.

Library/Homebrew/brew.sh
```diff
@@ -1,20 +1,35 @@
...
+  highlight 33 Warning >&2 # yellow
+  echo -n ": " >&2
+  o "$@" >&2
+}
+
+onoe() {
+  highlight 31 Error >&2 # red
+  echo -n ": " >&2
+  o "$@" >&2
+}
```
MikeMcQuaid: I reckone you can revert all this stuff for now.

MikeMcQuaid:
Some stuff to strip out but otherwise 👍

Library/Homebrew/cmd/update.sh
```diff
@@ -242,6 +242,14 @@ EOS
   INITIAL_BRANCH="$(git symbolic-ref --short HEAD 2>/dev/null)"
   if [[ "$INITIAL_BRANCH" != "$UPSTREAM_BRANCH" && -n "$INITIAL_BRANCH" ]]
   then
+
+    if [[ -z "$HOMEBREW_DEVELOPER" ]]
+    then
+      echo "Checking out $UPSTREAM_BRANCH in $DIR..."
+      echo "To checkout $INITIAL_BRANCH in $DIR run:"
+      echo "  'cd $DIR && git checkout $INITIAL_BRANCH"
```
zmwangx: To be pedantic, you need to quote `$DIR` and `$INITIAL_BRANCH` for robustness:

```bash
echo "  cd $(printf %q "$DIR") && git checkout $(printf %q "$INITIAL_BRANCH")"
```

But I guess I'm just being pedantic.

P.S. In Zsh the completely robust version (with even nicer output than the Bash version above) would be

```zsh
cd "${(qq)DIR}" && git checkout "${(qq)INITIAL_BRANCH}"
```

This is one of the things I dislike about Bash: wimpy parameter expansion.

MikeMcQuaid: Not quoting in `echo` is fine, I think, but thanks for the review anyway @zmwangx; definitely would rather have pedantry than the other extreme!

MikeMcQuaid:
Thanks again @penman!

PR #892, opened 2016-09-07
docs: Explain Tab instantiation

penman:

-----

The difference between `Tab.create` and `Tab.for_formula` was pretty confusing to me, and I had to go looking through the `Tab` history, where I found a comment explaining it that has since been deleted, which wasn't a great experience.

So I basically added that explanation back in, and then while I was there it seemed like a good idea to explain the other Tab instantiation methods if there was anything about their purpose that wasn't immediately obvious.


MikeMcQuaid:
Nice work again @penman!

PR #897, opened 2016-09-08
Add language method to `OS::Mac`.

reitermarkus:

-----

This adds a `language` method to the `MacOS` module, to be used in casks, such as https://github.com/caskroom/homebrew-cask/pull/24343.

cc @Homebrew/cask

### DIFF
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -41,6 +41,10 @@ module OS
       version.to_sym
     end
 
+    def language
+      @language ||= Utils.popen_read("defaults", "read", ".GlobalPreferences", "AppleLanguages").delete(" \n\"()").sub(/,.*/, "")
+    end
+
     # Locates a (working) copy of install_name_tool, guaranteed to function
     # whether the user has developer tools installed or not.
     def install_name_tool
--- a/Library/Homebrew/test/test_os_mac_language.rb
+++ b/Library/Homebrew/test/test_os_mac_language.rb
@@ -1,0 +1,8 @@
+require "testing_env"
+require "os/mac"
+
+class OSMacLanguageTests < Homebrew::TestCase
+  def test_language_format
+    assert_match %r{\A[a-z]{2}(-[A-Z]{2})?\Z}, OS::Mac.language
+  end
+end


Library/Homebrew/os/mac.rb
```diff
@@ -41,6 +41,10 @@ def cat
       version.to_sym
     end
 
+    def language
+      @language ||= `defaults read .GlobalPreferences AppleLanguages`.delete(" \n\"()").sub(%r{,.*}, '')
```
DomT4: Can we use `Utils.popen_read` instead here? Backticks are inconsistent with the rest of the file here... and kind of gross to look at 😄.

MikeMcQuaid: 👍

MikeMcQuaid: Also is it worth using `gsub`? I'd also favour `//` regexes unless they contain `/`.

MikeMcQuaid:
Feels a little weird for this to be on the MacOS class but I don't have strong feelings either way if there's no other logical place.

Library/Homebrew/os/mac.rb
reitermarkus: You mean `gsub` instead of `.delete.sub` or instead of only `sub`?

reitermarkus:
> Feels a little weird for this to be on the MacOS class

Where else would you put the OS language, if not in the `OS` module?

Library/Homebrew/os/mac.rb
MikeMcQuaid: Either.

MikeMcQuaid:
> Where else would you put the OS language, if not in the OS module?

👍

Library/Homebrew/os/mac.rb
reitermarkus: Leaving as is because I don't think having a complex regex here is worth it.

Library/Homebrew/os/mac.rb
```diff
@@ -41,6 +41,10 @@ def cat
       version.to_sym
     end
 
+    def language
+      @language ||= Utils.popen_read('defaults', 'read', '.GlobalPreferences', 'AppleLanguages').delete(" \n\"()").sub(/,.*/, '')
```
DomT4: Single quotes is probably a Rubocop violation given Homebrew's Rubocop config?

Library/Homebrew/os/mac.rb
MikeMcQuaid: Feels weird on hindsight to return a comma-separated string. `.delete("()\"\n ").split(",")` seems to do what you want and return an array.

Library/Homebrew/os/mac.rb
reitermarkus: Speaking of RuboCop, when was the last time someone actually ran it? Seems there are thousands of violations.

DomT4: We mass-fixed core code pre-separation, so a while ago. There is stuff Rubocop skips over automatically fixing  though, as well as us having changed various Rubocop rules between then & now.

reitermarkus:
Ok, should be good to go now.

reitermarkus:
Merging now in order for https://github.com/caskroom/homebrew-cask/pull/24343 and https://github.com/caskroom/homebrew-versions/pull/2558 to hopefully go green.

MikeMcQuaid:
@reitermarkus Did you miss this:

> Feels weird on hindsight to return a comma-separated string. `.delete("()\"\n ").split(",")` seems to do what you want and return an array.

I'd have preferred if you disagreed that it was discussed rather than merged?

jawshooah:
It doesn't return a comma-separated string. The `sub` method call strips away the first comma and everything after it, leaving only the first value.

MikeMcQuaid:
Ok. Why does this only care about the first value? That feels relatively opinionated for something that's a OS wrapper around a method? Regardless: this is discussion we should really be having before this was merged.

reitermarkus:
> @reitermarkus Did you miss this:

Yes, sorry, I missed that comment. 

For context, please take a look at https://github.com/caskroom/homebrew-cask/pull/24343/files#diff-d7a1b27ed779bc1601aadf2a1c3652f7R12 on how this is supposed to be used.

Alternatively, we could have `MacOS.languages` return an array, and `MacOS.language` return `languages.first`.

MikeMcQuaid:
@reitermarkus No worries. 👍 to having `languages` return an array and `language` return `languages.first` 👍

PR #901, opened 2016-09-09
completion: Avoid aliases with `command ls`

elyscape:

I haven't written new tests because there aren't really any tests for the bash completion. Additionally, when running `brew tests`, several tests fail with this error:
```
git config --file=/tmp/homebrew-tests-20160909-38939-rk5nwv/prefix/.git/config --replace-all homebrew.devcmdrun true
error: could not lock config file /tmp/homebrew-tests-20160909-38939-rk5nwv/prefix/.git/config: No such file or directory
```
The same tests fail without my changes, though. I'm not sure what that issue is.

-----

If the user has an alias overriding the `ls` command that modifies the output (e.g. `alias ls='ls -F'`), it will cause unexpected characters to appear in completions. Using `command ls` forces the shell to use the command directly, without alias expansion.

This also blackholes the stderr of `ls` in the `__brew_complete_logs` function to avoid printing  errors during completion if the Homebrew logs directory does not exist.

### DIFF
--- a/completions/bash/brew
+++ b/completions/bash/brew
@@ -52,7 +52,7 @@ __brew_complete_formulae() {
 
 __brew_complete_installed() {
   local cur="${COMP_WORDS[COMP_CWORD]}"
-  local inst="$(ls "$(brew --cellar)")"
+  local inst="$(command ls "$(brew --cellar)" 2>/dev/null)"
   COMPREPLY=($(compgen -W "$inst" -- "$cur"))
 }
 
@@ -71,7 +71,7 @@ __brew_complete_versions() {
 
 __brew_complete_logs() {
   local cur="${COMP_WORDS[COMP_CWORD]}"
-  local logs="$(ls "${HOMEBREW_LOGS:-~/Library/Logs/Homebrew/}")"
+  local logs="$(command ls "${HOMEBREW_LOGS:-${HOME}/Library/Logs/Homebrew/}" 2>/dev/null)"
   COMPREPLY=($(compgen -W "$logs" -- "$cur"))
 }
 


zmwangx:
Parsing `ls` output is one of the most basic mistakes in shell scripting (see e.g. http://mywiki.wooledge.org/ParsingLs). Using globbing (filename expansion) instead.

elyscape:
Indeed. I'll switch it to that, then.

elyscape:
@zmwangx Changed to using globs.

MikeMcQuaid:
> Additionally, when running brew tests, several tests fail with this error:

Will be fixed by https://github.com/Homebrew/brew/pull/907.

elyscape:
Rebased onto master. `brew tests` now succeeds.

MikeMcQuaid:
Seems reasonable but will let @zmwangx give final 🆗 and :ship

etc/bash_completion.d/brew
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
bfontaine: I don’t think we need this line because the file is sourced rather than executed.

elyscape: Oh, good catch. That was there for debugging purposes. Will remove.

etc/bash_completion.d/brew
```diff
@@ -71,8 +72,10 @@ __brew_complete_versions() {
 
 __brew_complete_logs() {
   local cur="${COMP_WORDS[COMP_CWORD]}"
-  local logs="$(ls "${HOMEBREW_LOGS:-~/Library/Logs/Homebrew/}")"
-  COMPREPLY=($(compgen -W "$logs" -- "$cur"))
+  local logdir="${HOMEBREW_LOGS:-~/Library/Logs/Homebrew}"
```
chaoren: Please change `~` to `$HOME` so it can get expanded in quotes.

etc/bash_completion.d/brew
```diff
@@ -52,8 +53,8 @@ __brew_complete_formulae() {
 
 __brew_complete_installed() {
   local cur="${COMP_WORDS[COMP_CWORD]}"
-  local inst="$(ls "$(brew --cellar)")"
-  COMPREPLY=($(compgen -W "$inst" -- "$cur"))
+  local inst=("$(brew --cellar)"/*)
+  COMPREPLY=($(compgen -W "${inst[*]#$(brew --cellar)/}" -- "$cur"))
```
chaoren: Suppose, in a dystopian future where packages with spaces in their names exist, wouldn't `inst[@]` be more correct than `inst[*]` here?

elyscape: Ideally yes, but unfortunately `compgen -W` only takes a single argument, which it then splits by the value(s) of `IFS`. `inst[@]` expands to multiple values, which causes this command to fail.

etc/bash_completion.d/brew
elyscape: Fixed.

etc/bash_completion.d/brew
elyscape: Fixed.

chaoren:
You should probably edit the title of the PR too for archival purposes.

MikeMcQuaid:
@zmwangx Merge if 👍

zmwangx:
@MikeMcQuaid Sorry I forgot about this. (I wasn't too enthusiastic because although I love shell scripting, I'm a hardcore Zsh user 😄 ).

---

@elyscape A few thoughts:

1. Globbing needs to be guarded by error checking (whether there are files in the cellar or whether the cellar even exists) — especially necessary when `failglob` option is set; unfortunately I don't have a good native solution for checking emptiness, everything I can thing of requires forking.

2. You're calling `brew --cellar` twice which is unnecessary, save it to a variable if you do need it twice.

3. To get a list of basenames you can simply do `"${filepaths[*]##*/}"` which should be robust in all cases; no need for another `$(brew --cellar)` or `$logdir`.

4. In order to be robust I would also set `IFS` locally everywhere, but maybe some people will find it an overkill.

In general, writing robust shell scripts is not easy, and Bash doesn't make it any easier (no nulls in strings, for instance). After thinking this through, I think `command ls` is reasonable after all; any added robustness will be defeated by the `-W` option to `complete` anyway.

So my conclusion is: let's switch to the simple `command ls` version, but make sure the cellar and the logdir exists (``[[ -d dir ]]``) before doing so. Sorry for changing my mind, and thanks for bearing with us.

elyscape:
No worries! Will do.

elyscape:
@zmwangx Okay, I'm able to work on this again. What should I have happen if the cellar or logdir don't exist?

zmwangx:
Actually, directing `command ls` error messages to `/dev/null` might be enough. (You know, there will be error messages when cellar or logdir doesnt' exist.)

elyscape:
Okay, blackholed the stderr for both calls.

zmwangx:
👍 Happy to ship (but have to wait for CI). If this is not shipped in 24 hours please ping us 😅

elyscape:
Will do! Thanks for your help.

zmwangx:
Thank you @elyscape!

PR #906, opened 2016-09-10
Make `MacOS.language` less opinionated and add `language` stanza.

reitermarkus:

-----

This adds a `languages` method which returns an array, which in turn is called by `language`.

A second addition is a `language` stanza to the cask DSL, which will hopefully make it easier to create and maintain multilingual casks.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/dsl.rb
+++ b/Library/Homebrew/cask/lib/hbc/dsl.rb
@@ -64,6 +64,7 @@ module Hbc
                             :depends_on,
                             :gpg,
                             :homepage,
+                            :language,
                             :license,
                             :name,
                             :sha256,
@@ -96,6 +97,31 @@ module Hbc
     def homepage(homepage = nil)
       assert_only_one_stanza_allowed :homepage, !homepage.nil?
       @homepage ||= homepage
+    end
+
+    def language(*args, &block)
+      @language ||= {}
+
+      if !args.empty? && block_given?
+        args.each do |arg|
+          MacOS.languages.each_with_index do |l, index|
+            string_or_regex = arg == :default ? %r{^en} : arg
+            next unless l.match(string_or_regex)
+            next unless @language[:level].nil? || @language[:level] > index
+            @language = {
+                          block: block,
+                          level: index,
+                        }
+          end
+        end
+
+        if args.include?(:default)
+          # :default has to be the last language block in order to assign return value of the selected `language` block to `@language`
+          @language = @language[:block].call
+        end
+      else
+        @language
+      end
     end
 
     def url(*args, &block)
--- a/Library/Homebrew/cask/lib/hbc/dsl.rb
+++ b/Library/Homebrew/cask/lib/hbc/dsl.rb
@@ -103,8 +103,15 @@ module Hbc
     def language(*args, default: false, &block)
       if !args.empty? && block_given?
         @language_blocks ||= {}
-        @language_blocks.default = block if default
         @language_blocks[args] = block
+
+        return unless default
+
+        unless @language_blocks.default.nil?
+          raise CaskInvalidError.new(token, "Only one default language may be defined")
+        end
+
+        @language_blocks.default = block
       else
         language_eval
         @language
--- a/Library/Homebrew/cask/lib/hbc/dsl.rb
+++ b/Library/Homebrew/cask/lib/hbc/dsl.rb
@@ -103,8 +103,7 @@ module Hbc
       if !args.empty? && block_given?
         args.each do |arg|
           MacOS.languages.each_with_index do |l, index|
-            string_or_regex = arg == :default ? %r{^en} : arg
-            next unless l.match(string_or_regex)
+            next unless arg == :default || l.match(arg)
             next unless @language.nil? || @language[:level].nil? || @language[:level] > index
             @language = {
                           block: block,
--- a/Library/Homebrew/cask/lib/hbc/dsl.rb
+++ b/Library/Homebrew/cask/lib/hbc/dsl.rb
@@ -1,4 +1,5 @@
 require "set"
+require "locale"
 
 require "hbc/dsl/appcast"
 require "hbc/dsl/base"
@@ -99,9 +100,10 @@ module Hbc
       @homepage ||= homepage
     end
 
-    def language(*args, &block)
+    def language(*args, default: false, &block)
       if !args.empty? && block_given?
         @language_blocks ||= {}
+        @language_blocks.default = block if default
         @language_blocks[args] = block
       else
         language_eval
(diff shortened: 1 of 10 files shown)


MikeMcQuaid:
👍 thanks!

reitermarkus:
Seems like the Mavericks `test-bot` is using `zh-Hans` as language. So the test in the regex should also match http://www.unicode.org/iso15924/iso15924-codes.html.

reitermarkus:
@Homebrew/cask, could someone look over this?

Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -114,8 +115,13 @@ def language(*args, &block)
                       }
         end
       end
+
+      if args.include?(:default)
+        # :default has to be the last language block in order to assign return value of the selected `language` block to `@language`
+        @language = @language[:block].call
+      end
```
reitermarkus: I'm sure there is a better place/way to do this. Could need some help here, @jawshooah and @mwean.

Library/Homebrew/os/mac.rb
```diff
@@ -41,8 +41,12 @@ def cat
       version.to_sym
     end
 
+    def languages
+      @languages ||= Utils.popen_read("defaults", "read", ".GlobalPreferences", "AppleLanguages").delete(" \n\"()").split(",")
```
UniqMartin: How about the following?

```ruby
Utils.popen_read("defaults", "read", ".GlobalPreferences", "AppleLanguages").scan(/[^ \n"(),]+/)
```

Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -99,6 +100,31 @@ def homepage(homepage = nil)
...
+    @language ||= {}
+
+    if !args.empty? && block_given?
+      args.each do |arg|
+        MacOS.languages.each_with_index do |l, index|
+          string_or_regex = arg == :default ? %r{^en} : arg
+          next unless l.match(string_or_regex)
+          next unless @language[:level].nil? || @language[:level] > index
+          @language = {
+                        block: block,
```
MikeMcQuaid: Up to you but I'd pull this back to being two-chars in from `@language`.

Library/Homebrew/os/mac.rb
```diff
@@ -41,8 +41,12 @@ def cat
       version.to_sym
     end
 
+    def languages
+      @languages ||= Utils.popen_read("defaults", "read", ".GlobalPreferences", "AppleLanguages").scan(/[^ \n"(),]+/)
```
MikeMcQuaid: 👍

Library/Homebrew/cask/lib/hbc/dsl.rb
reitermarkus: This is the style according to the current RuboCop config for Cask, so I'm leaving this as-is for now as it is used throughout `cask/*`.

mwean:
I'm trying to work on this, but I'm having a ton of trouble getting things working. The `develop_brew_cask` command isn't working, and I can't run the tests because of missing environment variables. I'm on Gitter if anyone can help me.

Library/Homebrew/cask/lib/hbc/dsl.rb
MikeMcQuaid: :cool:. Would be good after https://github.com/Homebrew/brew/pull/989 to try and get a single style and `rubocop`/`brew style` invocation working for both so we can add `brew style` to CI.

reitermarkus:
Never used the `develop_brew_cask` command, but this will probably need to be rewritten or be removed, as developing *only* Cask will not work as before anyways.

Library/Homebrew/os/mac.rb
```diff
@@ -41,8 +41,12 @@ def cat
...
+    end
+
     def language
-      @language ||= Utils.popen_read("defaults", "read", ".GlobalPreferences", "AppleLanguages").delete(" \n\"()").sub(/,.*/, "")
+      languages.first
```
UniqMartin: Technically, this will return `nil` if there's no default language; previously it returned an empty string. I guess it doesn't matter in practice as there's a slim (or even zero?) chance of this macOS setting to be set to an empty array. Still wanted to point this out.

reitermarkus: > there's a slim (or even zero?) chance of this macOS setting to be set to an empty array

There is only one way I can think of this could fail: Manually changing/deleting `.GlobalPreferences`, which is highly unlikely, so I think we're good here.

MikeMcQuaid:
@reitermarkus 🚢?

reitermarkus:
Would still like to get some feedback from @Homebrew/cask.

mwean:
I'd love to help, but I can't get things running locally. I'd also love to have some tests around this. I also don't like that we have to make `:default` the last declaration.

reitermarkus:
>  I also don't like that we have to make :default the last declaration.

This is not the case anymore. This is done using the `language_eval` method now.

ilovezfs:
:-1:

mwean:
@reitermarkus I made a slightly different DSL: https://github.com/reitermarkus/brew/compare/os-language...mwean:mw/language-dsl I wanted to avoid having a separate `default` block, and I wanted to make it easy to make the language name different from the pattern.

reitermarkus:
@mwean, I like the idea of only allowing strings in the DSL, so I have added a `Locale` class to enable matching locales by string only instead of regexes *and* strings. Still needs to be integrated into the `language` DSL.

reitermarkus:
I think this is pretty much ready. @mwean, I changed it mostly to your DSL, except for the `name:` parameter, as I think simply using the return value of the block is more flexible, and more in line with `url do`. @jawshooah, please also have a final look.

Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -98,6 +100,40 @@ def homepage(homepage = nil)
...
+        @language
+      end
+    end
+
+    def language_eval
+      return if instance_variable_defined?(:@language)
+
+      return unless instance_variable_defined?(:@language_blocks)
+
+      MacOS.languages.map(&Locale.method(:parse)).any? { |locale|
```
jawshooah: Seems like `each` is a better fit here than `any?` since you're not using the return value, as a boolean or otherwise.

Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -98,6 +100,40 @@ def homepage(homepage = nil)
...
+
+        return unless default
+
+        unless @language_blocks.default.nil?
+          raise CaskInvalidError.new(token, "Only one default language may be defined")
+        end
+
+        @language_blocks.default = block
+      else
+        language_eval
```
reitermarkus: I'll leave this as is because of https://github.com/Homebrew/brew/pull/906/files/3def3b805b139960f11af68e07e3295c6dbe92cb#diff-f85d4948a976b9727959680dbfe1118aR16.

Otherwise I'd have to call `@dsl.language` to actually “evaluate” the languages, which kind of defeats the purpose of having a `language_eval` method.

PR #917, opened 2016-09-11
caveats: highlight PKG_CONFIG_PATH if exists

DomT4:

-----

We keep getting issues every now & again about people being confused that `keg_only` includes hiding the `pkg-config` files. Perhaps seems reasonable to start pointing out those exist when they do & a hint on how to access them.

This could be complicated a little more by ensuring we only ever print one `PKG_CONFIG_PATH` hint if `pkg-config` files exist in both `lib/pkgconfig` _and_ `share/pkgconfig` but I'm not aware of a situation like that with any existing formula.

Hidden behind a `which` check to avoid redundant messages, although `pkg-config` is an incredibly popular formula. Ran some time tests, no appreciable difference from the status quo.

If accepted:
Closes https://github.com/Homebrew/homebrew-core/issues/4669.


MikeMcQuaid:
Good call 👍

PR #407, opened 2016-06-27
Call subcommands from zsh completion

joshka:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [ ] Have you successfully run `brew tests` with your changes locally?

(I'm not sure that these changes require tests so much as a code review and
a look at the other related changes.)

-----

This change is inspired by the way that the git zsh completions work by foisting
the responsibility for sub commands onto the command themselves.
It is paired with another change that takes the oh-my-zsh brew cask plugin and
makes it into a first class completion rather than it overriding this.

This change requires fixes in brew, oh-my-zsh and the zsh-completions, and will
likely cause the `brew cask` completions to break until these have been fully
accepted into each project. I'd anticipate doing brew first and the others after.

See related PRs:
https://github.com/zsh-users/zsh-completions/pull/430
https://github.com/robbyrussell/oh-my-zsh/pull/5191

PR #936, opened 2016-09-11
Install brew-cask zsh completions by default

joshka:
  - not required
  - not required

-----

Given that `brew-cask` is now part of Homebrew core, it might be nice to have the zsh completion just work out of the box per discussions on https://github.com/Homebrew/brew/pull/407. Similarly, the same could be said about the bash completions.

These commits mirror those in https://github.com/joshka/brew-cask-zsh-completions with modifications to the commit messages to align better with being part of homebrew.

@MikeMcQuaid

### DIFF
--- a/share/zsh/site-functions/_brew_cask
+++ b/share/zsh/site-functions/_brew_cask
@@ -1,0 +1,83 @@
+#compdef brew-cask
+#autoload
+
+# Autocompletion for homebrew-cask (https://github.com/caskroom/homebrew-cask/).
+#
+# Originally sourced from the oh-my-zsh
+# https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/brew-cask/brew-cask.plugin.zsh
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2009-2016 Robby Russell and contributors
+# See the full list at https://github.com/robbyrussell/oh-my-zsh/contributors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'm 🆒 with this but @apjanke @jawshooah @reitermarkus may want to take a look.

apjanke:
Generally 👍 on this, both on pulling `brew cask` completions in to core Homebrew and on how this is coded.

Minor coding note: I think the `_brew_cask` file actually needs a `_brew_cask` invocation at the bottom of the file (after the `_brew_cask()` definition), because it's using the "redefine-on-autoload" trick. Its absence means that nothing will happen the first time you hit tab to invoke completion on `brew cask` in a shell session, because that first invocation just loads the function definitions here. I'll try to test this to verify.

Might have a migration issue here: users who already have `brew-cask-zsh-completions` installed will have a symlink at the location where we're going to drop this file, so the update might fail. Might be few enough of them we should just go ahead with it and deal with the breakage; I don't know of a good easy migration mechanism.

joshka:
@apjanke you might also like to take a look at https://github.com/Homebrew/brew/pull/951

I'll take a look at the two issues you raised later tonight or tomorrow. Thanks :)

joshka:
I think that problems with having `brew-cask-zsh-completions` installed is likely fairly minimal exposure as https://github.com/Homebrew/homebrew-completions/pull/135 is not merged, so that project hasn't been installed automatically by anyone.

apjanke:
Cool. In that case, I think this is ready to ship.

PR #939, opened 2016-09-12
keg: make guile's site-dir persistent

DomT4:

-----
Things that install Guile scheme objects should be doing so in `HOMEBREW_PREFIX/share/guile/site`, which should be persistent rather than a symlink to any formula's Cellar, not even `guile`'s necessarily as Homebrew/homebrew-core@f061d86 managed to expose.

I believe `guile` isn't actually configured correctly to expect this directory, but since absolutely nobody has complained to date as far as I can find & `gnutls` seems to be the only formula using it this way the temptation is to leave it misconfigured & use this directory for site schemes.

I don't really know the deeper issues around messing with `guile`'s configuration so going for a minimal fix seems the safest option.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -326,6 +326,7 @@ class Keg
       when /^fish/ then :mkpath
       # Lua, Lua51, Lua53 all need the same handling.
       when /^lua\// then :mkpath
+      when %r{^guile/} then :mkpath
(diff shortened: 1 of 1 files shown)


DomT4:
First CI failure seems to be because Rubygems was having some issues. Rescheduled the build.

MikeMcQuaid:
👍

PR #951, opened 2016-09-13
Refactor and add commands to brew zsh completion

joshka:

(removed N/A checklist items)

-----

This is a full rewrite of the zsh completion based on the changes we
made to support external commands. A function for each brew command
handles only that command's arguments. The main completion function
dispatches the completion flow to the correct function. This
generically works with both internal and external commands.

This change includes all arguments for all internal brew commands.
Some commands are missing argument documentation due to missing help
docs.

@ilovezfs or @bfontaine care to comment / review this for errors?

### DIFF
--- a/completions/zsh/_brew
+++ b/completions/zsh/_brew
@@ -3,154 +3,825 @@
 
 # Brew ZSH completion function
 
-_brew_all_formulae() {
-  formulae=(`brew search`)
+# functions starting with __brew are helper functions that complete or list
+#   various types of items.
+# functions starting with _brew_ are completions for brew commands
+#   this mechanism can be extended by external commands by defining a function
+#   named _brew_<external-name>. See _brew_cask for an example of this.
+
+# a list of aliased internal commands
+__brew_list_aliases() {
+  local -a aliases
+  aliases=(
+    ls list
+    homepage home
+    '-S' search
+    up update
+    ln link
+    instal install
+    rm uninstall
+    remove uninstall
+    configure diy
+    abv info
+    dr doctor
+    '--repo' '--repository'
+    environment '--env'
+    '--config' config
+    '-v' '--version'
+  )
+  echo "${aliases}"
 }
 
-_brew_installed_formulae() {
-  installed_formulae=(`brew list`)
+__brew_formulae() {
+  local -a formulae
+  formulae=($(brew search))
+  _describe -t formulae 'all formulae' formulae
 }
 
-_brew_installed_taps() {
-  installed_taps=(`brew tap`)
+__brew_installed_formulae() {
+  local -a formulae
+  formulae=($(brew list))
+  _describe -t formulae 'installed formulae' formulae
 }
 
-_brew_official_taps() {
-  official_taps=(`brew tap --list-official`)
+__brew_outdated_formulae() {
+  local -a formulae
+  formulae=($(brew outdated))
+  _describe -t formulae 'outdated formulae' formulae
 }
 
-_brew_pinned_taps() {
-  pinned_taps=(`brew tap --list-pinned`)
+__brew_installed_taps() {
+  local -a taps
+  taps=($(brew tap))
+  _describe -t installed-taps 'installed taps' taps
 }
 
-_brew_outdated_formulae() {
-  outdated_formulae=(`brew outdated`)
+__brew_official_taps() {
+  local -a taps
+  taps=($(brew tap --list-official))
+  _describe -t official-taps 'official taps' taps
 }
 
-__brew_command() {
-  local command="$1"
-  local completion_func="_brew_${command//-/_}"
-  declare -f "$completion_func" >/dev/null && "$completion_func" && return
+__brew_pinned_taps() {
+  local -a taps
+  taps=($(brew tap --list-pinned))
+  _describe -t pinned-taps 'pinned taps' taps
 }
 
-local -a _1st_arguments
-_1st_arguments=(
-  'audit:check formulae for Homebrew coding style'
-  'cat:display formula file for a formula'
-  'cleanup:uninstall unused and old versions of packages'
-  'commands:show a list of commands'
-  'config:show homebrew and system configuration'
-  'create:create a new formula'
-  'deps:list dependencies and dependants of a formula'
-  'desc:display a description of a formula'
-  'doctor:audits your installation for common issues'
-  'edit:edit a formula'
-  'fetch:download formula resources to the cache'
-  'gist-logs:generate a gist of the full build logs'
-  'home:visit the homepage of a formula or the brew project'
-  'info:information about a formula'
-  'install:install a formula'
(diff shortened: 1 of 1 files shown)


share/zsh/site-functions/_brew
```diff
@@ -3,154 +3,825 @@
...
-      '1: :->forms' && return 0
-
-    if [[ "$state" == forms ]]; then
-      _brew_outdated_formulae
-      _wanted outdated_formulae expl 'outdated formulae' compadd -a outdated_formulae
-    fi ;;
-  *)
-    __brew_command "$command" ;;
-esac
+# functions starting with __brew are helper functions that  complete or list
```
apjanke: extra space in "that  complete" here

share/zsh/site-functions/_brew
```diff
@@ -3,154 +3,825 @@
...
+#   various types of items
+# functions starting with _brew_ are completions for brew commands
+#   this mechanism can be extended by external commands by defining a function
+#   named _brew_<external-name>. See _brew_cask for an example of this
```
apjanke: Line should end with a "."

apjanke:
This looks great!

I like how this is coded. The `_brew_<cmd>` and `__brew_<helper>` naming conventions are appropriate.

Looks good based on a quick read-through. I'll test it out locally and see if I have any detailed feedback.

joshka:
Fixed the small typos

MikeMcQuaid:
Will let @apjanke 🚢 this if 👍

apjanke:
Sorry for the delay. It's been a busy week, and I haven't had time to test this yet. Should have some time tomorrow.

MikeMcQuaid:
Merge conflicts need addressed here.

joshka:
I've amended this PR with a rebase on master. The merge conflict was due to https://github.com/Homebrew/install/pull/60

completions/zsh/_brew
```diff
@@ -3,154 +3,825 @@
...
+}
+
+# brew bump-formula-pr [--devel] [--dry-run] [--audit|--strict] --url=url --sha256=sha-256 formula
+# brew bump-formula-pr [--devel] [--dry-run] [--audit|--strict] --tag=tag --revision=revision formula
+_brew_bump_formula_pr() {
+  _arguments \
+      '--devel[bump the development rather than stable version]' \
+      '--dry-run[print what would be done rather than doing it]' \
+      '(--strict)--audit[run brew audit before opening the PR]' \
+      '(--audit)--strictrun brew audit --strict before opening the PR]' \
```
apjanke: Missing a "[" here: "--strictrun" --> "--strict[run".

completions/zsh/_brew
```diff
@@ -3,154 +3,825 @@
...
+}
+
+# brew command cmd
+_brew_command() {
+  _arguments \
+    ': :__brew_all_commands'
+}
+
+# brew commands [--quiet [--include-aliases]]
+_brew_commands() {
```
apjanke: I'm not sure the logic here is ideal. It seems to have the effect of not showing `--include-aliases` in the list of options if you just try to complete `brew commands --`. But if you complete `brew commands --i`, it'll complete to `--include-aliases`, regardless of whether `--quiet` is already specified. And if `--quiet` is specified, trying to complete anything fills in `--include-aliases`. Is that intentional? The comment above seems to indicate that `--include-aliases` is a sub-option of `--quiet`, and that matches my reading of the source for `brew commands`.

Strictly speaking, this might be the right behavior, but it feels a little odd from a usability perspective. I think it might be better to just present both `--quiet` and `--include-aliases` as options, along with a description for each, even though there's a dependency there.

Concrete example: if you do `brew commands ` and hit tab, it first fills in `--quiet`, and if you hit tab again, it fills in `--include-aliases`. This feels a little odd to me, even though it might be technically correct. I feel like as a user, I'd rather see a menu with both `--quiet` and `--include-aliases` and a description of each, instead of them each popping up in turn.

completions/zsh/_brew
```diff
@@ -3,154 +3,825 @@
...
+}
+
+__brew_all_commands() {
+  local -a commands
+  commands=($(_call_program brew brew commands --quiet --include-aliases))
+  _describe -t all-commands 'all commands' commands
+}
+
+__brew_commands() {
+  # TODO remove duplicates between common and all commands
```
apjanke: Good TODO. Was just going to leave a comment asking for this, but I see you've already considered it. Definitely would be nice to have, but not a reason to block shipping this.

apjanke:
Note for Oh My Zsh users: if you are running Oh My Zsh and have its `brew` plugin enabled, that plugin will mask and disable these completions. So if you are running Oh My Zsh, remove `brew` from your list of plugins when testing or using this.

With this PR, our completion is substantially better than the OMZ `brew` plugin's. So once this is shipped and we've been using it for a bit, I'll put in a PR at OMZ to deprecate and remove their `brew` plugin.

completions/zsh/_brew
joshka: This was the intent. It's a somewhat odd command in that it really only has the following valid forms according to the help (and reading the docs):
```shell
brew commands
brew commands --quiet
brew commands --quite --include-aliases
```

Learning Zsh completions is hampered by documentation that is fairly obtuse. I kind of like the idea of leaving this in here as it gives someone coming later a very simple example of how to make a dependent flag available.

The naked `--i` behavior seems like it's a bug (but one that very likely doesn't matter a bunch).

joshka:
Sorry about the delay in replying.
The OMZ PR deprecating the brew plugin is already merged ;)
https://github.com/robbyrussell/oh-my-zsh/pull/5190

completions/zsh/_brew
joshka: I'm generally not a fan of TODOs in code, but didn't want to have this fall through the cracks / add to the complexity of this PR. This also implies some not obvious questions about whether we should show a description for every command, or instead reduce the shown commands to only the common commands (like how git does). If it's the latter, what does that list look like?

completions/zsh/_brew
joshka: Fixing in next push

joshka:
I reckon, let's get this merged.

A couple of follow up tasks:
1. I can see the need for `brew tap` to allow taps to install completions in order to get the `brew-services` completions working from https://github.com/robbyrussell/oh-my-zsh/pull/4313 by @RomanKapitonov
2. Command de-duplication or common commands cleanup

completions/zsh/_brew
apjanke: Good questions, which I don't know the answer to. But I don't think we need to answer those before merging this.

completions/zsh/_brew
apjanke: That makes sense. I have no problem leaving this as is, then.

apjanke:
I agree. Merged! Thanks for the useful work here.

PR #954, opened 2016-09-13
Make scons use stdenv only if immediate build dep

aw1621107:

-----

formula_installer will use stdenv if scons is anywhere in the formula's
recursive dependency list:

    https://github.com/Homebrew/legacy-homebrew/issues/40401#issuecomment-110066355

Having scons as a dependency should only require stdenv if it is an
immediate build dependency, as otherwise scons shouldn't be invoked.

-----

This should address [#47260 from legacy-homebrew](https://github.com/Homebrew/legacy-homebrew/issues/47260), with more info in [#40401](https://github.com/Homebrew/legacy-homebrew/issues/40401).

This doesn't seem to break things horribly at first glance. `gringo`, which uses `scons` for its build, builds without a hitch, and it seems the `--env=std` flag is being properly passed in that case. Building LLVM with its OCaml bindings appears to be correctly missing the `--env=std` flag.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -562,7 +562,7 @@ class FormulaInstaller
 
     if ARGV.env
       args << "--env=#{ARGV.env}"
-    elsif formula.env.std? || formula.recursive_dependencies.any? { |d| d.name == "scons" }
+    elsif formula.env.std? || formula.deps.select(&:build?).any? { |d| d.name == "scons" }
       args << "--env=std"
     end
 


MikeMcQuaid:
CC @mistydemeo for thoughts

DomT4:
I'd prefer we just fix `scons`, really. `env :std` is a nightmare to control & keep build environments sane.

MikeMcQuaid:
@DomT4 Agreed about fixing `scons` but this is no worse than before and, in fact, a little better.

mistydemeo:
@domt4 i don't think an upstream change to accommodate this will come, unfortunately. The environment filtering is a core part of scons and the developers haven't expressed interest in providing a way to bypass it.

MikeMcQuaid:

DomT4:
@mistydemeo Aye, I remember. A while ago you mentioned there might be a fix for that on our side that we'd have to put up patching forever, and I think that's likely still the best option here, but my comment wasn't intended as a blocker on this specifically.

PR #955, opened 2016-09-13
Teach brew to recognize gcc-HEAD as valid compiler

aw1621107:

-----

Building gcc with --HEAD results in most of the executables not having a
version suffix, e.g. Building/installing gcc 6 would result in gcc-6,
g++-6, etc. being installed, while building/installing gcc --HEAD would
result in gcc-, g++-, etc. being installed.

The lack of a version suffix prevented brew from recognizing a valid gcc
install, resulting in brew instructing users to install gcc before
building certain formulae even though gcc is installed.

This, coupled with a patch to the gcc formula that makes the string
"HEAD" the version string for --HEAD builds should allow brew to
recognize gcc installs from trunk.

### DIFF
--- a/Library/Homebrew/compilers.rb
+++ b/Library/Homebrew/compilers.rb
@@ -1,7 +1,7 @@
 # @private
 module CompilerConstants
-  GNU_GCC_VERSIONS = %w[4.3 4.4 4.5 4.6 4.7 4.8 4.9 5 6]
-  GNU_GCC_REGEXP = /^gcc-(4\.[3-9]|[5-6])$/
+  GNU_GCC_VERSIONS = %w[4.3 4.4 4.5 4.6 4.7 4.8 4.9 5 6 7]
+  GNU_GCC_REGEXP = /^gcc-(4\.[3-9]|[5-7])$/
   COMPILER_SYMBOL_MAP = {
     "gcc-4.0"  => :gcc_4_0,
     "gcc-4.2"  => :gcc,


apjanke:
I tested this locally with openblas and the `gcc-HEAD` versioning and it still wasn't picking up the `gcc-HEAD` compiler.

Maybe version-parsing code in [`non_apple_gcc_version` in `development_tools.rb`](https://github.com/Homebrew/brew/blob/278ae8415cfac46e7a063ad61a3b3ce7a362e11b/Library/Homebrew/development_tools.rb#L74-L81) also needs to be changed.

```
    def non_apple_gcc_version(cc)
      (@non_apple_gcc_version ||= {}).fetch(cc) do
        path = HOMEBREW_PREFIX.join("opt", "gcc", "bin", cc)
        path = locate(cc) unless path.exist?
        printf "path = #{path}\n"
        version = `#{path} --version`[/gcc(?:-\d(?:\.\d)? \(.+\))? (\d\.\d\.\d)/, 1] if path
        @non_apple_gcc_version[cc] = version
      end
    end
```

That looks like it's still expecting a numeric pattern for the version, instead of the `HEAD` version it outputs under the accompanying change.

```
[/usr/local/Library/Taps/homebrew/homebrew-core on ⇄ pr-4783]
$ gcc-HEAD --version
gcc-HEAD (Homebrew gcc HEAD-240127) 7.0.0 20160914 (experimental)
```

apjanke:
I think simply relaxing

```
/gcc(?:-\d(?:\.\d)? \(.+\))? (\d\.\d\.\d)/
```

to

```
/gcc(?:-\d(?:\.\d)? \(.+\))? (\d\.\d\.\d)/
```

apjanke:
(Sorry, hit "Close" when I was looking for a "Cancel" button on my comment. Reopened.)

MikeMcQuaid:
I don't think this is the right approach. GCC-HEAD is not a version so we'll be unable to blacklist it. Instead it should create an actual semver version (with RC/beta/alpha flags as necessary) and use that version e.g. a simple version by adding +1 to the stable version's major or minor release.

aw1621107:
@apjanke Yeah, I'll take a look at that. Thanks for the catch!

@MikeMcQuaid Hmmm, that could work too. Makes sense, too, because GCC uses the +1 thing for versioning its trunk builds. And could you expand a bit on not being able to blacklist GCC-HEAD? Not quite sure what you're getting at, especially since it looks like the code I changed made HEAD a valid version

MikeMcQuaid:
I mean you can blacklist GCC-HEAD but that doesn't say much when the version of GCC-HEAD will change over time. You don't actually want to blacklist HEAD, you want to blacklist the version it's based on (as if that version gets e.g. released as a stable one you probably want to blacklist that too).

aw1621107:
Ah, I see. Makes sense.  I'll go make that change

aw1621107:
Just thought, would installing using `brew install gcc --HEAD` but calling gcc-7 (or -8, -9, etc. later) be confusing to users? Or should we assume that `--HEAD` installers know enough for this to not be an issue?

MikeMcQuaid:
> Or should we assume that --HEAD installers know enough for this to not be an issue?

this 👍

aw1621107:
Gotcha

aw1621107:
Alright, pushed the new version. Only change here is allowing GCC version 7. Seems that valid versions will have to be updated manually? Guess that's what's been happening, though, so hopefully not a big deal.

Still having issues with `openblas`, although for different reasons -- it's looking for a formula with the name `gcc7`, which I presume would normally be present in homebrew/versions.

aw1621107:
Yay, found the openblas issue. Should be fixed now.

MikeMcQuaid:
Seems reasonable and we'll want/need this eventually anyway 👍

PR #956, opened 2016-09-13
Xcode 8.0 changes.

DomT4:

-----


chdiza:
Is there a CLT release for Xcode8 for El Capitan?  I only see one for Sierra on the developer downloads page.

mistydemeo:
There was a beta for El Cap up through beta 6, but they haven't uploaded the GM.

DomT4:
For whatever reason CLT releases chasing Xcode releases have been considerably slower this cycle 😕. Sometimes 12+ hours behind Xcode's release.

On Sierra there's also been the fun fun situation where it would routinely try to shunt you back down to the 7.x CLT if you upgraded to the 8.x CLT.

DomT4:
Split the Xcode 8.0 is no longer prerelease commit into https://github.com/Homebrew/brew/pull/957 since the CLT update still isn't here. Might as well stop yelling at people that they're using a prerelease Xcode when, well, they aren't.

DomT4:
@MikeMcQuaid @mistydemeo @ilovezfs Made the lack of up-to-date CLT/Xcode fatal here, as well as adjusting various bits to the CLT releases no longer being kept in sync. Will merge this in after 🍏 if everyone's happy.

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -15,7 +15,7 @@ def latest_version
         when "10.8"  then "5.1.1"
         when "10.9"  then "6.2"
         when "10.10" then "7.2.1"
-        when "10.11" then "7.3.1"
+        when "10.11" then "8.0"
```
ilovezfs: note that this may be controversial to enforce on 10.11, but it doesn't bother me

DomT4: Aye, possibly. Since we've updated CI it's probably sensible to make sure people are using the same development tools we test against regardless. I'd tentatively consider 7.3.x on 10.11 an unsupported configuration by us at this point 😕.

ilovezfs:
:+1:

Library/Homebrew/os/mac/xcode.rb
chdiza: How can it be called unsupported if that's the only way to have the CLT?

ilovezfs: Because that's false.

DomT4: It's not the only way to have the CLT. Apple have noted in Xcode 8.1's release notes that if you want the CLT on 10.11 you should be using the 7.3.1 CLT, not any 8.x CLT (which doesn't exist for 10.11). Presumably from that they're perfectly happy for people to mix & match Xcode/CLT versions going forwards.

Also worth noting, if you only have the CLT and not Xcode, you won't be forced to update anything.

chdiza: > Presumably from that they're perfectly happy for people to mix & match Xcode/CLT versions going forwards.

I thought we'd been having a bunch of issues resulting from such mix-n-match on El Cap, and that people were being told "don't use Xcode 8 with the 7.3 CLT on El Cap".  Maybe I'm misremembering?  If I'm not, then we will be telling people "don't use Xcode8 with the 7.3 CLT", and also "don't use Xcode 7.3.1 with the 7.3 CLT", which amounts to "don't use the CLT on El_Capitan unless you want to forego Xcode".

DomT4: I thought it was dodgy, but if Apple themselves are pushing it in the Xcode 8.1 release notes I'm presuming they now consider it a valid configuration. I think the situation on 10.11 was that people _should_ install the CLT if they've upgraded to Xcode 8? Misty pushed a PR on that earlier this week IIRC.

simonbyrne:
Any chance you could make the help message clearer? I just spent 10 minutes trying to figure out why homebrew wanted me to upgrade to Xcode 8, but there were no Xcode 8 command line tools.

DomT4:
What specifically about the message that you saw was confusing? The Xcode one in particular is specific to the point of perhaps being overkill, but maybe there's something about that that isn't worded clearly enough.

simonbyrne:
I can't remember what the message was (as I've closed the window), but it was something like "homebrew requires Xcode 8". Since the only Xcode requirement of homebrew is the command line tools (and I didn't want to download 4GB right this moment), I took that to mean that I needed the Xcode 8 command line tools. 

Anyway, my solution was to uninstall the Xcode.app, which seemed to fix the problem.

DomT4:
Basically if you have Xcode installed we want it to be the latest version available for your platform, since that's what we test/build/etc against and know works _(or doesn't, in some cases)_. I'll fake a new Xcode version locally & double check what the message is; I thought it was quite comprehensive but perhaps we're printing something else.

simonbyrne:
Ah, I dug the app out of my trash. The message was:
```
├ brew upgrade
Error: Your Xcode (7.3.1) is outdated
Please update to Xcode 8.0.
```
Firstly, I couldn't actually update it from the App Store (despite much clicking of refresh). There just wasn't an option (if I searched for Xcode it said it was version 8, but when I opened it, it was 7.3.1). Then I gave up and googled "download xcode", which sent me to https://developer.apple.com/xcode/. Eventually I found the other downloads page, but couldn't understand why there is no Xcode 8 CLT for el capitan. 

If you're going to insist on upgrading, then maybe say something like "Please upgrade or remove Xcode"

simonbyrne:
or maybe: "Note: there are currently no Xcode 8.0 Command Line Tools available for El Capitan, you will need to either upgrade to Xcode 8.0, or remove Xcode and just use the command line tools"

ilovezfs:
>or maybe: "Note: there are currently no Xcode 8.0 Command Line Tools available for El Capitan, you will need to either upgrade to Xcode 8.0, or remove Xcode and just use the command line tools. Xcode can be updated from the App Store, or from https://developer.apple.com/download/"

That would be incorrect information.

PR #359, opened 2016-06-14
macOS: rebranding

DomT4:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you successfully run `brew tests` with your changes locally?

-----

This doesn't change everything, obviously, but I really should be in bed right now so the logical step was apparently to quickly write up a layer to determine whether a release should be referred to as `OS X` or `macOS`. Seriously.

But if people don't hate the idea I'll expand it a bit more. And yes, I'm aware I'm a giant, giant, giant, giant, giant, giant, _giant_ pedant.

PR #964, opened 2016-09-15
macOS: rebranding

### LINKED ISSUE #961: Adapt to macOS/OS X naming convention in diagnostics
There are 3 string templates including `OS X` in diagnostics:
`"You are using OS X #{MacOS.version}."`
`"OS X: #{MacOS.full_version}-#{kernel}"`
`"To restore it, you must reinstall OS X or restore the binary from"`

As Apple Inc. has renamed the desktop operating system from `OS X` to `macOS`, these diagnostic information will confuse people as time goes on, especially for the new-generation developers who learn developing from `macOS` 10.12.

## Proposed Solution
1. Add a utility function to determine os name from os version. Basically it returns `OS X` when `MacOS.version < :sierra` otherwise `macOS`. 
1. Use this function to replace fixed `OS X` string.

JLHwung:
This PR is a revived version of #359. It replace all `OS X` reference to `#{MacOS.os_name}` except that

- The files in the `vendor` folder
- The files in the `cask` folder
- The files without a `.rb` extension
- Reference of a specific version e.g. `OS X 10.8.2`, `OS X Yosemite`

It also replace all `OS X` reference to `macOS` on all comments except for reference of a specific version.

This PR fixes #961.

There are 2 `OS X` references in `brew.sh`. The tricky part is that even `/usr/bin/sw_ver -productName` returns `Mac OS X` on macOS 10.12, so I don't know a safe way (except if-else on productVersion) to replace `OS X` string in shell scripts.

There is similar pull request #359, this PR covers the works on #359.

-----


Library/Homebrew/os/mac.rb
```diff
@@ -37,6 +37,15 @@ def outdated_release?
       version < "10.9"
     end
 
+    def os_name
```
DomT4: You'll need to actually pick up the changes I suggested [here](https://github.com/Homebrew/brew/issues/961#issuecomment-247051002). Plain version comparisons like this don't work especially well, you get situations where `10.10` is correctly reported as `OS X` but `10.7` is flagged as `macOS`.

The proposed change in the linked issue also allows you to fix the situation where Requirements have their own `@version` we shouldn't be ignoring. My proposed fix in the other issue isn't perfect, and will need some additional testing & tweaking, but it's closer.

JLHwung: I'd better write a test for that.

Library/Homebrew/formula_support.rb
```diff
@@ -28,13 +28,14 @@ def valid?
 
   def to_s
     return @explanation unless @explanation.empty?
+    os_name = OS.mac? ? MacOS.os_name : 'macOS'
```
reitermarkus: Is this what it's supposed to be doing? If `OS.mac?` is `false`, how can it be `'macOS'`?

JLHwung: I use `MacOS.os_name` before but CI complains that `MacOS` is uninitialized constant. Test always pass locally. So I thought may be it's the case when `OS.mac` is false?

Library/Homebrew/os/mac.rb
```diff
@@ -37,6 +37,15 @@ def outdated_release?
       version < "10.9"
     end
 
+    def os_name(version = @version)
+      # TODO: replace os_name by constant string "macOS" when macOS 10.14 is released
+      if Version.create(version) >= Version.create("10.12")
```
reitermarkus: I think you could just use `version >= "10.12"` here.

Library/Homebrew/formula_support.rb
reitermarkus: Seems arbitrary that this is the only place where this is happening.

Library/Homebrew/os/mac.rb
```diff
@@ -37,6 +37,15 @@ def outdated_release?
       version < "10.9"
     end
 
+    def os_name(version = @version)
```
MikeMcQuaid: Could this maybe just be called `name`?

Library/Homebrew/formula_support.rb
MikeMcQuaid: That's because you need to not use `MacOS` (or `OS.mac?`) outside of `Library/Homebrew/os` or  `Library/Homebrew/extend/os/`. Instead, have an `OS.name` method which uses subclasses in `Library/Homebrew/extend/os/mac` to set the right macOS name.

Alternatively, edit `brew.sh` to set a `HOMEBREW_OS_NAME` (like `HOMEBREW_OS_VERSION` currently) which uses the correct version and then `OS.name` can just use `ENV["HOMEBREW_OS_NAME"]`

Shout if you need more help with this!

Library/Homebrew/os/mac.rb
JLHwung: Like @DomT4 [said before](https://github.com/Homebrew/brew/pull/964#r78884726), plain version comparison doesn't work well as `"10.7" > "10.12"` in the plain string comparison.

Library/Homebrew/formula_support.rb
JLHwung: @MikeMcQuaid I think why CI complains is that when `ENV["HOMEBREW_TEST_GENERIC_OS"]` is set, `MacOS` is not initialized but we can use `OS::Mac.os_name(ENV["HOMEBREW_OS_VERSION"])` here so that both CI and macOS should be happy.

> Alternatively, edit brew.sh to set a HOMEBREW_OS_NAME (like HOMEBREW_OS_VERSION currently) which uses the correct version and then OS.name can just use ENV["HOMEBREW_OS_NAME"]

It think this is over complicated. Basically I aim to adding the pedantry in a single place so that when macOS 10.14 is released we don't have to search everywhere to remove multiple methods and their usage.

Library/Homebrew/os/mac.rb
JLHwung: @MikeMcQuaid I think `name` is a bit too general to use as a method name 😄. Well to tell you the truth I just find it cumbersome to rewrite all the commit messages from "use os_name" to "use name".

MikeMcQuaid: In that case you can avoid making `version` a `Version` if it already is one and `10.12` doesn't need to be a `Version`.

Library/Homebrew/formula_support.rb
```diff
@@ -28,13 +28,14 @@ def valid?
 
   def to_s
     return @explanation unless @explanation.empty?
+    os_name = OS::Mac.os_name(ENV["HOMEBREW_OSX_VERSION"])
```
MikeMcQuaid: Sorry, this needs to use `ENV["HOMEBREW_OS_VERSION"]` (not `OSX_VERSION`) and be in the OS class so it works on non-OS X platforms. Alternatively, all the OS X references in code need to be moved into `extend/os`.

Library/Homebrew/os/mac.rb
JLHwung: @MikeMcQuaid use `Version` here because
- we need to pass `Version` to `os_name` on `macos_requirement`
- `"10.7" > "10.12"` so plain string comparison is not sufficient

Library/Homebrew/formula_support.rb
JLHwung: I think I'd better revert this change to use `MacOS.os_name` unconditionally, and require `os/mac` on `test_formula_support`. Apparently the error message is meaningless on a Linux, isn't it?

Library/Homebrew/os/mac.rb
MikeMcQuaid:
```ruby
version = Version(version) unless version.is_a? Version
```

Library/Homebrew/os/mac.rb
MikeMcQuaid:
```ruby
version = Version(version) unless version.is_a? Version
version >= "10.12"
```

will work

Library/Homebrew/formula_support.rb
MikeMcQuaid: Yep, ideally we'd not have any user-facing messages that say "OS X" outside of `Library/Homebrew/os/mac/` or `Library/Homebrew/extend/os/mac.`

Library/Homebrew/os/mac.rb
JLHwung: @MikeMcQuaid Ahh, this time I get your point. Thank you.

Library/Homebrew/formula_support.rb
JLHwung: So I think when `test_formula_support` makes assertions on the user-facing messages, it should require `os/mac` otherwise it does not simulate the real cases.
Apparently we are facing same `os/mac` loading problem if we write a test for `KegOnlyReason::valid` method.

MikeMcQuaid: `os/mac` should not be used/needed outside of those two directories.

JLHwung: @MikeMcQuaid Yep, I mean requiring `os/mac` on `test_formula_support`, just like what we do in `test_os_mac_language`.

Library/Homebrew/os/mac.rb
JLHwung: @MikeMcQuaid `version >= "10.12"` will fail because `ArgumentError: comparison of Version with String failed`

MikeMcQuaid: Check the methods above; if you create a version the same as them (probably with `.new`) then you should be able to compare the same as them.

MikeMcQuaid: Or you may need to use/try `>` rather than `>=`

JLHwung: @MikeMcQuaid Thank you! `Version.new` LGTM, I'd better add more test to ensure it does not break.

Library/Homebrew/formula_support.rb
JLHwung: @MikeMcQuaid How about I move `Class KegOnlyReason` in `formula_support` to `/extend/os/mac` (as we do for `formula_cellar_checks`) ? As the whole class is macOS-only and ideally we should not have it outside `Library/Homebrew/os/mac`, right? I am wondering how to specify that this method is macOnly and instruct the test should go into the macOs-only branch.

JLHwung: @MikeMcQuaid If necessary I can also move other `OS X` referencing user-facing messages to `/extend/os/mac`.

MikeMcQuaid: @JLHwung That'd be good, thanks!

JLHwung: @MikeMcQuaid CI complains that KegOnlyReason is uninitialized on linuxbrew because I move to os/mac. How would this class be used on Linux?

UniqMartin:
As mentioned in https://github.com/Homebrew/brew/issues/961#issuecomment-247779301 by now I'm strongly in favor of maximum simplicity and using “macOS” everywhere as soon as 10.12 is released. However, if we really wanted to be serious about adapting our output, the distinction will need to be three-fold (“Mac OS X” for 10.0-10.6, “OS X” for 10.7-10.11, and “macOS” for 10.12+).

MikeMcQuaid:
> As mentioned in #961 (comment):
> The API reference at https://developer.apple.com/reference/ only ever speaks of “macOS” when talking about when an API was introduced, even if it's “macOS 10.0+”.

This is by far the strongest argument for me so I'm 👍 with just using macOS all over.

@JLHwung I'm so sorry to send you on such a wild chase for this, my apologies. I hope you've learnt some stuff through doing so and will continue to contribute to Homebrew. Let me know if you'd rather doing the mass-`macOS` replacement or if you'd rather we did. Thanks so much for your contributions.

JLHwung:
@MikeMcQuaid Sorry for late reply as I am in vacation now. I've learnt a lot more than it appears at the beginning, and thank you for your review. I cannot spare time on replacing before 20th so it'd better assigned to other people.

As I am working on this pr, I find many macOS-dependent code does not locate in os/mac or extend/os/mac, shall we split the platform-dependent codes or just let it be?

UniqMartin:
Thanks you for your work here!

> As I am working on this pr, I find many macOS-dependent code does not locate in os/mac or extend/os/mac, shall we split the platform-dependent codes or just let it be?

This is an ongoing process and we started it fairly recently in an attempt to make Homebrew truly platform-agnostic (to the extent it makes sense) and to help move back code from Linuxbrew into Homebrew with proper abstractions and cleaner code structure. If that's a topic that interests you, we're more than happy to have someone help us with this endeavor.

PR #965, opened 2016-09-15
Warn when using Xcode 8 without CLT on 10.11

mistydemeo:

-----

Xcode 8 on 10.11 has more issues than the usual mismatched SDK problems, and we're going to get a decent degree of breakage. I think it's important that `brew doctor` tell people to install the CLT in this case.

It may be worth rethinking whether we want to support the Xcode-only-without-OS-specific-SDK usecase going forward. We've put a good amount of work into supporting Xcode-only over the years, but we're getting increasing amounts of pain for our efforts. I'd be happy to turn this PR into a blanket warn for that usecase, instead of something that only covers Xcode 8 on 10.11.

cc @DomT4

### DIFF
--- a/Library/Homebrew/extend/os/mac/diagnostic.rb
+++ b/Library/Homebrew/extend/os/mac/diagnostic.rb
@@ -9,6 +9,7 @@ module Homebrew
           check_for_installed_developer_tools
           check_xcode_license_approved
           check_for_osx_gcc_installer
+          check_xcode_8_without_clt_on_el_capitan
         ]
       end
 
@@ -87,6 +88,19 @@ module Homebrew
         <<-EOS.undent
           A newer Command Line Tools release is available.
           #{MacOS::CLT.update_instructions}
+        EOS
+      end
+
+      def check_xcode_8_without_clt_on_el_capitan
+        return unless MacOS::Xcode.without_clt?
+        # Scope this to Xcode 8 on El Cap for now
+        return unless MacOS.version == :el_capitan && MacOS::Xcode.version >= "8"
+
+        <<-EOS.undent
+          You have Xcode 8 installed without the CLT;
+          this causes certain builds to fail on OS X 10.11.
+          Please install the CLT via:
+            sudo xcode-select --install
         EOS
       end
 
--- a/Library/Homebrew/test/test_os_mac_diagnostic.rb
+++ b/Library/Homebrew/test/test_os_mac_diagnostic.rb
@@ -38,4 +38,12 @@ class OSMacDiagnosticChecksTest < Homebrew::TestCase
     MacOS::XQuartz.stubs(:version).returns("2.7.10_beta2")
     assert_match "The following beta release of XQuartz is installed: 2.7.10_beta2", @checks.check_for_beta_xquartz
   end
+
+  def test_check_xcode_8_without_clt_on_el_capitan
+    MacOS.stubs(:version).returns OS::Mac::Version.new("10.11")
+    MacOS::Xcode.stubs(:installed?).returns true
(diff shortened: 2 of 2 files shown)


Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -90,6 +91,18 @@ def check_clt_up_to_date
         EOS
       end
 
+      def check_xcode_8_without_clt_on_el_cap
```
MikeMcQuaid: Can we go for `el_capitan` here?

Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -90,6 +91,18 @@ def check_clt_up_to_date
...
+        return unless MacOS::Xcode.without_clt?
+        # Scope this to Xcode 8 on El Cap for now
+        return unless MacOS.version == :el_capitan && MacOS::Xcode.version >= "8"
+
+        <<-EOS.undent
+          You have Xcode 8 installed without the CLT; this causes certain builds to fail.
```
MikeMcQuaid: Worth spelling out this is El Capitan-only to avoid scaring Sierra people?

mistydemeo: The warning only prints on El Cap; are you worried about copy/pasted versions of it on the web being misinterpreted?

Library/Homebrew/extend/os/mac/diagnostic.rb
mistydemeo: Updated.

Library/Homebrew/extend/os/mac/diagnostic.rb
MikeMcQuaid: Yeh, or people reading this message when they have a mixed environment and thinking it applies everywhere.

DomT4: Agreed it's probably best to explicitly mention El Cap. It'll only be a matter of time until someone sticks the message on stackoverflow or something without much context 🙈.

Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -90,6 +91,18 @@ def check_clt_up_to_date
...
 
+      def check_xcode_8_without_clt_on_el_capitan
+        return unless MacOS::Xcode.without_clt?
+        # Scope this to Xcode 8 on El Cap for now
+        return unless MacOS.version == :el_capitan && MacOS::Xcode.version >= "8"
+
+        <<-EOS.undent
+          You have Xcode 8 installed without the CLT; this causes certain builds to fail.
+          Please install the CLT via:
+            xcode-select --install
```
DomT4: Maybe a good idea to recommend `sudo` here as `xcode-select` will only abort unless you give it elevated privs.

If we're using a vanilla non-sudo recommendation everywhere else though feel free to ignore me. No point introducing an inconsistency.

mistydemeo: It's totally inconsistent now I look at it - we have messages using both with and without. 😰 I'll update this now, and make a PR to update other parts for consistency later.

Library/Homebrew/extend/os/mac/diagnostic.rb
mistydemeo: Updated!

MikeMcQuaid:
@mistydemeo Is this still a problem? I'm wondering now the storm is over if it's worth reverting this. Alternatively, we wait until the 8.2 CLT on 10.11 (which is coming) and do it then.

CC @ilovezfs who talked about this with me.

PR #970, opened 2016-09-15
superenv: help Autotools with 10.12 SDK on 10.11

ilovezfs:
The GNU Autotools tests for whether a given symbol is defined are
reliably coming to incorrect conclusions on 10.11 with the 10.12 SDK
in Xcode 8. This overrides its decisions by forcing the right answer
in superenv using ac_cv_func_* environment variables and setting them to
"no" on 10.11. The list of problematic symbols is from

  grep 'weak$os10.11' MacOSX.sdk/usr/lib/system/libsystem_c.tbd

### DIFF
--- a/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
@@ -94,6 +94,15 @@ module Superenv
       self["SDKROOT"] = MacOS.sdk_path
     end
 
+    # Filter out symbols known not to be defined on 10.11 since GNU Autotools
+    # can't reliably figure this out with Xcode 8 on its own yet.
+    if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+      %w[basename_r clock_getres clock_gettime clock_settime dirname_r
+         getentropy mkostemp mkostemps].each do |s|
+        ENV["ac_cv_func_#{s}"] = "no"
+      end
+    end
+
     # On 10.9, the tools in /usr/bin proxy to the active developer directory.
     # This means we can use them for any combination of CLT and Xcode.
     self["HOMEBREW_PREFER_CLT_PROXIES"] = "1" if MacOS.version >= "10.9"


Library/Homebrew/extend/os/mac/extend/ENV/super.rb
```diff
@@ -94,6 +94,17 @@ def setup_build_environment(formula = nil)
       self["SDKROOT"] = MacOS.sdk_path
     end
 
+    # Filter out symbols known not to be defined on 10.11 since GNU Autotools
+    # can't reliably figure this out with Xcode 8 on its own yet. The Xcode
+    # version doesn't need to be checked since these are harmless with Xcode 7
+    # even if they're unnecessary.
```
MikeMcQuaid: Personally I'd add the Xcode check anyway because better safe than sorry but will leave that decision up to you.

ilovezfs:
```
if MacOS.version == "10.11" && DevelopmentTools.clang_build_version >= 800
```
yes?

MikeMcQuaid: Yeh, or you can do `MacOS::Xcode.version >= "8.0"`

ilovezfs: My concern with that is that it won't fix the problem on a CLT-only install assuming Apple puts the same stuff in the 10.11 Xcode 8 CLT (if there is one actually coming out).

MikeMcQuaid: Ah, I see. Yes, your version is better then. CC @DomT4 to confirm about the CLT?

DomT4: I'm not sure about the CLT on 10.11 yet to be honest because Apple releases CLT packages scoped to individual platforms, i.e. there's one available for macOS Sierra but not El Capitan, but when El Capitan gets its version it'll get its own `.pkg`.

It seems feasible they could push a CLT for 10.11 without the problems, but who knows until they do.

ilovezfs: let's avoid speculatively fixing bugs and just do
```
if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version.to_i >= 8
```

We can always expand its scope later as needed.

MikeMcQuaid: Agreed except avoid the `.to_i` and just do `MacOS::Xcode.version >= "8.0"` 😉

ilovezfs: Shall I change https://github.com/ilovezfs/brew/blob/2505da3182f48c960e896f42173b4d9809e01001/Library/Homebrew/extend/os/mac/extend/ENV/super.rb#L92 as well? hehe

MikeMcQuaid: Yeh, feel free 😁

ilovezfs:
PR refreshed.

>Part of me wonders if it's better to add this change as an unique method like determine_cccfg, etc. I can't imagine this problem getting smaller in future years, and it may be easier to keep track of that way.

It may eventually benefit from something like that, but I think for now that's probably over-engineering it.

DomT4:
👍. I'm happy unless you're planning to add a test for this change or anything. Otherwise, let's ship it and see what breaks.

ilovezfs:
:rocket:

DomT4:
Can confirm this at least fixes `xz`'s runtime failure of false `empty archive` messages & installing `python` on 10.11 with Xcode 8.

People will need to reinstall `xz` and other impacted-at-run formulae by the looks of it.

ilovezfs:
>People will need to reinstall xz and other impacted-at-run formulae by the looks of it.

indeed.

chdiza:
I got a malfunctioning `xz` with Xcode 8 on *Sierra*, which also was fixed by this PR (as far as I can tell).

chdiza:
That is to say: even with Xcode 8 *and* the associated CLT, on Sierra, `xz` gave me the "empty archive" nonsense and any formula I built whose tarball was an xz-ball was defective ("illegal instruction 4").

But after this PR it seemed to work.

zmwangx:
This PR specifically has a `MacOS.version == "10.11"` guard, so I doubt it should have any effect on your Sierra environment.

ilovezfs:
@chdiza what's `brew config` have to say for itself?

chdiza:
`brew --config` has this to say:
```
HOMEBREW_VERSION: 0.9.9
ORIGIN: https://github.com/Homebrew/brew
HEAD: a148aa3a41d34a07caab5d17daa26b4749ff3c50
Last commit: 15 hours ago
Core tap HEAD: 0e99665f347c1554f55fbecf448956a44ccd99e5
Core tap last commit: 14 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/hb
CPU: quad-core 64-bit ivybridge
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.8.4 => /Applications/Xcode.app/Contents/Developer/usr/bin/git
Perl: /usr/bin/perl
Python: /usr/bin/python
Ruby: /usr/bin/ruby => /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby
Java: N/A
OS X: 10.12-x86_64
Xcode: 8.0
CLT: 8.0.0.0.1.1472435881
X11: N/A
```

Maybe it wasn't this particular commit that fixed the problem, but another one that landed within around 24 hours.

PR #982, opened 2016-09-17
formula: set TERM to dumb during test

zmwangx:

-----

Software that tries to print different things (e.g. ANSI color) based on different termcaps often relies on the `TERM` environment variable, and could fail without it. This results in confusing test issues where certain tests can be successfully run by users testing locally with `TERM`
set, but choke up on our CI.

Always setting TERM to dumb leads to better consistency between local tests and CI tests, and saves much probing in certain cases.

---

Some examples:

- `aalib`: Existing test [fails on the CI](https://bot.brew.sh/job/Homebrew%20Testing/1084/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/test_aalib/) without TERM — I was investigating the failure just now;

- `hopenpgp-tools`: [I was confused by this just a short while ago](https://github.com/Homebrew/homebrew-core/pull/3997#issuecomment-240604992);

- In core:

    ```
$ grep -rI 'ENV\["TERM"\]'
Formula/asciiquarium.rb:    ENV["TERM"] = "xterm"
Formula/browser.rb:    ENV["TERM"] = "xterm"
Formula/bvi.rb:    ENV["TERM"] = "xterm"
Formula/clipsafe.rb:    ENV["TERM"] = "dumb"
Formula/diff-so-fancy.rb:    ENV["TERM"] = "xterm"
Formula/dshb.rb:    ENV["TERM"] = "xterm"
Formula/ee.rb:    ENV["TERM"] = "xterm"
Formula/eg.rb:    ENV["TERM"] = "xterm"
Formula/git-test.rb:    ENV["TERM"] = "xterm"
Formula/hopenpgp-tools.rb:    ENV["TERM"] = "dumb"
Formula/htop-osx.rb:    ENV["TERM"] = "xterm"
Formula/htop.rb:    ENV["TERM"] = "xterm"
Formula/jp2a.rb:    ENV["TERM"] = "xterm-256color"
Formula/le.rb:    ENV["TERM"] = "xterm"
Formula/mobile-shell.rb:    ENV["TERM"] = "xterm"
Formula/multitail.rb:    ENV["TERM"] = "xterm"
Formula/ne.rb:    ENV["TERM"] = "xterm"
Formula/newt.rb:    ENV["TERM"] = "xterm"
Formula/pwntools.rb:    ENV["TERM"] = "xterm"
Formula/slrn.rb:    ENV["TERM"] = "xterm"
Formula/sngrep.rb:    ENV["TERM"] = "xterm"
Formula/tnote.rb:    ENV["TERM"] = "xterm"
Formula/watch.rb:    ENV["TERM"] = "xterm"
Formula/weechat.rb:    ENV["TERM"] = "xterm"
```

    So many formulae require a workaround that could be easily resolved in general. Setting `TERM` to `xterm` is also "wrong" because on the CI the logging facilities definitely don't have xterm capabilities. If a formula specifically requires xterm in order to work, it can still be overridden in the test block as before.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1434,8 +1434,10 @@ class Formula
     old_tmpdir = ENV["TMPDIR"]
     old_temp = ENV["TEMP"]
     old_tmp = ENV["TMP"]
+    old_term = ENV["TERM"]
     ENV["CURL_HOME"] = old_curl_home || old_home
     ENV["TMPDIR"] = ENV["TEMP"] = ENV["TMP"] = HOMEBREW_TEMP
+    ENV["TERM"] = "dumb"
     mktemp("#{name}-test") do |staging|
       staging.retain! if ARGV.keep_tmp?
       @testpath = staging.tmpdir
@@ -1457,6 +1459,7 @@ class Formula
     ENV["TMPDIR"] = old_tmpdir
     ENV["TEMP"] = old_temp
     ENV["TMP"] = old_tmp
+    ENV["TERM"] = old_term
   end
 
   # @private

### CODE REVIEW

**Comment by ilovezfs** [comment]
Another one I see like this is `ENV["LC_ALL"] = "en_US.UTF-8"`

**Comment by zmwangx** [comment]
```
$ grep -rI 'ENV\["LC_ALL"\]'
Formula/alot.rb:      ENV["LC_ALL"] = "en_US.UTF-8"
Formula/asciinema.rb:    ENV["LC_ALL"] = "en_US.UTF-8"
Formula/cmake.rb:      ENV["LC_ALL"] = "en_US.UTF-8"
Formula/khal.rb:    ENV["LC_ALL"] = "en_US.UTF-8"
Formula/lensfun.rb:    ENV["LC_ALL"] = "en_US.UTF-8"
Formula/mpv.rb:    ENV["LC_ALL"] = "C"
Formula/percona-xtrabackup.rb:      ENV["LC_ALL"] = "en_US.UTF-8"
Formula/pixz.rb:    ENV["LC_ALL"] = "en_US.UTF-8"
Formula/twtxt.rb:    ENV["LC_ALL"] = "en_US.UTF-8"
Formula/vdirsyncer.rb:    inreplace prefix/"vdirsyncer.plist", "@@LOCALE@@", ENV["LC_ALL"] || ENV["LANG"] || "en_US.UTF-8"
Formula/vdirsyncer.rb:    ENV["LC_ALL"] = "en_US.UTF-8"
$ grep -rI 'ENV\["LANG"\]'
Formula/khal.rb:    ENV["LANG"] = "en_US.UTF-8"
Formula/lesstif.rb:    ENV["LANG"] = "C"
Formula/mitmproxy.rb:    ENV["LANG"] = "en_US.UTF-8"
Formula/twtxt.rb:    ENV["LANG"] = "en_US.UTF-8"
Formula/vdirsyncer.rb:    inreplace prefix/"vdirsyncer.plist", "@@LOCALE@@", ENV["LC_ALL"] || ENV["LANG"] || "en_US.UTF-8"
```

Reasonable.

**Comment by ilovezfs** [comment]
I guess the argument against globally setting the LC_ALL is that locally people probably don't have that set either.

**Comment by zmwangx** [comment]
"Reasonable" expands to "your suggestion is reasonable."

**Comment by zmwangx** [comment]
Oh I see what you're saying.

**Comment by ilovezfs** [comment]
probably best to stick to just TERM for now I guess

**Comment by zmwangx** [comment]
That won't be a problem once we have https://github.com/Homebrew/brew/issues/932 (although I haven't thought through all the ramifications it might have).

**Comment by zmwangx** [comment]
@MikeMcQuaid I did test on a bunch of listed formulae locally (not all) and I don't see any problem moving from `xterm` to `dumb`. CI might disagree but I highly doubt it.

This change isn't noticeable to users so I think we can merge first and test the affected formulae later. Even if a formula or two do require `xterm`, as I said above we can still override in the test block.

**Comment by MikeMcQuaid** [comment]
@zmwangx Cool :shipit:

**Comment by apjanke** [comment]
Agreed. This sounds like a good approach. And since most users won't see it, seems reasonable to merge first and test primarily through the CI.

**Comment by UniqMartin** [comment]
I agree with everyone else. Good move and good work! :tada:

**Comment by zmwangx** [comment]
To wrap this up: I've removed `ENV["TERM"]` from tests where possible, and that leaves us with seven formulae that require `xterm`:

```bash
$ grep -rIl 'ENV\["TERM"\]'
Formula/asciiquarium.rb
Formula/ee.rb
Formula/eg.rb
Formula/git-test.rb
Formula/ne.rb
Formula/newt.rb
Formula/slrn.rb
```

which is a still a good thing because it's now possible to determine locally that those settings are required; before this PR you can remove the `ENV["TERM"]` line from the tests, the tests would still work locally in xterm-capable terminals, but they'll fail (sometimes mysteriously) on the CI.

Note: To answer the possible question of why we don't set `TERM` to `xterm` universally: masquerading as xterm while not having the corresponding termcaps is a dangerous thing to do in general.

PR #990, opened 2016-09-17
no_weak_imports: always check for support

ilovezfs:
ENV.no_weak_imports should be a no-op when Xcode doesn't support the
feature to avoid breaking builds with <= Xcode 7.


ilovezfs:
CC @MikeMcQuaid 

See failed Yosemite build here: https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/7874/

zmwangx:
👍

nilstgmd:
That is exactly what I need. 👍 @ilovezfs

PR #1001, opened 2016-09-18
weak imports: skip CLT check if Xcode is installed

ilovezfs:
Xcode 8 plus CLT 7 should return true for weak imports support.

Thanks to @JCount for spotting this bug.

### DIFF
--- a/Library/Homebrew/extend/os/mac/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/shared.rb
@@ -2,14 +2,14 @@ module SharedEnvExtension
   def no_weak_imports_support?
     return false unless compiler == :clang
 
-    if MacOS::Xcode.version && MacOS::Xcode.version < "8.0"
-      return false
+    if MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+      return true
     end
 
-    if MacOS::CLT.version && MacOS::CLT.version < "8.0"
-      return false
+    if MacOS::CLT.version && MacOS::CLT.version >= "8.0"
+      return true
     end
 
-    true
+    false
   end
 end


MikeMcQuaid:
Does depend on the active developer directory?

MikeMcQuaid:
i.e. `xcode-select --print-path`

ilovezfs:
No ... developer directory is the default 
```
iMac-TMP:local joe$ xcode-select --print-path
/Applications/Xcode.app/Contents/Developer
```

Without this PR, or an equivalent change, you get `false` from the existing code if both Xcode and CLT are installed since 7 < 8 for the CLT.

MikeMcQuaid:
@ilovezfs Sorry, I mean: what if it's not the default and it's pointing to the CLT?

ilovezfs:
current code without PR:
false clt only
false clt + xcode, xcode-select pointing at xcode
false clt + xocde, xcode-select pointing at clt

with PR:
false clt only
true clt + xcode, xcode-select pointing at xcode
true clt + xcode, xcode-select pointing at clt

ilovezfs:
also I imagine this code is probably preferable
```
    if MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
        return true
    end

    if MacOS::CLT.version && MacOS::CLT.version >= "8.0"
      return true
    end
    false
```

ilovezfs:
Interestingly, with or without this PR, having your xcode-select pointing at `/Library/Developer/CommandLineTools` with both Xcode and CLT installed dies in a ball of flames:
```
==> ./configure --prefix=/usr/local/Cellar/ffmpeg/3.1.3 --enable-shared --enable-pthreads --enable-gpl --enable-version3 --enable-hardcoded-tables --enable-avresample --cc=clang --host-cflags= --host-ldflags= --enable-opencl --enable-libx264 --enable-libmp3lame --enable-libxvid --disable-lzma --enable-vda
clang is unable to create an executable file.
If clang is a cross-compiler, use the --enable-cross-compile option.
Only do this if you know what cross compiling means.
C compiler test failed.

If you think configure made a mistake, make sure you are using the latest
version from Git.  If the latest version fails, report the problem to the
ffmpeg-user@ffmpeg.org mailing list or IRC #ffmpeg on irc.freenode.net.
Include the log file "config.log" produced by configure as this will help
solve the problem.
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: ./configure --prefix=/usr/local/Cellar/ffmpeg/3.1.3 --enable-shared --enable-pthreads --enable-gpl --enable-version3 --enable-hardcoded-tables --enable-avresample --cc=clang --host-cflags= --host-ldflags= --enable-opencl --enable-libx264 --enable-libmp3lame --enable-libxvid --disable-lzma --enable-vda
1. raise
2. ignore
3. backtrace
4. irb
5. shell
```

ilovezfs:
@MikeMcQuaid in any case, since there seems to be some hesitation here I may just go with the ridiculous fix for now
```
    inreplace ["libavdevice/v4l2.c", "libavutil/time.c"], "HAVE_CLOCK_GETTIME",
                                                          "POKEMON"
```

zmwangx:
Recent Pokemon games have clock gettime too.

ilovezfs:
Ah, then we can do
```
diff --git a/Formula/ffmpeg.rb b/Formula/ffmpeg.rb
index 1b60d4b..4c03701 100644
--- a/Formula/ffmpeg.rb
+++ b/Formula/ffmpeg.rb
@@ -91,6 +91,12 @@ class Ffmpeg < Formula
   end
 
   def install
+    # Fixes "dyld: lazy symbol binding failed: Symbol not found: _clock_gettime"
+    if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+      inreplace %w[libavdevice/v4l2.c libavutil/time.c], "HAVE_CLOCK_GETTIME",
+                                                         "UNDEFINED_GIBBERISH"
+    end
+
     args = %W[
       --prefix=#{prefix}
       --enable-shared
```

ilovezfs:
:green_apple: but since this will make no_weak_imports actually work without breaking 10.11, and therefore encourage its use ...

UniqMartin:
Checking the CLT and/or Xcode version is probably not the best solution when checking for linker flag support. The problem is that this feature depends on a sufficiently recent `ld64` (the project behind the static linker `ld`); for `-no_weak_imports` it's version 274 or newer to be precise.

However, the static linker is rarely invoked directly and most often via some frontend like `clang`, `gcc`, etc. The problem here is that it's not always obvious which frontend will pick which version of `ld` (and from where). E.g. who can immediately answer what a Homebrew-provided GCC will use? (Spoiler: I cannot.) The first safe thing I can think of is to do trial compilations (`autoconf` style), but that's hardly a satisfying solution.

> Interestingly, with or without this PR, having your xcode-select pointing at `/Library/Developer/CommandLineTools` with both Xcode and CLT installed dies in a ball of flames:

That's a problem we might want to address and that is explained by the above paragraph. Though this should be a transient issue that only arises now because we've ended up in a situation where CLT and Xcode versions don't match (thanks Apple!), something we don't usually have.

ilovezfs:
Things we've learned today which may or may not change at a later date:
1. no one should be using ENV.no_weak_imports at this point
2. anyone with Xcode 8 and CLT 7.3.1 should `sudo xcode-select -r` if they set it to the CLT.

PR #1002, opened 2016-09-18
do not show curl progress bar in TTYs < 80 chars wide

### LINKED ISSUE #981: Progress information always print in new line.
Progress information (#) should be in one line, but the latest brew always print # in new line.

screen shot 2016-09-17 at 11 24 31 am

### Homebrew version :
```
Homebrew 0.9.9 (git revision a54c; last commit 2016-09-16)
Homebrew/homebrew-core (git revision 968d; last commit 2016-09-16)
```

apjanke:
Fixes #981

It looks like `curl` outputs a progress bar at a fixed 80 column width. This means that if you run `brew` with a `fetch` or other `curl`-using action in a terminal less than 80 columns wide, the progress bar gets copied to multiple lines, messing up the output.

This PR disables `curl` progress output when it's run in a terminal narrower than 80 columns.

Before:

screen shot 2016-09-17 at 11 19 40 pm

After:

screen shot 2016-09-17 at 11 57 23 pm

### DIFF
--- a/Library/Homebrew/utils/curl.rb
+++ b/Library/Homebrew/utils/curl.rb
@@ -8,10 +8,9 @@ def curl_args(extra_args = [])
 
   flags = HOMEBREW_CURL_ARGS
   flags -= ["--progress-bar"] if ARGV.verbose?
-
   args = [curl.to_s] + flags + extra_args
   args << "--verbose" if ENV["HOMEBREW_CURL_VERBOSE"]
-  args << "--silent" if !$stdout.tty? || ENV["TRAVIS"]
+  args << "--silent" if !$stdout.tty? || ENV["TRAVIS"] || Tty.width < 80
   args
 end
 


MikeMcQuaid:
What's the output if you omit `--silent` like? I'm agreed with hiding this for < 80 characters but I wonder if silent is what we want instead.

apjanke:
If you omit both `--silent` and `--progress-bar`, then you get the big multi-line progress meter, which also seems to be output for a fixed 80 column width, and screws up on narrow terminals.

screen shot 2016-09-18 at 7 25 56 pm

And I think that's all of our output options for `curl`.

I'm going to report this upstream, too. But there's zero chance of us getting a fix for it since we're using the system `curl`, which won't be updated until a new ~~OS X~~ macOS release.

MikeMcQuaid:
@apjanke Hmm, I'm not convinced silence is better than the current behaviour. A lot of tools barf if you have <80 characters so I think it may be better to just leave this as-is; it's been Homebrew (and Curl) behaviour for a very long time.

MikeMcQuaid:
Also worth noting: we optimise all our user-output messages for 80 characters so they will all wrap weirdly/unreadably also.

zmwangx:
IMO printing many lines (not everyone has a problem with that) is much better than printing nothing. Especially since it's been that way for a long time. If you suddenly suppress progress information with this semi-random terminal width cutoff, brew will feel broken to some extent, especially when downloading something large, say the GLC bottle.

MikeMcQuaid:
> If you suddenly suppress progress information with this semi-random terminal width cutoff, brew will feel broken to some extent, especially when downloading something large, say the GHC bottle.

A good point 👍

apjanke:
Fair enough. Closing this as withdrawn.

PR #1009, opened 2016-09-18
Fix some completion problems.

chaoren:

-----

* use $HOME instead of ~ so it gets expanded even when quoted.
* add --indicator-style=none in case the user aliased ls to use
  something else and extra indicator characters get added
  (e.g., directory/).

NOTE:
Why you shouldn't parse the output of ls(1)
http://mywiki.wooledge.org/ParsingLs

### DIFF
--- a/etc/bash_completion.d/brew
+++ b/etc/bash_completion.d/brew
@@ -52,7 +52,7 @@ __brew_complete_formulae() {
 
 __brew_complete_installed() {
   local cur="${COMP_WORDS[COMP_CWORD]}"
-  local inst="$(ls "$(brew --cellar)")"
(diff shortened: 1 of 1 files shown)


chaoren:
> Have you written new tests for your changes? Here's an example.

If you have tests for this, please show me where.

zmwangx:
`--indicator-style=none` is not a BSD ls option.

Use of `ls` should simply be removed. There's #901 which supposedly does that but I didn't get to reviewing it (should probably get that done today.)

chaoren:
Whoops, partial dupe with https://github.com/Homebrew/brew/pull/901. Just noticed.

chaoren:
I'll add the $HOME thing to the other PR, and this won't be necessary anymore.

PR #1014, opened 2016-09-18
dependency_collector: Fix ant_dep for Linux

rwhogg:

-----

This upstreams https://github.com/Linuxbrew/brew/pull/12. That pull request reflects that we always use a brewed `ant` on Linuxbrew, even if someone's installed it through `apt-get` or similar. It's not cherry-picked, but I for one am personally okay with a few merge conflicts if it means we can eventually make `linuxbrew/brew` unnecessary.

I'm expecting a bit of debate about the proper ways to upstream various Linuxbrew changes; my apologies if it isn't exactly what you want right off the bat.


Library/Homebrew/test/test_dependency_collector.rb
```diff
@@ -76,6 +76,20 @@ def test_x11_min_version_and_tag
     assert_predicate dep, :optional?
   end
 
+  def test_ant_dep_mavericks_or_newer
+    skip "Only for Mac OS" unless OS.mac?
```
rwhogg: This line (and the similar one in the second test) is what I expect to generate the most contention.

Without it, you get: `NameError: uninitialized constant DependencyCollectorTests::MacOS.`

It's not like these two tests make any sense on Linux, but I did add them at the same time of the original PR.

rwhogg: And now I see https://github.com/Homebrew/brew/pull/543, which pushes similar tests out of here. Perhaps it would be wisest to remove these two tests altogether...

Library/Homebrew/dependency_collector.rb
```diff
@@ -136,7 +136,7 @@ def parse_class_spec(spec, tags)
   end
 
   def ant_dep(spec, tags)
-    if MacOS.version >= :mavericks
+    if MacOS.version >= :mavericks || !OS.mac?
```
MikeMcQuaid: I'm afraid when upstreaming changes you're going to need to do more stylistically involved changes than this. Instead of scattering `if OS.mac?` throughout the codebase the pattern is to make use of `Library/Homebrew/extend/os/linux` to override methods.

rwhogg: No problem! Just wanted to get the conversation started.

Let's be honest, I wasn't expecting this to get merged straightaway.

MikeMcQuaid: 🆒! To be clear, I am excited about more stuff like this moving upstream. I still dream of a day when Linuxbrew/brew isn't needed.

Library/Homebrew/test/test_dependency_collector.rb
rwhogg: Decided that doing so was common sense.

rwhogg:
Slight complication with testing this change: you have to comment out the following line in `ant.rb`:

```ruby
keg_only :provided_by_osx if MacOS.version < :mavericks
```

The reason is that the MacOS constant isn't defined.

It might make sense to put this PR on hold and focus on https://github.com/Linuxbrew/legacy-linuxbrew/commit/4f5efb6377e1245626845d71e46abf1dd8f89858#diff-23335809c452169ea76c54df1816fb54 first. That commit alone fixes a huge number of builds.

Library/Homebrew/dependency_collector.rb
```diff
@@ -5,6 +5,12 @@
 require "requirements"
 require "set"
 
+if OS.mac?
+  require "extend/os/mac/dependency_collector"
+elsif OS.linux?
+  require "extend/os/linux/dependency_collector"
+end
```
MikeMcQuaid: This should live in `extend/dependency_collector.rb` like the other similar files.

Library/Homebrew/extend/os/linux/dependency_collector.rb
```diff
@@ -0,0 +1,4 @@
+def ant_dep(spec, tags)
+  # Always use brewed ant on Linux
```
MikeMcQuaid: Can drop the comment, it's self-explanatory.

Library/Homebrew/extend/os/mac/dependency_collector.rb
```diff
@@ -0,0 +1,5 @@
+def ant_dep(spec, tags)
```
MikeMcQuaid: Check out the other files in this directory to see the pattern for porting these.

Library/Homebrew/test/test_os_mac_dependency_collector.rb
```diff
@@ -19,4 +19,16 @@ def test_ld64_dep_leopard_or_newer
...
+    MacOS.stubs(:version).returns(MacOS::Version.new("10.9"))
+    @d.add :ant => :build
+    assert_equal find_dependency("ant"), Dependency.new("ant", [:build])
+  end
+
+  def test_ant_dep_pre_mavericks
+    MacOS.stubs(:version).returns(MacOS::Version.new("10.7"))
+    @d.add :ant => :build
+    assert_nil find_dependency("ant")
+  end
```
MikeMcQuaid: 👍

Library/Homebrew/dependency_collector.rb
rwhogg: Did you mean `extend/os/dependency_collector.rb`? That seems to be where all the `OS.mac?` checks are.

MikeMcQuaid: Sorry, yes. Brain: fried.

MikeMcQuaid:
Still got some failing tests: https://bot.brew.sh/job/Brew%20Pull%20Requests/1847/version=any_sierra/testReport/junit/brew-test-bot/sierra/tests___no_compat/

Shout if you need a hand with this.

Library/Homebrew/extend/os/linux/dependency_collector.rb
```diff
@@ -0,0 +1,5 @@
+class DependencyCollector
+  def ant_dep(spec, tags)
+    Dependency.new(spec.to_s, tags)
+  end
```
MikeMcQuaid: Sorry to be annoying and change my mind but: rather than making this in `linux` this could probably just be in `Library/Homebrew/dependency_collector.rb` and have the `os/mac` version override it as it currently does.

Library/Homebrew/extend/os/mac/dependency_collector.rb
```diff
@@ -0,0 +1,7 @@
+class DependencyCollector
+  def ant_dep(spec, tags)
+    if MacOS.version >= :mavericks
```
MikeMcQuaid: `return if MacOS.version < :mavericks `

Library/Homebrew/test/test_os_mac_dependency_collector.rb
```diff
@@ -19,4 +23,16 @@ def test_ld64_dep_leopard_or_newer
...
+    MacOS.stubs(:version).returns(MacOS::Version.new("10.9"))
+    @d.add ant: :build
+    assert_equal find_dependency("ant"), Dependency.new("ant", [:build])
+  end
+
+  def test_ant_dep_pre_mavericks
+    MacOS.stubs(:version).returns(MacOS::Version.new("10.7"))
+    @d.add ant: :build
+    assert_nil find_dependency("ant")
+  end
```
MikeMcQuaid: Ideally would be good to add a test for this into the non-mac tests just to verify that path works too.

Library/Homebrew/extend/os/linux/dependency_collector.rb
MikeMcQuaid: Like what we're going for is the code in `os` being OS-specific (including macOS specific) whereas this version is generic enough to be the default.

Library/Homebrew/test/test_os_mac_dependency_collector.rb
rwhogg: I had another test when I first added this to Linuxbrew. Something very similar should do.

Library/Homebrew/extend/os/linux/dependency_collector.rb
rwhogg: No worries!

That last comment is a helpful clarification for the future, to be sure.

Library/Homebrew/test/test_os_mac_dependency_collector.rb
rwhogg: Found it, it's now [here](https://github.com/Homebrew/brew/pull/1014/files#diff-658424c3ddbee14a90e8b8778471edb6R79)

MikeMcQuaid:
This is perfect. A great change and great contribution. Hopefully you understand the flow now @rwhogg it'll be much quicker to get future similar changes through. Thanks again!

PR #1025, opened 2016-09-19
Make `HOMEBREW_PREFIX/Caskroom` the default.

reitermarkus:

-----

Fixes https://github.com/caskroom/homebrew-cask/issues/24610.

### DIFF
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -231,12 +231,8 @@ module Homebrew
       FileUtils.cp_r "#{HOMEBREW_REPOSITORY}/Library/Taps", "#{new_homebrew_repository}/Library/Taps"
     end
 
-    if (HOMEBREW_REPOSITORY/"Caskroom").exist?
-      FileUtils.cp_r "#{HOMEBREW_REPOSITORY}/Caskroom", "#{new_homebrew_repository}/Caskroom"
-    end
-
     unremovable_paths = []
-    extra_remove_paths = [".git", "Library/Locks", "Library/Taps", "Caskroom",
+    extra_remove_paths = [".git", "Library/Locks", "Library/Taps",
                           "Library/Homebrew/cask", "Library/Homebrew/test"]
     (repo_files + extra_remove_paths).each do |file|
       path = Pathname.new "#{HOMEBREW_REPOSITORY}/#{file}"
--- a/Library/Homebrew/cask/lib/hbc.rb
+++ b/Library/Homebrew/cask/lib/hbc.rb
@@ -49,6 +49,7 @@ module Hbc
     Hbc::Cache.ensure_cache_exists
     Hbc::Cache.migrate_legacy_cache
 
+    Hbc::Caskroom.migrate_caskroom_from_repo_to_prefix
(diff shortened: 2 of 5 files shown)


MikeMcQuaid:
@reitermarkus May also be worth tweaking https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/update-report.rb#L234-L239. Happy to merge this once we have one job with 👍 to avoid waiting for coverage completion.

MikeMcQuaid:
Also: will this break any symlinks?

reitermarkus:
I'm not sure about tweaking `update-report.rb`, because if the Caskroom isn't moved to the new repository first, there won't be a reference to it.

MikeMcQuaid:
It feels pretty odd to move it and then move it back to the same location?

Library/Homebrew/cmd/update-report.rb
```diff
@@ -232,11 +232,11 @@ def migrate_legacy_repository_if_necessary
     end
 
     if (HOMEBREW_REPOSITORY/"Caskroom").exist?
-      FileUtils.cp_r "#{HOMEBREW_REPOSITORY}/Caskroom", "#{new_homebrew_repository}/Caskroom"
+      FileUtils.mv "#{HOMEBREW_REPOSITORY}/Caskroom", "#{HOMEBREW_PREFIX}/Caskroom"
```
MikeMcQuaid: This is only run if `HOMEBREW_REPOSITORY` and `HOMEBREW_PREFIX` are both `/usr/local` so it's a no-op and can be removed.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -232,11 +232,16 @@ def migrate_legacy_repository_if_necessary
     end
 
     if (HOMEBREW_REPOSITORY/"Caskroom").exist?
-      FileUtils.cp_r "#{HOMEBREW_REPOSITORY}/Caskroom", "#{new_homebrew_repository}/Caskroom"
+      repo_caskroom   = HOMEBREW_REPOSITORY}/"Caskroom"
+      prefix_caskroom = HOMEBREW_PREFIX}/"Caskroom"
+
+      unless File.identical?(repo_caskroom, prefix_caskroom)
+        FileUtils.mv repo_caskroom, prefix_caskroom
+      end
```
MikeMcQuaid: These will always be the same path.

reitermarkus: Ah, ok, will remove this whole section then.

reitermarkus:
> Also: will this break any symlinks?

It might. But it will probably fix more symlinks than it breaks, since the move to `/usr/local/Homebrew/Caskroom` broke all links pointing to `/usr/local/Caskroom`.

reitermarkus:
`/usr/local/Caskroom` was migrated to `/usr/local/Homebrew/Caskroom` with the migration of the Homebrew repo to `/usr/local/Homebrew`, so all symlinks would be broken if we left the caskroom at `/usr/local/Homebrew/Caskroom`.

jawshooah:
Ah, sorry, read your comment too fast and thought you were talking about the earlier migration from `/opt/homebrew-cask/Caskroom` to `/usr/local/Caskroom`.

PR #1028, opened 2016-09-19
Use user GEM_HOME for installing gems

### LINKED ISSUE #944: `brew audit` creates ~/.gem
I have `GEM_HOME` set to a non-default location to avoid gems being installed in my home directory. brew.sh unsets this variable, so when <kbd>brew audit</kbd> installs rubocop, this directory is created and gems are installed into a directory that otherwise doesn't exist. This means my gems are spread across two different locations, which doesn't seem like the correct behavior.

I understand that brew.sh unsettling this variable must be deliberate and prevent some bad behaviour, so if this is working as intended that's okay, but it doesn't seem like great behaviour to me.

penman:

-----

Save the value of GEM_HOME before unsetting it, and re-set it when a gem needs to be installed (or a previously-installed gem needs to be used). Fixes #944.

Here's what I've been using to test:

```sh
mkdir ~/Desktop/gem_home

GEM_HOME="$HOME/Desktop/gem_home" GEM_PATH="$HOME/Desktop/gem_home" brew tests
# Make sure gems are installed correctly to ~/Desktop/gem_home

GEM_HOME="$HOME/Desktop/gem_home" GEM_PATH="$HOME/Desktop/gem_home" brew tests
# Make sure gems aren't installed again

# Make sure ~/.gem hasn't changed

unset GEM_HOME
unset GEM_PATH

brew tests
# Make sure gems are installed to ~/.gem

brew tests
# Make sure gems aren't installed again
```

I'm also checking all these commands (all the <ins>other</ins> ones that use `Homebrew.install_gem_setup_path!`) with custom GEM_HOMEs and GEM_PATHs set to verify they still work and don't leak into ~/.gem:

- [x] `brew cask-tests`
- [x] `brew cask style`
- [x] `brew style`
- [x] `brew man`

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -59,6 +59,10 @@ then
   odie "Cowardly refusing to continue at this prefix: $HOMEBREW_PREFIX"
 fi
 
+# Save value to use for installing gems
+export GEM_OLD_HOME="$GEM_HOME"
+export GEM_OLD_PATH="$GEM_PATH"
+
 # Users may have these set, pointing the system Ruby
 # at non-system gem paths
 unset GEM_HOME
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -270,17 +270,24 @@ module Homebrew
   end
 
   def self.install_gem_setup_path!(name, version = nil, executable = name)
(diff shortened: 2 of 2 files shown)


MikeMcQuaid:
Fantastic quality work again, @penman.

Issue #905, opened 2016-09-10
[Homebrew Cask] Disable Accessibility Access override for OSX 10.12


adidalal:
`TCC.db` is (rightfully) protected by SIP in Sierra.

We should prevent [`enable_accessibility_access`](https://github.com/Homebrew/brew/blob/9c09f0573020e253e979af44d623ce049a4a2bd6/Library/Homebrew/cask/lib/hbc/installer.rb#L225) ` if MacOS.release >= :sierra` (throw up a caveat instead?)

vitorgalvao:
Closing in favour of the PR at https://github.com/Homebrew/brew/pull/1029.

PR #1029, opened 2016-09-19
Prevent Accessibility access for macOS >= 10.12

wickles:

-----

See https://github.com/caskroom/homebrew-cask/issues/24519 and/or https://github.com/Homebrew/brew/issues/905

Feel free to suggest alternative wording and such.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -253,12 +253,7 @@ class Hbc::Installer
 
   def disable_accessibility_access
     return unless @cask.accessibility_access
-    if MacOS.version <= :mountain_lion
-      opoo <<-EOS.undent
-        Accessibility access was enabled for #{@cask}, but it is not safe to disable
-        automatically on this version of macOS.  See System Preferences.
-      EOS
-    elsif MacOS.version <= :el_capitan
+    if MacOS.version >= :mavericks && MacOS.version <= :el_capitan
       ohai "Disabling accessibility access"
       @command.run!("/usr/bin/sqlite3",
                     args: [
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -225,12 +225,7 @@ class Hbc::Installer
   def enable_accessibility_access
     return unless @cask.accessibility_access
     ohai "Enabling accessibility access"
-    if MacOS.version >= :sierra
-      opoo <<-EOS.undent
-        Accessibility access cannot be enabled automatically on this version of macOS.
-        See System Preferences to enable it manually.
-      EOS
-    elsif MacOS.version <= :mountain_lion
+    if MacOS.version <= :mountain_lion
       @command.run!("/usr/bin/touch",
                     args: [Hbc.pre_mavericks_accessibility_dotfile],
                     sudo: true)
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -225,7 +225,12 @@ def print_caveats
   def enable_accessibility_access
     return unless @cask.accessibility_access
     ohai "Enabling accessibility access"
-    if MacOS.version <= :mountain_lion
+    if MacOS.version >= :sierra
+      opoo <<-EOS.undent
+        Access denied to accessibility DB (TCC.db) by SIP on this version of macOS.
+        If needed, enable manually via the app or System Preferences.
+      EOS
+    elsif MacOS.version <= :mountain_lion
```
reitermarkus: I'd change the wording to:

Accessibility access cannot be enabled on this version of macOS.
Enable it manually via System Preferences.

Same with “disable”.

wickles: How about
```
Accessibility access cannot be enabled automatically on this version of macOS.
See System Preferences to enable it manually.
```
? Is there a line length it should not exceed (i.e. should I wrap 'macOS' to the next line)?

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -225,7 +225,12 @@ def print_caveats
   def enable_accessibility_access
     return unless @cask.accessibility_access
     ohai "Enabling accessibility access"
-    if MacOS.version <= :mountain_lion
+    if MacOS.version >= :sierra
```
vitorgalvao: We generally favour using `<=` and `else` while “climbing”. So always say `if <= oldest … elsif <= old … elsif <= less_old … else`. So we climb up versioning and leave `else` as something that means “from this version on”.

reitermarkus: Yes, the order should be:

```ruby
if MacOS.version <= :mountain_lion
elsif MacOS.version <= :yosemite
elsif MacOS.version <= :el_capitan
else
end


```

wickles: Done. Now I'm wondering if the two `disable` messages should just be combined? 

<= Mountain Lion: 
```
        Accessibility access was enabled for #{@cask}, but it is not safe to disable
        automatically on this version of macOS.  See System Preferences.
```
Sierra: 
```
        Accessibility access cannot be disabled automatically on this version of macOS.
        See System Preferences to disable it manually.
```

reitermarkus: Not sure why “it is not safe” on `<= Mountain Lion`, but I think as it is worded now, it can be used for both.

reitermarkus: So this would make it:

```ruby
if MacOS.version >= :mavericks &&  MacOS.version <= :el_capitan
else
end
```

wickles: Done. 👍

wickles:
Checks are failing for [accessibility_test.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/test/cask/accessibility_test.rb). Errors reported for 10.10, 10.11, 10.12: 
```
Failure:
Accessibility Access::uninstall#test_0002_warns about disabling accessibility access on old OS X releases [/Users/brew/Jenkins/pr-brew/version/yosemite/Library/Homebrew/cask/test/cask/accessibility_test.rb:57]
Minitest::Assertion: Expected /Warning:\ Accessibility\ access\ was\ enabled\ for\ with\-accessibility\-access,\ but\ it\ is\ not\ safe\ to\ disable/ to match "Warning: Accessibility access cannot be disabled automatically on this version of macOS.\nSee System Preferences to disable it manually.\n".
```

Does anything need to be changed besides updating the string in 
```
capture_io { @installer.disable_accessibility_access }[1]
  .must_match("Warning: Accessibility access was enabled for with-accessibility-access, but it is not safe to disable")
```
???

reitermarkus:
Yes, you should update the string in the current test. And you should also add another test which tests if the same message appears on `>= :sierra`.

Library/Homebrew/cask/test/cask/accessibility_test.rb
```diff
@@ -48,13 +55,21 @@
...
       end
     end
-    it "warns about disabling accessibility access on old OS X releases" do
+    it "warns about disabling accessibility access on old macOS releases" do
       MacOS.stubs(version: MacOS::Version.new("10.8"))
 
       @installer.stubs(bundle_identifier: "com.example.BasicCask")
 
       capture_io { @installer.disable_accessibility_access }[1]
         .must_match("Warning: Accessibility access was enabled for with-accessibility-access, but it is not safe to disable")
```
jawshooah: This string needs to be updated as well.

wickles: Shoot, my bad. Thanks!

PR #1042, opened 2016-09-20
Updated documentation links to relative paths

pkuczynski:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,4 @@
-- [ ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
+- [ ] Have you followed the guidelines in our [Contributing](CONTRIBUTING.md) document?
```
zmwangx: This won't work when you're on the PR screen.

MikeMcQuaid:
I think it's best for these to use direct paths that always point to `master`. If any are broken, though, please update them. Thanks!

.github/PULL_REQUEST_TEMPLATE.md
pkuczynski: Fixed. Have a look at the updated PR...

pkuczynski:
Updated PR to use full path for pull request template

MikeMcQuaid:
> Updated PR to use full path for pull request template

Would be good to use them everywhere, thanks.

pkuczynski:
Why would they need to point to master? What if someone works one documentation in branch? Or documentation changes for a new version?

MikeMcQuaid:
@pkuczynski These docs are focused at users and we're rather they can always get the newest version even if clicking on outdated commit versions.

pkuczynski:
I see. Then how would you explain your latest changes to README.md done in https://github.com/Homebrew/brew/commit/7a63a753609558e14f20080dd7ab5f4fc6e3f985#diff-04c6e90faac2675aa89e2176d2eec7d8

You are using relative path over there... :)

MikeMcQuaid:
Inconsistency, basically 😀. Will leave this open and let more maintainers chime in with thoughts. Thanks!

MikeMcQuaid:
Passing on this but thanks for the PR anyway.

pkuczynski:
Seems that nobody cared except me... Shame...

PR #1043, opened 2016-09-20
Alter mkdir wrapper use mkdir_p to create intermediate directories

msavy:

-----

Proposal - https://github.com/Homebrew/brew-evolution/pull/13
Relevant PR discussion https://github.com/Homebrew/homebrew-core/pull/4976#issuecomment-248258491 - https://github.com/Homebrew/brew/pull/1043


msavy:
@MikeMcQuaid - Made a PR plus document as per the conversation we had earlier [here](https://github.com/Homebrew/homebrew-core/pull/4976#issuecomment-248258491).

Please note that I get a failure when running `brew tests` on latest master *without* my change applied (i.e. issue completely unrelated to this change). 

No (additional) tests fail due to this change.

msavy:
Failure I see [FYI](https://gist.github.com/msavy/bb3ca05c7faa8ac7016beffb2b710c63).

Looks like a path issue or perhaps something missing from Gemfile. Either way, seems OK on CI :-).

DomT4:
Why wouldn't you just use `mkpath`, out of curiosity?

MikeMcQuaid:
@DomT4 is there a `mkpath do`?

DomT4:
@MikeMcQuaid No, like `mkdir_p` it doesn't natively take a block, but unlike `mkdir` we haven't extended on top of that.

We've been loosely using `mkpath` instead of `mkdir_p` for a while, so my temptation would be to extend `mkpath` rather than `mkdir` behaving _too_ differently than it does natively, but it's not a blocking objection.

We tend to use both `mkpath` and `mkdir_p` like this:
```ruby
path = buildpath/"test/test2"
path.mkpath
cd path do
  system "blah"
  system "make", "blah"
end
```

MikeMcQuaid:
I'd be in favour of perhaps only having the `mkdir_p` behaviour when in the block form.

msavy:
@MikeMcQuaid I think having such inconsistency between block/non-block form would be a violation of the principle of least surprise. 

I would strongly suggest it being the same, in whichever form we land upon.

DomT4:
To be fair, almost all of Homebrew's magic violates the principle of least surprise, hence why we see a bunch of PRs with people using things like `cd path` followed by `make` followed by `cd buildpath` etc even though our `cd` can take a block to handle a lot of that automatically. 

That principle is why I wondered if it was less invasive to modify `mkpath` to take a block here rather than the vanilla `mkdir` to create multiple layers. That argument can swing either way though.

msavy:
@DomT4 I understand what you're saying, but don't think that it is a useful parallel to this particular situation. 

Having substantially different behaviour on a single method call when with/without a block is confusing. 

I think that's distinct in meaningful way to having consistently behaving (but enhanced) functions.

MikeMcQuaid:
@DomT4 Cool, agreed, feel free to make `mkdir` do `mkdir_p` (and it could call it instead of `old_mkdir` as well).

DomT4:
> Having substantially different behaviour on a single method call when with/without a block is confusing.

Sure, that's a fair point 👍. Interested in your comments on the gap between surprising behaviour on tweaking it like this and tweaking `mkpath` to take a block instead though, but unless any other maintainers care I'm happy enough to leave it.

msavy:
@DomT4 @MikeMcQuaid I think either solution is fine, as long as it behaves consistently.

The `mkdir` one appeals to me because it's more UNIXy than `mkpath`, and appears to have no side-effects. 

But, that's a minor distinction and I would be equally happy with adding block support to `mkdir_p` and/or `mkpath`.

Happy to contribute it, either way :-).

DomT4:
Nobody else has rushed to agree with my thoughts, so I'm happy here whenever @MikeMcQuaid is.

MikeMcQuaid:

PR #1050, opened 2016-09-20
RuboCop changes.

reitermarkus:

-----

This should get us one step closer to being able to merge Homebrew's and Cask's RuboCop  config.


MikeMcQuaid:
Looking good but let's hold off merging until after 1.0.0 (so probably this time tomorrow).

Library/Homebrew/extend/ARGV.rb
```diff
@@ -216,7 +216,8 @@ def build_from_source?
   end
 
   def build_all_from_source?
-    !!ENV["HOMEBREW_BUILD_FROM_SOURCE"]
+    return true if ENV["HOMEBREW_BUILD_FROM_SOURCE"]
```
jawshooah: Why not just replace `!!value` with `!value.nil?` here and elsewhere?

reitermarkus: To make sure, `!value.nil?` will only work for `ENV` variables, since those can't be `false`, or can they?

MikeMcQuaid:
```
mikebook # brew irb                (update-between-tags) [~/Documents/Homebrew]
>> ENV["HOMEBREW_DEVELOPER"] = false
TypeError: no implicit conversion of false into String
	from (irb):1:in `[]='
	from (irb):1
	from /Users/mike/Documents/Homebrew/Library/Homebrew/brew.rb:99:in `<main>'
>> 
```

so 👍 to `.nil?` from me.

MikeMcQuaid:
Happy to 🚢 this when conflicts are resolved. @reitermarkus after that it'd be cool to unify Cask's rubocop and style too.

MikeMcQuaid:
Nice. Would love to see us get a unified style on cask and the rest of Homebrew/brew as a next step.

reitermarkus:
Yep, will create another PR *before* that though, to clean up most of the remaining offenses in `rubocop_todo.yml`.

MikeMcQuaid:
🆒 👍

ilovezfs:
This changed the API on search_taps, which now dies when passed a string.

Issue #931, opened 2016-09-11
Store the formula used to build a bottle inside the bottle


MikeMcQuaid:
When building a bottle, store the formula used to build the keg inside the bottle in a file named `NAME/VERSION/.bottle/NAME.rb` after removing the `bottle do ... end` block.

Extracted from https://github.com/Homebrew/brew-evolution/blob/master/003-install-bottles-of-arbitrary-versions.md merged in https://github.com/Homebrew/brew-evolution/pull/6.

CC @sjackman

sjackman:
I have an implementation of this feature that needs just a bit more work. I'll submit a PR this coming week (hopefully early in the week).

MikeMcQuaid:
Cool, thanks @sjackman.

sjackman:
I've created a PR at https://github.com/Homebrew/brew/pull/1052

PR #1052, opened 2016-09-20
Store the formula used to build the keg in the keg

sjackman:

-----

Store the formula used to build the keg inside the keg in a
file named NAME/VERSION/.bottle/NAME.rb after removing the
bottle do ... end block.

See https://github.com/Homebrew/brew-evolution/pull/6.
Closes https://github.com/Homebrew/brew/issues/931.

### DIFF
--- a/Library/Homebrew/cmd/list.rb
+++ b/Library/Homebrew/cmd/list.rb
@@ -140,6 +140,8 @@ class PrettyListing
           # dylibs have multiple symlinks and we don't care about them
           (pnn.extname == ".dylib" || pnn.extname == ".pc") && !pnn.symlink?
         end
+      when ".brew"
+        # Ignore .brew
       else
         if pn.directory?
           if pn.symlink?
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -260,6 +260,12 @@ class FormulaInstaller
       compute_and_install_dependencies if not_pouring && !ignore_deps?
       build
       clean
+
+      # Store the formula used to build the keg in the keg.
+      s = formula.path.read.gsub(/  bottle do.+?end\n\n?/m, "")
+      brew_prefix = formula.prefix/".brew"
+      brew_prefix.mkdir
+      Pathname(brew_prefix/"#{formula.name}.rb").atomic_write(s)
     end
 
     build_bottle_postinstall if build_bottle?
--- a/Library/Homebrew/test/test_formula_installer.rb
+++ b/Library/Homebrew/test/test_formula_installer.rb
@@ -58,6 +58,7 @@ class InstallTests < Homebrew::TestCase
       bin = HOMEBREW_PREFIX+"bin"
       assert_predicate bin, :directory?
       assert_equal 3, bin.children.length
+      assert_predicate f.prefix/".brew/testball.rb", :readable?
     end
   end
 


Library/Homebrew/cmd/list.rb
```diff
@@ -140,6 +140,8 @@ def initialize(path)
           # dylibs have multiple symlinks and we don't care about them
           (pnn.extname == ".dylib" || pnn.extname == ".pc") && !pnn.symlink?
         end
+      when ".bottle"
+        # Ignore .bottle
```
MikeMcQuaid: Just to save me a bit: what command do you use to see this? Thanks!

Library/Homebrew/formula_installer.rb
```diff
@@ -786,7 +791,7 @@ def pour
       path.extend(InstallRenamed)
       path.cp_path_sub(formula.bottle_prefix, HOMEBREW_PREFIX)
     end
-    FileUtils.rm_rf formula.bottle_prefix
+    FileUtils.rm_rf Dir["#{formula.bottle_prefix}/{etc,var}"]
```
MikeMcQuaid: This makes me wonder if it's worth picking another path for storing this.

Library/Homebrew/formula_installer.rb
```diff
@@ -259,6 +259,11 @@ def install
       end
       build
       clean
+
+      # Store the formula used to build the keg in the keg.
+      s = File.open(formula.path, "rb", &:read).gsub(/  bottle do.+?end\n\n?/m, "")
```
MikeMcQuaid: `formula.path.read` may be sufficient?

Library/Homebrew/cmd/list.rb
sjackman: Do you mean `brew list hello`? Sorry if I misunderstand the question. The goal of this line is to elide `.bottle/hello.rb` from the output of `brew list hello`.

Library/Homebrew/formula_installer.rb
sjackman: With this change, line 794 now parallels line 790. The two directories `etc` and `var` are installed and then removed from `.bottle`.
https://github.com/sjackman/brew/blob/dd27921d6a096fc7f6bbcf8f04786cb669283d95/Library/Homebrew/formula_installer.rb#L790-L794
```ruby
    Pathname.glob("#{formula.bottle_prefix}/{etc,var}/**/*") do |path|
…
    end
    FileUtils.rm_rf Dir["#{formula.bottle_prefix}/{etc,var}"]
```

My other suggestion would be `FORMULA/VERSION/.homebrew/FORMULA.rb`

Library/Homebrew/formula_installer.rb
sjackman: Fixed. Thanks for the tip!

Library/Homebrew/cmd/list.rb
MikeMcQuaid: Perfect thanks 👍

Library/Homebrew/formula_installer.rb
MikeMcQuaid: How about just `FORMULA/VERSION/.formula/FORMULA.rb`? If you're 👍 on that and this PR is updated accordingly think this is good to 🚢. Thanks for your patience here 🙇

sjackman: We have two metafiles so far, `INSTALL_RECEIPT.json` and `FORMULA.rb`. On the chance that one day there's another metafile to install, I'd be inclined to name this directory something generic, like `.brew`, so that all the metafiles can live in one suitably-named directory.

I've updated the PR using `.brew`. I'm happy to change it to whichever name you prefer.

sjackman:
```
 > git checkout -f 7ee02c791683f69dd1fced797b311df662aab927
FATAL: Could not checkout 7ee02c791683f69dd1fced797b311df662aab927
```
https://bot.brew.sh/job/Brew%20Pull%20Requests/2025/version=el_capitan/console

sjackman:
@BrewTestBot test this please

MikeMcQuaid:
> Is this a transient error?

Yes.

sjackman:
Thanks for the positive review, Mike. Merge?

MikeMcQuaid:
@Homebrew/maintainers Anyone else have thoughts on the `.brew` naming here? I'm not 💯 on it but I'm unsure of what I'd rather instead. If no-one else cares or if people think we should 🚢 this as-is I probably just will.

DomT4:
`.metafiles` would work as well, or `.internal`. No objections to `.brew` though really.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #1058, opened 2016-09-21
Implement mechanism to suppress unsupported macOS warnings

### LINKED ISSUE #1056: Warning: You are using macOS 10.9
```bash
~ $ brew upgrade
Warning: You are using macOS 10.9.
We (and Apple) do not provide support for this old version.
You may encounter build failures or other breakages.
Please create pull-requests instead of filing issues.
```

I understand that Homebrew has formally ended support for 10.9 and that I should be aware of this when installing formulae. I want to take my chances, is it possible to suppress this message? I could not immediately find anything in `man brew`.

zmwangx:

-----

An idea to entertain, because getting a warning on each brew invocation is genuinely annoying (but that's kind of the point). Fixes #1056.

Design (to quote comments in `brew.sh`):

> Handle suppression of unsupported macOS (prerelease or outdated release)
warnings. Remember the `HOMEBREW_NO_WARN_UNSUPPORTED_MACOS` environment
variable through recording the target macOS major version (e.g. 10.9) as
`homebrew.nowarnmacos` and the "warning epoch" as `homebrew.nowarnmacosepoch`
in the Homebrew repository's git config.

> "Warning epoch" is a positive integer that should be bumped every time we
add basic support for a new prerelease version of macOS or deprecate an old
version. This ensures that a user who has suppressed warnings previously
will start getting warnings again (if applicable) and will have to review
their decision when one of the above happens.

Basically, if you run brew with the `HOMEBREW_NO_WARN_UNSUPPORTED_MACOS` env var once,  your warnings on this particular major macOS version and this particular brew installation will be suppressed until we adjust the criterion.

I'm not sure where we should put this in our docs yet (need to look at the docs) and whether we should document it in the man page — this should be something that is only revealed if you really look for it. I won't work on the docs until the team think this is worth inclusion.

### DIFF
--- a/Library/Homebrew/extend/os/mac/diagnostic.rb
+++ b/Library/Homebrew/extend/os/mac/diagnostic.rb
@@ -26,7 +26,8 @@ module Homebrew
       end
 
       def check_for_unsupported_macos
-        return if ARGV.homebrew_developer?
+        return if ARGV.homebrew_developer? ||
+                  ARGV.suppress_unsupported_macos_warnings?
 
         who = "We"
         if OS::Mac.prerelease?
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -59,6 +59,8 @@ then
   odie "Cowardly refusing to continue at this prefix: $HOMEBREW_PREFIX"
 fi
 
+HOMEBREW_GIT_CONFIG_FILE="$HOMEBREW_REPOSITORY/.git/config"
+
 # Save value to use for installing gems
 export GEM_OLD_HOME="$GEM_HOME"
 export GEM_OLD_PATH="$GEM_PATH"
@@ -120,6 +122,7 @@ export HOMEBREW_LIBRARY
 export HOMEBREW_VERSION
 export HOMEBREW_CACHE
 export HOMEBREW_CELLAR
+export HOMEBREW_GIT_CONFIG_FILE
 export HOMEBREW_SYSTEM
 export HOMEBREW_CURL
 export HOMEBREW_PROCESSOR
@@ -197,7 +200,6 @@ esac
 
 if [[ -z "$HOMEBREW_DEVELOPER" ]]
 then
-  export HOMEBREW_GIT_CONFIG_FILE="$HOMEBREW_REPOSITORY/.git/config"
   HOMEBREW_GIT_CONFIG_DEVELOPERMODE="$(git config --file="$HOMEBREW_GIT_CONFIG_FILE" --get homebrew.devcmdrun)"
   if [[ "$HOMEBREW_GIT_CONFIG_DEVELOPERMODE" = "true" ]]
   then
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -162,6 +162,42 @@ You have not agreed to the Xcode license. Please resolve this by running:
 EOS
     fi
   fi
+
+  # Handle suppression of unsupported macOS (prerelease or outdated release)
+  # warnings. Remember the HOMEBREW_NO_WARN_UNSUPPORTED_MACOS environment
+  # variable through recording the target macOS major version (e.g. 10.9) as
+  # homebrew.nowarnmacos and the "warning epoch" as homebrew.nowarnmacosepoch
+  # in the Homebrew repository's git config.
+  #
+  # "Warning epoch" is a positive integer that should be bumped every time we
+  # add basic support for a new prerelease version of macOS or deprecate an old
+  # version. This ensures that a user who has suppressed warnings previously
+  # will start getting warnings again (if applicable) and will have to review
+  # their decision when one of the above happens.
+  [[ "$HOMEBREW_MACOS_VERSION" =~ ^[0-9]+\.[0-9]+ ]]
+  HOMEBREW_MACOS_MAJOR_VERSION="${BASH_REMATCH[0]}"
+  HOMEBREW_MACOS_WARNING_EPOCH="1"
+  if [[ -n "$HOMEBREW_NO_WARN_UNSUPPORTED_MACOS" ]]
+  then
+    git config --file="$HOMEBREW_GIT_CONFIG_FILE" --replace-all homebrew.nowarnmacos "$HOMEBREW_MACOS_MAJOR_VERSION"
+    git config --file="$HOMEBREW_GIT_CONFIG_FILE" --replace-all homebrew.nowarnmacosepoch "$HOMEBREW_MACOS_WARNING_EPOCH"
+  else
+    # HOMEBREW_NO_WARN_UNSUPPORTED_MACOS environment variable not set. Try to
+    # retrieve the macOS version for which unsupported warnings has been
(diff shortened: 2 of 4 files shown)


chdiza:
I'm not sure why there has to be more fiddling with people's gitconfig files.  Why can't this just be a shellvar that suppresses warnings if set and doesn't suppress warnings if unset?

For the record, I am in favor of such a shellvar.  I just don't see why brew needs to look for that shellvar and then write to or read from to the repo's gitconfig.

chdiza:
If multiple suppressed "versions" need to be tracked, then the shellvar can be a colon or comma separated list.

zmwangx:
Having to set an env bar all the time does actual damage, especially when it's as long as this one. That's all I'm going to comment on this topic until someone with every power asks, by which point I'll explain the technical details.

chdiza:
> Having to set an env var all the time does actual damage

Why would one set it "all the time"?  Wouldn't a user just put it in their `~/.profile`?

zmwangx:
At all times.

chdiza:
> At all times.

But then why stick it in the gitconfig, where it will be set "at all times"?  As far as I can see, it will be set even if the user *stops* setting the shallvar.

Related: how will the user who has asked for the warnings to be suppressed change her mind?  Unsetting the shellvar will not do it (because of lines 185-200 of brew.sh in the PR).  Manually dive into the gitconfig?

jawshooah:
`git config unset {{whatever}}`?

jawshooah:
And I assume @zmwangx's problem with Yet Another Environment Variable is that it pollutes the global environment for the entire shell session, whether you're using Homebrew at the moment or not. Too many environment variables could also presumably degrade performance, though I have done no research to find out if there is an appreciable effect.

chdiza:
> git config --unset {{whatever}}?

Not very user-friendly.

chdiza:
If performance is the concern, I'd be very, very shocked to find that one more shellvar has more of a negative effect than yet another git-config call.

zmwangx:
> But then why stick it in the gitconfig, where it will be set "at all times"?

Okay, I really don't want to explain this, but let me explain. There's nothing wrong with enabling this behavior in brew at all times. The problem is having long and almost useless environment variables at all times in your environment. Every environment variable is part of `envv[]` that counts toward `ARG_MAX`, which is 262144 on Darwin. Having a single 34 character long environment variable may not be too big a deal, but when every piece of software asks you to set a bunch of junk, they do add up to something considerable. Currently my `env | wc -c` is around 5000 and I don't even have many variables set. I would try hard to avoid more bloat and I urge software developers to be considerate. Keep in mind, this is not unlimited resource (unlike another CPU cycle).

> If performance is the concern, I'd be very, very shocked to find that one more shellvar has more of a negative effect than yet another git-config call.

If you care about performance no that front you probably want to optimize Ruby away.

> how will the user who has asked for the warnings to be suppressed change her mind? 

I doubt it. If someone is seeking to re-enable a warning, I'd say the warning has served its purpose already. Plus it's very easy once you understand the mechanism.

MikeMcQuaid:
I'm with @chdiza: we should just make this another environment variable. The only reason we use the `gitconfig` stuff is for things that we want to set in the background for people without them having to specify an environment variable and I don't think this is that.

Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -26,7 +26,8 @@ def fatal_development_tools_checks
       end
 
       def check_for_unsupported_macos
-        return if ARGV.homebrew_developer?
+        return if ARGV.homebrew_developer? ||
+                  ARGV.suppress_unsupported_macos_warnings?
```
MikeMcQuaid: This will stop `brew doctor` showing it which is definitely not what we want.

MikeMcQuaid:
I'd suggest we wait for more than two people to complain about this before changing it. I think the ideal would be to only change the message we print before every e.g. `brew install` command if we set the environment variable and also in that case tell people to not report issues but instead file PRs. It may just be worth removing that warning altogether, though.

mistydemeo:
> I'd suggest we wait for more than two people to complain about this before changing it.

Anecdotally I've had several people complain to me personally about this, it just hasn't made it to the tracker.

MikeMcQuaid:
> Anecdotally I've had several people complain to me personally about this, it just hasn't made it to the tracker.

Good enough for me 👍.

Perhaps we just drop the warning from non-doctor and append something to the troubleshooting links? Honestly just having it in doctor is probably enough if we're prepared to instaclose 10.9 issues.

zmwangx:
The environment variable should definitely be set once and persist until we have a new criterion (hence my "migration epoch" thing). If you require it to be always set in order to suppress the warning, people will just drop it in their dotfiles, and never receive warnings again, ever.

MikeMcQuaid:
> If you require it to be always set in order to suppress the warning, people will just drop it in their dotfiles, and never receive warnings again, ever.

That seems fine, honestly?

zmwangx:
> That seems fine, honestly?

Um, not for me it's not. @DomT4 disagrees too according to his comment on the other thread.

MikeMcQuaid:
I'm 👎 on adding a git config line here. If we have an environment variable per-OS version that seems fine. Honestly, though, rather than making it optional I really think we should just adjust our defaults.

zmwangx:
> I'm 👎 on adding a git config line here.

Fair enough, I don't feel strongly about the whole idea. 

> If we have an environment variable per-OS version

We don't need more env var bloat, thanks.

> rather than making it optional I really think we should just adjust our defaults.

What do you mean?

MikeMcQuaid:
> rather than making it optional I really think we should just adjust our defaults.

Every time we consider adding an option or environment variable it's worth asking: should we just change the default behaviour? Options are often a crutch for not doing design.

> We don't need more env var bloat, thanks.

I'm far more concerned about code bloat than environment variable bloat (which has never been an actual problem for us).

zmwangx:
> Every time we consider adding an option or environment variable it's worth asking: should we just change the default behaviour? Options are often a crutch for not doing design.

You mean we should consider not issuing such warnings?

MikeMcQuaid:
We should consider that, yes, and consider whether it'll cause more harm than good, whether there's enough users who would benefit from it (or the option). 

My personal opinion is that we can tweak existing messaging rather than adding an option to opt-in to this behaviour.

zmwangx:
Even if you just print one line each time, it will be annoying to those who are annoyed by it. It's not about printing what, it's about printing *at all*.

chdiza:
The shellvar can be a colon-separated list of major versions for which to suppress.  That way, even if someone leaves (say) a setting to suppress warnings for Mavericks in their `~/.profile` and totally forgets about it, eventually there will be warnings again if they are later on (say) Yosemite and it goes out of date.

> We don't need more env var bloat, thanks.

If that is even a thing, it should not be replaced with bloat in the number of git-config calls.  (That would be OK, though, if we only read the whole gitconfig once and cached it for later lookup of particular settings inside it.  I'm not really sure why we aren't doing that, but that's a side point.)

zmwangx:
Closing due to lack of interest.

PR #1059, opened 2016-09-21
uninstall: improve pronoun for multiple version message

josephfrazier:

-----

When exactly two versions of a package were installed, the uninstall message should read "Remove it with..." rather than "Remove them all with...", since only one version remains.

See related commit: 62addcfce8d942eaa96f48f80b8b82ad7d934853

### DIFF
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -25,7 +25,7 @@ module Homebrew
             versions = rack.subdirs.map(&:basename)
             verb = versions.length == 1 ? "is" : "are"
             puts "#{keg.name} #{versions.join(", ")} #{verb} still installed."
-            puts "Remove them all with `brew uninstall --force #{keg.name}`."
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I guess it depends on intent; to me the previous message read as a description of what the command does in general rather than the next time you run it necessarily. Will let others chime in though. Your approach is nice.

josephfrazier:
Oh yeah, I now see how it could be interpreted either way, and agree on waiting for other opinions. Thanks for the quick feedback!

josephfrazier:
Hmm, maybe using "Remove all versions with..." regardless of `versions.length` would be better. It's flexible enough to avoid being interpreted as grammatically incorrect, and it still accurately describes the general behavior of `brew uninstall --force`.

MikeMcQuaid:
> Hmm, maybe using "Remove all versions with..." regardless of versions.length would be better. It's flexible enough to avoid being interpreted as grammatically incorrect, and it still accurately describes the general behavior of brew uninstall --force.

@josephfrazier Sounds great!

MikeMcQuaid:

PR #1064, opened 2016-09-21
Ask for `sudo` when moving Caskroom.

reitermarkus:

-----

Addresses https://github.com/caskroom/homebrew-cask/issues/24674, since people need to run `sudo chown` anyway, so might as well prompt them to.

--

@Homebrew/cask

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/caskroom.rb
+++ b/Library/Homebrew/cask/lib/hbc/caskroom.rb
@@ -5,6 +5,12 @@ module Hbc::Caskroom
     repo_caskroom = Hbc.homebrew_repository.join("Caskroom")
     if !Hbc.caskroom.exist? && repo_caskroom.directory?
       ohai "Moving Caskroom from HOMEBREW_REPOSITORY to HOMEBREW_PREFIX"
+
+      unless Hbc.caskroom.parent.writable?
+        opoo "#{Hbc.caskroom.parent} is not writable, changing owner to current user."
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Could you tell people to change it back afterwards? Also, how would you feel about printing a message with either the command to run and letting them do it themselves or, alternatively, the command you're about to run before you run it?

Library/Homebrew/cask/lib/hbc/caskroom.rb
```diff
@@ -5,6 +5,12 @@ def migrate_caskroom_from_repo_to_prefix
     repo_caskroom = Hbc.homebrew_repository.join("Caskroom")
     if !Hbc.caskroom.exist? && repo_caskroom.directory?
       ohai "Moving Caskroom from HOMEBREW_REPOSITORY to HOMEBREW_PREFIX"
+
+      unless Hbc.caskroom.parent.writable?
+        opoo "#{Hbc.caskroom.parent} is not writable, changing owner to current user."
+        system "/usr/bin/sudo", "--", "/usr/sbin/chown", "--", "#{Hbc::Utils.current_user}:staff", Hbc.caskroom.parent.to_s
```
vitorgalvao: Brought me back to when we used to have to do something similar when using `/opt`.

reitermarkus:
We could alternatively use `sudo mv` instead of `chmod`ding `/usr/local`, this way we don't need to change the permissions back.

PR #1073, opened 2016-09-22
readme: fix maintainer link

penman:

-----

The previous URL for @vladshablinsky 404ed for me.

Out of curiosity, I checked a similar URL for an organisation I'm a member of, and it also 404ed. 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This is our PGP key which is valid until May 24, 2017.
 ## Who Are You?
 Homebrew's lead maintainer is [Mike McQuaid](https://github.com/mikemcquaid).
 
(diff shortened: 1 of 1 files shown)


zmwangx:
When I log in with a sock puppet (shh!) it does 404 for me. But when I'm logged into this account it doesn't. I think only an org owner can access that page.

By the way, the ordering looks totally inscrutable to me...

penman:
Ah, yes, when I check with an organisation I'm an _owner_ of, rather than just a member, I can access it.

MikeMcQuaid:
Thanks @penman!

PR #1078, opened 2016-09-22
formula: help CMake with 10.12 SDK on 10.11

ilovezfs:
This adds a CMake cache entry to std_cmake_args specifying that the
function clock_gettime is not available on 10.11 in order to avoid
runtime errors such as

  dyld: lazy symbol binding failed: Symbol not found: _clock_gettime

when the build system is confused by Xcode 8's weak symbols.

Other weak symbols on 10.11, which may merit the same treatment in the
future, can be found with

  grep 'weak$os10.11' MacOSX.sdk/usr/lib/system/libsystem_c.tbd


zmwangx:
Found the override? Nice.

MikeMcQuaid:
@ilovezfs 👍 merge when 💚

PR #1032, opened 2016-09-19
Save runtime dependencies in INSTALL_RECEIPT.json

penman:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Fixes #930

PR #1082, opened 2016-09-22
uninstall: refuse when dependents still installed

### LINKED ISSUE #934: Prevent uninstalling in-use dependencies
If you try to `uninstall` a formula that's currently a runtime `depends_on` (i.e. not `:build`) for another installed formula it should refuse to `uninstall`. This should still be possible but require the use of `uninstall --force`.

alyssais:

---

Makes `brew uninstall foo` (but not `brew uninstall --force foo`) refuse to uninstall `foo` if kegs that depend on it are still installed.

Closes #934.

### DIFF
--- a/Library/Homebrew/cmd/missing.rb
+++ b/Library/Homebrew/cmd/missing.rb
@@ -18,8 +18,13 @@ module Homebrew
       ARGV.resolved_formulae
     end
 
-    Diagnostic.missing_deps(ff, ARGV.value("hide")) do |name, missing|
-      print "#{name}: " if ff.size > 1
+    hide = (ARGV.value("hide") || "").split(",")
+
+    ff.each do |f|
+      missing = f.missing_dependencies(hide: hide)
+      next if missing.empty?
+
+      print "#{f}: " if ff.size > 1
       puts missing.join(" ")
     end
   end
--- a/Library/Homebrew/cmd/missing.rb
+++ b/Library/Homebrew/cmd/missing.rb
@@ -3,7 +3,7 @@
 #:    given, check all installed brews.
 #:
 #:    If `--hide=`<hidden> is passed, act as if none of <hidden> are installed.
-#:    <hidden> should be a comma-seperated list of formulae.
+#:    <hidden> should be a comma-separated list of formulae.
 
 require "formula"
 require "tab"
--- a/Library/Homebrew/cmd/missing.rb
+++ b/Library/Homebrew/cmd/missing.rb
@@ -18,7 +18,7 @@ module Homebrew
       ARGV.resolved_formulae
     end
 
-    Diagnostic.missing_deps(ff) do |name, missing|
+    Diagnostic.missing_deps(ff, ARGV.value("hide")) do |name, missing|
       print "#{name}: " if ff.size > 1
       puts missing.join(" ")
     end
--- a/Library/Homebrew/cmd/missing.rb
+++ b/Library/Homebrew/cmd/missing.rb
@@ -1,6 +1,9 @@
-#:  * `missing` [<formulae>]:
+#:  * `missing` [`--hide=`<hidden>] [<formulae>]:
 #:    Check the given <formulae> for missing dependencies. If no <formulae> are
 #:    given, check all installed brews.
+#:
+#:    If `--hide=`<hidden> is passed, act as if none of <hidden> are installed.
+#:    <hidden> should be a comma-seperated list of formulae.
 
 require "formula"
 require "tab"
--- a/Library/Homebrew/cmd/missing.rb
+++ b/Library/Homebrew/cmd/missing.rb
@@ -18,10 +18,8 @@ module Homebrew
       ARGV.resolved_formulae
     end
 
-    hide = (ARGV.value("hide") || "").split(",")
-
     ff.each do |f|
-      missing = f.missing_dependencies(hide: hide)
+      missing = f.missing_dependencies(hide: ARGV.values("hide"))
       next if missing.empty?
 
       print "#{f}: " if ff.size > 1
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -15,18 +15,18 @@ module Homebrew
     raise KegUnspecifiedError if ARGV.named.empty?
 
     kegs_by_rack = if ARGV.force?
-      Hash[ARGV.named.map do |name|
+      Hash[ARGV.named.map { |name|
         rack = Formulary.to_rack(name)
         [rack, rack.subdirs.map { |d| Keg.new(d) }]
-      end]
+      }]
     else
       ARGV.kegs.group_by(&:rack)
     end
 
     # --ignore-dependencies, to be consistent with install
     unless ARGV.include?("--ignore-dependencies") || ARGV.homebrew_developer?
-      kegs = kegs_by_rack.values.flatten(1)
-      return if check_for_dependents kegs
+      all_kegs = kegs_by_rack.values.flatten(1)
+      return if check_for_dependents all_kegs
     end
 
     kegs_by_rack.each do |rack, kegs|
@@ -77,13 +77,13 @@ module Homebrew
   def check_for_dependents(kegs)
     kegs.each do |keg|
       dependents = keg.installed_dependents - kegs
-      if dependents.any?
-        dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
-        conjugation = dependents.count == 1 ? "is" : "are"
-        ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
-        puts "You can override this and force removal with `brew uninstall --ignore-dependencies #{keg.name}`."
-        return true
-      end
+      next if dependents.empty?
+
+      dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
+      conjugation = dependents.count == 1 ? "is" : "are"
+      ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
+      puts "You can override this and force removal with `brew uninstall --ignore-dependencies #{keg.name}`."
+      return true
     end
     false
   end
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -25,7 +25,7 @@ module Homebrew
     end
 
     # --ignore-dependencies, to be consistent with install
-    unless ARGV.include?("--ignore-dependencies") || ARGV.homebrew_developer?
+    if !ARGV.include?("--ignore-dependencies") && !ARGV.homebrew_developer?
       all_kegs = kegs_by_rack.values.flatten(1)
       return if check_for_dependents all_kegs
     end
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -15,12 +15,25 @@ module Homebrew
     raise KegUnspecifiedError if ARGV.named.empty?
 
     kegs_by_rack = if ARGV.force?
-      ARGV.named.map do |name|
+      Hash[ARGV.named.map do |name|
         rack = Formulary.to_rack(name)
         [rack, rack.subdirs.map { |d| Keg.new(d) }]
-      end
+      end]
     else
       ARGV.kegs.group_by(&:rack)
+    end
+
+    kegs = kegs_by_rack.values.flatten(1)
+
+    kegs.each do |keg|
+      dependents = keg.installed_dependents - kegs
+      if dependents.any?
+        dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
+        conjugation = dependents.count == 1 ? "is" : "are"
+        ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
+        # puts "You can override this and force removal with `brew uninstall --force #{keg.name}`."
+        next
+      end
     end
 
     kegs_by_rack.each do |rack, kegs|
@@ -38,15 +51,6 @@ module Homebrew
         rm_pin rack
       else
         kegs.each do |keg|
-          dependents = keg.installed_dependents - ARGV.kegs
-          if dependents.any?
-            dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
-            conjugation = dependents.count == 1 ? "is" : "are"
-            ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
-            puts "You can override this and force removal with `brew uninstall --force #{keg.name}`."
-            next
-          end
-
           keg.lock do
             puts "Uninstalling #{keg}... (#{keg.abv})"
             keg.unlink
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -14,47 +14,54 @@ module Homebrew
   def uninstall
     raise KegUnspecifiedError if ARGV.named.empty?
 
-    if !ARGV.force?
-      ARGV.kegs.each do |keg|
-        dependents = keg.installed_dependents - ARGV.kegs
-        if dependents.any?
-          dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
-          conjugation = dependents.count == 1 ? "is" : "are"
-          ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
-          puts "You can override this and force removal with `brew uninstall --force #{keg.name}`."
-          next
-        end
-
-        keg.lock do
-          puts "Uninstalling #{keg}... (#{keg.abv})"
-          keg.unlink
-          keg.uninstall
-          rack = keg.rack
-          rm_pin rack
-
-          if rack.directory?
-            versions = rack.subdirs.map(&:basename)
-            verb = versions.length == 1 ? "is" : "are"
-            puts "#{keg.name} #{versions.join(", ")} #{verb} still installed."
-            puts "Remove all versions with `brew uninstall --force #{keg.name}`."
-          end
-        end
+    kegs_by_rack = if ARGV.force?
+      ARGV.named.map do |name|
+        rack = Formulary.to_rack(name)
+        [rack, rack.subdirs.map { |d| Keg.new(d) }]
       end
     else
-      ARGV.named.each do |name|
-        rack = Formulary.to_rack(name)
+      ARGV.kegs.group_by(&:rack)
+    end
+
+    kegs_by_rack.each do |rack, kegs|
+      if ARGV.force?
         name = rack.basename
 
         if rack.directory?
           puts "Uninstalling #{name}... (#{rack.abv})"
-          rack.subdirs.each do |d|
-            keg = Keg.new(d)
+          kegs.each do |keg|
             keg.unlink
             keg.uninstall
           end
         end
 
         rm_pin rack
+      else
+        kegs.each do |keg|
+          dependents = keg.installed_dependents - ARGV.kegs
+          if dependents.any?
+            dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
+            conjugation = dependents.count == 1 ? "is" : "are"
+            ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
+            puts "You can override this and force removal with `brew uninstall --force #{keg.name}`."
+            next
+          end
+
+          keg.lock do
+            puts "Uninstalling #{keg}... (#{keg.abv})"
+            keg.unlink
+            keg.uninstall
+            rack = keg.rack
+            rm_pin rack
+
+            if rack.directory?
+              versions = rack.subdirs.map(&:basename)
+              verb = versions.length == 1 ? "is" : "are"
+              puts "#{keg.name} #{versions.join(", ")} #{verb} still installed."
+              puts "Remove all versions with `brew uninstall --force #{keg.name}`."
+            end
+          end
+        end
       end
     end
   rescue MultipleVersionsInstalledError => e
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -16,7 +16,7 @@ module Homebrew
 
     if !ARGV.force?
       ARGV.kegs.each do |keg|
-        dependents = keg.installed_dependents
+        dependents = keg.installed_dependents - ARGV.kegs
         if dependents.any?
           dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
           conjugation = dependents.count == 1 ? "is" : "are"
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -21,7 +21,7 @@ module Homebrew
           dependants_output = dependants.map { |k| "#{k.name} #{k.version}" }.join(", ")
           conjugation = dependants.count == 1 ? "is" : "are"
           ofail "Refusing to uninstall #{keg} because it is required by #{dependants_output}, which #{conjugation} currently installed."
-          puts "Remove it anyway with `brew uninstall --force #{keg.name}`."
+          puts "You can override this and force removal with `brew uninstall --force #{keg.name}`."
           next
         end
 
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -16,6 +16,15 @@ module Homebrew
 
     if !ARGV.force?
       ARGV.kegs.each do |keg|
+        dependants = keg.installed_dependants
+        if dependants.any?
+          dependants_output = dependants.map { |k| "#{k.name} #{k.version}" }.join(", ")
+          conjugation = dependants.count == 1 ? "is" : "are"
+          ofail "Refusing to uninstall #{keg} because it is required by #{dependants_output}, which #{conjugation} currently installed."
+          puts "Remove it anyway with `brew uninstall --force #{keg.name}`."
+          next
+        end
+
         keg.lock do
           puts "Uninstalling #{keg}... (#{keg.abv})"
           keg.unlink
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -92,39 +92,6 @@ module Homebrew
     true
   end
 
-  # Will return some kegs, and some dependencies, if they're present.
-  # For efficiency, we don't bother trying to get complete data.
-  def find_some_installed_dependents(kegs)
-    kegs.each do |keg|
-      dependents = keg.installed_dependents - kegs
-      dependents.map! { |d| "#{d.name} #{d.version}" }
-      return [keg], dependents if dependents.any?
-    end
-
-    # Find formulae that didn't have dependencies saved in all of their kegs,
-    # so need them to be calculated now.
-    #
-    # This happens after the initial dependency check because it's sloooow.
-    remaining_formulae = Formula.installed.select { |f|
-      f.installed_kegs.any? { |k| Tab.for_keg(k).runtime_dependencies.nil? }
-    }
-
-    keg_names = kegs.map(&:name)
-    kegs_by_name = kegs.group_by(&:to_formula)
-    remaining_formulae.each do |dependent|
-      required = dependent.missing_dependencies(hide: keg_names)
-      required.select! do |f|
-        kegs_by_name.key?(f)
-      end
-      next unless required.any?
-
-      required_kegs = required.map { |f| kegs_by_name[f].sort_by(&:version).last }
-      return required_kegs, [dependent]
-    end
-
-    nil
-  end
-
   def rm_pin(rack)
     Formulary.from_rack(rack).unpin
   rescue
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -16,11 +16,11 @@ module Homebrew
 
     if !ARGV.force?
       ARGV.kegs.each do |keg|
-        dependants = keg.installed_dependants
-        if dependants.any?
-          dependants_output = dependants.map { |k| "#{k.name} #{k.version}" }.join(", ")
-          conjugation = dependants.count == 1 ? "is" : "are"
-          ofail "Refusing to uninstall #{keg} because it is required by #{dependants_output}, which #{conjugation} currently installed."
+        dependents = keg.installed_dependents
+        if dependents.any?
+          dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
+          conjugation = dependents.count == 1 ? "is" : "are"
+          ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
           puts "You can override this and force removal with `brew uninstall --force #{keg.name}`."
           next
         end
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -16,10 +16,10 @@ module Homebrew
     raise KegUnspecifiedError if ARGV.named.empty?
 
     kegs_by_rack = if ARGV.force?
-      Hash[ARGV.named.map { |name|
+      Hash[ARGV.named.map do |name|
         rack = Formulary.to_rack(name)
         [rack, rack.subdirs.map { |d| Keg.new(d) }]
-      }]
+      end]
     else
       ARGV.kegs.group_by(&:rack)
     end
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -76,7 +76,7 @@ module Homebrew
   end
 
   def check_for_dependents(kegs)
-    return false unless result = find_some_installed_dependents(kegs)
+    return false unless result = Keg.find_some_installed_dependents(kegs)
 
     requireds, dependents = result
 
@@ -108,8 +108,16 @@ module Homebrew
     remaining_formulae = Formula.installed.select { |f|
       f.installed_kegs.any? { |k| Tab.for_keg(k).runtime_dependencies.nil? }
     }
-    Diagnostic.missing_deps(remaining_formulae, kegs.map(&:name)) do |dependent, required|
-      kegs_by_name = kegs.group_by(&:to_formula)
+
+    keg_names = kegs.map(&:name)
+    kegs_by_name = kegs.group_by(&:to_formula)
+    remaining_formulae.each do |dependent|
+      required = dependent.missing_dependencies(hide: keg_names)
+      required.select! do |f|
+        kegs_by_name.key?(f)
+      end
+      next unless required.any?
+
       required_kegs = required.map { |f| kegs_by_name[f].sort_by(&:version).last }
       return required_kegs, [dependent]
     end
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -24,8 +24,7 @@ module Homebrew
       ARGV.kegs.group_by(&:rack)
     end
 
-    # --ignore-dependencies, to be consistent with install
-    if !ARGV.include?("--ignore-dependencies") && !ARGV.homebrew_developer?
+    if should_check_for_dependents?
       all_kegs = kegs_by_rack.values.flatten(1)
       return if check_for_dependents all_kegs
     end
@@ -73,6 +72,13 @@ module Homebrew
         rack.unlink if rack.symlink? && !rack.resolved_path_exists?
       end
     end
+  end
+
+  def should_check_for_dependents?
+    # --ignore-dependencies, to be consistent with install
+    return false if ARGV.include?("--ignore-dependencies")
+    return false if ARGV.homebrew_developer?
+    true
   end
 
   def check_for_dependents(kegs)
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -6,6 +6,7 @@
 
 require "keg"
 require "formula"
+require "diagnostic"
 require "migrator"
 
 module Homebrew
@@ -75,17 +76,45 @@ module Homebrew
   end
 
   def check_for_dependents(kegs)
+    return false unless result = find_some_installed_dependents(kegs)
+
+    requireds, dependents = result
+
+    msg = "Refusing to uninstall #{requireds.join(", ")} because "
+    msg << (requireds.count == 1 ? "it is" : "they are")
+    msg << " required by #{dependents.join(", ")}, which "
+    msg << (dependents.count == 1 ? "is" : "are")
+    msg << " currently installed."
+    ofail msg
+    print "You can override this and force removal with "
+    puts "`brew uninstall --ignore-dependencies #{requireds.map(&:name).join(" ")}`."
+
+    true
+  end
+
+  # Will return some kegs, and some dependencies, if they're present.
+  # For efficiency, we don't bother trying to get complete data.
+  def find_some_installed_dependents(kegs)
     kegs.each do |keg|
       dependents = keg.installed_dependents - kegs
-      next if dependents.empty?
-
-      dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
-      conjugation = dependents.count == 1 ? "is" : "are"
-      ofail "Refusing to uninstall #{keg} because it is required by #{dependents_output}, which #{conjugation} currently installed."
-      puts "You can override this and force removal with `brew uninstall --ignore-dependencies #{keg.name}`."
-      return true
+      dependents.map! { |d| "#{d.name} #{d.version}" }
+      return [keg], dependents if dependents.any?
     end
-    false
+
+    # Find formulae that didn't have dependencies saved in all of their kegs,
+    # so need them to be calculated now.
+    #
+    # This happens after the initial dependency check because it's sloooow.
+    remaining_formulae = Formula.installed.select { |f|
+      f.installed_kegs.any? { |k| Tab.for_keg(k).runtime_dependencies.nil? }
+    }
+    Diagnostic.missing_deps(remaining_formulae, kegs.map(&:name)) do |dependent, required|
+      kegs_by_name = kegs.group_by(&:to_formula)
+      required_kegs = required.map { |f| kegs_by_name[f].sort_by(&:version).last }
+      return required_kegs, [dependent]
+    end
+
+    nil
   end
 
   def rm_pin(rack)
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -23,17 +23,10 @@ module Homebrew
       ARGV.kegs.group_by(&:rack)
     end
 
-    kegs = kegs_by_rack.values.flatten(1)
-
-    kegs.each do |keg|
-      dependents = keg.installed_dependents - kegs
-      if dependents.any?
-        dependents_output = dependents.map { |k| "#{k.name} #{k.version}" }.join(", ")
-        conjugation = dependents.count == 1 ? "is" : "are"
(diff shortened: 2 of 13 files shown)


penman:
Here's what I've tested with:

```sh
brew install feedgnuplot
brew uninstall gnuplot # check it refuses
brew uninstall --force gnuplot # check it uninstalls
brew uninstall feedgnuplot # check it uninstalls (because nothing depends on it)
```

vladshablinsky:
Does it make sense to rename `Keg#formula` to `Keg#to_formula` similar to `Dependency#to_formula` method?

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -14,6 +14,15 @@ def uninstall
 
     if !ARGV.force?
       ARGV.kegs.each do |keg|
+        dependants = keg.installed_dependants
+        if dependants.any?
+          dependants_output = dependants.map { |k| "#{k.name} #{k.version}" }.join(", ")
+          conjugation = dependants.count == 1 ? "is" : "are"
+          ofail "Refusing to uninstall #{keg} because it is required by #{dependants_output}, which #{conjugation} currently installed."
+          puts "Remove it anyway with `brew uninstall --force #{keg.name}`."
```
MikeMcQuaid: Maybe `You can override this and force removal with brew uninstall --force ...`

MikeMcQuaid:
@penman From my reading this just handles direct and not recursive dependencies. What do you think about handling this recursively?

penman:
@MikeMcQuaid it should handle recursive dependencies! It uses the list of dependencies from the tab, which is [based on](https://github.com/Homebrew/brew/blob/bc52932e5ab83d253850cc622c7a0032e9d2de20/Library/Homebrew/formula.rb#L1464) the recursive dependency list.

MikeMcQuaid:
@penman @ilovezfs made a good point to me privately: this probably shouldn't overload `--force` (although I realise that was my idea originally) so instead should be something like `--ignore-dependents` or something?

ilovezfs:
I think the brew code mostly uses "dependent" not "dependant"

penman:
@MikeMcQuaid @ilovezfs so `brew uninstall --force` would also refuse to uninstall a keg with dependents without `--ignore-dependents`?

ilovezfs:
Right, I think the two options shouldn't directly interact with each other.

DomT4:
I wonder if this PR breaks `brew bundle` as-is, given `brew bundle` IIRC doesn't try to remove things in order of leaves but simply alphabetically.

> --ignore-dependents or something?

I'm quite keen to avoid `dependents` or `dependants` because people will absolutely & regularly mistype that flag. It should be something fairly simple, and less likely to be bashed wrongly into the keyboard.

MikeMcQuaid:
> I'm quite keen to avoid dependents or dependants because people will absolutely & regularly mistype that flag. It should be something fairly simple, and less likely to be bashed wrongly into the keyboard.

Was thinking the same (but we probably don't really want people typing it anyway...).

ilovezfs:
`--ignore-dependents` and `--ignore-dependencies` (anyone want to volunteer to write a paragraph of documentation for the man page explaining the difference) seem like developer language spilling over into the CLI.

As for whether we want people typing. If you need to uninstall and reinstall something (for example: we didn't bump the revision but someone needs the change), they're going to have to type it, unless we want to force them to uninstall every dependent, which would almost certainly be unnecessary since otherwise there would have been a revision bump. If this PR is merged, that option will very much so be exposed to users whether we like it or not.

zmwangx:
IMO this is definitely going to be more trouble than its worth, plus it won't solve every problem with uninstall (think opportunistic linking which is plenty). If someone uninstalls something and accidentally break s another, most of the time it's just one bottle-pouring away.

MikeMcQuaid:
> IMO this is definitely going to be more trouble than its worth

This is the default behaviour of pretty much every other package manager. What makes them different to us?

zmwangx:
> This is the default behaviour of pretty much every other package manager. What makes them different to us?

The difference is users are accustomed to this behavior and you're going to break user scripts, include official ones it seems.

MikeMcQuaid:
> The difference is users are accustomed to this behavior and you're going to break user scripts

If the status quo is the only justification I don't think it's good enough to not implement the functionality.

ilovezfs:
>This is the default behaviour of pretty much every other package manager. What makes them different to us?

We're not managing system components. I'd start there.

zmwangx:
ILZ mentioned privately that one major difference with say Linux package managers is that we don't have anything mission critical and uninstalling a few thing wrong will at most leave your video player unfunctional, while on the Linux side if you uninstalling something wrong could leave you with an unbootable machine.

zmwangx:
And what about our "the command line is our API" promise (paraphrased)?

MikeMcQuaid:
@zmwangx Many of our users would disagree that we don't have anything mission critical. I'm open to different approaches to this functionality but needs to go in and be on by default to be useful.

ilovezfs:
When was the last time someone filed an issue due to uninstalling something they shouldn't have? I don't think we actually see that user problem in the wild much since people tend to install things and then leave them alone.

If anything this PR will probably inadvertently cause _more_ such issues by sending people on cleaning sprees since uninstall is "safe" now, when in fact due to undeclared dependencies, dependencies hidden behind`if build.with?` options, and anomalies of the requirement system, it won't actually be safe.

zmwangx:
If this branch really working at the moment?

I'm trying to test a potential bug pointed out by ILZ, that is,

```
brew uninstall foo bar
```

failing when `bar` depends on `foo`.

But then I realized I can uninstall `ant` just fine with `abcl` installed. (After `brew pull https://github.com/Homebrew/brew/pull/1082` of course.)

ilovezfs:
@zmwangx but I bet it fails if you try to uninstall `rlwrap`, yes?

MikeMcQuaid:
> If anything this PR will probably inadvertently cause more such issues by sending people on cleaning sprees since uninstall is "safe" now, when in fact due to undeclared dependencies, dependencies hidden behindif build.with? options, and anomalies of the requirement system, it won't actually be safe.

That doesn't matter because we track the actual used dependencies.

zmwangx:
@ilovezfs: No it happily uninstalled rlwrap. Maybe I'm doing something wrong.

MikeMcQuaid:
> I don't think we actually see that user problem in the wild much since people tend to install things and then leave them alone.

We regularly see issues with `brew missing` output from `brew doctor`.

ilovezfs:
@zmwangx interesting.

ilovezfs:
>We regularly see issues with brew missing output from brew doctor.

doctor complaining about that is not the same as it causing a user reported issue :)

MikeMcQuaid:
I've seen enough of them. I'm also fed up of people complaining in person and on the internets about this being another example of our dependency system being shitty.

MikeMcQuaid:
And if in doubt: we should do what other package managers do unless there's a really good reason.

ilovezfs:
>And if in doubt: we should do what other package managers do unless there's a really good reason.

Be careful what you wish for :)

ilovezfs:
Yup, reproduced what @zmwangx described.

zmwangx:
It appears that for slightly older bottles without `runtime_dependencies` in the install receipts this wouldn't work. Need to rebottle everything for this to work consistently, it seems.

ilovezfs:
And the predicted bug does occur as well
```
iMac-TMP:~ joe$ brew uninstall ant abcl
Error: Refusing to uninstall /usr/local/Cellar/ant/1.9.7 because it is required by abcl 1.3.3, which is currently installed.
Remove it anyway with `brew uninstall --force ant`.
Uninstalling /usr/local/Cellar/abcl/1.3.3... (7 files, 10M)
```

zmwangx:
Follow up to my last comment: even if we rebooted everything,, we can't affect existing installations. So every existing user's cellar will become a wonderful mix of inconsistency. Of course, you need someone who actually cares to notice.

ilovezfs:
We could install a SQLite database with the relevant information so that it would work for existing installations too.

zmwangx:
With info of all the versions that have ever existed? And for user taps?

MikeMcQuaid:
Can we try to focus on solutions instead of just problems, please?

zmwangx:
I don't know about solutions, but I can at least point out problems so that this doesn't get merged and lead to a catastrophe.

ilovezfs:
A solution would be to rewrite all of the `INSTALL_RECEIPT.json` files for the installed formulae in a user's Cellar based on information from `$somwhere`.

MikeMcQuaid:
> this doesn't get merged and lead to a catastrophe.

Isn't the catastrophe in this case: `brew uninstall` has the previous behaviour unless you reinstall/rebottle?

zmwangx:
> Isn't the catastrophe in this case: brew uninstall has the previous behaviour unless you reinstall/rebottle?

The catastrophe is inconsistent behavior that's going to be very confusing.

penman:
The inconsistency here is that under some circumstances, we can stop a user accidentally doing something they probably don't want to do with bad consequences for the user.

Why does it matter that we can't do that all the time? There's no need to announce "hey everyone you can run brew uninstall on all your formulae it's 100% safe and won't remove anything you need", so it's not like people are going to start relying on the error.

MikeMcQuaid:
I agree.

penman:
(I realise there a are a few things that still need fixed in this PR, but I'm going to leave it for a couple of days so you all can decide if you want it before I put the effort in.)

zmwangx:
> Why does it matter that we can't do that all the time?

Good luck explaining that.

MikeMcQuaid:
> Good luck explaining that.

Well, @penman won't need to explain anything because she isn't a maintainer here (yet, at least).

I'm in favour of quietly shipping this and not documenting it for now and iterating unless anyone can point out a way it breaks on its actual intent.

ilovezfs:
>I'm in favour of quietly shipping this and not documenting it for now and iterating unless anyone can point out a way it breaks on its actual intent.

That has indeed been pointed out https://github.com/Homebrew/brew/pull/1082#issuecomment-249373228

ilovezfs:
Extending `Diagnostic.missing_deps` would depend on neither up-to-date receipts nor the order in which the arguments are passed. Quick proof of concept:
```
Josephs-MacBook-Pro:Homebrew joe$ git diff
diff --git a/Library/Homebrew/cmd/missing.rb b/Library/Homebrew/cmd/missing.rb
index bbd5142..abdd01e 100644
--- a/Library/Homebrew/cmd/missing.rb
+++ b/Library/Homebrew/cmd/missing.rb
@@ -16,7 +16,7 @@ module Homebrew
       ARGV.resolved_formulae
     end
 
-    Diagnostic.missing_deps(ff) do |name, missing|
+    Diagnostic.missing_deps(ff, ARGV.value("hide")) do |name, missing|
       print "#{name}: " if ff.size > 1
       puts (missing * " ").to_s
     end
diff --git a/Library/Homebrew/diagnostic.rb b/Library/Homebrew/diagnostic.rb
index 743c653..29c3e07 100644
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -7,7 +7,7 @@ require "utils/shell"
 
 module Homebrew
   module Diagnostic
-    def self.missing_deps(ff)
+    def self.missing_deps(ff, hide = nil)
       missing = {}
       ff.each do |f|
         missing_deps = f.recursive_dependencies do |dependent, dep|
@@ -20,7 +20,13 @@ module Homebrew
         end
 
         missing_deps.map!(&:to_formula)
-        missing_deps.reject! { |d| d.installed_prefixes.any? }
+        if hide
+          missing_deps.reject! do |d|
+            !hide.include?(d.name) && d.installed_prefixes.any?
+          end
+        else
+          missing_deps.reject! { |d| d.installed_prefixes.any? }
+        end
 
         unless missing_deps.empty?
           yield f.full_name, missing_deps if block_given?
```

```
Josephs-MacBook-Pro:Homebrew joe$ brew missing --hide=mpfr
gawk: mpfr
libmpc: mpfr
Josephs-MacBook-Pro:Homebrew joe$ brew missing --hide=mpfr,readline
bash: readline
gawk: mpfr readline
libmpc: mpfr
python3: readline
ruby: readline
sqlite: readline
```

penman:
@ilovezfs would that be less accurate? If you're not reading from the install receipt, a dependency could have been added in a later version of a formula that isn't relevant to the currently installed version, right? There's a reason runtime dependencies are stored at install time in the first place.

I think the argument order is pretty easy to work around, by the way. Haven't tested, but I think a `dependants -= ARGV.formulae` before checking if there any would do the trick.

ilovezfs:
>would that be less accurate?

It will be no more and no less accurate than `brew missing`.

MikeMcQuaid:
Just to be explicit where it's been more implicit: I opened https://github.com/Homebrew/brew/issues/934 and planned on working on this myself. @penman has picked this up before I did (hence the `help-wanted` tag) and it will be merged once we figure out how best to document and implement it. If anyone has specific review or suggestions for the documentation or implementation they are very welcome but a 👎 on the feature as a whole is not.

ilovezfs:
In terms of the option naming, I was wondering if instead of
--force retains current behavior
--ignore-dependents overrides the removal block

it would be more user friendly language to have it be

--force overrides the removal block
--all current behavior of --force

I'm happy either way, as long as we're not overloading --force, but thought that ceding --force to this PR, and renaming current --force behavior to --all might address @DomT4's concern above.

In terms of tweaks:
A) as already discussed above, no overloading of --force
B) as discussed on Slack, making overriding removal the default if HOMEBREW_DEVELOPER is set. I have no desire for an opt-in option when HOMEBREW_DEVELOPER is set, but I could imagine maybe someone else might? --safe or --no-force or --no-ignore-dependents come to mind if anyone cares about an opt-in.
C) as already discussed above, if the dependent that is blocking the removal is on the list of things to be removed, order of the arguments to uninstall should not matter, and the uninstall should proceed without refusing or requiring the user to reorder the arguments

Potential enhancements:
D) in light of (C), doing the uninstalls in the least dangerous order might be advisable, but uninstalls are pretty unlikely to fail except on weird permissions stumbling blocks
E) consider falling back to the mechanism I suggested above for extending the missing diagnostic when any of the receipts are old to avoid the problem zmwangx found

DomT4:
> If anyone has specific review or suggestions for the documentation or implementation they are very welcome but a 👎 on the feature as a whole is not.

Not especially specific to this, or @penman's excellent work here & around Homebrew in recent weeks, I don't think it's unreasonable to 👎 a feature if we explore how to do it and people end up thinking its impractical or causes more confusion than it resolves. I don't think because a maintainer wants x feature that x feature absolutely has to become part of Homebrew. 

Given how much `upgrade --all` ended up causing us hassle and remains in people's aliases, etc to this day I don't think considering these things in depth & with flexibility on outcome is particularly a bad idea.

I agree there's a lot of improvement room for Homebrew around being able to yank dependencies out of existence even though they are very much required, and broadly I think the idea pushed here isn't a bad one as long as there are opt-outs for developers in particular, but bearing in mind if we break Homebrew now users have to wait for the next tag for things to become unbroken it's not unfair to be hyper-careful.

> In terms of tweaks:
A) as already discussed above, no overloading of --force
B) as discussed on Slack, making overriding removal the default if HOMEBREW_DEVELOPER is set. I have no desire for an opt-in option when HOMEBREW_DEVELOPER is set, but I could imagine maybe someone else might? --safe or --no-force or --no-ignore-dependents come to mind if anyone cares about an opt-in.
C) as already discussed above, if the dependent that is blocking the removal is on the list of things to be removed, order of the arguments to uninstall should not matter, and the uninstall should proceed without refusing or requiring the user to reorder the arguments

This seems like a reasonable list to me, at least.

penman:
> B) as discussed on Slack, making overriding removal the default if HOMEBREW_DEVELOPER is set. I have no desire for an opt-in option when HOMEBREW_DEVELOPER is set, but I could imagine maybe someone else might?

@ilovezfs I obviously can't see the conversation on Slack, but as a `HOMEBREW_DEVELOPER` I'd like dangerous uninstalls to be blocked by default. If I'm accidentally installing something without realising something else depends on it, I'd like to know about it, and if I'm doing it deliberately typing `--force` or something to indicate my intention wouldn't be too much overhead.

penman:
> In terms of the option naming, I was wondering if instead of
> --force retains current behavior
> --ignore-dependents overrides the removal block

> it would be more user friendly language to have it be

> --force overrides the removal block
> --all current behavior of --force

I like this idea. `--force` never made sense to me as the name for this.

zmwangx:
> but as a HOMEBREW_DEVELOPER I'd like dangerous uninstalls to be blocked by default.

As another HOMEBREW_DEVELOPER I vote for "dangerous uninstalls" to not be blocked by default, again.

> If I'm accidentally installing something without realising something else depends on it, I'd like to know about it,

I assume you meant "accidentally uninstalling something". In that case, at least 90% of the time you'll find out when you get a very clear dylib not found error.

> and if I'm doing it deliberately typing --force or something to indicate my intention wouldn't be too much overhead.

I'll say this again: it may not be a problem or overhead for you because you don't have any script that depends on it, but you *will* break user scripts.

And let me respond to an earlier comment in this thread.

> Many of our users would disagree that we don't have anything mission critical.

Then maybe said users should learn their mission critical software, including their dependency chains, and not mess with them unnecessarily. This is pretty much the general attitude with any mission-critical software deployment.

penman:
> I'll say this again: it may not be a problem or overhead for you because you don't have any script that depends on it, but you will break user scripts.

@zmwangx Fair enough. Ignoring for `HOMEBREW_DEVELOPER`s seems like a reasonable compromise.

MikeMcQuaid:
> I assume you meant "accidentally uninstalling something". In that case, at least 90% of the time you'll find out when you get a very clear dylib not found error.

I don't think those errors are universally understood or debuggable.

> I'll say this again: it may not be a problem or overhead for you because you don't have any script that depends on it, but you will break user scripts.

Which user scripts? Please provide links. Also, again, stop making this personal and keep it technical.

> Then maybe said users should learn their mission critical software, including their dependency chains, and not mess with them unnecessarily. This is pretty much the general attitude towards any mission-critical software deployment.

We should take responsibility for providing safety for our users.

In the interests of actually getting this 🚢d we'll disable it on HOMEBREW_DEVELOPER at first but longer-term we're going to figure out something better than just adding HOMEBREW_DEVELOPER to things a few maintainers don't like. Every other package manager does things this way and it's a good idea. If you have a flag to override and don't like typing it: you can create an alias.

MikeMcQuaid:
> I don't think it's unreasonable to 👎 a feature if we explore how to do it and people end up thinking it's impractical or causes more confusion than it resolves. I don't think because a maintainer wants x feature that x feature absolutely has to become part of Homebrew.

To clarify a bit more after thinking about this some more: I think it's unreasonable for us as a project to open issues like https://github.com/Homebrew/brew/issues/934, mark them as "help wanted" and then to 👎 the feature for a non-maintainer who comes around and implements it. That makes people not want to submit pull requests here and not trust our "help wanted" which in turn makes it harder to keep this project alive long-term. In this case I should have personally been quicker to develop product feedback (i.e. discussion about the merits of the feature rather than the implementation details in this PR) into that issue as the product side of this is on me and not @penman. 

> Given how much upgrade --all ended up causing us hassle and remains in people's aliases, etc to this day I don't think considering these things in depth & with flexibility on outcome is particularly a bad idea.

I don't think it caused us much hassle, really. It doesn't really matter if it's in aliases if it's a no-op. Sometimes we're going to go in a product direction that we later reverse on. That's actively a good thing as without being willing to experiment and make mistakes we will be unable to progress.

> I agree there's a lot of improvement room for Homebrew around being able to yank dependencies out of existence even though they are very much required, and broadly I think the idea pushed here isn't a bad one as long as there are opt-outs for developers in particular, but bearing in mind if we break Homebrew now users have to wait for the next tag for things to become unbroken it's not unfair to be hyper-careful.

Sure. That said, there's nothing stopping us pushing two tags e.g. 5m apart if things are ever broken that badly.

DomT4:
> we'll disable it on HOMEBREW_DEVELOPER at first but longer-term we're going to figure out something better than just adding HOMEBREW_DEVELOPER to things a few maintainers don't like.

This isn't a fair reflection of my stance on the issue, at least. We often leave breaking things on, or turn things from a soft breakage into hard breakage, for developers. Not protecting developers is the default, and I think a perfectly sane default. That was the root of my agreement with ILZ on this particular point. I don't think it's necessarily prudent to start trying to read motivations into comments too much.

> To clarify a bit more after thinking about this some more: I think it's unreasonable for us as a project to open issues like #934

This goes back to my original comment though. You opened the issue, not @HomebrewBot or whatever.

As far as anyone looking into that thread knows the issue was never discussed amongst the maintainer team, so yeah, if a single maintainer opening issues and marking it `help wanted` is merit enough for inclusion of something in Homebrew we're back to my original point; we shouldn't be judging ideas on theoretical discussions but real implementations such as this one. Otherwise I'd quite like a function where Homebrew triggers my kettle pouring a cup of tea every time I type `brew install qt5 -s` or `brew install llvm -s`.

If the real implementation is problematic in some way it's not unreasonable to ask again if we want to do this. And I support the broad idea of making it harder to break other formulae installed, always have, but the points raised above aren't entirely invalid or worthy of being ignored.

Could a better job have been done communicating them? Absolutely. @penman completely has the right to feel put off by 30 or so messages pulling flaws out of the idea, and ideally in future brew PRs we'll be better at proposing fixes more proactively, but it doesn't invalidate the concerns either party had.

> I don't think it caused us much hassle, really.

I was counting user confusion amongst the hassle. Even this far down the line from it being reverted at least once a week I end up telling people the `--all` is no longer required. It may not be bothering us, but there's evidently still users out there who think `--all` is mandated.

penman:
> It may not be bothering us, but there's evidently still users out there who think `--all` is mandated.

I still thought I had to use `--all` until ~last week when I read the code! Not that it caused me any major incovience using `--all` all those times.

penman:
Am I correct in thinking this is what's left to be done on this?:

- [ ] Disable checking for HOMEBREW_DEVELOPERs
- [ ] Make this check independent of `--force`

  @ilovezfs talked above about switching the options (making the current `--force` behaviour use `--all` instead, and making `--force` skip dependant checks instead). A potential issue with that approach is that changing what `--force` does might surprise some users used to the old behaviour. Maybe it would be better to completely deprecate `--force` and come up with two new option names?

  Either solution works for me, just need to know which one to do.

MikeMcQuaid:
> This isn't a fair reflection of my stance on the issue, at least. We often leave breaking things on, or turn things from a soft breakage into hard breakage, for developers. Not protecting developers is the default, and I think a perfectly sane default. That was the root of my agreement with ILZ on this particular point. I don't think it's necessarily prudent to start trying to read motivations into comments too much.

Sorry, to be clear I'm not saying any maintainer is taking that position but just that it's not something that's sustainable.

> As far as anyone looking into that thread knows the issue was never discussed amongst the maintainer team, so yeah, if a single maintainer opening issues and marking it help wanted is merit enough for inclusion of something in Homebrew we're back to my original point

The lack of discussion is on the maintainers who had that appear in their inboxes and said nothing about it. Fair or unfair I take silence on a feature issue as tacit approval.

> we shouldn't be judging ideas on theoretical discussions but real implementations such as this one.

I disagree. We should judge ideas on the idea (i.e. is this even a good idea) and judge implementations on the implementation (i.e. is this code going to do what's described). That prevents people wasting time on implementing something where people disagree with a one-sentence summary of the idea.

zmwangx:
Before I go into what I really want to say, let me make this clear: I (at least as a member of the project) appeciate @penman's past contributions and work in this PR, and although I disagree with the feature, I fully respect the effort that goes in here. None of my comments were personal, whether there was room for other interpretations.

While this discussion might have been a better fit for #934, damage has already been done by the "help wanted" tag, so I'll just continue here.

In case anyone is still concerned about my tone, please contact me directly.

---

Regarding

> Which user scripts? Please provide links. Also, again, stop making this personal and keep it technical.

I can write a one line script on spot:

```sh
for f in $(brew ls); do brew uninstall $f; done
```

That's a script from me, a user. It might feel like trolling, but it's not. Uninstalling one thing at a time is perfectly reasonable, in fact `--help` doesn't even allude to the fact that you can specify multiple formulae:

```
$ brew uninstall -h
brew uninstall, rm, remove [--force] formula:
    Uninstall formula.

    If --force is passed, and there are multiple versions of formula
    installed, delete all installed versions.
```

User scripts don't need to be published user scripts. I have many personal scripts which I don't publish; same can be said for many people.

Now, one might dismiss this as not enough evidence, but I also don't see evidence of legacy behavior causing problems. If "people complaining in person and on the internets about this being another example of our dependency system being shitty" is evidence, may I politely ask them to report the problem to our issue tracker? If they don't, I would argue they aren't really concerned enough, and people complain about random things on the Internet all the time it isn't enough reason to worry.

---

Regarding #934 and help wanted, I agree with Dom that

> we shouldn't be judging ideas on theoretical discussions but real implementations such as this one.

I'm pretty sure that was also the attitude of at least certain maintainers on my past feature requests or feature requests I have monitored: one needs a concrete implementation to discuss the merits and pitfalls of certain features.

Of course features proposed by maintainers should carry more weight and are more likely to be included, but IMO the approval and quality assurance process shouldn't be very much different and doubts should be taken seriously.

---

Regarding

> ideally in future brew PRs we'll be better at proposing fixes more proactively

Ideally yes, but in reality it's hard, unfortunately. For instance, when I posted https://github.com/Homebrew/brew/pull/1082#issuecomment-249370925 I didn't even know the cause yet. I don't think anyone deserves a "shut up about problems until you find a solution", even team members.

---

> I still thought I had to use --all until ~last week when I read the code! Not that it caused me any major incovience using --all all those times.

IMO the problem isn't the inconvenience of `--all`. The problem lies in the period between when a user first finds out that their daily routine is suddenly broken and when they finally finds out that the behavior has changed and they need an additional `--all` flag. This is the period where we could cause user confusion, undermine user trust, create a lot of support burden for ourselves, and potentially lose users.

(Maintainers, if you have time, please read for instance https://news.ycombinator.com/item?id=12546688, and reflect on whether we lived up to that comment recently. As a Homebrew user myself I would almost concur with that comment until recent events. That's why I really hope we think thrice and test carefully before introducing breaking behaviors going forward.)

---

> Am I correct in thinking this is what's left to be done on this?

I still disagree on this feature as a whole, but I'm afraid I'm not and won't be heard.

zmwangx:
> We should judge ideas on the idea (i.e. is this even a good idea) and judge implementations on the implementation (i.e. is this code going to do what's described).

Sorry, pretty sure there's double standard here. See comment above.

MikeMcQuaid:
> Disable checking for HOMEBREW_DEVELOPERs
>  Make this check independent of --force

@penman Yep! We should have a new argument called `--ignore-dependencies` (to match `brew install`s; I think the consistency is valuable unless there's other suggestions for names) and if you have `HOMEBREW_DEVELOPER` set it defaults to being like that parameter was passed.

MikeMcQuaid:
> Now, one might dismiss this as not enough evidence, but I also don't see evidence of legacy behavior causing problems. If "people complaining in person and on the internets about this being another example of our dependency system being shitty" is evidence, may I politely ask them to report the problem to our issue tracker? If they don't, I would argue they aren't really concerned enough, and people complain about random things on the Internet all the time it isn't enough reason to worry.

That's not how product works if you want to make something that's good for all users rather than just users with the time and inclination to post on our issue tracker.

MikeMcQuaid:
> This goes back to my original comment though. You opened the issue, not @HomebrewBot or whatever.
> 
> As far as anyone looking into that thread knows the issue was never discussed amongst the maintainer team, so yeah, if a single maintainer opening issues and marking it help wanted is merit enough for inclusion of something in Homebrew we're back to my original point

With that in mind: any maintainers reading this thread: please read the issues in https://github.com/homebrew/brew/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22 and make any comments you have in those issues, thanks.

MikeMcQuaid:
@penman One more thing: as pointed out by @ilovezfs earlier it's probably worth using something like:

> Extending Diagnostic.missing_deps would depend on neither up-to-date receipts nor the order in which the arguments are passed.

as a fall back when the dependencies in the receipt are not available. This will make it slightly less performant but over time as receipts get updated this will become negligible and could even be removed.

DomT4:
> The lack of discussion is on the maintainers who had that appear in their inboxes and said nothing about it. Fair or unfair I take silence on a feature issue as tacit approval.

> With that in mind: any maintainers reading this thread: please read the issues in https://github.com/homebrew/brew/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22 and make any comments you have in those issues, thanks.

To be honest, I thought we'd killed issues like that in favour of the evolution process, which was set up to stop us opening these vague tracking issues which people didn't really contribute any discussion towards until PRs became available _(if PRs ever became available, which they mostly did not)_ on the subject.

Like [this document](https://github.com/Homebrew/brew-evolution/blob/master/003-install-bottles-of-arbitrary-versions.md) is a lot easier to discuss because it's a detailed proposal, with justification explained. I don't understand what we gain with an issue like [this](https://github.com/Homebrew/brew/issues/931) on top of that. I look at that issue and go _"... okay, why? What's the benefit? What's the goal? Is that goal desirable?"_. It feels like we're duping evolution, but perhaps I'm misunderstanding your intention.

Going back to my intentionally ridiculous example of getting `brew install <long-build-time-formula> -s` to ping my kettle to refill me on caffeine so I don't fall asleep on the laptop waiting for it to finish, I'm sure I could convince most developers it's not a _bad idea_, but there's no sensible way to implement that and if I can't show an implementation people broadly agree with surely the idea is more or less dead.

And again, I'll take a moment to say that doesn't apply to @penman's work here, which is neither a bad idea or bad implementation, merely an implementation there were valid concerns on which are in the process of being addressed. And my apologies to @penman for the lengthy comments in an already lengthy thread, I'm happy to continue this discussion off-thread as desired.

> We should have a new argument called --ignore-dependencies (to match brew installs; I think the consistency is valuable unless there's other suggestions for names) and if you have HOMEBREW_DEVELOPER set it defaults to being like that parameter was passed.

👍.

MikeMcQuaid:
> To be honest, I thought we'd killed issues like that in favour of the evolution process, which was set up to stop us opening these vague tracking issues which people didn't really contribute any discussion towards until PRs became available (if PRs ever became available, which they mostly did not) on the subject.

We had for user-submitted feature requests but maintainer-submitted issues are ones we've already provided some level of green light for or are actively planning ourselves.

> Like this document is a lot easier to discuss because it's a detailed proposal, with justification explained. I don't understand what we gain with an issue like this on top of that.

Having something in which we can track the work to be done and have it in projects/milestones/etc. The "help wanted" label is a common OSS pattern and there's various aggregator sites that point contributors in that direction.

> Going back to my intentionally ridiculous example of getting brew install <long-build-time-formula> -s to ping my kettle to refill me on caffeine so I don't fall asleep on the laptop waiting for it to finish, I'm sure I could convince most developers it's not a bad idea, but there's no sensible way to implement that and if I can't show an implementation people broadly agree with surely the idea is more or less dead.

In that case (and in this!) those criticisms can be raised from both a product and technical perspective without needing to see code.

> And again, I'll take a moment to say that doesn't apply to @penman's work here, which is neither a bad idea or bad implementation, merely an implementation there were valid concerns on which are in the process of being addressed. And my apologies to @penman for the lengthy comments in an already lengthy thread, I'm happy to continue this discussion off-thread as desired.

The comments are welcome but agreed it's probably worth leaving them for now or pulling to another issue 👍

penman:
Still need to integrate the `brew missing` fallback. Getting there!

alyssais:
As expected, adding the `Diagnostic.missing_deps` fallback has slowed down uninstall. On my machine, with ~300 kegs installed, it took about a second per formula to be uninstalled to check if it had dependencies. The current implementation calculates `recursive_dependencies` for every formula that was installed since the user got #1032. This means that the extra time this takes will decrease to zero over time.

We could probably get around this with some caching, but it would probably require quite a bit of change in the test suite to clear the cache where appropriate.

Since this problem will eventually go away on its own, worth it?

alyssais:
An alternative to caching could be pulling the call to `Diagnostic.missing_deps` from `Keg#installed_dependencies` into `Homebrew#uninstall` so all the formulae can be passed as the `hide` argument at once. This would get messy in `uninstall.rb` and be difficult to test efficiently, though.

ilovezfs:
+1 to simplifying it down to a single call passing the full list for hide if you can

alyssais:
I'll have a go, @ilovezfs. :)

alyssais:
I'm going to tentatively say I think this is ready now.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -6,44 +6,58 @@
...
-            verb = versions.length == 1 ? "is" : "are"
-            puts "#{keg.name} #{versions.join(", ")} #{verb} still installed."
-            puts "Remove all versions with `brew uninstall --force #{keg.name}`."
-          end
-        end
-      end
-    else
-      ARGV.named.each do |name|
+    kegs_by_rack = if ARGV.force?
+      Hash[ARGV.named.map { |name|
```
MikeMcQuaid: Probably worth using a `do` if it's multi-line?

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -6,44 +6,58 @@
...
         rack = Formulary.to_rack(name)
+        [rack, rack.subdirs.map { |d| Keg.new(d) }]
+      }]
+    else
+      ARGV.kegs.group_by(&:rack)
+    end
+
+    # --ignore-dependencies, to be consistent with install
+    unless ARGV.include?("--ignore-dependencies") || ARGV.homebrew_developer?
```
MikeMcQuaid: Make an `if` as `unless ||` makes my brain hurt.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -59,6 +73,48 @@ def uninstall
...
+      dependents = keg.installed_dependents - kegs
+      dependents.map! { |d| "#{d.name} #{d.version}" }
+      return [keg], dependents if dependents.any?
+    end
+
+    # Find formulae that didn't have dependencies saved in all of their kegs,
+    # so need them to be calculated now.
+    #
+    # This happens after the initial dependency check because it's sloooow.
+    remaining_formulae = Formula.installed.select { |f|
```
MikeMcQuaid: Worth having a cache here so it can be shared between multiple formulae?

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -59,6 +73,48 @@ def uninstall
...
+      f.installed_kegs.any? { |k| Tab.for_keg(k).runtime_dependencies.nil? }
+    }
+    Diagnostic.missing_deps(remaining_formulae, kegs.map(&:name)) do |dependent, required|
```
MikeMcQuaid: Probably want to extract this method from `Diagnostic` into somewhere else that makes sense to be shared.

Library/Homebrew/keg.rb
```diff
@@ -292,6 +292,23 @@ def version
     PkgVersion.parse(path.basename.to_s)
   end
 
+  def to_formula
+    Formulary.from_keg(self)
+  end
+
+  def installed_dependents
```
MikeMcQuaid: Maybe worth reference the `tab` in the name as it's not checking the fallback?

MikeMcQuaid: Or putting the fallback code on here rather than on `uninstall`?

Library/Homebrew/cmd/uninstall.rb
jawshooah: AFK so can't verify right now, but I believe the new RuboCop rules enforce the semantic style for blocks, rather than always using `do...end` for multiline.

alyssais: That's how I wrote them, but Rubocop was unhappy

Library/Homebrew/keg.rb
alyssais: I had the fallback code in here, but it meant we had to call `missing_deps` multiple times; see earlier conversation between me and @ilovezfs.

Happy to rename though.

Library/Homebrew/cmd/uninstall.rb
alyssais: Yep. It's "use braces for functional blocks" or something

Library/Homebrew/cmd/uninstall.rb
alyssais: Not sure what you mean? This method is only called once per run.

Library/Homebrew/cmd/uninstall.rb
MikeMcQuaid: Weird, ignore me then.

Library/Homebrew/cmd/uninstall.rb
MikeMcQuaid: I see that now, ignore me!

Library/Homebrew/keg.rb
MikeMcQuaid: Could make it a static method or pull out into another class entirely so there's less going on in `uninstall`.

Library/Homebrew/cmd/uninstall.rb
reitermarkus: Yes, using braces for functional blocks is nicer for method chaining.

An example from `caveats.rb`:

```ruby
    @keg ||= [f.prefix, f.opt_prefix, f.linked_keg].map do |d|
      begin
        Keg.new(d.resolved_path)
      rescue
        nil
      end
    end.compact.first
```

vs.

```ruby
    @keg ||= [f.prefix, f.opt_prefix, f.linked_keg].map { |d|
      begin
        Keg.new(d.resolved_path)
      rescue
        nil
      end
    }.compact.first
```

MikeMcQuaid: The prior reads nicer to me FWIW 😕

alyssais: Me too

reitermarkus: Well, what would we be without different tastes? 😄 

I'd really like to keep this though, as you know at first sight if the return value of the block is used or not.

MikeMcQuaid: I don't agree unfortunately; I think the method-names do that alone.

MikeMcQuaid: Given what Rubocop says today: feel free to leave as-is.

Library/Homebrew/cmd/missing.rb
```diff
@@ -16,8 +16,13 @@ def missing
       ARGV.resolved_formulae
     end
 
-    Diagnostic.missing_deps(ff) do |name, missing|
-      print "#{name}: " if ff.size > 1
+    hide = (ARGV.value("hide") || "").split(",")
```
MikeMcQuaid: Can you document this arg at the top of the file? Thanks!

Library/Homebrew/tab.rb
```diff
@@ -60,7 +60,6 @@ def self.from_file(path)
   def self.from_file_content(content, path)
     attributes = Utils::JSON.load(content)
     attributes["tabfile"] = path
-    attributes["runtime_dependencies"] ||= []
```
MikeMcQuaid: ❓

alyssais: We need to be able to differentiate between "there aren't any runtime dependencies" and "runtime dependencies weren't specified" to know whether to use the fallback.

MikeMcQuaid: Great, thanks! Was just curious and that makes sense 👍

alyssais:
I have no idea why this would be failing (seemingly only on Linux???)

MikeMcQuaid:
@alyssais I've got a Linux VM handy, I'll take a look at this when I get the chance.

Library/Homebrew/test/test_uninstall.rb
```diff
@@ -1,8 +1,135 @@
...
   def test_uninstall
     cmd("install", testball)
     assert_match "Uninstalling testball", cmd("uninstall", "--force", testball)
+    assert_empty Formulary.factory(testball).installed_kegs
+  end
+
+  def test_uninstall_with_unrelated_missing_deps_in_tab
+    setup_test_formula "testball"
+    run_as_not_developer do
+      cmd("install", testball)
```
MikeMcQuaid: This is failing because `testball` and `"testball"` are not interchangeable in this case. Additionally, I think all the new tests in this file should be made into non-integration tests instead; the integration tests are meant just as an overall smoke test and are too slow to be used for more complex cases like this.

alyssais:
Haven't forgotten about this, just so you know! This is a busy week.

MikeMcQuaid:
@alyssais No rush!

alyssais:
> Additionally, I think all the new tests in this file should be made into non-integration tests instead; the integration tests are meant just as an overall smoke test and are too slow to be used for more complex cases like this.

Agree in principle. I might leave one or two for now just to make sure that the options are understood correctly, and move the rest somewhere else. (Although, it sure would be nice if they were just faster 🤔.)

MikeMcQuaid:
> Agree in principle. I might leave one or two for now just to make sure that the options are understood correctly, and move the rest somewhere else. (Although, it sure would be nice if they were just faster 🤔.)

I think it's probably still worth just removing them. Agreed about them being faster but integration tests are always going to be significantly slower so I think we really only want one per command.

alyssais:
Okay, I'll remove the two extra integration tests. (I'm just desperate to get this merged, heh.)

However, I'm still very wary of leaving Homebrew with no tests ensuring that it doesn't inadvertently block uninstalling things that shouldn't be blocked.

MikeMcQuaid:
> However, I'm still very wary of leaving Homebrew with no tests ensuring that it doesn't inadvertently block uninstalling things that shouldn't be blocked.

I appreciate the concern ❤️ but we'll do some manual testing and you've added some new unit tests which should help here.

MikeMcQuaid:
Something else that could be useful is writing out manual tests cases you've done so I can evaluate and rerun them.

alyssais:
I totally missed these comments. Oops!

This should be a pretty complete test:

```sh
# These probably should all be run with HOMEBREW_BUILD_FROM_SOURCE set.

# Run the follow tests as a developer

brew reinstall abcl
brew uninstall ant # should succeed

# Run the following tests as a non-developer

brew reinstall abcl
brew uninstall ant # should fail
brew uninstall --force ant # should fail
brew uninstall --ignore-dependencies ant # should work

brew reinstall abcl
brew uninstall abcl # should work

brew reinstall abcl
brew uninstall ant abcl # should work

brew reinstall
brew uninstall abcl ant # should work

brew reinstall abcl
brew uninstall rlwrap # should work

# To test the fallback for formulae installed before runtime
# dependencies were recorded, delete the runtime_dependencies
# array from the abcl INSTALL_RECEIPT after every `reinstall`.
```

Library/Homebrew/cmd/missing.rb
```diff
@@ -1,6 +1,9 @@
-#:  * `missing` [<formulae>]:
+#:  * `missing` [`--hide=`<hidden>] [<formulae>]:
 #:    Check the given <formulae> for missing dependencies. If no <formulae> are
 #:    given, check all installed brews.
+#:
+#:    If `--hide=`<hidden> is passed, act as if none of <hidden> are installed.
+#:    <hidden> should be a comma-seperated list of formulae.
```
bfontaine: Typo: `seperated` ➡️  `separated`

alyssais:
Anything else I should be doing that I've missed?

MikeMcQuaid:
Sorry, on vacation so will be a bit before I can test/merge this.

alyssais:
No problem, just making sure. :)

MikeMcQuaid:
Great work @alyssais! My testing is all good and I'll merge this tomorrow.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -61,6 +77,30 @@ def uninstall
...
+    return false unless result = Keg.find_some_installed_dependents(kegs)
+
+    requireds, dependents = result
+
+    msg = "Refusing to uninstall #{requireds.join(", ")} because "
+    msg << (requireds.count == 1 ? "it is" : "they are")
+    msg << " required by #{dependents.join(", ")}, which "
+    msg << (dependents.count == 1 ? "is" : "are")
+    msg << " currently installed."
+    ofail msg
```
MikeMcQuaid: It can be in another PR but I reckon a non-fatal version of this message should be printed in the developer case as a "Warning" rather than "Error".

alyssais: Good catch. I'll do a separate PR.

MikeMcQuaid:
🎉 Great work @alyssais!

Library/Homebrew/cmd/uninstall.rb
alyssais: I'm mostly done already! Just need to fix a test failure.

MikeMcQuaid: Great, looking forward to it!

PR #1041, opened 2016-09-20
Parallelise test suite

penman:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Integrates [parallel_tests](https://github.com/grosser/parallel_tests) with `brew tests` and `brew cask-tests`, as suggested in #933.

- `brew tests` now runs from `Library/Homebrew`, not `Library/Homebrew/test`, because parallel_tests demands it be so. I had to modify one test to make this work.
- Can somebody look at brew-cask-tests and check the code coverage is still set up right?
- The test output is weird because the different processes all output on top of each other.
- On my machine, there's a slight speed improvement, but a pretty small one.

PR #1084, opened 2016-09-22
tests: speed up integration tests coverage

penman:

-----

This is the speed-up I was referring to in #1041.

Previously, .simplecov called `SimpleCov.result` to store the coverage result, and ignored the return value. `SimpleCov.result`'s return can be slow to calculate, which wastes a lot of time when it's ignored.

This pull request extracts the code needed to store the SimpleCov result from `SimpleCov.result`, and calls it directly, without doing the busywork to compute the return value every time.

In my testing, this more than halves the time taken to run all the integration tests.

### DIFF
--- a/Library/Homebrew/.simplecov
+++ b/Library/Homebrew/.simplecov
@@ -20,7 +20,13 @@ SimpleCov.start do
     at_exit do
       exit_code = $!.nil? ? 0 : $!.status
       $stdout.reopen("/dev/null")
-      SimpleCov.result # Just save result, but don't write formatted output.
+
+      # Just save result, but don't write formatted output.
+      coverage_result = Coverage.result
+      SimpleCov.add_not_loaded_files(coverage_result)
+      simplecov_result = SimpleCov::Result.new(coverage_result)
+      SimpleCov::ResultMerger.store_result(simplecov_result)
+
       exit! exit_code
     end
   else


penman:
Looks like CI is about 5 minutes faster with this change. 😃

Library/Homebrew/test/.simplecov
```diff
@@ -20,7 +20,13 @@ SimpleCov.start do
     at_exit do
       exit_code = $!.nil? ? 0 : $!.status
       $stdout.reopen("/dev/null")
-      SimpleCov.result # Just save result, but don't write formatted output.
+
+      # Just save result, but don't write formatted output.
+      coverage_result = Coverage.result
+      SimpleCov.add_not_loaded_files(coverage_result)
+      simplecov_result = SimpleCov::Result.new(coverage_result)
```
reitermarkus: Is there a speed difference here between `SimpleCov::Result.new(Coverage.result)` and [`SimpleCov::Result.new(SimpleCov.add_not_loaded_files(Coverage.result))`](https://github.com/colszowka/simplecov/blob/342a67c986949d9b4280b0e267cd9066fe9fcdaa/lib/simplecov.rb#L83).

MikeMcQuaid:
Massive speedup, great work. Is it work updating both the Homebrew and Cask tests `Gemfile.lock` to pull in the `master` branch and your commit in https://github.com/colszowka/simplecov/pull/520 to further speed things up?

penman:
@MikeMcQuaid I tried it out, but now that we're not accidentally loading in 10000 files it only ended up making a 1 second difference. It looks like #1089 does this anyway.

MikeMcQuaid:
@penman Cool. Happy to wait on that one. One conflict to resolve here then happy to 🚢.

Library/Homebrew/test/.simplecov
penman: I don't think so. `SimpleCov.add_not_loaded_files(x)` just returns `x`, so I split it out of one long method chain for readability.

penman:
@MikeMcQuaid great! I'll have that done in a few hours when I'm back at a computer. :)

MikeMcQuaid:
👏 for dramatic speedup!

PR #1085, opened 2016-09-22
docs: more accurate update-test parameter name

penman:

-----

`sha1` isn't a very accurate description of what this value can be because it can be any git commit reference. For example, `brew test-bot` uses `brew update-test --commit=HEAD`.

While I was changing the documentation, it seemed like a good idea to change the variable names too, but I can see reasons for and against doing this, so happy to leave them as is if people would prefer.

### DIFF
--- a/Library/Homebrew/dev-cmd/update-test.rb
+++ b/Library/Homebrew/dev-cmd/update-test.rb
@@ -1,9 +1,9 @@
-#:  * `update-test` [`--commit=<sha1>`] [`--before=<date>`] [`--keep-tmp`]:
+#:  * `update-test` [`--commit=<commit>`] [`--before=<date>`] [`--keep-tmp`]:
 #:    Runs a test of `brew update` with a new repository clone.
 #:
 #:    If no arguments are passed, use `origin/master` as the start commit.
 #:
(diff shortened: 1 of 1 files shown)


zmwangx:
Personally I'd say update the docs and leave the variable names so that git-blame's usefulness it's destroyed. By the way, commitish might be the most accurate word here. Not that you should use it...

penman:
@zmwangx

> Not that you should use it...

😂

I spent some time researching the git terminology and 10 minutes in I still couldn't tell if committish was correct or not so decided it wasn't the best term to use.

MikeMcQuaid:
> Personally I'd say update the docs and leave the variable names so that git-blame's usefulness isn't destroyed.

I think not changing code for `git blame` isn't a good enough reason, personally. 👍 to changing the names; arguably they could be `ref`, though.

penman:
> arguably they could be ref, though.

`ref` was actually what I thought about first for the docs, but decided `commit` had a more obvious meaning. From there, I think the variable names being consistent with the documentation is a good thing.

MikeMcQuaid:
More great work from @penman (who is on 🔥)

PR #1051, opened 2016-09-20
Delete old cctools-based relocation code.

woodruffw:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [ ] Have you successfully run `brew tests` with your changes locally?

-----

Final wrap-up now that #767 has been merged and working smoothly for a few weeks.

cc @UniqMartin, @MikeMcQuaid

Issue #371, opened 2016-06-17
Executables with RPATH are not relocatable


sjackman:
The `sailfish` formula is not `cellar :any` because `cmake` sets the `RPATH` of the executable. I believe `keg_relocate` does not modify the `RPATH`. Could someone please confirm that? The `RPATH` should be given the same relocation treatment as `install_name`.

sjackman:
See https://github.com/Homebrew/homebrew-science/pull/3753#issuecomment-226634919

MikeMcQuaid:
I'm not sure if it does and if it doesn't: we'd welcome a PR to do so.

sjackman:
Should the `RPATH` simply be deleted? I believe it's redundant with the absolute path `install_name` mechanism of finding shared libraries on Mac.

sjackman:
`sailfish` includes an `RPATH`, which makes the bottle non-relocatable. Removing the `RPATH` makes it `cellar :any`.

```
❯❯❯ brew install sailfish
❯❯❯ otool -l /usr/local/bin/sailfish |grep RPATH -A2
          cmd LC_RPATH
      cmdsize 56
         path /usr/local/Cellar/sailfish/0.10.1/lib (offset 12)
❯❯❯ brew bottle -v sailfish
Warning: String '/usr/local/Cellar' still exists in these files:
/usr/local/Cellar/sailfish/0.10.1/bin/sailfish
 --> match '/usr/local/Cellar/sailfish/0.10.1/lib' at offset 0x9e4
❯❯❯ chmod u+w /usr/local/bin/sailfish
❯❯❯ install_name_tool -delete_rpath /usr/local/Cellar/sailfish/0.10.1/lib /usr/local/bin/sailfish
❯❯❯ brew bottle -v sailfish
    cellar :any
```

MikeMcQuaid:
If it can be: yes, that makes sense.

sjackman:
There's a relevant discussion on RPATH in Conda at https://github.com/conda/conda-build/issues/279

sjackman:
After some reading, it looks as though `RPATH` aren't used on Mac OS unless explicitly requested by `@rpath` in the `install_name`. If none of the dependent libraries use `@rpath` in the `install_name`, then `RPATH` is not used and can be safely deleted.
e.g.
`install_name = /usr/lib/libncurses.5.4.dylib`
is equivalent to
`install_name = @rpath/libncurses.5.4.dylib` `RPATH=/usr/lib`

[...]

PR #1086, opened 2016-09-22
Remove old cctools references now that only ruby-macho is used.

woodruffw:

-----

Continuation of #1051. The big changes here are that `Keg#require_install_name_tool?` is now `Keg#require_relocation?` and `OS::Mac.{install_name_tool,otool}` are removed.

Other notes:

We still check for `otool` and `install_name_tool` behavior in *extend/os/mac/diagnostic.rb*, even though both are currently eliminated entirely from the codebase.

Is it worth keeping checks like `check_for_bad_install_name_tool`? A cursory search of the core shows that at least 11 formulae invoke it directly, rather than using our ruby implementation.

cc @UniqMartin, @MikeMcQuaid

### DIFF
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -255,7 +255,7 @@ module Homebrew
           if prefix != prefix_check
             relocatable = false if keg_contain_absolute_symlink_starting_with?(prefix, keg)
           end
-          skip_relocation = relocatable && !keg.require_install_name_tool?
+          skip_relocation = relocatable && !keg.require_relocation?
         end
         puts if !relocatable && ARGV.verbose?
       rescue Interrupt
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
> Is it worth keeping checks like check_for_bad_install_name_tool

In that case it's worth renaming but adding to our `compat/` layer.

MikeMcQuaid:
Thanks again @woodruffw!

woodruffw:
> In that case it's worth renaming but adding to our compat/ layer.

Got it! I'll also look into changing those formula to use ruby-macho instead.

UniqMartin:
Nice work here and in #1051! If you feel like leveraging more of `ruby-macho`'s capabilities in Homebrew, you could look into relocating and cleaning up bogus RPATHs. This could also help with making a few more bottles relocatable (cf. #371), though admittedly RPATHs are still not that prevalent among Homebrew formulae, so the impact won't be massive.

And then there was also the idea of adding some form of validation of the install name in https://github.com/Homebrew/homebrew-core/pull/4655#issuecomment-247753580 which also falls into the realm of Mach-O related tweaks in Homebrew, thus something you might want to look into.

woodruffw:
Thanks!

I'll read those issues and see what I can do about them. Expect PRs in a few days :smile:

PR #1088, opened 2016-09-22
warning only functionality

MatzFan:
Re [my proposal](https://github.com/MatzFan/brew-evolution/blob/invalid_install_option_warnings/XXX-invalid-install-option-warnings.md), have implemented simplest case of warnings of any invalid build option args - no user interaction. Be interested in any feedback :)

### DIFF
--- a/Library/Homebrew/build_options.rb
+++ b/Library/Homebrew/build_options.rb
@@ -101,6 +101,15 @@ class BuildOptions
     @options - @args
   end
 
+  # @private
+  def invalid_options
+    (@options | @args) - @options
+  end
+
+  def invalid_flag_names
+    invalid_options.map(&:flag).sort
+  end
+
   private
 
   def option_defined?(name)
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -272,7 +272,9 @@ module Homebrew
     f.print_tap_action
 
     fi = FormulaInstaller.new(f)
-    fi.options             = f.build.used_options
+    build_options = f.build
+    fi.options             = build_options.used_options
+    fi.invalid_flag_names  = build_options.invalid_flag_names
     fi.ignore_deps         = ARGV.ignore_deps?
     fi.only_deps           = ARGV.only_deps?
     fi.build_bottle        = ARGV.build_bottle?
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -30,7 +30,7 @@ class FormulaInstaller
   end
 
   attr_reader :formula
-  attr_accessor :options, :build_bottle
+  attr_accessor :options, :build_bottle, :invalid_flag_names
   mode_attr_accessor :show_summary_heading, :show_header
   mode_attr_accessor :build_from_source, :force_bottle
   mode_attr_accessor :ignore_deps, :only_deps, :interactive, :git
@@ -50,6 +50,7 @@ class FormulaInstaller
     @quieter = false
     @debug = false
     @options = Options.new
+    @invalid_flag_names = []
 
     @@attempted ||= Set.new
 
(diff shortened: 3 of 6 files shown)


Library/Homebrew/formula_installer.rb
```diff
@@ -212,6 +213,10 @@ def install
       opoo "#{formula.full_name}: #{old_flag} was deprecated; using #{new_flag} instead!"
     end
 
+    invalid_flag_names.each do |flag|
+      opoo "#{formula.full_name}: #{flag} is invalid for this formula and will be ignored!"
```
MikeMcQuaid: What do you think about actually erroring rather than warning on this?

MatzFan: I think ideal would be error & halt execution if no similarly named valid flag can be found. If a similarly named valid flag can be found, this should be suggested as alternative. User would then have option to choose it & continue build, or exit.

MikeMcQuaid: `deprecated_option` already handles the "existing to new flag" migration stuff.

MikeMcQuaid: So this can be fatal.

MatzFan: Sorry yes, I think see what `deprecated_option` does, but my idea here was to pick up typos - like `--with-postgreslq` and suggest `did you mean --with-postgresql?`. If the option given is not valid (or deprecated) and not close enough (e.g. Levenshtein - or whatever Homebrew uses) to a valid option it should be fatal yes - IMO. Could provide a continue & ignore option of course, if we go down UI route.

MikeMcQuaid: I think on typos it's worth just hard failing. We don't use Levenshtein or anything similar currently but would be 👍 on doing so in future.

MatzFan: ok I change to hard fail then - if we are sure users don't rely on invalid options silently being ignored behaviour? Oh, and do you want a `--no-invalid-option-warnings` flag to override?

Library/Homebrew/build_options.rb
```diff
@@ -101,6 +101,15 @@ def unused_options
     @options - @args
   end
 
+  # @private
+  def invalid_options
+    (@options | @args) - @options
```
bfontaine: Isn’t this the same as `@options - @args`, i.e. `unused_options` above?

MatzFan: No, `#unused options` lists valid options for the formula not in ARGV. `#invalid_options` lists ARGV 'options' that are not listed in the formula's options at all. See the tests for an example :)

MatzFan: It is to catch imperfect user scenario - most of the time `#invalid_options` will be empty.

bfontaine: Oh right, I read too fast :smile:

penman: It's the same as `@args - @options`, though, right?

bfontaine: It seems so.

MatzFan: Indeed @penman, will change it

Library/Homebrew/cmd/install.rb
```diff
@@ -272,7 +272,9 @@ def install_formula(f)
     f.print_tap_action
 
     fi = FormulaInstaller.new(f)
-    fi.options             = f.build.used_options
+    build_options = f.build
+    fi.options             = build_options.used_options
+    fi.invalid_flag_names  = build_options.invalid_flag_names
```
MikeMcQuaid: `invalid_options` because `flags` have a particular meaning for us.

Library/Homebrew/formula_installer.rb
MikeMcQuaid: Actually, I think you're right about making it a warning for now and we could make it a hard failure later.

MatzFan:
#1217 now supersedes, closing.

PR #1102, opened 2016-09-23
version: improve alpha and rc detection (elasticsearch)

zmwangx:

-----

Previously we were in a weird situation where `elasticsearch 5.0.0-beta1` can be detected but `5.0.0-alpha5` cannot.

### DIFF
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -340,8 +340,8 @@ class Version
     m = /-((?:\d+\.)*\d+(?:[abc]|rc|RC)\d*)$/.match(stem)
     return m.captures.first unless m.nil?
 
-    # e.g. foobar-4.5.0-beta1, or foobar-4.50-beta
-    m = /-((?:\d+\.)*\d+-beta\d*)$/.match(stem)
+    # e.g. foobar-4.5.0-alpha5, foobar-4.5.0-beta1, or foobar-4.50-beta
+    m = /-((?:\d+\.)*\d+-(?:alpha|beta|rc)\d*)$/.match(stem)
     return m.captures.first unless m.nil?
 
     # e.g. http://ftpmirror.gnu.org/libidn/libidn-1.29-win64.zip
--- a/Library/Homebrew/test/test_versions.rb
+++ b/Library/Homebrew/test/test_versions.rb
@@ -203,6 +203,10 @@ class VersionParsingTests < Homebrew::TestCase
     assert_version_detected "0.7.1", "https://codeload.github.com/gsamokovarov/jump/tar.gz/v0.7.1"
   end
 
+  def test_elasticsearch_alpha_style
+    assert_version_detected "5.0.0-alpha5", "https://download.elastic.co/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/5.0.0-alpha5/elasticsearch-5.0.0-alpha5.tar.gz"
+  end
+
   def test_gloox_beta_style
     assert_version_detected "1.0-beta7", "http://camaya.net/download/gloox-1.0-beta7.tar.bz2"
   end


MikeMcQuaid:
While you're playing with this might want to handle the openssl case @DomT4 mentioned?

zmwangx:
@MikeMcQuaid We seem to be parsing openssl versions just fine? The version comparison issue is handled by `version_scheme`.

MikeMcQuaid:
I believe `1.0.0a` is interpreted as an alpha, incorrectly.

zmwangx:
That's the expected behavior due to

```ruby
  class AlphaToken < CompositeToken
    PATTERN = /a(?:lpha)?[0-9]*/i

    def <=>(other)
      case other
      when AlphaToken
	rev <=> other.rev
      else
        super
      end
    end
  end
```

We might be able to change that pattern to `/a[0-9]+|alpha[0-9]*/`, but I'm not sure if that's going to break any existing formula.

MikeMcQuaid:
> We might be able to change that pattern to /a[0-9]+|alpha[0-9]*/, but I'm not sure if that's going to break any existing formula.

I think in this case given we know it breaks at least one pretty popular formula as-is it's worth making the change. @DomT4, thoughts?

zmwangx:
By the way I noticed that non-capturing groups are used all over the place in `versions.rb`. What's the point?

Library/Homebrew/version.rb
```diff
@@ -340,8 +340,8 @@ def self._parse(spec)
     m = /-((?:\d+\.)*\d+(?:[abc]|rc|RC)\d*)$/.match(stem)
     return m.captures.first unless m.nil?
 
-    # e.g. foobar-4.5.0-beta1, or foobar-4.50-beta
-    m = /-((?:\d+\.)*\d+-beta\d*)$/.match(stem)
+    # e.g. foobar-4.5.0-alpha5, foobar-4.5.0-beta1, or foobar-4.50-beta
+    m = /-((?:\d+\.)*\d+-(alpha|beta|rc)\d*)$/.match(stem)
```
zmwangx: Might want to use a non-capturing group here to be consistent with everything else, but I want to know why they're used in the first place.

MikeMcQuaid:
> What's the point (when we only need .captures.first anyway)?

Not sure. Could be it used to use multiple captures.

zmwangx:
Do we want to keep using non-capturing groups, or remove them where unnecessary en masse?

MikeMcQuaid:
I don't mind, up to you.

zmwangx:
I'll keep using non-capturing group here for consistency, and maybe open a separate PR afterwards to handle the removal.

DomT4:
> I think in this case given we know it breaks at least one pretty popular formula as-is it's worth making the change. @DomT4, thoughts?

I'm not too fussed about going out of our way to handle OpenSSL. They stand out as having a particularly unusual version scheme & it's not necessary a surprise that doesn't mesh well with our overall version logic in places. Everything else seems to be more or less working as intended/

If people want to tweak it & test that against a bunch of places where we use `a` in the version & that `a` is no longer detected as an alpha releases I'd be fine with that, but I don't think the status quo is a massive problem outside of OpenSSL's version weirdness.

zmwangx:
Maybe we should just override version code inside openssl formulae themselves?

MikeMcQuaid:
> I'm not too fussed about going out of our way to handle OpenSSL. They stand out as having a particularly unusual version scheme & it's not necessary a surprise that doesn't mesh well with our overall version logic in places. Everything else seems to be more or less working as intended/

I don't know if I agree; having `a` on the end of a version doesn't seem to imply alpha to me.

zmwangx:
I'm merging this because the openssl discussion isn't directly related to the fairly straightforward change in this PR. Please keep the discussion going if you will, though.

chdiza:
There are some other things that misleadingly put an "a" on the end of non-alpha versions at times.  I think at least one of the `gmp`/`mpfr` duo has done that.  I also vaguely recall `tmux` doing that at least once.  Maybe there should be a hash that holds all the formulas at a given time that are like this, and use it has a kind of whitelist against accusations of alphahood.

PR #1114, opened 2016-09-24
update.sh: do not restore stable (tag) branch

### LINKED ISSUE #1111: Updating to tag + homebrew.devcmdrun affecting update behavior = stuck forever on old tag
Since Homebrew developers, including those who have run a dev cmd, enjoy the special update behavior where the pre-update branch is checked out post-update, unsuspecting non-developer users could run a dev cmd on a stable (tagged) branch and subsequently stuck on that branch forever. Basically,

```sh
# You start out a nobody, on the v1.0.2 branch.
$ git branch
  master
* v1.0.2
# You accidentally open Pandora's box and become a superhero.
$ brew audit git
$ git config --get homebrew.devcmdrun
true
# Your newly gained power unbeknownst to you, you perform a brew update after getting news of 1.0.3.
$ brew update --verbose
Checking if we need to fetch /usr/local/Homebrew...
Checking if we need to fetch /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core...
Updating /usr/local/Homebrew...
Switched to branch 'master'
Current branch master is up to date.
Switched to branch 'v1.0.2'
# What???
# Learned about superpower; give up superpower.
$ git config --unset homebrew.devcmdrun
$ brew update --verbose
Checking if we need to fetch /usr/local/Homebrew...
Checking if we need to fetch /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core...
Updating /usr/local/Homebrew...
Switched to a new branch 'v1.0.3'
Current branch v1.0.3 is up to date.

Already up-to-date.
# Phew.
```

zmwangx:

-----

Restoring stable branch post-update could lead to unsuspecting users
with homebrew.devcmdrun being stuck forever on an old tag.

Fixes #1111. (A much more detailed description of the problem can be found there.)

I'm worried about the damage that's already done. There should be users that are already stuck, and since we no longer update to master, we can't modify their update logic.

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -297,7 +297,8 @@ EOS
 
   if [[ -n "$HOMEBREW_NO_UPDATE_CLEANUP" ]]
   then
-    if [[ "$INITIAL_BRANCH" != "$UPSTREAM_BRANCH" && -n "$INITIAL_BRANCH" ]]
+    if [[ "$INITIAL_BRANCH" != "$UPSTREAM_BRANCH" && -n "$INITIAL_BRANCH" &&
+          ! "$INITIAL_BRANCH" =~ ^v[0-9]+\.[0-9]+\.[0-9]$ ]]
     then
       git checkout "$INITIAL_BRANCH" "${QUIET_ARGS[@]}"
     fi


MikeMcQuaid:
I've been wondering whether this is better than just having everyone always on `master` and `master` reflecting either `origin/master` or the latest tag.

> I'm worried about the damage that's already done. There should be users that are already stuck, and since we no longer update to master, we can't modify their update logic.

I'm not. We set the dev-cmd stuff for people we expect to be working on Homebrew so have some familiarity with git. These are by their nature more technical users.

zmwangx:
> I'm not. We set the dev-cmd stuff for people we expect to be working on Homebrew so have some familiarity with git. These are by their nature more technical users.

The problem is they may not notice until quite some time later, if ever.

zmwangx:
> I've been wondering whether this is better than just having everyone always on master and master reflecting either origin/master or the latest tag.

I think developers should have their initial branch restored either way, unless you disable auto update out right for developers (which actually may be a good idea).

MikeMcQuaid:
> I think developers should have their initial branch restored either way, unless you disable auto update out right for developers (which actually may be a good idea).

I agree, I just mean for the branches we do have people on rather than `v1.0.1` maybe it makes sense to always have `master`.

> The problem is they may not notice until quite some time later, if ever.

Now we have a core/formula separation I'm less concerned about this. They will notice eventually and the workaround is pretty simple.

PR #1117, opened 2016-09-24
test-bot: do not require brew style in core tap

ilovezfs:
The hashrocket changes in particular have siginficantly degraded the
readability, consistency, clarity, and style of formulae, as well as
expanded the opportunities for ambiguity, in exchange for no tangible
benefits to the contributors and maintainers of core.

`brew style` enforcement should be restricted to homebrew/brew alone.

### DIFF
--- a/Library/Homebrew/dev-cmd/test-bot.rb
+++ b/Library/Homebrew/dev-cmd/test-bot.rb
@@ -634,7 +634,6 @@ module Homebrew
           install_passed = steps.last.passed?
         end
       end
-      test "brew", "style", formula_name
       test "brew", "audit", *audit_args
       if install_passed
         if formula.stable? && !ARGV.include?("--fast") && !ARGV.include?("--no-bottle") && !formula.bottle_disabled?
@@ -740,7 +739,6 @@ module Homebrew
           test "brew", "cask-tests", *coverage_args
         end
       elsif @tap
-        test "brew", "style", @tap.name if @tap.name == "homebrew/core"
         test "brew", "readall", "--aliases", @tap.name
       end
     end


zmwangx:
Good luck reverting and resolving a million conflicts 😗

ilovezfs:
@zmwangx You won't have to assist with that if you do not want to.

MikeMcQuaid:
I disagree, we should still enforce them but fix the rules so that homebrew/core is excluded from them (or has different rules). I will happily revert them all back for you as I made the first change.

ilovezfs:
Given the current set of rules, I don't think we should fail CI over not abiding by them. If the facts on the ground change at some later date, then I may come to some other conclusion.

DomT4:
I agree with ILZ that the `depends_on` situation on core now at least is confusing & inconsistent. It takes me a few seconds to get my head around:
```ruby
depends_on ant: :build
depends_on :python
depends_on python3: :optional
depends_on "openssl"
```
Let alone newer contributors. That's an extreme example, but there will be formulae similar to that. I've no objection to a lot of the changes, but increasingly do find myself agreeing (😱) with ILZ on `depends_on` statements.

I don't know if Rubocop can be trained to simply ignore `depends_on` statements or whether this is going to require something as drastic as reverting the check as in this PR or core having its own Rubocop configuration.

reitermarkus:
> or core having its own Rubocop configuration.

I think this would be the simplest and most flexible solution. Also might be worth taking a look at https://github.com/caskroom/rubocop-cask.

MikeMcQuaid:
> The hashrocket changes in particular have siginficantly degraded the
> readability, consistency, clarity, and style of formulae, as well as
> expanded the opportunities for ambiguity, in exchange for no tangible
> benefits to the contributors and maintainers of core.

This was fixed so passing on this PR. Considering we ask people repeatedly to change style in Homebrew formula PRs it's pretty silly to not rely on tooling to do that for us. We can selectively enable/disable rules for the taps and create our own rules so should aim to do that in future.

PR #1129, opened 2016-09-24
download_strategy: never verbosely untar

zmwangx:

-----

I was responsible for adding the `v` flag to `tar` in d7b6230, and I have to admit, I regretted the decision pretty soon. The verbose mode of `tar` is just too pointlessly verbose. Today I'm finally fed up after having to scroll through [this](https://bot.brew.sh/job/Homebrew%20Core%20Pull%20Requests/8211/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/install_v8/). Let's get rid of it.


zmwangx:
The verbose mode is especially crazy when you deal with something like boost, where there are 12k+ files in the bottle and you get 800KB+ worth of logs in a matter of seconds.

MikeMcQuaid:
I agree. It's also worth noting `tar xvf` can actually slow down the extraction as it blocks on writing to the terminal.

PR #1130, opened 2016-09-25
Prefer brewed Clang over gcc-6 for needs :openmp

aw1621107:
- [?] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?

---

Clang has fully implemented OpenMP support as of LLVM 3.7, so if OpenMP
is required by a formula gcc is no longer the only choice of compiler.

Clang should be preferred over gcc because using gcc meant linking
against libstdc++, which is ABI incompatible with libc++. This may be
unnoticeable for some users, but it causes other builds to fail, e.g.
pstoedit when imagemagick was built with OpenMP. pstoedit is required
for the octave formula, so for some users this could be a significant
problem.

Unfortunately, these changes appear to have broken the *.cc files used
by brew to log changes to the command emitted by Make/etc -- those files
only show changes for calls to ld.

In addition, the changes made to determine_cc are rather hacky, and
should be replaced with a cleaner solution.

---

This is a work in progress. Hopefully will have more commits coming in in the next few days.


Library/Homebrew/compilers.rb
```diff
@@ -6,6 +6,7 @@ module CompilerConstants
     "gcc-4.0"  => :gcc_4_0,
     "gcc-4.2"  => :gcc,
     "clang"    => :clang,
+    "brew_clang" => :brew_clang,
```
MikeMcQuaid: Not a fan of this name but not sure I can think of a better one. I'm leaning towards saying this should just be part of `clang` above.

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -171,7 +171,11 @@ def compiler
 
   # @private
   def determine_cc
-    COMPILER_SYMBOL_MAP.invert.fetch(compiler, compiler)
+    if compiler == :brew_clang
+      "/usr/local/opt/llvm/bin/clang"
```
MikeMcQuaid: Use `HOMEBREW_PREFIX`; can't assume `/usr/local`. Regardless: this definitely shouldn't be hardcoded here.

MikeMcQuaid:
Just to let you know: I'm broadly in favour of this change but it'll take a while to figure out the best approach.

Library/Homebrew/extend/ENV/shared.rb
aw1621107: Yeah. I agree that the approach is lacking.

Is `HOMEBREW_PREFIX` preferred over `brew --prefix`?

MikeMcQuaid: Yep.

Library/Homebrew/compilers.rb
aw1621107: Hmmm... That would require quite a few changes to `CompilerFailure`, wouldn't it, since it appears to assume that compiler capabilities can be determined by name and version alone?

MikeMcQuaid: I'm open to debate with other maintainers on that one.

Library/Homebrew/extend/ENV/shared.rb
aw1621107: Is there an environment variable version of `brew --prefix llvm`? Or should I just use the prefix + `opt/llvm`?

MikeMcQuaid: `Formulary.factory("llvm").opt_prefix`

aw1621107:
Alright, changed the strategy up a bit. Hope it's better.

`brew install imagemagick --with-openmp` actually gets past the configuration stage now; the previous issue with a missing `ltdl.h` appears to have gone away on its own. The build still fails, however, with linker issues.

Gist [here](https://gist.github.com/aw1621107/dd55842605309566016c0e6d77f4ab9f)

For example, this part of the build log looks a bit suspicious:

```
Undefined symbols for architecture x86_64:
  "std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::copy(char*, unsigned long, unsigned long) const", referenced from:
```

Something to do with the standard library not being found? Or the wrong one being used? Don't remember what libc++ uses as its namespace vs libstdc++...

aw1621107:
And `imagemagick --with-openmp` now builds. No idea what I changed to fix things.

MikeMcQuaid:
CC @mistydemeo any naming thoughts here?

aw1621107:
And `pstoedit` builds, so at least one problem is fixed

Library/Homebrew/shims/super/cc
```diff
@@ -154,7 +156,7 @@ class Cmd
       /^-O[0-9zs]?$/, "-fast", "-no-cpp-precomp",
       "-pedantic", "-pedantic-errors", "-Wno-long-double",
       "-Wno-unused-but-set-variable"
-    when "-fopenmp", "-lgomp", "-mno-fused-madd", "-fforce-addr", "-fno-defer-pop",
```
MikeMcQuaid: These should still be filtered for non-`brew_clang`

aw1621107: Fixed!

aw1621107:
Something to check out for later: `brew install fftw --with-fortran --with-mpi --with-openmp` fails:

```
configure:13302: checking for mpicc
configure:13329: result: /usr/local/bin/mpicc
configure:13347: checking for MPI_Init
configure:13347: /usr/local/bin/mpicc -o conftest -g -O2   conftest.c  >&5

Failed to execute brew_clang -pipe -Os -o conftest -g -O2 conftest.c -I/usr/local/Cellar/open-mpi/2.0.1/include -L/usr/local/opt/libevent/lib -L/usr/local/Cellar/open-mpi/2.0.1/lib -lmpi -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -L/usr/local/lib -L/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries -Wl,-headerpad_max_install_names

Xcode and/or the CLT appear to be misconfigured. Try one or both of the following:
  xcodebuild -license
  sudo xcode-select -switch /path/to/Xcode.app

configure:13347: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "fftw"
| #define PACKAGE_TARNAME "fftw"
| #define PACKAGE_VERSION "3.3.5"
| #define PACKAGE_STRING "fftw 3.3.5"
| #define PACKAGE_BUGREPORT "fftw@fftw.org"
| #define PACKAGE_URL ""
| #define PACKAGE "fftw"
| #define VERSION "3.3.5"
| #define FFTW_ENABLE_ALLOCA 1
| #define FFTW_SINGLE 1
| #define BENCHFFT_SINGLE 1
| #define HAVE_SSE2 1
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define HAVE_DLFCN_H 1
| #define LT_OBJDIR ".libs/"
| /* end confdefs.h.  */
| /* Define MPI_Init to an innocuous variant, in case <limits.h> declares MPI_Init.
|    For example, HP-UX 11i <limits.h> declares gettimeofday.  */
| #define MPI_Init innocuous_MPI_Init
| 
| /* System header to define __stub macros and hopefully few prototypes,
|     which can conflict with char MPI_Init (); below.
|     Prefer <limits.h> to <assert.h> if __STDC__ is defined, since
|     <limits.h> exists even on freestanding compilers.  */
| 
| #ifdef __STDC__
| # include <limits.h>
| #else
| # include <assert.h>
| #endif
| 
| #undef MPI_Init
| 
| /* Override any GCC internal prototype to avoid an error.
|    Use char because int might match the return type of a GCC
|    builtin and then its argument prototype would still apply.  */
| #ifdef __cplusplus
| extern "C"
| #endif
| char MPI_Init ();
| /* The GNU C library defines this for functions which it implements
|     to always fail with ENOSYS.  Some functions are actually named
|     something starting with __ and the normal name is an alias.  */
| #if defined __stub_MPI_Init || defined __stub___MPI_Init
| choke me
| #endif
| 
| int
| main ()
| {
| return MPI_Init ();
|   ;
|   return 0;
| }
configure:13347: result: no
configure:13354: checking for MPI_Init in -lmpi
configure:13379: /usr/local/bin/mpicc -o conftest -g -O2   conftest.c -lmpi   >&5

Failed to execute brew_clang -pipe -Os -o conftest -g -O2 conftest.c -lmpi -I/usr/local/Cellar/open-mpi/2.0.1/include -L/usr/local/opt/libevent/lib -L/usr/local/Cellar/open-mpi/2.0.1/lib -lmpi -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -L/usr/local/lib -L/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries -Wl,-headerpad_max_install_names

Xcode and/or the CLT appear to be misconfigured. Try one or both of the following:
  xcodebuild -license
  sudo xcode-select -switch /path/to/Xcode.app

configure:13379: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "fftw"
| #define PACKAGE_TARNAME "fftw"
| #define PACKAGE_VERSION "3.3.5"
| #define PACKAGE_STRING "fftw 3.3.5"
| #define PACKAGE_BUGREPORT "fftw@fftw.org"
| #define PACKAGE_URL ""
| #define PACKAGE "fftw"
| #define VERSION "3.3.5"
| #define FFTW_ENABLE_ALLOCA 1
| #define FFTW_SINGLE 1
| #define BENCHFFT_SINGLE 1
| #define HAVE_SSE2 1
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define HAVE_DLFCN_H 1
| #define LT_OBJDIR ".libs/"
| /* end confdefs.h.  */
| 
| /* Override any GCC internal prototype to avoid an error.
|    Use char because int might match the return type of a GCC
|    builtin and then its argument prototype would still apply.  */
| #ifdef __cplusplus
| extern "C"
| #endif
| char MPI_Init ();
| int
| main ()
| {
| return MPI_Init ();
|   ;
|   return 0;
| }
configure:13388: result: no
configure:13396: checking for MPI_Init in -lmpich
configure:13421: /usr/local/bin/mpicc -o conftest -g -O2   conftest.c -lmpich   >&5

Failed to execute brew_clang -pipe -Os -o conftest -g -O2 conftest.c -lmpich -I/usr/local/Cellar/open-mpi/2.0.1/include -L/usr/local/opt/libevent/lib -L/usr/local/Cellar/open-mpi/2.0.1/lib -lmpi -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -L/usr/local/lib -L/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries -Wl,-headerpad_max_install_names

Xcode and/or the CLT appear to be misconfigured. Try one or both of the following:
  xcodebuild -license
  sudo xcode-select -switch /path/to/Xcode.app

configure:13421: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "fftw"
| #define PACKAGE_TARNAME "fftw"
| #define PACKAGE_VERSION "3.3.5"
| #define PACKAGE_STRING "fftw 3.3.5"
| #define PACKAGE_BUGREPORT "fftw@fftw.org"
| #define PACKAGE_URL ""
| #define PACKAGE "fftw"
| #define VERSION "3.3.5"
| #define FFTW_ENABLE_ALLOCA 1
| #define FFTW_SINGLE 1
| #define BENCHFFT_SINGLE 1
| #define HAVE_SSE2 1
| #define STDC_HEADERS 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_MEMORY_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_UNISTD_H 1
| #define HAVE_DLFCN_H 1
| #define LT_OBJDIR ".libs/"
| /* end confdefs.h.  */
| 
| /* Override any GCC internal prototype to avoid an error.
|    Use char because int might match the return type of a GCC
|    builtin and then its argument prototype would still apply.  */
| #ifdef __cplusplus
| extern "C"
| #endif
| char MPI_Init ();
| int
| main ()
| {
| return MPI_Init ();
|   ;
|   return 0;
| }
configure:13430: result: no
configure:13469: error: could not find mpi library for --enable-mpi
```

aw1621107:
`openblas --with-openmp` also fails, although this time it is OpenBLAS' own Makefile that is failing the build:

```
make
CC=brew_clang
FC=/usr/local/bin/gfortran
libs
netlib
shared

Makefile.system:409: *** OpenBLAS: Clang didn't support OpenMP yet..  Stop.
```

I wonder if mixing `clang` and `gfortran` would cause issues, too...

Library/Homebrew/compilers.rb
MikeMcQuaid: `clang-llvm`? @ilovezfs @zmwangx any thoughts?

ilovezfs: Doesn't `fails_with` refer to the formula just as `llvm`?

MikeMcQuaid: `llvm` is/was `llvm-gcc`. We'll need a new name as a result, unfortunately.

ilovezfs: How about `:clang_3_8` since
```
lrwxr-xr-x  1 joe  admin         9 Jun  3 20:43 clang -> clang-3.8
lrwxr-xr-x  1 joe  admin         5 Jun  3 20:43 clang++ -> clang
-r-xr-xr-x  1 joe  admin  44760792 Jun  3 20:43 clang-3.8
```

MikeMcQuaid: That'll change every release and I think we don't actually need to bump it each version like we do with GCC. I think ideally we'd figure out a way eventually to have this just be `:clang` and `:gcc` and specify the build/version inside the DSL if necessary.

MikeMcQuaid: I think it's fine; the version numbers only matter for `fails_with` invocations.

aw1621107: Done!

Library/Homebrew/shims/super/brew_clang++
```diff
@@ -0,0 +1 @@
+cc
```
MikeMcQuaid: Should be called `llvm_clang` I think?

Library/Homebrew/shims/super/cc
```diff
@@ -58,6 +58,8 @@ class Cmd
     @tool ||= case @arg0
     when "ld" then "ld"
     when "cpp" then "cpp"
+    when /brew_(clang(\+\+)?)/
```
MikeMcQuaid: Should be called `llvm_clang` I think?

aw1621107:
Fixed, I think.

Library/Homebrew/compilers.rb
```diff
@@ -6,6 +6,7 @@ module CompilerConstants
     "gcc-4.0"  => :gcc_4_0,
     "gcc-4.2"  => :gcc,
     "clang"    => :clang,
+    "llvm_clang" => :brew_clang,
```
MikeMcQuaid: Symbol should probably match the string here.

Library/Homebrew/development_tools.rb
```diff
@@ -74,6 +74,14 @@ def clang_build_version
         end
     end
 
+    def llvm_clang_build_version
+      @llvm_clang_build_version ||=
+        if Tab.for_name "llvm"
+          path = Formulary.factory("llvm").opt_prefix.join("bin", "clang")
```
MikeMcQuaid: `path = Formulary.factory("llvm").opt_prefix/"bin/clang"`

Library/Homebrew/development_tools.rb
```diff
@@ -74,6 +74,14 @@ def clang_build_version
...
+          `#{path} --version`.split(" ")[2]
```
MikeMcQuaid: May want to handle what happens if the version format is invalid here.

MikeMcQuaid: i.e. there's no `[2]` in the array as it's blank or doesn't contain spaces.

Library/Homebrew/development_tools.rb
```diff
@@ -74,6 +74,14 @@ def clang_build_version
         end
     end
 
+    def llvm_clang_build_version
+      @llvm_clang_build_version ||=
+        if Tab.for_name "llvm"
```
MikeMcQuaid: Put this line on the preceding line

MikeMcQuaid:
@aw1621107 A few more comments. After that can you let me know what formulae you've tested this with? Thanks!

ilovezfs:
I think we may want homebrew science's opinions here. Since Linux builds with gcc not clang, switching the openmp default compiler may cause problems for their CI due to the divergence, where currently there is unity.

MikeMcQuaid:
> switching the openmp default compiler may cause problems for their CI due to the divergence, where currently there is unity.

Good point. Let's make this PR prioritise `llvm_clang` last for now to ease getting support in and merged sooner rather than later and then we can have an immediate follow-up PR which changes the default priority.

Library/Homebrew/compilers.rb
aw1621107: I'm somewhat surprised that sed didn't catch that, especially since the symbol in `COMPILER_PRIORITY` got changed. In any case, fixed!

Library/Homebrew/development_tools.rb
aw1621107: Fixed!

Library/Homebrew/development_tools.rb
MikeMcQuaid: `@llvm_clang_build_version ||= if Tab.for_name "llvm"`

Library/Homebrew/compilers.rb
```diff
@@ -6,6 +6,7 @@ module CompilerConstants
     "gcc-4.0"  => :gcc_4_0,
     "gcc-4.2"  => :gcc,
     "clang"    => :clang,
+    "llvm_clang" => :llvm_clang,
```
aw1621107: Done!

Library/Homebrew/development_tools.rb
aw1621107: Ah, ok. Fixed!

Library/Homebrew/development_tools.rb
aw1621107: I just changed it to use a regex similar to the one `clang_version` uses. Would that work?

MikeMcQuaid: @aw1621107 Sure, probably 😁

MikeMcQuaid:
This looks good to me now. Will want to do some local testing with a few formulae and then hopefully we can get this shipped.

aw1621107:
Alright, made llvm_clang lowest priority.

The formulae with OpenMP that I have installed using this patch are:
  - `fftw`
  - `imagemagick`
  - `llvm` (not sure if this counts?)
  - `metis`
  - `openblas`
  - `pngquant`
  - `suitesparse`
  - `superlu`

I don't use many of these directly, so I don't think I can really say anything besides that they compiled.

MikeMcQuaid:
`brew style` failing: https://bot.brew.sh/job/Brew/2945/version=any_sierra/testReport/junit/brew-test-bot/sierra/style/. After that's fixed I'll test locally and if it looks good: merge.

aw1621107:
Alright, pushed the style fixes and cleaned up the commit message a bit.

Just finished building the newest imagemagick locally, and it finishes compiling. However, the logs say that imagemagick was not built with OpenMP despite the presence of `with-openmp`. Seems the linker cannot find the OpenMP library. Pretty sure this didn't happen before, but it's something I need to fix regardless.

MikeMcQuaid:
Ok, cool, will wait for that.

aw1621107:
I think the issue is that `gcc` is linked by default, while `llvm` isn't. Thus, the linker can find libgomp just fine, but cannot find libomp. In addition, the imagemagick build actually uses OpenMP if `brew link --force llvm` is run.

Seems like this could be an issue. Linking formulae that use OpenMP against LLVM might not play well with all the other formulae that use Apple's LLVM, and trying to link just libomp with Apple LLVM doesn't sound like it'd work too well. Thoughts?

MikeMcQuaid:
You'll probably need to make sure the relevant paths are added in superenv so they can link against `libomp`. 

> Linking formulae that use OpenMP against LLVM might not play well with all the other formulae that use Apple's LLVM

I think we'll need to accept or fix this or this whole concept is a non-starter, unfortunately.

aw1621107:
>You'll probably need to make sure the relevant paths are added in superenv so they can link against libomp.

If I'm understanding this correctly, superenv already has a mechanism to add library search paths for keg only dependencies, right? If so, would it be possible to make `needs :openmp` also add a dependency on LLVM? Or will `depends_on "llvm"` have to be added manually to the relevant formulae? Or is that not a good path to take?

In terms of LLVM libs that may conflict with Apple's LLVM libs, seems like that main culprits would be libc++, libLTO, and libunwind. Looks to me that the other ones are more compiler-specific libraries that most formulae wouldn't use. I'd hope that those are reasonably API/ABI stable across versions, so with any luck formulae shouldn't have too much of an issue...

aw1621107:
Also, going through the OpenMP-using formulae I have installed:

  - `fftw`: Something weird broke and configure can't find `MPI_Init` for OpenMPI. Compiling with gcc *also* fails, saying "OpenMP enabled but not using an OpenMP compiler".
  - `imagemagick`: compiles fine
  - `metis`: compiles fine
  - `openblas`: Needs editing to Makefile.system to remove check for clang, and build fails in line 1849 kernel/x86_64/dgemm_kernel_4x8_sandy.S with "error: unexpected token at start of statement". The "unexpected token" is a series of `#`s used as comments, so not really sure what's going on. Build with gcc *also* fails (with linker errors), so looks like there's something up with my installation...
  - `pngquant`: compiles fine
  - `suite-sparse`: compiles fine
  - `superlu`: compiles fine

So I guess the change didn't break anything that wasn't already broken.

MikeMcQuaid:
@BrewTestBot test this please

MikeMcQuaid:

ilovezfs:
This PR broke homebrew/science/trinity with or without llvm installed.

```
g++-6 -DHAVE_CONFIG_H -I. -I..    -ansi -pedantic -fopenmp -Wall -Wextra -Wno-deprecated -m64 -I/Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home/include -I/Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home/include/darwin -fopenmp -MT IRKE.o -MD -MP -MF .deps/IRKE.Tpo -c -o IRKE.o IRKE.cpp
IRKE.cpp: In member function 'void IRKE::populate_Kmers_from_kmers(const string&)':
IRKE.cpp:89:62: error: 'omp_set_num_threads' was not declared in this scope
     omp_set_num_threads(IRKE_COMMON::MAX_THREADS_READ_PARSING); // place limit here because performance deteriorates beyond this value.
                                                              ^
IRKE.cpp: In member function 'void IRKE::compute_sequence_assemblies(KmerCounter&, float, unsigned int, unsigned int, bool, bool, std::__cxx11::string)':
IRKE.cpp:461:53: error: 'omp_set_num_threads' was not declared in this scope
         omp_set_num_threads(IRKE_COMMON::NUM_THREADS);
                                                     ^
IRKE.cpp:464:30: error: 'omp_set_num_threads' was not declared in this scope
         omp_set_num_threads(1); // turn off multithreading for the contig building.
                              ^
make[2]: *** [IRKE.o] Error 1
make[1]: *** [install-recursive] Error 1
make: *** [inchworm_target] Error 2
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
```

ilovezfs:
It also doesn't work as designed. Even if I remove the dependency on gcc and explicitly add a dependency on llvm, g++-6 is still used.

MikeMcQuaid:
> It also doesn't work as designed. Even if I remove the dependency on gcc and explicitly add a dependency on llvm, g++-6 is still used.

That is intentional. It's the lowest priority so should require being manually specified with `--cc` or `fails_with` every other compiler for now.

ilovezfs:
Ah OK. Then the bug is straightforward: it filters out fopenmp for anything that isn't llvm-clang now, including gcc.

Oh, and If I uninstall gcc, I get
```
Error: undefined method `null?' for "3.9.0":String
Please report this bug:
/usr/local/Homebrew/Library/Homebrew/compilers.rb:132:in `block in find_compiler'
/usr/local/Homebrew/Library/Homebrew/compilers.rb:120:in `each'
/usr/local/Homebrew/Library/Homebrew/compilers.rb:120:in `find_compiler'
/usr/local/Homebrew/Library/Homebrew/compilers.rb:113:in `compiler'
/usr/local/Homebrew/Library/Homebrew/compilers.rb:96:in `select_for'
/usr/local/Homebrew/Library/Homebrew/extend/ENV/shared.rb:167:in `compiler'
/usr/local/Homebrew/Library/Homebrew/extend/ENV/super.rb:42:in `setup_build_environment'
/usr/local/Homebrew/Library/Homebrew/extend/os/mac/extend/ENV/super.rb:89:in `setup_build_environment'
/usr/local/Homebrew/Library/Homebrew/build.rb:85:in `install'
/usr/local/Homebrew/Library/Homebrew/build.rb:189:in `<main>'
```

ilovezfs:
With
1) https://gist.githubusercontent.com/ilovezfs/21a8649d5ba53a4e9dba1868989f91e0/raw/c6f3bbb5b33d97c5336d583dad8e6fd4e76b55b0/gistfile1.txt
2) gcc not installed
3) llvm installed
4) https://github.com/Homebrew/brew/pull/1546 pulled locally

I get

```
checking whether make sets $(MAKE)... yes
checking for C++ compiler default output file name... 
configure: error: in `/private/tmp/trinity-20161121-33709-5fj2js/trinityrnaseq-Trinity-v2.3.2/Inchworm':
configure: error: C++ compiler cannot create executables
See `config.log' for more details.
make: *** [inchworm_target] Error 77
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
BuildError: Failed executing: make all
```
if I run
```
brew install -dvs trinity
```

ilovezfs:
@aw1621107 only way I've found to fix the above is to an explicit `depends_on "llvm"` to the trinity formula.

aw1621107:
>Then the bug is straightforward: it filters out fopenmp for anything that isn't llvm-clang now, including gcc.

Oops. That was quite the oversight.

@ilovezfs The error is due to the linker being unable to find OpenMP, because LLVM isn't linked by default. `brew link --force llvm` would also fix that.

I would guess that means that `needs :openmp` needs to also implicitly set `depends_on "llvm"`, since linking LLVM isn't a great option, but not sure what else can be done. Thoughts?

MikeMcQuaid:
@aw1621107 I think the best solution would be for OpenMP's library to be added by superenv if `needs :openmp` is used.

aw1621107:
@MikeMcQuaid Sounds good to me! Only thing is that LLVM's libomp is in the same directory as everything else... Would creating a temp directory with only a symlink to libomp work?

MikeMcQuaid:
Yes, something like that sounds sensible assuming you want to avoid e.g. libc++.

PR #1135, opened 2016-09-25
Fix `search_tap` regex/string matching.

reitermarkus:

-----

Fixes the error introduced in https://github.com/Homebrew/brew/commit/52ff98853068c03b3bfa777932da1da69e35e583#diff-791b01b40fb03181802be9c2315258b2R153.

Not sure if the test I have added is the best way to go, but I didn't find another command which even get's to this stage in `search_tap`. Couldn't even get it to work locally yet as I am getting a GitHub API error, despite having set `HOMEBREW_GITHUB_API_TOKEN`.

### DIFF
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -40,8 +40,8 @@ module Homebrew
       exec_browser "http://packages.ubuntu.com/search?keywords=#{ARGV.next}&searchon=names&suite=all&section=all"
     elsif ARGV.include? "--desc"
       query = ARGV.next
-      rx = query_regexp(query)
-      Descriptions.search(rx, :desc).print
+      regex = query_regexp(query)
+      Descriptions.search(regex, :desc).print
     elsif ARGV.first =~ HOMEBREW_TAP_FORMULA_REGEX
       query = ARGV.first
       user, repo, name = query.split("/", 3)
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/search.rb
```diff
@@ -112,15 +112,17 @@ def query_regexp(query)
...
-      Thread.new { search_tap(user, repo, rx) }
+      Thread.new { search_tap(user, repo, regex_or_string) }
     end.inject([]) do |results, t|
       results.concat(t.value)
     end
   end
 
-  def search_tap(user, repo, rx)
+  def search_tap(user, repo, regex_or_string)
+    regex = regex_or_string.is_a?(String) ? /^#{regex_or_string}$/ : regex_or_string
```
MikeMcQuaid: May need to use `#{Regexp.escape(regex_or_string)}` here.

Library/Homebrew/test/test_integration_cmds.rb
```diff
@@ -707,6 +707,10 @@ def test_search
     assert_predicate desc_cache, :exist?, "Cached file should exist"
   end
 
+  def test_search_remote_tap
+    assert_match "Warning: Failed to search tap: homebrew/not-a-tap.", cmd("search", "homebrew/not-a-tap/not-a-formula")
```
MikeMcQuaid: Probably worth adding a unit rather than integration test for this, if that's 🆒

MikeMcQuaid:
Test failing for this: https://travis-ci.org/Homebrew/brew/builds/162568024

reitermarkus:
Added a unit test instead, which I'm letting fail first before pushing the fix again.

Library/Homebrew/test/test_integration_cmds.rb
reitermarkus: Added a unit test instead.

MikeMcQuaid:
Thanks @reitermarkus!

PR #1139, opened 2016-09-25
ARGV: kegs: error on empty arguments

### LINKED ISSUE #1138: Bug in `brew remove`?
This isn't a case of normal usage, but:  `brew remove ""` results in:

```
/usr/local/Cellar/autoconf is not a valid keg
```

There's nothing special about `autoconf`; testing reveals that whatever formula is the alphabetically earliest that's installed will be the one that appears in the message.

This kind of thing is probably supposed to be blocked by line 13 of `uninstall.rb`, but the KegUnspecifiedError isn't getting raised AFAICT.  I suspect that the `named` method insufficiently strict to use as the guard here.

zmwangx:

-----

When an argument is empty, `Formulary.to_rack(name.downcase)` returns `HOMEBREW_CELLAR`, then the list of kegs `dirs` become the list of racks, which is wrong, and leads to the bewildering behavior in #1138.

We need to error out on an empty argument, but whether `NoSuchKegError` is the most appropriate error here is debatable.

Fixes #1138.

**Update.** Switched to `UsageError`.

### DIFF
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -65,6 +65,8 @@ module HomebrewArgvExtension
     require "keg"
     require "formula"
     @kegs ||= downcased_unique_named.collect do |name|
+      raise UsageError if name.empty?
+
       rack = Formulary.to_rack(name.downcase)
 
       dirs = rack.directory? ? rack.subdirs : []


zmwangx:
Alternatively this can be fixed in `Formulary.to_rack`. I wonder what's the appropriate error there; `FormulaUnavailableError`? Anyway, as it currently stands, `Formulary.to_rack` does very little for validation, so not sure we should enforce non-emptiness there.

reitermarkus:
Seems like `ArgumentError` would be a better choice here. Also, `NoSuchKegError` doesn't seem to be `rescue`d anywhere, that would have been something to consider.

zmwangx:
> `ArgumentError`

Sure.

zmwangx:
I ended up changing it to a `UsageError` which seems most appropriate when an arg that shouldn't be empty is empty (and I can't think of a case where an empty argument is valid for a brew command).

`ArgumentError` isn't handled in `brew.rb` except as a generic exception that leads to a "Please report this bug:", which isn't right (although we can certainly change that).

PR #1145, opened 2016-09-25
Use `module_function` for `Homebrew` module.

reitermarkus:

-----

Use `module_function` rather than `extend self` to convert instance methods to class methods.


MikeMcQuaid:
Can you explain a bit why https://github.com/Homebrew/homebrew-test-bot/pull/2 is needed?

reitermarkus:
This is needed because without `extend self`, the methods inside `module Homebrew` blocks will not be converted to class methods globally, but rather on a per-block basis.

MikeMcQuaid:
@reitermarkus I see. I think we need to avoid breaking the previous API because we support it for third-party commands and can't update them globally.

reitermarkus:
We could put 

```ruby
module Homebrew
  extend self
end
```

in `compat/global.rb`.

MikeMcQuaid:
@reitermarkus That works for me. Is there any way to detect usage of the old functions and print a warning on them somehow? We don't need to do that eventually but trying to figure out how we'd deprecate that eventually.

reitermarkus:
Took me quite a while to figure this out (because of course I had a typo in `module Hombrew` while testing), but I found a way.

penman:
fwiw `module_function` will (IIRC) prevent you having any methods that are `private` or `protected` to the module, whereas this works with `extend self`.

reitermarkus:
@penman, `private_class_method`: https://github.com/Homebrew/brew/pull/1145/files#diff-3003aac9196017d495a416e89622da87R58

For `protected` you'll have to use `class << self` though.

penman:
TIL @reitermarkus :D

MikeMcQuaid:
A style failure here but it looks like the bot is behaving at least: https://bot.brew.sh/job/Brew%20Pull%20Requests/2103/version=any_sierra/testReport/junit/brew-test-bot/sierra/style/

reitermarkus:
The style failure seems to be a bug in RuboCop. Will open an issue.

reitermarkus:
If someone wants to test this, uncomment the `odeprecated` line and try running the `test-bot`, while [the PR over there](https://github.com/Homebrew/homebrew-test-bot/issues/2) isn't merged.

Library/Homebrew/cmd/link.rb
```diff
@@ -71,11 +73,10 @@ def link
...
   end
 
-  private
-
   def keg_only?(rack)
     Formulary.from_rack(rack).keg_only?
   rescue FormulaUnavailableError, TapFormulaAmbiguityError, TapFormulaWithOldnameAmbiguityError
     false
   end
+  private_class_method :keg_only?
```
MikeMcQuaid: I think for the `cmd`s we can remove the `private_class_method`s as we don't really care if they are private or not.

Library/Homebrew/compat/global.rb
```diff
@@ -0,0 +1,20 @@
...
+    if instance_methods.include?(method)
+      # odeprecated "#{self}##{method}", "'module_function' or 'def self.#{method}' to convert it to a class method"
+      instance_method(method).bind(self).call(*args, &block)
+    else
+      super
+    end
+  end
+
+  def respond_to_missing?(method, include_private = false)
+    if method_defined?(method)
```
MikeMcQuaid:
```
return true if method_defined?(method)
super(method, include_private)
```

Library/Homebrew/compat/global.rb
```diff
@@ -0,0 +1,20 @@
+module Homebrew
+  module_function
+
+  def method_missing(method, *args, &block)
+    if instance_methods.include?(method)
+      # odeprecated "#{self}##{method}", "'module_function' or 'def self.#{method}' to convert it to a class method"
+      instance_method(method).bind(self).call(*args, &block)
```
MikeMcQuaid: Maybe `return` this and skip the `else`?

Library/Homebrew/global.rb
```diff
@@ -39,14 +39,25 @@
 require "tap_constants"
 
 module Homebrew
-  include FileUtils
-  extend self
+  module_function
 
-  attr_accessor :failed
```
MikeMcQuaid: There's no equivalent?

reitermarkus: I didn't check, but why would you keep both?

`failed == failed?`

MikeMcQuaid: Sorry, not this specifically but just being able to use `attr_accessor`

reitermarkus:
```
class << self
  attr_accessor :something
end
```

MikeMcQuaid: Reasons to not do that here?

reitermarkus: No, just didn't check before. :+1:

MikeMcQuaid:
I'm 🆒 with 🚢 ing this when you are @reitermarkus when you're confident in it.

PR #1146, opened 2016-09-26
boneyard-formula-pr: respect HOMEBREW_NO_AUTO_UPDATE

zmwangx:

-----

Very simple: make `boneyard-formula-pr` respect `HOMEBREW_NO_AUTO_UPDATE`. (People who are at the pay grade of running this command typically should be on top of updates.)

Apparently `bump-formula-pr` should receive the same treatment, but I don't use it, and ILZ will get to it (on an infinite time scale®).

### DIFF
--- a/Library/Homebrew/dev-cmd/boneyard-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/boneyard-formula-pr.rb
@@ -30,7 +30,7 @@ module Homebrew
     boneyard_tap = Tap.fetch("homebrew", "boneyard")
     tap_migrations_path = formula.tap.path/"tap_migrations.json"
     if ARGV.dry_run?
-      puts "brew update"
+      puts "brew update" if ENV["HOMEBREW_NO_AUTO_UPDATE"].nil?
       puts "brew tap #{boneyard_tap.name}"
       puts "cd #{formula.tap.path}"
       cd formula.tap.path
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I don't agree with this because the no-op case is so fast and because it's particularly important to avoid merge conflicts. It may make sense to do a "proper" auto-update (https://github.com/Homebrew/brew/blob/6e5735565b9a31ec2b01e02a479a361189729c22/Library/Homebrew/brew.sh#L277) before this, though? I also think bump and boneyard should have the same behaviour.

zmwangx:
@MikeMcQuaid 

> the no-op case is so fast

It's not exactly fast when you have more than a dozen taps tapped. Takes a few seconds for me even in the no-op case.

> it's particularly important to avoid merge conflicts.

I disagree. If you are boneyarding a formula that somehow changed in core since your last upgrade, chances are it shouldn't be boneyarded. Moreover, it's trivial to fix a merge conflict in this case, and boneyard PRs don't hold up the CI however many times you push.

> It may make sense to do a "proper" auto-update 

I don't think it deserves special treatment in `brew.sh`.

> I also think bump and boneyard should have the same behaviour.

It's actually somewhat justifiable to say "it's particularly important to avoid merge conflicts" in the case of bump, but it's an entirely different story for boneyard. They are not exactly comparably (despite the similarity in their names) the more I think about it.

MikeMcQuaid:
> It's not exactly fast when you have more than a dozen taps tapped. Takes a few seconds for me even in the no-op case.

I'd rather see us lean on `HOMEBREW_AUTO_UPDATE_SECS` rather than removing `brew update` calls or overloading `HOMEBREW_NO_AUTO_UPDATE` to mean more than `brew update --preinstall` does. For this use-case it makes sense to make it another auto-update command and allow it to be disabled that way.

MikeMcQuaid:
Passing on this.

PR #1153, opened 2016-09-26
fixtures: add test.m4a

zmwangx:
`test.m4a` contains an AAC LC stream inside an M4A (MPEG-4 Part 14) container. It is generated by FFmpeg 3.1.3 from `test.wav`:

    ffmpeg -y -i test.wav -c:a aac -profile:a aac_low -b:a 32k test.m4a

`test.m4a` facilitates testing utilities or libraries that work with the AAC codec or the M4A container, e.g., `alac`.

`test.m4a` is 1689 bytes in size.


zmwangx:
This can be used in many formulae (it can be a drop-in replacement for `test.mp3` in most cases, for instance), though it's not necessary.

However, finding two instances isn't hard:

- `alac`: current sample (downloaded as a resource) is broken, we can switch to this one:

    ```
$ alac -t test.m4a
file type: mp4a
```

- `aacgain`: "AAC-supporting version of mp3gain", so it would be nice to make sure it actually supports AAC in addition to MP3, so replacing `test.mp3` with `test.m4a` is a pretty natural move.

MikeMcQuaid:
@zmwangx Sounds good!

zmwangx:
Promised test rewrites in https://github.com/Homebrew/homebrew-core/pull/5259.

PR #1156, opened 2016-09-26
Speed up `parallel_tests` by splitting integration tests.

reitermarkus:

-----

`brew tests` on my iMac Core i5:
  - before: ~105 seconds
  - after: ~45 seconds

Damn.


MikeMcQuaid:
Nice. I think if there's existing files like e.g. `bottle_tests.rb` it's worth integrating them in there and otherwise creating a new `bottle_tests.rb` rather than having dedicated integration tests files.

penman:
> Took 14 min on master

:O

reitermarkus:
> if there's existing files like e.g. bottle_tests.rb it's worth integrating them in there

I think this would be counter-productive, as this would again result in *less* files being split across multiple processes.

penman:
@reitermarkus I don't think it would make much of a difference. The issue before wasn't so much that the files weren't small enough; it was that a single file took longer to run than all the other ones combined. As long as that doesn't happen again, the difference between the way it is now and if some of the tests were moved back in to several other files should be negligible.

reitermarkus:
> I don't think it would make much of a difference.

Yes, probably not. Renamed/merged the integration test files with existing files now.

apjanke:
LGTM.

On my quad-core i7 MacBook Pro:

* before: 92 seconds
* after: 30 seconds

MikeMcQuaid:
Great work, thanks @reitermarkus!

MikeMcQuaid:
And for @penman for the previous PR using this gem, too.

PR #1169, opened 2016-09-27
Use separate RuboCop configs for formulae and core code.

### LINKED ISSUE #1168: Rubocop tap problems.
Our Rubocop configuration has exclusions which can be gleefully ignored by both Rubocop itself and things like Atom's [Rubocop package](https://atom.io/packages/linter-rubocop). So for example, after https://github.com/Homebrew/brew/pull/1128 being merged I do something like `brew edit subversion18` and I get told to use a guard clause in _one place_, so it wants me to rewrite the formula like this:

```ruby
if build.with? "java"
  system "make", "javahl"
  system "make", "install-javahl"
end

return unless build.with? "ruby"
# Peg to system Ruby
system "make", "swig-rb", "EXTRA_SWIG_LDFLAGS=-L/usr/lib"
system "make", "install-swig-rb"
```

Which is a complete style mismatch. Same sort of thing with the Rubocop 1.9+ syntax preference both inside and outside of text editors:

```bash
cd /usr/local/Homebrew/Library/Taps/homebrew/homebrew-x11
rubocop --auto-correct --config /usr/local/Homebrew/Library/.rubocop.yml
```
Produces a [snipped] diff like this:
```diff
diff --git a/chezscheme.rb b/chezscheme.rb
index 0063006..0266b72 100644
--- a/chezscheme.rb
+++ b/chezscheme.rb
@@ -10,7 +10,7 @@ class Chezscheme < Formula
     sha256 "39c50e48f35bec6715bf3046e17034b458c997766762c1dad665f78763025866" => :mavericks
   end
 
-  depends_on :x11 => :build
+  depends_on x11: :build
 
   def install
     system "./configure",
diff --git a/clusterit.rb b/clusterit.rb
index 5a9d544..25bf9d9 100644
--- a/clusterit.rb
+++ b/clusterit.rb
@@ -13,8 +13,8 @@ class Clusterit < Formula
 
   depends_on :x11
 
-  conflicts_with "couchdb-lucene", :because => "both install a `run` binary"
-  conflicts_with "pdsh", :because => "both install `dshbak`"
+  conflicts_with "couchdb-lucene", because: "both install a `run` binary"
+  conflicts_with "pdsh", because: "both install `dshbak`"
 
   def install
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
diff --git a/curlftpfs.rb b/curlftpfs.rb
index f9e9721..cce36f2 100644
--- a/curlftpfs.rb
+++ b/curlftpfs.rb
@@ -3,7 +3,7 @@ class Curlftpfs < Formula
   url "https://downloads.sourceforge.net/project/curlftpfs/curlftpfs/0.9.2/curlftpfs-0.9.2.tar.gz"
   sha256 "4eb44739c7078ba0edde177bdd266c4cfb7c621075f47f64c85a06b12b3c6958"
 
-  head ":pserver:anonymous:@curlftpfs.cvs.sourceforge.net:/cvsroot/curlftpfs", :using => :cvs
+  head ":pserver:anonymous:@curlftpfs.cvs.sourceforge.net:/cvsroot/curlftpfs", using: :cvs
 
   depends_on "pkg-config" => :build
   depends_on "autoconf" => :build
diff --git a/dmenu.rb b/dmenu.rb
index 56bf091..97afc92 100644
```

The only way I can find to get it to behave as-expected is to:
```bash
cd /usr/local/Homebrew/Library
rubocop --auto-correct --config .rubocop.yml Taps/homebrew/homebrew-x11
```
which is semi-practical at best, but seems impossible to enact inside of something like Atom, even if I try something like:
```bash
cd /usr/local/Homebrew/Library
$EDITOR Taps/homebrew/homebrew-core/Formula/subversion.rb
```

I presume users are seeing this issue as well because several people have tried to change to the 1.9+ syntax despite us scoping out exclusions for the taps.

CC @reitermarkus @MikeMcQuaid

reitermarkus:

-----

`Lint/NestedMethodDefinition` seems to be bugged, as it is the only cop that is not inherited from the `.rubocop_todo.yml`, at least on my machine. Let's see if CI has the same problem.

Fixes #1168


MikeMcQuaid:
This looks good. Do you use an editor with Rubocop integration? If so, can you try to open a single `Library/Taps/**/*.rb` file and see if it uses the correct file?

MikeMcQuaid:
I'm wondering if that'd be fixed by having the core file live in `Library/Homebrew`.

reitermarkus:
> Do you use an editor with Rubocop integration?

No, @DomT4 could you test this? I guess the RuboCop integration probably expects to find a `.rubocop.yml` instead of `.rubocop_formula` and `.rubocop_core.yml`. In this case we could probably have `Library/Homebrew/.rubocop.yml` and `Library/Taps/.rubocop.yml`.

DomT4:
Sure, I'll yank it down locally and test it in the next hour or so.

> I guess the RuboCop integration probably expects to find a .rubocop.yml instead of .rubocop_formula and .rubocop_core.yml.

I agree with your suspicion here though. Seems likely.

jawshooah:
> In this case we could probably have Library/Homebrew/.rubocop.yml and Library/Taps/.rubocop.yml.

Yup, that would be ideal I think. It's also how RuboCop itself expects configs to be structured, regardless of editor integrations.

DomT4:
Sorry for the massive delay, one of those days. One of those months really 😕.

This seems to have wildly confused Atom's Rubocop integration, heh. I now see messages about the class having too many lines, Ruby 1.9 hash syntax & even reverting double quotes to single quotes (which I'm not sure where that's coming from).

The plugin I use is [this one](https://github.com/AtomLinter/linter-rubocop), which you can't pass configurations to explicitly, it "sniffs" them, so I guess this somehow upset that sniffing. The same thing happens with Rubocop as well though, so perhaps a Rubocop issue rather than a linter issue (?). if I do:
```bash
cd /usr/local/Homebrew/Library/Taps/homebrew/homebrew-x11
rubocop --auto-correct
```
It changes all the double quotes to single. If I:
```bash
cd /usr/local/Homebrew/Library/Taps/homebrew/homebrew-x11
rubocop --auto-correct --config /usr/local/Homebrew/Library/.rubocop_formula.yml
```
It does behave as expected.

reitermarkus:
> I now see messages about the class having too many lines, Ruby 1.9 hash syntax & even reverting double quotes to single quotes (which I'm not sure where that's coming from).

This is the default config of RuboCop, so it just doesn't pick up the `_core.yml` and `_formula.yml` files.

MikeMcQuaid:
> In this case we could probably have Library/Homebrew/.rubocop.yml and Library/Taps/.rubocop.yml.

Agreed this sounds ideal 👍

reitermarkus:
Ok, I renamed/moved the separate configs. @DomT4, you wanna test this again?

DomT4:
Seem to be back where we started 😢. Atom now nags me to change to the 1.9 hash syntax, and nags about guard clauses. Rubocop concurs:

```bash
cd /usr/local/Homebrew && rubocop --auto-correct --config /usr/local/Homebrew/Library/.rubocop.yml /usr/local/Homebrew/Library/Taps/homebrew/homebrew-x11
```
```diff
diff --git a/clusterit.rb b/clusterit.rb
index 5a9d544..25bf9d9 100644
--- a/clusterit.rb
+++ b/clusterit.rb
@@ -13,8 +13,8 @@ class Clusterit < Formula

   depends_on :x11

-  conflicts_with "couchdb-lucene", :because => "both install a `run` binary"
-  conflicts_with "pdsh", :because => "both install `dshbak`"
+  conflicts_with "couchdb-lucene", because: "both install a `run` binary"
+  conflicts_with "pdsh", because: "both install `dshbak`"
```
etc. Same thing if I `cd` into the tap directory and run a simple `rubocop --auto-correct`.

reitermarkus:
What if you rename `/usr/local/Homebrew/Library/.rubocop.yml` to something else that rubocop won't pick up, since this is not a *complete* config, but shared between core and formulae? (You'll need to update the other two to `inherit_from` the new name.)

MikeMcQuaid:
I've updated this so that:
- `Library/.rubocop.yml` is the file that's then inherited and overridden by `Library/Homebrew/.rubocop.yml`. This means it'll work for all taps by default without needing to create `Library/Taps` in Git (which I'm slightly 😨 of for no real reason).
- Homebrew/core passes `brew style` and also editing a random file uses the `Library/.rubocop.yml` rules.
- Disabled the rules that were stopping Homebrew/core from passing. We can iterate on more rules in a future PR.

Going to merge this when 💚 to avoid more Rubocop weirdness creeping into core CC @ilovezfs @DomT4

reitermarkus:
Thanks, @DomT4, for confirming. Merging this then.

DomT4:
😄 Lots of green, but not the right green. Could be a while, CI queue is kind of... intense at the moment. Thanks for working on this @reitermarkus, very much appreciate you taking the time.

PR #1177, opened 2016-09-28
Man exit code

scpeters:

-----

I noticed a comment in #1126 that there is no CI test that the manpages are up to date. I started adding such a test in https://github.com/Homebrew/homebrew-test-bot/pull/4 and @MikeMcQuaid had a great suggestion to add some of that code directly to `dev-cmd/man.rb`. This PR now displays a `git diff --stat` on the auto-generated manpage files whenever you run `brew man`. Furthermore, if you pass the `--exit-code` argument, it will return a failing status code if the diff is not empty. This will make it trivial to use for CI.

I've also updated the man pages, since they were out of date. Feel free to checkout 8839d8b if you want to test the `brew man` behavior when there are changes.

### DIFF
--- a/Library/Homebrew/dev-cmd/man.rb
+++ b/Library/Homebrew/dev-cmd/man.rb
@@ -1,5 +1,9 @@
-#:  * `man`:
+#:  * `man` [`--fail-if-changed`]:
 #:    Generate Homebrew's manpages.
+#:
+#:    If `--fail-if-changed` is passed, the command will return a failing
+#:    status code if changes are detected in the manpage outputs.
+#:    This can be used for CI to be notified when the manpages are out of date.
 
 require "formula"
 require "erb"
@@ -17,6 +21,12 @@ module Homebrew
       odie "`brew man --link` is now done automatically by `brew update`."
     else
       regenerate_man_pages
+    end
+
+    if system "git", "-C", HOMEBREW_REPOSITORY, "diff", "--quiet", "docs/brew.1.html", "manpages"
+      puts "No changes to manpage output detected."
+    elsif ARGV.include?("--fail-if-changed")
+      Homebrew.failed = true
     end
   end
 
--- a/docs/brew.1.html
+++ b/docs/brew.1.html
@@ -467,7 +467,8 @@ for instance, for implementing pre-commit hooks.</p></dd>
 <dt><code>bump-formula-pr</code> [<code>--devel</code>] [<code>--dry-run</code>] [<code>--audit</code>|<code>--strict</code>] <code>--tag=</code><var>tag</var> <code>--revision=</code><var>revision</var> <var>formula</var></dt><dd><p>Creates a pull request to update the formula with a new url or a new tag.</p>
 
 <p>If a <var>url</var> is specified, the <var>sha-256</var> checksum of the new download must
-also be specified.</p>
+also be specified. A best effort to determine the <var>sha-256</var> and <var>formula</var>
+name will be made if either or both values are not supplied by the user.</p>
 
 <p>If a <var>tag</var> is specified, the git commit <var>revision</var> corresponding to that
 tag must also be specified.</p>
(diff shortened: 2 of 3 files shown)


zmwangx:
> it will return a failing status code if the diff is not empty.

Returning a failing status code when `brew man` actually did its job feels a little bit weird. What about the other way round.

scpeters:
I used `--exit-code` to match the `git` argument, but I think it is semantically equivalent to `--check-outdated`, in that it returns true if the man pages are already up to date. I think the new option is mainly intended for CI.

I will defer to the maintainers' preferences of course.

zmwangx:
Your argument is very reasonable, but another way to think about it is "nothing to do" => failing exit code. Just a thought, I don't feel strongly either way.

scpeters:
The CI is :red: because it thinks coverage will go down, but I think this will improve coverage once `brew man` is added to `test-bot` in https://github.com/Homebrew/homebrew-test-bot/pull/4

Library/Homebrew/dev-cmd/man.rb
```diff
@@ -18,6 +18,10 @@ def man
     else
       regenerate_man_pages
     end
+
+    git_diff_args = ["-C", HOMEBREW_REPOSITORY, "diff", "--stat"]
+    git_diff_args << "--exit-code" if ARGV.flag? "--exit-code"
+    exit system "git", *git_diff_args, "docs/brew.1.html", "manpages"
```
MikeMcQuaid: I think you can avoid doing any of this unless `--exit-code` is passed. Also, use `ARGV.include? ` instead of `flag?`. I'd also favour setting `Homebrew.failed` instead of calling `exit`.

scpeters: I had thought it would be fun to display a `git diff --stat` whenever `brew man` was invoked, so you could see if there are changes:

~~~
$ brew man
Writing HTML fragment to /usr/local/Homebrew/docs/brew.1.html
Writing man page to /usr/local/Homebrew/manpages/brew.1
Writing man page to /usr/local/Homebrew/manpages/brew-cask.1
 docs/brew.1.html | 12 +++++++++++-
 manpages/brew.1  | 11 ++++++++++-
 2 files changed, 21 insertions(+), 2 deletions(-)
~~~

If we only want that to be displayed when `--exit-code` is passed, then I will wrap those commands in an `if` block as you suggest.

MikeMcQuaid: I think we can avoid showing any `git diff` output and just saying "no changes made" or something and setting the exit code accordingly to `--exit-code`.

scpeters: Ok, since you don't want to display any `git diff`, I switched from `git diff --stat` to `git diff --quiet`, which implies `--exit-code`. I also renamed the `brew man --exit-code` option to `--fail-if-changed`, which is a bit more descriptive, and hopefully not too cumbersome.

Also, adding documentation of the `--fail-if-changed` option has made the manpages out of date in a bit of self-reference (updated in e748881).

The test-bot PR will be out of date if we stick with the `--fail-if-changed` option, since it's still using `--exit-code`.

Library/Homebrew/dev-cmd/man.rb
```diff
@@ -18,6 +22,11 @@ def man
     else
       regenerate_man_pages
     end
+
+    args = ["-C", HOMEBREW_REPOSITORY, "diff", "--quiet"]
```
MikeMcQuaid: I'd just inline these as you're not using the var in more than one place and it makes the command easier to follow.

Library/Homebrew/dev-cmd/man.rb
```diff
@@ -18,6 +22,11 @@ def man
...
+    changed = !system("git", *args, "docs/brew.1.html", "manpages")
+    puts "no changes made" if !changed
```
MikeMcQuaid: `No changes to manpage output detected.`

Library/Homebrew/dev-cmd/man.rb
```diff
@@ -18,6 +22,11 @@ def man
...
+    Homebrew.failed = true if changed && ARGV.include?("--fail-if-changed")
```
MikeMcQuaid: I'd just go for an `if system ... else ... end`  block.

Library/Homebrew/dev-cmd/man.rb
```diff
@@ -1,5 +1,9 @@
-#:  * `man`:
+#:  * `man` [`--fail-if-changed`]:
 #:    Generate Homebrew's manpages.
+#:
+#:    If `--fail-if-changed` is passed, the command will return a failing
+#:    status code if changes are detected in the manpage outputs.
+#:    This can be used for CI to be notified when the manpages are out of date.
```
MikeMcQuaid: Self-testing PR 👏

scpeters:
@MikeMcQuaid thanks for the recommendations; I've implemented them, and I think it makes the logic more readable. I squashed and force pushed; sorry if that makes it hard to see what changed.

MikeMcQuaid:

scpeters:
cheers!

PR #1178, opened 2016-09-28
superenv: more 10.11 clock_gettime Autotools fixes

ilovezfs:
some build systems check ac_cv_search_clock_gettime instead of
ac_cv_func_clock_gettime so the former should also be set to "no"

libev.m4 unsets ac_cv_func_clock_gettime, but if ac_have_clock_syscall
is defined, it will leave ac_cv_func_clock_gettime alone:

  http://cvs.schmorp.de/libev/libev.m4?view=markup#l23


MikeMcQuaid:
Thanks @ilovezfs!

ilovezfs:
Thanks for the merge!

PR #332, opened 2016-06-07
tests: sanitize env more for consistency

DomT4:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Ensures that like we do for formulae install & test processes we "borrow" the `$HOME` env to ensure consistency for the duration of `tests`. Comments above the changes in the file explain the rest, really.

Testing with `puts` shows things are being set as expected:

```
BEFORE_TESTS = /Users/Dominyk

==> Installing or updating 'bundler' gem
Fetching: bundler-1.12.5.gem (100%)
Successfully installed bundler-1.12.5
1 gem installed

DURING_TESTS = /tmp/tests20160607-26482-1w5ubie

Finished in 46.985927s, 18.1969 runs/s, 44.3750 assertions/s.

855 runs, 2085 assertions, 0 failures, 0 errors, 0 skips
AFTER_ENSURE = /Users/Dominyk
```

Checked with a deliberate failure to ensure it still fails as expected, which it does:
```
  1) Failure:
TapTest#test_git_variant [/usr/local/Library/Homebrew/test/test_tap.rb:155]:
--- expected
+++ actual
@@ -1 +1 @@
-"e1893a6bd191ba895c71b652ff8376a6114c7fa6"
+"e1893a6bd191ba895c71b652ff8376a6114c7fa7"
```

Alternative to https://github.com/Homebrew/brew/pull/324.

PR #1186, opened 2016-09-29
formula: unset XDG env vars during test

zmwangx:

-----

Having XDG environment variables set during test means the possibility of sandbox violation (software writing to `XDG_DATA_HOME` or `XDG_CACHE_HOME` outside the temporary `HOME`, see e.g. `buku`) or undesired configuration pick-up (see e.g. `git-test` and in fact any git-related
tests) when testing XDG Basedir Spec compliant software.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1558,6 +1558,9 @@ class Formula
     old_temp = ENV["TEMP"]
     old_tmp = ENV["TMP"]
     old_term = ENV["TERM"]
+    old_xdg_config_home = ENV.delete("XDG_CONFIG_HOME")
+    old_xdg_data_home = ENV.delete("XDG_DATA_HOME")
+    old_xdg_cache_home = ENV.delete("XDG_CACHE_HOME")
     ENV["CURL_HOME"] = old_curl_home || old_home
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'd like us to lean into the "proper" solution for this instead of adding more variables and instead whitelist the variables we allow into `test` (and `install` but that could be done afterwards).

zmwangx:
A proper solution is great but I doubt it will come from me (at least not any time soon).

DomT4:
Is it possible these variables could cause attempts to write outside the sandbox during build as well, or is this purely a testing issue?

MikeMcQuaid:
> A proper solution is great but I doubt it will come from me (at least not any time soon).

Ok, no worries. Personally I'm 👎 on the whack-a-mole that is adding more individual variables (which I also think applies to https://github.com/Homebrew/brew/pull/332 CC @DomT4)

zmwangx:
> Is it possible these variables could cause attempts to write outside the sandbox during build as well, or is this purely a testing issue?

It's certainly "possible" because build scripts could do anything, but a sane build script probably shouldn't do that, and I've yet to encounter one. XDG Basedir Spec is for users.

MikeMcQuaid:
Passing on this.

PR #1190, opened 2016-09-29
brew.sh: don't autoupdate if --help passed

### LINKED ISSUE #1188: brew autoupdate runs on brew install --help
TL;DR: `brew install --help` calls autoupdate

And now to the long story. I was surprised by the update of readline 7; i.e. it crashed other packages I'm using on a daily basis. Therefore, the easiest way seamed to revert to an older commit and pin readline. Searching for the command on how to disable the autoupdate triggered by `brew install` (Since this would checkout the master again) I called `brew install --help`. Sadly this command also triggered the autoupdate.
I know that this is a rare scenario and this behavior may also be by design but for me this is inconvenient.

vladshablinsky:

-----

Seems to fix #1188.

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -32,20 +32,17 @@ begin
 
   empty_argv = ARGV.empty?
   help_flag_list = %w[-h --help --usage -?]
-  help_flag = false
+  help_flag = !ENV["HOMEBREW_HELP"].nil?
   internal_cmd = true
   cmd = nil
 
   ARGV.dup.each_with_index do |arg, i|
     break if help_flag && cmd
 
-    if help_flag_list.include?(arg)
-      # Option-style help: Both `--help <cmd>` and `<cmd> --help` are fine.
-      help_flag = true
-    elsif arg == "help" && !cmd
+    if arg == "help" && !cmd
       # Command-style help: `help <cmd>` is fine, but `<cmd> help` is not.
       help_flag = true
-    elsif !cmd
+    elsif !cmd && !help_flag_list.include?(arg)
       cmd = ARGV.delete_at(i)
     end
   end
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -179,6 +179,15 @@ then
   set -- "$@" -v
 fi
 
+for arg in "$@"
+do
+  if [[ $arg = "--help" || $arg = "-h" || $arg = "--usage" || $arg = "-?" ]]
+  then
+    export HOMEBREW_HELP="1"
+    break
+  fi
+done
+
 HOMEBREW_ARG_COUNT="$#"
 HOMEBREW_COMMAND="$1"
 shift
@@ -268,6 +277,7 @@ setup-analytics
 report-analytics-screenview-command
 
 update-preinstall() {
+  [[ -z "$HOMEBREW_HELP" ]] || return
   [[ -z "$HOMEBREW_NO_AUTO_UPDATE" ]] || return
   [[ -z "$HOMEBREW_UPDATE_PREINSTALL" ]] || return
 


Library/Homebrew/brew.sh
```diff
@@ -179,6 +179,15 @@ then
   set -- "$@" -v
 fi
 
+# Check if user passes --help command which should prevent auto-update.
+for arg in "$@"
+do
+  if [[ $arg = "--help" ]]
+  then
+    export HOMEBREW_COMMAND_HELP="1"
```
dunn: Since it only persists for the duration of the subshell, should we just set `HOMEBREW_NO_AUTO_UPDATE` instead of adding another variable?

vladshablinsky: :+1: amending.

Library/Homebrew/brew.sh
```diff
@@ -179,6 +179,15 @@ then
   set -- "$@" -v
 fi
 
+# Check if user passes --help command which should prevent auto-update.
+for arg in "$@"
+do
+  if [[ $arg = "--help" ]]
```
zmwangx: Same goes for `"-h"`.

vladshablinsky: Seems like even more than that. https://github.com/Homebrew/brew/blob/master/Library/Homebrew/brew.rb#L34
```ruby
 help_flag_list = %w[-h --help --usage -?]
```

zmwangx: Yeah. `-?` — didn't know that. Weird one.

reitermarkus:
Can't we set `HOMEBREW_NO_AUTO_UPDATE` in `brew.rb` by using the already existing `help_flag_list` instead of duplicating the flags, or is it already too late by then? Otherwise I would suggest removing the `help_flag_list` and instead exporting `HOMEBREW_COMMAND_HELP` like in the first diff.

zmwangx:
By the time you get to `brew.rb` the battle is already lost. `update-preinstall` is in `brew.sh`.

reitermarkus:
Ah, ok, let's get rid of `help_flag_list` then, as it is used only once anyways: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/brew.rb#L42

zmwangx:
Getting rid of `help_flag_list` and using an in place list literal is good, but I doubt we want to set another env var.

vladshablinsky:
> Getting rid of help_flag_list and using an in place list literal is good

With `help_flag_list` it looks more readable since variable's name explains variable's purpose.

reitermarkus:
We could completely remove the `help_flag_list` by calling the `help` command instead of also parsing the help flags in `brew.rb`.

```bash
# check for help flags and call `help` instead
arg_n=0
for arg in "$@"
do
  ((arg_n++))

  if [[ $arg = "--help" || $arg = "-h" || $arg = "--usage" || $arg = "-?" ]]
  then
    set -- "$HOMEBREW_COMMAND" "${@:1:arg_n-1}" "${@:arg_n+1}"
    HOMEBREW_COMMAND="help"
    break
  fi
done
```

MikeMcQuaid:
> I doubt we want to set another env var.

It's fine to set another environment variable to pass state between Bash and Ruby code. Personally I'd rather see us do that than overload `HOMEBREW_NO_AUTO_UPDATE`.

vladshablinsky:
@MikeMcQuaid @reitermarkus @zmwangx 

What if we transform any command like `brew (--help|--usage|-?|-h|<some_command>)+` into
`brew help <some_command>` e.g. 
```
brew --help install --usage -> brew help install
brew install -h -> brew help install
```

Basically we just remove all the help flags and unshift `help` to `$@`.

```shell
arg_n=0
for arg in "$@"
do
  ((arg_n++))

  if [[ $arg = "--help" || $arg = "-h" || $arg = "--usage" || $arg = "-?" ]]
  then
    set -- "${@:1:arg_n-1}" "${@:arg_n+1}" && ((arg_n--))

    if [[ $1 != "help" ]]
    then
      set -- "help" "$@" && ((arg_n++))
    fi
  fi
done

```

MikeMcQuaid:
> What if we transform any command like brew (--help|--usage|-?|-h|<some_command>)+ into
> brew help <some_command> e.g.

Seems like a good idea (even if the Bash code currently makes me want to cry a bit). It may be cleaner to consider just setting `HOMEBREW_HELP=1` rather than modifying the argument list.

vladshablinsky:
>It may be cleaner to consider just setting HOMEBREW_HELP=1 rather than modifying the argument list.

The idea was to prevent parsing `--help`-like flags twice: in `sh` and then in `rb`. Sorry for such a Bash code, I'm quite new to writing my own Bash code.

Assuming what you have said, can this be a solution?

```diff
diff --git a/Library/Homebrew/brew.sh b/Library/Homebrew/brew.sh
index 22ebb87..8ecaadd 100644
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -179,6 +179,15 @@ then
   set -- "$@" -v
 fi
 
+for arg in "$@"
+do
+  if [[ $arg = "--help" || $arg = "-h" || $arg = "--usage" || $arg = "-?" ]]
+  then
+    HOMEBREW_HELP="1"
+    break
+  fi
+done
+
 HOMEBREW_ARG_COUNT="$#"
 HOMEBREW_COMMAND="$1"
 shift
@@ -268,6 +277,7 @@ setup-analytics
 report-analytics-screenview-command
 
 update-preinstall() {
+  [[ "$HOMEBREW_HELP" != "1" ]] || return
   [[ -z "$HOMEBREW_NO_AUTO_UPDATE" ]] || return
   [[ -z "$HOMEBREW_UPDATE_PREINSTALL" ]] || return
 
```

MikeMcQuaid:
> The idea was to prevent parsing --help-like flags twice: in sh and then in rb. Sorry for such a Bash code, I'm quite new to writing my own Bash code.

Not your bad, this stuff is just far messier to do in Bash/

> Assuming what you have said, can this be a solution?

Yeh, that looks good. You could also `export` it and then delete the `--help` handling code from `brew.rb` in favour of checking if the environment variable is et.

Library/Homebrew/brew.rb
```diff
@@ -32,20 +32,17 @@ def require?(path)
 
   empty_argv = ARGV.empty?
   help_flag_list = %w[-h --help --usage -?]
-  help_flag = false
+  help_flag = ENV["HOMEBREW_HELP"]
```
reitermarkus: `!ENV["HOMEBREW_HELP"].nil?`, maybe, so we get a *nice* boolean.

MikeMcQuaid:
Thanks @vladshablinsky!

PR #1192, opened 2016-09-30
keg: create symlinks in opt for formula aliases

ilovezfs:


ilovezfs:
based on that test run, it sure looks like it will need to be bootstrapped with the existing ones rebuilt against the versioned name

MikeMcQuaid:
CC @Homebrew/maintainers for thoughts on this change and @penman too.

zmwangx:
Could someone write up the rationale behind this change? I see discussions in team chat but not everyone read transcripts after the fact, and @alyssais (looks like the handle changed) don't have access to that.

ilovezfs:
The existing bottles are linked against opt/boost not opt/boost@1.61. So renaming boost to boost@1.61 breaks them. The alternatives are not to rename boost to boost@1.61 and just to go from boost to boost@1.62 (that pattern is already underway with openssl), or have duplicate boost and boost@1.61 formulae alongside boost@1.62, until dependents are transferred from boost to boost@1.61 or boost@1.62, and then delete the duplicate boost formula, and create a boost alias instead.

zmwangx:
Do we want to limit opt linking for aliases to @ formulae only? Having both `cowthink` and `cowsay` in opt is pointless (but harmless either).

alyssais:
I like the idea that there's nothing special about @ formulae (that they're just normal aliases), but that could definitely end up not being practical.

ilovezfs:
@alyssais I agree with you. I think we'll want a full-on DSL for this stuff at some point relatively soon hehe

alyssais:
I think this looks good. Is it cool if I make a couple of comments on the code?

ilovezfs:
of course!

based on that test-bot run, I'm :-1: on my own PR though :)

alyssais:
> based on that `brew tests` run, I'm 👎 on my own PR though :)

Ah, sure. I'll wait a bit until that's resolved then. :)

MikeMcQuaid:
> I like the idea that there's nothing special about @ formulae (that they're just normal aliases), but that could definitely end up not being practical.

I agree with this.

DomT4:
> The alternatives are not to rename boost to boost@1.61 and just to go from boost to boost@1.62 (that pattern is already underway with openssl)

This would be my recommendation, FWIW.

DomT4:
Although we need to resolve the issues around `keg_only` and so on either way with Boost, especially if my comments on `boost` and `boost-python` last night prove accurately remembered. It didn't matter for `openssl` and `openssl@1.1` because both are always going to be `keg_only` when Homebrew is located in `/usr/local`.

MikeMcQuaid:
>> The alternatives are not to rename boost to boost@1.61 and just to go from boost to boost@1.62 (that pattern is already underway with openssl)
>
> This would be my recommendation, FWIW.

Even if that's the case I think this alias change would be a good one.

DomT4:
Yes, I have no issues with aliases also being in `opt` pointing to the "true" source of the alias, even for non-`@` formulae. Might be a good idea to add some tests, if possible, for this change though to ensure we don't bork it in future somehow.

Library/Homebrew/keg.rb
```diff
@@ -197,6 +197,11 @@ def optlinked?
 
   def remove_opt_record
     opt_record.unlink
+    unless aliases.empty?
+      aliases.each do |a|
```
MikeMcQuaid: Can do the each even if `empty?` as it'll be a no-op.

Library/Homebrew/keg.rb
```diff
@@ -390,9 +395,25 @@ def remove_oldname_opt_record
     @oldname_opt_record = nil
   end
 
+  def aliases
+    formula_name = rack.basename.to_s
+    aliases_path = Formula[formula_name].tap.path/"Aliases"
+    result = aliases_path.children.select do |c|
+      c.symlink? && c.readlink.basename(".rb").to_s == formula_name
+    end
+    result.map(&:basename).map(&:to_s)
```
MikeMcQuaid: Can get this from `Formula#aliases` (which also handles the no-tap case)

Library/Homebrew/keg.rb
```diff
@@ -390,9 +395,25 @@ def remove_oldname_opt_record
...
+  end
+
   def optlink(mode = OpenStruct.new)
     opt_record.delete if opt_record.symlink? || opt_record.exist?
     make_relative_symlink(opt_record, path, mode)
+    unless aliases.empty?
```
MikeMcQuaid: Again, `aliases.each` is a no-op if this is `empty` so you can just use `.each` without the check.

Library/Homebrew/keg.rb
```diff
@@ -390,9 +395,25 @@ def remove_oldname_opt_record
...
+      aliases.each do |a|
+        alias_opt_record = opt_record.parent/a
+        alias_opt_record.delete if alias_opt_record.symlink? || alias_opt_record.exist?
```
MikeMcQuaid: Will a `symlink?` not always `exist?`?

ilovezfs:
@MikeMcQuaid I'm not sure it matters but do you prefer `opt/boost -> boost@1.62` or `opt/boost -> ../Cellar/boost@1.62`?

MikeMcQuaid:
@ilovezfs Then latter just in case people are relying on being able to follow one layer of symlinks to find the Cellar :+1:

Library/Homebrew/keg.rb
ilovezfs: Broken symlinks return false for exist?

MikeMcQuaid: TIL.

MikeMcQuaid:
There's some test failures but code and local testing look good to me.

MikeMcQuaid:
@ilovezfs Thoughts on this? Shout if you want help with test failures.

ilovezfs:
@MikeMcQuaid this has introduced a bug for anything that has an alias but was installed before this PR shipped:
```
iMac-TMP:~ joe$ brew uninstall ag
Uninstalling /usr/local/Cellar/the_silver_searcher/1.0.2... (10 files, 113.1K)
Error: No such file or directory - /usr/local/opt/ag
```

MikeMcQuaid:
@ilovezfs Working on it.

MikeMcQuaid:
@ilovezfs Opened https://github.com/Homebrew/brew/pull/2085 to address this.

PR #1193, opened 2016-09-30
fix tests failing when verbose set

MatzFan:
- [√ ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [ √] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [ √] Have you added an explanation of what your changes do and why you'd like us to include them?
- [√ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [√ ] Have you successfully run `brew tests` with your changes locally?

-----

### DIFF
--- a/Library/Homebrew/test/test_resource.rb
+++ b/Library/Homebrew/test/test_resource.rb
@@ -117,7 +117,7 @@ class ResourceTests < Homebrew::TestCase
   end
 
   def test_verify_download_integrity_mismatch
-    fn = stub(file?: true)
+    fn = stub(file?: true, basename: "Ftest")
     checksum = @resource.sha256(TEST_SHA256)
 
     fn.expects(:verify_checksum).with(checksum)
--- a/Library/Homebrew/test/test_resource.rb
+++ b/Library/Homebrew/test/test_resource.rb
@@ -117,7 +117,7 @@ class ResourceTests < Homebrew::TestCase
   end
 
   def test_verify_download_integrity_mismatch
-    fn = stub(file?: true, basename: "Ftest")
+    fn = stub(file?: true)
     checksum = @resource.sha256(TEST_SHA256)
 
     fn.expects(:verify_checksum).with(checksum)
--- a/Library/Homebrew/test/test_tap.rb
+++ b/Library/Homebrew/test/test_tap.rb
@@ -25,7 +25,7 @@ class IntegrationCommandTestTap < IntegrationCommandTestCase
     assert_match "Unpinned homebrew/foo", cmd("tap-unpin", "homebrew/foo")
     assert_match "Tapped", cmd("tap", "homebrew/bar", path/".git")
     assert_match "Untapped", cmd("untap", "homebrew/bar")
-    assert_match /.*/, cmd("tap", "homebrew/bar", path/".git", "-q", "--full")
+    cmd("tap", "homebrew/bar", path/".git", "-q", "--full")
     assert_match "Untapped", cmd("untap", "homebrew/bar")
   end
 end
--- a/Library/Homebrew/test/test_tap.rb
+++ b/Library/Homebrew/test/test_tap.rb
@@ -25,7 +25,7 @@ class IntegrationCommandTestTap < IntegrationCommandTestCase
     assert_match "Unpinned homebrew/foo", cmd("tap-unpin", "homebrew/foo")
     assert_match "Tapped", cmd("tap", "homebrew/bar", path/".git")
     assert_match "Untapped", cmd("untap", "homebrew/bar")
-    assert_equal "", cmd("tap", "homebrew/bar", path/".git", "-q", "--full")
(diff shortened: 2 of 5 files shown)


MatzFan:
Fixes 4 tests failing outside of CI because $VERBOSE = 1 [set here](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/sh.rb#L25)

scpeters:
Out of curiosity I just tried reproducing these test failures, and I'm having trouble getting reproducible results. On master, successive invocations of `VERBOSE=1 brew tests --verbose` at first give 6 failures, then 7. Also `VERBOSE=1 brew tests` (without `--verbose`) generally gives a different number of failures. The good news is that this branch appears to reduce the number of failures for whatever combination of `VERBOSE` and `--verbose` in my limited testing.

If this is important, we could consider running a matrix of verbose options in one of the CI jobs.

MatzFan:
> If this is important, we could consider running a matrix of verbose options in one of the CI jobs.

I'd agree with this suggestion, as it is clear some tests only pass because `verbose` is assumed `false` (which must be that case in CI I assume). Editing out the `ENV["VERBOSE"] = "1"` line in `sh.rb` passed all tests on my machine, so I expect the others still failing locally are caused by the same issue. I'll try and identify the others & do a follow up PR to fix them too. I also get between 1 and 4 failures on this branch depending on seed, so I'll try and fix the sensitivity to ordering too.

MatzFan:
Just fixed the other failing tests. It seems a failing assertion [here](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/test_tap.rb#L28) in `master` when `verbose` is set was the cause. Note this last commit skips the assertion of empty output on line 28 in `master`, as I couldn't see what it was trying to prove.

This fixes the other failures in `test_tap.rb` as it ensures the following line executes - which untap's `homebrew/bar`. Without that, 2or 3 other tests were failing, depending on seed.

I am totally new to minitest and come from an RSpec background where 1 test == 1 assertion, so this sort of test interdependence is new (for speed I guess?). At least it was not hard to debug in this case.

Be interested if @scpeters  or anyone else can validate that all tests now pass consistently with `verbose` set.

reitermarkus:
I think stubbing `ARGV.verbose?` to `false` is not a good solution, as that should be the default case.

I would unset `VERBOSE` and `HOMEBREW_VERBOSE` during tests and only stub `ARGV.verbose?` to `true` if we actually want to test verbosity.

MikeMcQuaid:
> I would unset VERBOSE and HOMEBREW_VERBOSE during tests and only stub ARGV.verbose? to true if we actually want to test verbosity. Re-running the whole test suite just to test verbosity is overkill.

Agreed 👍

MatzFan:
Instead of stubbing, I've made assertions conditional on verbosity - better still?

MikeMcQuaid:
@MatzFan do `ENV.delete "HOMEBREW_VERBOSE"` in `cmd/tests.rb` instead.

MatzFan:
@MikeMcQuaid much better. Will do same for VERBOSE - as it appears to be set [here](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/sh.rb#L25)

MatzFan:
I see codecov has failed this. Should I write a `test_tests.rb` to cover the changes (with system stubbed - so it doesn't recurse!) please advise, thanks.

MikeMcQuaid:
You can ignore the codecov failures.

Library/Homebrew/test/test_resource.rb
```diff
@@ -117,7 +117,7 @@ def test_verify_download_integrity_missing
   end
 
   def test_verify_download_integrity_mismatch
-    fn = stub(file?: true)
+    fn = stub(file?: true, basename: "Ftest")
```
MikeMcQuaid: Can you explain why this change is needed?

Library/Homebrew/test/test_tap.rb
```diff
@@ -25,7 +25,7 @@ def test_tap
     assert_match "Unpinned homebrew/foo", cmd("tap-unpin", "homebrew/foo")
     assert_match "Tapped", cmd("tap", "homebrew/bar", path/".git")
     assert_match "Untapped", cmd("untap", "homebrew/bar")
-    assert_equal "", cmd("tap", "homebrew/bar", path/".git", "-q", "--full")
+    cmd("tap", "homebrew/bar", path/".git", "-q", "--full")
```
MikeMcQuaid: Can you explain why this change is needed?

MatzFan: Not now needed, will change

Library/Homebrew/test/test_resource.rb
MatzFan: Not now needed, will change

MikeMcQuaid:
Thanks @MatzFan!

PR #1195, opened 2016-10-01
Install command completions when installing a tap

joshka:

---

Taps can include completion scripts for external commands under
`completions/bash`, `completions/fish`, or `completions/zsh`. `brew tap`
will automatically install these into the correct directories during
install.

See https://github.com/Homebrew/homebrew-services/pull/73 for a concrete need for completions in a repo.


Library/Homebrew/tap.rb
```diff
@@ -251,6 +252,18 @@ def link_manpages
     link_path_manpages(path, "brew tap --repair")
   end
 
+  def link_completions
+    link_completion("bash", "etc/bash_completion.d")
+    link_completion("zsh", "share/zsh/site-functions")
+    link_completion("fish", "share/fish/vendor_completions.d")
+  end
+
+  def link_completion(shell, dest)
```
MikeMcQuaid: Maybe just go for `src`, `dst` here for consistency.

MikeMcQuaid: Or at least `shell`, `dst`

joshka:
Fixed:
s/dest/dst/g
Left: shell (as would be inconsistent with the unlink side which has another variable named src).

apjanke:
+1 to the general approach here.

I think the `brew tap --repair` command may also need to be updated to include this. Maybe just:

```
    if ARGV.include? "--repair"
      Tap.each(&:link_manpages)
      Tap.each(&:link_completions)
```

There's also a [section in `brew update-report`](https://github.com/Homebrew/brew/blob/564fa8867dd865c686d243ba48c202e0cb6a35fe/Library/Homebrew/cmd/update-report.rb#L297-L306) that deals with completion symlinks. Maybe that should be refactored to use `Tap#link_completions` instead of separate `link_src_dst_dirs` calls? (That would make the `fish` completion linking behavior consistent between `brew tap` and `brew update`, too.) Seems to me that whole `link_completions_and_docs` may belong in `Tap` instead of `update-report`.

```
  def link_completions_and_docs(repository = HOMEBREW_REPOSITORY)
    command = "brew update"
    link_src_dst_dirs(repository/"completions/bash",
                      HOMEBREW_PREFIX/"etc/bash_completion.d", command)
    link_src_dst_dirs(repository/"docs",
                      HOMEBREW_PREFIX/"share/doc/homebrew", command, link_dir: true)
    link_src_dst_dirs(repository/"completions/zsh",
                      HOMEBREW_PREFIX/"share/zsh/site-functions", command)
    link_src_dst_dirs(repository/"manpages",
                      HOMEBREW_PREFIX/"share/man/man1", command)
```

joshka:
You're right on the missing --repair.

I agree on the refactor for DRY, but it likely belongs in some shared class not Tap. Wanted to keep the PR small though for my first proper core change. I'll look into fixing this up tomorrow. 

Will look into this tomorrow.

joshka:
@apjanke I misunderstood you before. Reading more code, I think you're right about this going in tap.rb and will add to this PR tomorrow.

Was this all a clever ploy by @MikeMcQuaid to make me learn ruby? ;)

MikeMcQuaid:
> Was this all a clever ploy by @MikeMcQuaid to make me learn ruby? ;)

😉

joshka:
@apjanke the reason that this exists in `update-report` is that the main Homebrew repository isn't a tap. I'm looking at how to best refactor this to make it neat (and cleanup the inconsistency in manpages between the two at the same time).

joshka:
@MikeMcQuaid you might want to review this for the changes to the man folders.
I think it makes sense to either keep the man/man1 folder hierarchy (perhaps the formula / cask file formats might someday be documented in man5?), or instead explicitly only support man1 for the main repo and taps.

.gitignore
```diff
@@ -38,7 +38,7 @@
 # Unignore our documentation.
 !/.github
 !/docs
-!/manpages
+!/man
```
MikeMcQuaid: I'd rather leave this as-is.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -560,3 +555,13 @@ def installed?(formula)
...
 end
+
+class Docs
+  def self.install(path, command)
+    link_src_dst_dirs(path/"docs", HOMEBREW_PREFIX/"share/doc/homebrew", command, link_dir: true)
+  end
+
+  def self.uninstall(path)
+    unlink_src_dst_dirs(path/"docs", HOMEBREW_PREFIX/"share/doc/homebrew", unlink_dir: true)
+  end
```
MikeMcQuaid: Don't think we need this method.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -560,3 +555,13 @@ def installed?(formula)
     (HOMEBREW_CELLAR/formula.split("/").last).directory?
   end
 end
+
+class Docs
```
MikeMcQuaid: This class shouldn't live here.

Library/Homebrew/tap.rb
```diff
@@ -229,7 +229,7 @@ def install(options = {})
       raise
     end
 
-    link_manpages
```
MikeMcQuaid: I'd rather we kept this into two separate function calls. Apologies if that contradicts other maintainers.

Library/Homebrew/tap.rb
```diff
@@ -630,3 +622,27 @@ def []=(key, value)
...
+    link_src_dst_dirs(path/"completions/bash", HOMEBREW_PREFIX/"etc/bash_completion.d", command)
+    link_src_dst_dirs(path/"completions/zsh", HOMEBREW_PREFIX/"share/zsh/site-functions", command)
+    link_src_dst_dirs(path/"completions/fish", HOMEBREW_PREFIX/"share/fish/vendor_completions.d", command)
+  end
+
+  def self.uninstall(path)
+    unlink_src_dst_dirs(path/"completions/bash", HOMEBREW_PREFIX/"etc/bash_completion.d")
+    unlink_src_dst_dirs(path/"completions/zsh", HOMEBREW_PREFIX/"share/zsh/site-functions")
+    unlink_src_dst_dirs(path/"completions/fish", HOMEBREW_PREFIX/"share/fish/vendor_completions.d")
+  end
```
MikeMcQuaid: Make all these just standalone methods rather than in classes.

.gitignore
joshka: Noted. I'll make the convention in taps the same (manpages rather than man).

Library/Homebrew/cmd/update-report.rb
joshka: Per the comment on the completions / manpages class, it's likely not a class - just a method. Which likely belongs here as the only user of this. Or should this be extended to allow taps to install to docs?

Library/Homebrew/cmd/update-report.rb
joshka: Agreed

Library/Homebrew/tap.rb
joshka: This allows us to add a unit test for the result of `brew --repair`. See https://github.com/Homebrew/brew/pull/1195/files/d0539cd93c826f0129b2dae148761424d59a94e9#diff-3676702e08834bbec8ee89f8f9869981R279

Would you be more likely to accept this if it were named just `repair`?

Library/Homebrew/tap.rb
joshka: Agreed. I wasn't quite sure of Homebrew's OOP vs Procedural code conventions here and kind of expected that this might be your take.

Library/Homebrew/cmd/update-report.rb
MikeMcQuaid: Let's just make it a single method in the existing class

Library/Homebrew/tap.rb
MikeMcQuaid: I'm ok with it being one method after all but let's call it `link_additional_files` or similar rather than `repair` anything

Library/Homebrew/tap.rb
MikeMcQuaid: Thanks, sorry we're a bit inconsistent.

joshka:
Just a quick heads up that I'll likely not hit this until later this week.

MikeMcQuaid:
@joshka No rush, thanks for the heads up.

MikeMcQuaid:
@joshka Any news on this?

MikeMcQuaid:
Passing on this for now but will accept when issues and conflicts are addressed, thanks!

PR #1207, opened 2016-10-01
Change `A Cask for #{token} is already installed.` message.

reitermarkus:

----

Changed the message from suggesting `brew cask install --force` to `brew cask uninstall --force`, followed by a normal `brew cask install`.

Also fixed the indentation of `CaskCommandFailedError`.

--

@Homebrew/cask

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/exceptions.rb
+++ b/Library/Homebrew/cask/lib/hbc/exceptions.rb
@@ -65,21 +65,19 @@ module Hbc
     end
 
     def to_s
-      <<-EOS
-  Command failed to execute!
-
-  ==> Failed command:
-  #{@cmd}
-
-  ==> Standard Output of failed command:
-  #{@stdout}
-
-  ==> Standard Error of failed command:
-  #{@stderr}
-
-  ==> Exit status of failed command:
-  #{@status.inspect}
-      EOS
+      s = "Command failed to execute!\n"
+      s.concat("\n")
+      s.concat("==> Failed command:\n")
+      s.concat(@cmd).concat("\n")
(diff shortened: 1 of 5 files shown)


Library/Homebrew/cask/lib/hbc/exceptions.rb
```diff
@@ -29,13 +29,30 @@ def to_s
...
+
+      s.concat("\n").concat(reinstall_message)
+    end
+
+    private
+
+    def reinstall_message
+      <<-EOS.undent
+        To re-install #{token}, run:
+          brew cask uninstall --force #{token}; brew cask install #{token}
```
DomT4: Should this be `&&` instead of `;` or is it unimportant if the first command somehow fails?

reitermarkus: You know, the more I think about this, the more undecided I am.

If `uninstall` fails, you're basically back to the beginning, as then `install` will most likely fail due to the cask still being installed.

vitorgalvao: I’d say `&&` makes more sense. If it were `brew cask uninstall --force #{token}; brew cask install --force #{token}` ( `--force` also in `install`), `;` would suffice (as a kind of “it’ll work one way or another”, but as is, we don’t want the user to be stuck in a loop of it always showing the same message (because it could not install). We want it to fail as soon as possible so they can file a bug.

reitermarkus:
@vitorgalvao, changed it to `&&`.

reitermarkus:
@vitorgalvao, could you approve this?

PR #1208, opened 2016-10-02
Refactor `puts_columns` to allow `$stderr.puts_columns`.

reitermarkus:

-----

`puts_columns` currently only outputs to `$stdout`. This change makes it an instance method of the `IO` class, so it can be used with any `IO` object.

Also added `Kernel::puts_columns` which outputs to `$stdout` by default to be in line with `Kernel::puts`.

Also modified it so it doesn't output trailing whitespace anymore.

--
Fixes https://github.com/caskroom/homebrew-cask/issues/24642.


Library/Homebrew/utils/puts_columns.rb
```diff
@@ -0,0 +1,41 @@
+require "utils/tty"
+
+module Kernel
```
MikeMcQuaid: Let's put this in another module if we move it around.

MikeMcQuaid: But I think doing that in another PR would be best as it's tricky to review as-is and compare what's changed. Best to leave it as-is in `utils.rb` for now but 👍 to moving it after that.

reitermarkus: It needs to be in `Kernel` to work with `IO` objects, i.e. `$stderr.puts_columns`.

MikeMcQuaid: I see. Personally I'd rather we had a class method that took an argument and move away from monkeypatching system classes.

reitermarkus: We could change it to only format the columns, and not output them, and let default `puts` to the rest.

```ruby
puts Formatter.columns(something)
```

MikeMcQuaid: That could work.

Library/Homebrew/utils/formatter/columns.rb
```diff
@@ -0,0 +1,46 @@
+require "utils/tty"
+
+module Formatter
```
MikeMcQuaid: Just stick this inside the existing Formatter class; I dislike objects across multiple files

Library/Homebrew/cask/spec/formatter_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "spec_helper"
+require "utils/formatter"
+require "utils/tty"
+
+describe Formatter do
+  describe "::columns" do
```
MikeMcQuaid: Can you make this part of `brew tests` instead?

MikeMcQuaid:
@reitermarkus Thoughts on this?

reitermarkus:
@MikeMcQuaid, I moved the `columns` method into the existing file but left the test as-is, because we will need to merge the test suites anyways at some point.

MikeMcQuaid:
@reitermarkus As the tests are already written and working I'll make an exception this time but definitely don't want to be adding any more Homebrew functionality tests primary in Cask tests in future.

reitermarkus:
> definitely don't want to be adding any more Homebrew functionality tests primary in Cask tests

Then I guess I now have an incentive to merge the test suites. 😄

MikeMcQuaid:
@reitermarkus Yes 😀

PR #1215, opened 2016-10-02
diagnostic: add check for missing header file

javian:

-----
In the PHP tap we see an error with the zlib header missing quite often. This is generally caused by an upgrade of the OS starting in El Capitan where SIP was introduced which seems to delete the /usr/include directory at each operating system upgrade. The PR only checks the zlib.h file since that's the one we've seen problems with but it should be indicative enough to tell if the files are missing or not. I had some problems with the wording due to that but I'm willing to take feedback and make updates =)

### DIFF
--- a/Library/Homebrew/extend/os/mac/diagnostic.rb
+++ b/Library/Homebrew/extend/os/mac/diagnostic.rb
@@ -281,6 +281,23 @@ module Homebrew
           We recommend only installing stable releases of XQuartz.
         EOS
       end
+
+      def check_for_sip_deleted_header_file_installed_by_xcode
+        # Starting in El Capitan the upgrade of macOS lead to the deletion of the
+        # /usr/include directory due to SIP.
+        return if MacOS::CLT.installed? && File.file?("/usr/include/zlib.h")
+
+        <<-EOS.undent
+          A header file installed by xcode is missing and can lead to builds failing.
+          Due to SIP (System Integrity Protection) this can occur after a macOS
+          system has been upgraded.
+
+          Please run:
+            xcode-select --install
+
+          to reinstall the header file.
+        EOS
+      end
     end
   end
 end
--- a/Library/Homebrew/extend/os/mac/diagnostic.rb
+++ b/Library/Homebrew/extend/os/mac/diagnostic.rb
@@ -281,23 +281,6 @@ module Homebrew
           We recommend only installing stable releases of XQuartz.
         EOS
       end
-
-      def check_for_sip_deleted_header_file_installed_by_xcode
-        # Starting in El Capitan the upgrade of macOS lead to the deletion of the
-        # /usr/include directory due to SIP.
-        return if MacOS::CLT.installed? && File.file?("/usr/include/zlib.h")
-
-        <<-EOS.undent
-          A header file installed by xcode is missing and can lead to builds failing.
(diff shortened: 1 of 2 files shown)


Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -281,6 +281,23 @@ def check_for_beta_xquartz
           We recommend only installing stable releases of XQuartz.
         EOS
       end
+
+      def check_for_sip_deleted_header_file_installed_by_xcode
+        # Starting in El Capitan the upgrade of macOS lead to the deletion of the
+        # /usr/include directory due to SIP.
+        return if MacOS::CLT.installed? && File.file?("/usr/include/zlib.h")
```
MikeMcQuaid: I'd suggest this check by added to the CLT/development tools installed check as if `MacOS::CLT.installed?` returns true but this header is missing then that check is bad.

javian: Why would it be bad ? I'll add it to the dev tools installed check.

MikeMcQuaid: `xcode-select --install` just installs the CLT so if that's the fix then they aren't currently installed correctly. I mean that check is currently bad because it's not checking for a working CLT.

javian:
hmm that went bad

javian:
I've made the update now assuming this is what needed to be done. I tested it but I could not make it trigger while using `brew doctor` and it didn't seem to matter what I threw at it. It seemed to return the right things but it didn't seem to have an effect on anything and I couldn't figure out why  :'(

MikeMcQuaid:
> It seemed to return the right things but it didn't seem to have an effect on anything and I couldn't figure out why :'(

`brew doctor` won't complain if you don't have the CLT installed, if that's what you're aiming to do. It will complain if you have neither the CLT or Xcode installed. With your change here it'll bypass using the CLT and use Xcode instead if possible which seems like the right thing to do.

Additionally, this may be something you want to extend all the way back to e.g. 10.9. If you like I can check the status of this file on those older versions (I have VMs).

javian:
The thing is that if you have a working system with clt and/or xcode (which I both assume install the headers in /usr/include), upgrade to Sierra from El Cap and at that point the folder is removed and the building of some Formulas (in my case, php) will fail. Brew doctor in that case won't help you to detect that your system is missing the header files. This is what I wanted to accomplish.

MikeMcQuaid:
PHP needs specifically the CLT and Xcode is not enough? Xcode doesn't install those headers but we can find the headers in the other locations using superenv.

Given what you're suggesting something like your original check may be worth having after all (sorry for asking you to change it) where you state that the CLT is installed but broken and can be fixed by running `xcode-select --install`.

MikeMcQuaid:
@javian Any thoughts?

javian:
I thought I had responded to this but it must have gotten lost in a forgotten tab somewhere. Do you think it is worthwhile to keep the check that it was changed to ? I'm not sure that would ever trigger since doctor isn't checking up on it so I'm wondering if I should revert the change or include both.

MikeMcQuaid:
@javian Let's pass on both for now but if you see more issues reported that we could debug this way let us know 👍

PR #1218, opened 2016-10-03
Formula-Cookbook: fix /usr/local/Homebrew references.

neutric:


MikeMcQuaid:
Thanks for the fix! In future a better commit message and PR title would be great.

PR #1220, opened 2016-10-03
Make `core_tap_version_string` reusable.

reitermarkus:

-----

`brew --version` and `brew cask --version` both use the same format to output the tap version, so it makes sense to have a common method for that in the `Tap` class.

@Homebrew/cask, also updated `HBC_VERSION` to `0.60.1`, since that is the actual latest tag, but that also doesn't make sense anymore, so we should probably just use `HOMEBREW_VERSION` instead.

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -14,8 +14,9 @@ $:.unshift(HOMEBREW_LIBRARY_PATH.to_s)
 require "global"
 
 if ARGV == %w[--version] || ARGV == %w[-v]
+  require "tap"
   puts "Homebrew #{HOMEBREW_VERSION}"
-  puts "Homebrew/homebrew-core #{Homebrew.core_tap_version_string}"
+  puts "Homebrew/homebrew-core #{CoreTap.instance.version_string}"
   exit 0
 end
 
--- a/Library/Homebrew/cask/lib/hbc/version.rb
+++ b/Library/Homebrew/cask/lib/hbc/version.rb
@@ -1,13 +1,10 @@
-HBC_VERSION = "0.60.0".freeze
-
 module Hbc
   def self.full_version
     @full_version ||= begin
-      revision, commit = Dir.chdir(Hbc.default_tap.path) do
-        [`git rev-parse --short=4 --verify -q HEAD 2>/dev/null`.chomp,
-         `git show -s --format="%cr" HEAD 2>/dev/null`.chomp]
-      end
-      "#{HBC_VERSION} (git revision #{revision}; last commit #{commit})"
+      <<-EOS.undent
+        Homebrew-Cask #{HOMEBREW_VERSION}
+        caskroom/homebrew-cask #{Hbc.default_tap.version_string}
+      EOS
     end
   end
 end
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -138,6 +138,13 @@ class Tap
     name
   end
 
+  def version_string
+    return "N/A" unless installed?
+    pretty_revision = git_short_head
+    return "(no git repository)" unless pretty_revision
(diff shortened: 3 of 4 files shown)


vitorgalvao:
> @Homebrew/cask, also updated `HBC_VERSION` to `0.60.1`, since that is the actual latest tag, but that also doesn't make sense anymore, so we should probably just use `HOMEBREW_VERSION` instead.

Fine by me.

Library/Homebrew/brew.rb
```diff
@@ -15,7 +15,7 @@
 
 if ARGV == %w[--version] || ARGV == %w[-v]
   puts "Homebrew #{HOMEBREW_VERSION}"
-  puts "Homebrew/homebrew-core #{Homebrew.core_tap_version_string}"
+  puts "Homebrew/homebrew-core #{CoreTap.instance.version_string}"
```
jawshooah: Looks like `CoreTap` hasn't been initialized yet at this point: https://travis-ci.org/Homebrew/brew/builds/164644934#L518-L520

Library/Homebrew/cask/lib/hbc/version.rb
```diff
@@ -1,13 +1,10 @@
-HBC_VERSION = "0.60.0".freeze
+HBC_VERSION = "0.60.1".freeze
```
MikeMcQuaid: Is it worth maintaining a separate version for the two projects? Longer-term at least this should go away.

MikeMcQuaid: I see this was answered to 👍 to using Homebrew's version.

MikeMcQuaid:
Tests are failing: https://bot.brew.sh/job/Brew%20Pull%20Requests/2263/version=yosemite/testReport/junit/brew-test-bot/yosemite/tests___no_compat/

Otherwise 👍 to the general approach.

reitermarkus:
Should be ready to merge now.

Library/Homebrew/brew.rb
```diff
@@ -14,8 +14,9 @@
 require "global"
 
 if ARGV == %w[--version] || ARGV == %w[-v]
+  require "tap"
```
MikeMcQuaid: Maybe consider just requiring that in `global`?

Library/Homebrew/cask/lib/hbc/version.rb
```diff
@@ -1,13 +1,8 @@
...
 module Hbc
   def self.full_version
     @full_version ||= begin
-      revision, commit = Dir.chdir(Hbc.default_tap.path) do
-        [`git rev-parse --short=4 --verify -q HEAD 2>/dev/null`.chomp,
-         `git show -s --format="%cr" HEAD 2>/dev/null`.chomp]
-      end
-      "#{HBC_VERSION} (git revision #{revision}; last commit #{commit})"
+      "Homebrew-Cask #{HOMEBREW_VERSION}\n" \
+      "caskroom/homebrew-cask #{Hbc.default_tap.version_string}"
```
MikeMcQuaid: Favour `EOS.undent` over embedded `\n`.

Library/Homebrew/tap.rb
```diff
@@ -138,6 +138,16 @@ def to_s
     name
   end
 
+  def version_string
+    return "N/A" unless installed?
+    if pretty_revision = git_short_head
```
MikeMcQuaid: Could always make it more readable with:
```
pretty_revision = git_short_head
return "(no git repository)" unless pretty_revision
...
```

Library/Homebrew/brew.rb
reitermarkus: Let's do that in a separate PR, as `tap` is `require`d in many other places.

MikeMcQuaid: 🆒 👍

PR #1225, opened 2016-10-04
Immortalize virtualenvs better

tdsmith:
virtualenvs remember the prefix of the python that created them and do
magic to load packages from them; help them remember a more durable
path.

### DIFF
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -215,6 +215,13 @@ module Language
             f.unlink
             f.make_symlink new_target
           end
+
+          Pathname.glob(@venv_root/"lib/python*/orig-prefix.txt").each do |prefix_file|
+            prefix_path = prefix_file.read
+            python = prefix_path.include?("python3") ? "python3" : "python"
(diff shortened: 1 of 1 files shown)


tdsmith:
Ideally this lands before #5562.

Library/Homebrew/language/python.rb
```diff
@@ -215,6 +215,14 @@ def create
             f.unlink
             f.make_symlink new_target
           end
+
+          Pathname.glob(@venv_root/"lib/python*/orig-prefix.txt").each do |prefix_file|
+            prefix_path = prefix_file.read
+            next unless prefix_path.start_with? HOMEBREW_CELLAR
+            python = prefix_path.include?("python3") ? "python3" : "python"
+            prefix_path.sub! %r{#{HOMEBREW_CELLAR}/#{python}/[^/]+}, Formula[python].opt_prefix
```
ilovezfs: is this going to break if the requirement is fulfilled by non-brew python/python3?

tdsmith: No; `String#sub!` doesn't error if it doesn't find a match.

tdsmith:
Squashed commits.

PR #1227, opened 2016-10-05
apr_requirement: use Homebrew's Apr for Sierra

DomT4:

-----

Sierra ships the headers/libraries still but for some reason decided to bin the config scripts, which whilst seemingly not an issue for `mesos` or `ganglia` it has broken `subversion`, `log4cxx`, `ctail`, `shibboleth` and `passenger` that we know of so far. Let's assume more often than not things are going to break without those config scripts around.

### DIFF
--- a/Library/Homebrew/requirements/apr_requirement.rb
+++ b/Library/Homebrew/requirements/apr_requirement.rb
@@ -4,16 +4,17 @@ class AprRequirement < Requirement
   fatal true
   default_formula "apr-util"
 
-  # APR shipped in Tiger is too old, but Leopard+ is usable
-  satisfy(build_env: false) { MacOS.version > :leopard && MacOS::CLT.installed? }
+  # APR shipped in Tiger is too old, but Leopard+ is usable.
+  # The *-config scripts were removed in Sierra, which is widely breaking.
+  satisfy(build_env: false) do
+    MacOS.version > :leopard && MacOS.version < :sierra && MacOS::CLT.installed?
+  end
 
   env do
-    unless MacOS::CLT.installed?
-      ENV.prepend_path "PATH", Formula["apr-util"].opt_bin
-      ENV.prepend_path "PATH", Formula["apr"].opt_bin
-      ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["apr"].opt_libexec}/lib/pkgconfig"
-      ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["apr-util"].opt_libexec}/lib/pkgconfig"
-    end
+    ENV.prepend_path "PATH", Formula["apr-util"].opt_bin
+    ENV.prepend_path "PATH", Formula["apr"].opt_bin
+    ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["apr"].opt_libexec}/lib/pkgconfig"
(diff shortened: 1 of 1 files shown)


DomT4:
Travis failure looks unrelated.

Library/Homebrew/requirements/apr_requirement.rb
```diff
@@ -4,11 +4,16 @@ class AprRequirement < Requirement
   fatal true
   default_formula "apr-util"
 
-  # APR shipped in Tiger is too old, but Leopard+ is usable
-  satisfy(build_env: false) { MacOS.version > :leopard && MacOS::CLT.installed? }
+  # APR shipped in Tiger is too old, but Leopard+ is usable.
+  # The *-config scripts were removed in Sierra, which is widely breaking.
+  satisfy(build_env: false) do
+    MacOS.version > :leopard && MacOS.version < :sierra &&
```
MikeMcQuaid: Use `next if MacOS.version > :leopard` etc. to split this into multiple statements.

Library/Homebrew/requirements/apr_requirement.rb
```diff
@@ -4,11 +4,16 @@ class AprRequirement < Requirement
...
+      MacOS::CLT.installed? || Formula["apr-util"].installed?
+  end
 
   env do
-    unless MacOS::CLT.installed?
+    # Prefer system Apr as much as possible, even if our's is installed.
+    unless MacOS.version > :leopard && MacOS.version < :sierra && MacOS::CLT.installed?
```
MikeMcQuaid: Similarly with the `next if` approach here.

AnimiVulpis:
FYI: For me it prevents me from upgrading `git` (I could most probably install `apr` on my own, but I still wanted to point it out)
`OS: macOS Sierra 10.12 16A323 x86_64`


```
git 2.10.1
==> Upgrading git
==> Using the sandbox
Error: /usr/local/opt/apr not present or broken
Please reinstall apr. Sorry :(
```

ilovezfs:
@AnimiVulpis you need to reinstall subversion.

Library/Homebrew/requirements/apr_requirement.rb
MikeMcQuaid: Sorry, it would be `next if MacOS.version < :leopard` and `next if MacOS.version > :el_capitan`

Library/Homebrew/requirements/apr_requirement.rb
```diff
@@ -4,16 +4,22 @@ class AprRequirement < Requirement
   fatal true
   default_formula "apr-util"
 
-  # APR shipped in Tiger is too old, but Leopard+ is usable
-  satisfy(build_env: false) { MacOS.version > :leopard && MacOS::CLT.installed? }
+  # APR shipped in Tiger is too old, but Leopard+ is usable.
+  # The *-config scripts were removed in Sierra, which is widely breaking.
+  satisfy(build_env: false) do
+    next false if MacOS.version <= :leopard
```
ilovezfs: what does `next false if` do?

MikeMcQuaid: Leaves the current block and return the value `false`. `next` without `false` will return `nil` which is falsy but better to be more explicit.

ilovezfs: reads like gibberish but ok ...

PR #1228, opened 2016-10-05
audit: forbid `deprecated_option` in new formulae

jonchang:

-----

It doesn't make sense to include `deprecated_options` in a new formula. This PR adds an check on `brew audit --new formula` to restrict this. I didn't add a test because it seems like there weren't any tests for options; let me know if there's a place that the test belongs.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -448,6 +448,11 @@ class FormulaAuditor
         problem "Use '--with#{$1}-test' instead of '--#{o.name}'. Migrate '--#{o.name}' with `deprecated_option`."
       end
     end
+
+    return unless @new_formula
+    unless formula.deprecated_options.empty?
+      problem "New formulae should not use `deprecated_option`."
+    end
   end
 
   def audit_desc


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -448,6 +448,11 @@ def audit_options
         problem "Use '--with#{$1}-test' instead of '--#{o.name}'. Migrate '--#{o.name}' with `deprecated_option`."
       end
     end
+
+    return unless @new_formula
+    if formula.deprecated_options.size
```
vladshablinsky: `formula.deprecated_options.size` is always true.
Probably this is better: `unless formula.deprecated_options.empty?`

jonchang:
This is mainly for the CI builders, which run brew audit --new-formula for new formulae and --strict --online for existing formulae. Usually --new-formula implies both strict and online, but this change would only apply for new formulae.

vladshablinsky:
@jonchang yes, I see. There shouldn't be `deprecated_option` in new formula. But I wonder if it's likely to happen. This is my point.

jonchang:
Ok I just I just encountered this in https://github.com/Homebrew/homebrew-science/pull/4407 and was surprised that it wasn't already an audit. I don't think I've seen new formula with this before though.

vladshablinsky:
> Ok I just I just encountered this in Homebrew/homebrew-science#4407 and was surprised that it wasn't already an audit

Ah, then this makes sense! Thanks for the link.

bfontaine:
How does this work with formulae that are moved between taps?

MikeMcQuaid:
@bfontaine Similarly to other audit checks we have on new tap formulae: we'd require them to be cleaned up.

MikeMcQuaid:

PR #1235, opened 2016-10-06
Deprecate license

abookyun:

-----

> @reitermarkus: Every cask has a license stanza, which is, and never was, used for anything.

So, let's deprecate all license stanza.

discussions from here: https://github.com/caskroom/homebrew-cask/issues/16692 and https://github.com/caskroom/homebrew-cask/issues/17437


reitermarkus:
cc @jawshooah, @vitorgalvao

Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -206,16 +204,6 @@ def sha256(arg = nil)
       @sha256 ||= arg
     end
 
-    def license(arg = nil)
```
jawshooah: We don't want to remove the method entirely, as that would break Casks that still have a `license` stanza. Just replace the method body with `odeprecated "Hbc::DSL#license"`

abookyun: Something went wrong..😑

jawshooah: Looks like this breaks the `brew cask list` call in `test/test_cask.rb` because `odeprecated` throws an error when running as `HOMEBREW_DEVELOPER`. In `test_cask.rb`, change `cmd("cask", "list")`, to `cmd("cask", "list", "--caskroom=#{HOMEBREW_PREFIX}/Caskroom")`

abookyun: @jawshooah Fixed the IntegrationCommandTestCask failure, now here is File leak detected. 🤔 Trying dig into it.

abookyun: @jawshooah fixed it by adding `#{HOMEBREW_PREFIX}/Caskroom` into `paths_to_delete`. Please have a look make sure it's okay. thanks!

jawshooah:
@reitermarkus @vitorgalvao Any objections to merging this?

Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -207,13 +205,7 @@ def sha256(arg = nil)
     end
 
     def license(arg = nil)
```
reitermarkus: `arg` is unused, so change this to 

```ruby
def license(*)
```

vitorgalvao:
No objections. So much red. So good to see.

Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -207,13 +205,7 @@ def sha256(arg = nil)
...
-      return @license if arg.nil?
-      assert_only_one_stanza_allowed :license, !arg.nil?
-      @license ||= begin
-        DSL::License.new(arg) unless arg.nil?
-      rescue StandardError => e
-        raise CaskInvalidError.new(token, e)
-      end
+      odeprecated "Hbc::DSL#license"
```
reitermarkus: Let's comment this out for now until `license` has been removed from all official taps.

vitorgalvao: @reitermarkus After this is merged and tagged, I don’t foresee even a day passing before `license` is removed from all official taps.

reitermarkus: Ok then, let's merge this.

Library/Homebrew/cask/lib/hbc/dsl.rb
abookyun: ok!

Library/Homebrew/cask/lib/hbc/dsl.rb
abookyun: Append another commit, and let's merge it. Thanks guys! 👍

jawshooah:
@BrewTestBot Test this please

abookyun:
Yeah! @jawshooah thanks!

reitermarkus:
Thanks, @abookyun!

reitermarkus:
@MikeMcQuaid, would be great to get a tagged release with this soonish.

MikeMcQuaid:
Will likely do one on Monday. That 🆒?

toonetown:
Before tagging it on monday, can we remove the `license:` stanza in all the current casks (from https://github.com/caskroom/homebrew-cask/issues/17437)?

vitorgalvao:
@toonetown No, because then every cask will break. We need to do it concurrently.

We have already two or three reports of users getting the deprecation warning, which means they have set at some point homebrew developer flags (confirmed at least once). So far not as bad as it used to be, but it’s an awful throwback to the days when we used to need tags, before 1.0. We need to find a way to also include in error messages that the error the user is seeing might be do to that, and that they should remove whatever option they have set.

@abookyun Mind also updating the documentation on https://github.com/caskroom/homebrew-cask?

@reitermarkus @jawshooah @reitermarkus There’s no point in having duplicate documentation (and it’s already out of sync). Any objections to removing all HBC documentation from this repo and keeping it in the old repo?

MikeMcQuaid:
> We need to find a way to also include in error messages that the error the user is seeing might be do to that, and that they should remove whatever option they have set.

@vitorgalvao We can't do that as it'll be reset again if they want to contribute to Homebrew. 

Too late now, I realise, but the few step approach I'd personally take when doing this:

- make the license field optional
- remove the license from all core casks but don't print deprecation warning
- deprecate the license DSL and print messages warning about it
- after a few months: remove the license DSL so things break if it's still used

Library/Homebrew/cask/lib/hbc/dsl.rb
MikeMcQuaid: I'd suggest commenting this out again for now.

vitorgalvao:
@MikeMcQuaid Like I said, this is only a few users, so I don’t see it as a huge issue. My concern was for future cases.

> We can't do that as it'll be reset again if they want to contribute to Homebrew.

To clarify, what I meant was just something in the lines of: something errors out, it spits out the error but if it also detects the user is on a non-tagged version it tells them to do so and try again before reporting issues. As in, it’s no problem if it resets, and it should indeed reset, but when something errors-out, they should be aware that might the cause of the error.

If it isn’t done in the core itself, we can at least add it to our bug reporting template.

MikeMcQuaid:
@vitorgalvao That's a fairly unpleasant user experience. Might be better to write up a document on how to deprecate a DSL element and suggest similar to the above steps to avoid issues with other users.

reitermarkus:
> Any objections to removing all HBC documentation from this repo and keeping it in the old repo?

I thought I already removed it when I moved it back to `caskroom/cask`, so no objections.

vitorgalvao:
@MikeMcQuaid Alright, fair enough.

toonetown:
> @vitorgalvao 
> @toonetown No, because then every cask will break. We need to do it concurrently.

What do you mean "every cask will break"?  Isn't `license` optional?  Removing it via a script in every cask shouldn't break anything.  However, this change is now causing deprecation warnings to be spit out every time I run, for example `brew cask info <<cask>>`.

> @vitorgalvao 
> We have already two or three reports of users getting the deprecation warning, which means they have set at some point homebrew developer flags (confirmed at least once).

This might be my issue...what is meant by "setting developer flags"?  How can I check if I have set those or not (I don't think that I have).  The only flag I think I have set is the `HOMEBREW_GITHUB_API_TOKEN` environment variable.

vitorgalvao:
> Isn't `license` optional?

If it were, we would have gotten rid of it long ago. It’s mandatory even to pass `audit`.

As for the flag, search for `HOMEBREW_DEVELOPER`. I think the man page explains it (on a phone, now, can’t check).

reitermarkus:
@toonetown, if you only get a `Warning`, then you don't have the developer flag set. If you had, you would get an `Error`, and `brew` would exit immediately.

vitorgalvao:
> That's a fairly unpleasant user experience.

@MikeMcQuaid Thought a bit more about that, and I completely agree. So much so, I’m not even thinking of adding that to the bug reporting template anymore.

PR #1237, opened 2016-10-07
migrate-formula-pr: new command

ilovezfs:


MikeMcQuaid:
I love the idea here but I'm concerned we'd have three `dev-cmd/*-formula-pr.rb` with pretty similar code. Any chance you could get some code sharing going on between them?

zmwangx:
`boneyard-formula-pr` is a special case of `migrate-formula-pr` as far as I can tell.

ilovezfs:
@zmwangx yup, it can be replaced with something that runs `brew migrate-formula-pr --tap=boneyard`, or an equivalent api.

MikeMcQuaid:
@ilovezfs Any news here? Given we're trying to pull things from taps rather than the other way around will this still be useful?

ilovezfs:
It would be used as
```
brew migrate-formula-pr --tap=core foo
```
so the question would be whether those migrations are likely to be done in bulk, or one at a time.

MikeMcQuaid:
@ilovezfs Probably a couple at a time until we're confident then en masse.

MikeMcQuaid:
Let's pass on this given we're migrating things in here rather than out there now.

PR #1245, opened 2016-10-08
Fix getting cask path for queries.

reitermarkus:

-----

This fixes queries that result in `join` being called on `NilClass` when a tap's `cask_dir` doesn't exist.

Also moved the actual loading of cask files to a separate class `CaskLoader`.

--

@Homebrew/cask 

--
Fixes https://github.com/caskroom/homebrew-cask/issues/24813.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cask_loader.rb
+++ b/Library/Homebrew/cask/lib/hbc/cask_loader.rb
@@ -6,14 +6,7 @@ module Hbc
       raise CaskError, "File '#{path}' is not a plain file" unless path.file?
 
       token = path.basename(".rb").to_s
-
-      content = File.open(path, "rb") do |caskfile|
-        if defined?(Encoding)
-          caskfile.read.force_encoding("UTF-8")
-        else
-          caskfile.read
-        end
-      end
+      content = IO.read(path).force_encoding("UTF-8")
 
       new(token, content, path).load
     end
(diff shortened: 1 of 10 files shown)


Library/Homebrew/cask/lib/hbc/cask_loader.rb
```diff
@@ -0,0 +1,61 @@
...
+  class CaskLoader
+    def self.load_from_file(path)
+      raise CaskError, "File '#{path}' does not exist"      unless path.exist?
+      raise CaskError, "File '#{path}' is not readable"     unless path.readable?
+      raise CaskError, "File '#{path}' is not a plain file" unless path.file?
+
+      token = path.basename(".rb").to_s
+
+      content = File.open(path, "rb") do |caskfile|
+        if defined?(Encoding)
```
jawshooah: We don't support Ruby < 2.0 anymore, so this check can probably go.

Library/Homebrew/cask/lib/hbc/cask_loader.rb
```diff
@@ -6,14 +6,7 @@ def self.load_from_file(path)
...
       token = path.basename(".rb").to_s
-
-      content = File.open(path, "rb") do |caskfile|
-        if defined?(Encoding)
-          caskfile.read.force_encoding("UTF-8")
-        else
-          caskfile.read
-        end
-      end
+      content = File.open(path, "rb", &:read).force_encoding("UTF-8")
```
jawshooah: One more nitpick: `IO.read(path)` rather than `File.open(path, "rb", &:read)`

PR #1252, opened 2016-10-09
FIX: cask changes /usr/local ownership recursively

niw:
**Problems**

Relatively old code in `Hbc::Caskroom` recursively changes the ownership
of the directory where the `Caskroom` directory exists, that changes
entire files in `/usr/local` to `user:staff` if Homebrew setup with default
configuration.

This is really dangerous because it's easy to trigger (just simply type
`brew cask something` by following some installation documentation.)

**Solution**

This patch removes entire `chown` with `-R` option and make the logic
simply creating `Caskroom` directory with default Homebrew directories
ownership and permission.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/caskroom.rb
+++ b/Library/Homebrew/cask/lib/hbc/caskroom.rb
@@ -13,7 +13,7 @@ module Hbc
         FileUtils.mv repo_caskroom, Hbc.caskroom
       else
         opoo "#{Hbc.caskroom.parent} is not writable, sudo is needed to move the Caskroom."
-        sudo "/bin/mv", repo_caskroom.to_s, Hbc.caskroom.parent.to_s
+        command "/bin/mv", repo_caskroom, Hbc.caskroom.parent, :use_sudo => true
       end
     end
 
@@ -22,16 +22,23 @@ module Hbc
 
       ohai "Creating Caskroom at #{Hbc.caskroom}"
       ohai "We'll set permissions properly so we won't need sudo in the future"
+      use_sudo = !Hbc.caskroom.parent.writable?
 
-      sudo "/bin/mkdir", "-p", Hbc.caskroom
-      sudo "/bin/chmod", "g+rwx", Hbc.caskroom
-      sudo "/usr/sbin/chown", Utils.current_user, Hbc.caskroom
-      sudo "/usr/bin/chgrp", "admin", Hbc.caskroom
+      command "/bin/mkdir", "-p", Hbc.caskroom, :use_sudo => use_sudo
+      command "/bin/chmod", "g+rwx", Hbc.caskroom, :use_sudo => use_sudo
+      command "/usr/sbin/chown", Utils.current_user, Hbc.caskroom, :use_sudo => use_sudo
+      command "/usr/bin/chgrp", "admin", Hbc.caskroom, :use_sudo => use_sudo
     end
 
-    def sudo(*args)
-      ohai "/usr/bin/sudo #{args.join(" ")}"
-      system "/usr/bin/sudo", *args
+    def command(*args)
+      options = args.last.is_a?(Hash) ? args.pop : {}
+
+      if options[:use_sudo]
+        args.unshift "/usr/bin/sudo"
+      end
+
+      ohai args.join(" ")
+      system *args
     end
   end
 end
--- a/Library/Homebrew/cask/lib/hbc/caskroom.rb
+++ b/Library/Homebrew/cask/lib/hbc/caskroom.rb
@@ -13,7 +13,7 @@ module Hbc
         FileUtils.mv repo_caskroom, Hbc.caskroom
       else
         opoo "#{Hbc.caskroom.parent} is not writable, sudo is needed to move the Caskroom."
-        SystemCommand.run("/bin/mv", :args => [repo_caskroom, Hbc.caskroom.parent], :sudo => true)
+        SystemCommand.run("/bin/mv", args: [repo_caskroom, Hbc.caskroom.parent], sudo: true)
       end
     end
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
👍 to not changing permissions of `toplevel_dir`.

Library/Homebrew/cask/lib/hbc/caskroom.rb
```diff
@@ -13,32 +13,25 @@ def migrate_caskroom_from_repo_to_prefix
...
-        unless Hbc.caskroom.parent == toplevel_dir
-          system "/usr/bin/sudo", "--", "/usr/sbin/chown", "-R", "--", "#{Utils.current_user}:staff", Hbc.caskroom.parent.to_s
-        end
-      end
+      ohai "We'll set permissions properly so we won't need sudo in the future"
+
+      sudo "/bin/mkdir", "-p", Hbc.caskroom
+      sudo "/bin/chmod", "g+rwx", Hbc.caskroom
+      sudo "/usr/sbin/chown", Utils.current_user, Hbc.caskroom
+      sudo "/usr/bin/chgrp", "admin", Hbc.caskroom
```
reitermarkus: Not sure if this should be `admin` or `wheel`.

Library/Homebrew/cask/lib/hbc/caskroom.rb
```diff
@@ -13,32 +13,25 @@ def migrate_caskroom_from_repo_to_prefix
...
-        system "/usr/bin/sudo", "--", "/bin/mv", "--", repo_caskroom.to_s, Hbc.caskroom.parent.to_s
+        sudo "/bin/mv", repo_caskroom.to_s, Hbc.caskroom.parent.to_s
       end
     end
 
     def ensure_caskroom_exists
       return if Hbc.caskroom.exist?
 
       ohai "Creating Caskroom at #{Hbc.caskroom}"
-      if Hbc.caskroom.parent.writable?
```
reitermarkus: This should still be checked. No need to use `sudo` unless necessary.

Library/Homebrew/cask/lib/hbc/caskroom.rb
MikeMcQuaid: Homebrew installer sets `admin`: https://github.com/Homebrew/install/blob/8dcb739d4356f16079172dc9a5aa66a8a36f9860/install#L196-L204

Library/Homebrew/cask/lib/hbc/caskroom.rb
niw: True, just I was cheating. Let me rework on a patch.

MikeMcQuaid:
@reitermarkus How's this looking?

Library/Homebrew/cask/lib/hbc/caskroom.rb
```diff
@@ -13,32 +13,32 @@ def migrate_caskroom_from_repo_to_prefix
...
+      ohai "We'll set permissions properly so we won't need sudo in the future"
+      use_sudo = !Hbc.caskroom.parent.writable?
+
+      command "/bin/mkdir", "-p", Hbc.caskroom, :use_sudo => use_sudo
+      command "/bin/chmod", "g+rwx", Hbc.caskroom, :use_sudo => use_sudo
+      command "/usr/sbin/chown", Utils.current_user, Hbc.caskroom, :use_sudo => use_sudo
+      command "/usr/bin/chgrp", "admin", Hbc.caskroom, :use_sudo => use_sudo
+    end
+
+    def command(*args)
```
reitermarkus: Use `SystemCommand` instead of defining another wrapper for `system`.

```ruby
SystemCommand.run("/bin/mkdir", args: ["-p", Hbc.caskroom], sudo: use_sudo)
```

Library/Homebrew/cask/lib/hbc/caskroom.rb
```diff
@@ -13,32 +13,32 @@ def migrate_caskroom_from_repo_to_prefix
...
-        unless Hbc.caskroom.parent == toplevel_dir
-          system "/usr/bin/sudo", "--", "/usr/sbin/chown", "-R", "--", "#{Utils.current_user}:staff", Hbc.caskroom.parent.to_s
-        end
+      ohai "We'll set permissions properly so we won't need sudo in the future"
+      use_sudo = !Hbc.caskroom.parent.writable?
+
+      command "/bin/mkdir", "-p", Hbc.caskroom, :use_sudo => use_sudo
+      command "/bin/chmod", "g+rwx", Hbc.caskroom, :use_sudo => use_sudo
+      command "/usr/sbin/chown", Utils.current_user, Hbc.caskroom, :use_sudo => use_sudo
+      command "/usr/bin/chgrp", "admin", Hbc.caskroom, :use_sudo => use_sudo
```
reitermarkus: Not sure if we should set the group to `admin`, even when `sudo` is not needed. I think we can leave it at:

```ruby
if Hbc.caskroom.parent.writable?
  Hbc.caskroom.mkpath
else
  # stuff that needs sudo
end
```

niw: I think, if we can create a new directory, e.g. `/usr/local` has logged in owner ship, if it's newly create directory, following what the other Homebrew directory does might be better?
Setting `admin` to the directory is not because of using `sudo` but have a consistency with how Homebrew manages these directories. Thought?

Library/Homebrew/cask/lib/hbc/caskroom.rb
niw: Ah, there is that. Sure!

Library/Homebrew/cask/lib/hbc/caskroom.rb
```diff
@@ -13,32 +13,21 @@ def migrate_caskroom_from_repo_to_prefix
...
-        # sudo in system is rude.
-        system "/usr/bin/sudo", "--", "/bin/mkdir", "-p", "--", Hbc.caskroom
-        unless Hbc.caskroom.parent == toplevel_dir
-          system "/usr/bin/sudo", "--", "/usr/sbin/chown", "-R", "--", "#{Utils.current_user}:staff", Hbc.caskroom.parent.to_s
-        end
-      end
+      ohai "We'll set permissions properly so we won't need sudo in the future"
+      sudo = !Hbc.caskroom.parent.writable?
+
+      SystemCommand.run("/bin/mkdir", args: ["-p", Hbc.caskroom], sudo: sudo)
```
vitorgalvao: I’s a bit unclear to me why `sudo: true` above and `sudo: sudo` here. What’s the difference between the two?

niw: We can use similar code (switching `sudo` by variable) in `migrate_caskroom_from_repo_to_prefix` as well, is it preferable?

jawshooah:
Thanks @niw!

Issue #1250, opened 2016-10-09
Keg.relocate_text_files performance is awful


zmwangx:
`Keg.relocate_text_files` reads every single text file and libtool file in a bottle, replaces the cellar, prefix and repository placeholders (if any), and rewrites the files that need modifications. This process is single-handedly inflating bottle pouring time 10–100 times. A few examples I tested:

- `git`: untar ~0.3s, relocate text files ~6s;
- `go`: untar ~1.6s, relocate text files ~70s;
- `mysql`: untar ~2.7s, relocate text files ~170s.

MikeMcQuaid:
@zmwangx Thanks for filing this!

jawshooah:
Looks like we're currently doing [three separate passes](https://github.com/Homebrew/brew/blob/e851c9bf6c3adda268f1a307d6b49acb1ad1852d/Library/Homebrew/keg_relocate.rb#L28-L30) over each file when we could just do one with multiple replacements. Not an asymptotic improvement, but ~3x isn't bad.

jawshooah:
Scratch that; looks like the real bottleneck is in `text_files`, particularly [running `/usr/bin/file` on every file and parsing the output](https://github.com/Homebrew/brew/blob/e851c9bf6c3adda268f1a307d6b49acb1ad1852d/Library/Homebrew/keg_relocate.rb#L78). After some quick googling, I can't find a better way to determine whether a file is text or binary without bringing in external dependencies like [ruby-filemagic](https://rubygems.org/gems/ruby-filemagic/).

MikeMcQuaid:
A few options:
- do that in parallel to try and speed it up
- cache the files we actually rewrite in `bottle.rb` and store them with inside the bottle

PR #1253, opened 2016-10-10
keg_relocate: cache files rewritten during brew bottle

jawshooah:

-----

`brew bottle` replaces instances of the Homebrew prefix, cellar, and repository with placeholders in all text files. Cache these files in INSTALL_RECEIPT.json so that we don't have to check every single text file for placeholders on install.

Also add a couple of other optimizations to speed up building bottles.

Fixes https://github.com/Homebrew/brew/issues/1250.

### DIFF
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -187,12 +187,13 @@ module Homebrew
 
     keg.lock do
       original_tab = nil
+      changed_files = nil
 
       begin
         unless ARGV.include? "--skip-relocation"
           keg.relocate_dynamic_linkage prefix, Keg::PREFIX_PLACEHOLDER,
             cellar, Keg::CELLAR_PLACEHOLDER
-          keg.relocate_text_files prefix, Keg::PREFIX_PLACEHOLDER,
+          changed_files = keg.relocate_text_files prefix, Keg::PREFIX_PLACEHOLDER,
             cellar, Keg::CELLAR_PLACEHOLDER,
             repository, Keg::REPOSITORY_PLACEHOLDER
         end
@@ -205,6 +206,7 @@ module Homebrew
         tab.poured_from_bottle = false
         tab.HEAD = nil
         tab.time = nil
+        tab.changed_files = changed_files
         tab.write
 
         keg.find do |file|
@@ -268,7 +270,7 @@ module Homebrew
               Keg::CELLAR_PLACEHOLDER, cellar
             keg.relocate_text_files Keg::PREFIX_PLACEHOLDER, prefix,
               Keg::CELLAR_PLACEHOLDER, cellar,
-              Keg::REPOSITORY_PLACEHOLDER, repository
+              Keg::REPOSITORY_PLACEHOLDER, repository, changed_files
           end
         end
       end
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -191,11 +191,7 @@ module Homebrew
 
       begin
         unless ARGV.include? "--skip-relocation"
-          keg.relocate_dynamic_linkage prefix, Keg::PREFIX_PLACEHOLDER,
-            cellar, Keg::CELLAR_PLACEHOLDER
-          changed_files = keg.relocate_text_files prefix, Keg::PREFIX_PLACEHOLDER,
-            cellar, Keg::CELLAR_PLACEHOLDER,
-            repository, Keg::REPOSITORY_PLACEHOLDER
+          changed_files = keg.replace_locations_with_placeholders
         end
 
         keg.delete_pyc_files!
@@ -266,11 +262,7 @@ module Homebrew
         ignore_interrupts do
           original_tab.write if original_tab
           unless ARGV.include? "--skip-relocation"
-            keg.relocate_dynamic_linkage Keg::PREFIX_PLACEHOLDER, prefix,
-              Keg::CELLAR_PLACEHOLDER, cellar
-            keg.relocate_text_files Keg::PREFIX_PLACEHOLDER, prefix,
-              Keg::CELLAR_PLACEHOLDER, cellar,
-              Keg::REPOSITORY_PLACEHOLDER, repository, changed_files
+            keg.replace_placeholders_with_locations changed_files
           end
         end
       end
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -762,18 +762,12 @@ class FormulaInstaller
     end
 
     keg = Keg.new(formula.prefix)
-    tab_file = formula.prefix.join(Tab::FILENAME)
-    # Skip the cache since the receipt will be rewritten
-    orig_tab = Tab.from_file_content(tab_file.read, tab_file)
-    changed_files = orig_tab.changed_files.map { |f| formula.prefix.join(f) }
 
     unless formula.bottle_specification.skip_relocation?
-      keg.relocate_dynamic_linkage Keg::PREFIX_PLACEHOLDER, HOMEBREW_PREFIX.to_s,
-        Keg::CELLAR_PLACEHOLDER, HOMEBREW_CELLAR.to_s
+      tab = Tab.for_keg(keg)
+      Tab.clear_cache
+      keg.replace_placeholders_with_locations tab.changed_files
     end
-    keg.relocate_text_files Keg::PREFIX_PLACEHOLDER, HOMEBREW_PREFIX.to_s,
-      Keg::CELLAR_PLACEHOLDER, HOMEBREW_CELLAR.to_s,
-      Keg::REPOSITORY_PLACEHOLDER, HOMEBREW_REPOSITORY.to_s, changed_files
 
     Pathname.glob("#{formula.bottle_prefix}/{etc,var}/**/*") do |path|
       path.extend(InstallRenamed)
@@ -781,7 +775,7 @@ class FormulaInstaller
     end
     FileUtils.rm_rf formula.bottle_prefix
 
-    tab = Tab.for_keg(formula.prefix)
+    tab = Tab.for_keg(keg)
 
     CxxStdlib.check_compatibility(
       formula, formula.recursive_dependencies,
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -762,13 +762,18 @@ class FormulaInstaller
     end
 
     keg = Keg.new(formula.prefix)
+    tab_file = formula.prefix.join(Tab::FILENAME)
+    # Skip the cache since the receipt will be rewritten
+    orig_tab = Tab.from_file_content(tab_file.read, tab_file)
+    changed_files = orig_tab.changed_files.map { |f| formula.prefix.join(f) }
+
     unless formula.bottle_specification.skip_relocation?
       keg.relocate_dynamic_linkage Keg::PREFIX_PLACEHOLDER, HOMEBREW_PREFIX.to_s,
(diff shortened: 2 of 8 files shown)


jawshooah:
@zmwangx This cuts the install time for `mysql` on my machine down from ~100s to ~10s.

Library/Homebrew/formula_installer.rb
```diff
@@ -777,13 +777,18 @@ def pour
     end
 
     keg = Keg.new(formula.prefix)
+    tab_file = formula.prefix.join(Tab::FILENAME)
+    # Skip the cache since the receipt will be rewritten
```
MikeMcQuaid: Not sure I understand this comment, can you elaborate a bit?

Library/Homebrew/keg_relocate.rb
```diff
@@ -75,8 +78,8 @@ def text_files
       path.find do |pn|
         next if pn.symlink? || pn.directory?
         next if Metafiles::EXTENSIONS.include? pn.extname
-        if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
-           pn.text_executable?
+        if pn.text_executable? ||
```
MikeMcQuaid: Good idea on flipping these 👍

Library/Homebrew/metafiles.rb
```diff
@@ -1,13 +1,15 @@
+require "set"
+
 class Metafiles
   # https://github.com/github/markup#markups
   EXTENSIONS = %w[
     .adoc .asc .asciidoc .creole .html .markdown .md .mdown .mediawiki .mkdn
     .org .pod .rdoc .rst .rtf .textile .txt .wiki
-  ].freeze
+  ].to_set.freeze
```
MikeMcQuaid: If this is a frozen array I'm not sure I see the advantages of making it a set?

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -205,6 +206,7 @@ def bottle_formula(f)
         tab.poured_from_bottle = false
         tab.HEAD = nil
         tab.time = nil
+        tab.changed_files = changed_files
```
MikeMcQuaid: My only concern with putting it in the tab is that this may be a very long list. How many values are there for e.g. `mysql`?

MikeMcQuaid:
Loving the work here so far, great job @jawshooah!

Library/Homebrew/dev-cmd/bottle.rb
jawshooah: Here's my `INSTALL_RECEIPT.json` for `mysql`:

```json
{
  "used_options": [],
  "unused_options": [
    "--with-test",
    "--with-embedded",
    "--with-archive-storage-engine",
    "--with-blackhole-storage-engine",
    "--with-local-infile",
    "--with-debug"
  ],
  "built_as_bottle": true,
  "poured_from_bottle": true,
  "changed_files": [
    "INSTALL_RECEIPT.json",
    "bin/mysql_config",
    "bin/mysqld_multi",
    "bin/mysqld_safe",
    "homebrew.mxcl.mysql.plist",
    "include/mysql/my_config.h",
    "lib/pkgconfig/mysqlclient.pc",
    "mysql-test/Makefile",
    "mysql-test/suite/ndb/include/have_java.inc",
    "share/doc/mysql/INFO_BIN",
    "share/man/man1/myisam_ftdump.1",
    "share/man/man1/mysql.server.1",
    "share/man/man1/mysql_config.1",
    "share/man/man1/mysql_install_db.1",
    "share/man/man1/mysqld_multi.1",
    "share/man/man1/mysqld_safe.1",
    "share/man/man1/mysqldumpslow.1",
    "support-files/mysql-log-rotate",
    "support-files/mysql.server",
    "support-files/mysqld_multi.server"
  ],
  "time": 1476101424,
  "source_modified_time": 1472126415,
  "HEAD": null,
  "stdlib": "libcxx",
  "compiler": "clang",
  "runtime_dependencies": [],
  "source": {
    "path": "/usr/local/Library/Taps/homebrew/homebrew-core/Formula/mysql.rb",
    "tap": "homebrew/core",
    "spec": "stable",
    "versions": {
      "stable": "5.7.15",
      "devel": null,
      "head": null,
      "version_scheme": 0
    }
  }
}
```

Not too bad, I think. Would you rather output to a separate file?

Library/Homebrew/metafiles.rb
jawshooah: O(1) lookup versus O(n). If order doesn't matter and there are no duplicates, why use an array when you could use a set?

jawshooah: Keep in mind that `Keg#text_files` queries this collection once for every file, so bringing the complexity of that lookup down is worthwhile if we care about `brew bottle` runtime.

Library/Homebrew/keg_relocate.rb
```diff
@@ -19,17 +19,19 @@ def relocate_dynamic_linkage(_old_prefix, _new_prefix, _old_cellar, _new_cellar)
     []
   end
 
-  def relocate_text_files(old_prefix, new_prefix, old_cellar, new_cellar,
-                          old_repository, new_repository)
-    files = text_files | libtool_files
+  def relocate_text_files(old_prefix, new_prefix, old_cellar, new_cellar, # rubocop:disable Metrics/ParameterLists
```
jawshooah: Agreed, I'll clean this up.

Library/Homebrew/formula_installer.rb
jawshooah: Sure. `Tab.from_file` and `Tab.for_keg` both cache the resulting `Tab` instance so that when either is called with the same path, the same instance is returned. `Tab.from_file_content` bypasses the cache. This is desirable here because `keg.relocate_text_files` will rewrite the tab file, and we want to load in the rewritten file before modifying it and writing it out again a few lines down.

Library/Homebrew/dev-cmd/bottle.rb
MikeMcQuaid: Nah that seems fine, thanks 👍

Library/Homebrew/metafiles.rb
MikeMcQuaid: The algorithmic speedup is always nice but given Sets aren't strictly interchangeable with arrays I figure it's usually worth using arrays unless profiling shows using sets produces a measurable speedup. Based on your previous comments it sounds like it does so 👍

Library/Homebrew/formula_installer.rb
MikeMcQuaid: I think it would be clearer to nuke the cache before rereading rather than use a method that's not using the cache (and should probably be `private` as a result).

Library/Homebrew/keg_relocate.rb
MikeMcQuaid: Thanks!

Library/Homebrew/keg_relocate.rb
```diff
@@ -75,8 +78,8 @@ def text_files
       path.find do |pn|
         next if pn.symlink? || pn.directory?
         next if Metafiles::EXTENSIONS.include? pn.extname
-        if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
-           pn.text_executable?
+        if pn.text_executable? ||
+           Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text")
```
MikeMcQuaid: @reitermarkus Some local testing looks like it should be, good catch. There's probably a maximum number of arguments we can specify but batching them may well produce a speedup 👍

reitermarkus: I tried to get rid of https://github.com/Homebrew/brew/blob/c9c7f462d37500549127efba96c7a25e5c70de4a/Library/Homebrew/cask/lib/hbc/artifact/uninstall_base.rb#L11 before by using `getconf ARG_MAX`, but somehow got side-tracked. 😄

jawshooah:
@MikeMcQuaid are you still interested in this now that #1258 and friends have been merged?

MikeMcQuaid:
> It still offers a significant boost in install time, ~4x from my rough testing.

This was going to be my next question! In that case: yes, definitely! Please also check a bottle with/without this data can still be poured/relocated correctly both without/with this change (i.e. that all four combinations work fine).

Great work here!

MikeMcQuaid:
Any news here @jawshooah? Shout if I can help at all.

Library/Homebrew/keg_relocate.rb
```diff
@@ -19,17 +19,39 @@ def relocate_dynamic_linkage(_old_prefix, _new_prefix, _old_cellar, _new_cellar)
     []
   end
 
-  def relocate_text_files(old_prefix, new_prefix, old_cellar, new_cellar,
-                          old_repository, new_repository)
-    files = text_files | libtool_files
+  def replace_locations_with_placeholders
+    relocate_dynamic_linkage(HOMEBREW_PREFIX.to_s, PREFIX_PLACEHOLDER,
```
MikeMcQuaid: Tiny nit: might want to make the two prefix args a key/value of a hash or something? Doesn't really matter but I always get a little 😨 about things that rely on arg ordering like this.

MikeMcQuaid: In fact: like you did below with replacements 😆

jawshooah:
[**@MikeMcQuaid**](https://github.com/MikeMcQuaid) commented on [Oct 23, 2016, 4:34 AM EDT](https://github.com/Homebrew/brew/pull/1253#issuecomment-255576019 "2016-10-23T08:34:08Z - Replied by Github Reply Comments"):
> Any news here @jawshooah? Shout if I can help at all.

Just need to get around to re-verifying that everything works in the scenarios you mentioned. Nothing blocking me but time :smile:

Library/Homebrew/keg_relocate.rb
jawshooah: Will do!

jawshooah: How about a Struct? https://github.com/Homebrew/brew/pull/1253/commits/f798ba4a6ee53795e704d66bc298f73f4b1bff59

jawshooah:
@MikeMcQuaid I've verified that there are no problems installing with/without the data and with/without the change. Merging now as CI is :green_heart: and you've already approved the changes.

scpeters:
this may have broken the mercurial bottle for me, still investigating...

scpeters:
~~~
$ head -1 /usr/local/opt/mercurial/bin/hg
#!@@HOMEBREW_PREFIX@@/opt/python/bin/python2.7
~~~

scpeters:
I reverted to e6bce5ec27306e85a8312400f44d8b9685f9e981, and the problem went away, so I think it's an issue with this PR.

Here's what I see when I install mercurial:

~~~
brew install mercurial
==> Downloading https://homebrew.bintray.com/bottles/mercurial-3.9.2.yosemite.bottle.tar.gz
Already downloaded: /Users/jenkins/Library/Caches/Homebrew/mercurial-3.9.2.yosemite.bottle.tar.gz
==> Pouring mercurial-3.9.2.yosemite.bottle.tar.gz
Error: No such file or directory - /usr/local/opt/mercurial/bin/hg
~~~

There is a file at that location though:

~~~
$ head /usr/local/opt/mercurial/bin/hg
#!@@HOMEBREW_PREFIX@@/opt/python/bin/python2.7
#
# mercurial - scalable distributed SCM
#
# Copyright 2005-2007 Matt Mackall <mpm@selenic.com>
#
# This software may be used and distributed according to the terms of the
# GNU General Public License version 2 or any later version.

import os
~~~

zmwangx:
There's a bug here with deleting pyc files in the wrong order. Causing a failure: https://bot.brew.sh/job/Homebrew%20Core/10080/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/bottle_buku/. Fix in #1375.

PR #1255, opened 2016-10-10
brew.sh: allow --prefix to be run as root without warning

akirk:

-----

In the files `.bashrc` or `.zshrc` the command `brew --prefix xyz` is often used to set the correct `$PATH` for executables. For example: `export PATH="$(brew --prefix php70)/bin:$PATH"`

When the `*rc` files are executed as `root` (e.g. via `sudo -s`) then `--prefix` should be considered a legitimate use of the `brew` command while under uid `0`.

Old behavior:
```
$ brew --prefix gcc
Error: Running Homebrew as root is extremely dangerous. As Homebrew does not
drop privileges on installation you are giving all build scripts full access
to your system. As a result of the macOS sandbox not handling the root user
correctly HOMEBREW_NO_SANDBOX has been set so the sandbox will not be used. If
we have not merged a pull request to add privilege dropping by November 1st
2016 running Homebrew as root will be disabled. No Homebrew maintainers plan
to work on this functionality.
/usr/local/Cellar/gcc/6.2.0
$ brew --prefix gcc 2>/dev/null
/usr/local/Cellar/gcc/6.2.0
```

New behavior:
```
$ brew --prefix gcc
/usr/local/Cellar/gcc/6.2.0
```

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -243,6 +243,7 @@ check-run-command-as-root() {
 
   [[ "$HOMEBREW_COMMAND" = "cask" ]] && return
   [[ "$HOMEBREW_COMMAND" = "services" ]] && return
+  [[ "$HOMEBREW_COMMAND" = "--prefix" ]] && return
 
   onoe <<EOS
 Running Homebrew as root is extremely dangerous. As Homebrew does not


MikeMcQuaid:
Seems reasonable. Can you also add `--repository` in there too? Thanks!

akirk:
@MikeMcQuaid that's already handled https://github.com/Homebrew/brew/blob/master/Library/Homebrew/brew.sh#L63, also `--prefix` without arguments is covered by that `case` block.

This PR is about the case when `--prefix` has an argument.

MikeMcQuaid:
Thanks for the quick response!

PR #1258, opened 2016-10-10
keg_relocate: run /usr/bin/file in batches

jawshooah:

-----

Rather than running `/usr/bin/file` on each individual file to determine its type, split all files into batches as large as possible and run `/usr/bin/file` on each batch.

This is an alternative solution to #1250 as suggested by @reitermarkus in https://github.com/Homebrew/brew/pull/1253#discussion_r82602685, though the two solutions could be combined.

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -72,13 +72,14 @@ class Keg
     # file with that fix is only available in macOS Sierra.
     # http://bugs.gw.com/view.php?id=292
     with_custom_locale("C") do
-      path.find do |pn|
-        next if pn.symlink? || pn.directory?
-        next if Metafiles::EXTENSIONS.include? pn.extname
-        if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
-           pn.text_executable?
-          text_files << pn
-        end
+      files = path.find.reject { |pn|
+        pn.symlink? || pn.directory? || Metafiles::EXTENSIONS.include?(pn.extname)
+      }
+      output, _status = Open3.capture2("/usr/bin/xargs -0 /usr/bin/file --no-dereference --brief",
+                                       stdin_data: files.join("\0"))
+      output.each_line.with_index do |line, i|
+        next unless line.include?("text")
+        text_files << files[i]
       end
     end
 


jawshooah:
@reitermarkus This cuts install time for `mysql` down to ~40s. Not as big of an improvement as #1253, but there's no reason we couldn't do both!

MikeMcQuaid:
hqdefault

Library/Homebrew/keg_relocate.rb
```diff
@@ -72,14 +72,30 @@ def text_files
...
+          file_lists.last << file
+          current_length += file_length
         end
       end
+
+      file_lists.each do |list|
+        Utils.popen_read("/usr/bin/file", "--no-dereference", "--brief", *list)
+             .each_line.with_index do |line, i|
+               text_files << list[i] if line.include?("text")
+             end
```
MikeMcQuaid: Indent this with `end` and the `text_files` two-characters in

Library/Homebrew/keg_relocate.rb
```diff
@@ -72,14 +72,30 @@ def text_files
...
+        else
+          file_lists.last << file
+          current_length += file_length
         end
       end
+
+      file_lists.each do |list|
+        Utils.popen_read("/usr/bin/file", "--no-dereference", "--brief", *list)
+             .each_line.with_index do |line, i|
+               text_files << list[i] if line.include?("text")
```
MikeMcQuaid: `next unless line.include?("text")`

Library/Homebrew/keg_relocate.rb
jawshooah: Sorry, do you mean align `end` with `Utils`?

MikeMcQuaid: Sorry, yeh.

MikeMcQuaid: Aye 👍

jawshooah: Done in 513c712e4e39b219e405b305c56cc333a59c5550

Library/Homebrew/utils/shell.rb
```diff
@@ -90,6 +85,28 @@ def prepend_path_in_shell_profile(path)
         "echo 'set -g fish_user_paths \"#{sh_quote(path)}\" $fish_user_paths' >> #{shell_profile}"
       end
     end
-    module_function :prepend_path_in_shell_profile
+
+    # cut the actual value of ARG_MAX in half for safety since
+    # environment variables count toward the total
+    def max_command_length
+      Utils.popen_read("/usr/bin/getconf", "ARG_MAX").to_i / 2
```
reitermarkus: I think halving the `ARG_MAX` is kinda radical.

```ruby
ENV.inject(0) { |c, (k, v)| c + 2 + k.length + v.length }
```

get's the exact length of the environment variables. Doubling that should be enough safety margin.

Library/Homebrew/utils/shell.rb
```diff
@@ -89,7 +89,12 @@ def prepend_path_in_shell_profile(path)
     # cut the actual value of ARG_MAX in half for safety since
     # environment variables count toward the total
     def max_command_length
-      Utils.popen_read("/usr/bin/getconf", "ARG_MAX").to_i / 2
+      Utils.popen_read("/usr/bin/getconf", "ARG_MAX").to_i - 2 * environment_variables_length
```
reitermarkus: Could probably make this

```ruby
@max_command_length ||= Utils.popen_read("/usr/bin/getconf", "ARG_MAX").to_i
@max_command_length - 2 * environment_variables_length
```

since the `ARG_MAX` will never change.

jawshooah: Actually, looks like this broke something somehow: https://gist.github.com/jawshooah/c072dd613cdefd383cebba8883e2244f

Something other than environment variables must be eating up command characters.

jawshooah: Nice, I figured there was a reason it felt right to cut it in half!

MikeMcQuaid: That makes me wonder a bit if it's worth using `xargs` to do the argument splitting to save us some code rather than duplicate theirs?

MikeMcQuaid: If it's no slower: less code == 😄

jawshooah: Done in d8a6863283048cd2cc1f7fb78178b2fc19420e0c. Much less code indeed! :smile:

Library/Homebrew/keg_relocate.rb
```diff
@@ -72,13 +72,14 @@ def text_files
...
     # http://bugs.gw.com/view.php?id=292
     with_custom_locale("C") do
-      path.find do |pn|
-        next if pn.symlink? || pn.directory?
-        next if Metafiles::EXTENSIONS.include? pn.extname
-        if Utils.popen_read("/usr/bin/file", "--brief", pn).include?("text") ||
-           pn.text_executable?
-          text_files << pn
-        end
+      files = path.find.reject { |pn|
```
jawshooah: We have `Style/BlockDelimiters` set to `semantic`, which enforces curly braces around functional blocks and `do...end` around procedural blocks. If you'd rather we just use `do...end` for all multiline blocks, we can change that in a separate PR, as it will require many more changes elsewhere.

MikeMcQuaid: I like C lookalikes a lot but not in my Ruby 😆

reitermarkus: > not in my Ruby

Let the record show that I am a proponent of the semantic style.

tdsmith:
Something in this series is making `brew bottle python3` explode: https://bot.brew.sh/job/Homebrew%20Core/9501/version=any_sierra/console (ref https://github.com/Homebrew/homebrew-core/pull/5954)

tdsmith:
Er, commented on the wrong PR

PR #1259, opened 2016-10-10
audit: check patch sha matches when --online

alyssais:

-----

See Homebrew/homebrew-core#5353 for motivation.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -710,6 +710,15 @@ class FormulaAuditor
     when %r{^http://bugs\.debian\.org}
       problem "Patches from Debian should be https://, not http:\n#{patch.url}"
     end
+
+    return unless @online
+    audit_resource_checksum(patch.resource)
+  end
+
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
`brew fetch -s` may already do this. If so, I think this is already covered by the test-bot? If not: that'd be the best place to add it. Thanks for jumping on this anyway, though!

alyssais:
I can't figure out what `-s` does, but `fetch` does seem to do this, so no point duplicating it here.

I think there is some value to having a command that would check this for all formulae like `brew audit` does though. This would let people find problems and fix them in bulk before they affected anyone. So a way to bulk fetch all formulae?

MikeMcQuaid:
Sorry, `-s` is `--build-from-source`'s alias.

Could see an argument for a `brew fetch --force --all` but that's going to do all tarballs and patches too which I'm not sure will scale well.

alyssais:
> which I'm not sure will scale well.

Good point. I do think it would be a good idea to have some way of identifying broken formulae because resources have changed, but I don't know how to go about this, so I'll close this.

PR #1260, opened 2016-10-11
tab: don't interpret unknown deps as no deps

alyssais:
- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change? <ins>See below.</ins>

-----

I extracted this change from #1082 in the hope it can be merged as soon as possible, because without this change incorrect data could be being recorded in install receipts.

When I made `runtime_dependencies` default to `[]`, it didn't occur to me that tabs could be modified outside of the context of a new installation. However, I've just noticed that `brew bottle` does this. So, if there was a keg installed before #1032 was merged, and it was later bottled, I think the install receipt in the bottle would incorrectly record that there were no runtime dependencies, rather than correctly leaving them unspecified.

### DIFF
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -60,7 +60,6 @@ class Tab < OpenStruct
   def self.from_file_content(content, path)
     attributes = Utils::JSON.load(content)
     attributes["tabfile"] = path
-    attributes["runtime_dependencies"] ||= []
     attributes["source_modified_time"] ||= 0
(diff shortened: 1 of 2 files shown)


alyssais:
Travis is failing, but it seems unrelated.

MikeMcQuaid:
Restarting Travis build.

MikeMcQuaid:
Thanks again @alyssais!

PR #1266, opened 2016-10-12
Remove Cask’s RuboCop configuration.

reitermarkus:

-----

Removed Cask's RuboCop config and removed Cask from the main one's exclusion list.


jawshooah:
Did you mean to merge `cask_loader` back into `path_base`, or is this branch just based on an old commit?

Library/.rubocop.yml
```diff
@@ -1,5 +1,7 @@
 AllCops:
   TargetRubyVersion: 2.0
+  Exclude:
+    - '**/Casks/**/*'
```
MikeMcQuaid: I'm 🆒 with this but just wondering what it's doing that you don't like?

jawshooah: The `rubocop-cask` gem includes a default RuboCop configuration for Casks that is quite different from the one for brew core: https://github.com/caskroom/rubocop-cask/blob/master/config/default.yml

Library/Homebrew/cask/lib/hbc/source/path_base.rb
```diff
@@ -1,28 +1,72 @@
 require "rubygems"
-require "hbc/cask_loader"
```
jawshooah: Did you mean to merge `CaskLoader` back into `PathBase`, or is this branch just based on an old commit?

reitermarkus: I think I started this branch before it got merged into master.

Library/.rubocop.yml
MikeMcQuaid: 🆒

jawshooah:
Given https://github.com/Homebrew/brew/pull/1258#discussion_r82949331, let's go ahead and change `Style/BlockDelimiters` to either [`line_count_based`](https://github.com/bbatsov/rubocop/blob/6a3442b9d2bb3e6a6abacd15889560057463f165/config/default.yml#L207-L209) or [`braces_for_chaining`](https://github.com/bbatsov/rubocop/blob/6a3442b9d2bb3e6a6abacd15889560057463f165/config/default.yml#L220-L224). @MikeMcQuaid any preference?

MikeMcQuaid:
> Given #1258 (comment), let's go ahead and change Style/BlockDelimiters to either line_count_based or braces_for_chaining. @MikeMcQuaid any preference?

@jawshooah I'm happy with either, actually. I'm guessing @reitermarkus will probably prefer the latter and that's fine with me 👍

reitermarkus:
> I'm happy with either, actually.

I'm happy with neither, actually. The first one really messes up the RSpec `let` syntax, and in addition to that, the second one leads to very inconsistent style, where you can have e.g. both `map do end` and `map { }`, even when both are doing the same thing, simply because one of them is in a chain.

jawshooah:
@reitermarkus I agree, and I think using `do...end` with functional code makes no sense, but it doesn't bother me too much as long as it's consistent. @MikeMcQuaid, on the other hand, seems to find curly braces on multiline blocks nauseating, so I don't mind backing down on this one. Is it equally nauseating for you to see `do...end` used non-procedurally?

maxnordlund:
There is a semantic option that essentially enforces certain methods, e.g. `map`, to use curly brackets, and everything else to use `do ... end`. It's what we use at work and I find it the best. It's consistent per method, and you avoid the nasty `map do ... end.filter do ... end`.

It also helps differentiate between stuff that returns something useful and stuff that has side-effects.

https://github.com/bbatsov/rubocop/blob/master/config/default.yml#L204

jawshooah:
@maxnordlund Yup, that's what we're currently using.

maxnordlund:
Oh, sorry.

MikeMcQuaid:
> The first one really messes up the RSpec let syntax

We can filter out the RSpec tests. Long-term I don't think it makes sense for a single project to have three test suites across two test frameworks; we should aim to have `brew tests` run everything (and I have no strong feelings whether that's RSpec or Minitest).

> the second one leads to very inconsistent style, where you can have e.g. both map do end and map { }, even when both are doing the same thing, simply because one of them is in a chain.

I think that's better than the current version because the use of `{}` indicates whether it's currently in a chain rather than whether it could potentially be used in a chain.

That said, again, I don't have strong feelings either way. It seems excluding RSpec and picking the first may be the most agreeable?

MikeMcQuaid:
> That said, again, I don't have strong feelings either way. It seems excluding RSpec and picking the first may be the most agreeable?

@reitermarkus @jawshooah Thoughts on this? Would like to get this PR shipped in some form.

jawshooah:
> It seems excluding RSpec and picking the first may be the most agreeable?

I'm fine with that. @reitermarkus?

reitermarkus:
It may be that I missed something, but I haven't seen a *strong* argument against using semantic style (other than @MikeMcQuaid's skin crawling 😂, which is pretty subjective), so I'm still pretty much in favor of keeping the semantic style.

MikeMcQuaid:
Another subjective argument: it's not the wider Homebrew style.

Less subjective argument: it means that it's not at all obvious whether you need to use `do` or `{` without knowing something about the thing you're calling. I get that wrong frequently and working with Ruby is my job; we're trying to attract drive-by contributors here and that's not helpful for them. I should be able to infer from either the fact that I'm chaining or where I put my linebreaks whether I use `do` or `{`.

reitermarkus:
> Another subjective argument: it's not the wider Homebrew style.

This is basically like reasoning with “because it has always been this way“.

> it's not at all obvious whether you need to use `do` or `{` without knowing something about the thing you're calling

Could you give an example? I fail to see how you can call something and not know anything about it.

MikeMcQuaid:
> This is basically like reasoning with “because it has always been this way“.

Which for legacy codebases is not an unreasonable or uncommon position to take.


> Could you give an example? I fail to see how you can call something and not know anything about it.

Well, the cop itself says "but exceptions are permitted in the `ProceduralMethods`, `FunctionalMethods` and `IgnoredMethods` sections below". Without looking: could you name each of those methods? I certainly couldn't. 

@jawshooah and I are both happy picking between those two methods so I think it's worth just picking one of the other two options.

reitermarkus:
> Without looking: could you name each of those methods? I certainly couldn't.

You don't have to be able to name them, you just have to know what a given method does. If it returns something, use `{}`, else use `do … end`.

MikeMcQuaid:
> You don't have to be able to name them, you just have to know what a given method does. If it returns something, use {}, else use do … end.

It seems I'm not going to convince you but: to be fair, I don't think I have to here. @jawshooah and I agree here; can we reach some sort of compromise?

reitermarkus:
> It seems I'm not going to convince you.

Yeah, we're in the same boat in this regard. 😂 

> @jawshooah and I agree here; can we reach some sort of compromise?

@jawshooah, do you, though? You were the one who [suggested semantic style](https://github.com/caskroom/homebrew-cask/pull/22855#issuecomment-233227290).

jawshooah:
@reitermarkus While personally I prefer a semantic style, and that's the style I would use in any small or personal projects, this repo attracts contributors of varying levels of experience, and I think it makes sense to enforce a style that is more immediately understandable for beginners.

maxnordlund:
When I first started coding ruby at work, I didn't know it was legal syntax to write `do ... end.method`. Then I got it, and while we use the semantic style at work, I think curly brackets is the lesser of two evils here. Even though `do ... end` looks more ruby, I don't think people will be as confused by `{ ... }.method` as the `do ... end.method` one.

MikeMcQuaid:
> Given #1258 (comment), let's go ahead and change Style/BlockDelimiters to either line_count_based or braces_for_chaining. @MikeMcQuaid any preference?

I've gone for `line_count_based` and done it in https://github.com/Homebrew/brew/pull/1345 because that's the style we're using more widely in Homebrew/brew already (i.e. it required fewer fixes than `braces_for_chaining`). Once merged and this PR is rebased I think we should be good to go.

MikeMcQuaid:
@reitermarkus Could you resolve conflicts on this when you get the chance? Would be good to get it merged in.

MikeMcQuaid:
Thanks @reitermarkus! Looks good to me.

PR #1287, opened 2016-10-14
Haggins -> Hagins

jawshooah:
I'm no hobbit!


MikeMcQuaid:
So sorry!

reitermarkus:
The URL is still wrong. How hard can it be? :joy:

PR #1290, opened 2016-10-14
Fix wrong ordering on when clauses in audit.rb

maxnordlund:

-----

It looks like a rebase/merge conflict remmenet. I'm not totally sure if I got it right, but it is definitely not right as is.


MikeMcQuaid:
There's a `brew style` complaint after this I'm not sure I understand: https://bot.brew.sh/job/Brew/2444/version=any_sierra/testReport/junit/brew-test-bot/sierra/style/

maxnordlund:
I think I get it, any splats when clauses e.g. `when FOO*` must come last since they catch more then the plain ones. I'll fix it right away.

PR #971, opened 2016-09-15
Make `brew upgrade` and `brew outdated` follow alias changes

penman:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).

  <del>I've written <em>a few</em> tests, but plan on writing more.</del> <ins>Done!</ins>
- [x] Have you successfully run `brew tests` with your changes locally?

-----

This pull request makes `brew outdated` and `brew update` follow alias changes, as described in #567.

Issue #1162, opened 2016-09-27
Unable to install devel if stable installed.


vladshablinsky:
## Description

```
➜ vlad:Homebrew$ brew install ack                                        master
==> Using the sandbox
==> Downloading http://beyondgrep.com/ack-2.14-single-file
Already downloaded: /Users/vlad/Library/Caches/Homebrew/ack-2.14.14-single-file
==> pod2man /usr/local/Cellar/ack/2.14/bin/ack ack.1
🍺  /usr/local/Cellar/ack/2.14: 3 files, 182.9K, built in 1 second
➜ vlad:Homebrew$ brew install ack --devel                                master
Warning: ack-2.14 already installed
```
## Workaround

We'd probably want to warn user that `stable` keg is linked and suggest user unlinking it. After the keg have been unlinked we should be able to install any version that is newer than the latest version installed in the `Cellar` e.g. `HEAD` or `devel`.
## TODO
- [ ]  WIP Improve `install` #1343
- [ ]  Improve `upgrade`

CC @Homebrew/maintainers

MikeMcQuaid:
Seems sensible but out of interest what happens if you use `brew upgrade ack --devel`?

vladshablinsky:
> Seems sensible but out of interest what happens if you use brew upgrade ack --devel?

Wow, it works! Will experiment with it a bit and then close the issue if I don't find anything strange. Thanks!

vladshablinsky:
Seems that the warning message should be updated, though.

ilovezfs:
Yup, that works, though it does still block any transition from HEAD to devel or devel to stable or HEAD to stable.

MikeMcQuaid:
I think requiring intervention on a downgrade is sensible but agreed the messaging should be updated in both cases.

ilovezfs:
What intervention actually works for a downgrade currently other than uninstall?

MikeMcQuaid:
`unlink`, I think?

ilovezfs:
Nope!

ilovezfs:
```
iMac-TMP:~ joe$ brew install --devel ack
Warning: ack-HEAD-f0ad2b9 already installed
iMac-TMP:~ joe$ brew unlink ack
Unlinking /usr/local/Cellar/ack/HEAD-f0ad2b9... 2 symlinks removed
iMac-TMP:~ joe$ brew install --devel ack
Warning: ack-HEAD-f0ad2b9 already installed, it's just not linked
```

[...]

PR #1296, opened 2016-10-15
cmd/install: improve logic for following aliases

vladshablinsky:

---

Changes for `cmd/install`:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/install.rb#L138-L139
`f.installed?` means that last prefix is installed for formula spec. However it seems to me that here it's treated as a check for any version being installed.

Before the change `f.installed?` did everything properly, it just refused to install formula if its latest prefix was installed. Not sure it was right to do so, but if `f.installed?` check had failed and formula still had been installed then `formula_installer` would have handled this.
Here is the diff with that change:
https://github.com/Homebrew/brew/pull/971/files#diff-b82c90101b7f1236459d3ba357ccc6e0

Also, 
`current ||= f.old_installed_formulae.first` doesn't always mean that `f.old_installed_formulae.first.installed?` is true.

These lines confuse me a little. As for now we can have formula installed and its revision changed, which means that `f.installed?` is false and let's suppose `f.old_installed_formulae.first` is not `nil`. 

The fix is to check if `f.installed?` is true or old formulae installed
with `f`'s current alias is `installed?` which implies that its latest prefix
is installed.

Maybe this is not the best fix for the issue, but I think we'll figure out what is best and then I'll update the PR.

CC @MikeMcQuaid @alyssais

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -136,7 +136,12 @@ module Homebrew
         end
 
         current = f if f.installed?
-        current ||= f.old_installed_formulae.first
+        if current.nil? && f.installed_prefixes.empty?
+          old_formula_with_current_alias = f.old_installed_formulae.first
+          if old_formula_with_current_alias && old_formula_with_current_alias.installed?
+            current ||= old_formula_with_current_alias
+          end
+        end
 
         if current
           msg = "#{current.full_name}-#{current.installed_version} already installed"
@@ -148,7 +153,7 @@ module Homebrew
           # Check if the formula we try to install is the same as installed
           # but not migrated one. If --force passed then install anyway.
           opoo "#{f.oldname} already installed, it's just not migrated"
-          puts "You can migrate formula with `brew migrate #{f}`"
+          puts "You should migrate formula with `brew migrate #{f}`"
           puts "Or you can force install it with `brew install #{f} --force`"
         else
           formulae << f
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1188,10 +1188,11 @@ class Formula
   end
 
   def old_installed_formulae
+    return [] if alias_path.nil?
     # If this formula isn't the current target of the alias,
     # it doesn't make sense to say that other formulae are older versions of it
     # because we don't know which came first.
-    return [] if alias_path.nil? || installed_alias_target_changed?
+    return [] if installed_alias_target_changed?
     self.class.installed_with_alias_path(alias_path) - [self]
   end
 


alyssais:
Yeah, I think what must have happened is that I assumed that `Formula#installed?` whether _any_ version was installed, rather than just the latest one — I must have figured we wouldn't want `brew install x` with any x installed to work, preferring `brew upgrade x` instead.

If that is the behaviour we want, the correct fix would be to replace the `f.installed?` check with one that checked whether any version was installed, and leave `old_installed_formulae` check as is.

However, it seems that the way this worked before was that `brew install x` when an old version of `x` was installed would just install the new version. If we want to keep it that way, we should just check `f.installed?` and skip any `old_installed_formulae` check, because if an old formula is installed, by definition it's not the current version.

vladshablinsky:
> I must have figured we wouldn't want brew install x with any x installed to work, preferring brew upgrade x instead.

`Formula#installed?` can really confuse, so no worries. 

> However, it seems that the way this worked before was that brew install x when an old version of x was installed would just install the new version.

Yes, that's true if an old one is unlinked, otherwise it says that you need to unlink the old one to install new one. This behaviour is fine I guess, however I'm working on extending it so the user can install the latest version of any spec.

> If that is the behaviour we want, the correct fix would be to replace the f.installed? check with one that checked whether any version was installed, and leave old_installed_formulae check as is.

This check is performed in `formula_installer` for now and I'm not sure we need to change it right now, although it'll probably be changed when I finish with this: https://github.com/Homebrew/brew/issues/1162.

> leave old_installed_formulae check as is.

We need to check `old_installed_formulae` if new formula is not installed, right? AFAIU this change is done for better version support in future, could you elaborate why we need to perform this check at all?

Supposing we have the following mechanism for upgrading. does that mean that we get `boost@1.62 is already installed` or something when we're trying to install `boost@1.63`?

```
Upgrading a multiple version formula

In separate pull-requests:

Add the new keg_only version formula e.g. boost@1.63.rb
Modify the boost alias to point to boost@1.63.rb. Any formulae that need the old version of boost
should have their depends_on "boost" to be updated to depends_on "boost@1.62". boost@1.63.rb 
should no longer be keg_only and boost@1.62.rb should become keg_only.
When boost@1.63 has two major/minor versions newer than it (e.g. boost@1.65) then consider 
removing boost@1.63.rb and anything that depends on it.
```
(https://github.com/MikeMcQuaid/brew/blob/c3771336cff5630ce09248a57728e67393008087/docs/Migrate-A-Formula-To-Multiple-Versions.md)

MikeMcQuaid:
> Yeah, I think what must have happened is that I assumed that Formula#installed? whether any version was installed, rather than just the latest one

I think many people make this assumption. It would be great to rename this method to `latest_installed?` or something and add `installed?` as a fallback to `compat`.

MikeMcQuaid:
> Supposing we have the following mechanism for upgrading. does that mean that we get boost@1.62 is already installed or something when we're trying to install boost@1.63?

You should only get that if you do `brew install boost@1.62` not not for 1.63 through either `boost@1.63` or the `boost` alias.

Library/Homebrew/cmd/install.rb
```diff
@@ -136,7 +136,12 @@ def install
         end
 
         current = f if f.installed?
-        current ||= f.old_installed_formulae.first
+        if current.nil? && f.installed_prefixes.empty?
+          old_formula_with_current_alias = f.old_installed_formulae.first
+          if old_formula_with_current_alias && old_formula_with_current_alias.installed?
+            current ||= old_formula_with_current_alias
```
MikeMcQuaid: Can make this `current =` as you've already verified it's `nil`.

Library/Homebrew/cmd/install.rb
```diff
@@ -148,7 +153,7 @@ def install
           # Check if the formula we try to install is the same as installed
           # but not migrated one. If --force passed then install anyway.
           opoo "#{f.oldname} already installed, it's just not migrated"
-          puts "You can migrate formula with `brew migrate #{f}`"
+          puts "You should migrate formula with `brew migrate #{f}`"
```
MikeMcQuaid: `You should migrate the formula with`

alyssais:
> We need to check `old_installed_formulae` if new formula is not installed, right? AFAIU this change is done for better version support in future, could you elaborate why we need to perform this check at all?

`old_installed_formulae` is an extension of the check for old versions of the current formula. We should be checking one if and only if we're checking the other.

alyssais:
This does look to me like the correct adjustment to account for alias changes while preserving the same behaviour. 👍

MikeMcQuaid:
@vladshablinsky Any news here? Thanks!

vladshablinsky:
@MikeMcQuaid let's close it for know and see if it's needed when #1343 merged.

PR #1297, opened 2016-10-16
Complain about HEP 6 violations.

tdsmith:
The shebang check may be too stringent; these may be small scripts that
don't touch the network and installing Python for them could be
overkill. Or maybe not!

### DIFF
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -60,6 +60,7 @@ module FormulaCellarChecks
   end
 
   def check_python_virtualenv
+    return if default_python_is_system_python?
     framework = formula.libexec/".Python"
     return unless framework.symlink?
     return unless framework.realpath.to_s.start_with?("/System")
@@ -72,6 +73,7 @@ module FormulaCellarChecks
 
   def check_python_shebangs
     return unless formula.bin.directory?
+    return if default_python_is_system_python?
     system_python_shebangs = formula.bin.children.select do |bin|
       (bin.open { |f| f.read(32) }).start_with?("#!/usr/bin/python")
     end
@@ -107,5 +109,25 @@ module FormulaCellarChecks
     audit_check_output(check_python_virtualenv)
     audit_check_output(check_python_shebangs)
     check_linkage
+  end
+
+  def default_python_is_system_python?
+    sanitized_path = ENV["PATH"]
+    begin
+      # Run `python` in the user's environment to get the real answer because
+      # the python we find in the user's PATH might be a pyenv shim
+      ENV["PATH"] = ENV["HOMEBREW_USER_PATH"]
+      which_python = which("python")
+      python_exec, _ = Open3.capture2(which_python, "-c", "import sys; print(sys.executable)")
+      python_exec = Pathname.new(python_exec.strip).realpath
+    rescue => e
+      opoo "Inconsistent Python environment: #{e}"
+      python_exec = Pathname.new("")
+    ensure
+      ENV["PATH"] = sanitized_path
+    end
+
+    return nil unless python_exec.exist?
+    return python_exec.to_s.start_with?("/usr/bin/python")
   end
 end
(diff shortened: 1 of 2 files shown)


Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -59,6 +59,33 @@ def check_python_framework_links(lib)
...
   end
 
+  def check_python_virtualenv
+    framework = formula.libexec/".Python"
+    return unless framework.symlink?
+    return unless framework.realpath.to_s.start_with?("/System")
+    <<-EOS.undent
+      virtualenv created against system Python
+      This formula created a virtualenv using system Python.
+      Please add `depends_on :python` to the formula.
```
MikeMcQuaid: `depends_on :python` may use the system Python, though, if there's no Homebrew one installed.

tdsmith:
Wait, that last commit was dumb.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -59,6 +59,35 @@ def check_python_framework_links(lib)
...
+    EOS
+  end
+
+  def check_python_shebangs
+    return unless formula.bin.directory?
+    return if formula_requires_python?(formula) && default_python_is_system_python?
+    system_python_shebangs = formula.bin.children.select do |bin|
+      (bin.open { |f| f.read(32) }).start_with?("#!/usr/bin/python")
```
MikeMcQuaid: Why 32? If it's the string length can you make it a variable and get the length dynamically that way?

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
MikeMcQuaid: Again, still not quite sure I understand this message given the system Python behaviour.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -59,6 +59,35 @@ def check_python_framework_links(lib)
...
+    end
+    return if system_python_shebangs.empty?
+
+    <<-EOS.undent
+      python scripts run with system python
```
MikeMcQuaid: Again, still not quite sure I understand this message given the system Python behaviour.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
tdsmith: The alert only fires if `depends_on :python` was not specified.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
tdsmith: Arbitrary short length. Sure, doesn't have to be magical.

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
tdsmith: — The rationale is that `depends_on :python` will never fall back on system Python on the test bot, which is all I care about detecting.

tdsmith:
Can you comment on the message you'd prefer to see and when this messaging will be confusing?

Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
MikeMcQuaid: By "on the test bot" you mean "when building bottles", right? I'm not 100% sure if that happens if e.g. `bottle :disabled` or `bottle :unused` is set; may be worth testing.

MikeMcQuaid:
> Can you comment on the message you'd prefer to see and when this messaging will be confusing?

I think it may be worth filtering and/or explicitly referencing in both messaging and code whether the formula was built as a bottle and that's the failure case. Arguably that's a good reason to perhaps punt this check to `brew bottle` too.

MikeMcQuaid:
> these may be small scripts that don't touch the network and installing Python for them could be overkill

Given this assumes things communicating over the network and I'm unconvinced that's the majority case I'm probably 👎 on having an automated check for this. False positives are fine, false negatives are not.

MikeMcQuaid:
Let's pass on this for now but thanks for the work @tdsmith.

PR #1273, opened 2016-10-12
keg_relocate: fix for multiline /usr/bin/file output

jawshooah:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Sometimes, `/usr/bin/file` spits out multiple lines for a single file for some reason. Fix the optimization from #1258, #1267, and #1269 to handle this case.

PR #1298, opened 2016-10-16
Don't choke on invalid UTF-8 in `file` output

tdsmith:
Sometimes `file` output contains data from the file under examination,
which may include binary data that does not represent valid UTF-8
codepoints. String#split dies if it doesn't understand the encoding, so
tell Ruby to treat `file` output as a bytestring.

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -84,9 +84,12 @@ class Keg
       }
       output, _status = Open3.capture2("/usr/bin/xargs -0 /usr/bin/file --no-dereference --print0",
                                        stdin_data: files.to_a.join("\0"))
+      # `file` output sometimes contains data from the file, which may include
+      # invalid UTF-8 entities, so tell Ruby this is just a bytestring
+      output.force_encoding(Encoding::ASCII_8BIT)
       output.each_line do |line|
-        path, info = line.split("\0")
-        next unless info.to_s.include?("text")
+        path, info = line.split("\0", 2)
+        next unless info.include?("text")
         path = Pathname.new(path)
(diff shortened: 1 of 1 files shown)


tdsmith:
cc @jawshooah

tdsmith:
A failing line looks like:
`"/usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/__pycache__/reprlib.cpython-35.opt-2.pyc\u0000:                               dBase III DBT, version number 0, next free block index 168627478, 1st item \"\x83\u0001\"\n"`

Library/Homebrew/keg_relocate.rb
```diff
@@ -84,9 +84,12 @@ def text_files
...
       output, _status = Open3.capture2("/usr/bin/xargs -0 /usr/bin/file --no-dereference --print0",
                                        stdin_data: files.to_a.join("\0"))
+      # `file` output sometimes contains data from the file, which may include
+      # invalid UTF-8 entities, so tell Ruby this is just a bytestring
+      output.force_encoding(Encoding::ASCII_8BIT)
       output.each_line do |line|
-        path, info = line.split("\0")
-        next unless info.to_s.include?("text")
+        path, info = line.split("\0", 2)
+        next unless info.include?("text")
```
jawshooah: The `to_s` needs to stay for the reasons given in https://github.com/Homebrew/brew/pull/1273#discussion_r83545997.

tdsmith: oh, gross, okay! this is surprising enough i think being more explicit about it is good

tdsmith:
Thanks for the reviews!

PR #1300, opened 2016-10-16
Update FAQ.md

delamoe:
results in 404 error...
### Is there a glossary of terms around?
All your terminology needs can be [found here](docs/Formula-Cookbook.md#homebrew-terminology). 

fixed with...
(Formula-Cookbook.md#homebrew-terminology).


vladshablinsky:
Thanks for fixing this @delamoe! 🎉

PR #1305, opened 2016-10-16
install: prune build-only deps of deps

### LINKED ISSUE #1272: Needlessly building ":build" deps for formulas that are already installed
[Transferred from Homebrew/homebrew-core#5740 at @MikeMcQuaid's request]

I noticed this problem with `diff-pdf`, but it's probably because of a bug in `brew`'s deciding which deps need to be installed.

`brew install --build-from-source diff-pdf` will incorrectly try to build and install `:build` deps for `openjpeg`, even when `openjpeg` has already been poured from a bottle.

Reproduce: Pour the bottles for `diff-pdf`'s dependencies: `brew install wxmac cairo poppler`  Notice that `openjpeg` is now installed, since it's a recursive dep.  Other recursive deps are poured too.

Now, do `brew install --build-from-source diff-pdf`.  What should happen is that `diff-pdf` starts being built forthwith (all of its deps are already installed).  What actually happens is that `brew` starts trying to build `cmake`, `doxygen`, and `sphinx-doc`, because those are deps of `openjpeg`.  But, those are *:build* deps of `openjpeg`, and `openjpeg` has already been poured and is therefore already built.

My *guess* is that `brew` is simply failing to filter out the ":build" deps from its list of things it needs to do before it can build `diff-pdf`.  Even though `diff-pdf` is going to be built rather than poured, that doesn't mean that all of its deps need to be :built.

apjanke:

-----

This avoids needlessly installing `:build` dependencies of dependencies which are not themselves being built from source.

Fixes #1272. See it for details of the problematic behavior.

Before:

```
$ brew info diff-pdf | grep Build:
Build: pkg-config, automake, autoconf
$ brew info openjpeg | grep Build:
Build: cmake, doxygen
$ brew install -s diff-pdf
==> Installing dependencies for diff-pdf: cmake, doxygen
==> Installing diff-pdf dependency: cmake
```

`cmake` and `doxygen` are `:build` deps of `openjpeg`, which is not being built from source in this scenario, so they shouldn't be picked up.

After:

```
$ brew install -s diff-pdf
==> Using the sandbox
```

(No attempt to install `cmake` or `doxygen`.)

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -114,7 +114,7 @@ class FormulaInstaller
 
   def install_bottle_for?(dep, build)
     return pour_bottle? if dep == formula
-    return false if build_from_source?
+    return false if ARGV.build_formula_from_source?(dep)
     return false unless dep.bottle && dep.pour_bottle?
     return false unless build.used_options.empty?
     return false unless dep.bottle.compatible_cellar?


Library/Homebrew/formula_installer.rb
```diff
@@ -393,7 +393,8 @@ def expand_dependencies(deps)
 
       if (dep.optional? || dep.recommended?) && build.without?(dep)
         Dependency.prune
-      elsif dep.build? && install_bottle_for?(dependent, build)
+      elsif dep.build? && (install_bottle_for?(dependent, build) ||
+           !ARGV.build_formula_from_source?(dependent))
```
MikeMcQuaid: May be worth pulling this and other `(...||...)` into a separate function or variable.

MikeMcQuaid: Yeh, that sounds more correct.

ilovezfs:
Will this cause it not to install missing run-time deps? :(

Library/Homebrew/formula_installer.rb
```diff
@@ -114,7 +114,7 @@ def pour_bottle?(install_bottle_options = { warn: false })
 
   def install_bottle_for?(dep, build)
     return pour_bottle? if dep == formula
-    return false if build_from_source?
+    return false if ARGV.build_formula_from_source?(dep)
```
apjanke: I don't know. The existing `build_from_source?` method is an accessor on `@build_from_source`, which is twiddled from `cmd/install.rb` and the `DependencyInstaller` subclass. So unifying them would be a more extensive refactor of `FormulaInstaller` and `cmd/install.rb`. Could simplify things, but I'm reluctant to do the refactoring here, because it would change the control flow pattern for `Homebrew#install` a bit. The context in which they're used (the main formula vs a discovered dependency) is different enough that the existing control flow seems more readable to me.

apjanke:
> Will this cause it not to install missing run-time deps? :(

No, they still get picked up. They just no longer look like they're being built from source inside the dependency expansion logic (unless they have explicitly been requested to be built from source).

Here's an example with `diff-pdf`'s dependency `wxmac` locally converted to a run-time dependency.

```
$ brew rm wxmac jpeg
Uninstalling /usr/local/Cellar/wxmac/3.0.2_3... (809 files, 23.6M)
Uninstalling /usr/local/Cellar/jpeg/8d... (19 files, 731.5K)
[/usr/local/Homebrew on ⇄ prune-build-deps-of-deps]
$ brew install -s diff-pdf
==> Installing dependencies for diff-pdf: jpeg, wxmac
==> Installing diff-pdf dependency: jpeg
```

Or is there something else I'm overlooking?

Library/Homebrew/formula_installer.rb
MikeMcQuaid: Ok, that seems fine, thanks.

MikeMcQuaid:
Thanks @apjanke!

PR #1307, opened 2016-10-17
Formatter: identifier: only reset color, not full styling

apjanke:

-----

`Formatter.identifier` does a full `reset` to undo the green it applies. This also clobbers any bold/underline/italic or other styling, which could be a problem if it is called inside a span of text which has other styling on it.

You can see this when doing a `brew install` which has multiple dependencies. The `oh1` line should be all bold. But only the first dependency is bold; after that the line gets reset to normal text.

before

This PR changes `Formatter.identifier` to reset only the color of the text. It's not a perfect restoration of state, but it's better than doing a full clobber.

After:
after

Note that now, all the dependencies are listed in bold.

### DIFF
--- a/Library/Homebrew/utils/formatter.rb
+++ b/Library/Homebrew/utils/formatter.rb
@@ -12,7 +12,7 @@ module Formatter
   end
 
   def identifier(string)
-    "#{Tty.green}#{string}#{Tty.reset}"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @apjanke!

PR #1308, opened 2016-10-17
brew info: include options to dependencies in display

apjanke:

-----

Currently, `brew info` does not take the options of dependencies in to account when displaying them.

For example, `libgroove` depends on `ffmpeg --with-libvorbis`, not plain `ffmpeg`. But the `brew info` output looks like it needs plain `ffmpeg`.

Before:

before

This PR changes `brew info` to list the options for dependencies which have them, and to take those options in to account when determining whether the dependency is satisfied.

After:

after

Note that not only does it add `--with-libvorbis` to the display, but it's now a red X, which I think is correct here, because this is run on a system with the default `ffmpeg` formula installed, and `--with-libvorbis` was not included in the build options. Now the display correctly reflects a formula which will need to be (re-)installed during the `brew install libgroove` process.

Other formulae you can test this on:

anjuta.rb:  depends_on "libxml2" => "with-python"
cmu-sphinxbase.rb:  depends_on "libsamplerate" => "with-libsndfile"
freeling.rb:  depends_on "boost" => "with-icu4c"
git-ftp.rb:  depends_on "curl" => "with-libssh2"
libgroove.rb:  depends_on "ffmpeg" => "with-libvorbis"

If people are in favor of this change, similar options display could be added to `brew deps`.

### DIFF
--- a/Library/Homebrew/cmd/info.rb
+++ b/Library/Homebrew/cmd/info.rb
@@ -165,9 +165,13 @@ module Homebrew
 
   def decorate_dependencies(dependencies)
     deps_status = dependencies.collect do |dep|
-      dep.installed? ? pretty_installed(dep) : pretty_uninstalled(dep)
+      if dep.satisfied?([])
+        pretty_installed(dep_display_s(dep))
+      else
+        pretty_uninstalled(dep_display_s(dep))
+      end
     end
-    deps_status * ", "
+    deps_status.join(", ")
   end
 
   def decorate_requirements(requirements)
@@ -176,5 +180,10 @@ module Homebrew
       req.satisfied? ? pretty_installed(req_s) : pretty_uninstalled(req_s)
     end
     req_status.join(", ")
+  end
+
+  def dep_display_s(dep)
+    return dep.name if dep.option_tags.empty?
+    "#{dep.name} #{dep.option_tags.map { |o| "--#{o}" }.join(" ")}"
   end
 end


MikeMcQuaid:
Seems good 👍 

> If people are in favor of this change, similar options display could be added to brew deps.

This will likely need to be a non-default option as people are likely relying on parsing the current format and we want `brew install $(brew deps foo)` to work.

Library/Homebrew/cmd/info.rb
```diff
@@ -165,9 +165,9 @@ def info_formula(f)
 
   def decorate_dependencies(dependencies)
     deps_status = dependencies.collect do |dep|
-      dep.installed? ? pretty_installed(dep) : pretty_uninstalled(dep)
+      dep.satisfied?([]) ? pretty_installed(dep_display_s(dep)) : pretty_uninstalled(dep_display_s(dep))
```
MikeMcQuaid: Probably worth splitting onto multiple lines now.

Library/Homebrew/cmd/info.rb
```diff
@@ -177,4 +177,12 @@ def decorate_requirements(requirements)
     end
     req_status.join(", ")
   end
+
+  def dep_display_s(dep)
+    if dep.option_tags.empty?
+      dep.name
```
MikeMcQuaid: `return dep.name if dep.option_tags.empty?` to avoid the `else`.

apjanke:
> This will likely need to be a non-default option as people are likely relying on parsing the current format and we want `brew install $(brew deps foo)` to work. 

Yeah, I was thinking like a `--verbose` option for the main listings. It could probably be on by default in `--tree` mode, since that's unlikely to be parsed, though maybe it should be optional there too for consistency.

Library/Homebrew/cmd/info.rb
apjanke: What's the right style for a muti-line trinary expression? Or should it be converted to an `if ... else ... end`?

Library/Homebrew/cmd/info.rb
apjanke: Amended to fix.

Library/Homebrew/cmd/info.rb
MikeMcQuaid: Yeh, the latter, thanks.

MikeMcQuaid:
> Yeah, I was thinking like a --verbose option for the main listings. It could probably be on by default in --tree mode, since that's unlikely to be parsed, though maybe it should be optional there too for consistency.

Sounds good.

Library/Homebrew/cmd/info.rb
apjanke: Amended to do so.

MikeMcQuaid:
@apjanke :shipit:?

apjanke:
Merged!

apjanke:
Merged!

PR #1317, opened 2016-10-18
Refactor `brew cask style` not to require `.rubocop.yml`.

reitermarkus:

-----

Since `rubocop-cask` already includes a `default.yml`, we don't need to have a separate YAML file inside `caskroom/cask`. This also fixes the rare case that `brew cask style` fails when `caskroom/cask` is untapped, this is useful so we can have tests for `brew cask style` without having `caskroom/cask` tapped.

--

@Homebrew/cask

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cli/style.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/style.rb
@@ -54,15 +54,15 @@ module Hbc
       end
 
       def default_args
-        ["--format", "simple", "--force-exclusion", "--config", rubocop_config]
+        [
+          "--require", "rubocop-cask",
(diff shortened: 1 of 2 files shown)


reitermarkus:
Had to add `--config /dev/null` to make sure that only `rubocop-cask`'s `default.yml` is used. Not sure if there is a better/official way to force using the default configuration.

jawshooah:
@reitermarkus Went ahead and asked for you: https://github.com/bbatsov/rubocop/issues/3647

jawshooah:
Merging for now since this works as-is, but will follow up in a separate PR if necessary once we hear back on bbatsov/rubocop#3647.

PR #1330, opened 2016-10-19
Refactor `spec_helper`.

reitermarkus:

-----

This adds Homebrew's test config to the load path and gets rid of most of the directory overrides.

--

@Homebrew/cask

### DIFF
--- a/Library/Homebrew/cask/spec/spec_helper.rb
+++ b/Library/Homebrew/cask/spec/spec_helper.rb
@@ -6,23 +6,19 @@ if ENV["HOMEBREW_TESTS_COVERAGE"]
   require "simplecov"
 end
 
-project_root = Pathname.new(File.expand_path("../..", __FILE__))
-
 # add Homebrew to load path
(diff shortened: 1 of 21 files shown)


Library/Homebrew/cask/spec/spec_helper.rb
```diff
@@ -6,59 +6,35 @@
   require "simplecov"
 end
 
-project_root = Pathname.new(File.expand_path("../..", __FILE__))
-
 # add Homebrew to load path
 $LOAD_PATH.unshift(File.expand_path("#{ENV["HOMEBREW_REPOSITORY"]}/Library/Homebrew"))
+$LOAD_PATH.unshift(File.expand_path("#{ENV["HOMEBREW_REPOSITORY"]}/Library/Homebrew/test/lib"))
```
jawshooah: May as well change this and the line above to use `HOMEBREW_REPOSITORY.join(…)`

jawshooah: Ah, right. Never mind then

PR #1333, opened 2016-10-19
Compare names not formulae

vladshablinsky:

-----

* Causes a bug in Formula#installed_alias_target_changed? when
  `Formula#superseds_an_installed_formula?` returns true because
  `Formula#old_installed_formulae` includes f for some Formula f.

* Causes a bug when foo@2.4 with alias foo has HEAD or devel version and
  we try to `brew upgrade foo --devel|--HEAD` from stable. The upgrade fails
  while since we installing formula to the same prefix it's alredy installed.
  The reason for that is that we use
  `formula_to_install = outdated.map(&:latest_formula)` in cmd/upgrade
  before calling `upgrade_formula` on foo.

  ```ruby
  def latest_formula
    installed_alias_target_changed? ? current_installed_alias_target : self
  end
  ```

  `Formula#installed_alias_target_changed?` compares formulae using
  `Formula#==`, which is wrong for this case, thus `Formula#latest_formula` doesn't
  return self and returns `Formula#current_installed_alias_target` with spec
  foo was initially installed instead of devel or HEAD, causing the error.

### Reproduce

1. Make `foo@2.4` with stable and devel specs.
2. Make an alias `foo` to `foo@2.4`
3.  `brew install foo`
4.  `brew upgrade foo --devel`

Output looks like this
```bash
➜ vlad:Aliases$ brew upgrade mack --HEAD                                     master ✗
mack 2.4
==> Upgrading mack
==> Using the sandbox
==> Downloading http://beyondgrep.com/ack-2.14-single-file
Already downloaded: /Users/vlad/Library/Caches/Homebrew/mack@2.4-2.4.14-single-file
==> pod2man /usr/local/Cellar/mack@2.4/2.4/bin/ack ack.1
Error: File exists - /usr/local/Cellar/mack@2.4/2.4/.brew
```

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1167,7 +1167,7 @@ class Formula
   # Returns false if the formula wasn't installed with an alias.
   def installed_alias_target_changed?
     target = current_installed_alias_target
-    target && target != self
+    target && target.name != name
   end
 
   # Is this formula the target of an alias used to install an old formula?
@@ -1192,7 +1192,7 @@ class Formula
     # it doesn't make sense to say that other formulae are older versions of it
     # because we don't know which came first.
     return [] if alias_path.nil? || installed_alias_target_changed?
-    self.class.installed_with_alias_path(alias_path) - [self]
+    self.class.installed_with_alias_path(alias_path).reject { |f| f.name == name }
   end
 
   # @private
--- a/Library/Homebrew/test/test_formula.rb
+++ b/Library/Homebrew/test/test_formula.rb
@@ -799,8 +799,8 @@ end
 class AliasChangeTests < Homebrew::TestCase
   attr_reader :f, :new_formula, :tab, :alias_path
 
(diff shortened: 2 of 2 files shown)


vladshablinsky:
Also, I wonder if we need to preserve spec here: 

```ruby
# formula.rb#L257-L260

  def installed_alias_path
    path = build.source["path"] if build.is_a?(Tab)
    path if path =~ %r{#{HOMEBREW_TAP_DIR_REGEX}/Aliases}
  end

# formula.rb#L1162-L1164

  def current_installed_alias_target
    Formulary.factory(installed_alias_path) if installed_alias_path
  end


```
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L1162-L1164

vladshablinsky:
CC @MikeMcQuaid

MikeMcQuaid:
Makes sense, thanks @vladshablinsky!

PR #1342, opened 2016-10-21
Move cask's fixtures to `test/fixtures`.

reitermarkus:

-----

Move cask's fixtures as another step before merging test suites.


Library/Homebrew/test/helper/fixture.rb
```diff
@@ -0,0 +1,13 @@
+module Fixture
+  module_function
+
+  FIXTURE_DIR = File.expand_path(File.join(__FILE__, "..", "..", "fixtures")).freeze
+
+  def path(name)
```
MikeMcQuaid: `HOMEBREW_TEST_FIXTURES` as a constant should be sufficient and can be joined with `/` rather than needing a new method and class defined.

Library/Homebrew/test/helper/fixture.rb
```diff
@@ -0,0 +1,13 @@
...
+    File.join(FIXTURE_DIR, name)
+  end
+
+  def url(name)
+    "file://".concat(path(name))
```
MikeMcQuaid: String interpolation makes more sense to me here. Additionally, this should live in Cask's test_helper or similar as `fixture_url` and cask-specific fixtures should live in `test/fixtures/cask/`

MikeMcQuaid:
> before merging test suites.

Can you detail a bit how you're planning on doing this? Of particular concern is that Cask's test suite is in RSpec and MiniTest and takes almost twice as long to run.

reitermarkus:
> Can you detail a bit how you're planning on doing this?

First, I will move the current tests to a directory structure, so we can easily exclude `os/mac/**`, and later `cask/**` from `--generic` tests. The `--only=` option will have to be refactored to support this, but that's about it.

> Cask's test suite is in RSpec and MiniTest and takes almost twice as long to run.

Twice as long compared to what?

MikeMcQuaid:
> First, I will move the current tests to a directory structure, so we can easily exclude os/mac/**, and later cask/** from --generic tests. The --only= option will have to be refactored to support this, but that's about it.

Sounds great.

> Twice as long compared to what?

`brew tests`

Related to both of these: I'm more interested in how you're going to reconcile the use of both MiniTest and RSpec?

reitermarkus:
> `brew tests`

Really? `brew cask-tests` is faster than `brew tests` on my machine.

MikeMcQuaid:
https://bot.brew.sh/job/Brew/version=any_sierra/2571/testReport/brew-test-bot/sierra/ for with coverage (dramatically slower) https://bot.brew.sh/job/Brew/version=el_capitan/2571/testReport/brew-test-bot/el_capitan/ without (a bit slower).

reitermarkus:
I think we can merge this, as this doesn't have any effect on the speed of the tests, we'll need to look into that separately.

PR #1349, opened 2016-10-23
Merge Gemfiles.

reitermarkus:

---

### DIFF
--- a/Library/Homebrew/dev-cmd/tests.rb
+++ b/Library/Homebrew/dev-cmd/tests.rb
@@ -27,6 +27,7 @@ module Homebrew
       end
 
       ENV["BUNDLE_GEMFILE"] = "#{Dir.pwd}/test/Gemfile"
+      ENV["BUNDLE_PATH"] = "#{Dir.pwd}/test/vendor/bundle"
 
       # Override author/committer as global settings might be invalid and thus
       # will cause silent failure during the setup of dummy Git repositories.
@@ -37,7 +38,7 @@ module Homebrew
 
       Homebrew.install_gem_setup_path! "bundler"
       unless quiet_system("bundle", "check")
-        system "bundle", "install", "--path", "vendor/bundle"
+        system "bundle", "install"
       end
 
       # Make it easier to reproduce test runs.
--- a/Library/Homebrew/test/Gemfile
+++ b/Library/Homebrew/test/Gemfile
@@ -1,9 +1,13 @@
 source "https://rubygems.org"
 
-gem "mocha", "~> 1.1"
-gem "minitest", "~> 5.3"
-gem "rake", "~> 10.3"
-gem "parallel_tests", "~> 2.9"
+gem "minitest"
+gem "minitest-reporters"
+gem "mocha"
+gem "rake"
+gem "rspec"
+gem "rspec-its", require: false
+gem "rspec-wait", require: false
+gem "parallel_tests"
 
 group :coverage do
   # This is SimpleCov v0.12.0 with two fixes merged on top, that finally resolve
@@ -13,7 +17,8 @@ group :coverage do
   #  - https://github.com/colszowka/simplecov/pull/520
   gem "simplecov", "0.12.0",
     git: "https://github.com/colszowka/simplecov.git",
-    branch: "master", # commit 83d8031ddde0927f87ef9327200a98583ca18d77
+    branch: "master",
+    ref: "83d8031ddde0927f87ef9327200a98583ca18d77",
     require: false
   gem "codecov", require: false
 end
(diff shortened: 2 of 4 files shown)


Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -37,7 +37,7 @@ def tests
 
       Homebrew.install_gem_setup_path! "bundler"
       unless quiet_system("bundle", "check")
-        system "bundle", "install", "--path", "vendor/bundle"
+        system "bundle", "install"
```
MikeMcQuaid: Is this specified somewhere else now?

Library/Homebrew/test/Gemfile
```diff
@@ -1,9 +1,15 @@
 source "https://rubygems.org"
 
-gem "mocha", "~> 1.1"
-gem "minitest", "~> 5.3"
-gem "rake", "~> 10.3"
-gem "parallel_tests", "~> 2.9"
+group :test do
```
MikeMcQuaid: Given these are only used for tests this seems a bit redundant but don't have strong feelings.

Library/Homebrew/test/Gemfile
```diff
@@ -1,9 +1,15 @@
...
+  gem "minitest", "~> 5.9"
+  gem "minitest-reporters"
+  gem "mocha", "~> 1.2"
+  gem "rake", "~> 10.5"
```
MikeMcQuaid: Given a bunch of these are being bumped let's just remove all the version pinning here (we get it from `.lock` anyway) and restrict it to the cases where the newest version breaks `brew tests`

Library/Homebrew/test/Gemfile
```diff
@@ -1,9 +1,15 @@
...
+  gem "rspec", "~> 3.5.0"
```
MikeMcQuaid: As mentioned before: I'm a very strong 👎 on installing MiniTest and RSpec for the unified `brew tests`. We need to pick one test suite and use that.

Library/Homebrew/dev-cmd/tests.rb
reitermarkus: It's the default.

jawshooah: It's not the default, but there is likely a `.bundle/config` in the same directory as the `Gemfile` that specifies this behavior. It is created the first time you run `bundle install` with `--path`.

MikeMcQuaid: Indeed and conversely user global configs will override this unless specified. Better to leave as-is, I think.

Library/Homebrew/test/Gemfile
reitermarkus: I'm very much in favor of the RSpec syntax. Still, we will need to get both working first in order to transition from one to the other. And installing everything in one place is certainly better than it was before, where everything was duplicated also in `cask/vendor/bundle`.

MikeMcQuaid:
> I'm very much in favor of the RSpec syntax. Still, we will need to get both working first in order to transition from one to the other. And installing everything in one place is certainly better than it was before, where everything was duplicated also in cask/vendor/bundle.

I think it would be better to remove the RSpec dependency first before we add it to `brew tests`.

Library/Homebrew/test/Gemfile
```diff
@@ -1,9 +1,13 @@
 source "https://rubygems.org"
 
-gem "mocha", "~> 1.1"
-gem "minitest", "~> 5.3"
-gem "rake", "~> 10.3"
-gem "parallel_tests", "~> 2.9"
+gem "minitest"
```
jawshooah: We can remove this as well, since `minitest` has been part of the standard library since 1.9.3.

reitermarkus: I'd keep this, since we aren't using the one from the standard library, but a more up-to-date one.

MikeMcQuaid: I'm cool with using the newer one.

Library/Homebrew/dev-cmd/tests.rb
MikeMcQuaid: Still need this changed and then I think we're good.

MikeMcQuaid:
@reitermarkus Any news here? Thanks!

MikeMcQuaid:
@reitermarkus Any news on this one? Thanks!

reitermarkus:
I'll close this, as the gems are now already stored in the same directory for both test suites. So by merging the test suites, cask's Gemfile will eventually be unnecessary anyway.

MikeMcQuaid:
Thanks @reitermarkus!

PR #1353, opened 2016-10-23
Use proper path to `caskroom/cask` tap in `test_helper`.

reitermarkus:

-----

`HOMEBREW_LIBRARY` is pointing to a path in `TEST_TMPDIR`.

### DIFF
--- a/Library/Homebrew/cask/test/test_helper.rb
+++ b/Library/Homebrew/cask/test/test_helper.rb
@@ -50,7 +50,7 @@ Hbc.default_tap = Tap.fetch("caskroom", "test").tap do |tap|
 end
 
 # pretend that the caskroom/cask Tap is installed
-FileUtils.ln_s HOMEBREW_LIBRARY.join("Taps", "caskroom", "homebrew-cask"), Tap.fetch("caskroom", "cask").path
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cask/test/test_helper.rb
```diff
@@ -50,7 +50,7 @@ class TestCask < Cask; end
 end
 
 # pretend that the caskroom/cask Tap is installed
-FileUtils.ln_s HOMEBREW_LIBRARY.join("Taps", "caskroom", "homebrew-cask"), Tap.fetch("caskroom", "cask").path
+FileUtils.ln_s Pathname.new(ENV["HOMEBREW_LIBRARY"]).join("Taps", "caskroom", "homebrew-cask"), Tap.fetch("caskroom", "cask").path
```
jawshooah: This feels like the wrong approach. Are we using the `HOMEBREW_LIBRARY` environment variable instead of the constant somewhere in the cask code?

reitermarkus: Well, we need to link the *actual* `caskroom/cask` tap, and since `HOMEBREW_LIBRARY` is pointing to the one in the test environment, we need to use the environment variable used by `brew cask-tests`. Sooner or later, once we have converted the tests to only use the `test` tap, we can get rid of this completely.

jawshooah: Ah, right, I forgot we still have tests that use actual casks :disappointed:

PR #1355, opened 2016-10-23
Fix locally failing `depends_on` test.

reitermarkus:

-----

`[*arr] != Array(*arr)`

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/dsl/depends_on.rb
+++ b/Library/Homebrew/cask/lib/hbc/dsl/depends_on.rb
@@ -50,8 +50,7 @@ module Hbc
       def load(pairs = {})
         pairs.each do |key, value|
           raise "invalid depends_on key: '#{key.inspect}'" unless VALID_KEYS.include?(key)
-          writer_method = "#{key}=".to_sym
-          @pairs[key] = send(writer_method, value)
+          @pairs[key] = send(:"#{key}=", *value)
         end
       end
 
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cask/lib/hbc/dsl/depends_on.rb
```diff
@@ -84,23 +84,23 @@ def cask=(*arg)
...
-        macos = if arg.count == 1 && arg.first =~ /^\s*(<|>|[=<>]=)\s*(\S+)\s*$/
+        macos = if args.count == 1 && args.first =~ /^\s*(<|>|[=<>]=)\s*(\S+)\s*$/
           raise "'depends_on macos' comparison expressions cannot be combined" unless @macos.empty?
           operator = Regexp.last_match[1].to_sym
           release = self.class.coerce_os_release(Regexp.last_match[2])
           [[operator, release]]
         else
           raise "'depends_on macos' comparison expressions cannot be combined" if @macos.first.is_a?(Symbol)
-          [*arg].map(&self.class.method(:coerce_os_release)).sort
+          Array(*args).map(&self.class.method(:coerce_os_release)).sort
```
jawshooah: Not sure this is right:

```
>> arr = [1,2,3]
=> [1, 2, 3]
>> Array(*arr)
ArgumentError: wrong number of arguments (3 for 1)
	from (irb):2:in `Array'
	from (irb):2
	from /Volumes/git/github/brew/Library/Homebrew/brew.rb:94:in `<main>'
>> [*arr]
=> [1, 2, 3]
```

Why not just use `args`?

reitermarkus: I changed this to `[*arr]` today in the RuboCop PR, and after that I got errors locally.

I think this is due to the following scenario:

```ruby
# in cask

depends_on macos: ["10.9", "10.10"]

# inside `macos=`

args = [["10.9", "10.10"]]

Array(*args) == ["10.9", "10.10"]
[*args] == [["10.9", "10.10"]]

```

jawshooah: Actually, it looks like the only reason `Array(*args)` works here and elsewhere in this file is that `DependsOn#load` always calls the writer method with a single argument (hence the name `arg` rather than `args`), which in this case is an array.

So if a cask had `depends_on macos: ['10.8', '10.9']`, this method would get called with `send(:macos=, ['10.8', '10.9'])`, so `args` would be `[['10.8', '10.9']]`, and `Array(*args)` would be `['10.8', '10.9']`.

The best way to disambiguate this, I think, would be to remove the splat from `arg` in the method signature and explicitly convert it to an array at the beginning of the method with `Array(*arg)`. Same applies to the other `key=` methods.

jawshooah: Ha, didn't see your reply. We're on the same page :+1:

reitermarkus: I changed the `load` method to also use the splat operator instead.

PR #1373, opened 2016-10-25
build: copy formula file to logs

zmwangx:

-----

It's useful to include the exact formula file at the time of build in `gist-logs`:

- Users could be building an outdated version of the formula, which we can't tell from gist-logs (`00.config.out` includes the latest core commit, sure, but (1) it's takes effort to pull up the formula at that commit; (2) it's easily masked by a `brew update` post build);

- Occasionally users make modifications to formulae they're building, see https://github.com/Homebrew/homebrew-core/issues/6232 for instance.

### DIFF
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -132,6 +132,9 @@ class Build
       else
         formula.prefix.mkpath
 
+        formula.logs.mkpath
+        FileUtils.cp formula.path, formula.logs/"00.#{formula.name}.rb"
+
         formula.install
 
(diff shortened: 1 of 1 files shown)


apjanke:
👍 Would be nice to have this extra info for debugging.

MikeMcQuaid:
Rather than duplicating every time can we consider only copying it in full if it's modified from `origin/master` and writing the file commit instead? I'm not sure we ever actually need the file as the answer to a modified version is: unmodify it and to an outdated version: `brew update`.

PR #1374, opened 2016-10-25
Do not remove unknown bottle tags

sjackman:

---

Building bottles for Sierra in Homebrew-Science is removing the `:x86_64_linux` bottles for Linuxbrew. See for example https://github.com/Homebrew/homebrew-science/commit/db6b3ed8b40b39c4dc423f6422150b402d81d917

``` diff
--- a/a5.rb
+++ b/a5.rb
@@ -14,10 +14,10 @@ class A5 < Formula

   bottle do
     cellar :any_skip_relocation
+    sha256 "5694a2ae1e9cf95c9c4fdcc2dcb23069c89fc305a05e18f4c3f759caa56c2124" => :sierra
     sha256 "712a52ae946e36d3410f6ea98425605aceffebaa738a397e6721100de6474e12" => :el_capitan
     sha256 "589bb490b43853bd8bbb0817018141efd3e65794844fbcfaf950d6ea55d4d7bb" => :yosemite
     sha256 "c6ae86aad65c91dbcf2b647acdbd25ac5672f8c04f34e7aa9fb991992d9e8fbd" => :mavericks
-    sha256 "174ba48713d53b612375e2ac4b850455625a276a9078edb53f9e02882b11bfc4" => :x86_64_linux
   end

   def install
```

### DIFF
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -336,20 +336,19 @@ class BottleSpecification
   end
 
   def checksums
-    checksums = {}
-    os_versions = collector.keys
-    os_versions.map! do |macos|
+    tags = collector.keys.sort_by do |tag|
+      # Sort non-MacOS tags below MacOS tags.
       begin
-        MacOS::Version.from_symbol macos
+        MacOS::Version.from_symbol tag
       rescue
-        nil
+        "0.#{tag}"
       end
-    end.compact!
-    os_versions.sort.reverse_each do |os_version|
-      macos = os_version.to_sym
-      checksum = collector[macos]
+    end
+    checksums = {}
+    tags.reverse_each do |tag|
+      checksum = collector[tag]
       checksums[checksum.hash_type] ||= []
-      checksums[checksum.hash_type] << { checksum => macos }
+      checksums[checksum.hash_type] << { checksum => tag }
     end
     checksums
   end


MikeMcQuaid:
Can you explain a bit more about this fix?

sjackman:
Sure thing. When a bottle is built for Sierra, it's removing any bottle architectures not found in [`OS::Mac::Version::SYMBOLS`](https://github.com/sjackman/brew/blob/8d451b4a6c0559daa4e849812e11d68ae3c20b38/Library/Homebrew/os/mac/version.rb#L6). The result is a commit like https://github.com/Homebrew/homebrew-science/commit/db6b3ed8b40b39c4dc423f6422150b402d81d917 that adds a bottle for `:sierra` and removes the bottle for `:x86_64_linux`. This patch modifies `--keep-old` to keep all old bottle tags, including those not found in `OS::Mac::Version::SYMBOLS`.

MikeMcQuaid:
Ok, makes sense. I'm 👍 to fixing that but I'm not sure this is the right location for a fix.

sjackman:
This particular function `checksums` is unnecessarily clunky. It converts symbols to strings, sorts strings, then convert strings back to symbols, and loses symbols with no entry in `OS::Mac::Version::SYMBOLS` as a side effect (not so much intentional behaviour, I don't think). I could rewrite this function to simply sort the symbols using a different comparison operator, which would make this function tidier and have the desired behaviour effect of this PR. Would that suit you?

sjackman:
Alternatively (or in addition) I could change [`OS::Mac::Version.from_symbol`](https://github.com/sjackman/brew/blob/8d451b4a6c0559daa4e849812e11d68ae3c20b38/Library/Homebrew/os/mac/version.rb#L18) not to raise an error on an unknown symbol but return `to_s`, but I'm less comfortable with that change.

MikeMcQuaid:
> I could rewrite this function to simply sort the symbols using a different comparison operator, which would make this function tidier and have the desired behaviour effect of this PR. Would that suit you?

Yes, something like that would be good, agreed with the clunkiness.

MikeMcQuaid:
@sjackman Any news here?

sjackman:
No progress yet. Perhaps by the end of the week I can find some time for this issue.

sjackman:
I've updated this PR to use `sort_by` to avoid the round-trip conversion from symbol to string and then back to symbol.

Library/Homebrew/software_spec.rb
```diff
@@ -336,20 +336,18 @@ def checksum_for(tag)
   end
 
   def checksums
-    checksums = {}
-    os_versions = collector.keys
-    os_versions.map! do |macos|
+    tags = collector.keys.sort_by do |tag|
```
MikeMcQuaid: Does this maintain the current ordering?

Library/Homebrew/software_spec.rb
```diff
@@ -336,20 +336,18 @@ def checksum_for(tag)
...
       begin
-        MacOS::Version.from_symbol macos
+        MacOS::Version.from_symbol tag
       rescue
-        nil
+        "0.#{tag}"
```
MikeMcQuaid: Can you elaborate more on what this does and why? It's pretty cryptic.

Library/Homebrew/software_spec.rb
sjackman: Yes, it does maintain the current ordering, which is reverse order by version number.

MikeMcQuaid: Thanks.

Library/Homebrew/software_spec.rb
MikeMcQuaid: It looks like this means it's changing the actual tag that's returned by this function, though, which doesn't seem to be desirable. I think the higher-level refactoring you mentioned earlier would be preferable; this reads like a bit of a "hack to make it work" unfortunately.

MikeMcQuaid: Ok, sorry, I see this is a `sort_by` and misread as a `map`. Could you add a comment to explain this above the rescue and then I think we're good to go.

sjackman: Done.

MikeMcQuaid:
Thanks @sjackman!

sjackman:
Thanks for merging, Mike!

PR #1383, opened 2016-10-27
pathname: add support for creating app bundles.

midchildan:

-----

These changes introduces helper methods to create a `.app` bundle for unix GUI apps. It makes a  wrapper script that does things like:
- Add `#{HOMEBREW_PREFIX}/bin` to `PATH`
- Set `GTK_PATH`
- Setup locale-related environment  variables

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -431,6 +431,117 @@ class Pathname
     EOS
   end
 
+  # Writes an exec script intended for use by unix GUI apps.
+  def write_gui_script(*command, **options)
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Homebrew doesn't support GUI applications well so I don't think we want to go down this route as I'd rather we scaled this back rather than doubled-down. A better approach is to work with upstream (or create yourself) `.app` bundles that can be used with Homebrew Cask (and without Homebrew at all).

midchildan:
Okay, closing this PR. But not using Homebrew is difficult due to all the dependencies that has to be dealt with, so I'll submit a formula without the GUI wrappers.

PR #1364, opened 2016-10-24
Reorder and rename test files.

reitermarkus:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

---

The `--only=` option was changed to account for the new directory-based structure. So e. g. if you call `brew tests --only=utils` it will call `utils_test.rb` **and** all tests under the `utils/` directory.

PR #1400, opened 2016-10-29
Move `test/vendor/bundle` to `vendor/bundle`.

reitermarkus:

-----

This is to clean up the `test` directory a bit as per https://github.com/Homebrew/brew/pull/1364#issuecomment-255860189.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,15 +8,13 @@
 !/Library/
 
 # Ignore generated files within `Library` (again).
+/Library/Homebrew/**/.bundle
+/Library/Homebrew/**/vendor/bundle
 /Library/Homebrew/doc
-/Library/Homebrew/cask/.bundle
 /Library/Homebrew/cask/bin
-/Library/Homebrew/cask/vendor
 /Library/Homebrew/cask/coverage
 /Library/Homebrew/cask/tmp
-/Library/Homebrew/test/.bundle
 /Library/Homebrew/test/bin
-/Library/Homebrew/test/vendor
 /Library/Homebrew/test/coverage
 /Library/Homebrew/test/fs_leak_log
 /Library/Homebrew/tmp
@@ -26,7 +24,7 @@
 /Library/PinnedTaps
 /Library/Taps
 
-# Ignore vendored files within `Library`
+# Ignore vendored files within `Library`.
 /Library/Homebrew/vendor/portable-ruby
 
 # Ignore `bin` contents (again).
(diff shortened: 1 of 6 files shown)


MikeMcQuaid:
> Also not sure why vendor/bundle has to live inside test. Any reason not to move it up a level?

I missed that comment at the time but: because these gems only apply to the test suite it feels strange to have it in `Library/Homebrew`. It's particularly unusual for the `vendor/bundle` to live in the directory above the `Gemfile`? I'm mostly on vacation so it'd be good to hold off a bit on moving some of these things around. Ideally I think this PR would be reverted (although doesn't need to happen today).

MikeMcQuaid:
Then again, if this is literally the only way to avoid that glob causing issues then so be it (but feels pretty weird considering the layout of most Ruby projects that would presumably have an identical problem).

MikeMcQuaid:
@reitermarkus @jawshooah Would like your thoughts on above.

jawshooah:
> It's particularly unusual for the `vendor/bundle` to live in the directory above the `Gemfile`

Yes it is, I forgot that the `Gemfile` was still in `test`.

> Because these gems only apply to the test suite it feels strange to have it in `Library/Homebrew`

I disagree only because the vast majority of Ruby projects I've seen place the `Gemfile` in the root directory of the project, regardless of whether any gems are used outside of tests. Don't feel particularly strongly about this though.

reitermarkus:
I also have only every seen projects that have the `Gemfile` at the same level as the `test` directory, but not inside it. And since we already have a `vendor` directory at the root anyways, why not also use this for `bundle`?

Hombrew's layout is very different from any “standard” Ruby project, considering that most have their load path in `/lib`, while we have it in `Library/Hombrew`, and our `test` directory is also accessibly within this same load path, while the `test` directory for every other project I've seen is completely separated from the library path.

MikeMcQuaid:
Homebrew is different to most projects using `Gemfile`s in that we don't use any libraries at runtime. These gems are specific to our tests so I think it makes sense for it to live there. 

> if this is literally the only way to avoid that glob causing issues then so be it

@reitermarkus Can you elaborate if this is the case?

reitermarkus:
It's not the only way, but by far the simplest.

Another solution would be to move the tests to `test/test` as in the referenced PR, which you said looks weird.

Another way, which I dislike, would be to first glob all, and then remove all test's in `vendor/bundle` from the first glob.

reitermarkus:
Anothing solution which just came to mind would be to just use the default, `test/.bundle`, since dotfiles are not globbed by default.

MikeMcQuaid:
I think I'm fine with this as-is for now, thanks though.

PR #1408, opened 2016-10-30
Add :node special dependency

torifat:
screenshot 2016-10-30 19 46 31

### DIFF
--- a/Library/Homebrew/requirements.rb
+++ b/Library/Homebrew/requirements.rb
@@ -14,6 +14,7 @@ require "requirements/tuntap_requirement"
 require "requirements/unsigned_kext_requirement"
 require "requirements/x11_requirement"
 require "requirements/emacs_requirement"
+require "requirements/node_requirement"
 
 class XcodeRequirement < Requirement
   fatal true
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -123,8 +123,8 @@ class DependencyCollector
     when :ld64       then LD64Dependency.new if MacOS.version < :leopard
     # Tiger doesn't ship expat in /usr/lib
     when :expat      then Dependency.new("expat", tag) if MacOS.version < :leopard
-    when :python2
-      PythonRequirement.new(tags)
+    when :node       then NodeRequirement.new(tags)
+    when :python2    then PythonRequirement.new(tags)
     else
       raise ArgumentError, "Unsupported special dependency #{spec.inspect}"
     end
--- a/Library/Homebrew/requirements/node_requirement.rb
+++ b/Library/Homebrew/requirements/node_requirement.rb
@@ -1,0 +1,36 @@
+require "language/node"
+
+class NodeRequirement < Requirement
+  fatal true
+  default_formula "node"
+
+  def initialize(tags)
+    @version = tags.shift if /^\d+\.\d+$/ =~ tags.first
+    raise "NodeRequirement requires a version!" unless @version
+    super
+  end
+
+  satisfy build_env: false do
+    next false unless which "node"
+    node_version = Utils.popen_read("node", "--version")[/v(\d+\.\d+)(?:\.\d+)?/, 1]
+    Version.create(node_version) >= Version.create(@version)
+  end
+
+  def message
+    s = "Node #{@version} is required to install this formula."
+    s += super
+    s
+  end
+
+  def inspect
+    "#<#{self.class.name}: #{name.inspect} #{tags.inspect} version=#{@version.inspect}>"
+  end
+
+  def display_s
+    if @version
+      "#{name} >= #{@version}"
+    else
+      name
+    end
+  end
+end
--- a/Library/Homebrew/requirements/node_requirement.rb
+++ b/Library/Homebrew/requirements/node_requirement.rb
@@ -1,5 +1,3 @@
-require "language/node"
-
 class NodeRequirement < Requirement
   fatal true
   default_formula "node"
--- a/Library/Homebrew/test/test_node_requirement.rb
+++ b/Library/Homebrew/test/test_node_requirement.rb
@@ -1,0 +1,38 @@
+require "testing_env"
+require "requirements/node_requirement"
+require "fileutils"
+
+class NodeRequirementTests < Homebrew::TestCase
+  def setup
+    @dir = Pathname.new(mktmpdir)
+    (@dir/"bin/node").write <<-EOS.undent
+      #!/bin/bash
+      echo v5.0
+    EOS
+    FileUtils.chmod 0755, @dir/"bin/node"
+  end
+
+  def teardown
+    FileUtils.rm_rf @dir
+  end
+
+  def test_raises_for_missing_version_tag
+    assert_raises(RuntimeError) { NodeRequirement.new([]) }
+  end
+
+  def test_raises_for_non_version_tag
+    assert_raises(RuntimeError) { NodeRequirement.new(%w[test]) }
+  end
+
+  def test_satisfied
+    with_environment("PATH" => @dir/"bin") do
+      assert_predicate NodeRequirement.new(%w[4.0]), :satisfied?
+    end
+  end
+
+  def test_not_satisfied
+    with_environment("PATH" => @dir/"bin") do
+      refute_predicate NodeRequirement.new(%w[6.0]), :satisfied?
+    end
+  end
+end


Library/Homebrew/compat/requirements.rb
```diff
@@ -17,3 +17,4 @@
 TuntapDependency           = TuntapRequirement
 X11Dependency              = X11Requirement
 ConflictsWithBinaryOsxfuse = NonBinaryOsxfuseRequirement
+NodeRequirement            = NodeRequirement
```
reitermarkus: There is no need to add a compatibility layer for a new requirement.

torifat: Ok, thanks for pointing it out 😄

torifat:
@reitermarkus after remove `NodeRequirement = NodeRequirement`. One build failed 😕

Library/Homebrew/requirements.rb
```diff
@@ -125,3 +125,9 @@ class GitRequirement < Requirement
   default_formula "git"
   satisfy { Utils.git_available? }
 end
+
+class NodeRequirement < Requirement
+  fatal true
+  default_formula "node"
+  satisfy { which("node") }
+end
```
MikeMcQuaid: Please add this to a separate `node_requirement.rb` file and, like the `JavaRequirement`, support using a `version` argument. This should perhaps be compulsory like the `RubyRequirement.

torifat: Ok, Thanks 👍

torifat:
@MikeMcQuaid I'm not good at Ruby. Can you please guide me here? I have implemented the version checking and it's working.

But, `brew` still trying to install `node` even if return value of `satisfy build_env: false do` is `true`.

screenshot 2016-11-03 02 35 09

Library/Homebrew/requirements.rb
DomT4: > This should perhaps be compulsory like the `RubyRequirement`.

I wouldn't recommend this, personally.

DomT4:
> But, brew still trying to install node even if return value of satisfy build_env: false do is true.

Are you pouring the bottle of a formula to check? If so, that's expected and correct behaviour. To avoid the `default_formula` you'd need to build from source, `brew install blah -s`, etc.

torifat:
@DomT4 ah, thanks. So, the current code should work fine. Though I forgot to remove `require "language/node"` which is not needed there anymore.

torifat:
Any idea why this test fails only it Jenkins?

```
NodeRequirementTests#test_satisfied [/Users/brew/Jenkins/pr-brew/version/any_sierra/Library/Homebrew/test/test_node_requirement.rb:29]:
Expected #<NodeRequirement: "node" [] version="4.0">
 to be satisfied?.
```

torifat:
@MikeMcQuaid Can you have a look here, please?

MikeMcQuaid:
> Any idea why this test fails only in Jenkins?

Likely because there's no Node installed there. You may need to do some mocking.

As @DomT4 mentions above some things that are worth considering on this once it's 💚:

1. can something that uses `npm install` from `node` at build-time use `iojs` or a (very) different `node` version at runtime?
2. as-is all bottled formulae using `depends_on :node` will end up being `depends_on "node"` by default. This could be worked around if 1) is resolved.

torifat:
> Likely because there's no Node installed there. You may need to do some mocking.

But, I have mocked `node` and it works locally & in travis too.

https://github.com/Homebrew/brew/pull/1408/files#diff-595241f18c3635cda19d357a78dc5c52R6

1. As long as it meets the version requirement it should work.

Daniel15:
> can something that uses npm install from node at build-time use iojs or a (very) different node version at runtime?

This works most of the time, the main exception is when using native modules, where it only works if you install the package on a version of Node.js that's binary (ABI) compatible with the version you use at runtime. I think generally some package installed using npm 6.x should run on any other 6.x version of npm.

Also iojs is completely unsupported now, so it's not worth worrying about it.

MikeMcQuaid:
> In that case, I think generally some package installed using npm 6.x should run on any other 6.x version of npm.

Ok. We'll need to figure out a way to encode that into this requirement.

> The vast majority of npm packages are non-native packages that just contain JavaScript code so it doesn't matter which version of Node.js you used to install them.

Good to know.



> Also iojs is completely unsupported now, so it's not worth worrying about it.

The voice of a non-packager 😉. As we're just looking for a `node` in `$PATH` we will likely find some people/systems with weird stuff like `iojs` still around so we'll need to detect and handle these cases.

torifat:
@MikeMcQuaid `which node` should work for `iojs` too.

screenshot 2016-11-07 00 08 08

There's no version conflict too:
Node & IO versions

And, finally I mocked `node` in test but it's failing in Jenkins 😢  Working in both local & travis.

https://github.com/Homebrew/brew/pull/1408/files#diff-595241f18c3635cda19d357a78dc5c52R6

MikeMcQuaid:
> And, finally I mocked node in test but it's failing in Jenkins 😢 Working in both local & travis.

Let's not worry about that until the PR handles the other issues above. When it's working there I'll help fix the issues.

torifat:
I don't know `ruby`. I created this PR by reading other codes. Had a hard time figuring out `brew tests --only=node_requirement`. That's why I was worried. But, you guys are awesome ❤️ Thanks for the assurance 😄

ilovezfs:
Are we sure we want yet another one of these special requirements given the difficulties we've had with the others?

MikeMcQuaid:
@ilovezfs I think this may be one where we can handle bottles sensible due to this being interpreted with vendored dependencies rather than using linkage. This PR is still a bit off, though, and there's a bunch of questions to be answered and implemented.

torifat:
@MikeMcQuaid I tried to answer all the questions you asked and also pursed @Daniel15 to answer too.

> Ok. We'll need to figure out a way to encode that into this requirement.

Any suggestions regarding how we can do that?

> bunch of questions to be answered and implemented.

I'm not sure which questions are still unanswered.

MikeMcQuaid:
> As long as it meets the version requirement it should work.

We'll need confirmation that it does work and has been tested with a couple of different `depends_on "node"` formulae.

> Any suggestions regarding how we can do that?

I'm not sure. Are you saying both the Node and NPM version matter at runtime?

torifat:
> We'll need confirmation that it does work and has been tested with a couple of different depends_on "node" formulae.

I can do that for a few popular formulae.

> I'm not sure. Are you saying both the Node and NPM version matter at runtime?

Only `node` version matters. `npm` is mostly backward compatible.

IMO `brew` shouldn't handle `npm install` or `yarn install`. Formulae tarball should comes with `node_modules` like yarn tarball.

MikeMcQuaid:
> I can do that for a few popular formulae.

Thanks! Report back in here and try with `node`, `homebrew/versions/node4-lts`, `homebrew/versions/node6-lts` and `homebrew/versions/iojs` formulae.

> npm is mostly backward compatible.

Can you elaborate on "mostly"?



> IMO brew shouldn't handle npm install or yarn install. Formulae tarball should comes with node_modules like yarn tarball.

Check out the e.g. `yarn.rb` and `jsonlint.rb` formulae. We're open to doing it in a different way but may need some formula PRs for that. Thanks!

torifat:
> Can you elaborate on "mostly"?

We never faced anything like that. But, I saw a few people complaining about edge cases like `peerDependencies`.

> Check out the e.g. yarn.rb and jsonlint.rb formulae. We're open to doing it in a different way but may need some formula PRs for that. Thanks!

I have already checked `yarn.rb` and `npm install` there is not necessary. `jsonlint.rb` doesn't come with `node_modules` but I can create an issue in their repo.

The problem of entertaining `npm install` is, there are lots of npm package which doesn't support semver. So, unless the project doesn't maintain fixed versions in `package.json` it might fail any time.

torifat:
Currently there are only 10 packages using `std_npm_install_args` including `yarn` & `jsonlint`.

MikeMcQuaid:
> I have already checked yarn.rb and npm install there is not necessary.

@torifat Could you create a PR for a single formula to update it to not use that so we can discuss a new pattern? Thanks!

torifat:
It's as simple as

```
class Yarn < Formula
  desc "Javascript package manager"
  homepage "https://yarnpkg.com/"
  url "https://yarnpkg.com/downloads/0.17.0/yarn-v0.17.0.tar.gz"
  sha256 "bb87332c23baec5680e13c9afa858d851276eca27e33e215a84338fb4acb0026"

  bottle do
    cellar :any_skip_relocation
    sha256 "7f6f2621524c87e309fbea9ad7d8f49086dd481358f76f3f57f682ea0f57c2ab" => :sierra
    sha256 "aafe13b85ed27f46f884f38675927f29c4b5ed4fd78e92a3d2c803dcdb743c54" => :el_capitan
    sha256 "bd3f2ece7e38bd73d6090cb478893658515107ca8cee334ba895b030cfaf88f8" => :yosemite
  end

  depends_on :node => "4.0"

  def install
    bin.install_symlink Dir["#{libexec}/bin/*"]
  end

  test do
    (testpath/"package.json").write('{"name": "test"}')
    system bin/"yarn", "add", "jquery"
  end
end
```

MikeMcQuaid:
@torifat Can you create a pull request? Also, note that that won't work because it's not installing anything.

torifat:
@MikeMcQuaid Yeah sorry, I need to copy the contents. Ok, I will send a PR in `homebrew-core`

torifat:
@MikeMcQuaid I have updated the code I have posted above & also sending a PR in `homebrew-core`. Do you want me to use `depends_on :node => "4.0"` in the PR or just `depends_on "node"`?

MikeMcQuaid:
@torifat For now avoid using the `:node` requirement and just demonstrate how `npm install` isn't required, thanks!

gsklee:
Is something blocking further progress of this MR?

tonytonyjan:
wonder if there is any progress

MikeMcQuaid:
There's no news or progress. If there was it'd be posted on here. If anyone would like to try and create their own PR: please do.

sjmueller:
Why is this taking so long? Right now yarn from homebrew is broken for lots and lots of people who use nvm. Simply check if node or nvm is installed and skip as necessary. Even if it doesn't handle every case, surely it will improve from what's out there now...

ilovezfs:
Sorry I don't understand. The requirement system is only intended to provide non-mandatory flexibility. The default formula should always be sufficient to build any formulae, so the urgency here is pretty low unless you're doing something peculiar.

DomT4:
Having used `nvm` and Homebrew together for quite some time in the past I'm fairly sure the most "broken" I ever found things was the minor irritation of having two Node's installed. As ILZ points out, you'll still have to remember to stick an `-s` on `brew install xyz` where applicable to avoid that happening even if this is merged.

MikeMcQuaid:
> Why is this taking so long?

Because it's neither the maintainers nor contributors paid employment to work on Homebrew or submit a PR to fix this. This PR has changes requested that need addressed and it's failing CI.

If you'd like to create your own competing PR, we'd welcome it. [This document](http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html) should help and we're happy to walk you through anything else.

Thanks!

> Simply check if node or nvm is installed and skip as necessary. Even if it doesn't handle every case, surely it will improve from what's out there now...

If it were a simple problem to solve, it would have been solved already.

sjmueller:
So are you saying there is a way right now to install yarn via homebrew without installing a second copy of node (when using nvm)?

Daniel15:
@sjmueller I no longer have a Mac so I can't test this, but the `--ignore-dependencies` flag should work

jeanregisser:
Yes this flag works, but the next time you do a `brew upgrade` and there's an update for yarn, it will install it with the node deps.

ilovezfs:
@sjmueller as @DomT4 said above, "the most 'broken' I ever found things was the minor irritation of having two Nodes installed."

MikeMcQuaid:
@sjmueller two current options:

1. Stop caring about having two nodes on your system
2. Install Yarn through NPM or other means

Daniel15:
If you don't want to use Homebrew to install Yarn, there's an installation
shell script you can use.

I don't think it's bad to have two Node.js versions installed though,
particularly if you just use nvm for your own apps and the "system" /
Homebrew-installed Node.js for utilities. In the end, I don't think it
really matters which Node.js version is used for Yarn, as long as it works
properly?

torifat:
Sorry, I'm migrating to another country. So, currently not getting enough time to work on this. Anyone interested should take the lead.

DomT4:
If the primary reason most people want this is to avoid dual installations that's a slippery slope towards:
```ruby
depends_on :rust
depends_on :go
depends_on :node
depends_on :swift
depends_on :mono
depends_on :maven
depends_on :opam
```
And so on. I can see an argument for it but I also see endless new possibilities for people to break their local builds in interesting ways and then expect Homebrew to know how to fix those use cases, and quickly. Homebrew has already seen quite a bit of that with the `:ruby` and to a slightly lesser degree `:perl` requirements. 

I think it would be useful if the argument gets fleshed out a bit more than _"Having two of these is annoying"_, especially when a plain Homebrew `node` installation clocks in at <50MB, which even on the smallest SSDs Apple ships these days is a sneeze of disk space.

sjmueller:
In my experience it's been quite a hassle having two versions of node competing on the same system. When something goes wrong, which version is running? Did this particular terminal instance use a different node because a path variable was changed? If I'm hunting down a global dependency that I linked locally, am I looking in the right dir?

Much easier to troubleshoot when nvm is given autonomy to manage node versions, resulting in a single source of truth.

MikeMcQuaid:
@sjmueller A solution has been provided for that. Let's keep the discussion off this PR now, thanks,.

michael-yx-wu:
Would it also be possible to make this work for folks who use `nodenv` instead of `nvm` to manage node versions?

DomT4:
A requirement won't care where it's coming from as long as the `node` and `npm` executables are found. But someone would still need to complete any remaining work here, and convince the remaining maintainers there's an inconvenience beyond a sneeze of disk space.

ilovezfs:
@DomT4 Wait, I'm confused. So you're not volunteering to implement `:rust`, `:go`, `:node`, `:swift`, `:mono`, `:maven`, and `:opam`? I had my hopes up!

DomT4:
@ilovezfs All of those being a thing might actually be enough to push me to find an alternative package manager, and I used to help maintain this one 😆.

ilovezfs:
I think you and `fink` will get along swimmingly.

DomT4:
Come on now, I didn't say I hated myself and felt the deep need for self-punishment 😉. (Mike's going to tell us off for dragging this off-topic any second, heh).

MikeMcQuaid:
Closing as this has stalled.

SimenB:
> For now avoid using the `:node` requirement and just demonstrate how `npm install` isn't required, thanks!

Homebrew/homebrew-core#11881/

Noiwex:
yarn needs to replace homebrew

PR #1413, opened 2016-10-31
brew.sh: display message when auto update runs for too long

zmwangx:
---

Auto update could be very slow in at least two cases:
1. Many taps are tapped (e.g., my primary brew installation on my primary machine);
2. brew is used infrequently (e.g., on many of my test VMs and numerous servers that I only install things on once in a long while).

And when auto update is slow, brew seems broken when you type `brew install foo` and no output appears for half a minute or longer. Many users don't know brew is silently updating itself.

Adding a brief message before `brew update --preinstall` communicates what's happening to the user and should greatly reduce user anxiety.

**Update**. Now the message is only printed when `brew update --preinstall` has been running for 3 seconds. Also, the message is not printed when `HOMEBREW_VERBOSE` is set, because the update progress will be displayed in that case.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -296,6 +296,13 @@ source "$HOMEBREW_LIBRARY/Homebrew/utils/analytics.sh"
 setup-analytics
 report-analytics-screenview-command
 
+# Let user know we're still updating Homebrew if brew update --preinstall
+# exceeds 3 seconds.
+update-preinstall-timer() {
+  sleep 3
+  echo 'Updating Homebrew...' >&2
+}
+
 update-preinstall() {
   [[ -z "$HOMEBREW_HELP" ]] || return
   [[ -z "$HOMEBREW_NO_AUTO_UPDATE" ]] || return
@@ -306,7 +313,19 @@ update-preinstall() {
 
   if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || "$HOMEBREW_COMMAND" = "tap" ]]
   then
+    if [[ -z "$HOMEBREW_VERBOSE" ]]
+    then
+      update-preinstall-timer &
+      timer_pid=$!
+    fi
+
     brew update --preinstall
+
+    if [[ -n "$timer_pid" ]]
+    then
+      kill "$timer_pid" 2>/dev/null
+      wait "$timer_pid" 2>/dev/null
+    fi
   fi
 
   # If brew update --preinstall did a migration then export the new locations.


MikeMcQuaid:
> Many taps are tapped (e.g., my primary brew installation on my primary machine);
> brew is used infrequently (e.g., on many of my test VMs and numerous servers that I only install things on once in a long while).
> And when auto update is slow, brew seems broken when you type brew install foo and no output appears for half a minute or longer.

It feels like these are the problems that should be resolved instead. A suggestion: output this message using a background timer if `brew update --preinstall` hasn't completed after a given number of seconds e.g. 5s rather than printing it always.

zmwangx:
> A suggestion: output this message using a background timer if brew update --preinstall hasn't completed after a given number of seconds e.g. 5s rather than printing it always.

That's reasonable, implemented. I think 3s is better though.

> It feels like these are the problems that should be resolved instead.

Not sure I understand this part, those are valid usage patterns.

If you mean update should be faster — sure, but there's a fundamental limit imposed by `git`.

MikeMcQuaid:
> That's reasonable, implemented.

Thanks!

> If you mean update should be faster — sure, but there's a fundamental limit imposed by git.

Sure. I guess I mean e.g. the "when it takes too long it's confusing" means stop that case (it being too confusing when it takes too long) rather than the general case.

Library/Homebrew/brew.sh
```diff
@@ -296,6 +296,13 @@ source "$HOMEBREW_LIBRARY/Homebrew/utils/analytics.sh"
 setup-analytics
 report-analytics-screenview-command
 
+# Let user know we're still updating Homebrew if brew update --preinstall
+# exceeds 3 seconds.
+update-preinstall-timer() {
+  sleep 3
+  echo 'Still updating Homebrew, please wait...' >&2
```
MikeMcQuaid: I think this could just be the original message about `Updating Homebrew`

Library/Homebrew/brew.sh
```diff
@@ -306,7 +313,17 @@ update-preinstall() {
 
   if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || "$HOMEBREW_COMMAND" = "tap" ]]
   then
+    if [[ -z "$HOMEBREW_VERBOSE" ]]
+    then
+      update-preinstall-timer & timer_pid=$!
```
MikeMcQuaid: Can you set `timer_pid=$!` on a new line?

Library/Homebrew/brew.sh
```diff
@@ -306,7 +313,17 @@ update-preinstall() {
...
+    fi
+
     brew update --preinstall
+
+    if [[ -n "$timer_pid" ]]
+    then
+      { kill "$timer_pid"; wait "$timer_pid"; } 2>/dev/null
```
MikeMcQuaid: Worth doing `kill -9` and skipping the `wait`?

Library/Homebrew/brew.sh
zmwangx: Sure.

Library/Homebrew/brew.sh
zmwangx: `SIGTERM` or `SIGKILL` doesn't make a difference to `sleep(1)`. I'm `wait`ing here only because it's good practice in general (at least to me), not because it's necessary.

Library/Homebrew/brew.sh
zmwangx: That would sound like it just started "updating Homebrew", leaving the user to wonder what it was doing in the past three seconds, or five. But I'll change it back if you insist.

MikeMcQuaid: Yeh, I think it's still probably preferable, thanks.

Library/Homebrew/brew.sh
MikeMcQuaid: Ok. Can you split onto multiple lines? Does this need handled on `exit` or will that be done automatically by process cleanup?

Library/Homebrew/brew.sh
```diff
@@ -296,6 +296,13 @@ source "$HOMEBREW_LIBRARY/Homebrew/utils/analytics.sh"
 setup-analytics
 report-analytics-screenview-command
 
+# Let user know we're still updating Homebrew if brew update --preinstall
+# exceeds 3 seconds.
+update-preinstall-timer() {
+  sleep 3
+  echo 'Updating Homebrew...' >&2
+}
```
MikeMcQuaid: Was testing this out and figured it may make more sense to live in `update.sh` to make it easier to test. Thoughts?

MikeMcQuaid: Yeh, that's fair enough.



> What kind of testing are you thinking about?

Just manually reproducing this locally. Amtrak made this easier today, though 😁

MikeMcQuaid:
Thanks @zmwangx!

PR #1415, opened 2016-10-31
Update macOS Xcode tool checks do use Version class for comparisons

robinkunde:

-----

Right now, Library/Homebrew/os/mac/xcode.rb uses string comparisons to determine if the version of a local tool is outdated. This will create false results if the two version strings are of different lengths, ie. "1000" < "800" will return true instead of false. Brew already has a class for proper comparison of semantic version strings and this PR updates xcode.rb to make use of it.

I have not added any tests since that would require somehow mocking the data returned by the system commands that xcode.rb uses to determine the versions of the installed tools, and I don't know how to do that at the moment.

### DIFF
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -27,15 +27,15 @@ module OS
 
       def prerelease?
         # TODO: bump to version >= "8.3" after Xcode 8.2 is stable.
-        Version.create(version) >= Version.create("8.2")
+        Version.new(version) >= "8.2"
       end
 
       def outdated?
-        Version.create(version) < Version.create(latest_version)
+        Version.new(version) < latest_version
       end
 
       def without_clt?
-        installed? && Version.create(version) >= Version.create("4.3") && !MacOS::CLT.installed?
+        installed? && Version.new(version) >= "4.3" && !MacOS::CLT.installed?
       end
 
       # Returns a Pathname object corresponding to Xcode.app's Developer
@@ -56,7 +56,7 @@ module OS
       end
 
       def toolchain_path
-        Pathname.new("#{prefix}/Toolchains/XcodeDefault.xctoolchain") if installed? && Version.create(version) >= Version.create("4.3")
+        Pathname.new("#{prefix}/Toolchains/XcodeDefault.xctoolchain") if installed? && Version.new(version) >= "4.3"
       end
 
       # Ask Spotlight where Xcode is. If the user didn't install the
@@ -71,7 +71,7 @@ module OS
       end
 
       def update_instructions
-        if Version.create(MacOS.version) >= Version.create("10.9") && !OS::Mac.prerelease?
+        if Version.new(MacOS.version) >= "10.9" && !OS::Mac.prerelease?
           <<-EOS.undent
             Xcode can be updated from the App Store.
           EOS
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> This will create false results if the two version strings are of different lengths, ie. "1000" < "800" will return true instead of false.

Can you provide an example within the usage of this class that currently breaks? Thanks!

robinkunde:
I can't, and given the way version numbers have been increasing it _shouldn't_ be a problem until Xcode 10 becomes available. However, because the code is wrong I saw this as an appropriate thing to fix now, before we get blindsided.

For example, the length of the clang version string just changed. It's conceivable that Apple might put out version "800.0.42.2", and then "800.0.42.10" some time later. Using string comparison, "800.0.42.10" evaluate as smaller than "800.0.42.2"

MikeMcQuaid:
@robinkunde I'm happy to edit files to get a reproduction but I'd like to be able to reproduce this issue before making changes here otherwise we're speculatively fixing something. Thanks!

robinkunde:
@MikeMcQuaid oh, I see what you mean.

Edit Library/Homebrew/os/mac/xcode.rb. Hardcode uncached_version to always return "10.0" (line 94 or so). Then run "brew install git" for example. I now get this error:
Error: Your Xcode (10.0) is outdated.
Please update to Xcode 8.1 (or delete it).

Now revert the change and bump the latest xcode version for your macOS release to "10.0" (line 19 or so). Run "brew install git" again and you get no complaint about Xcode being out of date.

MikeMcQuaid:
@robinkunde Beautiful 😉! Nice spot. I think this code could be cleaned up perhaps with an `XcodeVersion` subclass that allows string comparisons which would avoid needing quite as many new `Version`s being created. Thoughts?

robinkunde:
@MikeMcQuaid Is your concern the code repetition or the overhead of creating the objects? 
If it's the former, it could make sense to abstract that away. I thought about having the methods that return version strings right now return version objects directly, but I have no idea in how many different places those methods are called, and I didn't want to trigger any refactoring since there's poor test coverage right now.
If it's the latter, the overhead would just be moved into this subclass, unless we reimplement simple semantic version comparison by splitting the string on each dot and then compare each component as integers.

MikeMcQuaid:
It was the former, yeh 👍

robinkunde:
@MikeMcQuaid Ok, I gave it a go. Let me know what you think. I don't have a good grasp on idiomatic ruby yet, as I'm quite new to it.

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -225,16 +225,36 @@ def version
...
+          if XcodeVersion.is_higher_or_equal(MacOS.version, than: :mavericks.to_s)
             next unless File.exist?("#{MAVERICKS_PKG_PATH}/usr/bin/clang")
           end
           version = MacOS.pkgutil_info(id)[/version: (.+)$/, 1]
           return version if version
         end
       end
     end
+
+    module XcodeVersion
```
MikeMcQuaid: Make this a class that inherits from `Version` instead and override `<=>`. Also, `Xcode::Version` is probably a preferable name. Thanks!

robinkunde:
@MikeMcQuaid Sorry, it took me awhile to make time. I've amended the last commit with your suggestions. Let me know if that's what you were looking for. Thank you!

MikeMcQuaid:
Much cleaner, thanks!

MikeMcQuaid:

PR #1426, opened 2016-11-03
audit: restore check that was lost in #927

scpeters:

-----

This audit check ensures that certain build dependencies
are explicitly marked either as `:build` or `:run`.
It seems to have been lost in #927.
It was also adjusted in #1290.

I tried adding a test in https://github.com/scpeters/brew/commit/fd6e1896efcc5b6f4e3582f93719e7d10751515f, but it doesn't work because the test fixture can't load the `cmake` formula:

~~~
  1) Failure:
FormulaAuditorTests#test_audit_build_time_deps [/usr/local/Homebrew/Library/Homebrew/test/test_audit.rb:484]:
--- expected
+++ actual
@@ -1 +1 @@
-["cmake dependency should be"]
+["Can't find dependency \"cmake\"."]

~~~

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -415,6 +415,12 @@ class FormulaAuditor
             EOS
         when *BUILD_TIME_DEPS
           next if dep.build? || dep.run?
+          problem <<-EOS.undent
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @scpeters, nice spot!

PR #1432, opened 2016-11-03
metafiles: warn about missing metafiles 

DomT4:

-----

Have a PR instead of idle rambling in https://github.com/Homebrew/homebrew-core/pull/6544. Not that you have to accept it these days 😉. Explanation is obvious enough, really.

Some impacted formulae, although they aren't flagged in your standard `brew audit` currently because of the `@strict` block:
```
ack:
  * Metafiles such as README, LICENSE or COPYING do not seem
    to have been installed. Please ensure this is correct
    otherwise you may need to add `prefix.install_metafiles`
    to the formula.
boost:
  * Metafiles such as README, LICENSE or COPYING do not seem
    to have been installed. Please ensure this is correct
    otherwise you may need to add `prefix.install_metafiles`
    to the formula.
boost-python:
  * Metafiles such as README, LICENSE or COPYING do not seem
    to have been installed. Please ensure this is correct
    otherwise you may need to add `prefix.install_metafiles`
    to the formula.
bsdmake:
  * Metafiles such as README, LICENSE or COPYING do not seem
    to have been installed. Please ensure this is correct
    otherwise you may need to add `prefix.install_metafiles`
    to the formula.
```

`boost` and `boost-python` are actually flagging an issue with the `metafiles` code itself, because the license there is `LICENSE_1_0.txt` which the `metafiles` code currently ignores.


MikeMcQuaid:
I think validating those files are provided by the tarball would be a good start. You could make this an `--online` check and check GitHub for their presence, perhaps.

DomT4:
I'll try and think of some validation method, but of all 4 of the formulae flagged before the audit issue earlier 2 host their code on GitHub. If that's a consistent sort of ratio there's still going to be a lot of gaps 😕.

Something like that would be adding:
```
curl https://api.github.com/repos/harr999y/boost/contents
```
to the `audit` check and then matching some generic metafile names, I guess.

`ack` embeds the license in the single file, `boost` and `boost-python` provide a license file that currently escapes the metafiles logic, `bsdmake` claims it should have a separate license included but for whatever reason does not.

MikeMcQuaid:
I think false positives (i.e. we don't catch it when we should) are preferable to false negatives.

DomT4:
Sure, no disagreement there. Just think it's a difficult thing to balance without negating the usefulness of a check.

I thought about not checking unless there's a cached download for that formula, not checking if the formula isn't a tarball and then running `tar -tf` on tarballs to peek at the contents instead of trying to check online, but I'm not sure how workable that is. That would remove the requirement for stuff to be on GitHub though.

MikeMcQuaid:
I think installing metafiles is most useful/common for GitHub repositories anyway where the buildsystems tend to be more limited around e.g. installing documentation. At least it feels like it'd be a good start.

An alternative to an `audit` is making this something checked automatically after installation and fails on our CI. At that point, though, I wonder if it's not worth thinking about how to do this automatically instead?

MikeMcQuaid:
@DomT4 Any more thoughts here?

DomT4:
To be honest, I forgot I'd opened this around all the election stuff. I'll try to think about it a bit more. 

The metafiles stuff has proven difficult to get "right" over time; Alex modified it for the better a while back but doing so exposed other issues. That makes me quite wary about fiddling with the logic itself directly, but I'm also sympathetic to the argument Homebrew doesn't necessarily want `prefix.install_metafiles` to become a manual fix all over the place.

Theoretically you could try to be clever if the `buildpath` doesn't contain any metafiles the installation process goes hunting for them in subdirectories, but I'm not sure that doesn't spawn other issues, such as:

Formula x
```
- buildpath
  - configure
  - Makefile.in
  - docs
    - Building
    - README
    - LICENSE
    - Changelog
```
Installs `README` & `LICENSE` correctly.

Formula y
```
- buildpath
  - src/github.com/ducks/pond
    - README
  - src/github.com/horses/horse
    - README
    - COPYING
  - src/github.com/test/test2
    - README
    - LICENSE
    - vendor
      - golang.org/x/net
        - README
        - LICENSE
```
Where `src/github.com/test/test2` contains the Go formula we're building, but instead the `README` from `src/github.com/ducks/pond` gets installed. It's not an especially common scenario, but there are formula that use both Go vendoring and Go resources because the resources cover a build tool upstream don't need to vendor permanently, etc.

Due to the way it operates the metafiles logic needs to more-or-less work for everything, to an extent, and I'm not immediately sure how to draw up an automated solution that doesn't risk working even less well than the current logic.

MikeMcQuaid:
> Due to the way it operates the metafiles logic needs to more-or-less work for everything, to an extent, and I'm not immediately sure how to draw up an automated solution that doesn't risk working even less well than the current logic.

I'm thinking the right solution may be to copy the metafiles somewhere temporary before `install` is run and, if there's a non-empty installation, install them afterwards. This avoids the need to have a separate checks for each location they could be moved around to during `install`.

PR #1435, opened 2016-11-04
Add "null version" class, and return compiler versions/build versions as Version objects

mistydemeo:

-----

Right now, we return compiler versions and builds as raw strings/fixnums or `nil`. The latter has caused problems for us in the past: comparing `nil` vs anything will always fail, and comparing strings/numbers against `nil` similarly raises exceptions. This is a particular problem for non-typical platforms like Linux, which are commonly missing compilers which are present on OS X.

This patch makes some major changes to this model:

* A new `Version::NullVersion` class has been introduced to represent the absence of a version. It supports the same API as `Version`, but always returns as smaller in any comparison, even against itself. An instance of it is available as `Version::NULL`.
* Unparseable strings passed to `Version.parse` will now return `Version::NULL` instead of `nil`.
* Compiler versions and builds are now returned as instances of `Version` instead of raw strings or integers.
* For compatibility with existing formulae which compare compiler versions/builds against other strings or integers, `Version#<=>` will now coerce strings or integers into versions before proceeding with a comparison.
* `Version#<=>` will always report a larger value when compared against a raw `nil`.

### DIFF
--- a/Library/Homebrew/compilers.rb
+++ b/Library/Homebrew/compilers.rb
@@ -14,7 +14,14 @@ end
 
 class CompilerFailure
   attr_reader :name
-  attr_rw :version
+
+  def version(val = nil)
+    if val
+      @version = Version.parse(val.to_s)
+    else
+      @version
+    end
+  end
 
   # Allows Apple compiler `fails_with` statements to keep using `build`
   # even though `build` and `version` are the same internally
@@ -45,7 +52,7 @@ class CompilerFailure
 
   def initialize(name, version, &block)
     @name = name
-    @version = version
+    @version = Version.parse(version.to_s)
     instance_eval(&block) if block_given?
   end
 
--- a/Library/Homebrew/compilers.rb
+++ b/Library/Homebrew/compilers.rb
@@ -122,13 +122,13 @@ class CompilerSelector
         GNU_GCC_VERSIONS.reverse_each do |v|
           name = "gcc-#{v}"
           version = compiler_version(name)
-          yield Compiler.new(name, version) if version
+          yield Compiler.new(name, version) unless version.null?
         end
       when :llvm
         # no-op. DSL supported, compiler is not.
       else
         version = compiler_version(compiler)
-        yield Compiler.new(compiler, version) if version
+        yield Compiler.new(compiler, version) unless version.null?
       end
     end
   end
--- a/Library/Homebrew/development_tools.rb
+++ b/Library/Homebrew/development_tools.rb
@@ -44,7 +44,9 @@ class DevelopmentTools
     def gcc_40_build_version
       @gcc_40_build_version ||=
         if (path = locate("gcc-4.0"))
-          `#{path} --version 2>/dev/null`[/build (\d{4,})/, 1].to_i
+          Version.new `#{path} --version 2>/dev/null`[/build (\d{4,})/, 1].to_i
+        else
+          Version::NULL
         end
     end
     alias gcc_4_0_build_version gcc_40_build_version
@@ -54,7 +56,9 @@ class DevelopmentTools
         begin
           gcc = locate("gcc-4.2") || HOMEBREW_PREFIX.join("opt/apple-gcc42/bin/gcc-4.2")
           if gcc.exist? && !gcc.realpath.basename.to_s.start_with?("llvm")
-            `#{gcc} --version 2>/dev/null`[/build (\d{4,})/, 1].to_i
+            Version.new `#{gcc} --version 2>/dev/null`[/build (\d{4,})/, 1]
+          else
+            Version::NULL
           end
         end
     end
@@ -63,14 +67,18 @@ class DevelopmentTools
     def clang_version
       @clang_version ||=
         if (path = locate("clang"))
-          `#{path} --version`[/(?:clang|LLVM) version (\d\.\d)/, 1]
+          Version.new `#{path} --version`[/(?:clang|LLVM) version (\d\.\d)/, 1]
+        else
+          Version::NULL
         end
     end
 
     def clang_build_version
       @clang_build_version ||=
         if (path = locate("clang"))
-          `#{path} --version`[/clang-(\d{2,})/, 1].to_i
+          Version.new `#{path} --version`[/clang-(\d{2,})/, 1]
+        else
+          Version::NULL
         end
     end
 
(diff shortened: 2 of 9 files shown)


Library/Homebrew/resource.rb
```diff
@@ -145,7 +145,10 @@ def url(val = nil, specs = {})
   end
 
   def version(val = nil)
-    @version ||= detect_version(val)
+    @version ||= begin
+      version = detect_version(val)
+      version.is_a?(Version::NullVersion) ? nil : version
```
mistydemeo: This probably isn't the right fix, but I didn't ferret out where the `nil` check is happening. This fixed an issue where resources w/o parseable versions weren't inheriting the version of their parent since they now had non-nil values.

Library/Homebrew/system_config.rb
```diff
@@ -143,8 +143,8 @@ def dump_verbose_config(f = $stdout)
       f.puts "HOMEBREW_BOTTLE_DOMAIN: #{BottleSpecification::DEFAULT_DOMAIN}"
       f.puts hardware if hardware
       f.puts "Homebrew Ruby: #{describe_homebrew_ruby}"
-      f.puts "GCC-4.0: build #{gcc_40}" if gcc_40
-      f.puts "GCC-4.2: build #{gcc_42}" if gcc_42
+      f.puts "GCC-4.0: build #{gcc_40}" if gcc_40 > "0"
```
mistydemeo: This probably also isn't the right approach. `Version#null?` maybe?

Library/Homebrew/version.rb
```diff
@@ -171,6 +171,41 @@ def detected_from_url?
     end
   end
 
+  class NullVersion
```
jawshooah: Since this is effectively a singleton class, wouldn't it make more sense to just do `NULL = Version.new do …`? That way, instead of checking `version.is_a?(Version::NullVersion)` or `version > "0"`, you could just check `version.equal?(Version::NULL)`

Library/Homebrew/version.rb
```diff
@@ -171,6 +171,41 @@ def detected_from_url?
...
+    def <=>(other)
+      -1
+    end
+
+    def eql?(other)
+      # Makes sure that the same instance of Version::NULL
+      # will never equal itself; normally Comparable#==
+      # will return true for this regardless of the return
+      # value of #<=>
+      false
```
jawshooah: Why should `Version::NULL` never equal itself?

mistydemeo: This mirrors `NaN` - it reflects a value that's not a number. (And it's consistent with the `#to_f` value returned by this class.)

mistydemeo:
Are the Linuxbrew tests expected to be broken right now? Doesn't look related to this branch.

MikeMcQuaid:
> Are the Linuxbrew tests expected to be broken right now?

They are not expected to be.

Library/Homebrew/compilers.rb
```diff
@@ -14,7 +14,10 @@ module CompilerConstants
 
 class CompilerFailure
   attr_reader :name
-  attr_rw :version
+
+  def version(val = nil)
+    val ? @version = Version.new(val.to_s) : @version
```
MikeMcQuaid: I think an `if`/`else` would make this a little clearer.

Library/Homebrew/compilers.rb
```diff
@@ -45,7 +48,7 @@ def self.create(spec, &block)
 
   def initialize(name, version, &block)
     @name = name
-    @version = version
+    @version = Version.new(version.to_s)
```
MikeMcQuaid: Worth checking it's not empty/nil?

Library/Homebrew/development_tools.rb
```diff
@@ -63,23 +67,27 @@ def gcc_42_build_version
...
         end
     end
 
     def non_apple_gcc_version(cc)
       (@non_apple_gcc_version ||= {}).fetch(cc) do
         path = HOMEBREW_PREFIX.join("opt", "gcc", "bin", cc)
         path = locate(cc) unless path.exist?
-        version = `#{path} --version`[/gcc(?:-\d(?:\.\d)? \(.+\))? (\d\.\d\.\d)/, 1] if path
-        @non_apple_gcc_version[cc] = version
+        version = Version.new(`#{path} --version`[/gcc(?:-\d(?:\.\d)? \(.+\))? (\d\.\d\.\d)/, 1]) if path
```
MikeMcQuaid: I know it's not your code but: sticking the `if` in the block form would make it more obvious given the line length.

Library/Homebrew/resource.rb
```diff
@@ -145,7 +145,10 @@ def url(val = nil, specs = {})
...
```
MikeMcQuaid: Why not return a `NullVersion` here, out of interest?

Library/Homebrew/system_config.rb
MikeMcQuaid: Yeh, something like that makes sense to me 👍

Library/Homebrew/version.rb
MikeMcQuaid: May be worth having a separate `null_version.rb`?

Library/Homebrew/resource.rb
mistydemeo: I haven't tracked down the nil check yet, but this has to do with versions being propagated from the parent. If a resource's version isn't parseable from the URL, it inherits the version of its parent resource (e.g., the formula). If the version is assigned as a `NullVersion` instead of `nil`, then it fails to inherit the parent's version.

MikeMcQuaid: Maybe this is too strict but I see explicitly checking for `is_a?` to be a bit of a smell; particularly when converting to `nil` so would be 🆒 to try and avoid them when possible.

mistydemeo:
I'm pretty sure the Linuxbrew failures are from bad compiler selection, I'll dig into the bug. The tests on OS X didn't surface this.

mistydemeo:
Fixed the compiler selection bugs, and beefed up the compiler selection tests a little.

Library/Homebrew/version.rb
mistydemeo: Good idea, separated out.

Library/Homebrew/development_tools.rb
mistydemeo: Updated!

Library/Homebrew/version.rb
mistydemeo: Yeah, good point. There really shouldn't be any other instances. Updated.

Library/Homebrew/compilers.rb
mistydemeo: Maybe use `Version.parse` on the stringified value? That'll assign `Version::NULL` as appropriate.

mistydemeo: Updated.

MikeMcQuaid:
@mistydemeo Feel free to 🚢 any time when you'll be around for e.g. ~4 hours to handle any post-merge 💥.

mistydemeo:
I'm on for about four hours now so I'm ready to handle the things I've inevitably broken. ✨

PR #1436, opened 2016-11-04
Disable auto update at tap command for listing.

rcmdnk:

-----

Currently "Auto Update" is launched at commands of `install`, `upgrade` and `tap`.

Sometime I want to obtain a list of taps, and use `brew tap` command.
But if Auto Update is launched, there are additional lines about Auto Update.

Any installation is done at `brew tap` only when a repository is given,
and Auto Update is necessary only for such a case.

Therefore, this update disables Auto Update  if no additional arguments are given to `brew tap` or listing options of `--list*` are given.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -213,6 +213,7 @@ case "$HOMEBREW_COMMAND" in
   environment) HOMEBREW_COMMAND="--env" ;;
   --config)    HOMEBREW_COMMAND="config" ;;
 esac
+HOMEBREW_COMMAND_ARGS=("$@")
 
 if [[ -z "$HOMEBREW_DEVELOPER" ]]
 then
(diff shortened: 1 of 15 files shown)


Library/Homebrew/brew.sh
```diff
@@ -304,7 +305,8 @@ update-preinstall() {
   # Allow auto-update migration now we have a fix in place (below in this function).
   export HOMEBREW_ENABLE_AUTO_UPDATE_MIGRATION="1"
 
-  if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || "$HOMEBREW_COMMAND" = "tap" ]]
+  echo HOMEBREW_COMMAND_ARGS=${HOMEBREW_COMMAND_ARGS[0]}
```
scpeters: I assume this `echo` statement was for debugging; it's probably best to remove it for the PR

scpeters:
I ran into a similar issue where I wanted to untap any non-core taps in a script. I think I was using `brew untap $(brew tap | grep -v homebrew/homebrew-core)`, but the auto-update console messages were causing problems. My solution was to call `brew update` right before that command.

I'm +1 on the idea of this PR.

MikeMcQuaid:
@rcmdnk For future reference it'd be good to push changes to the same PR rather than opening a new one.

rcmdnk:
sorry, I missed that I deleted this PR branch, so that I made a new.

PR #1439, opened 2016-11-04
disable oputput of brew tap at initialize

rcmdnk:

-----

Currently "Auto Update" is launched at commands of install, upgrade and tap.

Sometime I want to obtain a list of taps, and use brew tap command.
But if Auto Update is launched, there are additional lines about Auto Update.

Any installation is done at brew tap only when a repository is given,
and Auto Update is necessary only for such a case.

Therefore, this update disables Auto Update if no additional arguments are given to brew tap or listing options of --list* are given.

(this is renewed PR of #1436 w/o debug output.)

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -213,6 +213,7 @@ case "$HOMEBREW_COMMAND" in
   environment) HOMEBREW_COMMAND="--env" ;;
   --config)    HOMEBREW_COMMAND="config" ;;
 esac
+HOMEBREW_COMMAND_ARGS=("$@")
 
 if [[ -z "$HOMEBREW_DEVELOPER" ]]
 then
@@ -304,7 +305,7 @@ update-preinstall() {
   # Allow auto-update migration now we have a fix in place (below in this function).
   export HOMEBREW_ENABLE_AUTO_UPDATE_MIGRATION="1"
 
-  if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || "$HOMEBREW_COMMAND" = "tap" ]]
+  if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || ( "$HOMEBREW_COMMAND" = "tap" && $HOMEBREW_ARG_COUNT -gt 1 && ! "${HOMEBREW_COMMAND_ARGS[0]}" =~ ^--list ) ]]
   then
     brew update --preinstall
   fi


MikeMcQuaid:
I'm not sure the extra code is worth this slight workaround. Is there a simpler possible version?

rcmdnk:
Though I think this is simple enough,
if mind to add "HOMEBREW_COMMAND_ARG", then 
it can be like:

    - { upadate-preinstall; "homebrew-$HOMEBREW_COMMAND" "$@"; exit $?; }
    + { upadate-preinstall "$@"; "homebrew-$HOMEBREW_COMMAND" "$@"; exit $?; }

and use $1 in update-preinstall.

or if you are thinking 

     ( "$HOMEBREW_COMMAND" = "tap" && $HOMEBREW_ARG_COUNT -gt 1 && ! "${HOMEBREW_COMMAND_ARGS[0]}" =~ ^--list )

is too long, it may be like:


    HOMEBREW_COMMAND_ARGS="$*"
    ...
    ( "$HOMEBREW_COMMAND" = "tap" && ! "${HOMEBREW_COMMAND_ARGS[0]}" =~ ^--list )

or combine.

MikeMcQuaid:
I don't like setting the precedent that `brew.sh` should know how other commands function; it'll get pretty messy pretty fast. I think it's probably worth passing on this and setting a higher timeout if you find the current one annoying.

zmwangx:
> Sometime I want to obtain a list of taps, and use brew tap command.

Assuming you need that for a script, you can run

```sh
HOMEBREW_NO_AUTO_UPDATE=1 brew tap
```

> I think it's probably worth passing on this and setting a higher timeout if you find the current one annoying.

Hmm, what's the "timeout" you're talking about here?

zmwangx:
Okay, `HOMEBREW_AUTO_UPDATE_SECS`.

rcmdnk:
Currently I do check an output and remove unnecessary lines like:

> https://github.com/rcmdnk/homebrew-file/blob/4889745952e1d8d09d35a28ee3443877e2de79e1/bin/brew-file#L1606

And HOMEBREW_NO_AUTO_UPDATE opt is also good possibility to avoid it.

On the other hand, you enabled this auto update by default because most of people use as default and don't check options, isn't it?
Making a list with `brew tap` is not so special thing,
so that it should be more kindly like this, I think.

In addition, in brew.sh, it already have such

    if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || "$HOMEBREW_COMMAND" = "tap" ]]

so it is not so much, I feel.

scpeters:
Thanks for the reminder about `HOMEBREW_NO_AUTO_UPDATE` @zmwangx ! That's perfect

MikeMcQuaid:
Passing on this for now but may reconsider in future.

PR #1449, opened 2016-11-06
bottle: improve relocatability check

ilovezfs:
Given how common it is for formulae to hard-code `etc` and `var`, check
for those paths (`/usr/local/etc` and `/usr/local/var`) when determing
relocatability.

Fixes https://github.com/Homebrew/homebrew-core/issues/6454.


MikeMcQuaid:
Nice work :+1:

ilovezfs:
Thanks :)

PR #1460, opened 2016-11-08
os/mac: MachO.dynamically_linked_libraries filtering.

woodruffw:

-----

This allows us to filter out weak linkages during audits, preventing
a false error from occurring when the dylib cannot be found.

ref Homebrew/homebrew-core#6645
cc @MikeMcQuaid, @UniqMartin

### DIFF
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -23,7 +23,10 @@ class LinkageChecker
     @keg.find do |file|
       next if file.symlink? || file.directory?
       next unless file.dylib? || file.mach_o_executable? || file.mach_o_bundle?
-      file.dynamically_linked_libraries.each do |dylib|
+
+      # weakly loaded dylibs may not actually exist on disk, so skip them
+      # when checking for broken linkage
+      file.dynamically_linked_libraries(except: :LC_LOAD_WEAK_DYLIB).each do |dylib|
         @reverse_links[dylib] << file
         if dylib.start_with? "@"
           @variable_dylibs << dylib
--- a/Library/Homebrew/os/mac/mach.rb
+++ b/Library/Homebrew/os/mac/mach.rb
@@ -1,7 +1,7 @@
 require "vendor/macho/macho"
 require "os/mac/architecture_list"
 
-module MachO
+module MachOShim
   # @private
   def macho
     @macho ||= begin
--- a/Library/Homebrew/os/mac/mach.rb
+++ b/Library/Homebrew/os/mac/mach.rb
@@ -51,8 +51,10 @@ module MachO
     end
   end
 
-  def dynamically_linked_libraries
-    macho.linked_dylibs
+  def dynamically_linked_libraries(except: :none)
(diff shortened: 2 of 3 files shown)


woodruffw:
P.S.: I noticed that we're using the `MachO` namespace in Homebrew as well as in `ruby-macho`. This hasn't been an issue yet, but it could lead to some method clobbering in the future. Thoughts on changing the module name in *os/mac/mach.rb* to `MachOShim` or something similar?

MikeMcQuaid:
> Thoughts on changing the module name in os/mac/mach.rb to MachOShim or something similar?

Fine with me 👍

Library/Homebrew/os/mac/linkage_checker.rb
```diff
@@ -23,7 +23,7 @@ def check_dylibs
     @keg.find do |file|
       next if file.symlink? || file.directory?
       next unless file.dylib? || file.mach_o_executable? || file.mach_o_bundle?
-      file.dynamically_linked_libraries.each do |dylib|
+      file.dynamically_linked_libraries(except: :LC_LOAD_WEAK_DYLIB).each do |dylib|
```
MikeMcQuaid: A comment above here explaining why would be good.

Library/Homebrew/os/mac/mach.rb
```diff
@@ -51,8 +51,10 @@ def mach_data
     end
   end
 
-  def dynamically_linked_libraries
-    macho.linked_dylibs
+  def dynamically_linked_libraries(except: [])
+    lcs = macho.dylib_load_commands.reject { |lc| [*except].include?(lc.type) }
```
MikeMcQuaid: Looks like you're expecting `expect` to be an array here but it's a non-array above. I'm happy with either but would be good to be consistent. Additionally, if you're expecting an array I think `expect` is equivalent to `[*except]`, no?

woodruffw: I did this to allow someone to pass either an array of symbols to filter on or a single symbol. The `[*except]` will correctly pack either a single value or an array.

That's pretty hard to read though, so I could replace it with something like this:

```ruby
unless except.is_a? Array
  except = [except]
end
```

woodruffw: Got it!

MikeMcQuaid:
Thanks again @woodruffw!

MikeMcQuaid:
Apologies, merged this prematurely because I didn't realise that this was reliant on a new ruby-macho.  Would be good to have tests that catch this in future.

woodruffw:
Yikes. This was a huge oversight on my part - we never added `MachO::FatFile#dylib_load_commands`, only the corresponding `MachOFile` method. I'll make the necessary changes to `ruby-macho`, submit a revendoring PR, and then this should be good to merge.

woodruffw:
#1469 will fix that bug. This is a difficult case to test, as `MachOFile` and `FatFile` can't be complete clones of each other (each has semantics that don't make sense for the other), but I'll add something to `ruby-macho`'s test suite that enforces expected compatibility between the two.

Issue #712, opened 2016-08-14
Installed Homebrew formulas are always preferred over other ways to satisfy dependencies


tdsmith:
Build#expand_reqs has the following logic:

``` ruby
      elsif req.satisfied? && req.default_formula? && (dep = req.to_dependency).installed?
        deps << dep
        Requirement.prune
      end
```

This is what we want _if_ the requirement was satisfied by the installed default_formula. It is probably _not_ what we want if `satisfy` found some other way to satisfy the requirement.

MikeMcQuaid:
Can you elaborate on why this is a problem and what formulae (if any) it's currently negatively affecting? Thanks!

tdsmith:
The way I discovered this was I was trying to test macvim against a pyenv python3 with Homebrew python3 installed. Even though the pyenv python3 was first in PATH, `--with-python3` (which triggers a `Python3Requirement`) ended up building against the Homebrew python3.

We sometimes describe the behavior of the `PythonRequirement`s as "building against the first python in PATH," which is not exactly the observed behavior.

We could wontfix as a documentation bug; I'd prefer that we finally fully-bake an API for adding paths to superenv environments as the result of a satisfy operation.

tdsmith:
(Sorry, filed this on a plane, didn't want to forget about it)

MikeMcQuaid:
> Even though the pyenv python3 was first in PATH, --with-python3 (which triggers a Python3Requirement) ended up building against the Homebrew python3.

I see. Makes sense to tweak this (but obviously will need tested with a bunch of other things as this code is prone to 💥).

> I'd prefer that we finally fully-bake an API for adding paths to superenv environments as the result of a satisfy operation.

Something beyond Requirements? Feels like that may be overkill.

ilovezfs:
@tdsmith I think this may work:
```
diff --git a/Library/Homebrew/build.rb b/Library/Homebrew/build.rb
index 8915e94..22b1dd9 100644
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -49,7 +49,7 @@ class Build
         Requirement.prune
       elsif req.build? && dependent != formula
         Requirement.prune
-      elsif req.satisfied? && req.default_formula? && (dep = req.to_dependency).installed?
+      elsif req.satisfied? && req.default_formula? && req.uses_default? && (dep = req.to_dependency).installed?
         deps << dep
         Requirement.prune
       end
[...]

PR #1471, opened 2016-11-09
Add --with-custom-python option to virtualenv formulæ

tdsmith:
Installing with `-s` in order that `:python` dependencies do not install
Homebrew's python has some poor ergonomics: it's not discoverable and it
doesn't persist with `brew upgrade`. This adds an option to every
virtualenv-based formula with a discoverable name that has no effect
except to force a source build and which will be memorized for upgrades.

### DIFF
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -102,6 +102,10 @@ module Language
     module Virtualenv
       def self.included(base)
         base.class_eval do
+          # This is a no-op option intended to force a source build in a way
+          # that `brew upgrade` will memorize.
+          option "with-custom-python", "Use the python in PATH instead of Homebrew's python distribution"
+
           resource "homebrew-virtualenv" do
             url PYTHON_VIRTUALENV_URL
             sha256 PYTHON_VIRTUALENV_SHA256


ilovezfs:
Isn't this what :python and :python3 are supposed to accomplish?

tdsmith:
> Isn't this what :python and :python3 are supposed to accomplish?

This complements the :python requirements. The problem in this instance is that :python resolves to Homebrew's python when a bottle is poured, but there is no persistent flag to indicate that reinstalls upgrades should also be built from source. This provides that flag.

> I think the requirements need to be fixed to handle a non-brew python that is first in the user's path.

I agree; this solves a different and non-overlapping problem.

ilovezfs:
They are the same problem as far as I can tell. If the :python requirement were to correctly use the first python in the path, then there would be no reason for this hack since the correct Python would be used at upgrade time.

tdsmith:
It would not be, because the upgrade would be poured from a bottle, and therefore depend on Homebrew's python.

tdsmith:
The problem is that --build-from-source isn't sticky.

ilovezfs:
That is a "problem" (I'm not convinced it's actually a bug) that is in no way specific to the Python requirement. This is a one-off hack not a proper general fix.

tdsmith:
It is only a "problem" in the context of wanting to install an application with a custom Python. I do not propose that changing the behavior of --build-from-source is desirable. That's exactly why this is the right place to make this change.

ilovezfs:
The exact same behavior occurs for :ruby, :perl, :mysql, etc. I'm definitely :-1: on only changing the behavior for Python as you're proposing.

tdsmith:
Why? We have a lot of Python applications, we've only had requests for this in the context of Python applications, and this is a change that efficiently improves the UX for Python applications and doesn't block a more generic solution.

ilovezfs:
Since there's no emergency, I'd prefer to see it changed once in a general way, rather than allowing an immediate hack for python and then waiting for someone else to clean up the mess later.

tdsmith:
cf #6728; there is a desire to see this implemented in the near term.

I think it ought to be possible to add a --always-build-from-source or --build-from-source=always or --some-other-spelling flag but I think --with-custom-python is actually better because it shows up in `brew info` and suggests why you might find it useful. Otherwise, because the bottle-induced dependency switch is relatively subtle, this behavior is harder to discover and explain.

I know you have good ideas; do you have any thoughts?

"Hack" and "mess" are content-free words expressing a negative opinion about a solution; please consider using more neutral language in code reviews.

ilovezfs:
I will use the words "hack" and "mess" when I think something is a hack, or a mess, or both.

>there is a desire to see this implemented in the near term.

That does not constitute an emergency.

Regarding  a more general fix, why not handle it the same way we handle it when you don't have the CLT installed, and we refuse to pour the bottle for python and refuse to pour the bottle for gcc? If brew python is not first in your path (or ruby, or perl, etc.), we could refuse to pour the bottle, and build from source when there is a :python dependency. A `--force-bottle` could potentially be provided, or just advise in the bottle-reason about why the bottle is not being poured that making brew python first in the path will allow the bottle to be poured, just as installing the CLT would allow the python and gcc bottles to be poured.

tdsmith:
I think I can see how to do that and I think I like it.

Some thoughts:

* We don't want to accidentally use system Python for the reasons described in HEP 6. Should it be possible to intentionally use system Python? Maybe the :python dependencies should just never be satisfied by system Python. (I would expect that the alternative ways to satisfy :python will then be a pyenv or python.org distribution.)

* we should make sure that the python in PATH is not part of a virtualenv itself.

* Issuing a "run `brew install python` to get a bottle for this" notice is probably better than mentioning `--force-bottle` so we don't cause user confusion about cellar-any vs cellar-constrained formulas.

This should land at the same time as the fix to #712.

I'll open a PR for this soon unless other opinions appear in this thread.

I'm not here to censor you and I'm not personally offended because I've been working with you for a while but I do think more neutral language would be consistent with the values of consideration and respect described in our CoC. If you disagree, so be it; I don't want to drag this out. :)

ilovezfs:
>more neutral language would be consistent with the values of consideration and respect described in our CoC

I'd highly recommend not publicly invoking the CoC in reference to other maintainers in the future.

tdsmith:
I'm not trying to scold or threaten a tribunal. I think you felt I overstepped and I'm sorry you felt mentioning the CoC was inappropriate. This seems like something the team should establish shared norms about; cc @MikeMcQuaid, maybe this is something to discuss at some point (at another time, in another venue!). I _don't_ think anything you (ilovezfs) said in this thread needs to be discussed at any further length; I _do_ think it's worth discussing (or having Mike think about and set out, actually, would be fine with me) how we want to approach the CoC and how we can encourage each other to live up to it; I think it's an aspirational standard and I don't think that mentioning it should be threatening. :) Our conversations set the example for every contributor and I think it would be great if we could model a constructive response in public but if you/Mike/the team would prefer those discussions happen in private I'll bear that in mind moving forward.

ilovezfs:
No comment.

MikeMcQuaid:
My thoughts on this PR:
- I agree with @ilovezfs that a more general solution would be good but I do think that we should add a `with-custom-python` option for the short-term to alleviate complaints with our approach if @tdsmith thinks it's the best thing
- I think we need to figure out why Homebrew Python is "broken" for people's scripts etc. This is hard with little debugging information but I think it's the most important thing we do
- Perhaps there's a way we can/should install Python for things like Ansible but not put it first in the PATH. Make Homebrew's Python keg-only I think could solve a lot of problems 

My thoughts on CoC conversation:
- I don't think "hack" or "mess" are CoC-violating/unacceptable language
- I think ideally maintainers should sort out issues between themselves privately. In this case after doing "requested changes" stage (or a couple of comments later) it'd be good to just talk on e.g. Slack
- I think calling out maintainers on CoC violations is acceptable if you're unable to resolve issues privately and at least one other maintainer (privately) agrees. Otherwise (which 100% isn't what happened here but could in future) the CoC risks being a way of trying to win arguments

You're both doing good work and coming at this from different angles. In the short-term I agree with @tdsmith here, in the medium/long-term @ilovezfs.

MikeMcQuaid:
> Perhaps there's a way we can/should install Python for things like Ansible but not put it first in the PATH. Make Homebrew's Python keg-only I think could solve a lot of problems

Thinking about this more: this is the main issue I think people have. I `brew install ansible` and `/usr/local/bin/python` is now around to mess everything up. We've previously always solved this with `keg_only` and I think this is the right solution here, too. I'd propose the same thing with Ruby/Perl if we had the same complaints there but we don't. An alternative version is adding `/usr/local/bin/python2` but _not_ `/usr/local/bin/python`. Regardless, I think we need to do one of these two and pretty short-term. Thoughts @tdsmith?

MikeMcQuaid:
Thinking about this even more: I'm 👎 on this option being added for applications written in Python (e.g. stuff that can be installed with `pip`). If you install through Homebrew, you get it configured how we want. I think it's acceptable for e.g. MacVim/Vim to have a `custom-python` because of how bindings and stuff interact with it but I think it's a bad road to go down to add this option widely; we're giving people are asking for rather than what they want (or even need). I think what they want/need it for `brew install ansible` to not change `which python` and I think that's a very reasonable request.

tdsmith:
> I think we need to figure out why Homebrew Python is "broken" for people's scripts etc. This is hard with little debugging information but I think it's the most important thing we do

I think "broken" has been strictly hyperbolic; we've received I think a _very_ small volume of complaints relative to the volume of software we've deployed with a) virtualenv encapsulation and specifically b) virtualenvs that install Homebrew's python distribution. I haven't seen any genuinely confused users.

> Perhaps there's a way we can/should install Python for things like Ansible but not put it first in the PATH. Make Homebrew's Python keg-only I think could solve a lot of problems ... An alternative version is adding /usr/local/bin/python2 but not /usr/local/bin/python. Regardless, I think we need to do one of these two and pretty short-term.

I think people expect to get be able to run `python` after they run `brew install python` and I think changing that behavior will break a lot of scripts, invalidate a lot of community knowledge and documentation, and make setting up a development environment harder for new users or inexperienced developers. Generally I think the keg-only user experience is not good. :(

One alternative, about which I have some complex feelings but which I would strongly prefer to making the `python` formula keg-only, is to introduce a keg-only "Homebrew system python" that gets a dependency for the virtualenv-driven stuff and which we never expose and which has no interaction with the `python` formula.

tdsmith:
> "Homebrew system python"

This is terrible nomenclature and I apologize; "Homebrew-private python"?

MikeMcQuaid:
> I think "broken" has been strictly hyperbolic

I believe people when they say we've broken their scripts.

> One alternative, about which I have some complex feelings but which I would strongly prefer to making the python formula keg-only, is to introduce a keg-only "Homebrew system python" that gets a dependency for the virtualenv-driven stuff and which we never expose and which has no interaction with the python formula.

This is pretty ick but may be a better option. I do think changing the `python` formula itself may be a better option, though.

tdsmith:
I mean—I'll jump on it as soon as someone describes the problem they're having. The overwhelmingly most likely situation I can conjure is that someone had put modules in python distro A's default site-packages, installed Homebrew's python, Homebrew's python ended up first in PATH, Homebrew's python didn't have the distro-A site-packages in its search path, and `import foo` failed when Homebrew python ran the script.

I don't think there's any evidence at all that there's some other incompatibility specific to Homebrew's python distribution.

To be comprehensive: Ian also observed that Apple has applied a couple of patches to system Python that change stdlib API behavior. Those aren't going to land upstream and I don't think we should pull them; if someone's actually running into that, c'est la guerre, they should hardcode #!/usr/bin/python.

MikeMcQuaid:
> they should hardcode #!/usr/bin/python.

This is the problem I think we need to address. I don't think they should have to do this just because they want to `brew install ansible`.

tdsmith:
> I don't think they should have to do this just because they want to brew install ansible.

I... agree? I don't understand.

tdsmith:
To clarify, I guess, my position is that if an application user depends on the non-standard behavior of the Apple system python, they should manage the application themselves using the system python and assume the associated technical risk and labor on their own. I think this is doubly true since the trust-store behavior is most relevant for corporate environments.

`brew install ansible` has always delivered / will continue to deliver a functional ansible installation notwithstanding.

tdsmith:
Python Application Runtime proposal: https://github.com/Homebrew/brew-evolution/pull/24

MikeMcQuaid:
Passing on this as we've gone far enough that it'd probably introduce more rather than less confusion at this stage. Thanks for the PR anyway though @tdsmith 🙇

PR #1472, opened 2016-11-09
cask-test: fix reinstall a non installed Cask

umireon:
This is the change for the test itself.
- [x] Have you successfully run `brew cask-tests` with your changes locally?

-----

Trying to uninstall a non installed Cask made this test fail.
This commit remove that and thus this test passes.
Note that every Cask is uninstalled after each test case.


umireon:
I noticed that the result of `brew cask-tests` is completely ignored on BrewTestBot.

PR #1476, opened 2016-11-10
Disallow running `brew cask` as `root`.

reitermarkus:

-----

As discussed in https://github.com/Homebrew/brew/pull/1452.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -240,9 +240,6 @@ fi
 check-run-command-as-root() {
   [[ "$(id -u)" = 0 ]] || return
 
-  # Homebrew Cask may need `sudo` for system-wide installation.
-  [[ "$HOMEBREW_COMMAND" = "cask" ]] && return
-
   # Homebrew Services may need `sudo` for system-wide daemons.
   [[ "$HOMEBREW_COMMAND" = "services" ]] && return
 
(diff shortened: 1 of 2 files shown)


bcg62:
This breaks a lot of CM tools like chef/puppet that function with out having a tty available for a user to enter a password. 

Theres been a lot of back and fourth about this it seems and I'm still not sure what the correct solution is, could someone point me in the right direction?

https://github.com/caskroom/homebrew-cask/issues/5667
https://github.com/caskroom/homebrew-cask/issues/19180#issuecomment-188522310

bcg62:
thanks @reitermarkus 

This seems fine for some simple bash scripting but when trying to do something similar in a real CM tool its very intrusive and dangerous to modify sudo on the fly to preform an action, or do something like a keep alive in the background. Its also a broader discussion for the tool responsible referenced above.

reitermarkus:
Ok, I just read through the two discussions you linked above, and it seems like adding support for `SUDO_ASKPASS` is pretty trivial, already have it working locally.

bcg62:
Thanks, would you be able to share your example?

reitermarkus:
First you will have to save your askpass-script to `/path/to/script`.

Then you can either do this

```bash
SUDO_ASKPASS="/path/to/script" brew cask install cask-which-requires-sudo
```

or this

```bash
export SUDO_ASKPASS="/path/to/script"
brew cask install cask-which-requires-sudo
brew cask install another-cask-which-requires-sudo
```

PR #1497, opened 2016-11-13
keg: don't rely on #to_formula

### LINKED ISSUE #1496: `brew uninstall` can't uninstall formulae that don't exist in any tap
Note that, because #1082 is only enabled for non-developers, this bug can't be reproduced in developer mode.

This is the bug I referred to a few minutes ago in https://github.com/Homebrew/brew/pull/1082#discussion_r87722822.

If I've installed a formula from a file path or URL, and a formula with that same name doesn't exist in any of the taps I've installed, I can't uninstall it:

```sh
$ brew install hello2.rb
# ...
$ brew uninstall hello2
Error: No available formula with the name "hello2"
```

This happens because `uninstall` calls `to_formula` on the "hello2" `Keg`, which then raises an exception because it can't find the formula.

alyssais:

-----

See #1496 for an explanation of the problem.

This PR removes the call to `Keg#to_formula` in `Keg.find_some_installed_dependents`. Rather than grouping kegs by formula, it instead groups them by tap and name, which should have the same result but be more robust.

Fixes #1496.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -109,13 +109,20 @@ class Keg
     end
 
     keg_names = kegs.map(&:name)
-    kegs_by_name = kegs.group_by(&:to_formula)
+    kegs_by_source = kegs.group_by { |k| [k.name, Tab.for_keg(k).tap] }
+
     remaining_formulae.each do |dependent|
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Thanks for the quick turnaround here @alyssais 👏

PR #1498, opened 2016-11-13
Warn developers when uninstalling a dependency

alyssais:
- [ ] Have you successfully run `brew tests` with your changes locally? <ins>Only with #1497 merged.</ins>

-----

Suggested by @MikeMcQuaid in https://github.com/Homebrew/brew/pull/1082#discussion_r87691728.

Tests will fail as a side effect of #1496. When #1497 is merged, run the CI here again. It should go green and this should then be ready to merge.

I fixed the existing error message for non-developers as part of this too, so that it's output entirely to `stderr`. Previously, the first line was sent to `stderr` and subsequent ones went to `stdout`.

### DIFF
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -87,36 +87,64 @@ module Homebrew
     return false unless result = Keg.find_some_installed_dependents(kegs)
 
     if ARGV.homebrew_developer?
-      dependents_output_for_developers(*result)
+      DeveloperDependentsMessage.new(*result).output
     else
-      dependents_output_for_nondevelopers(*result)
+      NondeveloperDependentsMessage.new(*result).output
     end
 
     true
   end
 
-  def dependents_output_for_developers(requireds, dependents)
-    msg = requireds.join(", ")
-    msg << (requireds.count == 1 ? " is" : " are")
-    msg << " required by #{dependents.join(", ")}, which "
-    msg << (dependents.count == 1 ? "is" : "are")
-    msg << " currently installed."
-    msg << "\nYou can silence this warning with "
-    msg << "`brew uninstall --ignore-dependencies "
-    msg << "#{requireds.map(&:name).join(" ")}`."
-    opoo msg
+  class DependentsMessage
+    attr :reqs, :deps
+
+    def initialize(requireds, dependents)
+      @reqs = requireds
+      @deps = dependents
+    end
+
+    protected
+
+    def is(items)
+      items.count == 1 ? "is" : "are"
+    end
+
+    def it(items)
+      items.count == 1 ? "it" : "they"
+    end
+
+    def list(items)
+      items.join(", ")
+    end
+
+    def sample_command
+      "brew uninstall --ignore-dependencies #{list reqs.map(&:name)}"
+    end
+
+    def is_required_by_deps
+      "#{is reqs} required by #{list deps}, which #{is deps} currently installed"
+    end
   end
 
-  def dependents_output_for_nondevelopers(requireds, dependents)
-    msg = "Refusing to uninstall #{requireds.join(", ")} because "
-    msg << (requireds.count == 1 ? "it is" : "they are")
-    msg << " required by #{dependents.join(", ")}, which "
-    msg << (dependents.count == 1 ? "is" : "are")
-    msg << " currently installed."
-    msg << "\nYou can override this and force removal with "
-    msg << "`brew uninstall --ignore-dependencies "
-    msg << "#{requireds.map(&:name).join(" ")}`."
-    ofail msg
+  class DeveloperDependentsMessage < DependentsMessage
+    def output
+      opoo <<-EOS.undent
+        #{list reqs} #{is_required_by_deps}.
+        You can silence this warning with:
+          #{sample_command}
+      EOS
+    end
+  end
+
+  class NondeveloperDependentsMessage < DependentsMessage
+    def output
+      ofail <<-EOS.undent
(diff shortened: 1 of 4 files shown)


Library/Homebrew/cmd/uninstall.rb
```diff
@@ -78,28 +76,47 @@ def uninstall
...
 
+  def dependents_output_for_developers(requireds, dependents)
+    msg = requireds.join(", ")
+    msg << (requireds.count == 1 ? " is" : " are")
+    msg << " required by #{dependents.join(", ")}, which "
+    msg << (dependents.count == 1 ? "is" : "are")
+    msg << " currently installed."
+    msg << "\nYou can silence this warning with "
+    msg << "`brew uninstall --ignore-dependencies "
+    msg << "#{requireds.map(&:name).join(" ")}`."
```
MikeMcQuaid: Will merge this as-is but would be nice to refactor this to use `EOS.undent` and a bunch of variables declared before instead. Similarly below.

Library/Homebrew/utils.rb
```diff
@@ -376,6 +376,14 @@ def ignore_interrupts(opt = nil)
   trap("INT", std_trap)
 end
 
+def capture_stderr
```
MikeMcQuaid: Think you can use `shutup` here. Again, won't let it block this PR but for post-merge refactoring that'd be good if it's possible.

Library/Homebrew/utils.rb
```diff
@@ -376,6 +376,14 @@ def ignore_interrupts(opt = nil)
...
+  old, $stderr = $stderr, StringIO.new
```
MikeMcQuaid: `brew style` complains about the double assignment.

MikeMcQuaid:
Unfortunately still some (different) failing tests after rerunning.

alyssais:
Will aim to get this fixed today so you can merge.

MikeMcQuaid:
Thanks @alyssais!

Library/Homebrew/utils.rb
alyssais: I don't think `shutup` returns the captured output. Would you be open to it being refactored to return the captured `stdout` and `stderr`?

MikeMcQuaid: Yeh, consolidation there makes sense although might want to tweak the name if we do that too. Definitely one for another PR 👍

alyssais:
Looks like the failures went away after rebased. Thought CI would merge into latest master, but I guess not?

MikeMcQuaid:
It does normally. Maybe Something Weird Happened.

Library/Homebrew/cmd/uninstall.rb
```diff
@@ -87,36 +87,64 @@ def check_for_dependents(kegs)
...
-    msg = requireds.join(", ")
-    msg << (requireds.count == 1 ? " is" : " are")
-    msg << " required by #{dependents.join(", ")}, which "
-    msg << (dependents.count == 1 ? "is" : "are")
-    msg << " currently installed."
-    msg << "\nYou can silence this warning with "
-    msg << "`brew uninstall --ignore-dependencies "
-    msg << "#{requireds.map(&:name).join(" ")}`."
-    opoo msg
+  class DependentsMessage
```
alyssais: Put all the shared stuff between the two messages into a class so it can be re-used between both of them. It's a bit longer, but I think much cleaner now.

ilovezfs:
Can we have an environment variable to make these warnings go away?

alyssais:
@ilovezfs could do. What's the motivation?

ilovezfs:
To not have a regression where I'm spammed with warnings.

alyssais:
@ilovezfs If you're getting spammed with warnings where you shouldn't be, should we not be fixing that rather than hiding it under an environment variable?

We've already had one bug (#1496) that went undetected for longer than it should have because developers had different functionality than non-developers, and I'd like to stop something like that happening again.

Can you clarify if there's an actual regression where Homebrew is warning you where it shouldn't be, or if you just consider the fact that there's a warning in the first place to be a regression?

ilovezfs:
> the fact that there's a warning in the first place to be a regression

this.

alyssais:
That `cask-tests` failure looks unrelated but I'll investigate a little bit.

MikeMcQuaid:
> Can we have an environment variable to make these warnings go away?

No.

ilovezfs:
Because ...

MikeMcQuaid:
I'd like to see far more than one person request such a thing before we add it.

> We've already had one bug (#1496) that went undetected for longer than it should have because developers had different functionality than non-developers, and I'd like to stop something like that happening again.

This is a good argument.

ilovezfs:
The original feature was said to be disabled for developers. This makes it no longer disabled, and annoying in a novel way.

MikeMcQuaid:
Quoting myself:

> In the interests of actually getting this 🚢d we'll disable it on HOMEBREW_DEVELOPER at first but longer-term we're going to figure out something better than just adding HOMEBREW_DEVELOPER to things a few maintainers don't like. Every other package manager does things this way and it's a good idea. If you have a flag to override and don't like typing it: you can create an alias.

ilovezfs:
Not cool.

alyssais:
Yep, I don't think the `cask-tests` failure is related to this.

MikeMcQuaid:
@alyssais Have trigged a rebuild. CC `@reitermarkus` or `@jawshooah` if this is still broken afterwards.

MikeMcQuaid:
@alyssais 💚 Will merge later, nice work!

PR #1502, opened 2016-11-14
Load cask file in `.metadata` directory directly.

reitermarkus:

-----

Fixes https://github.com/caskroom/homebrew-cask/issues/26630.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cli/reinstall.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/reinstall.rb
@@ -19,7 +19,7 @@ module Hbc
                                                            "Casks", "#{cask_token}.rb")
 
                 # use the same cask file that was used for installation, if possible
-                installed_cask = Hbc.load(latest_installed_cask_file) if latest_installed_cask_file.exist?
+                installed_cask = CaskLoader.load_from_file(latest_installed_cask_file) if latest_installed_cask_file.exist?
               end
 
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -27,7 +27,7 @@ def metadata_master_container_path
     end
 
     def metadata_versioned_container_path
-      cask_version = version ? version : :unknown
+      cask_version = version ? version.sanitize : :unknown
```
jawshooah: Let's move the `version.sanitize` stuff to a separate PR since it's not necessary to load the cask file from the `.metadata` directory.

jawshooah: In fact, it seems like https://github.com/Homebrew/brew/pull/1502/commits/a799a3979f907307657ef9baddede51fba7d4d18 is sufficient to fix https://github.com/caskroom/homebrew-cask/issues/26630, and the `sanitize` method isn't necessary. Colons are fine, the problem was how we were using `curl`.

jawshooah: We weren't prepending `file://` to the path. For example:

```sh
$ echo 'Hello, World!' > foo:bar
$ curl "$PWD/foo:bar"
curl: (3) <url> malformed
$ curl "file://$PWD/foo:bar"
Hello, World!
```

PR #1515, opened 2016-11-15
Fix flaky tests.

alyssais:

-----

The issue was that the `Formula.installed` result was getting cached as `[]` by a previous test.

I've made `Tab#write` clear the cache if the tab didn't exist before, which _should_ stop this coming up again, but I wonder if we should clear the cache automatically in the teardown of every test as well, which would _definitely_ stop this coming up again, and shouldn't have too much overhead. <ins>Thoughts?</ins>

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1337,10 +1337,14 @@ class Formula
     end
   end
 
-  # Clear caches of .racks and .installed.
-  # @private
-  def self.clear_cache
+  # Clear cache of .racks
+  def self.clear_racks_cache
     @racks = nil
+  end
+
+  # Clear caches of .racks and .installed.
+  def self.clear_installed_formulae_cache
+    clear_racks_cache
     @installed = nil
   end
 
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -317,6 +317,10 @@ class Tab < OpenStruct
   end
 
   def write
+    # If this is a new installation, the cache of installed formulae
+    # will no longer be valid.
+    Formula.clear_cache unless tabfile.exist?
+
     CACHE[tabfile] = self
     tabfile.atomic_write(to_json)
   end
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -319,7 +319,7 @@ class Tab < OpenStruct
   def write
     # If this is a new installation, the cache of installed formulae
     # will no longer be valid.
-    Formula.clear_cache unless tabfile.exist?
+    Formula.clear_installed_formulae_cache unless tabfile.exist?
 
     CACHE[tabfile] = self
     tabfile.atomic_write(to_json)
--- a/Library/Homebrew/test/keg_test.rb
+++ b/Library/Homebrew/test/keg_test.rb
@@ -327,7 +327,6 @@ class InstalledDependantsTests < LinkTests
     f = stub_formula_name(name)
     keg = super
     Tab.create(f, DevelopmentTools.default_compiler, :libcxx).write
-    Formula.clear_cache
     keg
   end
 
--- a/Library/Homebrew/test/uninstall_test.rb
+++ b/Library/Homebrew/test/uninstall_test.rb
@@ -34,13 +34,11 @@ class UninstallTests < Homebrew::TestCase
   end
 
   def test_check_for_testball_f2s_when_developer
-    skip "Flaky test"
     assert_match "Warning", handle_unsatisfied_dependents
     refute_predicate Homebrew, :failed?
   end
 
   def test_check_for_dependents_when_not_developer
-    skip "Flaky test"
     run_as_not_developer do
       assert_match "Error", handle_unsatisfied_dependents
       assert_predicate Homebrew, :failed?


alyssais:
What went wrong with the CI here?

vladshablinsky:
Seems like `Formula.clear_cache` should be updated, but not cleared. Is this better than clearing the cache on teardown? I can be wrong, though.

MikeMcQuaid:
> Seems like Formula.clear_cache should be updated, but not cleared.

Yeh, this would be ideal.

alyssais:
Can change so new `Tab`s are added to the cache, **but** I still think the cache should be cleared in teardown, because tests are supposed to clear up any data they write after them.

Currently, `Keg`s are removed at the end of tests, but the cache isn't cleared. This leads to the opposite problem to what was an issue here — no kegs installed, but some formulae still being present in the `Formula.installed` cache.

MikeMcQuaid:
@alyssais Nuking cache on teardown is fine, would just be nice to avoid nuking cache on write if possible.

alyssais:
Several hours in, this is ending up being way more complicated than it should be. I will make sure the next thing I do here is getting this working properly, but for unexpected reasons it's going to take a small refactoring elsewhere before it's possible (can go into detail if you want, but I'm posting this right before I go to bed and I'm 99.9% sure that _has_ to be done for this can work properly, and I know what I need to do now).

I'd like to suggest, however, that this is merged as-is. It replaces behaviour that is clearly incorrect (it's only a matter of time before there's another weird bug that takes hours to track down because of it) with behaviour that, while not optimal, _is_ correct.

I also think it's quite important to clear the cache on teardown, but again, I'd prefer to make a separate PR for that and get this merged sooner if possible, because it's definitely an improvement to the current behaviour (and it means tests don't need to be skipped!).

MikeMcQuaid:
@alyssais Can't merge it as-is due to conflicts 😭. I think `Formula.clear_cache` should probably be renamed to `Formula.clear_installed_formulae_cache` or similar. My concern is if we end up using more caches in `Formula` that nuking them all is not what we want to be doing. I'm still not sure about having the cache cleared in the location it is, too; from the placement/code it's not clear which method call it's being cleared for. If it were for e.g. `runtime_dependencies` I'd suggest clearing it just before then would be best and perhaps splitting it into another `Formula.clear_runtime_dependencies_cache` to signal intent better.

alyssais:
>  think Formula.clear_cache should probably be renamed to Formula.clear_installed_formulae_cache or similar. My concern is if we end up using more caches in Formula that nuking them all is not what we want to be doing.

Agreed.

> I'm still not sure about having the cache cleared in the location it is, too; from the placement/code it's not clear which method call it's being cleared for. If it were for e.g. runtime_dependencies I'd suggest clearing it just before then would be best and perhaps splitting it into another Formula.clear_runtime_dependencies_cache to signal intent better.

Can you explain what you mean by this? It's not being cleared for any particular method call, it's being cleared so that the list is up-to-date for anything that wants to use it.

MikeMcQuaid:
> It's not being cleared for any particular method call, it's being cleared so that the list is up-to-date for anything that wants to use it.

That's my point: what is it that needs to use it? Just from the e.g. diff it's unclear if it's `CACHE[tabfile] =`,  `atomic_write` or `to_json` or a method called from inside one of them (I know it's the latter but I had to dig in to realise that).

alyssais:
Nothing in `Tab` needs to use `Formula.installed`, it's just that `Tab#write` is the only centralised location (as far as I can tell) that will **always** be called when the cache needs to be cleared.

MikeMcQuaid:
> Nothing in Tab needs to use Formula.installed, it's just that Tab#write is the only centralised location (as far as I can tell) that will always be called when the cache needs to be cleared.

What does need it (in these tests I mean)? Would like to try and figure out how to get it in a "nearer" place to that.

alyssais:
In these tests, the call comes from `Keg.find_some_installed_dependents`.

Remember, though, that just because this is somewhere we _know_ that not clearing the cache is causing incorrect behaviour, doesn't mean that it's the only place.

MikeMcQuaid:
How about clearing the cache in `formula_installer.rb` instead? Presumably that's what's calling `Tab.write`?

alyssais:
Not always — a lot of tests create their own kegs without using `FormulaInstaller`.

Could always add it to `FormulaInstaller` and come up with a helper method for creating kegs in tests, I guess?

MikeMcQuaid:
> Not always — a lot of tests create their own kegs without using FormulaInstaller.
> 
> Could always add it to FormulaInstaller and come up with a helper method for creating kegs in tests, I guess?

Yep, that sounds best 👍

MikeMcQuaid:
Thanks again @alyssais!

vladshablinsky:
Oh, I'm late with some of my thoughts on this.

>get this merged sooner if possible, because it's definitely an improvement to the current behaviour (and it means tests don't need to be skipped!).

Good point @alyssais.

>  I also think it's quite important to clear the cache on teardown, but again, I'd prefer to make a separate PR for that ...

Teardown seems to be a good place.


This fails only for three tests:

```diff
diff --git a/Library/Homebrew/formula.rb b/Library/Homebrew/formula.rb
index 7fe00cb..90454cd 100644
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1360,6 +1360,23 @@ class Formula
     end
   end
 
+  def self.update_cached_with_formula(formula)
+    racks.push(formula.rack) unless racks.include?(formula.rack)
+    installed.push(formula) unless installed.include?(formula)
+  end
+
+  def self.update_cached_with_rack(rack)
+    racks.push(rack) unless racks.include?(rack)
+    formula_for_rack = begin
+      Formulary.from_rack(rack)
+    rescue FormulaUnavailableError, TapFormulaAmbiguityError, TapFormulaWithOldnameAmbiguityError
+    end
+
+    unless formula_for_rack.nil?
+      installed.push(formula_for_rack) unless installed.include?(formula_for_rack)
+    end
+  end
+
   # An array of all installed {Formula}
   # @private
   def self.installed
diff --git a/Library/Homebrew/tab.rb b/Library/Homebrew/tab.rb
index 3bb7d8b..3902c09 100644
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -317,9 +317,7 @@ class Tab < OpenStruct
   end
 
   def write
-    # If this is a new installation, the cache of installed formulae
-    # will no longer be valid.
-    Formula.clear_installed_formulae_cache unless tabfile.exist?
+    Formula.update_cached_with_rack(tabfile.parent.parent) unless tabfile.exist?
 
     CACHE[tabfile] = self
     tabfile.atomic_write(to_json)

```

Fails occur when we call `.write` on tabs, which is expected. And which also proves what @MikeMcQuaid  said here:

> I'm still not sure about having the cache cleared in the location it is, too; from the placement/code it's not clear which method call it's being cleared for. If it were for e.g. runtime_dependencies I'd suggest clearing it just before then would be best and perhaps splitting it into another Formula.clear_runtime_dependencies_cache to signal intent better.

The right place for updating is probably `formula_installer` as Mike already said, however the directories are created here:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/build.rb#L133

PR #1520, opened 2016-11-15
cmd/list: follow aliases

### LINKED ISSUE #1514: "brew list --versions" should follow formula-name aliases
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [X/n/a] If you're seeing permission errors tried running `sudo chown -R $(whoami) $(brew --prefix)`?

### Propose a feature:

`brew list --versions` should "follow" aliases.

When a formula has an alias (say `mobile-shell <-> mosh`), `brew list` and `brew info` for the formula *or* the alias displays the same info entry. `brew search` also "follows" the alias. `brew search` also displays a check-mark if the formula *or* the alias is supplied.

However, `brew list --versions` is inconsistent with this behavior; `brew list --versions $alias` returns nothing, even if a formula was installed under an aliased name, while `brew list --versions $real-name` returns the correctly installed version.

*Specific feature proposal:*
- `brew list --versions;` behaves exactly as before, and lists all installed formulae by their "true" names.
- `brew list --versions $formula;` will "follow" aliases and perform the `list` operations on the "true" formula. Thus, `brew list --versions mosh` and `brew list --versions mobile-shell` will return the exact same output.
- Optional: a flag for `brew list` that enables the display of aliases next to true names, e.g. `brew list --versions --show-aliases;` results in the normal list, but with aliased packages appearing like `mobile-shell (mosh,otheralias) 1.2.6_3` or similar.

vladshablinsky:

-----
From #1514:

* `brew list --versions` behaves exactly as before, and lists all installed formulae by their "true" names.
* `brew list --versions $formula` will "follow" aliases and perform the list operations on the "true" formula. Thus, `brew list --versions mosh` and `brew list --versions mobile-shell` will return the exact same output.

Closes #1514

### DIFF
--- a/Library/Homebrew/cmd/list.rb
+++ b/Library/Homebrew/cmd/list.rb
@@ -107,7 +107,7 @@ module Homebrew
     names = if ARGV.named.empty?
       Formula.racks
     else
-      racks = ARGV.named.map { |n| HOMEBREW_CELLAR+n }
+      racks = ARGV.named.map { |n| Formulary.to_rack(n) }
       racks.select do |rack|
         Homebrew.failed = true unless rack.exist?
         rack.exist?


MikeMcQuaid:
Nice work @vladshablinsky!

Issue #1525, opened 2016-11-17
How should installed dependency resolution handle versions?


alyssais:
Currently, keg _A_ is only [considered](https://github.com/Homebrew/brew/blob/e6933b90dc750dc9aadb3bfb6acf1bf395942d57/Library/Homebrew/keg.rb#L349) to be a dependency of keg _B_ if the version of _A_ listed in _B_'s install receipt matches _A_'s version.

This means that, after a `brew upgrade`, the installed dependency resolution might not work any more.

My thinking is that it would be more correct to say _B_ depends on _A_ if _A_ is the currently linked keg for its formula (and maybe if _A_'s version is >= the version of _A_ that _B_ was installed with?).

I'm not entirely sure if I've thought this wrong properly, though, so I'm looking for feedback on this before I try to fix it.

This _might_ be what's causing #1524, but I'm not sure yet.

MikeMcQuaid:
> Currently, keg A is only considered to be a dependency of keg B if the version of A listed in B's install receipt matches A's version.

What would be the downsides of ignoring versions entirely here?

alyssais:
I'm not sure. It feels wrong that we'd consider a _B_ to depend on _A_ if _A_ is older than the version in _B_'s install receipt, but if _B_ is the linked version I guess that doesn't matter..

I think we probably should ignore versions, but only consider the linked version of a keg to have dependents.

MikeMcQuaid:
> I think we probably should ignore versions, but only consider the linked version of a keg to have dependents.

Yep, sounds sensible.

Issue #1526, opened 2016-11-17
Some tests are run multiple times


alyssais:
Here's a weird one (wow, I'm getting a lot of these today!): some of our tests are being run twice, when they shouldn't be.

Here's a simplified example of when it happens:

```ruby
class ATest < Homebrew::TestCase
  def setup
    # some fancy setup
  end

  def test_foo
    # …
  end
end

class BTest < ATest
  def setup
    super # get all A's fancy setup
    # some more fancy setup
  end

  def test_bar
    # …
  end
end
```

Rather counterintuitively (to me, at least), `test_foo` is run twice! Once on an `ATest` instance, and once on a `BTest` instance.

Now, usually, this wouldn't be much of a problem (other than the wasted time), but because `test_foo` isn't designed to be run in `BTest`'s environment, it can fail on the second run (it's really just luck that this hasn't been a problem before (to my knowledge)), when it shouldn't even have been run in the first place.

Instead, we need something like this:

```ruby
class ATestCase < Homebrew::TestCase
  # …
end

class ATest < ATestCase
  # …
end

class BTest < ATestCase
  # …
end
```

This is pretty clunky, but there doesn't seem to be a good way to do this with Minitest.

I think maybe I recall some discussion of transitioning Homebrew's tests to use `Minitest::Spec`. Was anything ever decided there? That would solve the problem because we could use contexts.

alyssais:
Oops, hit the create issue button before I'd finished writing the description. I've finished writing it now!

MikeMcQuaid:
> Rather counterintuitively (to me, at least), test_foo is run twice!

That's weird. Is this the case without using `parallel_test`? Has this always been the case or is a recent regression? Does it always/sometimes run a test twice? Would be great to try and have a failing PR that reproduces this.

alyssais:
I read through the Minitest source and it doesn't look like it does anything to prevent this. It's unlikely that it's a parallel_test issue because these tests tend to be in the same file, and it parallel_test works at a file level — it doesn't know about individual tests.

#1527 fails because of this if dfbf07f isn't included. I can put that up as a separate PR if you want, but all you need to do to see the failure is drop that commit.

MikeMcQuaid:
I guess what I'm trying to understand is: why does this happen for certain tests but not e.g. every test? It's any time you use `super`?

alyssais:
[...]

PR #1527, opened 2016-11-17
keg: update installed dependency resolution

alyssais:

-----

See #1525 for an explanation.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -385,7 +385,7 @@ class Keg
   end
 
   def installed_dependents
-    return [] unless linked?
+    return [] if !linked? && !to_formula.keg_only?
     tap = Tab.for_keg(self).source["tap"]
     Keg.all.select do |keg|
       tab = Tab.for_keg(keg)
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -385,7 +385,7 @@ class Keg
   end
 
   def installed_dependents
-    return [] if !linked? && !to_formula.keg_only?
+    return [] unless optlinked?
     tap = Tab.for_keg(self).source["tap"]
     Keg.all.select do |keg|
       tab = Tab.for_keg(keg)
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -385,6 +385,7 @@ class Keg
   end
 
   def installed_dependents
+    return [] unless linked?
     tap = Tab.for_keg(self).source["tap"]
     Keg.all.select do |keg|
       tab = Tab.for_keg(keg)
@@ -394,12 +395,10 @@ class Keg
         # in case of conflicts between formulae from different taps.
         begin
           dep_formula = Formulary.factory(dep["full_name"])
-          next false unless dep_formula == to_formula
+          dep_formula == to_formula
         rescue FormulaUnavailableError
-          next false unless "#{tap}/#{name}" == dep["full_name"]
+          next "#{tap}/#{name}" == dep["full_name"]
         end
-
-        dep["version"] == version.to_s
       end
     end
   end
--- a/Library/Homebrew/test/keg_test.rb
+++ b/Library/Homebrew/test/keg_test.rb
@@ -473,4 +473,12 @@ class InstalledDependantsTests < LinkTestCase
     assert_empty @keg.installed_dependents
     assert_nil Keg.find_some_installed_dependents([@keg])
   end
+
+  def test_keg_only
+    @keg.unlink
+    Formula["foo"].class.keg_only "a good reason"
+    dependencies [{ "full_name" => "foo", "version" => "1.1" }] # different version
+    assert_equal [@dependent], @keg.installed_dependents
+    assert_equal [[@keg], ["bar 1.0"]], Keg.find_some_installed_dependents([@keg])
+  end
 end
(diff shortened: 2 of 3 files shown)


Library/Homebrew/test/keg_test.rb
```diff
@@ -44,7 +44,9 @@ def teardown
     rmtree HOMEBREW_PREFIX/"bin"
     rmtree HOMEBREW_PREFIX/"lib"
   end
+end
 
+class LinkTests < LinkTestCase
```
MikeMcQuaid: Let's hold off these cleanups for now or at least punt to a separate PR.

Library/Homebrew/keg.rb
```diff
@@ -339,14 +339,14 @@ def to_formula
   end
 
   def installed_dependents
+    return [] unless linked?
```
MikeMcQuaid: This will presumably return for anything `keg_only`? I'm not sure that's what we want. Perhaps we should still do it for things that aren't linked, too, but just go purely by `name` (which I realise contradicts what I said earlier).

Library/Homebrew/test/keg_test.rb
alyssais: This isn't a cleanup for the sake of it. The tests fail if I don't do this. That's how I discovered #1526!

Library/Homebrew/keg.rb
alyssais: I'm not entirely sure I understand how `keg_only` works, exactly. If a keg isn't linked, is it even possible for it to be a runtime dependency? How does that work, generally?

Library/Homebrew/test/keg_test.rb
MikeMcQuaid: Given your elaborate I understand now and this is 👍

Library/Homebrew/keg.rb
alyssais: Ah, I see. In that case, wouldn't the hard-coded path contain a version?

MikeMcQuaid: Not necessarily; they can also use `opt/`

alyssais: Okay, in that case I think you're right. `keg_only` should just check name.

MikeMcQuaid: I'm guessing I'm wondering at that point why it's worth anything else checking not just the name?

alyssais: If I have two versions of _A_ installed, and _B_ depends on _A_, if we just checked the name, I wouldn't be able to uninstall my old version of _A_ that isn't needed anymore. We need a way to differentiate the "current" version of _A_ from the other installed versions, and checking if it's linked seems to be the most appropriate way to do that.

MikeMcQuaid: Yeh, of course, thanks. I think checking the opt link is probably the best way of doing this and then opt-out (pun partially intended) of this safety feature if there's no `opt` link at all.

alyssais: So, a non-opt-linked `keg_only` keg isn't considered to have any installed dependents?

MikeMcQuaid: Yep. Anything without an opt-link can be considered outdated or just plain broken.

Library/Homebrew/keg.rb
```diff
@@ -339,14 +339,14 @@ def to_formula
   end
 
   def installed_dependents
+    return [] if !linked? && !to_formula.keg_only?
```
MikeMcQuaid: Let's go for `unless optlinked?` rather than `if !linked? && !to_formula.keg_only?`

alyssais:
Oh, no! I totally lost track of this PR and didn't realise it only needed a minor change to (hopefully) be done. Sorry!

alyssais:
In the process of tracking down the test failures here.

ilovezfs:
Does this rewrite install receipts of dependents after the fact?

alyssais:
It shouldn't…

alyssais:
No test-bot here?

ilovezfs:
@BrewTestBot test this please

alyssais:
@ilovezfs is the test bot working? https://bot.brew.sh suggests it hasn't run for 19 days…

ilovezfs:
project was disabled

ilovezfs:
@BrewTestBot test this please

ilovezfs:
@alyssais I guess we're relying on Travis temporarily or something: https://travis-ci.org/Homebrew/brew/builds/186393542

MikeMcQuaid:
Not temporarily.

ilovezfs:
Poor @BrewTestBot

MikeMcQuaid:
Lucky @BrewTestBot; she no longer has to work quite as hard 😉

MikeMcQuaid:
Looking good now. @alyssais do you consider this good to 🚢? If so, can you provide a wee test plan I can run locally to reproduce the issue this is fixing? Thanks!

alyssais:
Aww, I'll miss you @BrewTestBot!

alyssais:
@MikeMcQuaid what do you think of https://github.com/Homebrew/brew/issues/1555#issuecomment-263463117? I had a bit of trouble following it.

MikeMcQuaid:
Not sure what this looks like after https://github.com/Homebrew/brew/issues/1737 is merged but probably worth a rebase anyway and we can discuss the diff then.

alyssais:
@mikemcquaid this was rebased

(Not sure if you get notifications for that or not?)

MikeMcQuaid:
You don't get notifications on force-pushes just regular pushes.

MikeMcQuaid:
Still a conflict here unfortunately 😭

alyssais:
@mikemcquaid rebased again again

MikeMcQuaid:
@alyssais Thanks again again

PR #1533, opened 2016-11-18
tests: ignore test files in test/vendor/bundle

jawshooah:

-----

Ensure that files matching `test/vendor/bundle/**/*_test.rb` are not picked up when running `brew tests`.

### DIFF
--- a/Library/Homebrew/dev-cmd/tests.rb
+++ b/Library/Homebrew/dev-cmd/tests.rb
@@ -46,6 +46,7 @@ module Homebrew
 
       files = Dir.glob("test/**/*_test.rb")
                  .reject { |p| !OS.mac? && p.start_with?("test/os/mac/") }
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -46,6 +46,7 @@ def tests
 
       files = Dir.glob("test/**/*_test.rb")
                  .reject { |p| !OS.mac? && p.start_with?("test/os/mac/") }
+                 .reject { |p| p.start_with?("test/vendor/bundle/") }
```
MikeMcQuaid: Maybe just `test/vendor` is worth doing here? Thoughts?

jawshooah: No strong feelings either way, just figured it was better to be specific in case we decided to test censored code for e.g. integrations at some point.

PR #1534, opened 2016-11-18
utils/json: Replace OkJson with Ruby's core JSON.

woodruffw:

-----

Removes OkJson from `vendor/`, as Ruby has included its own JSON parser in corelib since 1.9.x and Homebrew is now guaranteed to be running on 2.x+.

Some unscientific local tests with `brew info --json=v1 --all` show a pretty substantial performance increase.

With `OkJson`:

```
$ time brew info --json=v1 --all >/dev/null

real   0m18.275s
user   0m17.859s
sys    0m0.545s
```

With core `json`:

```
$ time brew info --json=v1 --all >/dev/null

real   0m10.142s
user   0m9.553s
sys    0m0.501s
```


MikeMcQuaid:
Nice work here.

Library/Homebrew/utils/json.rb
```diff
@@ -1,4 +1,4 @@
-require "vendor/okjson"
+require "json"
 
 module Utils
   module JSON
```
MikeMcQuaid: @woodruffw We could consider removing this entire module and replacing it with calls to these methods instead.

woodruffw:
Thanks! I also have some further ideas for speeding up JSON generation (I have ulterior motives in mind), but they're more involved.

MikeMcQuaid:
> Thanks! I also have some further ideas for speeding up JSON generation (I have ulterior motives in mind), but they're more involved.

Sounds exciting 😁

Library/Homebrew/utils/json.rb
woodruffw: Sounds good to me; that would avoid the ugly `::JSON` scoping. I'll have a PR ready tonight or tomorrow.

MikeMcQuaid: `stringify_keys` doesn't seem to be JSON-specific so could be moved (and we could vendor some of ActiveSupport for such things if we really wanted) and will need to check the exceptions behave correctly but otherwise should be pretty simple.

woodruffw: Done! #1542 has the details.

PR #1535, opened 2016-11-18
Cask: Add the choices option to pkg stanza

umireon:

-----

installer command accepts -applyChoiceChangesXML option to change customize options on the GUI installer from the commandline.
(`man installer` for more detailed information)

Some `.pkg` packages need to be supplied this choice changes XML.
Currently that is achieved by some tricks including writing a XML on preflight and execute `installer` command via the arbitrary command execution (`installer` stanza).
This is confusing because these Casks cannot use the standard `pkg` stanza for `.pkg` packages.

The introduced option `choice` enables the choice changes to be supplied
via pkg stanza without tricks in preflight code.

## Example

```
pkg 'MyFancyPkg/Fancy.pkg',
    choices: [
      {
        'choiceIdentifier' => 'choice1',
        'choiceAttribute'  => 'selected',
        'attributeSetting' => 1,
      },
    ]
```

`-applyChoiceChangesXML` option is used in [osxfuse Cask], for example.
This Cask can be written in [a more concise way] with choice option.

[osxfuse Cask]: https://github.com/caskroom/homebrew-cask/blob/master/Casks/osxfuse.rb
[a more concise way]: https://gist.github.com/umireon/ecb8908f8b2514abba907a5ca5c5e233/revisions#diff-15bc398082ca9019e0041ddbadc37bdf

## Related issues
caskroom/homebrew-cask#9578 (closed for lack of interest)
caskroom/homebrew-cask#10137 (stalled?)
caskroom/homebrew-cask#15431 (request for this feature)

Note that I don't mean adding the `--with-xxx` feature mentioned in caskroom/homebrew-cask#10137.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
@@ -54,17 +54,25 @@ module Hbc
         ]
         args << "-verboseR" if Hbc.verbose
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
-        if pkg_install_opts :choices
-          choices_file = choices_xml
-          args << "-applyChoiceChangesXML" << choices_file.path
+        with_choices_file pkg_install_opts(:choices) do |choices_path|
+          args << "-applyChoiceChangesXML" << choices_path if choices_path
+          @command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
         end
-        @command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
       end
 
-      def choices_xml
-        file = Tempfile.open(["", ".xml"])
-        file.write Plist::Emit.dump(pkg_install_opts(:choices))
-        file
+      def with_choices_file(choices)
+        unless choices
+          yield nil
+          return
+        end
+
+        begin
+          file = Tempfile.new(["choices", ".xml"])
+          file.write Plist::Emit.dump(choices)
+          yield file.path
+        ensure
+          file.close(true)
+        end
       end
     end
   end
--- a/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
@@ -2,6 +2,8 @@ require "hbc/artifact/base"
 
 require "hbc/utils/hash_validator"
 
+require "vendor/plist/plist"
+
 module Hbc
   module Artifact
     class Pkg < Base
@@ -16,7 +18,7 @@ module Hbc
         @pkg_install_opts = pkg_description.shift
         begin
           if @pkg_install_opts.respond_to?(:keys)
-            @pkg_install_opts.extend(HashValidator).assert_valid_keys(:allow_untrusted)
+            @pkg_install_opts.extend(HashValidator).assert_valid_keys(:allow_untrusted, :choices)
           elsif @pkg_install_opts
             raise
           end
@@ -52,7 +54,20 @@ module Hbc
         ]
         args << "-verboseR" if Hbc.verbose
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
+        if pkg_install_opts :choices
+          args << "-applyChoiceChangesXML"
+          args << choices_xml
+        end
         @command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
+      end
+
+      def choices_xml
+        path = @cask.staged_path.join("Choices.xml")
+        unless File.exist? path
+          choices = pkg_install_opts :choices
+          IO.write path, Plist::Emit.dump(choices)
+        end
+        path
       end
     end
   end
--- a/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
@@ -55,19 +55,16 @@ module Hbc
         args << "-verboseR" if Hbc.verbose
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
         if pkg_install_opts :choices
-          args << "-applyChoiceChangesXML"
-          args << choices_xml
+          choices_file = choices_xml
+          args << "-applyChoiceChangesXML" << choices_file.path
         end
         @command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
       end
 
       def choices_xml
-        path = @cask.staged_path.join("Choices.xml")
-        unless File.exist? path
-          choices = pkg_install_opts :choices
-          IO.write path, Plist::Emit.dump(choices)
-        end
-        path
+        file = Tempfile.open(["", ".xml"])
+        file.write Plist::Emit.dump(pkg_install_opts(:choices))
+        file
       end
     end
   end
--- a/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
@@ -54,24 +54,23 @@ module Hbc
         ]
         args << "-verboseR" if Hbc.verbose
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
-        with_choices_file pkg_install_opts(:choices) do |choices_path|
+        with_choices_file do |choices_path|
           args << "-applyChoiceChangesXML" << choices_path if choices_path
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -52,8 +54,21 @@ def run_installer(pkg_description)
...
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
+        if pkg_install_opts :choices
+          args << "-applyChoiceChangesXML"
+          args << choices_xml
+        end
         @command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
       end
+
+      def choices_xml
+        path = @cask.staged_path.join("Choices.xml")
```
reitermarkus: Could you change this to use a [Tempfile](https://github.com/Homebrew/brew/blob/8d15bddedb2f5796193f79a9a97cd6271367e251/Library/Homebrew/cask/lib/hbc/container/dmg.rb#L67-L77)? Since this is generated from the cask file, there is no need to keep this around and clutter up the `staged_path` with it.

Library/Homebrew/cask/test/cask/artifact/pkg_test.rb
```diff
@@ -30,4 +30,41 @@
...
+    before do
+      @cask = Hbc.load("with-choices")
+      shutup do
+        TestHelper.install_without_artifacts(@cask)
+      end
+    end
+
+    it "passes the choice changes xml to the system installer" do
+      pkg = Hbc::Artifact::Pkg.new(@cask,
+                                   command: Hbc::FakeSystemCommand)
```
reitermarkus: You can put this in one line.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -52,8 +54,21 @@ def run_installer(pkg_description)
         ]
         args << "-verboseR" if Hbc.verbose
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
+        if pkg_install_opts :choices
+          args << "-applyChoiceChangesXML"
+          args << choices_xml
```
reitermarkus: This can be one line, also:

```ruby
args << "-applyChoiceChangesXML" << choices_xml if pkg_install_opts :choices
```

umireon:
Thank you for reviewing, @reitermarkus 🐱 
I have changed the code.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -52,8 +54,19 @@ def run_installer(pkg_description)
         ]
         args << "-verboseR" if Hbc.verbose
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
+        if pkg_install_opts :choices
+          choices_file = choices_xml
```
umireon: Holding the Tempfile instance as a variable to ensure the file is not deleted until install_phase ends.

reitermarkus: As it it now, it isn't actually deleted. You should add an `ensure` block at the end of the method.

Library/Homebrew/cask/test/cask/artifact/pkg_test.rb
```diff
@@ -30,4 +30,42 @@
...
+        </array>
+        </plist>
+      EOS
+      file.stubs path: Pathname.new("/tmp/choices.xml")
+      Tempfile.expects(:open).returns(file)
+      Hbc::FakeSystemCommand.expects_command(["/usr/bin/sudo", "-E", "--", "/usr/sbin/installer", "-pkg", @cask.staged_path.join("MyFancyPkg", "Fancy.pkg"), "-target", "/", "-applyChoiceChangesXML", @cask.staged_path.join("/tmp/choices.xml")])
+
+      shutup do
+        pkg.install_phase
+      end
```
reitermarkus: Also test that the `choices.xml` is deleted after the `install_phase`.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
umireon: BTW, I'm interested why that does not removes the file though `Tempfile` does so with GC according to doc.

umireon:
@reitermarkus I changed the code to explicitly close and delete the xml file and tests to check that.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -52,7 +54,25 @@ def run_installer(pkg_description)
...
+          return
+        end
+
+        begin
+          file = Tempfile.new(["choices", ".xml"])
+          file.write Plist::Emit.dump(choices)
+          yield file.path
+        ensure
+          file.close(true)
+        end
```
reitermarkus: Now you could also use a `Tempfile.open do … end` block.

umireon: I'm afraid I don't understand your point.

`::open` with a block does not delete the file but the finalizer does.
This behaviour is essentially same as `::new`; both delete no files until Ruby's GC runs.
The difference between them is `::open` closes the IO handle, which is also eventually closed with `::new` with GC.
I am sure ensuring the temp file deleted needs to call `#close(true)` or `#unlink` explicitly, or use `::create`, which is unavailable at 2.0.0.

See http://ruby-doc.org/stdlib-2.0.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-open

I wish I did not miss any major points.

umireon:
@reitermarkus Do you concern that the temp file remains at the abnormal exit (e.g. `exit!`) ...?
`Tempfile.open` does not remove the file at the abnormal exit, neither.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -61,17 +61,15 @@ def run_installer(pkg_description)
       end
 
       def with_choices_file(choices)
-        unless choices
-          yield nil
-          return
-        end
+        return yield nil unless choices
 
+        file = Tempfile.new(["choices", ".xml"])
```
reitermarkus: This can stay inside the block, as this doesn't change anything. Other than that everything is looking fine.

reitermarkus:
> @reitermarkus Do you concern that the temp file remains at the abnormal exit (e.g. `exit!`) ...?
> Tempfile.open does not remove the file at the abnormal exit, neither.

My only motivation was that it would look nicer with a block:

```ruby
Tempfile.open(["choices", ".xml"]) do |file|
  begin
    file.write Plist::Emit.dump(choices)
    file.close
    yield file.path
  ensure
    file.unlink
  end
end

```

umireon:
@reitermarkus I have never come up with that because of closing twice, but it is nice 👍

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -52,7 +54,24 @@ def run_installer(pkg_description)
...
+        end
+      end
+
+      def with_choices_file(choices)
+        return yield nil unless choices
+
+        Tempfile.open(["choices", ".xml"]) do |file|
+          begin
+            file.write Plist::Emit.dump(choices)
+            file.close
```
umireon: Close immediately after the file handle is no longer needed.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -52,7 +54,24 @@ def run_installer(pkg_description)
...
         args << "-verboseR" if Hbc.verbose
         args << "-allowUntrusted" if pkg_install_opts :allow_untrusted
-        @command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
+        with_choices_file pkg_install_opts(:choices) do |choices_path|
+          args << "-applyChoiceChangesXML" << choices_path if choices_path
+          @command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
+        end
+      end
+
+      def with_choices_file(choices)
```
reitermarkus: You can remove the `choices` argument and directly use `pkg_install_opts(:choices)` since this is an instance method.

umireon:
Thank you for reviewing, @reitermarkus !
The trivial argument is removed.

reitermarkus:
Thanks, @umireon! Tested locally with `osxfuse`, everything is working.

PR #1537, opened 2016-11-18
install: suppress redundunt warnings with `depends_on` requirement

umireon:

-----

When a formula depends on any [requirements], they evaluated at most three times.

- Before locking: [verify_deps_exist in FormulaInstaller#prelude]
- Before installing: [compute_dependencies in FormulaInstaller#install]
- Before building formula: [compute_and_install_dependencies in FormulaInstaller#install]

These methods call [FormulaInstaller#check_requirements], which `puts` the warning messages generated by the specified requirement class.

The problem is when a non-fatal requirement is specified and thus evaluated three times,
mostly the same warning message is also emitted three times.

[requirements]: https://github.com/Homebrew/brew/blob/59e2d6772158d8df0735708ae78ddec6ccc68026/Library/Homebrew/requirement.rb
[verify_deps_exist in FormulaInstaller#prelude]: https://github.com/Homebrew/brew/blob/59e2d6772158d8df0735708ae78ddec6ccc68026/Library/Homebrew/formula_installer.rb#L127
[compute_dependencies in FormulaInstaller#install]: https://github.com/Homebrew/brew/blob/59e2d6772158d8df0735708ae78ddec6ccc68026/Library/Homebrew/formula_installer.rb#L200
[compute_and_install_dependencies in FormulaInstaller#install]: https://github.com/Homebrew/brew/blob/59e2d6772158d8df0735708ae78ddec6ccc68026/Library/Homebrew/formula_installer.rb#L262
[FormulaInstaller#check_requirements]: https://github.com/Homebrew/brew/blob/59e2d6772158d8df0735708ae78ddec6ccc68026/Library/Homebrew/formula_installer.rb#L336-L348

## Example
```
$ brew install python
$ brew install gdb --with-python
Updating Homebrew...
gdb: A build of GDB using a brewed Python was requested, but Python is not
a universal build.

GDB requires Python to be built as a universal binary or it will fail
if attempting to debug a 32-bit binary on a 64-bit host.
gdb: A build of GDB using a brewed Python was requested, but Python is not
a universal build.

GDB requires Python to be built as a universal binary or it will fail
if attempting to debug a 32-bit binary on a 64-bit host.
gdb: A build of GDB using a brewed Python was requested, but Python is not
a universal build.

GDB requires Python to be built as a universal binary or it will fail
if attempting to debug a 32-bit binary on a 64-bit host.
==> Using the sandbox
==> Downloading https://ftpmirror.gnu.org/gdb/gdb-7.12.tar.xz

```

This change restricts printing the warning messages only when a bottle is
successfully installed or before building.
Since this timing is after the final dependency computation for each cases,
the warnings will be most useful to check what is not yet satisfied.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -251,6 +251,7 @@ class FormulaInstaller
         opoo "Bottle installation failed: building from source."
         raise BuildToolsError, [formula] unless DevelopmentTools.installed?
       else
+        puts @requirement_messages
         @poured_bottle = true
       end
     end
@@ -260,6 +261,7 @@ class FormulaInstaller
     unless @poured_bottle
       not_pouring = !pour_bottle || @pour_failed
       compute_and_install_dependencies if not_pouring && !ignore_deps?
+      puts @requirement_messages
       build
       clean
 
@@ -334,17 +336,21 @@ class FormulaInstaller
   end
 
   def check_requirements(req_map)
+    @requirement_messages = []
     fatals = []
 
     req_map.each_pair do |dependent, reqs|
       next if dependent.installed?
       reqs.each do |req|
-        puts "#{dependent}: #{req.message}"
+        @requirement_messages << "#{dependent}: #{req.message}"
         fatals << req if req.fatal?
       end
     end
 
-    raise UnsatisfiedRequirements, fatals unless fatals.empty?
+    return if fatals.empty?
+
+    puts @requirement_messages
+    raise UnsatisfiedRequirements, fatals
   end
 
   def install_requirement_default_formula?(req, dependent, build)
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -52,6 +52,7 @@ class FormulaInstaller
     @debug = false
     @options = Options.new
     @invalid_option_names = []
+    @requirement_messages = []
 
     @@attempted ||= Set.new
 
@@ -251,7 +252,7 @@ class FormulaInstaller
         opoo "Bottle installation failed: building from source."
         raise BuildToolsError, [formula] unless DevelopmentTools.installed?
       else
-        puts @requirement_messages
+        puts_requirement_messages
         @poured_bottle = true
       end
     end
@@ -261,7 +262,7 @@ class FormulaInstaller
     unless @poured_bottle
       not_pouring = !pour_bottle || @pour_failed
       compute_and_install_dependencies if not_pouring && !ignore_deps?
-      puts @requirement_messages
+      puts_requirement_messages
       build
       clean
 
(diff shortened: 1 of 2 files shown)


Library/Homebrew/formula_installer.rb
```diff
@@ -251,6 +251,7 @@ def install
         opoo "Bottle installation failed: building from source."
         raise BuildToolsError, [formula] unless DevelopmentTools.installed?
       else
+        puts @requirement_messages
```
MikeMcQuaid: Make a new function with `puts @requirement_messages` like `def puts_requirement_messages` or similar which outputs nothing i.e. no newline which `@requirement_messages` is empty

umireon: `@requirement_messages` is an `Array` and AFAIK `puts []` outputs nothing (no newline).
Therefore, this code already works as you suggested.

umireon:
Thank you for reviewing, @MikeMcQuaid !
IMO `puts_requirement_messages` is too trivial to define, but it makes senses in encapsulation.
```ruby
def puts_requirement_messages
  puts @requirement_messages
end
```

umireon:
Not relevant to this PR, however, the actual problem might be evaluating the requirements three times.
When I was modifying a Formula with Requirement class, I expected the requirement evaluated once, but it was not the case.

MikeMcQuaid:
> IMO puts_requirement_messages is too trivial to define, but it makes senses in encapsulation.

As mentioned there's more logic for it to do: don't `puts` anything if it's empty.

MikeMcQuaid:
> As mentioned there's more logic for it to do: don't puts anything if it's empty.

So something like:

```ruby
def puts_requirement_messages
  return unless @requirement_messages
  return if @requirement_messages.empty?
  puts @requirement_messages
end
```

umireon:
Add puts_requirement_messages for the case `@requirement_messages` is a String or nil!

umireon:
`@requirement_messages` should be initialized with an empty string (`""`) instead of an empty Array (`[]`)?

vladshablinsky:
> @requirement_messages should be initialized with an empty string ("") instead of an empty Array ([])?

I think everything is fine with current initialization.

umireon:
I think the suggested guards `return unless @requirement_messages; return if @requirement_messages.empty?` are unnecessary when `@requirement_messages` is initialized with `[]`.

MikeMcQuaid:
> are unnecessary when @requirement_messages is initialized with [].

If it's called before initialisation in future then it will output an empty line if it's `nil` which is what we want to avoid.

umireon:
Anyway, the foolproof version of `puts_requirement_messages` is implemented in bcc9002 .
I understand this kind of robustness is important for such a big OSS product.
Thank you for discussing, @MikeMcQuaid , @vladshablinsky 🍎

MikeMcQuaid:

PR #1547, opened 2016-11-21
cask: Delete package dir if it is symlink on the uninstall

umireon:

-----

When a Formula is converted into a Cask (e.g. osxfuse),
a symlink remains at the place of package dir.
This change ensure to remove such leftovers.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/pkg.rb
+++ b/Library/Homebrew/cask/lib/hbc/pkg.rb
@@ -72,7 +72,12 @@ module Hbc
     end
 
     def _rmdir(path)
-      @command.run!("/bin/rmdir", args: ["--", path], sudo: true) if path.children.empty?
+      return unless path.children.empty?
+      if path.symlink?
+        @command.run!("/bin/rm", args: ["-f", "--", path], sudo: true)
+      else
+        @command.run!("/bin/rmdir", args: ["--", path], sudo: true)
+      end
     end
 
     def _with_full_permissions(path)


MikeMcQuaid:
Makes sense to me. @jawshooah @reitermarkus?

nilbus:
👍 See https://github.com/Homebrew/homebrew-fuse/issues/86 for the symptom.

jawshooah:
Strange, seems like [this bit](https://github.com/umireon/brew/blob/63e7fb5cae12ad751b4f3f83f6aa767d7c3c9f00/Library/Homebrew/cask/lib/hbc/pkg.rb#L21-L24) should take care of symlinks. Is the problem that these symlinks are listed as directories in the BOM?

umireon:
@jawshooah Yes, symlinks which links to a directory is not handled.
This situation is caused when the package moved from Formula to Cask, and then install the Cask without uninstalling the Formula, and then uninstall the Cask.

## Example
```
$ ls -ld /usr/local/include/osxfuse
lrwxr-xr-x  1 umireon  staff  26 Nov 24 02:06 /usr/local/include/osxfuse -> /usr/local/Cellar/osxfuse/x.x.x/include
$ brew cask install osxfuse
🍺  osxfuse was successfully installed!
$ ls /usr/local/include/osxfuse/
fuse	fuse.h
$ brew cask uninstall osxfuse
==> Running uninstall process for osxfuse; your password may be necessary
==> Unloading kernel extension com.github.osxfuse.filesystems.osxfusefs
==> Removing files from pkgutil Bill-of-Materials
==> rmdir: /usr/local/include/osxfuse: Not a directory
Error: Command failed to execute!

==> Failed command:
/usr/bin/sudo -E -- /bin/rmdir -- #<Pathname:/usr/local/include/osxfuse>

==> Standard Output of failed command:


==> Standard Error of failed command:
rmdir: /usr/local/include/osxfuse: Not a directory


==> Exit status of failed command:
#<Process::Status: pid 68354 exit 1>
```

jawshooah:
Got it, thanks for the explanation.

Issue #1558, opened 2016-11-24
Sudo restriction breaks linkapps and cask for non-admin users


JamesMcMahon:
I'm having some trouble with brew post `sudo` restrictions.

```
Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.
```

In general this sounds like good idea, but the problem I'm running into is my user doesn't is non-admin so I don't have access to `/Applications/` which breaks any operation that relies on write access to the directory, eg, `linkapps` and pretty much anything having to do with cask.

Now while I could give my user permissions to write into `/Applications/`, I feel like that is counter-productive in terms of security. I'd actually _prefer_ to be able to briefly `sudo` an install script then have my user constantly have full access to the the `/Applications/` directory. 

**Are their any work arounds to the current `sudo` restriction?**

Apologies in advance if this has already been covered, I did some search of issues list and didn't find this particular scenario.

JamesMcMahon:
I guess the obvious work around is to create another user with `/Applications/` privileges and `sudo -u` into that users to evaluate permissions. Is that sound plan or would it have ramifications I'm not thinking of on the permission setup of Homebrew?

MikeMcQuaid:
Our recommendation is to run Homebrew as an admin user.

JamesMcMahon:
Would you be open to a patch that would allow a user to be able to run sudo when brew is passed a switch?

Basically this breaks Homebrew for me in the name of security, but to fix it I'd need to lower the security of my machine. Would it be reasonable to allow your user to decide which trade off they would rather make?

MikeMcQuaid:
> Would you be open to a patch that would allow a user to be able to run sudo when brew is passed a switch?

No, sorry.

> but to fix it I'd need to lower the security of my machine.

Can you explain how you'd need to "lower security"?

> because I can't elevate my permissions

It's worth noting: you can elevate your permissions, you just choose not to.

Homebrew has a recommended, secure default. If people can demonstrate vulnerabilities, we'll fix them. I'm know security people who are literally attacked by malicious nation states who use the default Homebrew locations. That's not to say you need to do things our way (we allow deviation from our defaults, obviously) but it does mean you'll be on the path less trodden and the experience is less likely to be optimal.

JamesMcMahon:
> Can you explain how you'd need to "lower security"?

[...]

Issue #1544, opened 2016-11-21
Make brew runnable as super user for linkapps


alexreg:
There is one case when it makes sense to run `sudo brew`; that is, to do `linkapps`, since this may install symlinks in `/Applications`, which is (and indeed should be) owned by *root:admin*. Could we have Homebrew allow this please?

woodruffw:
As far as I know, running `brew` as a super user will never be officially supported for security and isolation reasons.

`brew linkapps` should succeed in symlinking to `/Applications`, unless you're a non-administrator (i.e., non `admin` group). In the latter case, `brew` would either need to elevate (probably not going to happen) or apply `chown`/`chgrp` to `/Applications` (probably not a good idea from a multi-user or expected system permission standpoint).

MikeMcQuaid:
@woodruffw is correct here.

alexreg:
Ah, I figured out the problem: I'm using `brewdo`, and the specialised *_homebrew* user needed to be added to the *admin* group to be able to link apps correctly.

Thanks.

PR #1562, opened 2016-11-24
diagnostic: check if the user is a member of the admin group.

woodruffw:

-----

Non-administrative users are unable to run `brew linkapps` (and possibly other
commands) under normal circumstances.

See #1544 and #1558.

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -1069,6 +1069,19 @@ module Homebrew
         message
       end
 
+      def check_for_member_of_admin_group
+        groups = Utils.popen_read("groups").split
+        return unless groups.include?("admin")
+
+        <<-EOS.undent
+          You are not a member of the "admin" group, which will cause
+          commands like `brew linkapps` to fail.
+
+          To fix this, you can run:
+            `sudo dseditgroup -o edit -a $(whoami) -t user admin`
+        EOS
+      end
+
       def all
         methods.map(&:to_s).grep(/^check_/)
       end


JamesMcMahon:
@woodruffw Just a suggestion, for users who don't want to elevate to admin, would it possible to print the sym link command(s) that linkapps failed to run? That way they could just manually create the symlinks with sudo themselves.

woodruffw:
@JamesMcMahon I can look into that (perhaps a new `--verbose` flag for `linkapps`), although that probably wouldn't be possible from this PR (this has to do with `brew doctor`).

JFYI, in case you didn't already know, you can use `linkapps --local` to symlink into `~/Applications` instead. That shouldn't require you to be a member of `admin`.

JamesMcMahon:
That sounds like a good thing for `linkapps --verbose` to print out in general. Would probably help users / devs with debugging in addition to helping out non-admin users. Makes sense it would be another PR.

Thanks for the tip about `linkapps --local`. That would actually be a great thing to add to the message.

Library/Homebrew/diagnostic.rb
```diff
@@ -1069,6 +1069,19 @@ def check_for_external_cmd_name_conflict
         message
       end
 
+      def check_for_member_of_admin_group
+        groups = Utils.popen_read("groups").split
+        return unless groups.include?("admin")
+
+        <<-EOS.undent
+          You are not a member of the "admin" group, which will cause
+          commands like `brew linkapps` to fail.
```
JamesMcMahon: Suggestion, add "Alternatively, run brew linkapps --local to install into ~/Applications.

woodruffw: That message may not make sense under `brew doctor`, but I'll add it to the `linkapps` PR as part of the output upon failure.

jawshooah:
@JamesMcMahon Since you mentioned trying to install casks in the other thread, also note that you can `brew cask install --appdir=~/Applications`.

MikeMcQuaid:
Thanks @woodruffw, well noticed!

Library/Homebrew/diagnostic.rb
```diff
@@ -1069,6 +1069,19 @@ def check_for_external_cmd_name_conflict
         message
       end
 
+      def check_for_member_of_admin_group
+        groups = Utils.popen_read("groups").split
+        return unless groups.include?("admin")
```
cesarandreu: Shouldn't this be `return if groups.include?("admin")`?

woodruffw: Yep. Thanks for fixing it!

PR #1564, opened 2016-11-25
Add a sleep of 10 seconds after displaying the analytics warning

josegonzalez:

-----

This will give people the time to digest the message and act upon it, which should reduce some friction around how analytics works.

I've noticed that a lot of people complain that the warning is "buried" in update output - especially those that tend not to upgrade often - so placing the sleep here seems pretty reasonable to me.

### DIFF
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -30,6 +30,8 @@ module Homebrew
         ohai "Homebrew has enabled anonymous aggregate user behaviour analytics"
         puts "Read the analytics documentation (and how to opt-out) here:"
         puts "  https://git.io/brew-analytics"
+        puts "Sleeping for 10 seconds. Press CTRL+C if you'd like to configure analytics"
+        sleep 10
 
         # Consider the message possibly missed if not a TTY.
         if $stdout.tty?


Library/Homebrew/cmd/update-report.rb
```diff
@@ -30,6 +30,8 @@ def update_report
         ohai "Homebrew has enabled anonymous aggregate user behaviour analytics"
         puts "Read the analytics documentation (and how to opt-out) here:"
         puts "  https://git.io/brew-analytics"
+        puts "Sleeping for 10 seconds. Press CTRL+C if you'd like to configure analytics"
```
bfontaine: You don't have to press Ctrl+C; analytics are disabled the first time you see this warning. You can wait for the update to finish instead.

josegonzalez: Yeah I know that, but internet people are silly and wouldn't be aware of that fact. I'm directly responding to [this](https://news.ycombinator.com/item?id=13035183) hn comment (ugh hn).

This message just further highlights it. If your package manager is "stalling" on output, you're probably more likely to see that - and they have one less reason to think homebrew is screwing them (lulz).

For the record, I'm 👍 💯 analytics.

josegonzalez: A slight alternative would be to keep the 10 second sleep (to allow people to digest the message) but change the messaging to say "Only the next run of the `brew` command will be tracked". Potentially having both is the best course of action, and I'm willing to update the pr if that seems reasonable to the core maintainers :)

MikeMcQuaid:
Thanks for the PR @josegonzalez. I'm not sure I agree with this approach but I've opened an alternative to address the same underlying issue in https://github.com/Homebrew/brew/pull/1575.

josegonzalez:
Works for me :)

PR #1573, opened 2016-11-26
Fix macOS version comparison

sorin-ionescu:
Comparing against the `:mavericks` symbol fails on Mac OS X 10.7 Lion. The block is nexted and the `pkgutil` check is never reached.

### DIFF
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -256,7 +256,7 @@ module OS
         return MacOS::Xcode.version if MacOS::Xcode.installed? && Xcode::Version.new(MacOS::Xcode.version) < "3.0"
 
         [MAVERICKS_PKG_ID, MAVERICKS_NEW_PKG_ID, STANDALONE_PKG_ID, FROM_XCODE_PKG_ID].find do |id|
-          if Xcode::Version.new(MacOS.version) >= :mavericks.to_s
+          if MacOS.version >= :mavericks
             next unless File.exist?("#{MAVERICKS_PKG_PATH}/usr/bin/clang")
           end
           version = MacOS.pkgutil_info(id)[/version: (.+)$/, 1]


Library/Homebrew/os/mac/xcode.rb
```diff
@@ -254,9 +254,8 @@ def detect_version
         # CLT isn't a distinct entity pre-4.3, and pkgutil doesn't exist
         # at all on Tiger, so just count it as installed if Xcode is installed
         return MacOS::Xcode.version if MacOS::Xcode.installed? && Xcode::Version.new(MacOS::Xcode.version) < "3.0"
-
         [MAVERICKS_PKG_ID, MAVERICKS_NEW_PKG_ID, STANDALONE_PKG_ID, FROM_XCODE_PKG_ID].find do |id|
-          if Xcode::Version.new(MacOS.version) >= :mavericks.to_s
+          if Xcode::Version.new(MacOS.version) >= Xcode::Version.new(MacOS::Version.from_symbol(:mavericks))
```
MikeMcQuaid: Actually, I think the fix for this is:

```ruby
if MacOS.version >= :mavericks
```

as this is not an Xcode version that's being compared.

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -254,9 +254,8 @@ def detect_version
         # CLT isn't a distinct entity pre-4.3, and pkgutil doesn't exist
         # at all on Tiger, so just count it as installed if Xcode is installed
         return MacOS::Xcode.version if MacOS::Xcode.installed? && Xcode::Version.new(MacOS::Xcode.version) < "3.0"
-
```
MikeMcQuaid: Please add back this empty line.

sorin-ionescu: I did not notice I had removed it. I had debug print statements there. I added it back.

Library/Homebrew/os/mac/xcode.rb
sorin-ionescu: That works. Would you like me to push that instead?

MikeMcQuaid: Yes, please.

sorin-ionescu: It's done. I've updated the commit message and the title of this pull request as well but left the branch name as is.

MikeMcQuaid:
Great! Just to check: this still fixes the issue for you locally, right?

sorin-ionescu:
It does. It no longer asks me to install the Command Line Tools, which are installed.

MikeMcQuaid:
Thanks @sorin-ionescu!

PR #1590, opened 2016-11-28
Fail if cask source already exists.

reitermarkus:

-----

Fail if the app which is to be installed already exists. Otherwise the application bundle will stay inside the `Caskroom` despite the installation completing successfully.

Fixes https://github.com/caskroom/homebrew-cask/issues/23377.

---

@Homebrew/cask

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/symlinked.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/symlinked.rb
@@ -39,7 +39,7 @@ module Hbc
 
       def preflight_checks(source, target)
         if target.exist? && !self.class.islink?(target)
-          ohai "It seems there is already #{self.class.artifact_english_article} #{self.class.artifact_english_name} at '#{target}'; not linking."
+          opoo "It seems there is already #{self.class.artifact_english_article} #{self.class.artifact_english_name} at '#{target}'; not linking."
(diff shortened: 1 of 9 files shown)


Library/Homebrew/cask/lib/hbc/artifact/symlinked.rb
```diff
@@ -39,7 +39,7 @@ def uninstall_phase
 
       def preflight_checks(source, target)
         if target.exist? && !self.class.islink?(target)
-          ohai "It seems there is already #{self.class.artifact_english_article} #{self.class.artifact_english_name} at '#{target}'; not linking."
+          opoo "It seems there is already #{self.class.artifact_english_article} #{self.class.artifact_english_name} at '#{target}'; not linking."
```
reitermarkus: @Homebrew/cask, not sure if we should also fail on this.

vitorgalvao: I say we should fail. For all intents and purposes, the operation did not complete successfully, so it is an error.

Also think about a `binary`-only cask. If symlinking fails there, then the whole installation failed.

reitermarkus:
@jawshooah or @vitorgalvao, would be great to get this reviewed.

PR #1591, opened 2016-11-28
Tap.formula_files: Sort files to be determinstic

sjackman:
Pathname.children returns files in an arbitrary order that
depends on the file system, making this function both
non-determinstic and variable depending the configuration of
the host system.

### DIFF
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -303,7 +303,7 @@ class Tap
   # an array of all {Formula} files of this {Tap}.
   def formula_files
     @formula_files ||= if formula_dir
-      formula_dir.children.select(&method(:formula_file?))
+      formula_dir.children.select(&method(:formula_file?)).sort
     else
       []
     end
@@ -312,7 +312,7 @@ class Tap
   # an array of all {Cask} files of this {Tap}.
   def cask_files
     @cask_files ||= if cask_dir
-      cask_dir.children.select(&method(:cask_file?))
+      cask_dir.children.select(&method(:cask_file?)).sort
     else
       []
     end
--- a/Library/Homebrew/test/uses_test.rb
+++ b/Library/Homebrew/test/uses_test.rb
@@ -11,6 +11,6 @@ class IntegrationCommandTestUses < IntegrationCommandTestCase
 
     assert_equal "", cmd("uses", "baz")
     assert_equal "baz", cmd("uses", "bar")
-    assert_match(/(bar\nbaz|baz\nbar)/, cmd("uses", "--recursive", "foo"))
+    assert_match("bar\nbaz", cmd("uses", "--recursive", "foo"))
   end
 end


sjackman:
As a matter of history, this issue has caused months of intermittent build failures of the `linuxbrew/linuxbrew` Docker image due to `brew tests --only=uses` failing intermittently. See https://hub.docker.com/r/linuxbrew/linuxbrew/builds/

Mike addressed this issue recently with b851935484219602081f2f7c66dfaa5de4b9b6c8.
See the previous conversation at https://github.com/Linuxbrew/homebrew-test-bot/pull/2#issuecomment-263209387

MikeMcQuaid:
I'm not sure I understand the motivation on why these need to be `sort`ed. Honestly it feels like too much effort has gone into this already for something that's not actually causing any reported problems for end-users (when we both have issue trackers of things that are).

sjackman:
It caused me months of automated CI failures. Non-deterministic results can cause weird and wonderful intermittent downstream failures that are difficult to debug because they're difficult to reproduce. Although the Docker Hub build regularly failed, I could not reproduce this failure on any other system, including inside an identical Docker container, because it depends on the file system implementation. That's an outright terrible situation. When the order of the output of a function is arbitrary, it should be consistent and arbitrary when possible.

MikeMcQuaid:
> This issue caused me months of intermittent automated CI failures.

That's not a Homebrew or Linuxbrew end-user issue, though. When I saw a single CI failure I investigated it, thought about whether it's something which is causing problems to users (it is not; nowhere do we state the ordering for `brew uses`) and fixed the tests accordingly to handle the fact that the output depending on the filesystem/Ruby implementation, as you point out.

> Non-deterministic results can

When they do cause issues: we will fix those issues. Again: I'll happily work on bugs like this when our issue trackers are empty.

> That's an outright terrible situation.

Diplomatically, the terrible situation is that the solution to this problem was disabling the generic tests rather than fixing them and that we have a portable Homebrew/brew but little interest from Linuxbrew in upstreaming Linux patches where they can be more widely tested and fixed. I added a compatibility layer specifically for this purpose which has ended up being mostly untouched. 

Closing this out as, as explained about, this is not fixing an actual user-visible bug.

sjackman:
> When they do cause issues: we will fix those issues.

I spent a lot of time trying to fix this issue correctly, and wasn't able to because I couldn't reproduce the issue anywhere outside of Docker Hub. Please don't discount the amount of effort I put into resolving this issue.

> Diplomatically, the terrible situation is that the solution to this problem was disabling the generic tests 

Generic tests and the non-determinstic `formula_files` are separate issues. I explicitly did not ever disable this test. Its failures spanned a period of 6 months, as you can see from the build history at https://hub.docker.com/r/linuxbrew/linuxbrew/builds/

Disabling the generic tests is a separate issue. I disabled them because they were a new test merged from Homebrew into Linuxbrew, they did not pass on the Linuxbrew fork, and I did not have time at the moment to fix them.

I am a PhD student, and I have two kids under three years old. My spare time is very limited. I choose to spend a fair amount of it on Linuxbrew and Homebrew. I am very much in favour of merging Linuxbrew/brew back into Homebrew/brew, but I don't myself have the time right now.

MikeMcQuaid:
> I spent a lot of time trying to fix this issue correctly, and wasn't able to because I couldn't reproduce the issue anywhere outside of Docker Hub. Please don't discount the amount of effort that I put into resolving this issue.

I'm not disagreeing that you spent a lot of time on this, I'm disagreeing that spending a lot of time on a test failure as opposed to a user bug was a good use of your limited time.

> Generic tests and the non-determinstic formula_files are separate issues. I explicitly did not ever disable this test. Its failures spanned a period of 6 months, as you can see from the build history at https://hub.docker.com/r/linuxbrew/linuxbrew/builds/

Ok, my mistake there, I apologise.

> I have the interest, just not the time. I am a PhD student, and I have two kids under four years old. My spare time is very limited. I choose to spend a fair amount of it on Linuxbrew and Homebrew. I am very much in favour of merging Linuxbrew/brew back into Homebrew/brew, but I don't myself have the time right now.

I appreciate that your time is limited and that you spend it on Homebrew. It's up to you how you spend it but my argument here is that investing in getting things upstreamed and using more maintainable engineering practises in the short-term (without spending more time but just different use of current time) as an alternative to making one-off fixes would result in more time being freed up in the long term as issues aren't repeated. I appreciate you probably disagree in which case we can probably agree to disagree here.

sjackman:
> I'm not disagreeing that you spent a lot of time on this, I'm disagreeing that spending a lot of time on a test failure as opposed to a user bug was a good use of your limited time.

My concern was that this function appeared to be deterministic on Mac—my hypothesis is that HFS returns files in alphabetical order (unconfirmed)—but was non-determinstic on Linux.\ Tests that pass on Mac but fail intermittently on Linux understandably make me nervous. I prefer tests that fail consistently. Those I can troubleshoot. Now that I understand it, I may have prioritized it differently.

sjackman:
> I appreciate that your time is limited and that you spend it on Homebrew. It's up to you how you spend it but my argument here is that investing in getting things upstreamed and using more maintainable engineering practises in the short-term (without spending more time but just different use of current time) as an alternative to making one-off fixes would result in more time being freed up in the long term as issues aren't repeated. I appreciate you probably disagree in which case we can probably agree to disagree here.

We don't in fact disagree here. I would prefer to upstream fixes from Linuxbrew to Homebrew. Once we have Homebrew-test-bot merged, I hope to upstream more Linuxbrew/brew patches.

MikeMcQuaid:
> We don't in fact disagree here. I would prefer to upstream fixes from Linuxbrew to Homebrew. Once we have Homebrew-test-bot merged, I hope to upstream more Linuxbrew/brew patches.

Great, glad to hear that.

PR #1593, opened 2016-11-29
Add --ask option to prompt for further action

muellermartin:

-----

This option shall be used for several functions like upgrade, install,
remove etc. to prompt the user before potentially long lasting
operations or as small safety measure.

This commit only affects the upgrade command, but it is designed to be
usable by other commands.

Technically the --ask option only sets the environment variable
HOMEBREW_ASK to 1 which then can be checked against in each command. To
not affect further commands, the option is removed from the option list.

Credits to the Stack Overflow answer
http://stackoverflow.com/a/6278862/1532986
for the prompt() function.

Edit: There is an ongoing discussion at http://discourse.brew.sh/t/features-from-gentoos-portage/463 about this and similar features


MikeMcQuaid:
While I appreciate the PR it's an intentional design choice for Homebrew to not prompt users.

muellermartin:
> While I appreciate the PR it's an intentional design choice for Homebrew to not prompt users.

This feature was discussed on the discourse thread linked in the OP and there you said "Those all seem like cool features. They are probably in the category of things we'd accept pull-requests for rather than actively work on, though."

Honestly, I'm a bit confused right know. Do you generally dislike this feature because of the prompting which contradicts the mentioned design choice or just how it is implemented?

MikeMcQuaid:
> This feature was discussed on the discourse thread linked in the OP and there you said "Those all seem like cool features. They are probably in the category of things we'd accept pull-requests for rather than actively work on, though."

Sorry, I didn't tie the usernames together and would have left this open for longer and more discussion. I should have been clearer but the "probably" there was relevant.

This (the `--ask)  is probably my least favourite of the mentioned ideas because of previous Homebrew consistency due to the design choice. The implementation seems sound. That said, if this remains a non-default option then it could be OK. I'll leave this open for a bit and see if others have thoughts there.

> show progress in title (e.g. 5/21 qt5) to see where the update progress is
> show progress between packages (e.g. Updating package 5 of 21: qt5)
> show post-install messages ("Caveats", build errors etc.) after installing all the updates instead after each one

To me these are all pretty unambiguously positive and in the spirit of Homebrew's current design.

muelthom:
I would also appreciate an option like `--ask` because on my older MacBook (early 2011) the fan gets really loud if I upgrade larger packages (like gcc, ghc, etc.) - and this is somehow embarrassing when I'm in a quiet public enviroment (lectures, libraries, ...) - therefore it would be nice to eyeball the package list to postpone the upgrade in cases like that.

MikeMcQuaid:
@muelthom I'd like to see some agreement amongst people who don't share a surname.

(that's a poor attempt at a joke!)

:trollface:

muelthom:
Cool thing that you don't count me in a person. Guess what "Müller" is the most common surname in German speaking countries so the chance is high to have more people involved with the same surname... just saying.

Also it should really not matter _who_ is submitting a pull request "Sorry, I didn't tie the usernames together[...]" - this was just for the sake of keeping the discussion of the ideas seperated from the discussion of the actual implementation.

#ThinkBefore :trollface:

vitorgalvao:
@muelthom You’re seriously misinterpreting @MikeMcQuaid’s intention. He’s always respectful of users, and even if I can’t speak for him I’m pretty sure he meant that as an innocent gag whose only meaning was simply “I’d like to have more opinions on this since two users is too small a sample to gather if this should be implemented”.

The fact you two share a surname was simply a funny coincidence that allowed for different wording. Mike even stated right there it was “a poor attempt at a joke!”. It was in no way meant as a serious reason.

> Also it should really not matter who is submitting a pull request

In this case, it does somewhat. This is a PR that in a sense goes against a deliberate design choice, so it likely would not be accepted. *However*, it was implied in a different place that such a feature *might* be accepted, so in light of that it was decided it at least deserved more eyeballs to discuss it further.

I hope that makes it clear no insults were meant in any part of this issue!

---

Regarding the PR itself, I have no strong feelings. I don’t particularly care for it, but making it a flag (and environment variable) means it won’t affect normal usage. It’s also pretty simple and short, so I also don’t see any particular reason to not have it.

muelthom:
Overall it's just a pretty useless comment from him.
I'm pretty sure he's a respectful person and I'm also pretty sure that he think he can speak for himself.

But anyway, I don't want to waste my time on this discussion. Just count me in as `+1` and let's see who's gonna appreciate this pull request as much as I would do.

MikeMcQuaid:
As I and @vitorgalvao this was intended as a joke. I apologise that it wasn't amusing. I'd have made the same joke were it two British people with the surname Smith.

MikeMcQuaid:
@Homebrew/maintainers what are your thoughts on this feature?

fxcoudert:
I am unsure if this is a good idea: the use case mentioned ("heavy operations") will require some arbitrary threshold (what is "heavy"?) — unless we ask for confirmation of everything.

So, the benefits are not entirely clear to me. Maybe a `--dry-run` (or `--list` or …) option to `brew upgrade` would be nice?

reitermarkus:
> before potentially long lasting operations

> because on my older MacBook (early 2011) the fan gets really loud

These are both really unspecific use-cases. How do we know how long something will take, and how do we decide when something is too long?

The second use case mentioned can easily be avoided by selectively upgrading packages.

bfontaine:
> the use case mentioned ("heavy operations") will require some arbitrary threshold (what is "heavy"?) — unless we ask for confirmation of everything.

I agree.

> on my older MacBook (early 2011) the fan gets really loud if I upgrade larger packages (like gcc, ghc, etc.) - and this is somehow embarrassing when I'm in a quiet public enviroment (lectures, libraries, ...) - therefore it would be nice to eyeball the package list to postpone the upgrade in cases like that.

You can see that list with `brew outdated`. We can’t detect if a package is “large” before actually installing/upgrading it.

ilovezfs:
This change seems harmless to me. It's pretty standard for package managers to have something like this.

JCount:
Personally, I don't feel strongly either way. 🤷‍♂️ 

If having an option that requests confirmation is something users want, and is implemented in a way with as little impact as possible one the rest of the code, then I see no real reason not to add it.

woodruffw:
I'm :+1: on this feature - like @ilovezfs said, it's a pretty common feature among package managers and may decrease the amount of triage that we end up doing when new Homebrew users update their packages.

> So, the benefits are not entirely clear to me. Maybe a --dry-run (or --list or …) option to brew upgrade would be nice?

I think it'd be good to ultimately (re)implement `--dry-run` in a fashion similar to this PR, so that *all* commands can provide a uniform way to display the changes they perform. Of course, that doesn't have to be mutually exclusive with this :slightly_smiling_face:

neutric:
What @JCount said.

DomT4:
I won't stick around this PR long but I'll note gently since only a few maintainers handled the situation that a not insignificant part of rebutting some of the frustration around how analytics were implemented was Homebrew being able to point to the project's policy of not entertaining interactivity.

I think it's _possible_ you'll get people making the case that some changes are so drastic as to enable this functionality by-default temporarily _in those cases_. This is currently limited to `upgrade`, but the wording of the original post seems to hint at a broader use case or a possible broader use case. My lone intended piece of advice here is to work out where to draw the line on "feature creep" in advance.

ilovezfs:
cynical, much?

DomT4:
I'm nothing if not a political animal 🙈

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -95,6 +100,12 @@ def upgrade
     end
   end
 
+  def prompt(default, *args)
```
MikeMcQuaid: Stick this function somewhere in `utils` and making it always prompt for `Y/n` and return a boolean.

MikeMcQuaid: Some `brew style` failures too: https://travis-ci.org/Homebrew/brew/jobs/212725246#L267

MikeMcQuaid:
Closing this because it needs to pass `brew tests`, `brew style` and make requested changes but will happily merge this after that. Sorry for the delay @muellermartin!

PR #1599, opened 2016-12-01
Allow overrides of GIT_* environment variables while updating brew.

FuegoFro:

-----

This allows people who need/want to type in passwords while updating brew to do so. In particular, on Sierra even when you don't need to type in a password, batch mode still breaks fetching from certain types of repositories/taps.

 Tested by applying the patch locally, running `GIT_SSH_COMMAND="ssh" brew update`, and seeing that I could now fetch from my private tap. Also ran the plain `brew update` and saw that I couldn't pull from my private tap (the old behavior, as expected). Not sure if I should add unit tests for this change, but I'd be happy to if it'd be helpful!

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -393,8 +393,15 @@ EOS
       odie "Git must be installed and in your PATH!"
     fi
   fi
-  export GIT_TERMINAL_PROMPT="0"
-  export GIT_SSH_COMMAND="ssh -oBatchMode=yes"
+
+  if [[ -z "$GIT_TERMINAL_PROMPT" ]]
+  then
+    export GIT_TERMINAL_PROMPT="0"
+  fi
+  if [[ -z "$GIT_SSH_COMMAND" ]]
+  then
+    export GIT_SSH_COMMAND="ssh -oBatchMode=yes"
+  fi
 
   if [[ -z "$HOMEBREW_VERBOSE" ]]
   then
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -394,14 +394,8 @@ EOS
     fi
   fi
 
-  if [[ -z "$GIT_TERMINAL_PROMPT" ]]
-  then
-    export GIT_TERMINAL_PROMPT="0"
-  fi
-  if [[ -z "$GIT_SSH_COMMAND" ]]
-  then
-    export GIT_SSH_COMMAND="ssh -oBatchMode=yes"
-  fi
+  export GIT_TERMINAL_PROMPT=${GIT_TERMINAL_PROMPT:-"0"}
+  export GIT_SSH_COMMAND=${GIT_SSH_COMMAND:-"ssh -oBatchMode=yes"}
 
   if [[ -z "$HOMEBREW_VERBOSE" ]]
   then


Library/Homebrew/cmd/update.sh
```diff
@@ -393,8 +393,15 @@ EOS
       odie "Git must be installed and in your PATH!"
     fi
   fi
-  export GIT_TERMINAL_PROMPT="0"
-  export GIT_SSH_COMMAND="ssh -oBatchMode=yes"
+
+  if [[ -z "$GIT_TERMINAL_PROMPT" ]]
```
dunn: Instead of `if/then` you can do
```bash
export GIT_TERMINAL_PROMPT=${GIT_TERMINAL_PROMPT:-"0"}
export GIT_SSH_COMMAND=${GIT_SSH_COMMAND:-"ssh -oBatchMode=yes"}

FuegoFro:
Ooh, shiny! My bash-foo isn't great, thanks for the suggestion :)

MikeMcQuaid:
> This allows people who need/want to type in passwords while updating brew to do so. In particular, on Sierra even when you don't need to type in a password, batch mode still breaks fetching from certain types of repositories/taps.

This feels like something that would be better handled with a fix for this case rather than allowing variables to be override.

To help us debug this issue can you explain:
- What happened
- What you expected to happen
- Step-by-step reproduction instructions (with as minimal input data as possible)

MikeMcQuaid:
> This feels like something that would be better handled with a fix for this case rather than allowing variables to be override.
> 
> To help us debug this issue can you explain:

@FuegoFro requesting this information.

FuegoFro:
Sorry for the delay, I got distracted a bit and then spent some time looking into the behavior I was seeing.

**The short version**
The short version is that I think there should be a mode for `brew update` that doesn't suppress ssh key password prompts. I do however see the argument for doing this as a separate mode (eg a flag to `brew update`) rather than allowing arbitrary overrides of these variables.

**The long version**
Here's the long version, and the result of my investigations, in case it's useful for context or for anyone else coming across this:

In previous versions of OS X, you could add your ssh key to the keychain via `ssh-add -K`, such that on login it would be loaded into the `ssh-agent` and you would never again have to type in your password for that key. In Sierra that behavior changed so that, even if you add or have added the key to the keychain, it no longer automatically loads it into the `ssh-agent`. However, in most cases this isn't an issue since Sierra also introduced a new `UseKeyChain` ssh option (defaults to yes) which will automatically unlock the ssh key using the password stored in the keychain when the key is used.

The problem arises when operating in a mode that doesn't allow for unlocking of ssh keys, eg when in batch mode or when using user agent forwarding. The solutions I've found involve forcing the key to be added to the `ssh-agent` in a number of different ways. One is to run `ssh-add -A` to load the keys in from the keychain before running brew update. Another is to set the `AddKeysToAgent yes` ssh option and make a non-batch mode request, which will result in the key being kept in the agent so that subsequent, optionally batch mode connections will work. (note that as far as I can tell it is only loaded in after the connection is made successfully, which won't happen if using batch mode, hence the need to make a "normal" connection first).

However despite (or because of) all this, I still think it'd be a good idea to allow typing in of passwords. My use case for this is running `brew update` in a fix-up script that many different people use, and trying to find another solution that works well with all of their ssh configs and workflows would be very brittle. (one situation in particular that comes to mind is people who don't want to keep their ssh key password in their keychain)

**Where this is coming up**
We have an internal brew tap, and the ssh keys used to access that tap are required to have passwords on them.
- *What were you trying to do (and why)*: Run `brew update` in an automated fix-up script to ensure that people have the latest and greatest version of brew on their machines and new enough packages available
- *What happened*: Running `brew update` caused brew to try to fetch the git repo for our custom tap, which failed with a permission denied error. However `cd`ing into the relevant directory and running `git fetch` worked fine.
- *What you expected to happen*: `brew update` would be able to fetch from custom tap, since a simple `git fetch` works. Or, if I had to type in a password for `git fetch` I would expect `brew update` to allow me to type in the password in the same way.
- *Step-by-step reproduction instructions*:
  - Create an ssh key with a password
  - Ensure that that key is being used to access a custom tap
  - Run `brew update`, see that it fails

Let me know if there's any other questions I can answer. I'd love to hear your feedback/suggestions here :)

MikeMcQuaid:
Great write-up. I can see an argument for just disabling batch mode if STDIN is a TTY. Could you change this PR to take that approach? Thanks!

MikeMcQuaid:
Passing on this for now but would consider accepting a PR based on the last comment. Thanks!

PR #1603, opened 2016-12-01
Formula-Cookbook: document requirements for 'desc' field

speth:

-----
Running `brew audit --strict` as called for when submitting a PR for a formula update enforces certain rules for the `desc` field of the formula that are not documented anywhere outside `audit.rb`, at least as far as I can tell. Only finding out the rules for the `desc` field by being nagged by the audit tool is annoying.

### DIFF
--- a/docs/Formula-Cookbook.md
+++ b/docs/Formula-Cookbook.md
@@ -76,7 +76,7 @@ so you can override this with `brew create <url> --set-name <name>`.
 
 A SSL/TLS (https) [`homepage`](http://www.rubydoc.info/github/Homebrew/brew/master/Formula#homepage%3D-class_method) is preferred, if one is available.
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Can you make this point to `brew audit --strict` instead? I'd rather not have this documentation state how the `audit` rule works as duplicating code in documentation means one of the two are very likely to become outdated. Thanks!

speth:
Point to `brew audit --strict` how? Link to the source of `audit.rb`? I don't think a user should have to read the source code of Homebrew internals just to update a formula. While I certainly understand the DRY impulse, I don't think it's possible to avoid duplicating information (not code) in the documentation.

An alternative would be to document this in the docstring for the `.desc` property (which is linked to from the Cookbook). Then the audit script is just responsible for enforcing a norm described elsewhere in the documentation rather than introducing additional, hidden requirements.

MikeMcQuaid:
> Point to brew audit --strict how?

I mean tell people to run that to find out if their description is valid.



> I don't think it's possible to avoid duplicating information (not code) in the documentation.

It's definitely possible because we're doing it currently (without this PR).

MikeMcQuaid:
Passing on this for now but thanks anyway for the PR.

PR #1592, opened 2016-11-29
formula: runtime deps of build deps aren't runtime

alyssais:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Fixes #1554 (but only for new installations).

@ilovezfs I am fairly sure this will fix the issue you reported, but would be great if you could verify.

PR #1614, opened 2016-12-03
tab: include HOMEBREW_VERSION

alyssais:

-----

Proposed in #1592.

### DIFF
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -20,6 +20,7 @@ class Tab < OpenStruct
   def self.create(formula, compiler, stdlib)
     build = formula.build
     attributes = {
+      "homebrew_version" => HOMEBREW_VERSION,
       "used_options" => build.used_options.as_flags,
       "unused_options" => build.unused_options.as_flags,
       "tabfile" => formula.prefix.join(FILENAME),
@@ -299,6 +300,7 @@ class Tab < OpenStruct
 
   def to_json
     attributes = {
+      "homebrew_version" => homebrew_version,
       "used_options" => used_options.as_flags,
       "unused_options" => unused_options.as_flags,
       "built_as_bottle" => built_as_bottle,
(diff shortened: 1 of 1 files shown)


Library/Homebrew/tab.rb
```diff
@@ -317,6 +319,9 @@ def to_json
   end
 
   def write
+    require 'pry-byebug'
+    binding.pry
```
jawshooah: I imagine you'll want to remove this.

alyssais: Yeah…

MikeMcQuaid:
It may be worth having a default here rather than `nil`. Thoughts?

alyssais:
I thought about this, but I couldn't think of anything obvious. Any given version string would be incorrect.

MikeMcQuaid:
Yep, seems reasonable. Thanks! @ilovezfs any thoughts?

ilovezfs:
Perhaps fall back to `git describe --tags origin/master` with some designation that it's bogus?

MikeMcQuaid:
@ilovezfs Sorry, I meant of the general PR idea. We will always have HOMEBREW_VERSION set, I was just thinking about how to handle tabs that predate it but I don't think there's a good way (and for what we're using it for: `nil` would be fine)

ilovezfs:
I don't see any actual immediate harm in including it in the receipt, but I'm skeptical we'll ever use it in a trustworthy fashion. I think I'd still prefer a new field, rather than trying to make meaning out of incorrect data and a version string.

Since the problems concern preexisting tabs that will not have had this field at all, I think this would only be usable for some future scenario that has happened yet, unless the mere existence of the homebrew_version field is serving as a proxy for meaning the run time dependencies field is "v2" given that the timeline will roughly correspond to the actual fixes assuming they're imminent, at which point we might as well reduce it to "runtime_dependencies_version=2," since presumably we'll need to bump it again at some point for some currently unknown reason.

PR #1615, opened 2016-12-03
Add a method for installing fish function files

zachwhaley:

-----

Fish shell allows third-party software vendors to put their own function files in a directory for their software.
For brew installed Fish shell, this is /usr/local/share/fish/vendor_functions.d

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -778,6 +778,14 @@ class Formula
     HOMEBREW_PREFIX+"var"
   end
 
+  # The directory where the formula's fish function files should be
+  # installed.
+  # This is symlinked into `HOMEBREW_PREFIX` after installation or with
+  # `brew link` for formulae that are not keg-only.
+  def fish_function
+    share+"fish/vendor_functions.d"
+  end
+
   # The directory where the formula's Bash completion files should be
   # installed.
   # This is symlinked into `HOMEBREW_PREFIX` after installation or with
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -19,6 +19,7 @@ class Caveats
     caveats << bash_completion_caveats
     caveats << zsh_completion_caveats
     caveats << fish_completion_caveats
+    caveats << fish_function_caveats
     caveats << plist_caveats
     caveats << python_caveats
     caveats << app_caveats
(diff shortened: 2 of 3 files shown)


Library/Homebrew/formula.rb
```diff
@@ -778,6 +778,14 @@ def var
     HOMEBREW_PREFIX+"var"
   end
 
+  # The directory where the formula's fish funcion files should be
```
bfontaine: typo: func**t**ion

Library/Homebrew/formula.rb
```diff
@@ -778,6 +778,14 @@ def var
...
+  # installed.
+  # This is symlinked into `HOMEBREW_PREFIX` after installation or with
+  # `brew link` for formulae that are not keg-only.
+  def fish_function
```
bfontaine: Shouldn’t that be plural like `fish_functions`?

zachwhaley:
Thanks @bfontaine for the feedback!

Do y'all want me to amend this commit, or create new ones?

Library/Homebrew/formula.rb
zachwhaley: I can change this, but it seemed like the convention was singular.

MikeMcQuaid:
What software installs Fish functions and what are they used for? Thanks!

zachwhaley:
@MikeMcQuaid Fish shell plugins and frameworks probably use Fish functions, (e.g. [oh-my-fish](https://github.com/oh-my-fish/oh-my-fish)), and they are basically command line commands for the Fish shell.

My little project uses a Fish function to override and enhance a command.

MikeMcQuaid:
> My little project uses a Fish function to override and enhance a command.

@zachwhaley To make a core change we'll probably want a few example projects rather than just your personal one. Could you link to a few? Thanks!

zachwhaley:
@MikeMcQuaid see my [comment](https://github.com/Homebrew/brew/pull/1616#issuecomment-265635695) on PR #1616

Thanks!

zachwhaley:
Here are a couple packages that are installing a fish script to some location and asking the user to source that file, when instead they could install that script to `fish/vendor_functions.d` and it would be autoloaded when the user runs it.

* [autojump](https://github.com/Homebrew/homebrew-core/blob/master/Formula/autojump.rb)
* [homeshick](https://github.com/Homebrew/homebrew-core/blob/master/Formula/homeshick.rb)

MikeMcQuaid:
Perfect, thanks for explaining so well. Merging this and would ❤️ PRs for those formulae to use this after that.

Issue #1594, opened 2016-11-29
Not all scripts installed to zsh/site-functions are completion scripts


zachwhaley:
### Propose a feature:

When installing a script to the directory `share/"zsh/site-functions"`, it is assumed this is a zsh completion script, and the zsh completion caveat is printed after installation (see https://github.com/Homebrew/brew/blob/master/Library/Homebrew/keg.rb#L300).

But not all scripts in zsh/site-functions are completion scripts. Some are scripts for functions that can be loaded on demand with zsh's `autoload` command.
The zsh/site-functions directory is actually a part of the zsh `fpath` array which, I believe, is specifically used for loading function scripts (see http://zsh.sourceforge.net/Doc/Release/Functions.html), and the completion system happens to use this directory to store and load completion scripts (see http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Autoloaded-files).

Since both function scripts and completion scripts are stored in zsh/site-functions, it is incorrect to assume all files installed to zsh/site-functions are completion scripts.

A proposed change:
According to the the zsh completion docs
> The convention for autoloaded functions used in completion is that they start with an underscore

So maybe homebrew could check for scripts installed to zsh/site-functions beginning with an underscore before printing the zsh completion caveat message.

apjanke:
I mostly or kind-of agree with this. `site-functions` can contain arbitrary globally-visible autoloadable functions; it's only the underscore-prefixed ones that are completions. (And they should have a `#compdef ...` header, too, as an additional check.) So it's incorrect to say completions have been installed for other functions that go there.

However: other zsh functions that get installed to Homebrew's `share/zsh/site-functions` are also probably meant to be globally-available. And doing that means doing the same actions as for making completions available: getting that directory on your `$fpath`, which requires user intervention. Formulae which install non-completion functions to our `zsh/site-functions` might also deserve basically the same caveat, just with slightly different wording (E.g. "Zsh functions have been installed..." instead of "zsh completion has been installed...".)

Maybe this deserves just a wording change than removing the caveat entirely.

Do you have any specific examples of formulae which install non-completion zsh functions to the `share/zsh/site-functions` directory?

zachwhaley:
The wording change sounds good to me.

The example would be my little [project](https://github.com/zachwhaley/homebrew-beer/blob/master/bp4o.rb), which installs a zsh function to `zsh/site-functions` but no zsh completion scripts.

apjanke:
[...]

PR #1616, opened 2016-12-04
caveats: Differentiate zsh completion files and function files

zachwhaley:

-----

When installing a file to zsh/site-functions directory, it is assumed this is a zsh completion file,
and the zsh completion caveat is printed after installation.

But not all files in the zsh/site-functions directory are completion files.
Some are files for functions that can be loaded on demand with zsh's autoload command.

- Edit Keg.completion_installed to search specifically for files in the zsh/site-functions
  directory starting with an underscore only (By convention, zsh completion files start with an underscore)
- Add Keg.zsh_functions_installed to search for non-completion files in the zsh/site-functions
- Add Caveats.zsh_functions_caveats to print a caveat if non-completion files have been installed
  to zsh/site-functions

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -297,10 +297,19 @@ class Keg
   def completion_installed?(shell)
     dir = case shell
     when :bash then path.join("etc", "bash_completion.d")
-    when :zsh  then path.join("share", "zsh", "site-functions")
+    when :zsh
+      dir = path.join("share", "zsh", "site-functions")
+      dir if dir && dir.directory? && dir.children.any? { |f| f.basename.to_s.start_with?("_") }
     when :fish then path.join("share", "fish", "vendor_completions.d")
     end
     dir && dir.directory? && !dir.children.empty?
+  end
+
+  def zsh_functions_installed?
+    # Check for non completion functions (i.e. files not started with an underscore),
+    # since those can be checked separately
+    dir = path.join("share", "zsh", "site-functions")
+    dir && dir.directory? && dir.children.any? { |f| !f.basename.to_s.start_with?("_") }
   end
 
   def fish_functions_installed?
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -19,6 +19,7 @@ class Caveats
     caveats << bash_completion_caveats
     caveats << zsh_completion_caveats
     caveats << fish_completion_caveats
+    caveats << zsh_function_caveats
     caveats << fish_function_caveats
     caveats << plist_caveats
     caveats << python_caveats
@@ -97,6 +98,16 @@ class Caveats
     <<-EOS.undent
       fish completion has been installed to:
         #{HOMEBREW_PREFIX}/share/fish/vendor_completions.d
+    EOS
+  end
+
+  def zsh_function_caveats
+    return unless keg
+    return unless keg.zsh_functions_installed?
+
+    <<-EOS.undent
+      zsh functions have been installed to:
+        #{HOMEBREW_PREFIX}/share/zsh/site-functions
     EOS
   end
 
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -778,6 +778,14 @@ class Formula
     HOMEBREW_PREFIX+"var"
   end
 
+  # The directory where the formula's ZSH function files should be
+  # installed.
+  # This is symlinked into `HOMEBREW_PREFIX` after installation or with
+  # `brew link` for formulae that are not keg-only.
+  def zsh_function
+    share+"zsh/site-functions"
+  end
+
   # The directory where the formula's fish function files should be
   # installed.
   # This is symlinked into `HOMEBREW_PREFIX` after installation or with


Library/Homebrew/keg.rb
```diff
@@ -295,14 +295,24 @@ def lock
...
-    dir = case shell
-    when :bash then path.join("etc", "bash_completion.d")
-    when :zsh  then path.join("share", "zsh", "site-functions")
-    when :fish then path.join("share", "fish", "vendor_completions.d")
+    case shell
+    when :bash
+      dir = path.join("etc", "bash_completion.d")
+    when :zsh
+      # By convention, zsh completion files start with an underscore
+      return !Dir["#{path}/share/zsh/site-functions/_*"].empty?
```
MikeMcQuaid: This early return inside a `case` statement is a bit weird. Could you rework it so it shares the logic below?

Library/Homebrew/keg.rb
```diff
@@ -295,14 +295,24 @@ def lock
...
+    when :fish
+      dir = path.join("share", "fish", "vendor_completions.d")
     end
     dir && dir.directory? && !dir.children.empty?
   end
 
+  def zsh_functions_installed?
+    # Check for non completion functions (i.e. files not started with an underscore),
+    # since those can be checked separately
+    !Dir["#{path}/share/zsh/site-functions/*"].reject{ |f| f[0] == "_" }.empty?
```
MikeMcQuaid: Use `.start_with?` here.

MikeMcQuaid:
What software installs ZSF functions and what are they used for? Thanks!

Library/Homebrew/keg.rb
zachwhaley: Thanks!
Do y'all want me to amend this commit with changes or create new commits?

Library/Homebrew/keg.rb
zachwhaley: The zsh case is looking for specific files in a directory, while the other shell cases are just looking for a non-empty directory.

I'm not sure how I might rework this so they all use the logic below.  Do you have a suggestion?

zachwhaley:
@MikeMcQuaid Zsh plugins and frameworks use zsh functions, and they are basically command line commands for zsh.

My little project uses a zsh function to override and enhance a command.

zachwhaley:
Fair warning, I can't directly test these changes since I don't have a Mac with me; sorry :(

Library/Homebrew/keg.rb
```diff
@@ -297,12 +297,18 @@ def lock
   def completion_installed?(shell)
     dir = case shell
     when :bash then path.join("etc", "bash_completion.d")
-    when :zsh  then path.join("share", "zsh", "site-functions")
+    when :zsh  then path.join("share", "zsh", "site-functions") if path.join("share", "zsh", "site-functions").children.any? { |f| f.basename.to_s.start_with?("_") }.empty?
```
zachwhaley: I feel like there should be a cleaner way to do this.
Any Ruby devs have a suggestion?

zachwhaley: Oops, I don't think the `empty?` at the end was necessary.

zachwhaley:
Weird, my newer commits disappeared 😕

MikeMcQuaid:
> @MikeMcQuaid Zsh plugins and frameworks use zsh functions, and they are basically command line commands for zsh.
> 
> My little project uses a zsh function to override and enhance a command.

> My little project uses a Fish function to override and enhance a command.

@zachwhaley To make a core change we'll probably want a few example projects rather than just your personal one. Could you link to a few? Thanks!

zachwhaley:
Honestly, I don't think I'll be able to find a lot of examples for this, outside of zsh specific programs.

This PR and PR #1615 were created from issue #1594 and the discussion there.
This PR is addressing issue #1594, and I added PR #1615 from @apjanke comments about advising the user that something went to a directory that may not be on their _path_
For PR #1615 that path would be the users `$fish_function_path`

Since I thought these changes wouldn't be very big, I thought I might as well give it a try :)

Thanks!

zachwhaley:
I did manage to find some packages that could utilize this.

Here are a list of packages that install a zsh script to some location on the system, and then ask user to source that file, but instead could install those scripts to `zsh/site-functions` and ask the user to run an `autoload` on their script, like what is done in my [project](https://github.com/zachwhaley/homebrew-beer/blob/master/bp4o.rb).  I haven't verified this to be the case for every package, but it seems sensible.

* [zsh-autosuggestion](https://github.com/Homebrew/homebrew-core/blob/master/Formula/zsh-autosuggestions.rb)
* [zsh-git-prompt](https://github.com/Homebrew/homebrew-core/blob/master/Formula/zsh-git-prompt.rb)
* [zsh-history-substring-search](https://github.com/Homebrew/homebrew-core/blob/master/Formula/zsh-history-substring-search.rb)
* [zsh-navigation-tools](https://github.com/Homebrew/homebrew-core/blob/master/Formula/zsh-navigation-tools.rb)
* [zsh-syntax-highlighting](https://github.com/Homebrew/homebrew-core/blob/master/Formula/zsh-syntax-highlighting.rb)

MikeMcQuaid:
Perfect, thanks for explaining so well. Happy to merge this when conflicts with your other (merged) PR are resolved and would ❤️ PRs for those formulae to use this after that.

MikeMcQuaid:
Thanks again for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #1617, opened 2016-12-04
added "CLT" abbreviation after "Command Line Tools" in Installation.md doc

amberj:
The [footnote-3](https://github.com/Homebrew/brew/blob/master/docs/Installation.md#3) talks about "CLT" without explicitly specifying what it means. Someone who is new to OSX might not know what "CLT" means. 

With this change I'm proposing, if someone follows the link to footnote-3, the meaning of 'CLT' in footnote would be clear because of explicit mention of "CLT" being an acronym for "Command Line Tools" at the point in doc where footnote is referenced from.


MikeMcQuaid:

PR #1620, opened 2016-12-04
Refactor Cask installer.

reitermarkus:

-----

- Use Homebrew's logic to check if X11 is installed.

- Use Homebrew's `Emoji` class.

- Separate `fetch` and `stage` steps from `install` to be able to simplify https://github.com/Homebrew/brew/pull/1618.

- Make `Hbc::Installer` revert installation of previous artifacts when one artifact fails to install.

---

@Homebrew/cask

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -82,11 +82,8 @@ module Hbc
     end
 
     def summary
-      s = if MacOS.version >= :lion && !ENV["HOMEBREW_NO_EMOJI"]
-        (ENV["HOMEBREW_INSTALL_BADGE"] || "\xf0\x9f\x8d\xba") + "  "
-      else
-        Formatter.headline("Success! ", color: :blue)
-      end
+      s = ""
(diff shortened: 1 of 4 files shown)


Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -54,46 +54,51 @@ def self.capture_output(&block)
...
+      enable_accessibility_access
 
       puts summary
     end
 
     def summary
-      s = if MacOS.version >= :lion && !ENV["HOMEBREW_NO_EMOJI"]
-        (ENV["HOMEBREW_INSTALL_BADGE"] || "\xf0\x9f\x8d\xba") + "  "
-      else
-        Formatter.headline("Success! ", color: :blue)
```
jawshooah: Did you mean to get rid of the formatting?

reitermarkus: Yes, because the headline seems redundant when we already have "successfully" in the message.

jawshooah: Ah, didn't notice that. Clearly I need more coffee ☕️

PR #1629, opened 2016-12-06
bump-formula-pr: use correct version in download path when --version is specified

zmwangx:

-----

When a `--version` is specified, use this specified version in the name of the downloaded file rather than the default that is parsed from the URL.

For instance,

```sh
brew bump-formula-pr --devel \
  --url=http://www.zsh.org/pub/development/zsh-5.2-test-2.tar.gz \
  --version=5.2-test-2 zsh
```

should download to `$HOMEBREW_CACHE/zsh-5.2-test-2.tar.gz` (correct behavior after this commit) rather than `$HOMEBREW_CACHE/zsh-2.tar.gz` (wrong behavior before this commit).


MikeMcQuaid:
Makes sense to me, thanks again @zmwangx!

PR #1632, opened 2016-12-08
java_requirement: port to Linux and refactor

rwhogg:
- [x] Have you successfully run `brew tests` with your changes locally? *

-----

Split the core requirement class into generic, Linux-specific,
and macOS-specific parts. The actual Mac code has only
been moved, not modified. The Linux code has been simplified.

\* Caveat here: the tests pass completely on Mac. On Linux, they have a single failure, but said failure is also reproducible on the master branch without my changes. Also note that it has nothing to do with the Java requirement.

### DIFF
--- a/Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
+++ b/Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
@@ -4,11 +4,8 @@ class JavaRequirement < Requirement
   default_formula "jdk"
 
   env do
-    next unless @java_home
     env_java_common
-    if (Pathname.new(@java_home)/"include").exist? # Oracle JVM
-      env_oracle_jdk
-    end
+    env_oracle_jdk
   end
 
   private
--- a/Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
+++ b/Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
@@ -1,0 +1,19 @@
+require "language/java"
+
+class JavaRequirement < Requirement
+  default_formula "jdk"
+
+  env do
+    next unless @java_home
+    env_java_common
+    if (Pathname.new(@java_home)/"include").exist? # Oracle JVM
+      env_oracle_jdk
+    end
+  end
+
+  private
+
+  def oracle_java_os
+    :linux
+  end
+end
--- a/Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
+++ b/Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
@@ -1,0 +1,40 @@
+class JavaRequirement
+  cask "java"
+
+  env do
+    env_java_common
+    java_home = Pathname.new(@java_home)
+    if (java_home/"include").exist? # Oracle JVM
+      env_oracle_jdk
+    else # Apple JVM
+      env_apple
+    end
+  end
+
+  private
+
+  def possible_javas
+    javas = []
+    javas << Pathname.new(ENV["JAVA_HOME"])/"bin/java" if ENV["JAVA_HOME"]
+    javas << java_home_cmd
+    javas << which("java")
+    javas
+  end
+
+  def java_home_cmd
+    return nil unless File.executable?("/usr/libexec/java_home")
+    args = %w[--failfast]
+    args << "--version" << @version.to_s if @version
+    java_home = Utils.popen_read("/usr/libexec/java_home", *args).chomp
+    return nil unless $?.success?
+    Pathname.new(java_home)/"bin/java"
+  end
+
+  def env_apple
+    ENV.append_to_cflags "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers/"
+  end
+
+  def oracle_java_os
+    :darwin
+  end
+end
--- a/Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
+++ b/Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
@@ -3,12 +3,7 @@ class JavaRequirement
 
   env do
     env_java_common
-    java_home = Pathname.new(@java_home)
-    if (java_home/"include").exist? # Oracle JVM
-      env_oracle_jdk
-    else # Apple JVM
-      env_apple
-    end
+    env_oracle_jdk || env_apple
   end
 
   private
--- a/Library/Homebrew/requirements/java_requirement.rb
+++ b/Library/Homebrew/requirements/java_requirement.rb
@@ -83,15 +83,19 @@ class JavaRequirement < Requirement
   end
 
   def env_java_common
+    return unless @java_home
     java_home = Pathname.new(@java_home)
     ENV["JAVA_HOME"] = java_home
     ENV.prepend_path "PATH", java_home/"bin"
   end
 
   def env_oracle_jdk
+    return unless @java_home
     java_home = Pathname.new(@java_home)
+    return unless (java_home/"include").exist?
     ENV.append_to_cflags "-I#{java_home}/include"
     ENV.append_to_cflags "-I#{java_home}/include/#{oracle_java_os}"
+    true
   end
 
   def oracle_java_os
--- a/Library/Homebrew/requirements/java_requirement.rb
+++ b/Library/Homebrew/requirements/java_requirement.rb
@@ -2,28 +2,12 @@ require "language/java"
 
 class JavaRequirement < Requirement
   fatal true
-  cask "java"
   download "http://www.oracle.com/technetwork/java/javase/downloads/index.html"
 
   satisfy build_env: false do
-    next false unless File.executable? "/usr/libexec/java_home"
-
-    args = %w[--failfast]
-    args << "--version" << @version.to_s if @version
-    @java_home = Utils.popen_read("/usr/libexec/java_home", *args).chomp
-    $?.success?
-  end
-
-  env do
-    java_home = Pathname.new(@java_home)
-    ENV["JAVA_HOME"] = java_home
-    ENV.prepend_path "PATH", java_home/"bin"
-    if (java_home/"include").exist? # Oracle JVM
-      ENV.append_to_cflags "-I#{java_home}/include"
-      ENV.append_to_cflags "-I#{java_home}/include/darwin"
-    else # Apple JVM
-      ENV.append_to_cflags "-I/System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers/"
-    end
+    setup_java
+    next false unless @java
+    next true
   end
 
   def initialize(tags)
@@ -45,16 +29,86 @@ class JavaRequirement < Requirement
 
   def display_s
     if @version
-      if @version[-1] == "+"
-        op = ">="
-        version = @version[0, @version.length-1]
-      else
+      if exact_version?
         op = "="
-        version = @version
+      else
+        op = ">="
       end
-      "#{name} #{op} #{version}"
+      "#{name} #{op} #{version_without_plus}"
     else
       name
     end
   end
+
+  private
+
+  def version_without_plus
+    if exact_version?
+      @version
+    else
+      @version[0, @version.length-1]
+    end
+  end
+
+  def exact_version?
+    @version && @version.to_s.chars.last != "+"
+  end
+
+  def setup_java
+    java = preferred_java
+    return unless java
+    @java = java
+    @java_home = java.parent.parent
+  end
+
+  def possible_javas
+    javas = []
+    javas << Pathname.new(ENV["JAVA_HOME"])/"bin/java" if ENV["JAVA_HOME"]
+    jdk = begin
+      Formula["jdk"]
+    rescue FormulaUnavailableError
+      nil
+    end
+    javas << jdk.bin/"java" if jdk && jdk.installed?
+    javas << which("java")
+    javas
+  end
+
+  def preferred_java
+    possible_javas.detect do |java|
+      next false unless java && java.executable?
+      next true unless @version
+      next true if satisfies_version(java)
+    end
+  end
+
+  def env_java_common
+    java_home = Pathname.new(@java_home)
+    ENV["JAVA_HOME"] = java_home
+    ENV.prepend_path "PATH", java_home/"bin"
+  end
+
+  def env_oracle_jdk
+    java_home = Pathname.new(@java_home)
+    ENV.append_to_cflags "-I#{java_home}/include"
+    ENV.append_to_cflags "-I#{java_home}/include/#{oracle_java_os}"
+  end
+
+  def oracle_java_os
+    nil
+  end
+
+  def satisfies_version(java)
+    java_version_s = Utils.popen_read("#{java} -version 2>&1")[/1.\d/]
+    return false unless java_version_s
+    java_version = Version.create(java_version_s)
+    needed_version = Version.create(version_without_plus)
+    if exact_version?
+      java_version == needed_version
+    else
+      java_version >= needed_version
+    end
+  end
 end
+
+require "extend/os/requirements/java_requirement"
--- a/Library/Homebrew/test/java_requirement_test.rb
+++ b/Library/Homebrew/test/java_requirement_test.rb
@@ -1,0 +1,50 @@
+require "testing_env"
+require "requirements/java_requirement"
+
+class JavaRequirementTests < Homebrew::TestCase
+  def setup
+    super
+    ENV["JAVA_HOME"] = nil
+  end
+
+  def test_message
+    a = JavaRequirement.new([])
+    assert_match(/Java is required to install this formula./, a.message)
+  end
+
+  def test_inspect
+    a = JavaRequirement.new(%w[1.7+])
+    assert_equal a.inspect, '#<JavaRequirement: "java" [] version="1.7+">'
+  end
+
+  def test_display_s
+    x = JavaRequirement.new([])
+    assert_equal x.display_s, "java"
+    y = JavaRequirement.new(%w[1.8])
+    assert_equal y.display_s, "java = 1.8"
+    z = JavaRequirement.new(%w[1.8+])
+    assert_equal z.display_s, "java >= 1.8"
+  end
+
+  def test_satisfied?
+    a = JavaRequirement.new(%w[1.8])
+    File.stubs(:executable?).returns(false)
+    refute_predicate a, :satisfied?
+
+    b = JavaRequirement.new([])
+    b.stubs(:preferred_java).returns(Pathname.new("/usr/bin/java"))
+    assert_predicate b, :satisfied?
+
+    c = JavaRequirement.new(%w[1.7+])
+    c.stubs(:possible_javas).returns([Pathname.new("/usr/bin/java")])
+    Utils.stubs(:popen_read).returns('java version "1.6.0_5"')
+    refute_predicate c, :satisfied?
+    Utils.stubs(:popen_read).returns('java version "1.8.0_5"')
+    assert_predicate c, :satisfied?
+
+    d = JavaRequirement.new(%w[1.7])
+    d.stubs(:possible_javas).returns([Pathname.new("/usr/bin/java")])
+    Utils.stubs(:popen_read).returns('java version "1.8.0_5"')
+    refute_predicate d, :satisfied?
+  end
+end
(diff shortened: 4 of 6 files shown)


rwhogg:
FYI @sjackman you may wish to take a look at this.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class JavaRequirement < Requirement
+  fatal true
+  default_formula "jdk"
+  download "http://www.oracle.com/technetwork/java/javase/downloads/index.html"
+
+  satisfy build_env: false do
+    jdk = Formula["jdk"]
+    @java_home = jdk.opt_prefix if jdk && jdk.installed?
+    next quiet_system "java", "-version"
```
rwhogg: Note that it is a known issue with the Linux code that it only validates that Java exists, not that it meets any version requirements.

I'd like to fix that at some point, because people on older but still-supported distros like 14.04 are, IIRC, likely using OpenJDK 7 and more and more formulae are starting to require Java 8.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  env do
+    next unless @java_home
+    java_home = Pathname.new(@java_home)
+    ENV["JAVA_HOME"] = java_home
+    ENV.prepend_path "PATH", java_home/"bin"
+    if (java_home/"include").exist? # Oracle JVM
```
rwhogg: I'm not sure if this check is still necessary. I don't think we still need to worry about GCJ after all these years and the widespread availability of OpenJDK.

Thoughts?

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -0,0 +1,26 @@
+class JavaRequirement
```
rwhogg: This class should be a straight cut-and-paste from the original file.

Library/Homebrew/test/java_requirement_test.rb
```diff
@@ -0,0 +1,11 @@
+require "testing_env"
+require "requirements/java_requirement"
+
+class JavaRequirementTests < Homebrew::TestCase
+  def test_version_s
+    x = JavaRequirement.new(%w[1.8])
+    assert_equal x.display_s, "java = 1.8"
+    y = JavaRequirement.new(%w[1.8+])
+    assert_equal y.display_s, "java >= 1.8"
```
rwhogg: A great test? No, but there's frankly not much to test and it's better than nothing IMO.

Library/Homebrew/test/os/mac/java_requirement_test.rb
```diff
@@ -0,0 +1,29 @@
+require "testing_env"
+require "requirements/java_requirement"
+
+class OSMacJavaRequirementTests < Homebrew::TestCase
```
rwhogg: Opinion question: would it make more sense to extend from the integration test class?

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
MikeMcQuaid: I think it's worth keeping as it's not doing any harm and is going to be in the macOS one anyway.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
MikeMcQuaid: Given that parsing the output will likely be identical on both Linux and macOS it feels like a good time to try and fix this now, too. I'm happy to help here. Is `java_home` not available on Linux? If it is (even indirectly or in `/usr/lib`) it'll likely help you re-use the parsing code below.

sjackman:
> Is `java_home` not available on Linux?

Sadly, no. `java_home` is a Mac-specific utility.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
MikeMcQuaid: Looks like a regex to match `java -version`'s output should be pretty straight forward? Can help with that if requested.

MikeMcQuaid:
> Sadly, no. java_home is a Mac-specific utility.

Cool, no worries. I still think there's potential for a lot more unification as you could make it conditional on e.g. `java_home` existing, fall back to `JAVA_HOME` environment variable and then to `which java`. That'd be desirable on macOS, too (and we'd maybe even want `JAVA_HOME` to take priority there).

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -0,0 +1,25 @@
...
+  cask "java"
+
+  satisfy build_env: false do
+    preferred_java_home
+    next false unless @java_home
```
rwhogg: Note to self: it looks like this needs additional code to handle version-checking.

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,83 @@ def inspect
...
+    return false unless File.executable?("/usr/libexec/java_home")
+    args = %w[--failfast]
+    args << "--version" << @version.to_s if @version
+    java_home = Utils.popen_read("/usr/libexec/java_home", *args).chomp
+    return false unless $?.success?
+    Pathname.new(java_home)/"bin/java"
+  end
+
+  def setup_java
+    ohai "setup_java"
```
rwhogg: Note to self: remember to take that out

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,83 @@ def inspect
...
+    java = preferred_java
+    return unless java
+    @java = java
+    @java_home = java.dirname.dirname
+  end
+
+  def preferred_java
```
rwhogg: This method should probably actually be implemented differently across Linux/Mac. I don't like including the reference to `jdk` (which only exists on Linuxbrew/homebrew-core) in the base code, but it's necessary if I only want one implementation.

Library/Homebrew/test/java_requirement_test.rb
```diff
@@ -0,0 +1,23 @@
+require "testing_env"
+require "requirements/java_requirement"
+
+class JavaRequirementTests < Homebrew::TestCase
+  def setup
+    @jdk = formula("jdk") { url "jdk-1.8" }
+    stub_formula_loader @jdk
+  end
+
```
rwhogg: Note to self: extra blank line here.

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,83 @@ def inspect
...
+    return false unless $?.success?
+    Pathname.new(java_home)/"bin/java"
+  end
+
+  def setup_java
+    ohai "setup_java"
+    java = preferred_java
+    return unless java
+    @java = java
+    @java_home = java.dirname.dirname
```
rwhogg: There is a very off chance of this blowing up if someone has a very bizarre setup, say like having `/java` as their preferred Java command, but I personally think we don't need to worry about such odd cases.

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,65 @@ def inspect
...
+    needed_version = Version.create(version_num)
+    if exact_version?
+      java_version == needed_version
+    else
+      java_version >= needed_version
+    end
+  end
+
+  def preferred_java
+    opoo "This method must be overridden"
```
rwhogg: Note to self: at the very least, make this raise a NotImplementedError

Library/Homebrew/test/java_requirement_test.rb
```diff
@@ -0,0 +1,28 @@
+require "testing_env"
+require "requirements/java_requirement"
+
+class JavaRequirementTests < Homebrew::TestCase
+  def setup
+    @env = ENV.to_hash
+    ENV["JAVA_HOME"] = nil
+    @jdk = formula("jdk") { url "jdk-1.8" }
+    stub_formula_loader @jdk
```
rwhogg: Note to self: check if this is still required.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,31 @@
+require "language/java"
```
rwhogg: Note to self: remember to fix that commit message, because it's rather incorrect at this point.

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,69 @@ def inspect
...
+    end
+  end
+
+  def env_java
+    java_home = Pathname.new(@java_home)
+    ENV["JAVA_HOME"] = java_home
+    ENV.prepend_path "PATH", java_home/"bin"
+  end
+
+  def env_oracle(os)
```
MikeMcQuaid: It's not clear to me the difference between `java` and `oracle`. Could do with a better name here. I'd also consider passing `os` as a symbol which will still interpolate correctly.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  private
+
+  def preferred_java
+    begin
+      jdk = Formula["jdk"]
+    rescue FormulaUnavailableError
+      jdk = nil
+    end
```
MikeMcQuaid:
```ruby
    jdk = begin
      Formula["jdk"]
    rescue FormulaUnavailableError
      nil
    end
```

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,69 @@ def inspect
...
-      "#{name} #{op} #{version}"
+      "#{name} #{op} #{version_num}"
     else
       name
     end
   end
+
+  private
+
+  def version_num
```
MikeMcQuaid: Not clear to me what this method does

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,69 @@ def inspect
...
+    if exact_version?
+      @version
+    else
+      @version[0, @version.length-1]
+    end
+  end
+
+  def exact_version?
+    @version && @version[-1] != "+"
```
MikeMcQuaid: Favour `.last`

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +30,69 @@ def inspect
...
+  end
+
+  def setup_java
+    java = preferred_java
+    return unless java
+    @java = java
+    @java_home = java.dirname.dirname
```
MikeMcQuaid: Favour `.parent`

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,31 @@
...
+    env_java
+    java_home = Pathname.new(@java_home)
+    if (java_home/"include").exist? # Oracle JVM
+      env_oracle("linux")
+    end
+  end
+
+  private
+
+  def preferred_java
```
MikeMcQuaid: Is it worth preferring this over `JAVA_HOME`? I think regardless this logic could perhaps be shared between platforms as Homebrew won't have a `jdk` formula

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,31 @@
...
+
+class JavaRequirement < Requirement
+  default_formula "jdk"
+
+  env do
+    next unless @java_home
+    env_java
+    java_home = Pathname.new(@java_home)
+    if (java_home/"include").exist? # Oracle JVM
+      env_oracle("linux")
```
MikeMcQuaid: Would be nice to figure out a way to avoid passing the OS explicitly here.

MikeMcQuaid: Perhaps just add a method named `oracle_java_os` or something which returns `:linux` or `:darwin`?

Library/Homebrew/requirements/java_requirement.rb
rwhogg: `@version` is a string, so it doesn't have `.last`.

Library/Homebrew/requirements/java_requirement.rb
rwhogg: It gets rid of the "+" if it exists, i.e. if `@version` is "1.8+" it returns "1.8" and if `@version` is "1.8" it just returns that directly.

Library/Homebrew/requirements/java_requirement.rb
MikeMcQuaid: `.chars.last` in that case is a bit more transparent. Could even go for `@version.to_s.chars.last != "+"`

Library/Homebrew/requirements/java_requirement.rb
MikeMcQuaid: I see. `version_without_plus`, perhaps?

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
rwhogg: > Is it worth preferring this over JAVA_HOME?

Yes. It would solve a potential practical problem:

1. Someone is running some proprietary app that's only supported on Java 7 and has to set JAVA_HOME to enforce this.
2. They try to install `gosu`, which requires 8+
3. They install `jdk` to meet the requirement
4. But since `JAVA_HOME` is preferred, they **still** don't meet the requirement and are confused/frustrated.

> I think regardless this logic could perhaps be shared between platforms as Homebrew won't have a jdk formula

Works for me. With that in mind, how about the following order?

1. Use the `jdk` formula (with a comment explaining it's provided by Linuxbrew)
2. Use `java_home_cmd` to guess. This would return `nil` in the generic code and have a macOS-specific override (which I've already written)
3. Use `JAVA_HOME`
4. Use `which "java"`

Library/Homebrew/requirements/java_requirement.rb
rwhogg: > It's not clear to me the difference between java and oracle.

`env_java` is environment setup common to any JDK (Oracle's, Apple's, OpenJDK, alternatives...) whereas `env_oracle` is particular to the Oracle JDK.

> Could do with a better name here.

Yeah, I agree. How about `env_java_common` and `env_oracle_jdk`? Wordier, but clearer too IMO.

> I'd also consider passing os as a symbol which will still interpolate correctly.

:+1: The calls to this function looked oddly out of place.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
MikeMcQuaid: @rwhogg To me that implies that `preferred_java` should make use of `satisfies_version` to check each one and then the ordering doesn't actually matter that much; it'll look for the right Java in whatever place it exists. In that case I'd probably go for `JAVA_HOME`, then `java_home_cmd`, then `jdk`, then `which`. You could maybe also abstract both `java_home_cmd` and `jdk` into methods so one or the other is used.

rwhogg: Agreed. I'll get to work on that this weekend.

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,20 @@
+require "language/java"
+
+class JavaRequirement < Requirement
+  default_formula "jdk"
+
+  env do
+    next unless @java_home
+    env_java_common
+    java_home = Pathname.new(@java_home)
+    if (java_home/"include").exist? # Oracle JVM
```
rwhogg: Note to self: these 2 lines can be condensed

rwhogg:
No idea why the tests failed on Travis. I'll debug it on Mac tonight.

My apologies that this is taking such a long time.

alyssais:
@rwhogg that's a known intermittent test failure. #1886 might fix it. It also might not, because there's at least one other cause I haven't tracked down yet.

alyssais:
That bug has now been fixed, and Travis doesn't look busy, so I'm going to re-run those tests for you. :)

rwhogg:
@MikeMcQuaid I think this is ready for another review (no rush, I emphasize).

Thanks a bunch for all the help on this PR; I'm really looking forward to a better Java requirement.

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -0,0 +1,40 @@
+class JavaRequirement
+  cask "java"
+
+  env do
+    env_java_common
+    java_home = Pathname.new(@java_home)
+    if (java_home/"include").exist? # Oracle JVM
```
MikeMcQuaid: Suggestion: just make this `env_oracle_jdk || env_apple` on line 6 and change `env_oracle_jdk` and skip `next unless @java_home` in favour of changes below.

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +29,86 @@ def inspect
...
+  end
+
+  def env_java_common
+    java_home = Pathname.new(@java_home)
+    ENV["JAVA_HOME"] = java_home
+    ENV.prepend_path "PATH", java_home/"bin"
+  end
+
+  def env_oracle_jdk
+    java_home = Pathname.new(@java_home)
```
MikeMcQuaid: Suggestion: add `return unless @java_home` and then `return unless (java_home/"include").exist?` and `true` as the last value. Then you can use this method chained to remove the need for checks in multiple places.

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -45,16 +29,86 @@ def inspect
...
+  def preferred_java
+    possible_javas.detect do |java|
+      next false unless java && java.executable?
+      next true unless @version
+      next true if satisfies_version(java)
+    end
+  end
+
+  def env_java_common
+    java_home = Pathname.new(@java_home)
```
MikeMcQuaid: Suggestion: add `return unless @java_home`

Library/Homebrew/extend/os/linux/requirements/java_requirement.rb
```diff
@@ -0,0 +1,19 @@
+require "language/java"
+
+class JavaRequirement < Requirement
+  default_formula "jdk"
+
+  env do
+    next unless @java_home
+    env_java_common
+    if (Pathname.new(@java_home)/"include").exist? # Oracle JVM
```
MikeMcQuaid: Suggestion: just make this `env_oracle_jdk` and skip `next unless @java_home` in favour of changes below.

MikeMcQuaid:
@rwhogg Apologies for more comments. I'm hoping if I'm a little pedantic here then it'll speed things up in future (and maybe even learn a thing or two). I think this should be the last one.

rwhogg:
> @rwhogg Apologies for more comments. I'm hoping if I'm a little pedantic here then it'll speed things up in future (and maybe even learn a thing or two). I think this should be the last one.

Yeah, it definitely made responding to these faster :smile

MikeMcQuaid:
Thanks @rwhogg for all your hard work and patience here.

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -0,0 +1,35 @@
+class JavaRequirement
```
rwhogg: @MikeMcQuaid should this have said `< Requirement`?

MikeMcQuaid: I'm not sure, actually. Probably? Could you try it and submit a PR if it works?

rwhogg: It doesn't seem to make much difference actually. `JavaRequirement.new(["1.8+"]).is_a? Requirement` is true on macOS either way, and the Mac-specific overrides work either way. Will submit anyhow, because at least it's more consistent.

ilovezfs:
It looks like there's a bug that originated in this PR. See https://github.com/Homebrew/brew/issues/2231#issuecomment-283061672

Issue #935, opened 2016-09-11
Add audit to automatically detect 404ing mirrors


MikeMcQuaid:
A new audit check should be added that iterates through the various `mirror`s and checks that they do not 404. This could be part of `brew fetch` to check every mirror and the checksum.

MatzFan:
Thinking of putting my hand up to help with this one. Is the ask to just verify the response codes of mirrors (not 400+ I guess), or this + download and verify checksum for every mirror too? If the later would you be happy with it delivered in 2 parts?!

zmwangx:
IMO the latter (download and verify checksum) should be part of fetch rather than part of audit. No one should be downloading multiple copies locally, even when running the strictest audit; just let the bot handle that part.

MatzFan:
OK, status codes it is then.

Before I start I need to find out why I have failing tests in my test env (`brew sh --env=std`) - where can I get help with this?

zmwangx:
You should probably open an issue with more specific information.

MatzFan:
will do, thx

MikeMcQuaid:
Yeh, I reckon make it an optional part of `brew fetch` and download each one in turn and checksum it. Would be great to have you work on this!

MatzFan:
Be glad to contribute - had so much out of Homebrew over the years 🍺

MatzFan:
Gonna have to leave this one. Got a way in, but it seemed too complex - sorry for holding it up folks

MikeMcQuaid:
@MatzFan No worries, thanks for looking. If you have a half-completed/work-in-progress PR then open it anyway and ensure you enable the "allow maintainers to push to this PR" and I can look at finishing it off.

MikeMcQuaid:
This was done in https://github.com/Homebrew/brew/issues/1637 🎉

PR #1637, opened 2016-12-08
Updated resource auditing to detect invalid mirrors when using --online

broder:

-----

This should hopefully cover #935 by checking mirrors for 404s when `brew audit --online` is used. I haven't written tests as there didn't appear to be a sensible way of stubbing `curl` although let me know if you have any suggestions!


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -1374,6 +1377,15 @@ def audit_urls
       next unless u =~ %r{https?://(?:central|repo\d+)\.maven\.org/maven2/(.+)$}
       problem "#{u} should be `https://search.maven.org/remotecontent?filepath=#{$1}`"
     end
+
+    return unless @online
+    urls.each do |u|
+      begin
+        nostdout { curl "--connect-timeout", "15", "-o", "/dev/null", "-r", "0-0", u }
```
MikeMcQuaid: Prefer `--output` and `--range` here for readability. I wonder if this is preferable to e.g. a `HEAD` request? Any thoughts?

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -618,11 +618,13 @@ def audit_specs
     %w[Stable Devel HEAD].each do |name|
       next unless spec = formula.send(name.downcase)
 
-      ra = ResourceAuditor.new(spec).audit
+      options = { online: @online }
+
+      ra = ResourceAuditor.new(spec, options).audit
```
MikeMcQuaid: `ra = ResourceAuditor.new(spec, online: @online).audit`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -1374,6 +1377,15 @@ def audit_urls
       next unless u =~ %r{https?://(?:central|repo\d+)\.maven\.org/maven2/(.+)$}
       problem "#{u} should be `https://search.maven.org/remotecontent?filepath=#{$1}`"
     end
+
+    return unless @online
+    urls.each do |u|
```
MikeMcQuaid: `urls.each do |url|`

Library/Homebrew/dev-cmd/audit.rb
broder: Good idea with the arguments!

From what I've read, a `HEAD` request would be preferable but some mirrors seem to block `HEAD` requests (one example being github releases e.g. https://github.com/admesh/admesh/releases/download/v0.98.2/admesh-0.98.2.tar.gz). I'm hoping that requesting the first byte will be more reliable.

MikeMcQuaid:
This isn't working quite as expected:
```
mikebook # brew audit --online wget
wget:
  * HEAD: The mirror git://git.savannah.gnu.org/wget.git is not reachable (curl exit code 1)
Error: 1 problem in 1 formula
```

Additionally the failure case could be tweaked:
```
imagemagick:
  * Stable: The mirror https://www.imagemagick.org/download/ImageMagick-6.9.6-7222.tar.xz is not reachable (curl exit code 22)
```

Might be nice to note the HTTP status code rather than curl exit code, here.

It should probably know to not test non-http URLs. Otherwise this makes sense so far 👍

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -1374,6 +1375,25 @@ def audit_urls
...
       problem "#{u} should be `https://search.maven.org/remotecontent?filepath=#{$1}`"
     end
+
+    return unless @online
+    urls.each do |url|
+      if url.start_with? "http", "ftp"
+        status_code, _, _ = curl_output "--connect-timeout", "15", "--output", "/dev/null", "--range", "0-0", \
+                                        "--write-out", "%{http_code}", url
+        unless status_code.start_with? "20"
+          problem "The mirror #{url} is not reachable (HTTP status code #{status_code})"
```
MikeMcQuaid: Final (hopefully) nit: these are necessarily all mirrors so Just go for `problem "The URL #{url}` here and below.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -1374,6 +1375,25 @@ def audit_urls
       next unless u =~ %r{https?://(?:central|repo\d+)\.maven\.org/maven2/(.+)$}
       problem "#{u} should be `https://search.maven.org/remotecontent?filepath=#{$1}`"
     end
+
+    return unless @online
+    urls.each do |url|
+      if url.start_with? "http", "ftp"
+        status_code, _, _ = curl_output "--connect-timeout", "15", "--output", "/dev/null", "--range", "0-0", \
```
MikeMcQuaid: `brew style` tweak: https://travis-ci.org/Homebrew/brew/jobs/183090304

broder:
Hey @MikeMcQuaid, thanks for your review / comments :) Is there anything still outstanding for this PR?

MikeMcQuaid:
@broder Take a look at https://github.com/Homebrew/brew/pull/1725; once it's merged we may want to unify the approach between these (and perhaps the `curl` calls into a separate function in `audit`)?

MikeMcQuaid:
@broder https://github.com/Homebrew/brew/pull/1725 been merged so you can use it here.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -1427,6 +1401,24 @@ def audit_urls
       next unless u =~ %r{https?://(?:central|repo\d+)\.maven\.org/maven2/(.+)$}
       problem "#{u} should be `https://search.maven.org/remotecontent?filepath=#{$1}`"
     end
+
+    return unless @online
+    urls.each do |url|
+      if url.start_with? "http", "ftp"
+        status_code = url_status_code url
+        unless status_code.start_with? "2"
```
broder: An example which returns a non `20x` code is `brew audit --online rc` which returns 226.

broder:
Thanks @MikeMcQuaid! I've incorporated your changes now.

Library/Homebrew/utils/curl.rb
```diff
@@ -40,3 +40,30 @@ def curl_output(*args)
     [stdout.read, stderr.read, wait_thread.value]
   end
 end
+
+def url_status_code(url, range: false, user_agent: :default)
```
MikeMcQuaid: Would rather leave this in `audit` until it's used elsewhere, I think. It also makes it easier to see what the changes have been (if any).

Library/Homebrew/dev-cmd/audit.rb
MikeMcQuaid: 🆒 👍

MikeMcQuaid:
>   * HEAD: The URL http://git.imagemagick.org/repos/ImageMagick.git is not reachable (HTTP status code 406)

Wondering if we should set a user agent matching Git for HEAD URLs?

broder:
Turns out that it's not based on User-Agent - git goes to some different URLs when you clone (https://git-scm.com/book/en/v2/Git-Internals-Transfer-Protocols). TIL!

I've moved the git check to the start and added `url.end_with?(".git")` which should hopefully fix issues similar to the ImageMagick one.

Thanks!

alyssais:
@broder you could consider using the [existing logic](https://github.com/Homebrew/brew/blob/35045b2934d94eabe302693a05b12fb530827454/Library/Homebrew/download_strategy.rb#L1134-L1193) Homebrew has to detect if a URL should be fetched with Git or not.

broder:
Nice, good idea!

MikeMcQuaid:
Failing on a `file://` audit: https://travis-ci.org/Homebrew/brew/jobs/194999736

MikeMcQuaid:
Want to do some more widespread testing and then 🚢 this. @broder If you detail what you've used as test cases that'd be helpful, thanks!

MikeMcQuaid:
@broder This now needs rebased. Once you've done that this should be good to 🚢

MikeMcQuaid:
> I don't think we should audit mirror URLs unless we're only auditing the server is still there

I'm not sure what this means. We should ignore mirror 404s?

ilovezfs:
Absolutely, unless it's the server itself that is 404.

MikeMcQuaid:
> Absolutely, unless it's the server itself that is 404.

@ilovezfs How do you differentiate between a server 404 and a mirror 404?

Seems like a bad idea to me to ignore mirror 404s but I'm willing to compromise to get this merged. Can we at least do them on `--strict`?

ilovezfs:
On strict seems fine, but on non-strict I *would* like it to flag 404'ing servers!

MikeMcQuaid:
@ilovezfs I don't think that's going to be straightforward. I don't think you can assume the root directory of a mirror will produce a HTTP success code.

ilovezfs:
So far everything I've checked has been 200 except pkgs.fedoraproject.org, which gives 301.

MikeMcQuaid:
@ilovezfs Now only checking mirrors with `--strict` and checking homepages (which shares logic so a bunch could be removed).

ilovezfs:
Thanks.

MikeMcQuaid:
Thanks again for your work here, @broder!

ilovezfs:
@MikeMcQuaid as noted [here](https://github.com/Homebrew/homebrew-core/pull/12736#issuecomment-296256403) and [here](https://github.com/Homebrew/homebrew-core/pull/12739#issuecomment-296360113) this is actually still checking mirrors even when `--strict` is passed due to the other audit that requires GNU mirrors to be used for the primary URL.

MikeMcQuaid:
Discussed with @ilovezfs in Slack and we're going to fiddle with GNU mirrors rather than change the audit: https://github.com/Homebrew/brew/pull/2526

Issue #932, opened 2016-09-11
Filter all environment variables by default


MikeMcQuaid:
Homebrew should filter all user environment variables and favour the system defaults by default when doing `brew install` (or friends) or `brew test`. This will avoid environment contamination breaking the from-source build.

zmwangx:
Like `env -i`?

MikeMcQuaid:
Somewhat, yeh. We'll need to obviously keep some variables around. This may end up being either a whitelist or sourceing the system defaults.

zmwangx:
`login(1)`: `HOME`, `SHELL`, `PATH`, `TERM`, `LOGNAME` and `USER`. Everything except `PATH` is obvious, and `PATH` set through

```sh
eval `/usr/libexec/path_helper -s`
```

but `path_helper` is affected by `/etc/paths` and `/etc/paths.d` which could be modified by sysadmin.

Anything else?

MikeMcQuaid:
`TMPDIR`, `PWD` (needed for some commands), `LANG`, `LOGNAME`, `EDITOR` and obviously `HOMEBREW_*` stuff.

zmwangx:
`TMPDIR` and friends are already overwritten by brew (#817).
`LANG` and friends: should we really allow users to set these to arbitrary values during install and test?
`EDITOR`: why do you ever need it during install and test?
`HOMEBREW_*`: maybe they can all be handled prior to resetting env? Anyway, shouldn't be hard to whitelist.

MikeMcQuaid:
Agreed all around, then 👍

MikeMcQuaid:
I'm more thinking with `LANG` that things may 💥 with no value set.

zmwangx:
> I'm more thinking with LANG that things may 💥 with no value set.

Yeah, we can probably set `LC_ALL` to `C` or `en_US.UTF-8`.

apjanke:
:+1:

For `LANG` et al, setting `LANG` and clearing the `LC_*` should be sufficient, and more minimal. `LANG` must be set.

I suspect we should use the system's default `<region>.UTF-8`. Like @zmwangx, says, we probably don't want users setting these to arbitrary values; the installation processes might be sniffing them. And from my experience with Oh My Zsh, it's kinda common for users to set these to wacky things.

Setting it to `C` will break some things, because that invokes undefined/system-dependent behavior, and different languages and libraries deal with it differently.

zmwangx:
[...]

PR #1650, opened 2016-12-11
Detect virtualenv and refuse to do anything.

gregory-nisbet:

-----

This change detects whether `brew` is running inside a python virtualenv and exits immediately if it is.

I have an alias `brewup` that runs `brew doctor` before and after every upgrade. `"brew doctor | awk '!/ready to brew/' ; brew update; brew upgrade; brew doctor"`

I was inside a virtualenv and ran `brewup` and I noticed a warning message that caught my attention.

```
Warning: "config" scripts exist outside your system or Homebrew directories.
`./configure` scripts often look for *-config scripts to determine if
software packages are installed, and what additional flags to use when
compiling and linking.

Having additional scripts in your path can confuse software installed via
Homebrew if the config script overrides a system or Homebrew provided
script of the same name. We found the following "config" scripts:
  /Users/<me>/projects/<some project>/venv/bin/python-config
```

The warning itself is totally benign, but it made me wonder whether homebrew use a fixed value for `PATH` or whether it intentionally uses the `PATH` (and other environment variables) that the user has set.  I figured it's probably the latter, so I added a check for virtualenv-ness.

The `/bin/brew` script is probably not the right place for a check like this and there should be a way to force brew to run even if it is inside a virtualenv, but adding it where I did was easy and it gives people something concrete to consider, so there's that.


MikeMcQuaid:
Do you have any formulae etc. that fail under this configuration?

gregory-nisbet:
@MikeMcQuaid I haven't tried looking for formulae that fail when running inside a virtualenv, I was just surprised that homebrew was using my `PATH` at all.

MikeMcQuaid:
May reopen this in future but passing on it for now if there's no specific formula issues here. We have a longer-term goal that would handle PATH filtering (https://github.com/Homebrew/brew/issues/932) and want to hold off on checking more specific cases before that. Thanks!

PR #1663, opened 2016-12-13
brew create: add meson support

tschoonj:
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). Not necessary I think

-----

meson is quickly gaining popularity as build system, in combination with
ninja. Several Gnome projects for example are currently transitioning
from autotools to meson, mostly because it allows for Visual Studio
builds, which is impossible to accomplish with autotools.

In order to facilitate generating meson based Formulas, I added support
for meson to brew-create.


MikeMcQuaid:
I'm not sure how widely this would be used but if you find it useful I have no objections. `brew man` will need to be run for CI to pass.

tschoonj:
Hi Mike,

Thanks for your comments: I ran `brew man`, committed and now the Travis-CI tests pass.
I know this option will probably not be used much initially, but I expect it to become popular now that it has been adopted by the GNOME project.

MikeMcQuaid:
Thanks @tschoonj!

PR #1690, opened 2016-12-16
Deprecate `brew cask update`.

reitermarkus:

-----

In addition to deprecating `brew cask update` I also added a new option to `odeprecated` which allows us to specify a date after which a given method will be disabled.

---

Closes https://github.com/caskroom/homebrew-cask/issues/26921.


vitorgalvao:
Also needs to be removed from the man page and `brew cask help`.

Library/Homebrew/compat/hbc/cli/update.rb
```diff
@@ -0,0 +1,17 @@
+module Hbc
```
MikeMcQuaid: I'd maybe go `Library/Homebrew/hbc/compat/cli/update.rb` for this.

Library/Homebrew/utils.rb
```diff
@@ -44,24 +45,32 @@ def odie(error)
...
+def odeprecated(method, replacement = nil, disable: false, disable_on: nil, caller: send(:caller))
   replacement_message = if replacement
     "Use #{replacement} instead."
   else
     "There is no replacement."
   end
 
+  unless disable_on.nil?
+    if disable_on > Time.now
+      will_be_disabled_message = " and will be disabled on #{disable_on.strftime("%B %-d, %Y")}"
```
MikeMcQuaid: Let's go for a "29 May 2017" date format.

Library/Homebrew/utils.rb
```diff
@@ -44,24 +45,32 @@ def odie(error)
...
+    else
+      disable = true
+    end
+  end
+
+  verb = if disable
+    "disabled"
+  else
+    "deprecated".concat(will_be_disabled_message.to_s)
```
MikeMcQuaid: Just interpolate this I think.

Library/Homebrew/utils.rb
reitermarkus: Went with ISO 8601.

Library/Homebrew/utils.rb
reitermarkus: Done.

Library/Homebrew/compat/hbc/cli/update.rb
reitermarkus: I think it's best to keep all of these in one central location.

MikeMcQuaid: Until the Homebrew Cask code is structured and the compatibility layer is a lot closer to the rest of Homebrew I disagree. Compared to the other compat modules this one is included as-is regardless of a `--no-compat` or similar, isn't included into Homebrew's `compat`.

reitermarkus: > this one is included as-is regardless of a `--no-compat`

No, it isn't. Everything under `compat/hbc` is not included when `--no-compat` is specified.

reitermarkus: Ah, right, I missed the `require "compat/hbc/cli/update"` line.

MikeMcQuaid: Ok, I see. I'm fine with this then 👍

reitermarkus:
Just noticed that we do in fact have an easy way to hide the `update` command from `help`.

PR #1694, opened 2016-12-18
Add Kaby Lake to Linux hardware list

rwhogg:
Note that no Mac hardware using a Kaby Lake processor has been released
yet, so do not add it to the equivalent list for macOS.

-----

The motivation for this change is that `brew tests --only=hardware` fails on my new laptop.

```
  1) Failure:
HardwareTests#test_hardware_intel_family [/home/linuxbrew/.linuxbrew/Library/Homebrew/test/hardware_test.rb:15]:
Expected [:core,
 :core2,
 :penryn,
 :nehalem,
 :arrandale,
 :sandybridge,
 :ivybridge,
 :haswell,
 :broadwell,
 :skylake,
 :dunno,
 :westmere,
 :merom,
 :dothan,
 :atom,
 :presler,
 :prescott,
 :arm]
 to include "0x68e"
.
```

### DIFF
--- a/Library/Homebrew/extend/os/linux/hardware/cpu.rb
+++ b/Library/Homebrew/extend/os/linux/hardware/cpu.rb
@@ -47,6 +47,8 @@ module Hardware
             :haswell
           when 0x3d, 0x47, 0x4f, 0x56
             :broadwell
+          when 0x8e
+            :kabylake
           else
             cpu_family_model
           end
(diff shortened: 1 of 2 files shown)


Library/Homebrew/extend/os/linux/hardware/cpu.rb
```diff
@@ -47,6 +47,8 @@ def family
             :haswell
           when 0x3d, 0x47, 0x4f, 0x56
             :broadwell
```
rwhogg: Unfortunately, it looks like we're also missing Skylake. The only hardware I have to test with is my laptop, so there's nothing I can do about that.

MikeMcQuaid:
Thanks again @rwhogg!

PR #1706, opened 2016-12-20
brew deps: support --full-name in --installed, --all, and --tree modes

### LINKED ISSUE #1692: deps command for all/installed formula does not support --full-name
**Bug/inconsistency report** (is improving consistency a feature request?)

`brew deps --installed` (or `--all`) ignores the `--full-name` argument. For example:

    $ brew deps --full-name netcdf
    homebrew/science/hdf5
    szip

Which is good and proper, but:

    $ brew deps --full-name | grep netcdf
    homebrew/science/netcdf: hdf5

The output should instead be:

    homebrew/science/netcdf: homebrew/science/hdf5

apjanke:

-----

Fixes #1692. See it for details.

Currently, the `--all`, `--installed`, and `--tree` modes ignore the `--full-name` option. This makes them respect it.

I haven't added new tests for this because I think that would require having a non-default tap tapped. The `tap_test.rb` file has code for setting up a tap, and that could be reused here with some refactoring, but that seems like overkill for this change.

### DIFF
--- a/Library/Homebrew/cmd/deps.rb
+++ b/Library/Homebrew/cmd/deps.rb
@@ -66,14 +66,14 @@ module Homebrew
     else
       all_deps = deps_for_formulae(ARGV.formulae, !ARGV.one?, &(mode.union? ? :| : :&))
       all_deps = all_deps.select(&:installed?) if mode.installed?
-      all_deps = if ARGV.include? "--full-name"
-        all_deps.map(&:to_formula).map(&:full_name)
-      else
-        all_deps.map(&:name)
-      end.uniq
+      all_deps = all_deps.map(&method(:dep_display_name)).uniq
       all_deps.sort! unless mode.topo_order?
       puts all_deps
     end
+  end
+
+  def dep_display_name(d)
+    ARGV.include?("--full-name") ? d.to_formula.full_name : d.name
   end
 
   def deps_for_formula(f, recursive = false)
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/deps.rb
```diff
@@ -66,16 +66,16 @@ def deps
     else
       all_deps = deps_for_formulae(ARGV.formulae, !ARGV.one?, &(mode.union? ? :| : :&))
       all_deps = all_deps.select(&:installed?) if mode.installed?
-      all_deps = if ARGV.include? "--full-name"
-        all_deps.map(&:to_formula).map(&:full_name)
-      else
-        all_deps.map(&:name)
-      end.uniq
+      all_deps = all_deps.map { |d| dep_display_name(d) }.uniq
```
MikeMcQuaid: Think you can do `.map(&method(:dep_display_name)).uniq` here

Library/Homebrew/cmd/deps.rb
```diff
@@ -127,7 +127,10 @@ def deps_for_formulae(formulae, recursive = false, &block)
   end
 
   def puts_deps(formulae)
-    formulae.each { |f| puts "#{f.full_name}: #{deps_for_formula(f).sort_by(&:name) * " "}" }
+    formulae.each do |f|
+      deps = deps_for_formula(f).sort_by(&:name).map { |d| dep_display_name(d) }
```
MikeMcQuaid: Similarly here.

Library/Homebrew/cmd/deps.rb
```diff
@@ -127,7 +127,10 @@ def deps_for_formulae(formulae, recursive = false, &block)
...
+      puts "#{f.full_name}: #{deps * " "}"
```
MikeMcQuaid: I think a `.join` is a bit more readable here but no strong preference.

MikeMcQuaid:
> I haven't added new tests for this because I think that would require having a non-default tap tapped. The tap_test.rb file has code for setting up a tap, and that could be reused here with some refactoring, but that seems like overkill for this change.

Seems fine 👍

apjanke:
Amended to address code comments and merged.

PR #1707, opened 2016-12-20
brew deps --tree: fix gap in line between reqs and deps

apjanke:

-----

Minor cosmetic change. In `brew deps --tree`, if a formula has both Requirements and Dependencies, there's a gap in the line between them. This fixes it to draw the line as continuous.

Example output from `brew deps --tree mbsystem`:

Before:

```
├── gdal
│   ├── :python3
│   ├── :gcc
│   └── :git
│   ├── libpng
│   ├── jpeg
```

(Note the broken line between `:git` and `libpng`.)

After:
```
├── gdal
│   ├── :python3
│   ├── :gcc
│   ├── :git
│   ├── libpng
│   ├── jpeg
```

### DIFF
--- a/Library/Homebrew/cmd/deps.rb
+++ b/Library/Homebrew/cmd/deps.rb
@@ -143,15 +143,23 @@ module Homebrew
 
   def recursive_deps_tree(f, prefix)
     reqs = f.requirements.select(&:default_formula?)
+    deps = f.deps.default
     max = reqs.length - 1
     reqs.each_with_index do |req, i|
-      chr = i == max ? "└──" : "├──"
+      chr = if i == max && deps.empty?
+        "└──"
+      else
+        "├──"
+      end
       puts prefix + "#{chr} :#{dep_display_name(req.to_dependency)}"
     end
-    deps = f.deps.default
     max = deps.length - 1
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/deps.rb
```diff
@@ -140,12 +140,12 @@ def puts_deps_tree(formulae)
 
   def recursive_deps_tree(f, prefix)
     reqs = f.requirements.select(&:default_formula?)
+    deps = f.deps.default
     max = reqs.length - 1
     reqs.each_with_index do |req, i|
-      chr = i == max ? "└──" : "├──"
+      chr = i == max && deps.empty? ? "└──" : "├──"
```
MikeMcQuaid: I'd make this an `if`/`else` as there's a bit too much logic going on here (even before, arguably) for a ternary operator.

apjanke: Amended PR to use if/else both here and in the following block, to keep parallel code consistent.

MikeMcQuaid:
Some conflicts to resolve but 👍 :shipit: after that.

apjanke:
Conflicts resolved and merged.

PR #1708, opened 2016-12-20
install tap cmd completions

joshka:

-----

Taps can include completion scripts for external commands under
completions/bash, completions/fish, or completions/zsh. brew tap
will automatically install these into the correct directories during
install.

See Homebrew/homebrew-services#73 for a concrete need for completions in a repo.

This PR follows on from https://github.com/Homebrew/brew/pull/1195 but github didn't pick up the extra commits automatically as part of the previous PR (possibly because of my local rebase to pickup changes made later).

### DIFF
--- a/Library/Homebrew/cmd/list.rb
+++ b/Library/Homebrew/cmd/list.rb
@@ -77,6 +77,7 @@ module Homebrew
     lib/ruby/site_ruby/[12].*
     lib/ruby/vendor_ruby/[12].*
     manpages/brew.1
+    manpages/brew-cask.1
     share/pypy/*
     share/pypy3/*
     share/info/dir
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -104,8 +104,8 @@ module Homebrew
       puts if ARGV.include?("--preinstall")
     end
 
-    link_completions_and_docs
-    Tap.each(&:link_manpages)
+    link_completions_manpages_and_docs
+    Tap.each(&:link_completions_and_manpages)
 
     Homebrew.failed = true if ENV["HOMEBREW_UPDATE_FAILED"]
 
@@ -281,7 +281,7 @@ module Homebrew
       EOS
     end
 
-    link_completions_and_docs(new_homebrew_repository)
+    link_completions_manpages_and_docs(new_homebrew_repository)
 
     ohai "Migrated HOMEBREW_REPOSITORY to #{new_homebrew_repository}!"
     puts <<-EOS.undent
@@ -302,16 +302,11 @@ module Homebrew
     $stderr.puts e.backtrace
   end
 
-  def link_completions_and_docs(repository = HOMEBREW_REPOSITORY)
+  def link_completions_manpages_and_docs(repository = HOMEBREW_REPOSITORY)
     command = "brew update"
-    link_src_dst_dirs(repository/"completions/bash",
-                      HOMEBREW_PREFIX/"etc/bash_completion.d", command)
-    link_src_dst_dirs(repository/"docs",
-                      HOMEBREW_PREFIX/"share/doc/homebrew", command, link_dir: true)
-    link_src_dst_dirs(repository/"completions/zsh",
-                      HOMEBREW_PREFIX/"share/zsh/site-functions", command)
-    link_src_dst_dirs(repository/"manpages",
-                      HOMEBREW_PREFIX/"share/man/man1", command)
+    Utils::Link.link_completions(repository, command)
+    Utils::Link.link_manpages(repository, command)
+    Utils::Link.link_docs(repository, command)
   rescue => e
     ofail <<-EOS.undent
       Failed to link all completions, docs and manpages:
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -9,6 +9,7 @@ require "utils/git"
 require "utils/github"
 require "utils/hash"
 require "utils/inreplace"
+require "utils/link"
 require "utils/popen"
 require "utils/tty"
 require "time"
@@ -479,38 +480,6 @@ def truncate_text_to_approximate_size(s, max_bytes, options = {})
   out.encode!("UTF-16", invalid: :replace)
   out.encode!("UTF-8")
   out
-end
-
-def link_src_dst_dirs(src_dir, dst_dir, command, link_dir: false)
-  return unless src_dir.exist?
-  conflicts = []
-  src_paths = link_dir ? [src_dir] : src_dir.find
-  src_paths.each do |src|
-    next if src.directory? && !link_dir
-    dst = dst_dir/src.relative_path_from(src_dir)
-    if dst.symlink?
-      next if src == dst.resolved_path
-      dst.unlink
-    end
-    if dst.exist?
-      conflicts << dst
-      next
-    end
-    dst_dir.parent.mkpath
-    dst.make_relative_symlink(src)
-  end
-
-  return if conflicts.empty?
-  onoe <<-EOS.undent
-    Could not link:
-    #{conflicts.join("\n")}
-
-    Please delete these paths and run `#{command}`.
-  EOS
-end
-
-def link_path_manpages(path, command)
-  link_src_dst_dirs(path/"man", HOMEBREW_PREFIX/"share/man", command)
 end
 
 def migrate_legacy_keg_symlinks_if_necessary
--- a/Library/Homebrew/utils/link.rb
+++ b/Library/Homebrew/utils/link.rb
@@ -1,0 +1,70 @@
+module Utils
+  module Link
+    module_function
+
+    def link_src_dst_dirs(src_dir, dst_dir, command, link_dir: false)
+      return unless src_dir.exist?
+      conflicts = []
+      src_paths = link_dir ? [src_dir] : src_dir.find
+      src_paths.each do |src|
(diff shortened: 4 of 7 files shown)


Library/Homebrew/cmd/list.rb
```diff
@@ -76,7 +76,8 @@ def list
     lib/ruby/gems/[12].*
     lib/ruby/site_ruby/[12].*
     lib/ruby/vendor_ruby/[12].*
-    manpages/brew.1
+    manpages/man1/brew.1
+    manpages/man1/brew-cask.1
```
MikeMcQuaid: Are there taps using non-`.1` manpages (or wanting to)?

Library/Homebrew/cmd/update-report.rb
```diff
@@ -302,16 +302,11 @@ def migrate_legacy_repository_if_necessary
     $stderr.puts e.backtrace
   end
 
-  def link_completions_and_docs(repository = HOMEBREW_REPOSITORY)
+  def link_additional_files(repository = HOMEBREW_REPOSITORY)
```
MikeMcQuaid: I think perhaps `link_completions_manpages_and_docs` would better signify the intent here.

Library/Homebrew/utils.rb
```diff
@@ -500,8 +500,39 @@ def link_src_dst_dirs(src_dir, dst_dir, command, link_dir: false)
...
+  src_paths = unlink_dir ? [src_dir] : src_dir.find
+  src_paths.each do |src|
+    next if src.directory? && !unlink_dir
+    dst = dst_dir/src.relative_path_from(src_dir)
+    dst.delete if dst.symlink? && src == dst.resolved_path
+    dst.parent.rmdir_if_possible
+  end
+end
+
+def install_manpages(path, command)
```
MikeMcQuaid: Can we keep the naming of these to be `(un)link_*`?  Also, probably worth splitting these into a new `utils/link.rb` or similar file.

MikeMcQuaid:
Nice work here so far. A few comments but looking good 👍

Library/Homebrew/cmd/list.rb
joshka: This allows us to handle the man pages just as a generic nested folder for install / uninstall rather than specifically handling the man case. I'd like to leave this like this for simplicity. It makes it obvious where a non.1 case would belong (arguablely brew should be .8)

Library/Homebrew/cmd/update-report.rb
joshka: But what happens when we add a config file, some other related thing, ...

Library/Homebrew/utils.rb
joshka: I like install because it tells the intent rather than the effect. How this occurs might change without too much hassle, but why it's happening remains the same

joshka: Kinda agree on link.rb

Library/Homebrew/cmd/list.rb
MikeMcQuaid: Do we have any non-`.1` cases currently, though? If not, I'd rather leave as-is (or at least make that another PR).

Library/Homebrew/cmd/update-report.rb
MikeMcQuaid: 1) YAGNI, 2) we can change the name then. `link_additional_files` doesn't really communicate anything.

Library/Homebrew/utils.rb
MikeMcQuaid: Let's leave it as `link` but move these to `link.rb` 👍

Library/Homebrew/cmd/list.rb
MikeMcQuaid: @joshka Ping on this, let's leave it as-is for now, please and save moving these around to another PR.

Library/Homebrew/cmd/update-report.rb
MikeMcQuaid: @joshka Let's use `link_completions_manpages_and_docs`, thanks.

Library/Homebrew/utils.rb
MikeMcQuaid: Again, let's leave these as `link`, thanks.

MikeMcQuaid:
@joshka Ping on the changes/conflicts here, thanks.

joshka:
@MikeMcQuaid I had a rather extended Christmas / new year vacation and then moved house and had to put this on the back burner. I'll take a look later today and get it back on track. :)

joshka:
@MikeMcQuaid  I've fixed up the review comments. I'm happy to rebase / squash / resolve conflicts once you've looked at it.

joshka:
I've rebased against master, squashed the WIP commits, and fixed the review item that I missed (the link functions are now in link.rb)

Library/Homebrew/cmd/update-report.rb
```diff
@@ -302,16 +303,11 @@ def migrate_legacy_repository_if_necessary
...
-                      HOMEBREW_PREFIX/"etc/bash_completion.d", command)
-    link_src_dst_dirs(repository/"docs",
-                      HOMEBREW_PREFIX/"share/doc/homebrew", command, link_dir: true)
-    link_src_dst_dirs(repository/"completions/zsh",
-                      HOMEBREW_PREFIX/"share/zsh/site-functions", command)
-    link_src_dst_dirs(repository/"manpages",
-                      HOMEBREW_PREFIX/"share/man/man1", command)
+    link_completions(repository, command)
+    link_manpages(repository, command)
+    link_docs(repository, command)
```
MikeMcQuaid: Given this function is only called from here and only applies to Homebrew/brew I wonder if the indirection here is a little confusing (but have no strong views either way).

Library/Homebrew/link.rb
```diff
@@ -0,0 +1,62 @@
+def link_src_dst_dirs(src_dir, dst_dir, command, link_dir: false)
```
MikeMcQuaid: This should be in `utils/link.rb` and be inside a `module Link`

Library/Homebrew/link.rb
```diff
@@ -0,0 +1,62 @@
...
+  src_paths = unlink_dir ? [src_dir] : src_dir.find
+  src_paths.each do |src|
+    next if src.directory? && !unlink_dir
+    dst = dst_dir/src.relative_path_from(src_dir)
+    dst.delete if dst.symlink? && src == dst.resolved_path
+    dst.parent.rmdir_if_possible
+  end
+end
+
+def link_manpages(path, command)
```
MikeMcQuaid: I realise I'm changing my mind but could always be `link_man1pages` or something similar if you wanted to communicate this only refers to `man1`

Library/Homebrew/link.rb
MikeMcQuaid: Additionally, this method could be `private.

Library/Homebrew/link.rb
```diff
@@ -0,0 +1,62 @@
...
+  return if conflicts.empty?
+  onoe <<-EOS.undent
+    Could not link:
+    #{conflicts.join("\n")}
+
+    Please delete these paths and run `#{command}`.
+  EOS
+end
+
+def unlink_src_dst_dirs(src_dir, dst_dir, unlink_dir: false)
```
MikeMcQuaid: This method could be `private`

Library/Homebrew/cmd/update-report.rb
joshka: I'm not sure that I understand your question. Are you talking about `link_completions_manpages_and_docs()` or link_docs() method here?
If it's the former, the method exists to DRY up the same code in the taps.
If it's the latter, the method exists for consistency with the other `link_` methods.

Or perhaps you're suggesting to inline the calls to `link_*` at the call sites? I'm a fan of smaller methods rather than larger generally, and we need to call `Tap.link_completions_and_manpages` externally so I think for consistency this means this stays.

MikeMcQuaid: `link_docs`, yes. If you feel strongly: I don't so leave as-is.

Library/Homebrew/link.rb
joshka: I think it makes enough sense as-is to leave it and only update the implementation if a situation presents that other man sections are needed. Adding manX would likely mean reorganizing a bunch of folders across various repos, so it would be a fairly heavy change that isn't just a matter of adding a new method

joshka:
Addressed module change

Library/Homebrew/utils/link.rb
```diff
@@ -0,0 +1,68 @@
+module Link
```
MikeMcQuaid:
```ruby
module Utils
  module Link
```

MikeMcQuaid:
Final bit of pedantry but otherwise looking good. Once that's addressed I'll give this some testing and get Homebrew/homebrew-services ready to go with it then 🚢.

Library/Homebrew/utils/link.rb
joshka: Cool with me. I wasn't sure about the idiomatic ruby way on this was and the modules under `/utils` seemed to follow both approaches.

joshka:
Done. I note that while this PR adds more tests as an absolute number, the code coverage check on the diff reports that it's not quite enough. I think this is due to a large portion of this diff being a refactor of methods to a new module.

MikeMcQuaid:
Thanks for your work and patience here @joshka!

PR #1721, opened 2016-12-23
Acceptable-Formulae.md: e.g. to i.e.

vitorgalvao:

-----

Doesn’t really make sense to have these as “e.g.”. They’re more explanations than examples.


MikeMcQuaid:
I don't quite agree with this; those are both examples but not the only reasons. In this case using `that is` makes less sense than `for example`. Thanks for the PR anyway, though, I do appreciate the attention to detail 👍

PR #1723, opened 2016-12-23
formula: don't expand unused optional dependencies

alyssais:

—--

This properly addresses the issue reported in Homebrew/homebrew-core#7826, so `Formula#runtime_dependencies` no longer attempts to expand optional dependencies that aren't being used in the current build.

As such, it reverts #1665, and so unreverts #1592.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1505,7 +1505,12 @@ class Formula
   # Returns a list of Dependency objects that are required at runtime.
   # @private
   def runtime_dependencies
-    recursive_dependencies { |_, dep| Dependency.prune if dep.build? }
+    recursive_dependencies do |_dependent, dependency|
+      Dependency.prune if dependency.build?
+      if dependency.optional? || dependency.recommended?
+        Dependency.prune unless build.with?(dependency)
+      end
+    end
   end
 
   # Returns a list of formulae depended on by this formula that aren't
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1505,7 +1505,7 @@ class Formula
   # Returns a list of Dependency objects that are required at runtime.
   # @private
   def runtime_dependencies
-    recursive_dependencies.reject(&:build?)
+    recursive_dependencies { |_, dep| Dependency.prune if dep.build? }
   end
 
   # Returns a list of formulae depended on by this formula that aren't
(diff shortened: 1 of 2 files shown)


Library/Homebrew/formula.rb
```diff
@@ -1505,7 +1505,12 @@ def recursive_requirements(&block)
   # Returns a list of Dependency objects that are required at runtime.
   # @private
   def runtime_dependencies
-    recursive_dependencies.reject(&:build?)
+    recursive_dependencies do |_dependent, dependency|
+      Dependency.prune if dependency.build?
+      if dependency.optional? || dependency.recommended?
```
MikeMcQuaid: `if build.without?(dependency) && (dependency.optional? || dependency.recommended?)` may be easier to follow. Could also consider having `optional_dependency = dependency.optional? || dependency.recommended?` before that, too.

MikeMcQuaid:
@alyssais when you consider this good to 🚢, can you provide a wee test plan I can run locally to reproduce the issue this is fixing? Thanks!

alyssais:
@MikeMcQuaid I reproduced the issue by running `Formula['ffmpeg'].runtime_dependencies` in `brew irb`, and seeing if it produced an exception or not. I imagine you'll want a more realistic test scenario, so I'll see if I can come up with one…

Library/Homebrew/formula.rb
alyssais: How's that @MikeMcQuaid?

alyssais:
Going to force push because the history has ended up a bit unclear here…

alyssais:
Okay, there are two things to check here (I assume building from source):

1. Does `brew install influxdb` record `go` as a runtime dependency in `influxdb`'s install receipt? The correct behaviour is not to.
2. Does `brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/636e5b8ee684c4885f0a296cf8bf27f6faa9eba5/Formula/ffmpeg.rb
&& bin/brew upgrade ffmpeg` succeed, or does the upgrade fail with a "No available formula with the name "homebrew/science/opencv"" error? The correct behaviour is for it to succeed.

Currently, only _2_ should succeed. As of 3f72482, only _1_ should succeed. With this PR as a whole, both _1_ and _2_ should succeed.

MikeMcQuaid:
Thanks again @alyssais!

PR #1737, opened 2016-12-27
keg: installed dependencies of unknown formulae

### LINKED ISSUE #1586: Cannot uninstall dependents unless the dependencies' formulae still exist
CC @alyssais @MikeMcQuaid 

Reproducer:
```
$ brew pull https://github.com/Homebrew/homebrew-core/pull/7288
$ brew install -dvs gtkglext
$ brew install open-scene-graph
$ brew uninstall open-scene-graph
Error: No available formula with the name "pangox-compat"
$ brew uninstall --ignore-dependencies open-scene-graph
Uninstalling /usr/local/Cellar/open-scene-graph/3.5.5... (782 files, 36.9M)
```

As a result, I will now have to revision bump open-scene-graph even though it's not at all broken.

Edit: Note that open-scene-graph does not itself even depend on pangox-compat directly, only via gtkglext, which no longer depends on pangox-compat if 7288 has been pulled. So pangox-compat is not actually anywhere in the dependency tree of open-scene-graph when the error occurs:
```
iMac-TMP:homebrew-core joe$ brew deps open-scene-graph|tr '\n' ' '
atk cairo fontconfig freetype gdk-pixbuf gettext glib gobject-introspection gtk+ gtkglext harfbuzz hicolor-icon-theme icu4c jpeg libffi libpng libtiff openssl pango pcre pixman pkg-config shared-mime-info wget
```

alyssais:

-----

Previously, trying to resolve the dependencies of a keg would raise an exception if the formulae for any of the dependencies could not be found (e.g. if it had been moved to another tap).

This commit updates the dependency finding logic to catch these exceptions, and fall back to comparing names and taps of formulae, which should give the correct behaviour.

Fixes #1586.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -357,7 +357,7 @@ class Keg
   end
 
   def installed_dependents
-    my_tab = Tab.for_keg(self)
+    tap = Tab.for_keg(self).source["tap"]
     Keg.all.select do |keg|
       tab = Tab.for_keg(keg)
       next if tab.runtime_dependencies.nil? # no dependency information saved.
@@ -368,7 +368,7 @@ class Keg
           dep_formula = Formulary.factory(dep["full_name"])
           next false unless dep_formula == to_formula
         rescue FormulaUnavailableError
-          next false unless my_tab["full_name"] = dep["full_name"]
+          next false unless "#{tap}/#{name}" == dep["full_name"]
         end
 
         dep["version"] == version.to_s
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -139,6 +139,10 @@ class Keg
     raise NotAKegError, "#{path} is not inside a keg"
   end
 
+  def self.all
+    Formula.racks.flat_map(&:subdirs).map { |d| new(d) }
+  end
+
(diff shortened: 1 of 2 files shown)


alyssais:
I had to cherry-pick 0a20edf from #1527 to work around #1526.

alyssais:
To test this:

1. Install formula X with a non-core-tap dependency
2. Untap that tap
3. `brew uninstall X`

Previously, this would give you an error. Now, it shouldn't.

Library/Homebrew/keg.rb
```diff
@@ -353,14 +357,21 @@ def to_formula
...
       tab.runtime_dependencies.any? do |dep|
         # Resolve formula rather than directly comparing names
         # in case of conflicts between formulae from different taps.
-        dep_formula = Formulary.factory(dep["full_name"])
-        dep_formula == to_formula && dep["version"] == version.to_s
+        begin
+          dep_formula = Formulary.factory(dep["full_name"])
+          next false unless dep_formula == to_formula
+        rescue FormulaUnavailableError
+          next false unless my_tab["full_name"] = dep["full_name"]
```
alyssais: Single equals… oops

MikeMcQuaid:
Thanks @alyssais!

PR #1743, opened 2016-12-28
Delete pkgutil directories that are really files.

toonetown:
- [x] Have you successfully run `brew cask-tests` with your changes locally?

-----

Sometimes, pkgutil will return actual files (usually .nib files) as if they were part of the directory.  Microsoft Office is an example of this: in a recent update the file `/Library/Application Support/Microsoft/MAU2.0/Microsoft AutoUpdate.app/Contents/SharedSupport/Microsoft Error Reporting.app/Contents/Resources/en.lproj/MainWindowAlt.nib` was returning from `/usr/sbin/pkgutil --only-dirs --files com.microsoft.package.component` even though it should have been a file instead of a directory.  This caused the `rmdir` command to fail.

This patch will check if we are trying to delete a “directory” that is really a “file” - and if we are, we just delete the file instead.  This will allow packages that get in this state to be uninstalled cleanly.

--

Fixes https://github.com/caskroom/homebrew-cask/issues/29804.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/pkg.rb
+++ b/Library/Homebrew/cask/lib/hbc/pkg.rb
@@ -26,6 +26,7 @@ module Hbc
       _deepest_path_first(pkgutil_bom_dirs).each do |dir|
         next unless dir.exist? && !MacOS.undeletable?(dir)
         _with_full_permissions(dir) do
+          _delete_broken_file_dir(dir) && next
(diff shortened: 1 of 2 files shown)


toonetown:
Oops - forgot to run `brew style` on my file... 😁

reitermarkus:
@toonetown, thanks!

PR #1713, opened 2016-12-21
cask depends_on arch: remove everything but 64-bit intel

vitorgalvao:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [ ] Have you successfully run `brew tests` with your changes locally?

-----

[`depends_on arch:`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/depends_on.md#depends_on-arch) is, quite simply, useless as of now. Homebrew-cask itself will likely only run on 64-bit intel, and anything else is too old to care for and support, anyway.

Already made [a PR to remove all of these from casks](https://github.com/caskroom/homebrew-cask/pull/28099) and [from the documentation](https://github.com/caskroom/homebrew-cask/pull/28102).

I’d be wary of removing `depends_on arch:` as a whole, though, lest ARM Macs become a reality.

Ping @Homebrew/cask.

PR #1745, opened 2016-12-29
cask depends_on arch: remove everything but 64-bit intel

vitorgalvao:

-----

Redo of #1713, with passing tests.

[`depends_on arch:`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/depends_on.md#depends_on-arch) is, quite simply, useless as of now. Homebrew-cask itself will likely only run on 64-bit intel, and anything else is too old to care for and support, anyway.

Already made [a PR to remove all of these from casks](https://github.com/caskroom/homebrew-cask/pull/28099) and [from the documentation](https://github.com/caskroom/homebrew-cask/pull/28102) (merged).

I’d be wary of removing `depends_on arch:` as a whole, though, lest ARM Macs become a reality.

Ping @Homebrew/cask.


Library/Homebrew/cask/test/support/Casks/with-depends-on-arch-failure.rb
```diff
@@ -1,12 +0,0 @@
-test_cask 'with-depends-on-arch-failure' do
-  version '1.2.3'
-  sha256 '67cdb8a02803ef37fdbf7e0be205863172e41a561ca446cd84f0d7ab35a99d94'
-
-  url "file://#{TEST_FIXTURE_DIR}/cask/caffeine.zip"
-  homepage 'http://example.com/with-depends-on-arch-failure'
-
-  # guarantee mismatched hardware
```
vitorgalvao: I’ve removed this test because this is no longer possible. If there is only one option, we can’t have multiple valid options that are mutually exclusive.

There is [a different test for invalid values](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/test/support/Casks/invalid/invalid-depends-on-arch-value.rb).

Library/Homebrew/cask/lib/hbc/dsl/depends_on.rb
```diff
@@ -13,30 +13,18 @@ class DependsOn
...
+        intel:    { type: :intel, bits: 64 },
         # specific
-        i386:     { type: :intel, bits: 32 },
         x86_64:   { type: :intel, bits: 64 },
-        ppc_7400: { type: :ppc,   bits: 32 },
-        ppc_64:   { type: :ppc,   bits: 64 },
       }.freeze
 
       # Intentionally undocumented: catch variant spellings.
       ARCH_SYNONYMS = {
```
reitermarkus: I feel like we should lose these synonyms.

vitorgalvao: Agreed. Done.

vitorgalvao:
Hadn’t noticed tests broke after the removal of the synonyms. Should be good now.

PR #1748, opened 2016-12-30
Get file extension with `curl`’s `filename_effective`.

reitermarkus:

-----

This fixes https://github.com/caskroom/homebrew-cask/issues/28348 in that the “real“ filename is stored as an extended attribute on the cached file named `token--version.ext`. Otherwise a `.pkg` may not be detectable without the right extension.

There is a bit of a problem here though, which could break some casks.

Let's take https://github.com/caskroom/homebrew-cask/issues/28348 as an example. If we fix it now, we would have to rename the `pkg` stanza to `airvpn--#{version}.pkg`, some casks are already using this naming, which will break, because with this PR they will now have the “real“ filename, i. e `airvpn_osx_x64_installer.pkg`.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/container/criteria.rb
+++ b/Library/Homebrew/cask/lib/hbc/container/criteria.rb
@@ -9,6 +9,14 @@ module Hbc
       end
 
       def extension(regex)
+        effective_filename = @command.run!("/usr/bin/xattr", args: ["-p", "curl.filename_effective", @path]).stdout.chomp
+
+        path = if effective_filename.empty?
+          @path
+        else
+          Pathname.new(effective_filename)
+        end
+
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cask/lib/hbc/container/criteria.rb
```diff
@@ -9,6 +9,12 @@ def initialize(path, command)
       end
 
       def extension(regex)
+        path = if (effective_filename = @command.run!("/usr/bin/xattr", args: ["-p", "curl.filename_effective", @path]).stdout.chomp).empty?
+          @path
+        else
+          Pathname.new(effective_filename)
+        end
+
```
reitermarkus: I am aware this is not DRY yet.

MikeMcQuaid:
@reitermarkus Any news on this?

MikeMcQuaid:
> @reitermarkus Any news on this?

@reitermarkus March version: any news on this? 😉

Issue #1524, opened 2016-11-17
brew uninstall can have counter-intuitive (and frustrating) results


telemachus:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?

### Bug reports:

Calling `brew uninstall x y` can have very unhelpful results in cases where `y` depends on `x`. For example:

    $ brew uninstall gnupg2 dirmngr
    Error: Refusing to uninstall /usr/local/Cellar/dirmngr/1.1.1_2
    because it is required by gnupg2, which is currently installed.
    You can override this and force removal with:
      brew uninstall --ignore-dependencies dirmngr

Ideally `brew` should be able to work out that in the same command where I'm trying to uninstall `dirmngr`, I'm also uninstalling `gnupg2`. No dependency will be left unfulfilled, so I shouldn't need to add an extra flag here.

I'm not sure what exactly `brew` is doing here behind the scenes, but I suppose what I'm asking is that it does it a bit smarter. That is, if the blocker (here `gnupg2`) is *also* in the call to uninstall, then `brew` should go ahead and remove the item, shouldn't it? (There may be something I'm missing here. Sorry, if so. Over the next week or so, I can try to take a look at how `brew remove` is implemented and see if I can help with a PR as well.)

MikeMcQuaid:
> I'm not sure what exactly brew is doing here behind the scenes, but I suppose what I'm asking is that it does it a bit smarter.

Yep, a good point. CC @alyssais for thoughts on this and if you'd like to jump in on this.

alyssais:
Oh, interesting. It's supposed to be able to work that out — I think we even have a test for it! I'll look into why this isn't working.

scpeters:
I ran into a similar issue when trying to:

~~~
brew remove `brew list`
~~~

In that case I just added `--ignore-dependencies` and now it's working.

alyssais:
Note: since #1723, recommended dependencies (like `dirmngr` for `gnupg2`) will no longer be recorded in install receipts as runtime dependencies, so `brew uninstall dirmngr` will no longer error, even on its own.

To test for this bug, then, we need to use a required dependency instead, so I've been using this:

```sh
brew uninstall libgpg-error gnupg2
```

alyssais:
As far as I can tell, this is fixed in `master`. I can't reproduce it any more. :D

alyssais:
[...]

PR #1752, opened 2016-12-30
keg: fix fallback dependencies with multiple kegs

alyssais:

-----

We didn't remove formulae to be uninstalled from the list of installed formulae to check for dependencies, but only in the fallback. This is what caused #1524, and made it difficult to reproduce (because one had to realise that it was only in the fallback).

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -87,11 +87,23 @@ class Keg
     mime-info pixmaps sounds postgresql
   ].freeze
 
-  # Will return some kegs, and some dependencies, if they're present.
+  # Given an array of kegs, this method will try to find some other kegs
+  # that depend on them.
+  #
+  # If it does, it returns:
+  # - some kegs in the passed array that have installed dependents
+  # - some installed dependents of those kegs.
+  #
+  # If it doesn't, it returns nil.
+  #
+  # Note that nil will be returned if the only installed dependents
+  # in the passed kegs are other kegs in the array.
+  #
   # For efficiency, we don't bother trying to get complete data.
   def self.find_some_installed_dependents(kegs)
     # First, check in the tabs of installed Formulae.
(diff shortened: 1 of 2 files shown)


Library/Homebrew/keg.rb
```diff
@@ -105,7 +105,13 @@ def self.find_some_installed_dependents(kegs)
     #
     # This happens after the initial dependency check because it's sloooow.
     remaining_formulae = Formula.installed.select do |f|
-      f.installed_kegs.any? { |k| Tab.for_keg(k).runtime_dependencies.nil? }
+      installed_kegs = f.installed_kegs
+
+      # All installed kegs are going to be removed anyway,
```
MikeMcQuaid: Can you elaborate a bit more on this? The "removed anyway" bit makes sense in context of the PR but I'm not sure it does in the context of the function itself which isn't necessarily relating specifically to `uninstall`.

alyssais: Sure. The purpose of `Keg.find_some_installed_dependents` is to take some kegs and see if there are any kegs outside of that array that depend on those kegs. (That's why it takes an array.) This is what it [does](https://github.com/Homebrew/brew/blob/c29a458dc15efc92e3649155d0ed44bf982cdc15/Library/Homebrew/keg.rb#L95) currently when the install receipts are reliable.

But, in the fallback, I forgot to remove the passed kegs from the list of dependents, so now I've fixed that.

MikeMcQuaid: Cool, makes sense. I think the comment could be expanded on a bit to make this clearer e.g. why they will be "removed"

alyssais: Will do 👍

MikeMcQuaid:
Perfect. Great comments.

MikeMcQuaid:
Thanks @alyssais!

PR #1753, opened 2016-12-30
Suggested solution for 'Filter all environment variables by default #932' issue

doughazell:
1) Code to list env vars to be readded after removing all
2) Changes necessary to be able to add " require 'byebug' " and 'debugger' to check passed ENV

-----


Library/Homebrew/brew.sh
```diff
@@ -76,7 +76,11 @@ export GEM_OLD_PATH="$GEM_PATH"
 # Users may have these set, pointing the system Ruby
 # at non-system gem paths
 unset GEM_HOME
-unset GEM_PATH
```
MikeMcQuaid: I'd just remove both these `unset` and let your script handle them. You can check through all the `unset` in this file and remove them all, I think.

Library/Homebrew/brew.sh
```diff
@@ -267,6 +271,12 @@ access to all bottles."
 EOS
 fi
 
+if [[ -z "$HOMEBREW_ENV_PRUNED" ]]
```
MikeMcQuaid: I like the double-guard here but probably warrants a comment explaining this is what it's doing.

Library/Homebrew/brew.sh
```diff
@@ -336,7 +346,12 @@ else
   # shellcheck source=/dev/null
   source "$HOMEBREW_LIBRARY/Homebrew/utils/ruby.sh"
   setup-ruby-path
-
+  
+  if [[ -n "$HOMEBREW_WHICH_RUBY" ]]
```
MikeMcQuaid: Not sure I understand this variable, can you elaborate?

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
+keepkeys=(
+HOMEBREW_*
```
MikeMcQuaid: Indent these if possible for readability.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+GEM_*
```
MikeMcQuaid: Can strip this.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
+keepkeys=(
```
MikeMcQuaid: Favour capitalised variable names and run `shellcheck` on this file and address any complaints.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+GEM_*
+HOME
+PWD
+LANG
+PATH
+)
+
+prune-shell-env() {
+  # Associative arrays are only a Bash 4 feature
+  #declare -A aEnv
```
MikeMcQuaid: Can remove this and the comment.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+
+  local verbose
```
MikeMcQuaid: Worth setting a default value at initialisation time.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+  
+  local envPATH=`which env`
```
MikeMcQuaid: favour `env_path` or similar for local variable names and `$(...)` although in this case you can just use `env` rather than looking up the path.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+  local echoPATH=`which echo`
+  local sedPATH=`which sed`
+
+  file=$HOMEBREW_LIBRARY/Homebrew/utils/shell-env.txt
```
MikeMcQuaid: Quote `$` usage. Also, I'd consider not making this a `txt` file and storing it somewhere in `tmp`.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+)
+
+prune-shell-env() {
+  # Associative arrays are only a Bash 4 feature
+  #declare -A aEnv
+
+  local verbose
+  
+  local envPATH=`which env`
+  local echoPATH=`which echo`
```
MikeMcQuaid: Just use `echo` rather than looking up the path.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+  local sedPATH=`which sed`
```
MikeMcQuaid: Can just use `/usr/bin/sed` here.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+    $echoPATH "Writing new env to '"$pfile"'"
+    $echoPATH "---------------------"
+    $echoPATH
+  fi
+  
+  $envPATH > $pfile
+
+  
+  export HOMEBREW_ENV_PRUNED=1
+}
```
MikeMcQuaid: Newline at end of file

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,71 @@
...
+
+  done
+
+  pfile=$HOMEBREW_LIBRARY/Homebrew/utils/shell-env-pruned.txt
+  if [[ -n "$verbose" ]]
+  then
+    $echoPATH "Writing new env to '"$pfile"'"
+    $echoPATH "---------------------"
+    $echoPATH
+  fi
```
MikeMcQuaid: Not sure I understand this file, can you explain?

Library/Homebrew/utils/shell_env.sh
MikeMcQuaid: Also, CI is currently failing because `TRAVIS` is not passed through.

MikeMcQuaid:
@doughazell Is `env -i` of any use here? Seems like it could save some code.

Library/Homebrew/brew.sh
doughazell: Yea ok, sounds like a good refactor.

Library/Homebrew/brew.sh
doughazell: I think that the double-guard was originally a ksh thing but it makes the test more obvious to me.

Library/Homebrew/brew.sh
doughazell: OK a little bit a dev history left in there.

I thought that it would be useful to add the debugger to the 'cmd/install.rb' to see what ENV it was getting after the various Homebrew additions but the gem couldn't be found originally due to using 'vendor/portable-ruby' (which gets updated if changed) or the '/System' one (due to being a Mac) but what I really wanted was my RVM one with the installed 'byebug' gem (also the reason for leaving in the 'GEM_*' since my dev env was to drop into a debugger shell on install).

I left it in due to the ability to use a debugger must be useful to other devs.

doughazell:
I originally looked at 'env -i' but that would just change the child shell process env rather than altering the parent bash shell which then passes its env to the homebrew commands written in ruby.  Also I wanted to write the script in ruby rather than bash (which I've not used for 10 years!) but that would lead to the same situation whereby the parent bash process' env is not changed which the homebrew command scripts then inherit.

Library/Homebrew/utils/shell_env.sh
doughazell: Yes I've never used 'shellcheck' before and a nice "gem", :), that has come out of this process...

Library/Homebrew/utils/shell_env.sh
doughazell: I left it in to remind about bash version portability but happy to remove it if you want.

Library/Homebrew/utils/shell_env.sh
doughazell: ok

Library/Homebrew/utils/shell_env.sh
doughazell: ok

Library/Homebrew/utils/shell_env.sh
doughazell: I read in the original change conversation that someone wanted to remove the PATH which would also prevent the utilities from being found that the script uses.

Library/Homebrew/utils/shell_env.sh
doughazell: Partly a dev feature left in and maybe the pruned path could be used in other homebrew commands like 'doctor'.  Happy to remove it if you want.

Library/Homebrew/brew.sh
MikeMcQuaid: Sure, needs a comment regardless.

Library/Homebrew/brew.sh
MikeMcQuaid: Ok, let's just remove the debugger, thanks.

Library/Homebrew/utils/shell_env.sh
MikeMcQuaid: That'd be good thanks.

Library/Homebrew/utils/shell_env.sh
MikeMcQuaid: We'll set want to set the PATH to the default system value e.g. `/usr/bin:/bin:/usr/sbin:/sbin`

Library/Homebrew/utils/shell_env.sh
MikeMcQuaid: Ok, let's just kill it for now, thanks.

MikeMcQuaid:
@doughazell We could run in a new process easily enough. Check out `/usr/local/bin/brew`; that's where the `env -i` call could be instead of a sourced script.

Library/Homebrew/brew.sh
```diff
@@ -336,7 +327,7 @@ else
   # shellcheck source=/dev/null
   source "$HOMEBREW_LIBRARY/Homebrew/utils/ruby.sh"
   setup-ruby-path
-
+
```
MikeMcQuaid: Remove this whitespace-only change, thanks!

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set +o posix
+
+KeepKeys=(
+  BASH
+  BASH_ARGC
+  BASH_ARGV
+  BREW_FILE_DIRECTORY
+  USR_LOCAL_BREW_FILE_DIRECTORY
```
MikeMcQuaid: This can be filtered.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,65 @@
...
+  USR_LOCAL_HOMEBREW_REPOSITORY
```
MikeMcQuaid: This can be filtered.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+set +o posix
+
+KeepKeys=(
+  BASH
+  BASH_ARGC
+  BASH_ARGV
+  BREW_FILE_DIRECTORY
```
MikeMcQuaid: This can be filtered.

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,65 @@
...
+  PWD
+  LANG
+  PATH
+  TERM
+)
+
+prune-shell-env() {
+  file=/tmp/homebrew-env
+
+  IFS="="
```
MikeMcQuaid: I think this could be simplified quite a bit by doing something like `env | grep -E '^(HOMEBREW.*|HOME|PATH)=.*$'` and just writing `env` output to the original file. Thoughts?

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,65 @@
...
+  HOMEBREW_*
+  HOME
+  PWD
+  LANG
+  PATH
+  TERM
+)
+
+prune-shell-env() {
+  file=/tmp/homebrew-env
```
MikeMcQuaid: Actually was thinking about this more and I think the ideal location for this would be just in a `HOMEBREW_OLD_ENV` variable that can be read later so e.g. `export HOMEBREW_OLD_ENV="$(env)"`. Thoughts?

Library/Homebrew/utils/shell_env.sh
```diff
@@ -0,0 +1,65 @@
...
+
+}
+
+verbose=''
+for option in "$@"
+  do
+    case "$option" in
+      --verbose) verbose=1 ;;
+    esac
+done
```
MikeMcQuaid: I reckon you can kill the `--verbose` logic. If you want to leave some sort of debugging logic in make it conditional on an environment variable i.e. `HOMEBREW_ENV_DEBUG`

bin/brew
```diff
@@ -44,4 +44,5 @@ fi
 
 HOMEBREW_LIBRARY="$HOMEBREW_REPOSITORY/Library"
 
-source "$HOMEBREW_LIBRARY/Homebrew/brew.sh"
+set > /tmp/homebrew-env
+env -i "$HOMEBREW_LIBRARY/Homebrew/utils/shell_env.sh" "$@"
```
MikeMcQuaid: Nice 👍. I think it's maybe worth making this conditional for now on the presence of the `HOMEBREW_ENV_FILTERING` environment variable. This means we can do a gradual rollout of this feature.

MikeMcQuaid:
Squash commits by running `brew update; git checkout $YOUR_BRANCH; git rebase --interactive origin/master` and then mark each commit other than the first as `squash` or `fixup` then `git push -f`. Thanks!

doughazell:
Is that for a future occasion if my git code gets itself in a mixed up state prior to pull-req commit update, because I was expecting to create a clean new branch with all the comments removed from the code if you liked the new code?

MikeMcQuaid:
I'm afraid I can't see your current version of the code unfortunately because it's all mixed in with the merges.

bin/brew
```diff
@@ -44,4 +44,78 @@ fi
 
 HOMEBREW_LIBRARY="$HOMEBREW_REPOSITORY/Library"
 
-source "$HOMEBREW_LIBRARY/Homebrew/brew.sh"
+# Refactor of code from '/Library/Homebrew/brew.sh'
+export HOMEBREW_BREW_FILE
+export HOMEBREW_PREFIX
+export HOMEBREW_REPOSITORY
+export HOMEBREW_LIBRARY
```
MikeMcQuaid: Don't think these need exported here.

bin/brew
```diff
@@ -44,4 +44,78 @@ fi
 
 HOMEBREW_LIBRARY="$HOMEBREW_REPOSITORY/Library"
 
-source "$HOMEBREW_LIBRARY/Homebrew/brew.sh"
+# Refactor of code from '/Library/Homebrew/brew.sh'
```
MikeMcQuaid: Remove this comment.

bin/brew
```diff
@@ -44,4 +44,78 @@ fi
...
+# from an original of:
+
+#HOMEBREW_LIBRARY=/usr/local/Homebrew/Library
+#TERM=xterm-color
+#HOMEBREW_REPOSITORY=/usr/local/Homebrew
+#PATH=/usr/bin:/bin:/usr/sbin:/sbin
+#HOMEBREW_BREW_FILE=/usr/local/bin/brew
+#HOME=/Users/doug
+#HOMEBREW_PREFIX=/usr/local
+# -------------------------------------------------------------------------------------------------------------------------
```
MikeMcQuaid: Remove all the comments above here.

bin/brew
```diff
@@ -44,4 +44,78 @@ fi
...
+
+if [[ -n "$HOMEBREW_ENV_FILTERING" ]]
+then
+  # Needed TERM to prevent "tput: No value for $TERM and no -T specified"
+  env | grep -E '^(HOMEBREW.*|HOME|PATH|TERM)=.*$' > /tmp/homebrew-env
+  env -i "$HOMEBREW_LIBRARY/Homebrew/utils/shell_env.sh" "$@"
```
MikeMcQuaid: Instead of these two lines and `shell_env.sh` I think you could just do something like:
```bash
env -I \
  HOME="$HOME" \
  PATH=/usr/bin:/bin:/usr/sbin:/sbin \
  TERM="$TERM" \
  HOMEBREW_BREW_FILE="$HOMEBREW_BREW_FILE" \
  HOMEBREW_PREFIX="$HOMEBREW_PREFIX" \
  HOMEBREW_REPOSITORY="$HOMEBREW_REPOSITORY" \
  HOMEBREW_LIBRARY="$HOMEBREW_LIBRARY" \
  /bin/bash -c "source $HOMEBREW_LIBRARY/Homebrew/brew.sh"
```

DomT4:
Haven't yet worked out why, but if you run with this PR applied locally & operating by default, things come out as expected in the `/tmp` file, but if you then open that file in GitHub's Atom editor any custom `HOMEBREW_CACHE` and `HOMEBREW_LOGS` variables lose their `/usr/local` and `HOMEBREW_GITHUB_API_TOKEN` is blanked, along with another `.zcompdump` file being created in `$HOME` 🤷‍♂️.

For what it's worth my `.zshrc` sets all three of those variables using a shell command, the cache/logs look for `$(brew --prefix)` and the API key is retrieved from the Keychain using the `security` command.

Not sure how much of that you'll see in the wild to be honest, I just enjoy breaking things, but perhaps interesting.

doughazell:
But thinking about it some more then wouldn't the current solution remove any 'HOMEBREW.*' var set in the user's shell like HOMEBREW_DEBUG?

MikeMcQuaid:
@doughazell It would, yeh. We should figure out a way to expand them all without dumping to a file. That may mean just listing all variables here that we want to allow through for now.

doughazell:
Have you had problems using cache files in the past?

MikeMcQuaid:
Yeh. Generally any temporary files are going to result in sporadic permission issues, information leakage and users deleting them when they shouldn't do.

zmwangx:
As stated in #932, I think you should at least keep `HOME`, `SHELL`, `PATH`, `TERM`, `LOGNAME` and `USER`, the set of environment vars set by `login(1)` that may very be expected at runtime (during tests).

Also, more env vars than the aforementioned and `HOMEBREW_*` needs to be kept if you decide to take the current approach of enforcing a clean slate on all commands. Examples include but are not limited to `EDITOR` and `VISUAL`. Check [the `ENVIRONMENT` section](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/manpages/brew.1.md.erb#L100) of `brew.1`. Even that list is not exhaustive. Homebrew maintainers need `BINTRAY_USER` and `BINTRAY_KEY` to be whitelisted, for instance.

I think it would much easier if the scope of environment wiping is limited to variants of `install` and `test`, as was the original intention of #932.

MikeMcQuaid:
> Also, more env vars than the aforementioned and HOMEBREW_* need to be kept if you decide to take the current approach of enforcing a clean slate on all commands.

@zmwangx As you're a shell scripting guru: do you have a suggestion for whitelisting all HOMEBREW_* without a temporary file?

doughazell:
So @MikeMcQuaid you didn't like https://github.com/Homebrew/brew/pull/1753/commits/5d1576784a5603662ca9910eb97c948e833f35c9 then...(which got out of sync since my broadband has just been fixed by Openreach after going down on wed)?

MikeMcQuaid:
> So @MikeMcQuaid you didn't like 5d15767 then...(which got out of sync since my broadband has just been fixed by Openreach after going down on wed)?

@doughazell Sorry Doug, missed that. I reckon something using just the output of `env | grep ^HOMEBREW_` would be sufficient and you could perhaps use something like `| sed -e 's/=/="/' -e 's/$/"/'` to handle the whitespace.

zmwangx:
> do you have a suggestion for whitelisting all HOMEBREW_* without a temporary file?

Here's what I would do, in pure Bash:

```bash
execution_env=(
  "HOME=$HOME"
  "SHELL=$SHELL"
  "PATH=$PATH"
  "TERM=$TERM"
  "LOGNAME=$LOGNAME"
  "USER=$USER"
  # Hard-code additional key=val here
)
for key in "${!HOMEBREW_@}"
do
  execution_env+=( "${key}=${!key}" )
done
/usr/bin/env -i "${execution_env[@]}" /bin/bash "$HOMEBREW_LIBRARY/Homebrew/brew.sh" "$@"
```

Note that all variables starting with `HOMEBREW_` will be passed to the env, not only the ones exported. If for some reason you do not want to export all non-exported `HOMEBREW_*` variables, it is doable, but checking whether a variable is exported in Bash is quite ugly. Let me know if you need that.

I wouldn't recommend parsing `env(1)` output, because BSD `env` does not have `-0`, which means you cannot reliably parse the output (can't distinguish the linefeed separating two variables and a linefeed in a variable's value).

MikeMcQuaid:
> Note that all variables starting with HOMEBREW_ will be passed to the env, not only the ones exported.

That's desirable, actually.

> I wouldn't recommend parsing env(1) output, because BSD env does not have -0, which means you cannot reliably parse the output (can't distinguish the linefeed separating two variables and a linefeed in a variable's value).

Good to know, thanks. Your solution seems like a good approach.

johnhawkinson:
> > Note that all variables starting with `HOMEBREW_` will be passed to the env, not only the ones exported.

> That's desirable, actually.

Why? And what's the deal with all the other variables in /usr/local/bin/brew that don't begin with `HOMEBREW_`? There sure seem to be a lot, e.g. `BREW_FILE_DIRECTORY`, `USR_LOCAL_BREW_FILE_DIRECTORY` ... ok, not that many. But it does seem needlessly complex?

zmwangx:
> Why?

They are exported in `brew.sh` to begin with.

```
# Declared in bin/brew
export HOMEBREW_BREW_FILE
export HOMEBREW_PREFIX
export HOMEBREW_REPOSITORY
export HOMEBREW_LIBRARY
```

> And what's the deal with all the other variables in /usr/local/bin/brew that don't begin with HOMEBREW_?

They are not. Variables meant to be exported to environment all begin with `HOMEBREW_`.

> But it does seem needlessly complex?

Not sure what "it" refers to. But what I can tell you is trying to discern which variable is exported and which is not would be needlessly complex.

johnhawkinson:
> > But it does seem needlessly complex?

> Not sure what "it" refers to. But what I can tell you is trying to discern which variable is exported and which is not would be needlessly complex.

I meant to say that `bin/brew`'s use of variables and indirection seems needlessly complex. Perhaps it is not, but it seems like it uses a lot of variables very briefly. I don't disagree about discerning.

zmwangx:
> Perhaps it is not, but it seems like it uses a lot of variables very briefly.

Not a valid argument unless you point out specific problems. As for me, I wouldn't use ALL CAPS for non env vars, but existing style is existing style; other than that I see no problem.

IMO arguing about complexity of a 48 line file is a waste of time. Also, your objections don't seem to belong to this thread, but of course you're entitled to your opinions.

reitermarkus:
> As for me, I wouldn't use ALL_CAPS for non env vars, but existing style is existing style; other than that I see no problem.

> But what I can tell you is trying to discern which variable is exported and which is not would be needlessly complex.

Then again, using `snake_case` for non-env variables would eliminate the need for discerning them altogether.

zmwangx:
> Then again, using snake_case for non-env variables would eliminate the need for discerning them altogether.

No, when I talked about discerning I meant programmatic detection and exclusion of non-exported variables, basically

```bash
execution_env=()
for key in HOME SHELL PATH TERM LOGNAME USER "${!HOMEBREW_@}"
do
  # ugly, unfortunate consequence of lack of `export -p NAME` in bash's export
  [[ "$(declare -p "$key")" =~ ^declare\ -[^\ ]*x[^\ ]*\  ]] || continue
  execution_env+=( "${key}=${!key}" )
done
```

(which would exclude ` HOMEBREW_BREW_FILE`, `HOMEBREW_PREFIX`, `HOMEBREW_REPOSITORY` and `HOMEBREW_LIBRARY` as they are not exported), not visual cues for human beings.

reitermarkus:
> which would exclude HOMEBREW_BREW_FILE, HOMEBREW_PREFIX, HOMEBREW_REPOSITORY and HOMEBREW_LIBRARY as they are not exported

Right, I just meant that if we were to use`snake_case` for all non-env variables, `"${!HOMEBREW_@}"` would give us only the variables that are actually going be exported somewhere along the line.

MikeMcQuaid:
> I meant to say that bin/brew's use of variables and indirection seems needlessly complex. Perhaps it is not, but it seems like it uses a lot of variables very briefly. I don't disagree about discerning.

@johnhawkinson Whether it is or not: that's not a helpful comment. If you'd like to help us fix it: pull requests are welcome.

doughazell:
Do we have closure on this PR since it will be guarded by 'HOMEBREW_ENV_FILTERING' and can always be adapted at a later date?

MikeMcQuaid:
@doughazell I'd like to see us take @zmwangx's approach (or something related but Bash-only) but I'm happy to commit to this PR if necessary.

doughazell:
Yes defintely since it is much better to keep it Bash-only rather than resource cost of starting a ruby shell (it was my hack to seed the later refactoring).  My summary of the best solution so far is from @zmwangx:

execution_env=()
for key in HOME SHELL PATH TERM LOGNAME USER "${!HOMEBREW_@}"
do
  # ugly, unfortunate consequence of lack of `export -p NAME` in bash's export
  [[ "$(declare -p "$key" 2>/dev/null)" =~ ^declare\ -[^\ ]*x[^\ ]*\  ]] || continue
  execution_env+=( "${key}=${!key}" )
done

/usr/bin/env -i "${execution_env[@]}" /bin/bash "$HOMEBREW_LIBRARY/Homebrew/brew.sh" "$@"

Should I update my branch with the above?

MikeMcQuaid:
> Should I update my branch with the above? I'm not sure of the GitHub etiquette.

Yeh, please! Totally fine to copy-paste stuff from comments.

zmwangx:
> Should I update my branch with the above? 

Yeah, feel free but please leave out

```bash
  # ugly, unfortunate consequence of lack of `export -p NAME` in bash's export
  [[ "$(declare -p "$key" 2>/dev/null)" =~ ^declare\ -[^\ ]*x[^\ ]*\  ]] || continue
```

Also, I still want to call to attention this:

> more env vars than the aforementioned and `HOMEBREW_*` need to be kept if you decide to take the current approach of enforcing a clean slate on all commands. Examples include but are not limited to `EDITOR` and `VISUAL`. Check [the `ENVIRONMENT` section](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/manpages/brew.1.md.erb#L100) of `brew.1`. Even that list is not exhaustive. Homebrew maintainers need `BINTRAY_USER` and `BINTRAY_KEY` to be whitelisted, for instance.
>
> I think it would much easier if the scope of environment wiping is limited to variants of `install` and `test`, as was the original intention of #932.

doughazell:
Updated before I read your last comment @zmwangx and adapted your solution to make it work on Bash 3.2.48 and include the pruned PATH.

Once we get an agreed solution then it can be adapted to include what you draw attention to above.

bin/brew
```diff
@@ -44,4 +44,17 @@ fi
 
 HOMEBREW_LIBRARY="$HOMEBREW_REPOSITORY/Library"
 
-source "$HOMEBREW_LIBRARY/Homebrew/brew.sh"
+if [[ -n "$HOMEBREW_ENV_FILTERING" ]]
+then
+  PATH="/usr/bin:/bin:/usr/sbin:/sbin"
```
zmwangx: Not really necessary. Superenv does its only PATH cleansing; stdenv^ and `env :userpaths`^^ would be broken by this.

^ Someone will shout at me saying stdenv is deprecated, but it apparently still exists as far as I can tell, although I personally don't use it and am not rooting for it.
^^ Also not popular.

MikeMcQuaid: @zmwangx Still necessary for i.e. not breaking `brew update` by putting stupid things like `head` in your PATH. This will break a bunch of this, I realise, but it's the first step to getting this actually working.

zmwangx: @MikeMcQuaid Sure, if we're concerned about that vector. No objection to getting this working first.

What would you suggest for stdenv and `env :userpaths` (and possibly other things affected which I don't realize at the moment) though? Death sentence to all of them? Persist original PATH in `ORIG_PATH` and optionally restore it later? Something else?

MikeMcQuaid: > Persist original PATH in ORIG_PATH and optionally restore it later?

Yeh, this. Won't necessarily need this for this PR but that seems like the right direction to take.

zmwangx:
Since it seems we're not limiting to install (and family) and test after all, here's a list of env vars that should be considered for whitelisting:

brew:

```
$ grep --exclude-dir=test --exclude-dir=cask -rIPoh '(?<=ENV\[")(?!HOMEBREW_)\w+(?="\](?! = ))' /usr/local/Homebrew/Library/Homebrew | sort | uniq
BINTRAY_KEY
BINTRAY_USER
BROWSER
CI
CURL_HOME
FC
GEM_HOME
GEM_OLD_HOME
GEM_OLD_PATH
GIT_URL
HOME
JAVA_HOME
JENKINS_HOME
LC_ALL
PATH
PYTHONPATH
SDKROOT
SHELL
SSL_CERT_DIR
TEMP
TERM
TMP
TMPDIR
TMUX
TRAVIS
USER
VERBOSE
```

cask has some additional stuff:

```
$ grep --exclude-dir=test -rIPoh '(?<=ENV\[")(?!HOMEBREW_)\w+(?="\](?! = ))' /usr/local/Homebrew/Library/Homebrew/cask | sort | uniq
CI
CODECOV_TOKEN
HOME
MACOS_VERSION
SUDO_ASKPASS
TRAVIS
VERBOSE
```

doughazell:
OK, as a newby for the brew internals (hence choosing this issue) I can only comment on stuff that I've seen in the code (without any formula experience).

GEM_OLD_HOME and GEM_OLD_PATH
are set in '$HOMEBREW_LIBRARY/Homebrew/brew.sh' and hence will get [re]added automatically and hence don't need to be considered in the sanitisation of the interface between the user's shell and the brew system.

ilovezfs:
>Since it seems we're not limiting to install (and family) and test after all, here's a list of env vars that should be considered for whitelisting

Yikes. It seems it would be easier to list what this won't break than what it will break.

zmwangx:
Quick note on something that just came to my mind: some env vars that are not on the list above (i.e., does not appear explicitly on the list above) might have an impact too. For instance, Emacs uses [these env vars](https://www.gnu.org/software/emacs/manual/html_node/emacs/General-Variables.html#General-Variables), and users are free to reference anything in their `init.el`, so a cleansed environment might cripple a user's editor in `brew create` or `brew edit`.

I propose that we at least whitelist some dev-oriented commands (not all dev commands, because `test`, `tests` etc. would certainly benefit from the filtering).

MikeMcQuaid:
Just to stop this thread going off the rails: this feature will be shipped in the following order:

1. opt-in with an environment variable (which I'll personally enable and fix things that crop up in my workflow)
1. CI will be opted-in
1. HOMEBREW_DEVELOPERs will be opted-in
1. everyone will be opted-in

Much like many other invasive changes like this: there will be a lot of changes required between and uncovered by each step and I don't think the initial implementation needs to handle anything beyond passing CI and code review at this point.

DomT4:
What's the argument against limiting this in scope to `install`, `test` and `tests`, out of curiosity?

MikeMcQuaid:
> What's the argument against limiting this in scope to install, test and tests, out of curiosity?

- It adds scope to an already long-open PR that will have no effect on any users without an opt-in
- There are many other commands you can break with many other environment variables. This isn't just environment variables we use, either, but subcommands like `gem` that we call.
- Different commands will need different whitelisted variables

DomT4:
Thanks. Was just curious as @zmwangx kept mentioning it & it wasn't being rebutted yet, and could reduce the amount of edge cases you're going to run into, but those reasons seem reasonable.

MikeMcQuaid:
Tested locally and works as expected (I can `brew reinstall -s wget`). `brew doctor` complains about the PATH but that can be addressed in a follow-up PR.

What I'd suggest for the additional variables is that we export them as stuff like `HOMEBREW_OLD_PATH` and `HOMEBREW_EDITOR` and then we can use them when needed rather than relying on particular variables which can interfere with build systems being whitelisted for specific (or all) commands.

zmwangx:
> What I'd suggest for the additional variables is that we export them as stuff like HOMEBREW_OLD_PATH and HOMEBREW_EDITOR and then we can use them when needed...

Not always possible, because environment variables can be expected by external commands, and (1) prefixing them makes them useless; (2) restoring them at the right point is both hard work and easily forgotten. Examples:

1. `CURL_HOME`. Without this user's curlrc may be ignored. (Yes, I do use a curlrc, and I'm well aware of the fact that doctor barks at its mere existence.)
2. `ZDOTDIR`. It doesn't appear anywhere in the source code, hence not in my list above. I use this to get a usable shell from debrew.
3. `EMACSPATH`, another example of an env var that doesn't appear in brew's code, but is useful for an external command that brew may invoke.

The point I'm trying to make is that it's both useful and easy to whitelist harmless environment variables that are *used by external commands invoked by brew that users are allowed to customize* (e.g., curl, emacs, zsh). On the ease of whitelisting:

```bash

WHITELIST=(
  HOME
  SHELL
  PATH
  TERM
  LOGNAME
  USER
  # Used by external commands
  CURL_HOME
  # ...
)
FILTERED_ENV=()
for VAR in "${WHITELIST[@]}" "${!HOMEBREW_@}"
do
  FILTERED_ENV+=( "${VAR}=${!VAR}" )
done

/usr/bin/env -i "${FILTERED_ENV[@]}" /bin/bash "$HOMEBREW_LIBRARY/Homebrew/brew.sh" "$@"
```

This structure also makes it easy for users to submit PRs when they notice something that should be whitelisted.

Of course there's also the possibility of whitelisting certain dev commands. For now I won't argue for it.

MikeMcQuaid:
> restoring them at the right point is both hard work and easily forgotten

I'd still rather try this approach before abandoning it before we've attempted it.

> CURL_HOME. Without this user's curlrc may be ignored.

I consider this a feature rather than a bug, I'm afraid.

> The point I'm trying to make is that it's both useful and easy to whitelist harmless environment variables that are used by external commands invoked by brew that users are allowed to customize

It remains to be seen how many are truly harmless i.e. misconfiguration of them cannot break Homebrew. It's also worth noting that all those variables you mention above would not be used by all Homebrew commands. Customisation is something that's directly proportional to our support burden.

> This structure also makes it easy for users to submit PRs when they notice something that should be whitelisted.

It's also going to introduce a list that's going to be endlessly added to.

---

Let's stick with the current approach for now and we can iterate on it. There's a fair bit of YAGNI in here which we can punt on for now.

zmwangx:
> CURL_HOME. Without this user's curlrc may be ignored.
>
> I consider this a feature rather than a bug, I'm afraid.

Except respecting user's curlrc was a design decision. #516. You and most users may not need a curlrc, but it could be very useful for people with bad connections, for instance (I say this from first-hand experience).

> It remains to be seen how many are truly harmless i.e. misconfiguration of them cannot break Homebrew.

They haven't been breaking Homebrew so far, AFAIK. (With the exception of misconfiguring curlrc, which is like misconfiguring your home router and end up not being able to connect to the Internet — there's always a way one can screw themself.)

One can also present convincing arguments on a case-by-case basis.

> It's also going to introduce a list that's going to be endlessly added to.

1. I quote:

    > harmless environment variables that are used by external commands invoked by brew that users are allowed to customize (e.g., curl, emacs, zsh)

    That list won't be endless (with an asterisk), because brew only ever invokes a handful of external commands that aren't dumb plumbing commands.

2. Even if the list becomes long (it won't), the practical downside (as opposed to theoretical support burden) is ...? It's just a straightforward, highly visible list.

> It's also worth noting that all those variables you mention above would not be used by all Homebrew commands.

Of course. I was proposing the simplest approach (which IMO has the biggest chance of winning favor).

> Let's stick with the current approach for now and we can iterate on it. 

What I proposed effortlessly fits into the current approach though.

MikeMcQuaid:
> With the exception of misconfiguring curlrc, which is like misconfiguring your home router and end up not being able to connect to the Internet — there's always a way one can screw themself.

The whole reason for this feature exists is to stop people from screwing themselves.

> One can also present convincing arguments on a case-by-case basis.

And environment variables can be added as `HOMEBREW_` and used by specific commands on a case-by-case basis which is the option that means they are applied where needed rather than universally.

zmwangx:
> The whole reason for this feature exists is to stop people from screwing themselves.

Real people who actually need or benefit immensely from a curlrc would be screwed to stop a couple of people who don't need it from screwing themselves (I looked at almost all threads containing the keyword curlrc, and only a few of them are failures that might be attributed to curlrc). That doesn't sound fair to me.

Anyway, this is not the thread to argue about curlrc, and I assume `CURL_HOME` can be whitelisted on a case-by-case basis, however that whitelisting is implemented. (This PR only invalidated `$CURL_HOME/.curlrc`, it doesn't even affect `$HOME/.curlrc`; it would take another PR to throw out #516.)

> And environment variables can be added as HOMEBREW_ and used by specific commands on a case-by-case basis which is the option that means they are applied where needed rather than universally.

Sure, then it's just a more conservative and complicated way to do the same thing. "Growing list" concern applies there, too, assuming it is a concern.

MikeMcQuaid:
> more conservative

That's the intention, yep.

PR #1756, opened 2016-12-31
keg: don't return nil dependencies

alyssais:

-----

This is a more correct solution to [this problem](https://github.com/Homebrew/homebrew-core/issues/3958#issuecomment-260632390) than @MikeMcQuaid's quick fix in #1510.

The problem arises when trying to uninstall at least two dependencies of the same keg (which must have `null` `runtime_dependencies` in its install receipt), when the name and tap information in some (but not all) of the install receipts of those dependencies do not match the current name and tap in the corresponding formulae (e.g. if the formula was renamed or moved, or it was installed from a file or URL). What an edge case!

Since #1751, it'll no longer produce a `NoMethodError`, but it will still produce a slightly broken error message without either fix (#1510 or this one). Note the floating comma:

<pre><samp>
Warning: , /Users/alyssa/code/brew/Cellar/readline/7.0.1 are required by clisp, which is currently installed.
You can silence this warning with:
  brew uninstall --ignore-dependencies libsigsegv readline
</samp></pre>

This PR fixes the problem by removing the `nil` values as soon as they occur. (Not doing this before was an oversight on my part — it was always the intended behaviour.)

To test:

1. Install a keg with a null tap in its install receipt: <kbd>brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/libsigsegv.rb</kbd>
2. Install a keg that depends on that keg and another (normal) one: <kbd>brew install clisp</kbd>
3. Edit `clisp`'s install receipt to set the `runtime_dependencies` to `null`.
4. Check if there are empty items in the list in the warning/error message generated by <kbd>brew uninstall libsigsegv readline</kbd>. (There shouldn't be.)

### DIFF
--- a/Library/Homebrew/cmd/uninstall.rb
+++ b/Library/Homebrew/cmd/uninstall.rb
@@ -102,8 +102,8 @@ module Homebrew
     attr_reader :reqs, :deps
 
     def initialize(requireds, dependents)
-      @reqs = requireds.compact
-      @deps = dependents.compact
+      @reqs = requireds
+      @deps = dependents
     end
 
     protected
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -119,7 +119,7 @@ class Keg
         next unless f_kegs
 
         f_kegs.sort_by(&:version).last
-      end
+      end.compact
 
       next unless required_kegs.any?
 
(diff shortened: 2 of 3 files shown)


MikeMcQuaid:
Thanks again @alyssais!

PR #1757, opened 2016-12-31
keg: handle dependencies of moved/renamed formulae

alyssais:

-----

In #1497 I switched from `Keg#to_formula` for comparing kegs to formulae to comparing the name and tap in the keg's tab to the name and tap of the formula.

However, this fails to match if the name and tap of the formula have changed since the keg was installed, so it's clearly better to use `Keg#to_formula` where possible, and fall back to the information in the tab when `#to_formula` can't be used.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -126,7 +126,18 @@ class Keg
     end
 
     keg_names = kegs.map(&:name)
-    kegs_by_source = kegs.group_by { |k| [k.name, Tab.for_keg(k).tap] }
+    kegs_by_source = kegs.group_by do |keg|
+      begin
+        # First, attempt to resolve the keg to a formula
+        # to get up-to-date name and tap information.
+        f = keg.to_formula
+        [f.name, f.tap]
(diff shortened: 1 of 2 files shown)


alyssais:
There's a test in #1756 that will fail if both PRs are merged, so once either is merged I'll need to update the other before it can be.

alyssais:
To reproduce the problem this fixes:

1. <kbd>brew install homebrew/science/ccfits</kbd>
2. Edit `ccfits`'s install receipt to change the runtime dependencies to `null` (so the fallback is used)
3. Edit `cfitsio`'s install receipt to change the tap to "homebrew/core" (where it used to be)
4. Check a warning or error message is shown by <kbd>brew uninstall cfitsio</kbd>

MikeMcQuaid:
> There's a test in #1756 that will fail if both PRs are merged, so once either is merged I'll need to update the other before it can be.

@alyssais It was merged.

alyssais:
Should be ready now, assuming CI succeeds.

MikeMcQuaid:
Thanks @alyssais!

PR #1763, opened 2017-01-02
New feature: Allow download from private GitHub repository

minamijoyo:
## Why?
I would like to download the release asset of the private GitHub repository.

I'm using a private brew tap at my company and the prebuild binaries for internal tools are uploaded to the private GitHub Release page.
For such in-house use, S3DownloadStrategy already exists, but most of the source code are managed by GitHub, I think it would be useful to download the binary directly from GitHub.

## What I did
I implemented a new DownloadStrategy to find and download asset using GitHub API.

## How I did it
First, resolve the asset_id associated with the release tag with the following API:
https://developer.github.com/v3/repos/releases/#get-a-release-by-tag-name

and then, download the asset with the following API:
https://developer.github.com/v3/repos/releases/#get-a-single-release-asset

## How to verify it
To use it, add `:using => GitHubReleaseDownloadStrategy` to the URL section of your formula.

```rb
require "formula"

class Hoge < Formula
  url "https://github.com/:owner/:repo/releases/download/:tag/hoge_v0.1.0_darwin_amd64.tar.gz", :using => GitHubReleaseDownloadStrategy
  ...

end
```

```bash
$ export HOMEBREW_GITHUB_API_TOKEN=xxxxxx
$ brew install hoge
```

Please review this.

## Check List

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -532,6 +532,82 @@ class S3DownloadStrategy < CurlDownloadStrategy
   end
 end
 
+# GitHubReleaseDownloadStrategy downloads tarballs from GitHub Release assets.
+# To use it, add ":using => GitHubReleaseDownloadStrategy" to the URL section
+# of your formula. This download strategy uses GitHub access tokens (in the
+# environment variables GITHUB_TOKEN) to sign the request.
+# This strategy is suitable for corporate use just like S3DownloadStrategy,
+# because it lets you use a private GttHub repository for internal distribution.
+# It works with public one, but in that case simply use CurlDownloadStrategy.
+class GitHubReleaseDownloadStrategy < CurlDownloadStrategy
+  require 'open-uri'
+
+  def initialize(name, resource)
+    super
+
+    @github_token = ENV["GITHUB_TOKEN"]
+    unless @github_token
+      puts "Environmental variable GITHUB_TOKEN is required."
+      raise CurlDownloadStrategyError, @url
+    end
+
+    url_pattern = %r|https://github.com/(\S+)/(\S+)/releases/download/(\S+)/(\S+)|
+    unless @url =~ url_pattern
+      puts "Invalid url pattern for GitHub Release."
+      raise CurlDownloadStrategyError, @url
+    end
+
+    _, @owner, @repo, @tag, @filename = *(@url.match(url_pattern))
+  end
+
+  def _fetch
+    puts "Download asset_id: #{asset_id}"
+    # HTTP request header `Accept: application/octet-stream` is required.
+    # Without this, the GitHub API will respond with metadata, not binary.
+    curl asset_url, "-C", downloaded_size, "-o", temporary_path, "-H", 'Accept: application/octet-stream'
+  end
+
+  private
+
+  def asset_url
+    "https://#{@github_token}@api.github.com/repos/#{@owner}/#{@repo}/releases/assets/#{asset_id}"
+  end
+
+  def asset_id
+    @asset_id ||= resolve_asset_id
+  end
+
+  def resolve_asset_id
+    release_metadata = fetch_release_metadata
+    assets = release_metadata["assets"].select{ |a| a["name"] == @filename }
+    if assets.empty?
+      puts "Asset file not found."
+      raise CurlDownloadStrategyError, @url
+    end
+
+    return assets.first["id"]
+  end
+
+  def release_url
+    "https://api.github.com/repos/#{@owner}/#{@repo}/releases/tags/#{@tag}"
+  end
+
+  def fetch_release_metadata
+    begin
+      release_response = open(release_url, {:http_basic_authentication => [@github_token]}).read
+    rescue OpenURI::HTTPError => e
+      if e.message == '404 Not Found'
+        puts "GitHub Release not found."
+        raise CurlDownloadStrategyError, @url
+      else
+        raise e
+      end
+    end
+
+    return JSON.parse(release_response)
+  end
+end
+
 class SubversionDownloadStrategy < VCSDownloadStrategy
   def initialize(name, resource)
     super
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -542,7 +542,7 @@ end
 # works with public one, but in that case simply use CurlDownloadStrategy.
 class GitHubPrivateRepositoryDownloadStrategy < CurlDownloadStrategy
   require "utils/formatter"
-  require 'utils/github'
+  require "utils/github"
 
   def initialize(name, resource)
     super
@@ -551,12 +551,12 @@ class GitHubPrivateRepositoryDownloadStrategy < CurlDownloadStrategy
   end
 
   def parse_url_pattern
-    url_pattern = %r|https://github.com/([^/]+)/([^/]+)/(\S+)|
+    url_pattern = %r{https://github.com/([^/]+)/([^/]+)/(\S+)}
     unless @url =~ url_pattern
       raise CurlDownloadStrategyError, "Invalid url pattern for GitHub Repository."
     end
 
-    _, @owner, @repo, @filepath = *(@url.match(url_pattern))
+    _, @owner, @repo, @filepath = *@url.match(url_pattern)
   end
 
   def download_url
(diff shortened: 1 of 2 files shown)


Library/Homebrew/download_strategy.rb
```diff
@@ -532,6 +532,82 @@ def _fetch
...
 
+# GitHubReleaseDownloadStrategy downloads tarballs from GitHub Release assets.
+# To use it, add ":using => GitHubReleaseDownloadStrategy" to the URL section
+# of your formula. This download strategy uses GitHub access tokens (in the
+# environment variables GITHUB_TOKEN) to sign the request.
+# This strategy is suitable for corporate use just like S3DownloadStrategy,
+# because it lets you use a private GttHub repository for internal distribution.
+# It works with public one, but in that case simply use CurlDownloadStrategy.
+class GitHubReleaseDownloadStrategy < CurlDownloadStrategy
+  require 'open-uri'
```
MikeMcQuaid: This should use `utils/curl` and/or `utils/github` instead.

Library/Homebrew/download_strategy.rb
```diff
@@ -532,6 +532,82 @@ def _fetch
...
+
+  def initialize(name, resource)
+    super
+
+    @github_token = ENV["GITHUB_TOKEN"]
+    unless @github_token
+      puts "Environmental variable GITHUB_TOKEN is required."
```
MikeMcQuaid: Perhaps this string should be part of the `raise` instead?

Library/Homebrew/download_strategy.rb
```diff
@@ -532,6 +532,82 @@ def _fetch
...
+      raise CurlDownloadStrategyError, @url
+    end
+
+    url_pattern = %r|https://github.com/(\S+)/(\S+)/releases/download/(\S+)/(\S+)|
+    unless @url =~ url_pattern
+      puts "Invalid url pattern for GitHub Release."
```
MikeMcQuaid: Perhaps this string should be part of the raise instead?

Library/Homebrew/download_strategy.rb
```diff
@@ -532,6 +532,82 @@ def _fetch
...
+  def initialize(name, resource)
+    super
+
+    @github_token = ENV["GITHUB_TOKEN"]
+    unless @github_token
+      puts "Environmental variable GITHUB_TOKEN is required."
+      raise CurlDownloadStrategyError, @url
+    end
+
+    url_pattern = %r|https://github.com/(\S+)/(\S+)/releases/download/(\S+)/(\S+)|
```
MikeMcQuaid: This should allow https://github.com/Homebrew/brew/archive/1.1.5.tar.gz format links.

Library/Homebrew/download_strategy.rb
minamijoyo: Oh, I did not know such a usuful lib. I changed to use `utils/github`.
I also changed the environment varaible used to this same `HOMEBREW_GITHUB_API_TOKEN` .

Library/Homebrew/download_strategy.rb
minamijoyo: Sure, I changed it.

Library/Homebrew/download_strategy.rb
minamijoyo: I updated it here as well.

Library/Homebrew/download_strategy.rb
minamijoyo: @MikeMcQuaid I am trying to support this, but the following URL does not contain any asset.
https://api.github.com/repos/Homebrew/brew/releases/tags/1.1.5

How do we download it? I could not find an GitHub API to get archive. Are there any good ideas?

Library/Homebrew/download_strategy.rb
```diff
@@ -532,6 +532,63 @@ def _fetch
...
+
+  def resolve_asset_id
+    release_metadata = fetch_release_metadata
+    assets = release_metadata["assets"].select{ |a| a["name"] == @filename }
+    raise CurlDownloadStrategyError, "Asset file not found." if assets.empty?
+
+    return assets.first["id"]
+  end
+
+  def release_url
```
MikeMcQuaid: I reckon you can just make this a variable or inline it rather than a function.

Library/Homebrew/download_strategy.rb
```diff
@@ -532,6 +532,63 @@ def _fetch
...
+    raise CurlDownloadStrategyError, "Environmental variable HOMEBREW_GITHUB_API_TOKEN is required." unless @github_token
+
+    url_pattern = %r|https://github.com/(\S+)/(\S+)/releases/download/(\S+)/(\S+)|
+    raise CurlDownloadStrategyError, "Invalid url pattern for GitHub Release." unless @url =~ url_pattern
+
+    _, @owner, @repo, @tag, @filename = *(@url.match(url_pattern))
+  end
+
+  def _fetch
+    puts "Download asset_id: #{asset_id}"
```
MikeMcQuaid: I think this shouldn't be outputted as it's not necessary for the user to see.

Library/Homebrew/download_strategy.rb
```diff
@@ -532,6 +532,63 @@ def _fetch
...
+# It works with public one, but in that case simply use CurlDownloadStrategy.
+class GitHubReleaseDownloadStrategy < CurlDownloadStrategy
+  require "utils/formatter"
+  require 'utils/github'
+
+  def initialize(name, resource)
+    super
+
+    @github_token = ENV["HOMEBREW_GITHUB_API_TOKEN"]
+    raise CurlDownloadStrategyError, "Environmental variable HOMEBREW_GITHUB_API_TOKEN is required." unless @github_token
```
MikeMcQuaid: Use `unless ... end` so this line is shorter.

MikeMcQuaid: Also, you probably need to handle the case where the token doesn't have e.g. private repository access.

MikeMcQuaid:
> @MikeMcQuaid I am trying to support this format, but the following URL does not contain any asset.
> https://api.github.com/repos/Homebrew/brew/releases/tags/1.1.5
> 
> How do we download it? I could not find an GitHub API to get archive. Are there any good ideas?

@minamijoyo I'm thinking this download strategy could be a lot more general and instead be e.g. `GitHubTokenDownloadStrategy`. You can access private resources like if e.g. https://github.com/Homebrew/brew/archive/1.1.5.tar.gz was private using the same token format you've used but without the API call i.e. just a `curl https://$TOKEN@github.com/Homebrew/brew/archive/1.1.5.tar.gz`. That means we can avoid any API calls or parsing and just allow this download strategy to take a token and pass that to GitHub for downloading any private resource. Thoughts?

minamijoyo:
@MikeMcQuaid Thanks for the review and advice.
First let me share my current understanding.

Certainly the archive url can be accessed with the curl.
`https://$TOKEN@github.com/:owner/:repo/archive/:filename`

But the releases url is not. In the case of the a private repository, even if you set a token, not found is returned. I don't know why there is such a difference. Try it with a private repository.
`https://$TOKEN@github.com/:owner/:repo/releases/download/:tag/:filename`
 
Although the url of assets API can be accessed with the curl, asset_id is not shown on Web,  I don't know how to get asset_id without API.
`https://$TOKEN@api.github.com/repos/:owner/:repo/releases/assets/:asset_id`

In addition, to download the asset's binary content, we must set the Accept header of the request to `application/octet-stream`.

Therefore it is difficult to make it a common implementation without API calls. Is there a way to go something good? If we assume that both logic are implemented respectively, I think that it is better to separate them into different classes, but one is not what I originally wanted to do.

MikeMcQuaid:
@minamijoyo I think it's fine to have API calls if they are needed for releases 👍 but I think this PR should probably handle both cases in order to be useful. The GitHub Releases download strategy could be a subclass from a GitHub Private Repository download strategy to share e.g. at least the token logic.



> But the releases url is not. In the case of the a private repository, even if you set a token, not found is returned. I don't know why there is such a difference.

It may be worth asking support@github.com why this is the case and their thoughts; they will definitely be able to help.

minamijoyo:
@MikeMcQuaid OK. I will try to support both cases. Which of the base class names should be `GitHubTokenDownloadStrategy` or `GitHubPrivateRepositoryDownloadStrategy` ?

Also I will also ask GitHub why there is such a difference.

MikeMcQuaid:
@minamijoyo Thanks! I think `GitHubPrivateRepositoryDownloadStrategy` and `GitHubPrivateRepositoryReleaseDownloadStrategy` are the best names.

minamijoyo:
@MikeMcQuaid OK.  I will do so.

Library/Homebrew/download_strategy.rb
minamijoyo: I've supported both cases.

Library/Homebrew/download_strategy.rb
minamijoyo: Fixed.

Library/Homebrew/download_strategy.rb
minamijoyo: Fixed.

Library/Homebrew/download_strategy.rb
minamijoyo: Fixed.

minamijoyo:
@MikeMcQuaid I've fixed it, so please review again.
In addition, as the class name changed, I've also changed the title of this pull request to a more appropriate name.

minamijoyo:
I contacted GitHub's support and got answers. The summary is as follows:

>> * Is it possible to download private release assets directly with curl without using API?

> No, not currently.

>> * If possible please tell me how to do it.

> It's not possible.

>> * If not, compared to the archive curl, why is there such a difference?

> The fact that it's possible to do for the archive doesn't mean it will be possible for everything else -- you should not have such expectations and I don't believe we'll support that for releases anytime soon. To clarify a bit why those two are different -- those two things are stored differently and go through different paths. In general, you should be using the API for all automation purposes.

After all, we can not get the private release assets without API.

MikeMcQuaid:
Fantastic work @minamijoyo. Thanks for your first contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

@joscha @stefansundin This should help you with your use-cases.

joscha:
Nice work @minamijoyo!

minamijoyo:
@MikeMcQuaid Thank you for review :beer:

Issue #1770, opened 2017-01-03
`brew upgrade` misbehaves when a formula's dependency is renamed


cartr:
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [X] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

Here's my [`brew config` and `brew doctor` output](https://gist.github.com/cartr/5d713e151cfa7ecdbfd56b3ca6e76dbd).

### Summary

If a formula (that is depended on by another installed formula) is renamed, `brew update` tries to install it on top of the old one and fails during the `brew link` step.

### Steps to Reproduce

Run these commands:

```
brew tap cartr/rename-bug-demo
brew install dependent
```

You'll get two new formulae installed: `somepackage` (which is GNU Hello) and `dependent` which depends on it. Next, run these to switch to a branch where `somepackage` has been renamed to `somepackage4` and upgrade:

```
cd /usr/local/Homebrew/Library/Taps/cartr/homebrew-rename-bug-demo
git config --add remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*" && git fetch
git checkout add-four
brew upgrade
```

### Expected Result

The upgrade completes successfully, migrating `somepackage` to `somepackage4`.

### Actual Result

```
bash-3.2$ brew upgrade
Warning: You are using a pre-release version of Xcode.
You may encounter build failures or other breakages.
Please create pull-requests instead of filing issues.
==> Upgrading 1 outdated package, with result:
cartr/rename-bug-demo/somepackage4 2.10_1
==> Upgrading cartr/rename-bug-demo/somepackage4
==> Downloading http://ftpmirror.gnu.org/hello/hello-2.10.tar.gz
==> Downloading from http://mirrors.kernel.org/gnu/hello/hello-2.10.tar.gz
######################################################################## 100.0%
==> ./configure --disable-silent-rules --prefix=/usr/local/Cellar/somepackage4/2.10_1
==> make install
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/hello
Target /usr/local/bin/hello
is a symlink belonging to somepackage. You can unlink it:
  brew unlink somepackage

To force the link and overwrite all conflicting files:
  brew link --overwrite somepackage4

To list all files that would be deleted:
[...]

PR #1772, opened 2017-01-03
upgrade: unlink kegs of renamed formulae

alyssais:

-----

This is a solution to #1770.

Because this change refactors some alias-resolving logic, I re-ran all the test commands I used in #971 to make sure that behaviour was still correct.

This PR changes the way <kbd>brew upgrade</kbd> finds linked kegs to be unlinked. Instead of looking for the linked keg based on the current name of the formula, it loops through all linked kegs to find any that were installed with the formula under any name. For efficiency (to avoid iterating all linked kegs for every formula) the list of linked kegs is generated for all formulae to be uninstalled in the same iteration and stored in an instance variable.

While this fixes the problem in #1770, I can see kegs installed with previous names of formulae causing lots more problems in future, because a lot of the code in `Formula` assumes that kegs can be found on the filesystem based on the current name of the formula. So, I'm wondering if, instead of working around that in this one specific case, it might be worth changing `brew upgrade` to migrate any kegs of renamed formulae?

### DIFF
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -75,6 +75,8 @@ module Homebrew
       puts pinned.map { |f| "#{f.full_specified_name} #{f.pkg_version}" } * ", "
     end
 
+    @linked_kegs_by_formula = linked_kegs_by_formula(formulae_to_install)
+
     formulae_to_install.each do |f|
       upgrade_formula(f)
       next unless ARGV.include?("--cleanup")
@@ -88,11 +90,7 @@ module Homebrew
   end
 
   def upgrade_formula(f)
-    formulae_maybe_with_kegs = [f] + f.old_installed_formulae
-    outdated_kegs = formulae_maybe_with_kegs
-                    .map(&:linked_keg)
-                    .select(&:directory?)
-                    .map { |k| Keg.new(k.resolved_path) }
+    outdated_kegs = @linked_kegs_by_formula[f]
 
     fi = FormulaInstaller.new(f)
     fi.options             = f.build.used_options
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
> This PR changes the way brew uninstall finds linked kegs to be unlinked.

Can you elaborate a bit more on this; I'm not seeing any `uninstall` changes here?



> So, I'm wondering if, instead of working around that in this one specific case, it might be worth changing brew upgrade to migrate any kegs of renamed formulae?

`brew update` does migrate kegs of renamed formulae so I'm not entirely convinced this is an issue that can be exposed to users rather than e.g. developers who haven't setup a tap migration properly.

alyssais:
> Can you elaborate a bit more on this; I'm not seeing any `uninstall` changes here?

Meant `upgrade`, sorry. I'll update the PR body.

alyssais:
> `brew update` does migrate kegs of renamed formulae so I'm not entirely convinced this is an issue that can be exposed to users rather than e.g. developers who haven't setup a tap migration properly.

That seems reasonable. I did have to disable auto-update to test this.

MikeMcQuaid:
I think there may be some rom for improvement here around this but will wait and see if there's a few more issues first.

vladshablinsky:
I think it's not the best solution here, because what happens in `upgrade` for renamed formula is just ignoring the rename and install formula with new name as some other formula which has same binaries. No rename is detected here. Still, this is a solution, because what we do on migrations is unlinking old name, moving it to the new cellar and then linking, right?

I think this solution leaves us with `oldname` and `newname` installed, no additional links created and we can't even migrate `oldname` since `newname` already exist. Which btw won't be the case if we merge #2359.

Issue #1764, opened 2017-01-02
Feature Request: display a check mark to the installed packages when searching


Clumsy-Coder:
### Propose a feature:

Displaying a check mark to the packages installed when searching for a package (including brew cask). 
Ex: you installed google chrome and you type 'brew search google' and a list of packages are listed (like usual), but the installed packages have a check mark next to it. In this case google chrome package will have a checkmark indicating the package is installed and the user doesn't need to type 'brew cask info google-chrome' to check if the package is installed or not (if the user wants to check later on).

MikeMcQuaid:
This seems like a good idea.

vitorgalvao:
This has been implemented.

PR #1774, opened 2017-01-03
[Advice needed] search: show checkmark next to installed casks

alyssais:

-----

I'm trying to implement #1764.

To do that, I've changed <kbd>brew search</kbd> to search local casks as well as formulae. However, for this to work I needed to use some cask code, which wasn't possible.

@MikeMcQuaid suggested I refactor whatever Cask I needed into Homebrew, but I'm not sure how to do that. Really what I need is something like `Hbc.all_tokens`, but there's no sensible place to put it. One option would be to add a `Cask` class to Homebrew that had some very basic metadata like `name` and `tap`, but that seems like repeating ourselves when the code's already there in Cask.

So, I'm looking for advice on how to refactor the Cask functionality I need to use here into Homebrew.

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -10,6 +10,7 @@ raise "Homebrew must be run under Ruby 2!" unless RUBY_TWO
 
 require "pathname"
 HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
+$:.unshift(HOMEBREW_LIBRARY_PATH.join("cask", "lib").to_s)
 $:.unshift(HOMEBREW_LIBRARY_PATH.to_s)
 require "global"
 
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
I think it's probably worth punting on showing checkmarks for Casks for now. When we do do so I think it's probably worth refactoring some of the logic to detect if a Cask is installed into shared code in `Library/Homebrew` rather than having Homebrew call into the Homebrew Cask code/modules. Another option would be to call out to `brew cask search`.

CC @reitermarkus and @jawshooah for thoughts.

reitermarkus:
> One option would be to add a `Cask` class to Homebrew that had some very basic metadata like name and tap, but that seems like repeating ourselves when the code's already there in Cask.

I thought about refactoring the whole `Hbc::Cask` class before to behave more like the `Formula` class, i. e. being `Enumerable` and letting you call `Cask["token"]` instead of `Hbc.load("token")`.

MikeMcQuaid:
We could also have a mixin or proxy to abstract formulae and casks in a case like this.

dolfly (approved):

vitorgalvao:
Looks like this has already been implemented. I seem to distinctly recall a PR to do it, but I’m not finding it. Either way, just tested it and it works.

PR #1750, opened 2016-12-30
keg: fall back for dependencies of buggy tabs

alyssais:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Fixes #1554.

If a keg was installed with Homebrew < 1.1.6 (which had bugs when populating the `runtime_dependencies` field in the install receipt), just treat it as if `runtime_dependencies` wasn't present in the tab and use the formula-based fallback.

To test:

1. `brew install hello`
2. Modify `hello`'s install receipt to add another installed formula as a dependency, e.g. `{ "full_name": "autoconf", "version": "2.14" }`.
3. Attempt to uninstall that formula e.g. `brew uninstall autoconf`. Verify that no warning or error is produced (because the information in the receipt isn't trusted).
4. Reinstall it: `brew install autoconf`
5. Modify `hello`'s install receipt to set `homebrew_version` to `1.1.6` or greater. You might want to try with different variations of `git describe --tags --dirty`, e.g. `1.1.6-1-gabcd12345` or `1.1.6-2-54321fedbca-dirty`.
6. Attempt to uninstall the dependency again e.g. `brew uninstall autoconf`. Verify that there is now a warning (or error if you're not a `HOMEBREW_DEVELOPER`).

PR #1777, opened 2017-01-04
tests: remove temporary tab homebrew_version override

alyssais:

***

This had to be added in #1750 to work around special-casing for tabs generated with Homebrew versions < 1.1.6. Now that 1.1.6 is the current version, we can remove this hack.

I set `homebrew_version` in `Tab.empty` so the [`#runtime_dependencies` override](https://github.com/Homebrew/brew/blob/fc11f633b29e6a2c9c4d49d5fd5452549b328193/Library/Homebrew/tab.rb#L253) isn't triggered (which would force `runtime_dependencies` to default to `nil` – unexpected behaviour that would cause a test to fail).

### DIFF
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -168,6 +168,7 @@ class Tab < OpenStruct
 
   def self.empty
     attributes = {
+      "homebrew_version" => HOMEBREW_VERSION,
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Thanks for the cleanup 🙇

PR #1782, opened 2017-01-04
Fix abv and prefix_linked?

ilovezfs:
CC @tdsmith @MikeMcQuaid

Tim suggested we may want to realpath all pathnames before computing
abv, but in case that's not always universally desirable, this is a
slightly less invasive approach.

I also noticed prefix_linked? was returning lies, so there's a fix for
that as well.

Fixes https://github.com/Homebrew/brew/issues/1791.

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -26,11 +26,17 @@ module DiskUsageExtension
   private
 
   def compute_disk_usage
-    if directory?
+    path = if symlink?
+      resolved_path
+    else
+      self
+    end
+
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Thanks for the PR @ilovezfs 🎉 

> Tim suggested we may want to realpath all pathnames before computing
> abv, but in case that's not always universally desirable, this is a
> slightly less invasive approach.

I think in this case I agree with @tdsmith given it's a mixin module we're providing we do want this functionality. I've made this and some other tweaks and pushed to this branch.

ilovezfs:
LGTM

PR #1784, opened 2017-01-05
uses: properly handle untapped formulae in recursive dependency expansion

### LINKED ISSUE #1776: `brew uses` with `--include-optional` doesn't return used formula

- [x] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

### Bug reports:

When I run `brew uses` with `--include`, it doesn't find a formula that is actually used.

```
$ brew uses --installed --include-build --recursive unixodbc
homebrew/php/php70

$ brew uses --installed --include-build --include-optional --recursive unixodbc

$
```

Notice that unixodc is used by php70, but when I run the `uses` command with `--include-optional`, the command no longer shows php70 as using unixodbc.

`brew config`:

```
$ brew config
HOMEBREW_VERSION: 1.1.6-2-gfc11f633b2
HEAD: fc11f633b29e6a2c9c4d49d5fd5452549b328193
Last commit: 3 hours ago
Core tap HEAD: 0a6cce09b2ff3e50128d2e0d3887b12f5bf78368
Core tap last commit: 3 hours ago
HOMEBREW_PREFIX: /usr/local
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.11.0 => /usr/local/bin/git
Perl: /usr/local/bin/perl => /usr/local/Cellar/perl/5.24.0_1/bin/perl
Ruby: /usr/local/bin/ruby => /usr/local/Cellar/ruby/2.4.0/bin/ruby
Java: 1.8.0_65, 1.6.0_65-b14-468
macOS: 10.11.6-x86_64
Xcode: 8.1
CLT: 8.2.0.0.1.1480973914
X11: 2.7.9 => /opt/X11
```

`brew doctor` (I can't upgrade Xcode because I'm still on El Capitan):

```
$ brew doctor

Warning: Found Ruby file outside beeftornado/rmtree tap formula directory
(/usr/local/Homebrew/Library/Taps/beeftornado/homebrew-rmtree/Formula):
  /usr/local/Homebrew/Library/Taps/beeftornado/homebrew-rmtree/brew-rmtree.rb

Warning: Your Xcode (8.1) is outdated.
Please update to Xcode 8.2 (or delete it).
```

alyssais:

-----

Fixes #1776.

If any known formula had a dependency on an untapped tap, `Formula#recursive_dependencies` would throw an exception, which would be caught by the outer exception handler, causing the rest of the dependencies for that formula to be skipped and incomplete output to be generated.

To fix this, I added a check to avoid analysing the dependencies of formulae from uninstalled taps.

Additionally, I removed the aforementioned outer exception handler added in 5fdb89a, because the only other place that should be capable of throwing such an exception is the
statement that was surrounded by another wider exception handler in Homebrew/legacy-homebrew#40682. I verified that the example failures in https://github.com/Homebrew/legacy-homebrew/issues/40636 have still not returned even with this change.

Also, `TapDependency#tap` previously returned a `String`, but a `Tap` instance seemed more sensible and useful, so I changed it. I couldn't find anywhere this method was actually used, so the change
shouldn't break anything.

### DIFF
--- a/Library/Homebrew/cmd/uses.rb
+++ b/Library/Homebrew/cmd/uses.rb
@@ -47,48 +47,49 @@ module Homebrew
 
     uses = formulae.select do |f|
       used_formulae.all? do |ff|
-        begin
-          if recursive
-            deps = f.recursive_dependencies do |dependent, dep|
-              if dep.recommended?
-                Dependency.prune if ignores.include?("recommended?") || dependent.build.without?(dep)
-              elsif dep.optional?
-                Dependency.prune if !includes.include?("optional?") && !dependent.build.with?(dep)
-              elsif dep.build?
-                Dependency.prune unless includes.include?("build?")
-              end
+        if recursive
+          deps = f.recursive_dependencies do |dependent, dep|
+            if dep.recommended?
+              Dependency.prune if ignores.include?("recommended?") || dependent.build.without?(dep)
+            elsif dep.optional?
+              Dependency.prune if !includes.include?("optional?") && !dependent.build.with?(dep)
+            elsif dep.build?
+              Dependency.prune unless includes.include?("build?")
             end
-            reqs = f.recursive_requirements do |dependent, req|
-              if req.recommended?
-                Requirement.prune if ignores.include?("recommended?") || dependent.build.without?(req)
-              elsif req.optional?
-                Requirement.prune if !includes.include?("optional?") && !dependent.build.with?(req)
-              elsif req.build?
-                Requirement.prune unless includes.include?("build?")
-              end
-            end
-          else
-            deps = f.deps.reject do |dep|
-              ignores.any? { |ignore| dep.send(ignore) } && !includes.any? { |include| dep.send(include) }
-            end
-            reqs = f.requirements.reject do |req|
-              ignores.any? { |ignore| req.send(ignore) } && !includes.any? { |include| req.send(include) }
-            end
-          end
-          next true if deps.any? do |dep|
-            begin
-              dep.to_formula.full_name == ff.full_name
-            rescue
-              dep.name == ff.name
-            end
-          end
 
-          reqs.any? do |req|
-            req.name == ff.name || [ff.name, ff.full_name].include?(req.default_formula)
+            # If a tap isn't installed, we can't find the dependencies of one
+            # its formulae, and an exception will be thrown if we try.
+            if dep.is_a?(TapDependency) && !dep.tap.installed?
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cmd/uses.rb
```diff
@@ -47,48 +47,49 @@ def uses
...
-              ignores.any? { |ignore| dep.send(ignore) } && !includes.any? { |include| dep.send(include) }
-            end
-            reqs = f.requirements.reject do |req|
-              ignores.any? { |ignore| req.send(ignore) } && !includes.any? { |include| req.send(include) }
+
+            # If a tap isn't installed, we can't find the dependencies of one
+            # its formulae, and an exception will be thrown if we try.
+            if dep.is_a?(TapDependency) && !dep.tap.installed?
+              Dependency.keep_but_prune_recursive_deps
             end
```
alyssais: Because I removed an enclosing block and outdented its contents, the diff is a bit messy, so I'll highlight my actual code changes here.

This is where I added the check to avoid the exception when a dependency is untapped.

Library/Homebrew/cmd/uses.rb
```diff
@@ -47,48 +47,49 @@ def uses
...
+          begin
+            dep.to_formula.full_name == ff.full_name
+          rescue
+            dep.name == ff.name
+          end
+        end
+
+        reqs.any? do |req|
+          req.name == ff.name || [ff.name, ff.full_name].include?(req.default_formula)
         end
```
alyssais: This is where I removed the `rescue FormulaUnavailableError` that is no longer necessary.

alyssais:
Not sure what to do about the coverage. There's already an integration test and all I did to most of the diff was change its indentation.

Library/Homebrew/cmd/uses.rb
MikeMcQuaid: For anyone else following along: https://github.com/Homebrew/brew/pull/1784/files?w=1 makes it a lot easier to follow.

MikeMcQuaid:
Great work again @alyssais!

Library/Homebrew/cmd/uses.rb
MikeMcQuaid: Unfortunately it looked like it still was so need to revert this 😭

PR #1786, opened 2017-01-05
PerlRequirement: Perl version string may lack parentheses

rwhogg:
There may be no parentheses around the version number. For example:
This is perl, v5.10.1 (*) built for x86_64-linux-thread-multi

-----

Note that this is a straight cherry-pick from Linuxbrew (excluding a slightly modified commit message), but might affect macOS too in certain situations (eg. using Perl 5.10.1 from perl-build).

### DIFF
--- a/Library/Homebrew/requirements/perl_requirement.rb
+++ b/Library/Homebrew/requirements/perl_requirement.rb
@@ -10,7 +10,7 @@ class PerlRequirement < Requirement
 
   satisfy(build_env: false) do
     which_all("perl").detect do |perl|
-      perl_version = Utils.popen_read(perl, "--version")[/\(v(\d+\.\d+)(?:\.\d+)?\)/, 1]
+      perl_version = Utils.popen_read(perl, "--version")[/v(\d+\.\d+)(?:\.\d+)?/, 1]
       next unless perl_version
       Version.create(perl_version.to_s) >= Version.create(@version)
     end


MikeMcQuaid:
Will this not fail if it does have parenthesis? This isn't making it optional, it's removing them entirely.

rwhogg:
> 

No, it won't fail. For a Perl version with parentheses, the result is the same as it was with the old regex:

```bash
irb(main):002:0> Utils.popen_read("perl", "--version")
=> "\nThis is perl 5, version 18, subversion 2 (v5.18.2) built for darwin-thread-multi-2level\n(with 2 registered patches, see perl -V for more detail)\n\nCopyright 1987-2013, Larry Wall\n\nPerl may be copied only under the terms of either the Artistic License or the\nGNU General Public License, which may be found in the Perl 5 source kit.\n\nComplete documentation for Perl, including FAQ lists, should be found on\nthis system using \"man perl\" or \"perldoc perl\".  If you have access to the\nInternet, point your browser at http://www.perl.org/, the Perl Home Page.\n\n"
irb(main):003:0> Utils.popen_read("perl", "--version")[/v(\d+\.\d+)(?:\.\d+)?/, 1]
=> "5.18"
irb(main):004:0> Utils.popen_read("perl", "--version")[/\(v(\d+\.\d+)(?:\.\d+)?\)/, 1]
=> "5.18"
```

MikeMcQuaid:
@rwhogg Yes, of course, my brain wasn't working right the first time! Thanks @rwhogg and @sjackman!

sjackman:
If you haven't yet encountered Regexper to visualize regular expressions, it's absolutely fantastic:
https://regexper.com/#v(%5Cd%2B%5C.%5Cd%2B)(%3F%3A%5C.%5Cd%2B)%3F

MikeMcQuaid:
Oh, neat! In this case I missed that this wasn't doing e.g. `perl (1.2.3)` vs. `perl 1.2.3` but was just `(1.2.3)` vs. `1.2.3`.

PR #1792, opened 2017-01-06
Whitelist PHP formulas in shadowed header audit check

javian:

-----
Instead of having to renaming the include directory in the PHP releases it was suggested by @MikeMcQuaid  that we whitelist the php formulas in https://github.com/Homebrew/homebrew-php/pull/3893

### DIFF
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -6,6 +6,8 @@ module FormulaCellarChecks
       formula.name.start_with?(formula_name)
     end
 
+    return if formula.name =~ /^php\d+$/
+
     return if MacOS.version < :mavericks && formula.name.start_with?("postgresql")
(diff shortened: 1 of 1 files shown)


Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -6,6 +6,8 @@ def check_shadowed_headers
       formula.name.start_with?(formula_name)
     end
 
+    return if formula.name =~ /^php[57]\d$/
```
MikeMcQuaid: `return if formula.name =~ /^php\d\d$/` may be more future proof?

bfontaine: Or even `/^php\d+$/`?

javian: I'm thinking it will change if we move into the new versioned model with php@... as well. I just wanted it to be as specific as it could be but I can change it and its likely that it won't cause any clashes anyhow. I'll change it.

PR #1797, opened 2017-01-07
Revert "uses: properly handle untapped formulae in recursive dependency expansion"

MikeMcQuaid:
Reverts Homebrew/brew#1784

CC @ilovezfs @alyssais

### DIFF
--- a/Library/Homebrew/cmd/uses.rb
+++ b/Library/Homebrew/cmd/uses.rb
@@ -47,49 +47,48 @@ module Homebrew
 
     uses = formulae.select do |f|
       used_formulae.all? do |ff|
-        if recursive
-          deps = f.recursive_dependencies do |dependent, dep|
-            if dep.recommended?
-              Dependency.prune if ignores.include?("recommended?") || dependent.build.without?(dep)
-            elsif dep.optional?
-              Dependency.prune if !includes.include?("optional?") && !dependent.build.with?(dep)
-            elsif dep.build?
-              Dependency.prune unless includes.include?("build?")
+        begin
+          if recursive
+            deps = f.recursive_dependencies do |dependent, dep|
+              if dep.recommended?
+                Dependency.prune if ignores.include?("recommended?") || dependent.build.without?(dep)
+              elsif dep.optional?
+                Dependency.prune if !includes.include?("optional?") && !dependent.build.with?(dep)
+              elsif dep.build?
+                Dependency.prune unless includes.include?("build?")
+              end
             end
+            reqs = f.recursive_requirements do |dependent, req|
+              if req.recommended?
+                Requirement.prune if ignores.include?("recommended?") || dependent.build.without?(req)
+              elsif req.optional?
+                Requirement.prune if !includes.include?("optional?") && !dependent.build.with?(req)
+              elsif req.build?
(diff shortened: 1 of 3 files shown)


alyssais:
Oh no! When did that exception handler get triggered?

ilovezfs:
@alyssais https://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/6759/version=el_capitan/console I think it's triggering here.

Library/Homebrew/cmd/uses.rb
```diff
@@ -47,49 +47,48 @@ def uses
...
+              elsif dep.build?
+                Dependency.prune unless includes.include?("build?")
+              end
             end
-
-            # If a tap isn't installed, we can't find the dependencies of one
-            # its formulae, and an exception will be thrown if we try.
-            if dep.is_a?(TapDependency) && !dep.tap.installed?
-              Dependency.keep_but_prune_recursive_deps
+            reqs = f.recursive_requirements do |dependent, req|
```
alyssais: This is where the exception is thrown. The correct behaviour, here, I think, would be to have a way to get `Requirement.expand` to ignore unavailable formulae. (With the current behaviour, if a formula has a dependency of any kind on any untapped formula, it isn't considered for `uses` at all.)

MikeMcQuaid: Sounds sensible. I think for now leaving the `rescue` is worthwhile.

alyssais: Is there any way to tell if the handler is ever hit? I don't like the idea of it being there unnecessarily, or being triggered unexpectedly — in this case, it silently caused incomplete results, whereas allowing the exceptions would have quickly revealed the bugs.

Some options I thought of off the top of my head:

- Capture the exceptions, but convert them to warnings (for developers only???)
- Remove the handler, but test `brew uses` against every formula in the official taps with every combination of arguments first (I'll happily run this)

MikeMcQuaid: You could reraise if an environment variable is set or warn for developers only. 



> Remove the handler, but test brew uses against every formula in the official taps with every combination of arguments first (I'll happily run this)

Sounds like a good idea.

alyssais: After thinking about this for a bit, it's not such a great idea, because e.g. in the issue that this fixed, the issue was that a formula wasn't tapped, so a script that ran through all Homebrew/homebrew-* formulae likely wouldn't have run into it.

Is there a way to use analytics to see if the `rescue` block is ever hit? That way we could monitor it for a month or whatever and remove it if it's never hit.

If not, I'll add a warning for developers encouraging them to open an issue if they see it. After a while, if nobody reports anything, we could enable the warning for everybody, and then finally if nobody has still reported anything we could remove the `rescue`.

MikeMcQuaid: > If not, I'll add a warning for developers encouraging them to open an issue if they see it. After a while, if nobody reports anything, we could enable the warning for everybody, and then finally if nobody has still reported anything we could remove the rescue.

This sounds good except: test it on all of the tapped Homebrew/* taps, make it reraise for developers rather than warn and then we'll remove it, yeh 👍

PR #1801, opened 2017-01-07
keg_relocate: Check HOMEBREW_TEMP's realpath when excluding name changes

woodruffw:

-----

Since `/tmp` (the default `HOMEBREW_TEMP`) is a symlink to `/private/tmp`,
some build systems (like Parrot's) will attempt to use the realpath
instead of the literal `/tmp` we supply it with. This breaks the relocation
code, which only tested the literal `HOMEBREW_TEMP` and not its realpath.

Fixes parrot bottling in Homebrew/homebrew-core#6408.

Ping @ilovezfs

### DIFF
--- a/Library/Homebrew/extend/os/mac/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/mac/keg_relocate.rb
@@ -6,7 +6,9 @@ class Keg
 
         each_install_name_for(file) do |bad_name|
           # Don't fix absolute paths unless they are rooted in the build directory
-          next if bad_name.start_with?("/") && !bad_name.start_with?(HOMEBREW_TEMP.to_s)
+          next if bad_name.start_with?("/") &&
+                  !bad_name.start_with?(HOMEBREW_TEMP.to_s) &&
+                  !bad_name.start_with?(HOMEBREW_TEMP.realpath.to_s)
 
           new_name = fixed_name(file, bad_name)
           change_install_name(bad_name, new_name, file) unless new_name == bad_name


ilovezfs:
This also fixed open-zwave. https://bot.brew.sh/job/Homebrew%20Sierra%20Testing/1185/

Thanks!

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -6,7 +6,9 @@ def fix_dynamic_linkage
 
         each_install_name_for(file) do |bad_name|
           # Don't fix absolute paths unless they are rooted in the build directory
-          next if bad_name.start_with?("/") && !bad_name.start_with?(HOMEBREW_TEMP.to_s)
+          next if bad_name.start_with?("/") &&
+                  !bad_name.start_with?(HOMEBREW_TEMP.to_s) &&
+                  !bad_name.start_with?(HOMEBREW_TEMP.realpath.to_s)
```
MikeMcQuaid: Will this not :boom: if `HOMEBREW_TEMP` isn't a symlink?

ilovezfs: you can call realpath on anything

MikeMcQuaid: Yeh, forgot it'll only 💥 on bad symlinks in which case you'll have other problems.

ilovezfs: Ah good to know. So `readlink` requires a symlink but can read a bad link, and `realpath` will accept non-symlinks, but dies on a bad link. Sometimes I wonder if these functions are out to get us.

MikeMcQuaid: (I was maybe thinking of `readlink`...). Yes, they definitely are.

PR #1828, opened 2017-01-12
audit: Check for invocations of cctools.

woodruffw:

-----

Now that ruby-macho is available, cctools should be invoked as little
as possible.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -1039,6 +1039,10 @@ class FormulaAuditor
       problem "'fails_with :llvm' is now a no-op so should be removed"
     end
 
+    if line =~ /system\s+['"](otool)|(install_name_tool)|(lipo)/
+      problem "Use ruby-macho instead of calling #{$1}"
+    end
+
     if formula.tap.to_s == "homebrew/core"
       ["OS.mac?", "OS.linux?"].each do |check|
         next unless line.include?(check)


woodruffw:
This shouldn't be merged until I revendor ruby-macho at least one more time, since the changes that add `lipo(1)`-like functionality aren't in the currently vendored version.

MikeMcQuaid:
> This shouldn't be merged until I revendor ruby-macho at least one more time, since the changes that add lipo(1)-like functionality aren't in the currently vendored version.

@woodruffw Any news on this?

woodruffw:
Yeah, I implemented the necessary features in ruby-macho a few months ago - I just haven't done a 1.1 release, which when revendored would include the changes.

I'm currently in the mountains with limited internet, but I can do one next week :smile:

MikeMcQuaid:
@woodruffw Cool, thanks!

woodruffw:
Just bumped ruby-macho to 1.1.0, I'll have a revendor PR ready soon and then this'll be good to go :tada:

woodruffw:
This should now be good to go. I'll merge it tonight, unless someone else feels like doing it first :smile:

ilovezfs:
:confetti_ball:

ilovezfs:
@woodruffw looks like this needs some kind of exception for `cctools` itself:
```
iMac-TMP:~ joe$ brew audit --strict --online cctools
cctools:
  * Use ruby-macho instead of calling 
Error: 1 problem in 1 formula
iMac-TMP:~ joe$ 
```

ilovezfs:
@woodruffw also the tool name not being printed isn't specific to cctools:
```
chicken:
  * Use ruby-macho instead of calling 
Error: 1 problem in 1 formula
iMac-TMP:~ joe$
```

woodruffw:
Yeah, the check should be something like :

```ruby
if line =~ /system\s+['"](otool|install_name_tool|lipo)/ && formula.name != "cctools"
```

to keep to capture all in `$1` and exclude `cctools`. Should I revert-and-fix or just update it in another PR?

ilovezfs:
No need for revert.

fxcoudert:
`wine` calls `lipo` like this:

```
        system "lipo", "-create", "#{dirs.first}/openssl",
                                  "#{dirs.last}/openssl",
                       "-output", "#{libexec}/bin/openssl"
```

What would be the ruby-macho equivalent?

ilovezfs:
I think you want
```
          MachO::Tools.merge_machos("#{libexec}/bin/openssl",
                                    "#{dirs.first}/openssl", 
                                    "#{dirs.last}/openssl")
```

woodruffw:
Yep, `merge_machos` is what you want. The docs are here: http://www.rubydoc.info/gems/ruby-macho/MachO/Tools#merge_machos-class_method, although I just realized there's an error: the `files` are strings like @ilovezfs showed, not `MachOFile`/`FatFile` instances like the doc says.

fxcoudert:
@woodruffw can you confirm that it will work the same way for dylibs and executables?

woodruffw:
@fxcoudert Yes. All `merge_machos` does is take *N* Mach-O binaries, flatten them down to single architectures, and then smush all of those together with a fat Mach-O header. It doesn't care about what kind of Mach-O they are (dylib, executable, object file, etc).

fxcoudert:
@woodruffw thanks!

woodruffw:
No problem :+1:

MikeMcQuaid:
@woodruffw Feature request: a `lipo` drop-in written in Ruby 😉

woodruffw:
@MikeMcQuaid That would actually be pretty easy to do, although having Homebrew use it would probably be a pretty ugly hack onto `$PATH` when the real `lipo` isn't available. We also benefit a fair bit in terms of performance from not having to `exec` in the current form :smile:

ilovezfs:
I'm not sure you two are on the same page. @MikeMcQuaid could you be more explicit about what you're asking for please.

MikeMcQuaid:
This isn't a feature request for Homebrew but just in general for Ruby MachO having command-line tools that can be used instead of e.g. `lipo`, `install_name_tool`, `otool` but produce identical output could make the tool more widely useful.

woodruffw:
Yeah, that's something I can do. I recently broke https://github.com/woodruffw/ruby-macho-utils off from the main repo to experiment with command-line tools (since I wanted to add extra dependencies that ruby-macho itself doesn't require), so I can implement `lipo`, etc over there.

PR #1835, opened 2017-01-13
audit: whitelist unstable versions already in core

ilovezfs:
also allow higher stable versions with the same version prefix

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -678,11 +678,47 @@ class FormulaAuditor
       end
     end
 
+    unstable_whitelist = %w[
+      aalib 1.4rc5
+      automysqlbackup 3.0-rc6
+      aview 1.3.0rc1
+      distcc 3.2rc1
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -678,11 +678,40 @@ def audit_specs
...
+      pwnat 0.3-beta
+      pxz 4.999.9
+      recode 3.7-beta2
+      speexdsp 1.2rc3
+      sqoop 1.4.6
+      tcptraceroute 1.5beta7
+      testssl 2.8rc3
+      tiny-fugue 5.0b8
+      vbindiff 3.0_beta4
+    ].each_slice(2).to_a
```
MikeMcQuaid: I wonder if it's worth allowing these to be higher stable versions with the same prefix e.g. `aalib-1.4r6` feels like something that's 🆗 but not `aalib-1.5alpha`

ilovezfs:
@MikeMcQuaid PR refreshed.

PR #1839, opened 2017-01-13
Add docs on persisting settings across upgrades

justinmayer:
There are some formulas in which Homebrew symlinks configuration files
into /usr/local/etc/ instead of copying them, resulting in custom
settings being overwritten on package upgrades. This addition to the
documentation illustrates one way to avoid this behavior by ensuring the
configure script is told where to copy the configuration file(s).

Refs https://github.com/Homebrew/homebrew-core/issues/8711

### DIFF
--- a/docs/Formula-Cookbook.md
+++ b/docs/Formula-Cookbook.md
@@ -783,6 +783,8 @@ The symlinks created by `install_symlink` are guaranteed to be relative. `ln_s` 
 
 For example, Ruby 1.9’s gems should be installed to `var/lib/ruby/` so that gems don’t need to be reinstalled when upgrading Ruby. You can usually do this with symlink trickery, or *better* a configure option.
 
+Another example would be configuration files that should not be overwritten on package upgrades. If after installation you find that to-be-persisted configuration files are not copied but instead *symlinked* into `/usr/local/etc/` from the Cellar, this can often be rectified by passing an appropriate argument to the package’s configure script. That argument will vary depending on a given package’s configure script and/or Makefile, but one example might be: `--sysconfdir=#{etc}`
+
 ### launchd plist files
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
I'm a bit confused here. Aren't those formula bugs?

justinmayer:
@ilovezfs: I imagine that's a matter of perspective. When you updated the dnscrypt-proxy formula to version 1.8 via https://github.com/homebrew/homebrew-core/commit/957afb5787aff88ccad6f4f67fd1406de66fea4c, you perhaps did not realize that there was a new configuration file — or that it and its example sibling would be symlinked into `/usr/local/etc/` unless changes were made to the `configure` invocation. Was that a formula bug? Personally, I'm not sure that's a useful way to look at it in this context.

The formula docs ostensibly exist to assist would-be contributors in authoring, fixing, and/or improving Homebrew packages.  In an effort to support dnscrypt-proxy's new configuration file, I spent hours trying to understand why the aforementioned symlinks were being created and what the optimal solution to the problem should be. I found nothing in any Homebrew-related documentation that described how and why the symlinks were being created nor how to properly prevent their creation. Having stumbled upon a solution mainly through trial and error, I thought it only appropriate to ensure that future contributors might be spared that frustration.

If there is a better way to improve the documentation to clarify this topic for future contributors, I'm certainly open to suggestions.

MikeMcQuaid:
I think this is something that should probably be part of `brew audit` instead as I also agree it's a formula bug. Would that be something you'd be able to address @justinmayer? If not, I'll give it a go.

justinmayer:
Thanks for the input, Mike. Regarding changes to `brew audit`, I think it would be best to defer to your expertise. I've no experience with Ruby and scarcely understand the problem domain enough to even know what to look for in the audit process.

MikeMcQuaid:

PR #1849, opened 2017-01-15
tests: test Formula#recursive_requirements

alyssais:

-----

I was hacking on `Requirement.expand` recently for an improvement I've since decided to handle a different way, but as part of that abandoned patch I wrote some tests for `Requirement.expand` (via `Formula#recursive_requirements`) because I wanted to make sure I didn't break anything and I noticed there currently weren't any.

So, here are some free tests. :D

### DIFF
--- a/Library/Homebrew/test/formula_test.rb
+++ b/Library/Homebrew/test/formula_test.rb
@@ -694,8 +694,8 @@ class FormulaTests < Homebrew::TestCase
       url "f1-1"
 
       depends_on :python
-      depends_on :x11 => :recommended
-      depends_on :xcode => ['1.0', :optional]
+      depends_on x11: :recommended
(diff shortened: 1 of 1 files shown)


alyssais:
What should I do about the <kbd>brew style</kbd> violation here? Switching to the 1.9 hash syntax would mean not mirroring what real formulae do.

Should I change it to the 1.9 syntax or add a `# rubocop:disable`?

MikeMcQuaid:
> Should I change it to the 1.9 syntax

Yeh, just do that. Doesn't matter if the style matches formulae or not as long as the code is equivalent.



> So, here are some free tests. :D

🎉

MikeMcQuaid:
Thanks @alyssais!

PR #1856, opened 2017-01-16
Revert "formula_installer: set specified_path on pour."

ilovezfs:
Reverts Homebrew/brew#1847

I'm not sure this was such a good idea. I think we'll want to be able to unalias things in the future but this will have the consequence of making aliases permanent, which doesn't really make sense.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -808,7 +808,6 @@ class FormulaInstaller
     tab.poured_from_bottle = true
     tab.time = Time.now.to_i
     tab.head = HOMEBREW_REPOSITORY.git_head
-    tab.source["path"] = formula.specified_path.to_s
     tab.write
   end
 


MikeMcQuaid:
> I'm not sure this was such a good idea. 

This just makes the behaviour consistent when pouring bottles for https://github.com/Homebrew/brew/pull/971 which was merged pre-1.0.0 so I don't think we can just revert this behaviour.

> I think we'll want to be able to unalias things in the future but this will have the consequence of making aliases permanent, which doesn't really make sense.

If we wish to avoid making aliases permanent that can be what's fixed instead. We still need this functionality for versioning.

ilovezfs:
>If we wish to avoid making aliases permanent that can be what's fixed instead. We still need this functionality for versioning.

Which functionality exactly?

MikeMcQuaid:
> Which functionality exactly?

Not making this make aliases permanent i.e. if the tab-referenced alias is missing then handle that gracefully.

ilovezfs:
I don't see why we need to alias it at all. Why not just have `brew install gcc@6` work without creating or storing an alias?

MikeMcQuaid:
> I don't see why we need to alias it at all. Why not just have brew install gcc@6 work without creating or storing an alias?

Having `brew install foo@1` install `foo` when they aren't named the same is an alias whether we store it on disk or not. I'd prefer to keep such aliases machine readable data rather than an implementation detail.

ilovezfs:
So you'd prefer to do alias bookkeeping even if it could be avoided?

MikeMcQuaid:
Yes, I'd prefer not introducing a new alias system that breaks existing assumptions about how our aliases work. If there's too much busywork: that's something that can be automated/scripted.

ilovezfs:
>Yes, I'd prefer not introducing a new alias system that breaks existing assumptions about how our aliases work.

I don't think anyone has an "existing assumption" that
```
brew install ag
brew install the_silver_searcher
```
is the same thing as
```
brew install gcc@6
brew install gcc
```

ilovezfs:
It would also be nice if this "just worked" for all formulae without making users learn, lookup, or remember which formulae are "non-versioned counter-parts of versioned formulae" (wow we need a word for that) and which are not.

MikeMcQuaid:
People have existing (correct) assumptions that `brew install foo` means `foo` is either the name of a formula file or an alias.

> It would also be nice if this "just worked" for all formulae without making users learn, lookup, or remember which formulae are "non-versioned counter-parts of versioned formulae" (wow we need a word for that) and which are not.

It would be nice and we can do that longer-term but that's significantly more involved so passing on it for now.

It's been (genuinely) a good discussion but passing on this revert for now. I'll make a note to address the problems you've addressed but we've already been writing tabs this way since September so can't just back it out now.

alyssais:
Can the branch for this be deleted now?

PR #1864, opened 2017-01-17
Add requirements class for gfortran

zbeekman:

[OpenCoarrays](https://github.com/Homebrew/homebrew-core/pull/8790) provides an implementation of Coarray Fortran compatible with GFortran. It is the official multi-image implementation, and as such has a somewhat tight coupling with GFortran through it's internal API/ABI. This is an attempt at providing formulae with a means of verifying that the fortran compiler is GFortran and that it's version matches or exceeds that required by the formula. However I am a complete blundering fool in Ruby, and therefore I may not have implemented this correctly (testing would suggest this to be the case). I am merely hoping that some kind soul, homebrew maintainer or contributor or otherwise, will take a look at what I have so far and point me in the right direction. The goal is to be able to specify `depends_on :gfortran => "5.4+"` in a formula...

Nope, and I have no idea how to write an effective test for these changes!

Yes and it's still running... it seems to take quite a while.

-----

### DIFF
--- a/Library/Homebrew/requirements/gfortran_requirement.rb
+++ b/Library/Homebrew/requirements/gfortran_requirement.rb
@@ -1,0 +1,33 @@
+class GfortranRequirement < Requirement
+  fatal true
+  default_formula "gcc"
+
+  satisfy build_env: false do
+    next false unless which "gfortran"
+    next true unless @version
+    gfortran_version = Utils.popen_read("gfortran", "--version").split("\n").first.split(" ").last.chomp
+    Version.create(gfortran_version) >= Version.create(@version)
+  end
+
+  env do
+    ENV.prepend_path "PATH", which("gfortran").dirname
+    ENV["FC"] = "gfortran"
+  end
+
+  def initialize(tags)
+    @version = tags.shift if /(\d\.)+\d/ =~ tags.first
+    super
+  end
+
+  def message
+    version_string = " #{@version}" if @version
+
+    s = "GFortran #{version_string} or later is required to install this formula."
+    s += super
+    s
+  end
+
+  def inspect
+    "#<#{self.class.name}: #{name.inspect} #{tags.inspect} version=#{@version.inspect}>"
+  end
+end
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -127,6 +127,7 @@ class DependencyCollector
     when :tuntap     then TuntapRequirement.new(tags)
     when :ant        then ant_dep(tags)
     when :emacs      then EmacsRequirement.new(tags)
+    when :gfortran   then GfortranRequirement.new(tags)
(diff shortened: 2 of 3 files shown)


zbeekman:
I certainly don't expect this to be merged as is... I have very little idea of what I am doing and I don't think it works as it should.

This PR was inspired by discussion with @MikeMcQuaid at https://github.com/Homebrew/homebrew-core/pull/8790 and https://github.com/Homebrew/legacy-homebrew/pull/46547#r55076413 before that.

I am trying to add something like `depends_on :java "1.8+"` but for GFortran... My apologies, I am a complete Ruby noob and don't understand the majority of Ruby syntax and semantics... I blundered through this with sheer mimicry of `requirements/java_requirement.rb` and `requirements/emacs_requirement.rb` and I don't think I have even been able to achieve the functionality I am after, but hoped that a beneficent homebrew maintainer or contributor may be able to look at my hacked together code and point out what needs changing to get it to work and any other style or logic errors I may have made.

MikeMcQuaid:
If this is for a single formula the requirement should probably live in the formula itself. We could work on that in homebrew/core after https://github.com/Homebrew/homebrew-core/issues/8790 is merged?

Library/Homebrew/requirements/gfortran_requirement.rb
```diff
@@ -0,0 +1,33 @@
+class GfortranRequirement < Requirement
+  fatal true
+  default_formula "gcc"
+
+  satisfy build_env: false do
+    next false unless which "gfortran"
+    next true unless @version
+    gfortran_version = Utils.popen_read("gfortran", "--version").split("\n").first.split(" ").last.chomp
```
MikeMcQuaid: This looks correct.

MikeMcQuaid:
This looks pretty good. I'm wondering if it's worth just making this a `depends_on "gcc" => :build` (if it's even build time?) now that that includes `gfortran`?

zbeekman:
> If this is for a single formula the requirement should probably live in the formula itself. We could work on that in homebrew/core after Homebrew/homebrew-core#8790 is merged?

yes I agree you're probably right about that... although it may end up being useful for other formulae in e.g., homebrew-science... but probably not worth introducing my horrid Ruby into brew itself :-)

Library/Homebrew/requirements/gfortran_requirement.rb
zbeekman: I think we can grab some of this logic and insert it into https://github.com/Homebrew/homebrew-core/pull/8790 perhaps after we merge a decent enough first cut?

zbeekman:
> This looks pretty good. I'm wondering if it's worth just making this a depends_on "gcc" => :build (if it's even build time?) now that that includes gfortran?

Yes, I replied at https://github.com/Homebrew/homebrew-core/pull/8790 ... I think you're right and that this will get us 90% of the way there (although it is runtime and build dep)

PR #1878, opened 2017-01-18
Simplify Keg class

alyssais:

-----

The `Keg` class is pretty complex. It's scattered with repetitive method definitions that just forward to `path`. These are a couple of straightforward changes to make it a bit less difficult to navigate and, well… long.

As far as I can tell, the conditional aliasing of `to_path` is a remnant of Ruby 1.8.

I had a one-off test failure [here](https://github.com/Homebrew/brew/blob/f0249643d47f1dee4a0d08c440eb3afd6da3558a/Library/Homebrew/test/formula_installer_test.rb#L23) as part of [this test](https://github.com/Homebrew/brew/blob/f0249643d47f1dee4a0d08c440eb3afd6da3558a/Library/Homebrew/test/formula_installer_test.rb#L82), but it seems unlikely to be related and I haven't been able to reproduce since…

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -174,6 +174,12 @@ class Keg
   attr_reader :path, :name, :linked_keg_record, :opt_record
   protected :path
 
+  extend Forwardable
+
+  def_delegators :path,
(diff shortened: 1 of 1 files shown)


alyssais:
Huh, got the failure on Travis too.

PR #1879, opened 2017-01-19
tests: add a missing Tab cache clear

alyssais:

-----

This was responsible for the sporadic failures I encountered in #1878.

It feels like constant whack-a-mole with these caches causing order-dependent test failures, and it's a hugely complicated process to reproduce, isolate and fix them (this one took me the better part of a day). We should try to implement some means of checking after each test to ensure they're not accidentally leaving behind persistent state.


alyssais:
Just found _another_ one of these (second commit). This one hasn't caused any issues as far as I can tell, but I bet it would have done at some point.

MikeMcQuaid:
Nice, thanks @alyssais!

PR #1881, opened 2017-01-19
`brew orphaned`

alyssais:

-----

This builds on @MikeMcQuaid's work in #1813 by adding a <kbd>brew unused</kbd> that will list kegs that were installed implicitly, but are no longer required by any explicitly installed kegs.

It takes a conservative approach, since I assume a common thing for people to do will be to uninstall all the kegs listed as unused. Specifically, if an install receipt doesn't specify whether a keg was installed by request or not (because it's too old), it will assume it was.

I haven't implemented the `--ignore-promotions` flag @ilovezfs posits in https://github.com/Homebrew/brew/pull/1813#issuecomment-272811547, but it would be very easy to implement in future with this foundation — we'd just pass a custom `Proc` to `Keg.unused`.

Something I'd like other people's thoughts on is what the output should be. Currently, each keg it detects to be "unused" is outputted by name and version. This seems the most readable option, but it would be nice to have something that could be passed to <kbd>brew uninstall</kbd> so brew would immediately support an `autoremove` equivalent in <kbd>brew uninstall $(brew unused)</kbd>. One solution here would be to add a `--full-paths` option or similar that would output the full keg path rather than name and version. This could, perhaps, be automatically enabled for non-TTY output?

### DIFF
--- a/Library/Homebrew/cmd/orphaned.rb
+++ b/Library/Homebrew/cmd/orphaned.rb
@@ -1,5 +1,5 @@
 #:  * `orphaned`:
-#:    Lists kegs that were installed as dependencies of other kegs, but are no
+#:    List kegs that were installed as dependencies of other kegs, but are no
 #:    longer required for any explicitly-installed keg.
 
 require "keg"
--- a/Library/Homebrew/cmd/orphaned.rb
+++ b/Library/Homebrew/cmd/orphaned.rb
@@ -1,4 +1,4 @@
-#:  * `unused`:
+#:  * `orphaned`:
 #:    Lists kegs that were installed as dependencies of other kegs, but are no
 #:    longer required for any explicitly-installed keg.
 
@@ -7,7 +7,7 @@ require "keg"
 module Homebrew
   module_function
 
-  def unused
-    puts Keg.unused.map { |k| "#{k.name} #{k.version}" }
+  def orphaned
+    puts Keg.orphaned.map { |k| "#{k.name} #{k.version}" }
   end
 end
--- a/Library/Homebrew/cmd/orphaned.rb
+++ b/Library/Homebrew/cmd/orphaned.rb
@@ -8,6 +8,6 @@ module Homebrew
   module_function
 
   def orphaned
-    puts Keg.orphaned.map { |k| "#{k.name} #{k.version}" }
+    puts Keg.orphaned.map(&:name).uniq
   end
 end
--- a/Library/Homebrew/cmd/orphaned.rb
+++ b/Library/Homebrew/cmd/orphaned.rb
@@ -1,6 +1,6 @@
 #:  * `orphaned`:
-#:    List kegs that were installed as dependencies of other kegs, but are no
-#:    longer required for any explicitly-installed keg.
+#:    Find formulae that were installed as dependencies and are no
+#:    longer required.
 
 require "keg"
 
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -170,11 +170,11 @@ class Keg
   end
 
   def might_have_been?(key)
-    Tab.for_keg(self).might_be?(key)
+    Tab.for_keg(self).absent_or_truthy?(key)
   end
 
   def definitely_was_not?(key)
-    Tab.for_keg(self).definitely_not?(key)
+    Tab.for_keg(self).present_and_falsey?(key)
   end
 
   # if path is a file in a keg then this will return the containing Keg object
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -165,16 +165,10 @@ class Keg
   end
 
   def orphaned?
-    return false if might_have_been? :installed_on_request
-    installed_dependents.none? { |d| d.might_have_been? :installed_on_request }
-  end
-
-  def might_have_been?(key)
-    Tab.for_keg(self).absent_or_truthy?(key)
-  end
-
-  def definitely_was_not?(key)
-    Tab.for_keg(self).present_and_falsey?(key)
+    return false if Tab.for_keg(self).absent_or_truthy? :installed_on_request
+    installed_dependents.none? do |dependent|
+      Tab.for_keg(dependent).absent_or_truthy? :installed_on_request
+    end
   end
 
   # if path is a file in a keg then this will return the containing Keg object
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -157,6 +157,14 @@ class Keg
     nil
   end
 
+  def might_have_been?(key)
+    Tab.for_keg(self).might_be?(key)
+  end
+
+  def definitely_was_not?(key)
+    Tab.for_keg(self).definitely_not?(key)
+  end
+
   # if path is a file in a keg then this will return the containing Keg object
   def self.for(path)
     path = path.realpath
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -160,13 +160,13 @@ class Keg
 
   # Finds kegs that were installed as dependencies of other kegs,
   # but that are no longer required by any explicitly-installed kegs.
-  def self.unused(*args)
-    all.select { |k| k.unused?(*args) }
+  def self.orphaned
+    all.select(&:orphaned?)
   end
 
-  # Pass a :needed proc to customise which kegs are determined to be unused.
-  def unused?(needed: ->(k) { k.might_have_been? :installed_on_request })
-    !needed.call(self) && installed_dependents.none?(&needed)
+  def orphaned?
+    return false if might_have_been? :installed_on_request
+    installed_dependents.none? { |d| d.might_have_been? :installed_on_request }
   end
 
   def might_have_been?(key)
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -166,9 +166,7 @@ class Keg
 
   def orphaned?
     return false if Tab.for_keg(self).absent_or_truthy? :installed_on_request
-    installed_dependents.none? do |dependent|
-      Tab.for_keg(dependent).absent_or_truthy? :installed_on_request
-    end
+    installed_dependents.all?(&:orphaned?)
   end
 
   # if path is a file in a keg then this will return the containing Keg object
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -1,5 +1,6 @@
 require "extend/pathname"
 require "keg_relocate"
+require "formula"
 require "formula_lock"
 require "ostruct"
 
@@ -155,6 +156,17 @@ class Keg
     end
 
     nil
+  end
+
+  # Finds kegs that were installed as dependencies of other kegs,
+  # but that are no longer required by any explicitly-installed kegs.
+  def self.unused(*args)
+    all.select { |k| k.unused?(*args) }
+  end
+
+  # Pass a :needed proc to customise which kegs are determined to be unused.
+  def unused?(needed: ->(k) { k.might_have_been? :installed_on_request })
+    !needed.call(self) && installed_dependents.none?(&needed)
   end
 
   def might_have_been?(key)
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -344,6 +344,20 @@ class Tab < OpenStruct
     tabfile.atomic_write(to_json)
   end
 
+  # Helper method for checking truthiness of a value that may not be specified.
+  # `Tab.might_be?(:foo)` differs from `tab.foo` in that if `foo` is not
+  # specified `#might_be?` will return true.
+  def might_be?(key)
+    !definitely_not?(key)
+  end
+
+  # Helper method for checking falsiness of a value that may not be specified.
+  # `Tab.definitely_not?(:foo)` differs from `tab.foo` in that if `foo` is not
+  # specified `#definitely_not?` will return false.
+  def definitely_not?(key)
+    __send__(key) == false
+  end
+
   def to_s
     s = []
     if poured_from_bottle
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -345,14 +345,14 @@ class Tab < OpenStruct
   end
 
   # Helper method for checking truthiness of a value that may not be specified.
-  # `Tab.might_be?(:foo)` differs from `tab.foo` in that if `foo` is not
+  # `Tab#might_be?(:foo)` differs from `tab.foo` in that if `foo` is not
   # specified `#might_be?` will return true.
   def might_be?(key)
     !definitely_not?(key)
   end
 
   # Helper method for checking falsiness of a value that may not be specified.
-  # `Tab.definitely_not?(:foo)` differs from `tab.foo` in that if `foo` is not
+  # `Tab#definitely_not?(:foo)` differs from `tab.foo` in that if `foo` is not
   # specified `#definitely_not?` will return false.
   def definitely_not?(key)
     __send__(key) == false
--- a/Library/Homebrew/tab.rb
+++ b/Library/Homebrew/tab.rb
@@ -345,16 +345,16 @@ class Tab < OpenStruct
   end
 
   # Helper method for checking truthiness of a value that may not be specified.
-  # `Tab#might_be?(:foo)` differs from `tab.foo` in that if `foo` is not
-  # specified `#might_be?` will return true.
-  def might_be?(key)
-    !definitely_not?(key)
+  # `Tab#absent_or_truthy?(:foo)` differs from `tab.foo` in that if `foo` is not
+  # specified `#absent_or_truthy?` will return true.
+  def absent_or_truthy?(key)
+    !present_and_falsey?(key)
   end
 
   # Helper method for checking falsiness of a value that may not be specified.
-  # `Tab#definitely_not?(:foo)` differs from `tab.foo` in that if `foo` is not
-  # specified `#definitely_not?` will return false.
-  def definitely_not?(key)
+  # `Tab#present_and_falsey?(:foo)` differs from `tab.foo` in that if `foo` is
+  # not specified `#present_and_falsey?` will return false.
+  def present_and_falsey?(key)
     __send__(key) == false
   end
 
--- a/Library/Homebrew/cmd/unused.rb
+++ b/Library/Homebrew/cmd/unused.rb
@@ -1,0 +1,13 @@
+#:  * `unused`:
+#:    Lists kegs that were installed as dependencies of other kegs, but are no
+#:    longer required for any explicitly-installed keg.
+
+require "keg"
+
+module Homebrew
+  module_function
+
+  def unused
+    puts Keg.unused.map { |k| "#{k.name} #{k.version}" }
+  end
+end
--- a/Library/Homebrew/test/keg_spec.rb
+++ b/Library/Homebrew/test/keg_spec.rb
@@ -116,26 +116,6 @@ describe Keg do
     end
   end
 
-  specify "#might_have_been?" do
-    expect(keg).to be_might_have_been(:foo)
-
-    alter_tab(keg) { |t| t.foo = true }
-    expect(keg).to be_might_have_been(:foo)
-
-    alter_tab(keg) { |t| t.foo = false }
-    expect(keg).not_to be_might_have_been(:foo)
-  end
-
-  specify "#definitely_was_not?" do
-    expect(keg).not_to be_definitely_was_not(:foo)
-
-    alter_tab(keg) { |t| t.foo = true }
-    expect(keg).not_to be_definitely_was_not(:foo)
-
-    alter_tab(keg) { |t| t.foo = false }
-    expect(keg).to be_definitely_was_not(:foo)
-  end
-
   specify "#empty_installation?" do
     %w[.DS_Store INSTALL_RECEIPT.json LICENSE.txt].each do |file|
       touch keg/file
--- a/Library/Homebrew/test/keg_spec.rb
+++ b/Library/Homebrew/test/keg_spec.rb
@@ -17,6 +17,13 @@ describe Keg do
     keg
   end
 
+  def alter_tab(keg)
+    tab = Tab.for_keg(keg)
+    tab.tabfile ||= keg/Tab::FILENAME
+    yield tab
+    tab.write
+  end
+
   around(:each) do |example|
     begin
       @old_stdout = $stdout
@@ -47,6 +54,26 @@ describe Keg do
   specify "::all" do
     Formula.clear_racks_cache
     expect(described_class.all).to eq([keg])
+  end
+
+  specify "#might_have_been?" do
+    expect(keg).to be_might_have_been(:foo)
+
+    alter_tab(keg) { |t| t.foo = true }
+    expect(keg).to be_might_have_been(:foo)
+
+    alter_tab(keg) { |t| t.foo = false }
+    expect(keg).not_to be_might_have_been(:foo)
+  end
+
+  specify "#definitely_was_not?" do
+    expect(keg).not_to be_definitely_was_not(:foo)
+
+    alter_tab(keg) { |t| t.foo = true }
+    expect(keg).not_to be_definitely_was_not(:foo)
+
+    alter_tab(keg) { |t| t.foo = false }
+    expect(keg).to be_definitely_was_not(:foo)
   end
 
   specify "#empty_installation?" do
@@ -349,16 +376,10 @@ describe Keg do
       keg.link
     end
 
-    def alter_tab(keg = dependent)
-      tab = Tab.for_keg(keg)
-      yield tab
-      tab.write
-    end
-
     # 1.1.6 is the earliest version of Homebrew that generates correct runtime
     # dependency lists in Tabs.
     def dependencies(deps, homebrew_version: "1.1.6")
-      alter_tab do |tab|
+      alter_tab dependent do |tab|
         tab.homebrew_version = homebrew_version
         tab.tabfile = dependent/Tab::FILENAME
         tab.runtime_dependencies = deps
--- a/Library/Homebrew/test/keg_spec.rb
+++ b/Library/Homebrew/test/keg_spec.rb
@@ -114,6 +114,18 @@ describe Keg do
       expect(keg).to be_orphaned
       expect(Keg.orphaned).to contain_exactly(keg, dependent)
     end
+
+    it "does not include kegs with non-orphaned dependents" do
+      dependent = make_dependent(keg)
+      recursive_dependent = make_dependent(dependent)
+
+      alter_tab(keg) { |t| t.installed_on_request = false }
+      alter_tab(dependent) { |t| t.installed_on_request = false }
+      alter_tab(recursive_dependent) { |t| t.installed_on_request = true }
+
+      expect(keg).not_to be_orphaned
+      expect(Keg.orphaned).to be_empty
+    end
   end
 
   specify "#empty_installation?" do
--- a/Library/Homebrew/test/keg_spec.rb
+++ b/Library/Homebrew/test/keg_spec.rb
@@ -59,7 +59,7 @@ describe Keg do
     expect(described_class.all).to eq([keg])
   end
 
-  describe "unused" do
+  describe "orphaned" do
     def make_dependent(keg)
       keg.optlink unless keg.optlinked?
       dependent = setup_test_keg("#{keg.name}_dependent", "1.0")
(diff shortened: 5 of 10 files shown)


ilovezfs:
This seems redundant with `brew leaves`, and also premature to add. I think most package managers implement this in terms of an `autoremove` command, if anything.

MikeMcQuaid:
> It takes a conservative approach, since I assume a common thing for people to do will be to uninstall all the kegs listed as unused. Specifically, if an install receipt doesn't specify whether a keg was installed by request or not (because it's too old), it will assume it was.

Good idea 👍 

> This seems the most readable option, but it would be nice to have something that could be passed to brew uninstall so brew would immediately support an autoremove equivalent in brew uninstall $(brew unused)

This fits in well with our CLI UI 👍 

> This seems redundant with brew leaves

This is the opposite of `brew leaves`, no? Like `brew uses` and `brew deps`? Regardless I agree `brew leaves` should probably use this logic before we add a new command.

Nice work 👏

ilovezfs:
```
This is the opposite of brew leaves
```
Right.

>Regardless I agree brew leaves should probably use this logic before we add a new command.

Right.

>Like brew uses and brew deps

Sure, but we don't have `brew present` to go with `brew missing` or `brew current` to go with `brew outdated`.

MikeMcQuaid:
> Sure, but we don't have brew present to go with brew missing or brew current to go with brew outdated.

We kinda do; `brew list` covers both those cases. Arguably it's `brew leaves` that's the weird/pointless command.

ilovezfs:
I think they could all probably be moved into `brew list` as `--` options.

alyssais:
> I think they could all probably be moved into `brew list` as `--` options.

Seems reasonable. I don't know if maybe that would make them slightly less discoverable though?

ilovezfs:
arguably more discoverable since almost everyone will use list at some point but I bet the vast majority of users have no idea leaves or outdated exist

alyssais:
A couple of other concerns I'd have about making them `--` options passed to <kbd>brew list</kbd>:

- The output format wouldn't be the same for all of them. For `unused` to be useful, it really has to list kegs, whereas other list commands list formulae. We could maybe standardise of listing kegs, though (maybe with some nicer way of referencing them than the full path).
- Each list command would take different arguments. For example, if the `--ignore-promotions` option was added to `--unused`, that would be a flag that only worked in conjunction with the `--unused` flag. What happens if `--ignore-promotions` is specified without `--unused`? What happens if two different list commands were specified (e.g. `--unused --leaves`)? We'd need to implement a lot of compatibility checking between all the combinations of options.

  This could be resolved by making them **subcommands** of <kbd>brew list</kbd>, rather than `--` options. Then they'd look like `brew list unused --ignore-promotions`. At this point, though, there's not really much difference between `brew list unused` and `brew unused` except five extra characters.

ilovezfs:
Mutually exclusive options are fairly common, and in general behavior is simply undefined when mutually exclusive options are specified at the same time.

If this is going to be a separate command, I don't think `unused` is the right name because we already have `uses` which has a very different meaning.

alyssais:
`unneeded`? `nonrequired`? `leftovers`?

I actually kind of like `leftovers` hehe but it's not particularly obvious.

ilovezfs:
I'm not sure yet, but I'd be curious to know what words other package managers are using for these thingies.

ilovezfs:
expendables_ver9_xlg

alyssais:
<dl>
<dt><code>apt</code></dt>
<dd>Has an `autoremove` command, but gets around the problem of what to call unused packages by not having a command that tells you what will be removed by `autoremove`, as far as I can tell. In their documentation for <a href="https://help.ubuntu.com/community/AptGet/Howto"><code>autoremove</code></a>, they say <q>packages that were installed by other packages and are no longer needed</q>.</dd>
<dt><code>pacman</code></dt>
<dd>Seems to call them <a href="https://wiki.archlinux.org/index.php/Pacman/Tips_and_tricks#Removing_unused_packages_.28orphans.29"><q>orphans</q></a>.</dd>
<dt><code>dnf</code>/<code>yum</code></dt>
<dd>Doesn't seem to have a name for them, just <a href="http://dnf.readthedocs.io/en/latest/command_ref.html#autoremove-command-label">calls them</a> <q>“leaf” packages from the system that were originally installed as dependencies of user-installed packages but which are no longer required by any such package</q>.</dd>
</dl>

I'll note that "unused" seems to be the most common term people use on Q&A sites to ask how to clean up leftover packages.

ilovezfs:
>I'll note that "unused" seems to be the most common term people use on Q&A sites to ask how to clean up leftover packages.

That's not surprising but I think we'd need to change `brew uses` to something else if we went with "used/unused" for this characteristic, and that would be an API breaking change insofar as the CLI is our API.

ilovezfs:
>pacman seems to call them "orphans".

That's not entirely awful.

>"“leaf” packages from the system that were originally installed as dependencies of user-installed packages but which are no longer required by any such package".

This is actually a good point. `brew leaves` is not in fact an antonym as these are indeed leaves.

alyssais:
@MikeMcQuaid what do you think of this? You suggested to me last week that I make `brew uses` (and similar) use the keg-based dependency checking anyway if possible. Would that fix our terminology problem here?

alyssais:
> > pacman seems to call them "orphans".
>
> That's not entirely awful.

Yeah, it's not. Not brilliant though — technically all `leaves` should be `orphans`, so I don't think it'd be immediately obvious to an end user what its purpose was like it would be with `unused` or `leftovers`.

ilovezfs:
>immediately obvious to an end user what its purpose was

I don't think it really needs to be immediately obvious, as that's why we have and man pages. Also, I think most users will never have an occasion to use this command and will instead use e.g., `autoremove --dry-run` whenever it comes into being.

ilovezfs:
Looks like Gentoo also uses orphan in their docs: https://wiki.gentoo.org/wiki/Knowledge_Base:Remove_orphaned_packages

chdiza:
One can already check, about a specific package, whether it is an "orphan".  If the user wonders "hey, what is foo in here for?" all it takes is `brew uses --installed foo` and check for empty output.

bfontaine:
@chdiza That only tells you that an installed formula is not depended on by another one. That doesn’t tell you if *you* installed it vs. it was installed as a dependency and is no longer needed.

chdiza:
@bfontaine Well, the name 'unused' suggests that all the user wants is know whether there are any unused kegs and if so what they are.  Whether a package is "used", i.e., is depended upon by some other installed package, is independent of whether it was installed manually on purpose or by some other package.

But now I see I didn't read carefully enough about what this command, name aside, is actually supposed to do.  If this command is supposed to only list the unused ones that were originally pulled in by something else that is no longer present, then the command should be called `brew orphans` (or whatever noun gets picked, like `brew residue` or `brew cruft`, or even `brew unnecessary`).

alyssais:
@chdiza if I install a formula that I directly use, e.g. <kbd>brew install git</kbd>, I wouldn't consider git to be "unused", because I use it!

That said, if there's a name that communicates it more effectively (I'm not convinced "orphans" does) we should go with that.

ilovezfs:
garbage?

alyssais:
> garbage?

Seems basically the same as `leftovers`, except "garbage" is a chiefly [North American](https://github.com/Homebrew/brew/blob/bbed7246bc5c5b7acb8c1d427d10b43e090dfd39/docs/Prose-Style-Guidelines.md#style-and-usage) word whereas `leftovers` and other synonyms are fairly universal in English varieties.

ilovezfs:
I mean `garbage` in the CS sense.

alyssais:
I guess that makes sense. I don't know how obvious it's meaning would be in this context though?

ilovezfs:
Yeah, again that would be a man page issue.

I think it might also help in thinking this over to consider what verbs we will/could use with these various adjectives for a CLI that transitions the state. For example,

divorced -> marry/wed(?) and divorce
orphans or orphaned -> adopt and abandon/disown
unused -> use and unuse(?)
uninvited -> invite and disinvite
? -> add and drop/subtract
downsized(?) -> promote and demote

alyssais:
I don't suppose there's a 🍺 metaphor that works here, is there?
Then we'd have a clear winner. :P

ilovezfs:
Then someone will object it's more jargon, which became a no-no of late. lol

Another one I just thought of: `removable`

ilovezfs:
Also: detached

alyssais:
🍺 spills could be mopped up/evaporated (deleted) or, uh, unspilled? 😛
A cellar with spills in it could be sticky!

***

`removable` could work.

alyssais:
Honestly I don't think we're going to be able to avoid jargon here. Might as well be consistent with our other jargon?

ilovezfs:
Which brew _commands_ actually use jargon?

alyssais:
Depends on your definition of "jargon", but I'd consider these to be jargon:

- `--cellar`
- `tap-info`
- `tap-pin`
- `tap-unpin`
- `tap`
- `untap`

And possibly also:

- `doctor`
- `leaves`
- `prune`

A bunch of them also use terms like "formula", "keg", etc. in their man pages.

ilovezfs:
Yeah the tap ones, I agree. doctor/leaves/prune not really.

>A bunch of them also use terms like "formula", "keg", etc. in their man pages.

Right. The jargon is pervasive everywhere except the commands themselves.

alyssais:
The problem with, e.g., `removable`, `redundant`, etc. is that I'd expect that to also include old versions of kegs that had since been upgraded (which should also be cleaned up, but that's a different discussion).

RandomDSdevel:
Just a quick end-user comment, but you guys _may_ want to take into account the fact that `brew cleanup` already exists and fulfills a similar purpose.  (Also, for the record, _I_ use `brew outdated`…a lot&thinsp;—&thinsp;as in, pretty much every time I run Homebrew as a sort of `brew upgrade --dry-run`, if you will.)

alyssais:
You're right @RandomDSdevel. In fact, when we implement automatic deletion of these kegs <kbd>brew cleanup</kbd> would probably be a good place to put it. Right now, though, as a starting point, I'd just like to have some way to see what the unneeded kegs _are_.

MikeMcQuaid:
`removable` or `unused` or `unneeded` are my preferences so far.

MikeMcQuaid:
I also would like to see this be a separate command and be handled sensibly by `brew cleanup` eventually. I don't think `brew uses` keg stuff relates particularly but would still be cool.

JCount:
I personally like the term `removable`. In my mind this creates a good distinction between the leaves that were implicitly installed to satisfy dependencies and the leaves that were explicitly installed by the user, and are therefore _not_ `removable`.

MikeMcQuaid:
`uninstallable` would arguably be more mouthy but consistent

ilovezfs:
I think a problem with `brew removable` and `brew uninstallable` is that they're both inaccurate descriptions given that the explicitly user-installed formulae are subject to `brew remove foo` or `brew uninstall foo` just like anything else.

JCount:
This just popped into my head, what about `cleanable`? This seems to tie well with how we use `cleanup` to remove other cruft from a user's system. (And in case anyone is wondering, it is an actual word.)

MikeMcQuaid:
`cleanable` is good to me. I'd probably prefer we get something in with some name which we can always alias and rename later if we need to. How about everyone who has already commented in here posts a single word comment with their current favourite (if it's not already posted) and anyone can 👍 if it's already posted.

ilovezfs:
orphaned

JCount:
cleanable

alyssais:
`leftovers`

bfontaine:
`unneeded`

chdiza:
Since the votes were requested, only `leftovers` and `orphaned` contain any hint that the command will look for installed formulas whose presence on the machine is due to a certain history of events.

Any installed formula that is not a dep of something other installed formula is "cleanable".  (Just `uninstall` and `cleanup`).  Also, if a user installs `foo` just to play around with it (and it's not a dep of anything subsequently installed), then `foo` is "unneeded".

ilovezfs:
Given that orphaned seems to be the term of art in the documentation of other package managers, and that leftovers sounds like it pertains to food not software, I'd advise we start with orphaned. It has the additional benefit of having the obvious associated commands "adopt" and "disown" or "abandon" whereas leftovers has no such obvious commands unless "brew unleaveover" and "brew leaveover" are acceptable :trollface:

JCount:
I relent to @ilovezfs unrelenting badgering on slack, and change my vote to `orphaned`. 😒   He won't be quiet... 😭

MikeMcQuaid:
Orphaned it is 👍

maxim-belkin:
Haven't looked through the suggested source code improvements, but such formulae as `pkg-config`, `automake`, etc. are needed to install other formulae (or non-Homebrew projects) only. Hence, they might be detected as `orphaned` or `unused`. What is the plan for these? Will there be a "safe-list"?

ilovezfs:
If you explicitly install them, they're automatically not considered orphans. But this is why we'll need some kind of adopt/abandon so you can toggle it yourself without reinstalling or manually editing receipts.

maxim-belkin:
got it. Hence `installed_on_request`, right?

ilovezfs:
Yup. And installed_as_dependency.

alyssais:
Renamed. Go democracy!

Library/Homebrew/cmd/orphaned.rb
```diff
@@ -0,0 +1,13 @@
+#:  * `orphaned`:
+#:    Lists kegs that were installed as dependencies of other kegs, but are no
```
MikeMcQuaid: Let's use `installed formulae` rather than `kegs`; it's widely misunderstood as a term.

Library/Homebrew/cmd/orphaned.rb
```diff
@@ -0,0 +1,13 @@
...
+#:    longer required for any explicitly-installed keg.
+
+require "keg"
+
+module Homebrew
+  module_function
+
+  def orphaned
+    puts Keg.orphaned.map { |k| "#{k.name} #{k.version}" }
```
MikeMcQuaid: I wonder if the version should be omitted when it's not a Tty so it can be used as a subshell more usefully?

Library/Homebrew/tab.rb
```diff
@@ -341,6 +341,20 @@ def write
     tabfile.atomic_write(to_json)
   end
 
+  # Helper method for checking truthiness of a value that may not be specified.
+  # `Tab.might_be?(:foo)` differs from `tab.foo` in that if `foo` is not
+  # specified `#might_be?` will return true.
+  def might_be?(key)
```
MikeMcQuaid: `def key_truthy_or_absent?`

Library/Homebrew/tab.rb
```diff
@@ -341,6 +341,20 @@ def write
...
+    !definitely_not?(key)
+  end
+
+  # Helper method for checking falsiness of a value that may not be specified.
+  # `Tab.definitely_not?(:foo)` differs from `tab.foo` in that if `foo` is not
+  # specified `#definitely_not?` will return false.
+  def definitely_not?(key)
```
MikeMcQuaid: `def key_falsey_or_absent?`

Library/Homebrew/cmd/orphaned.rb
maxim-belkin: I'd also suggest to change "List**s**" to "List" or "Find"
```
Find formulae that were installed as dependencies and are no longer required.
```

Library/Homebrew/cmd/orphaned.rb
maxim-belkin: something like:
```sh
brew orphaned | xargs brew remove
```

Library/Homebrew/tab.rb
maxim-belkin: Neither of these names give a hint what the function does.

Library/Homebrew/cmd/orphaned.rb
alyssais: If version 1.0 of a formula is orphaned but 1.1 is required, that `brew orphaned | xargs brew uninstall` would remove 1.1 and leave 1.0.

With that in mind, would it be better to output the full keg path when not outputting to a TTY?

Library/Homebrew/cmd/orphaned.rb
alyssais: Good catch @maxim-belkin. That's more consistent with the other Homebrew commands.

Library/Homebrew/cmd/orphaned.rb
ilovezfs: I think that starts to confuse this with cleanup.

MikeMcQuaid: Agreed; I think orphaned should be version agnostic and probably actually never print versions at all.

alyssais: So in that case, does it only show formulae where _no_ versions are required?

ilovezfs: Only one "version" can be opt-linked at a time, so not sure what you mean exactly.

alyssais: Oh, of course. Never mind. :)

wickles:
I didn't see this mentioned elsewhere, it's probably worth noting the terminology used by macports which always made sense to me. They use the keywords/pseudo-portnames (from the man pages)
```
o   requested: installed ports that were explicitly asked for
o   unrequested: installed ports that were installed only to satisfy dependencies
o   leaves: installed ports that are unrequested and have no dependents
```
So for example I might do (`port installed` is sort of analogous to `brew list --versions`)
```
$ port installed leaves
The following ports are currently installed:
  autoconf @2.69_5 (active)
  automake @1.15_1 (active)
  gsed @4.2.2_0 (active)
  pkgconfig @0.29.1_0 (active)
```
or
```
$ port installed requested
The following ports are currently installed:
  djvulibre @3.5.27_0 (active)
  pdf2djvu @0.9.4_5 (active)
```
or
```
$ port installed unrequested
The following ports are currently installed:
  autoconf @2.69_5 (active)
  automake @1.15_1 (active)
  bzip2 @1.0.6_0 (active)
  cairo @1.14.6_0+quartz+x11 (active)
  curl @7.50.2_0+ssl (active)
  curl-ca-bundle @7.50.2_0 (active)
  db48 @4.8.30_4 (active)
  expat @2.2.0_0 (active)
  fontconfig @2.12.1_0 (active)
  freetype @2.6.5_0 (active)
  ...
```
Basically they use the `leaves` keyword/command as well, but they also take into account whether a port/formula is requested or not which is pretty handy.

maxim-belkin:
@wickles `installed` in `port` terminology seems to be a bit redundant. Also, `leaves` to me is something that is 'attached' to the tree, though nothing 'depends on it'. I can't say that `orphaned` is 100% to my liking but it's closer

Library/Homebrew/tab.rb
alyssais: @maxim-belkin when you say "neither", do you mean `might_be?` and `key_truthy_or_absent?`, or `might_be?` and `definitely_not?`?

maxim-belkin: I was referring to `might_be?` and `key_truthy_or_absent?`. But my critique was mostly devoted to the description `truthiness of a value that may not be specified` that I don't (yet) understand 🙄

alyssais: Well, a tab can either have a key set, or it can not. If it's not specified, sometimes we want to be able to treat that as if the value was truthy, e.g.

```ruby
if tab.installed_on_request || tab.installed_on_request.nil?
```

In other words, do this if the tab was either installed on request, or if we don't have data for whether it was installed in request. That allows us to take actions only if we know for sure that a tab _wasn't_ installed on request.

Does that make sense? Any idea how I could improve the description or name of the method to get that idea across better while staying succinct?

MikeMcQuaid: I disagree that `def key_truthy_or_absent?` doesn't indicate that it does. I don't think @maxim-belkin's opinion should block this progressing.

alyssais: :+1: to progressing — but the delay here has been on me, not this minor comment! If we can improve it while it's still a work in progress, that's very welcome!

I'm not sure about `key_truthy_or_absent?` though — we're not checking if the key is truthy like that name implies.

MikeMcQuaid: > I'm not sure about key_truthy_or_absent? though — we're not checking if the key is truthy like that name implies.

Can you suggest a name in a similar format?

alyssais: How about just `absent_or_truthy?`?

MikeMcQuaid: Seems fine or `nil_or_truthy?`. Ideally I think `key` would be in the name but don't feel incredibly strongly.

monouser7dig:
This is a great effort to make brew more usable as a package manager! Thanks!

Library/Homebrew/keg.rb
```diff
@@ -157,6 +158,25 @@ def self.find_some_installed_dependents(kegs)
...
+  def self.orphaned
+    all.select(&:orphaned?)
+  end
+
+  def orphaned?
+    return false if might_have_been? :installed_on_request
+    installed_dependents.none? { |d| d.might_have_been? :installed_on_request }
+  end
+
+  def might_have_been?(key)
```
alyssais: You wanting these changed too @MikeMcQuaid?

MikeMcQuaid: Yeh, I think so, thanks. I'd be tempted to just call the tab methods, though; I think that's going to be more readable.

Library/Homebrew/keg.rb
```diff
@@ -157,6 +158,25 @@ def self.find_some_installed_dependents(kegs)
...
+    Tab.for_keg(self).absent_or_truthy?(key)
+  end
+
+  def definitely_was_not?(key)
+    Tab.for_keg(self).present_and_falsey?(key)
```
MikeMcQuaid: Where is this method called?

MikeMcQuaid: (If it's nowhere now it can probably be killed)

alyssais: It's not — I put it there for consistency with `#might_have_been?`, its inverse.

alyssais: 💀👍

alyssais:
There's one more thing I want to do before this is ready for merge — I just realised that there is (no longer?) a unit test for `Keg.outdated`, so I want to add one of those first.

Library/Homebrew/cmd/orphaned.rb
```diff
@@ -0,0 +1,13 @@
+#:  * `orphaned`:
+#:    Find formulae that were installed as dependencies and are no
+#:    longer required.
```
maxim-belkin: why do you break the line here?

maxim-belkin:
```sh
[mbelkin@MacBook-Pro] [Fri Mar 17 11:11:52] [last exit code: 1]
/usr/local/Homebrew
(master) $ brew orphaned
dirmngr
libusb
libusb-compat

[mbelkin@MacBook-Pro] [Fri Mar 17 11:12:09]
/usr/local/Homebrew
(master) $ brew orphaned | xargs -n 1 brew remove
Error: Refusing to uninstall /usr/local/Cellar/dirmngr/1.1.1_2
because it is required by gnupg2 2.0.30_3, which is currently installed.
You can override this and force removal with:
  brew uninstall --ignore-dependencies dirmngr
Error: Refusing to uninstall /usr/local/Cellar/libusb/1.0.21
because it is required by gnupg2 2.0.30_3, which is currently installed.
You can override this and force removal with:
  brew uninstall --ignore-dependencies libusb
Error: Refusing to uninstall /usr/local/Cellar/libusb-compat/0.1.5
because it is required by gnupg2 2.0.30_3, which is currently installed.
You can override this and force removal with:
  brew uninstall --ignore-dependencies libusb-compat
```

monouser7dig:
@maxim-belkin so this seems like it is not working at all? sorry for that dump question 😑

alyssais:
@maxim-belkin why do you have the `-n 1`? I think it would work without that.

ilovezfs:
If gnupg2 was installed, how are they orphans?

alyssais:
Oh, right, yes. Missed that. Interesting. Probably because there are old versions installed — I should make it check for only the latest versions.

maxim-belkin:
I think `orphaned` must be version-agnostic. thoughts?

ilovezfs:
There probably needs to be at least one option around this, on analogy to the `-s` for cleanup.

alyssais:
@maxim-belkin I think it should only affect linked kegs — not necessarily the latest version, but as discussed above, anything else is in `brew cleanup`'s scope.

maxim-belkin:
@ilovezfs I'm not sure I understand how it would work
@alyssais that's what I thought and that's why I think `orphaned` must be version-agnostic: it should just check whatever is installed. Now, I don't know whether `cleanup` uninstalls unlinked kegs but I doubt it does.

alyssais:
@maxim-belkin my understanding is that cleanup removes _all_ kegs that are not

- linked
- pinned
- the current version of the formula

maxim-belkin:
there should be `brew cleanup --unlinked`, `brew cleanup --unpinned`, `brew cleanup --outdated` then

alyssais:
I'm not sure how that's related to this PR, though?

maxim-belkin:
not related. just thinking out loud 🤔 📢

ilovezfs:
@maxim-belkin basically orphaned if only the opt-linked kegs are considered vs. orphaned if all kegs are considered.

alyssais:
I think "orphaned if all kegs are considered" doesn't have a clear use right now. We can add it in the future if a need becomes apparent.

ilovezfs:
The use would to protect specs other than the currently opt-linked one. If I have both devel and stable installed, and devel is currently opt-linked, should orphaned report things that are still used by stable?

JCount:
@alyssais Good work on this so far! 😄 I was playing around with this and observed some behavior that is probably not what you intend.

Steps to reproduce:
Checkout this PR
```
$ brew install `gnupg`
==> Installing dependencies for gnupg: npth, libtasn1, gmp, nettle, libunistring, libffi,
p11-kit, gnutls, libgpg-error, libgcrypt, libksba, libassuan, pinentry, gettext, adns,
libusb
$ brew orphaned
libffi
p11-kit
```

I think this probably happens because `libffi` is a required dependency of `p11-kit`, which in turn is a **recommended** dependency of `gnutls` which is a required dependency of `gnupg`. Unfortunately, I don't have time to try and debug this in the code right now, but I thought I should mention it.

alyssais:
@JCount thanks! I think this is actually the same issue as above — do you have old versions of libﬃ and p11-kit installed?

JCount:
@alyssais No, this was done inside a clean VM.

alyssais:
@JCount the issue is because I was only checking whether one level up of a given dependency was installed on request, so `p11-kit` would be reported as orphaned because the only thing that depended on it, `gnutls`, wasn't installed on request. I should be able to fix this by recursively checking if the dependents were also orphaned. Thanks for finding that for me!

alyssais:
@JCount could you give that a try and let me know if it fixes your issue?

JCount:
@alyssais This appears to have successfully fixed the issue that was revealed by `gnupg`. 

Unfortunately, I have uncovered a separate, but similar, issue in the process of confirming the fix.

```
[Test:~] $ brew install jq
==> Installing dependencies for jq: oniguruma
==> Installing jq dependency: oniguruma
==> Downloading https://homebrew.bintray.com/bottles/oniguruma-6.2.0.sierra.bottle.tar.gz
Already downloaded: /Users/Test/Library/Caches/Homebrew/oniguruma-6.2.0.sierra.bottle.tar.gz
==> Pouring oniguruma-6.2.0.sierra.bottle.tar.gz
==> Using the sandbox
🍺  /usr/local/Cellar/oniguruma/6.2.0: 17 files, 1.3MB
==> Installing jq 
==> Downloading https://homebrew.bintray.com/bottles/jq-1.5_2.sierra.bottle.tar.gz
Already downloaded: /Users/Test/Library/Caches/Homebrew/jq-1.5_2.sierra.bottle.tar.gz
==> Pouring jq-1.5_2.sierra.bottle.tar.gz
🍺  /usr/local/Cellar/jq/1.5_2: 18 files, 958KB
[Test:~] $ brew orphaned
oniguruma
[Test:~] $ brew rm oniguruma
Error: Refusing to uninstall /usr/local/Cellar/oniguruma/6.2.0
because it is required by jq, which is currently installed.
You can override this and force removal with:
  brew uninstall --ignore-dependencies oniguruma
```

alyssais:
Thanks for letting me know. Here we go again! :D

maxim-belkin:
@alyssais these guys have no mercy :/

monouser7dig:
Dear bot 
This is in important addition and definitely not stale!

balthild:
So... Any updates for this?

alyssais:
Lots of difficult bugs to figure out, but I'm getting there — happy to explain what needs to be done to anyone who wants to lend a hand.

apjanke:
I'd like to help out here. What's outstanding that I could lend a hand with?

alyssais:
@apjanke [this](https://github.com/Homebrew/brew/pull/1881#issuecomment-299720606) issue is still outstanding

apjanke:
Okay. I'll take a look at it and see if I can help iron it out.

MikeMcQuaid:
I think it's worth trying to slim this PR down into a few PRs that can be merged/reviewed more easily.

MikeMcQuaid:
Closing this out as there's been no activity in a while and although I'd really like to see it it should probably be multiple, smaller PRs that get merged one at a time.

PR #1883, opened 2017-01-20
Allow overriding `MacOS.version`.

reitermarkus:

-----

Bringing back `MACOS_RELEASE` doesn't make sense, since we already deprecated `MacOS.release`. So should we add a new variable `MACOS_VERSION` or only allow overriding `HOMEBREW_MACOS_VERSION`, or both?

---

Fixes https://github.com/caskroom/homebrew-cask/issues/28843.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cli.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli.rb
@@ -142,6 +142,10 @@ module Hbc
     end
 
     def self.process(arguments)
+      unless ENV["MACOS_VERSION"].nil?
+        MacOS.full_version = ENV["MACOS_VERSION"]
+      end
+
       command_string, *rest = *arguments
       rest = process_options(rest)
       command = Hbc.help ? "help" : lookup_command(command_string)
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -27,6 +27,11 @@ module OS
       @full_version ||= Version.new((ENV["HOMEBREW_MACOS_VERSION"] || ENV["HOMEBREW_OSX_VERSION"]).chomp)
     end
 
+    def full_version=(version)
+      @full_version = Version.new(version.chomp)
+      @version = nil
+    end
+
     def prerelease?
       # TODO: bump version when new OS is released
       version >= "10.13"


MikeMcQuaid:
What's the reasoning for overriding it?

vitorgalvao:
> Bringing back `MACOS_RELEASE` doesn't make sense, since we already deprecated `MacOS.release`.

Agreed.

> So should we add a new variable `MACOS_VERSION` or only allow overriding `HOMEBREW_MACOS_VERSION`, or both?

No strong preference, but only allow overriding `HOMEBREW_MACOS_VERSION` might make more sense. It’s clear as to what exactly it is doing.

> What's the reasoning for overriding it?

Being able to accurately test/audit casks with conditionals ([example](https://github.com/caskroom/homebrew-cask/blob/411a04dbf859a5700bbe791a35fcf76655f6a40e/Casks/calibre.rb#L2)).

MikeMcQuaid:
> Being able to accurately test/audit casks with conditionals (example).

🆒. I'd rather see this more tightly scoped to something like `HOMEBREW_CASK_FAKE_MACOS_VERSION` and handled in Ruby-land.

reitermarkus:
To be honest, `HOMEBREW_CASK_FAKE_MACOS_VERSION` feels unnecessarily specific, seeing as `MacOS.version` isn't specific to Homebrew-Cask.

MikeMcQuaid:
> To be honest, HOMEBREW_CASK_FAKE_MACOS_VERSION feels unnecessarily specific, seeing as MacOS.version isn't specific to Homebrew-Cask.

Overriding it is, though. I definitely don't want to add an environment variable that allows this to be overridden globally.

reitermarkus:
Ok then, I am not sure how we should handle this. @vitorgalvao?

vitorgalvao:
Since overriding is more of a cask-specific feature, and to address @MikeMcQuaid’s concerns, I think `HOMEBREW_MACOS_VERSION` is out, then.

In that vein, `MACOS_VERSION` seems acceptable to me, because it is incredibly generic. To me it looks like the type of variable you’d expect some tools to support but others not, making it fine for use in HBC but not HB.

MikeMcQuaid:
If it's an override purely for testing I think using a name that's unlikely to be set by users in their environment is a better idea (i.e. `HOMEBREW_CASK_*)` but ultimately: up to you.

neutric:
Thanks guys, works a treat!

PR #1885, opened 2017-01-20
Remove some remnants of Ruby 1.8.7 support

alyssais:

-----

Since the emoji hack is no longer required, and `Emoji.tick` and `.cross` are only called in one place, I just inlined them so there weren't methods on `Emoji` that just return a one-character `String`.

### DIFF
--- a/Library/Homebrew/test/emoji_test.rb
+++ b/Library/Homebrew/test/emoji_test.rb
@@ -1,0 +1,11 @@
+require "testing_env"
+require "emoji"
+
+class EmojiTest < Homebrew::TestCase
+  def test_install_badge
+    assert_equal "🍺", Emoji.install_badge
+
+    ENV["HOMEBREW_INSTALL_BADGE"] = "foo"
+    assert_equal "foo", Emoji.install_badge
+  end
+end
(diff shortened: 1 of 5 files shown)


alyssais:
Added some new tests for the diff to make CodeCov happy.

Library/Homebrew/test/emoji_test.rb
```diff
@@ -0,0 +1,11 @@
+require "testing_env"
+require "emoji"
+
+class EmojiTest < Homebrew::TestCase
+  def test_install_badge
+    assert_equal "🍺", Emoji.install_badge
+
+    ENV.stubs(:[]).with("HOMEBREW_INSTALL_BADGE").returns("foo")
```
alyssais: When #1890 is merged, these stubs can be replaced with `ENV#[]=` calls.

Library/Homebrew/test/utils_test.rb
```diff
@@ -14,6 +14,14 @@ def teardown
     ENV.replace @env
   end
 
+  # Helper for matching escape sequences.
+  def e(code)
```
MikeMcQuaid: Even although this is a test, let's use a longer name. Generally I think short name plus comment is worth than long name without comment.

PR #1886, opened 2017-01-20
tests: add another missing cache clear.

alyssais:

-----

Another one of these! (See #1879)

We really need to come up with a better solution for this.


alyssais:
CI failing due to **another** one of these.

MikeMcQuaid:
If this is still needed: 👍.  I think would be good to pull all these into the `super` teardown, even when they aren't necessary. Would rather have a simple, comprehensive teardown that takes a little longer.

alyssais:
I agree.

PR #1888, opened 2017-01-21
tests: enforce `super` in lifecycle hooks

alyssais:

-----

Made all existing `setup` and `teardown` methods call `super`, and added a `LifecycleEnforcer` test plugin that will enforce this for new tests.

This will allow us to have global setup and teardown for tests.

For example, we can automatically clear caches after each test, to avoid annoying intermittent failures like #1879 and #1886.


MikeMcQuaid:
Great work. Feels like the best solution.

PR #1890, opened 2017-01-21
tests: automatically restore ENV in teardown

alyssais:

-----

I'm going to enjoy all of these more-deletions-than-additions PRs.


Library/Homebrew/test/support/helper/env.rb
```diff
@@ -1,14 +1,32 @@
...
+      def using_git_env
+        initial_env = copy_env
+        %w[AUTHOR COMMITTER].each do |role|
+          ENV["GIT_#{role}_NAME"] = "brew tests"
+          ENV["GIT_#{role}_EMAIL"] = "brew-tests@localhost"
+          ENV["GIT_#{role}_DATE"] = "Thu May 21 00:04:11 2009 +0100"
         end
+        yield
+      ensure
+        restore_env initial_env
```
reitermarkus: You could even do this:

```ruby 
def using_git_env
  git_env = ["AUTHOR", "COMMITTER"].each_with_object({}) do |role, env|
    env["GIT_#{role}_NAME"]  = "brew tests"
    env["GIT_#{role}_EMAIL"] = "brew-tests@localhost"
    env["GIT_#{role}_DATE"]  = "Sun Jan 22 01:50:11 2017 +0100"
  end

  with_environment(git_env) do
    yield
  end
end
```

Library/Homebrew/test/support/helper/test_case.rb
```diff
@@ -16,12 +16,17 @@ class TestCase < ::Minitest::Test
 
     def setup
       super
+
       @__argv = ARGV.dup
+      @__env = copy_env # Call #to_hash to duplicate ENV
```
MikeMcQuaid: Consider renaming the function to avoid the comment being needed

Library/Homebrew/test/support/helper/env.rb
```diff
@@ -1,13 +1,31 @@
...
-        old = ENV.to_hash
+        old = copy_env
         ENV.update partial_env
-        begin
+        yield
+      ensure
+        restore_env old
+      end
+
+      def using_git_env
```
MikeMcQuaid: Any downside to just setting this globally?

Library/Homebrew/test/support/helper/test_case.rb
alyssais: Oh, that's a leftover from when I inlined the `ENV.to_hash`. (I actually put it in a named function to avoid the need for the comment!) So I think I'll just remove the comment.

Library/Homebrew/test/support/helper/env.rb
alyssais: Probably not. Most of it [already is](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/tests.rb#L32-L37)

MikeMcQuaid: @alyssais If you get 💚 tests: let's just set globally 👍

Library/Homebrew/test/support/helper/test_case.rb
MikeMcQuaid: 🆒

Library/Homebrew/test/support/helper/env.rb
alyssais: All green apart from the `brew bundle` integration test, because `brew bundle` [requires](https://github.com/Homebrew/homebrew-bundle/blob/575804057c7195f647d747683cc4c1592f28417a/cmd/brew-bundle.rb#L36) a commit on or after `Wed Aug 17 11:07:17 2016 +0100`.

So, three options here:

1. Give up on removing this method
2. Delete the `GIT_COMMITTER_DATE` environment variable before this test.
3. Update the date used so it's more recent, and also update all of the expected commit hashes that rely on the old test.

I think I'd probably prefer 3. What do you think?

MikeMcQuaid: > Update the date used so it's more recent, and also update all of the expected commit hashes that rely on the old test.

Seems best to me too 👍

alyssais: Trivia I just discovered: the previous value was the timestamp of the initial commit to Homebrew.

Library/Homebrew/test/tap_test.rb
```diff
@@ -184,10 +175,10 @@ def test_git_variant
     touch @path/"README"
     setup_git_repo
 
-    assert_equal "e1893a6bd191ba895c71b652ff8376a6114c7fa7", @tap.git_head
-    assert_equal "e189", @tap.git_short_head
-    assert_match "years ago", @tap.git_last_commit
```
alyssais: I'll draw particular attention to this test, because it was the only one that needed any modification other than an updated SHA. (I had to remove the "years".)

alyssais:
Hmm… I _could_ write an integration test for `brew tests` to stop CodeCov complaining. Thoughts?

Library/Homebrew/test/tap_test.rb
reitermarkus: I think testing for `"ago"` is enough. Although, if you want to be absolutely sure:

```ruby
assert_match /\d+ .+ ago/, @tap.git_last_commit
```

Library/Homebrew/test/support/helper/env.rb
```diff
@@ -1,14 +1,20 @@
...
     module Env
+      def copy_env
+        ENV.to_hash
+      end
+
+      def restore_env(env)
+        ENV.replace(env)
+      end
+
       def with_environment(partial_env)
```
reitermarkus: We could even get rid of this entirely now, since we are only using this five times, with only one variable each.

MikeMcQuaid:
> Hmm… I could write an integration test for brew tests to stop CodeCov complaining. Thoughts?

@alyssais If you can get it to work and not be hilariously ill-performant (i.e. set it to run a single, fast test) then 👍

Library/Homebrew/test/support/helper/env.rb
MikeMcQuaid: 👍

Library/Homebrew/test/tap_test.rb
MikeMcQuaid: 👍

MikeMcQuaid:
@alyssais Actually: the best option at some point would be to attempt to push a bunch of the coverage logic into `brew test-bot` so that its run of e.g. `brew audit` etc. is counted in the coverage count (as it is being run every build).

Library/Homebrew/test/support/helper/env.rb
alyssais: @reitermarkus It's used a few times in the cask tests. Do they automatically reset ENV like the Homebrew tests now do, or would I have to add that?

alyssais: (Alternatively, I could get rid of all of Homebrew's uses of it and move it into cask.)

reitermarkus: > @reitermarkus It's used a few times in the cask tests. Do they automatically reset ENV like the Homebrew tests now do, or would I have to add that?

They do not, but I have a branch, if you want to cherry-pick that commit: https://github.com/reitermarkus/brew/commit/c7c2482caf4cb35e179a52c3171e11f05431a5bb

PR #1892, opened 2017-01-21
tests: clean up file system for all tests

alyssais:

-----

(No longer just integration tests.)

This was a big change, and I wanted to avoid removing any `ensure`/`teardown` code that would later end up needing to be added back in, so it's likely there'll be a few lines I could have removed but missed.

Similarly, I'm sure there are more directories that could be purged in teardown safely, but I decided to do an 80/20 and just copy the list from the integration tests for now.

### DIFF
--- a/Library/Homebrew/test/ARGV_test.rb
+++ b/Library/Homebrew/test/ARGV_test.rb
@@ -22,8 +22,6 @@ class ArgvExtensionTests < Homebrew::TestCase
     keg.mkpath
     @argv << "mxcl"
     assert_equal 1, @argv.kegs.length
-  ensure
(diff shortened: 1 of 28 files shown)


MikeMcQuaid:
Yes, so much deleted code! Any noticeable difference in speed?

alyssais:
Looks like it averages to about four seconds per full (`--official-cmd-taps`) test suite run.

alyssais:
(Where a full test suite run after this change takes 104 seconds)

MikeMcQuaid:
So a 4% regression? Seems fine given the code complexity and (moreso) the better errors we'll get on failures.

PR #1894, opened 2017-01-22
Add internal command to calculate appcast checkpoint.

reitermarkus:

-----

@vitorgalvao, I think (*finally*) implementing this is better than adding yet another workaround (`tr "\n"`).

Closes https://github.com/caskroom/homebrew-cask/issues/24811, https://github.com/caskroom/homebrew-cask/pull/28966 and https://github.com/Homebrew/brew/pull/1876.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
@@ -6,7 +6,18 @@ module Hbc
         cask_tokens = cask_tokens_from(args)
         raise CaskUnspecifiedError if cask_tokens.empty?
 
-        appcask_checkpoint(cask_tokens, calculate)
+        if cask_tokens.all? { |t| t =~ %r{^https?://} && t !~ /\.rb$/ }
+          appcask_checkpoint_for_url(cask_tokens)
+        else
+          appcask_checkpoint(cask_tokens, calculate)
+        end
+      end
+
+      def self.appcask_checkpoint_for_url(urls)
+        urls.each do |url|
+          appcast = DSL::Appcast.new(url)
+          puts appcast.calculate_checkpoint[:checkpoint]
+        end
       end
 
       def self.appcask_checkpoint(cask_tokens, calculate)
@@ -39,7 +50,7 @@ module Hbc
       end
 
       def self.help
-        "prints (no flag) or calculates ('--calculate') a given Cask's appcast checkpoint"
+        "prints (no flag) or calculates ('--calculate') a given Cask's (or URL's) appcast checkpoint"
       end
 
       def self.needs_init?
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
@@ -2,13 +2,14 @@ module Hbc
   class CLI
     class InternalAppcastCheckpoint < InternalUseBase
       def self.run(*args)
+        calculate = args.include? "--calculate"
         cask_tokens = cask_tokens_from(args)
         raise CaskUnspecifiedError if cask_tokens.empty?
 
-        appcask_checkpoint(cask_tokens)
+        appcask_checkpoint(cask_tokens, calculate)
       end
 
-      def self.appcask_checkpoint(cask_tokens)
+      def self.appcask_checkpoint(cask_tokens, calculate)
         count = 0
 
         cask_tokens.each do |cask_token|
@@ -17,9 +18,13 @@ module Hbc
           if cask.appcast.nil?
             opoo "Cask '#{cask}' is missing an `appcast` stanza."
           else
-            result = cask.appcast.calculate_checkpoint
+            if calculate
+              result = cask.appcast.calculate_checkpoint
 
-            checkpoint = result[:checkpoint]
+              checkpoint = result[:checkpoint]
+            else
+              checkpoint = cask.appcast.checkpoint
+            end
 
             if checkpoint.nil?
               onoe "Could not retrieve `appcast` checkpoint for cask '#{cask}': #{result[:command_result].stderr}"
@@ -34,7 +39,7 @@ module Hbc
       end
 
       def self.help
-        "calculates a given Cask's appcast checkpoint"
+        "prints (no flag) or calculates ('--calculate') a given Cask's appcast checkpoint"
       end
 
       def self.needs_init?
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
@@ -50,7 +50,7 @@ module Hbc
       end
 
       def self.help
-        "prints (no flag) or calculates ('--calculate') a given Cask's (or URL's) appcast checkpoint"
+        "prints or calculates a given Cask's or URL's appcast checkpoint"
       end
 
       def self.needs_init?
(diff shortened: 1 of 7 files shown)


Library/Homebrew/cask/lib/hbc/dsl/appcast.rb
```diff
@@ -9,6 +11,24 @@ def initialize(uri, parameters = {})
         @checkpoint     = @parameters[:checkpoint]
       end
 
+      def calculate_checkpoint
+        result = SystemCommand.run("/usr/bin/curl", args: ["--compressed", "--location", "--user-agent", URL::FAKE_USER_AGENT, @uri], print_stderr: false)
+
+        checkpoint = if result.success?
+          processed_appcast_text = result.stdout.gsub(%r{<pubDate>[^<]*</pubDate>}, "")
+
+          # This step is necessary to replicate running `sed` from the command line
```
vitorgalvao: I’d say these can be removed. If this exists to give the same result the non-internal command does, but the non-internal command will ceased to be used in favour of this internal one, then there’s nothing to replicate. This will be the canonical and only recommended way to do it.

vitorgalvao:
Might it be worth adding mandatory flags to this? Something like `--calculate` and `--print` (gives the current one)?

Library/Homebrew/cask/lib/hbc/dsl/appcast.rb
reitermarkus: I just thought I'd leave this in for now as to not mess up existing values. But I guess this doesn't make any difference since it isn't officially used for anything yet.

vitorgalvao: Just noticed it also needs the `delete("\n")` step before `gsub`.

There’s no harm in messing with current values, since those will be outdated anyway, for missing the `delete` step.

reitermarkus: Isn't `delete("\n")` only needed because `sed` only handles one line at a time? Do we need it then, when we aren't going to use `sed` anymore?

vitorgalvao: The problem is that some casks have `<pubDate>` and `</pubDate>` on different lines, so it doesn’t actually get removed. By removing all newlines, we make sure `<pubDate>[^<]*</pubDate>` always works.

I guess we can just setup the regex to work multi-line, yes.

reitermarkus: Ah, right, I forgot `gsub` still needs the `m` option to be multi-line.

vitorgalvao:
@reitermarkus Instead of adding mandatory flags, it seemed sensible to make it just print the current appcast when no flags given, or calculate it when given `--calculate`.

Was going to submit it as a PR to your fork, but GitHub was erroring and not cooperating, so it was easier to add the commit directly: https://github.com/Homebrew/brew/pull/1894/commits/949d3deb9b8fb8ea362f6c73da80b56c4af6197e.

Looks good to me as is, unless you want to change something/disagree.

vitorgalvao:
While editing the documentation, just recalled something else. We can’t just take casks as arguments, we also need to take URLs (so users can calculate it the first time, when building the cask).

Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
```diff
@@ -2,13 +2,14 @@ module Hbc
   class CLI
     class InternalAppcastCheckpoint < InternalUseBase
       def self.run(*args)
-        cask_tokens = cask_tokens_from(args)
+        calculate = args.include? "--calculate"
+        cask_tokens = args.reject { |arg| arg.chars.first == "-" }
```
reitermarkus: Does `cask_tokens_from(args)` not work here, or was this overlooked?

vitorgalvao: It was overlooked. It does indeed work, but I was basing myself on `_stanza` and didn’t even notice. Will fix it and add to https://github.com/Homebrew/brew/pull/1896 as well.

Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
```diff
@@ -0,0 +1,61 @@
...
+        raise CaskUnspecifiedError if cask_tokens.empty?
+
+        if cask_tokens.all? { |t| t =~ %r{^https?://} && t !~ /\.rb$/ }
+          appcask_checkpoint_for_url(cask_tokens)
+        else
+          appcask_checkpoint(cask_tokens, calculate)
+        end
+      end
+
+      def self.appcask_checkpoint_for_url(urls)
```
vitorgalvao: This will work the same both with or without `--calculate`. Just pointing it out in case you intended to do it any other way. It’s fine by me as it is since the only sensical thing to do with a URL is calculate, anyway.

Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
```diff
@@ -0,0 +1,61 @@
...
+              count += 1
+            end
+          end
+        end
+
+        count == cask_tokens.count
+      end
+
+      def self.help
+        "prints (no flag) or calculates ('--calculate') a given Cask's appcast checkpoint"
```
vitorgalvao: We should add a note about it working on URLs as well.

reitermarkus: Let's maybe add this to the man page instead, otherwise the line is getting kinda long.

Library/Homebrew/cask/lib/hbc/cli/internal_appcast_checkpoint.rb
reitermarkus: > the only sensical thing to do with a URL is calculate

Exactly, you can't not calculate it.

reitermarkus:
@vitorgalvao, I added a manpage entry and shortened the help text a bit.

vitorgalvao:
I’d say it’s ready to merge as soon as all checks pass.

toonetown:
So - does this use a different method of calculating the hash of the app cast?  For example, if I run (after updating)

```bash
$ brew cask _appcast_checkpoint marked
caf003f39ca2a87922e2d1f6afe17df55bd6bb6f36dcfd16b52e257cf629f5e0
$ brew cask _appcast_checkpoint --calculate marked
f315ff8dfb5cbd0e1da1ef302c6edbcb738d083237ebc4b977e78a1c6ce97fd3
```

However - the hash (using cURL) was previously calculated thusly:

```bash
$ curl --compressed --location --user-agent 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/33.0.1750.152 Safari/537.36' "https://updates.marked2app.com/marked.xml" | /usr/bin/sed 's|<pubDate>[^<]*</pubDate>||g' | shasum --algorithm 256
caf003f39ca2a87922e2d1f6afe17df55bd6bb6f36dcfd16b52e257cf629f5e0
```

Which is the current (in the cask file) hash.  I'm fine if the internal call changes the way that the hash is calculated - but is there a separate task to update all current casks with their new hashes?

vitorgalvao:
> does this use a different method of calculating the hash of the app cast?

Yes. The change was necessary anyway as the old method wasn’t reliable on all cases.

>  is there a separate task to update all current casks with their new hashes?

No. That is unnecessary. Only now that this is finalised will I start working on remaking the scripts to fix casks based on appcast checkpoints. Those will end up being fixed organically.

PR #1895, opened 2017-01-22
bottle: fix regex

xu-cheng:
Noted that this is intended for the revision of this formula instead of
rebuild of a bottle.

-----------------

Sorry of being missing for long time.
But, I think I should start to find some time to contribute Homebrew again
from next month or so.


MikeMcQuaid:
Hey @xu-cheng. This seems correct but can you point to e.g. a formula PR or anything that made you spot this? Thanks.

xu-cheng:
While, during my absence, I sometimes will take a little time to follow Homebrew's current changes.

I accidentally spot this when scanning [a recent commit](https://github.com/Homebrew/brew/commit/42486c1181bdf4ed85f334ccb1edbb6632cfc4b2).

MikeMcQuaid:
Cool, thanks @xu-cheng!

PR #1912, opened 2017-01-25
Update outdated link

omegahm:
Old link gives SSL error.

Editing old link to same without `https` redirects to `http://docs.brew.sh`, which I assume is the correct link.


MikeMcQuaid:

PR #1913, opened 2017-01-25
Use array of constants to set up test directories.

reitermarkus:

-----

Extracted from https://github.com/Homebrew/brew/pull/1314 for easier review.

Move creation of test directories into `Homebrew::TestCase` and use the existing constants instead of duplicating them with an array of strings.

Also updates `minitest` and adds `minitest-reporters` to assimilate with Cask's `test_helper`.

### DIFF
--- a/Library/Homebrew/test/support/helper/test_case.rb
+++ b/Library/Homebrew/test/support/helper/test_case.rb
@@ -12,6 +12,20 @@ module Homebrew
     TEST_SHA1   = "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef".freeze
     TEST_SHA256 = "deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef".freeze
 
+    def before_setup
+      [
+        HOMEBREW_LIBRARY/"Taps/homebrew/homebrew-core/Formula",
+        HOMEBREW_CACHE,
+        HOMEBREW_CACHE_FORMULA,
+        HOMEBREW_LOCK_DIR,
+        HOMEBREW_CELLAR,
+        HOMEBREW_LOGS,
+        HOMEBREW_TEMP,
+      ].each(&:mkpath)
+
+      super
+    end
+
     def setup
       super
 
@@ -23,6 +37,10 @@ module Homebrew
       ARGV.replace(@__argv)
       ENV.replace(@__env)
 
+      super
+    end
+
+    def after_teardown
       Tab.clear_cache
 
       coretap = CoreTap.new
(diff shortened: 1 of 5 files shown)


Library/Homebrew/test/support/helper/test_case.rb
```diff
@@ -12,6 +12,20 @@ class TestCase < ::Minitest::Test
     TEST_SHA1   = "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef".freeze
     TEST_SHA256 = "deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef".freeze
 
+    def before_setup
```
alyssais: Can we use `setup` instead of `before_setup`?

Minitest has this to say, which is why I went to great lengths to avoid them:

> These are meant for library writers, NOT for regular test authors.

Library/Homebrew/test/support/helper/test_case.rb
```diff
@@ -12,6 +12,20 @@ class TestCase < ::Minitest::Test
...
+      [
```
alyssais: The array itself might as well be constant too, if we're doing this.

Library/Homebrew/test/support/helper/test_case.rb
reitermarkus: You know, I tried about an hour to get this to work, but now that I think about it again, the problem is probably that `FSLeakLogger` is using these also instead of `setup` and `teardown`.

alyssais: Heh, I hate it when I spend ages trying and failing to work out a problem, then come back to it hours later and immediately get it. But also I love it. 😁

Library/Homebrew/test/support/helper/test_case.rb
```diff
@@ -13,6 +13,16 @@ class TestCase < ::Minitest::Test
...
     def setup
+      [
+        HOMEBREW_LIBRARY/"Taps/homebrew/homebrew-core/Formula",
+        HOMEBREW_CACHE,
+        HOMEBREW_CACHE_FORMULA,
+        HOMEBREW_LOCK_DIR,
+        HOMEBREW_CELLAR,
+        HOMEBREW_LOGS,
+        HOMEBREW_TEMP,
+      ].each(&:mkpath)
```
alyssais: Might be worth having a comment explaining that this needs to be done before `super` because of `FSLeakLogger`?

Library/Homebrew/test/support/lib/config.rb
```diff
@@ -15,21 +15,21 @@
 
 # Paths pointing into the Homebrew code base that persist across test runs
 HOMEBREW_LIBRARY_PATH  = Pathname.new(File.expand_path("../../../..", __FILE__))
-HOMEBREW_SHIMS_PATH    = HOMEBREW_LIBRARY_PATH.parent+"Homebrew/shims"
+HOMEBREW_SHIMS_PATH    = HOMEBREW_LIBRARY_PATH.parent.join("Homebrew", "shims")
```
MikeMcQuaid: Leave these as they were; we don't want to use `.join` for paths

PR #1934, opened 2017-01-30
cleanup: consider version_scheme

xu-cheng:
`brew cleanup` should cleanup formulae which have the older
version scheme.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1843,7 +1843,16 @@ class Formula
       eligible_kegs = if head? && (head_prefix = latest_head_prefix)
         installed_kegs - [Keg.new(head_prefix)]
       else
-        installed_kegs.select { |k| pkg_version > k.version }
+        installed_kegs.select do |keg|
+          tab = Tab.for_keg(keg)
+          if version_scheme > tab.version_scheme
+            true
+          elsif version_scheme == tab.version_scheme
+            pkg_version > keg.version
+          else
+            false
+          end
+        end
       end
 
       unless eligible_kegs.empty?
--- a/Library/Homebrew/test/cleanup_test.rb
+++ b/Library/Homebrew/test/cleanup_test.rb
@@ -36,25 +36,40 @@ class CleanupTests < Homebrew::TestCase
   end
 
   def test_cleanup_formula
-    f1 = Class.new(Testball) { version "0.1" }.new
-    f2 = Class.new(Testball) { version "0.2" }.new
-    f3 = Class.new(Testball) { version "0.3" }.new
+    f1 = Class.new(Testball) do
+      version "1.0"
+    end.new
+    f2 = Class.new(Testball) do
+      version "0.2"
+      version_scheme 1
+    end.new
+    f3 = Class.new(Testball) do
+      version "0.3"
+      version_scheme 1
+    end.new
+    f4 = Class.new(Testball) do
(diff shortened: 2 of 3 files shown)


Library/Homebrew/formula.rb
```diff
@@ -1843,7 +1843,12 @@ def eligible_kegs_for_cleanup
       eligible_kegs = if head? && (head_prefix = latest_head_prefix)
         installed_kegs - [Keg.new(head_prefix)]
       else
-        installed_kegs.select { |k| pkg_version > k.version }
+        installed_kegs.select do |keg|
+          tab = Tab.for_keg(keg)
```
MikeMcQuaid: Maybe worth a `nil` check.

Library/Homebrew/formula.rb
```diff
@@ -1843,7 +1843,12 @@ def eligible_kegs_for_cleanup
...
+          next true if version_scheme > tab.version_scheme
```
MikeMcQuaid: Think it might be more readable to do a `if`, `elseif`, `else` on `>`, `==`, `<`

Library/Homebrew/formula.rb
xu-cheng: It won't be nil, as it has been handled by `Tab.for_keg`.

Library/Homebrew/formula.rb
xu-cheng: Like
```rb
          if version_scheme > tab.version_scheme
            true
          elsif version_scheme == tab.version_scheme
            pkg_version > keg.version
          else
            false
          end
```
??

Library/Homebrew/formula.rb
MikeMcQuaid: Good point. I think `version_scheme` could be, worth checking.

Library/Homebrew/formula.rb
MikeMcQuaid: Yeh 👍

Library/Homebrew/formula.rb
xu-cheng: `version_scheme` won't be nil as well. It defaults to 0.

MikeMcQuaid: That's the default but I was struggling to see where it's set to `0` if it's not present in an (old) tab? Happy to defer to you here though if you're sure.

xu-cheng: FYI https://github.com/Homebrew/brew/blob/master/Library/Homebrew/tab.rb#L305-L307

DomT4:
Thanks for fixing this! Noticed this the other day and flagged it to ILZ, but hadn't looked at it in-depth myself. Appreciated 🙇.

MikeMcQuaid:
> Noticed this the other day and flagged it to ILZ

@DomT4 Gentle request to create issues for such things 😉

DomT4:
I _try_ to avoid anything that may prove complicated or controversial around here these days, since my opinions on a lot of things are strong but I gave up trying to ram the rest of you into them 😸.

MikeMcQuaid:
@DomT4 Sure but in the cases like this where it's pretty unambiguously a bug it'd be appreciated 😉

DomT4:
One human's bug is another human's feature 😝, but I'll try to nudge myself in future.

MikeMcQuaid:
Cheers @DomT4.

PR #1948, opened 2017-02-03
gcc_version_formula: Use gcc@4.x rather than gcc4x

sjackman:
This PR modifies the compiler selection logic to use `Homebrew/core/gcc@4.x` rather than `Homebrew/versions/gcc4x`.

I haven't yet had a chance to test this PR. I wanted to first get feedback that this approach is correct.

### DIFF
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -269,7 +269,7 @@ module SharedEnvExtension
   # @private
   def gcc_version_formula(name)
     version = name[GNU_GCC_REGEXP, 1]
-    gcc_version_name = "gcc#{version.delete(".")}"
+    gcc_version_name = "gcc@#{version}"
 
     gcc = Formulary.factory("gcc")
     if gcc.version_suffix == version
@@ -286,7 +286,6 @@ module SharedEnvExtension
     rescue FormulaUnavailableError => e
       raise <<-EOS.undent
       Homebrew GCC requested, but formula #{e.name} not found!
-      You may need to: brew tap homebrew/versions
       EOS
     end
 


Library/Homebrew/compilers.rb
```diff
@@ -1,7 +1,7 @@
 # @private
 module CompilerConstants
-  GNU_GCC_VERSIONS = %w[4.3 4.4 4.5 4.6 4.7 4.8 4.9 5 6 7].freeze
-  GNU_GCC_REGEXP = /^gcc-(4\.[3-9]|[5-7])$/
+  GNU_GCC_VERSIONS = %w[4.6 4.7 4.8 4.9 5 6 7].freeze
+  GNU_GCC_REGEXP = /^gcc-(4\.[6-9]|[5-7])$/
```
MikeMcQuaid: Let's leave these as-is unless they break something as people may still want to use their own versions of the old formulae.

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -269,7 +269,7 @@ def ld64
   # @private
   def gcc_version_formula(name)
     version = name[GNU_GCC_REGEXP, 1]
-    gcc_version_name = "gcc#{version.delete(".")}"
+    gcc_version_name = "gcc@#{version}"
```
MikeMcQuaid: Looks right 👍

Library/Homebrew/compilers.rb
sjackman: Done.

sjackman:
I've updated the error message as well. Examples of the errors now:

```
❯❯❯ brew install hello --cc=gcc-4.1
Error: Invalid value for --cc: gcc-4.1
❯❯❯ brew install hello --cc=gcc-4.3
==> Using the sandbox
Error: Compiler gcc-4.3 requested, but formula gcc@4.3 not found!
The tap Homebrew/core does not provide gcc@4.3, gcc@4.4 nor gcc@4.5.
❯❯❯ brew install hello --cc=gcc-4.6
Error: The requested Homebrew GCC was not installed. You must:
  brew install gcc@4.6
```

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -285,8 +285,8 @@ def warn_about_non_apple_gcc(name)
       gcc_formula = gcc_version_formula(name)
     rescue FormulaUnavailableError => e
       raise <<-EOS.undent
-      Homebrew GCC requested, but formula #{e.name} not found!
-      You may need to: brew tap homebrew/versions
+      Compiler #{e.name.sub("@", "-")} requested, but formula #{e.name} not found!
+      The tap Homebrew/core does not provide gcc@4.3, gcc@4.4 nor gcc@4.5.
```
MikeMcQuaid: I think it's probably worth just omitting this line.

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -285,8 +285,8 @@ def warn_about_non_apple_gcc(name)
       gcc_formula = gcc_version_formula(name)
     rescue FormulaUnavailableError => e
       raise <<-EOS.undent
-      Homebrew GCC requested, but formula #{e.name} not found!
-      You may need to: brew tap homebrew/versions
+      Compiler #{e.name.sub("@", "-")} requested, but formula #{e.name} not found!
```
MikeMcQuaid: Would this ever be not GCC? If not I'd leave the old message but just add the name/version.

MikeMcQuaid:
CI failing on Linux: https://travis-ci.org/Homebrew/brew/jobs/198586674

Library/Homebrew/extend/ENV/shared.rb
sjackman: Okay.

sjackman:
I'm heading to a conference for a week. I can look into the Linux CI failure once I'm back.

sjackman:
I've made the requested changes. I haven't yet looked into the Linux CI failure.

sjackman:
CI is passing now on Linux. I didn't change anything myself to make that happen.

MikeMcQuaid:
@sjackman Weird, thanks!

sjackman:
Thanks for merging, Mike!

PR #1950, opened 2017-02-03
brew.sh: check for git repository first

maxim-belkin:


alyssais:
What problem is this attempting to solve? Would you mind filling in the [pull request template](https://raw.githubusercontent.com/Homebrew/brew/master/.github/PULL_REQUEST_TEMPLATE.md)? Thanks.

maxim-belkin:
There is no git repo when `brew` is installed by downloading `master.zip` from GitHub. In such a case, the first time user runs `brew`, it initializes an empty git repository which you want to avoid giving users a chance to do:
```
git remote add origin https://github.com/Homebrew/brew.git
```
In fact, there are couple of other places (in `brew.sh`) with the same behavior. Perhaps, it would make sense to inform the user of the problem (no git repository) and consequences (no updates) but not to run `git <something>` in a folder without repository. 

The lines that follow the code that I updated in this PR assume that "non-repositorial" cases are expected.

Library/Homebrew/brew.sh
```diff
@@ -1,5 +1,8 @@
-HOMEBREW_VERSION="$(git -C "$HOMEBREW_REPOSITORY" describe --tags --dirty 2>/dev/null)"
-HOMEBREW_USER_AGENT_VERSION="$HOMEBREW_VERSION"
+if [[ $(git -C "$HOMEBREW_REPOSITORY" status) ]]
```
alyssais: Running <kbd>git status</kbd> seems a little overkill. Wouldn't it be enough to check if `"$HOMEBREW_REPOSITORY"/.git` exists?

alyssais:
@maxim-belkin makes sense. Thanks for the explanation. 👍

Library/Homebrew/brew.sh
maxim-belkin: doh!

maxim-belkin: do we want to check `$GIT_DIR` here as well? or `.git` only?

MikeMcQuaid:
What's the user visible issue here?

maxim-belkin:
> What's the user visible issue here?

I tried to describe the issue in my comment above https://github.com/Homebrew/brew/pull/1950#issuecomment-277382079 so I'm not really sure what else to say... What exactly are you asking about?

MikeMcQuaid:
> it initializes an empty git repository

What does, `git describe`?

> I tried to describe the issue in my comment above #1950 (comment) so I'm not really sure what else to say... What exactly are you asking about?

I'm not sure if you were saying something about the version? If the issue is just "we initialise a Git repository when there isn't one" I'd consider that an accidental feature rather than bug.

maxim-belkin:
Whoops! Originally I thought that `git describe` creates an empty repository. Now that I've done some digging I see that it's not the case. The code that actually initializes a **non**-empty repository is [here](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/update.sh#L41-L49)

Sorry, guys! While I think the linked code should not be there (I don't see any reason to do that without user's permission) I can work around it by `mkdir "$(brew --repo)/.git` :)

thank you!

PR #1954, opened 2017-02-05
Refactor Cask `Artifact`s.

reitermarkus:

-----

Remove no-op `install_phase` and `uninstall_phase` methods and use `respond_to?` instead.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/zap.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/zap.rb
@@ -3,17 +3,8 @@ require "hbc/artifact/uninstall_base"
 module Hbc
   module Artifact
     class Zap < UninstallBase
-      def install_phase
-        odebug "Nothing to do. The zap artifact has no install phase."
-      end
-
       def uninstall_phase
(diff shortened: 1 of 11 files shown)


Library/Homebrew/cask/lib/hbc/artifact/zap.rb
```diff
@@ -3,17 +3,8 @@
...
   module Artifact
     class Zap < UninstallBase
-      def install_phase
-        odebug "Nothing to do. The zap artifact has no install phase."
-      end
-
       def uninstall_phase
-        odebug "Nothing to do. The zap artifact has no uninstall phase."
-      end
-
```
coeligena: It looks like this method has a function though it's overriding as an empty. Without this, uninstall commands function like zap commands because it looks for the Artifacts containing uninstall_phase [installer.rb: ... Artifact.for_cask( ... .respond_to?(:uninstall_phase) ...] then executes them including the zap artifact. When called via uninstall command, this artifact should do nothing as before. Zap function inside the installer.rb file should also call zap_phase instead of uninstall_phase.

reitermarkus:
@coeligena, thanks for pointing this out.

PR #1955, opened 2017-02-05
audit: enforce https://*.sourceforge.io/ homepages

vszakats:
Ref: https://sourceforge.net/blog/introducing-https-for-project-websites/

-----

I tested the new condition/suggestion in a minimised test source.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -584,6 +584,10 @@ class FormulaAuditor
       problem "Please use https:// for #{homepage}"
     end
 
+    if homepage =~ %r{^http://([^/]*)\.(sf|sourceforge)\.net(/|$)}
+      problem "#{homepage} should be `https://#{$1}.sourceforge.io/`"
+    end
+
     # There's an auto-redirect here, but this mistake is incredibly common too.
     # Only applies to the homepage and subdomains for now, not the FTP URLs.
     if homepage =~ %r{^http://((?:build|cloud|developer|download|extensions|git|glade|help|library|live|nagios|news|people|projects|rt|static|wiki|www)\.)?gnome\.org}
--- a/Library/Homebrew/test/audit_test.rb
+++ b/Library/Homebrew/test/audit_test.rb
@@ -431,6 +431,9 @@ class FormulaAuditorTests < Homebrew::TestCase
       "corge" => "http://savannah.nongnu.org/corge",
       "grault" => "http://grault.github.io/",
       "garply" => "http://www.gnome.org/garply",
+      "sf1" => "http://foo.sourceforge.net/",
+      "sf2" => "http://foo.sourceforge.net",
+      "sf3" => "http://foo.sf.net/",
       "waldo" => "http://www.gnu.org/waldo",
     }
 
@@ -455,6 +458,8 @@ class FormulaAuditorTests < Homebrew::TestCase
         end
       elsif homepage =~ %r{https:\/\/code\.google\.com}
         assert_match "#{homepage} should end with a slash", fa.problems.first
+      elsif homepage =~ /foo\.(sf|sourceforge)\.net/
+        assert_match "#{homepage} should be `https://foo.sourceforge.io/`", fa.problems.first
       else
         assert_match "Please use https:// for #{homepage}", fa.problems.first
       end


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -584,6 +584,12 @@ def audit_homepage
       problem "Please use https:// for #{homepage}"
     end
 
+    case homepage
+    when %r{^http://([^/]*)sourceforge\.net(/|$)}i,
+         %r{^http://([^/]*)sf\.net(/|$)}i
+      problem "#{homepage} should be `https://#{$1}sourceforge.io/`"
```
maveonair: Regarding the other existing [checks](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L553:L560), you could use the following problem message: 

`problem "Please use https:// for #{homepage}"`

to simply extend the existing [tests](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/audit_test.rb#L426:L434) with:

```ruby
    formula_homepages = {
      # other sites
      "projectlibre" => "http://sourceforge.net/projects/projectlibre/",
      "sevenzup" => "http://sf.net/projects/sevenzip/"     
    }
```

vszakats: Thanks for the test pointers, I've added some.
The problem message would be misleading as the domain needs to be updated as well.

maveonair: Yes, makes sense. But in this case you should add an additional assertion for your test case such as it is done here: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/audit_test.rb#L456 because the tests are now failing.

vszakats: I don't undertstand how the audit test is supposed to work. Can you help?

vszakats: I tried adding a line to catch results for new `foo.*` URLs.

maveonair:
This is not a problem and you did great work. 

Here is my suggestion to your contribution:

```ruby
# ...
elsif (homepage.include?("sf.net") || homepage.include?("sourceforge.net"))
  assert_match "#{homepage} should be `https://foo.sourceforge.io/`", fa.problems.first
else
# ...
```

Unfortunately, I had not yet the chance to run the tests to verify this approach. But the idea here is to catch the homepages that include the domains `sf.net` and `sourceforge.net` and to assert the problem statement.

vszakats:
Thank you @maveonair, that looks very good. I've updated the patch. Let's hope it passes.

maveonair:
@vszakats You are welcome. Thank you for your contribution  👍 .

reitermarkus:
According to [the documentation](https://sourceforge.net/p/forge/documentation/Project%20Web%20Services/), only *new* projects will use `https` by default, existing ones have to switch to `https` manually.

This means that `https://*.sourceforge.io` will still redirect to `http://*.sourceforge.net` for those that haven't switched to `https`.

vszakats:
@reitermarkus, Yes, I'm aware. Anyhow it's best to use the secure entry point by default and let the rest be sorted out by the pages themselves. Any redirection (to an insecure page) will be performed in the browser, so it will be visible to the user.

vszakats:
Is there anything that needs to be done in this PR to merit a merge?

Library/Homebrew/dev-cmd/audit.rb
reitermarkus: The regexes should probably be:

```ruby
%r{^http://([^/]*\.)?sourceforge\.net(/|$)}i

%r{^http://([^/]*\.)?sf\.net(/|$)}i

```

Otherwise `unrelated-sourceforge.net` and `unrelated-sf.net` would also be matched.

vszakats: Good point @reitermarkus, thank you! I've committed this fix.

vszakats:
Commits squashed.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -584,6 +584,12 @@ def audit_homepage
       problem "Please use https:// for #{homepage}"
     end
 
+    case homepage
+    when %r{^http://([^/]*)\.sourceforge\.net(/|$)}i,
+         %r{^http://([^/]*)\.sf\.net(/|$)}i
+      problem "#{homepage} should be `https://#{$1}.sourceforge.io/`"
+    end
```
MikeMcQuaid:
```ruby
if homepage =~ %r{^http://([^/]*)\.(sf|sourceforge)\.net(/|$)}
  problem "#{homepage} should be `https://#{$1}.sourceforge.io/`"
end
```

vszakats: 👍 Patch updated.

vszakats:
CI failure - unrelated:
```
error: RPC failed; curl 56 SSLRead() return error -36
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
Error: Failure while executing: git clone https://github.com/Homebrew/homebrew-core /Users/travis/build/Homebrew/brew/Library/Taps/homebrew/homebrew-core -q
```

vszakats:
Required test passed.

Library/Homebrew/test/audit_test.rb
```diff
@@ -455,6 +458,9 @@ class #{Formulary.class_s(name)} < Formula
         end
       elsif homepage =~ %r{https:\/\/code\.google\.com}
         assert_match "#{homepage} should end with a slash", fa.problems.first
+      elsif homepage.include?("foo.sourceforge.net") || \
+            homepage.include?("foo.sf.net")
```
reitermarkus: Could also make this `homepage =~ /foo\.(sf|sourceforge)\.net/`.

vszakats: Done.

PR #1957, opened 2017-02-06
Update docs for recently deprecated taps.

neutric:
Update for recently deprecated taps and fix link so that it is clickable on docs.brew.sh. Rewording of headline from "Main Taps" to "Official Taps" and addition of the two missing official taps to make list consistent with the taps `brew search` automatically uses.


MikeMcQuaid:
I'll leave this PR open for now as there's going to be quite a few more taps deprecated in the coming weeks so we may want to reconsider how we handle this whole section.

MikeMcQuaid:
Cheers for this @neutric but handled more widely in https://github.com/Homebrew/brew/pull/2392.

PR #1962, opened 2017-02-07
download_strategy: broaden https svn support/usage

DomT4:
- [ ] Have you written new tests for your changes? _Don't see much worthwhile to do on this, since CI will only have the system SVN installed._

-----

Homebrew, for a hilarious amount of years, has gone back and forth on which protocol to use for the Sourceforge SVN checkouts. This is complicated by the system SVN, out-of-the-box, being garbage at handling secure connections, ending up more often than not in situations like this:
```
~> brew fetch gpsim -v --HEAD
==> Cloning https://svn.code.sf.net/p/gpsim/code/trunk
...
Error validating server certificate for 'https://svn.code.sf.net:443':
 - The certificate is not issued by a trusted authority. Use the
   fingerprint to validate the certificate manually!
Certificate information:
 - Hostname: *.code.sf.net
 - Valid: from Apr  6 00:00:00 2016 GMT until Jun  5 23:59:59 2017 GMT
 - Issuer: GeoTrust SSL CA - G3, GeoTrust Inc., US
 - Fingerprint: 9C:63:DE:5C:24:C4:86:ED:D2:53:26:E1:26:6B:52:0D:DD:B0:95:A3
(R)eject, accept (t)emporarily or accept (p)ermanently?
```
This happens a fair bit with SVN `HEAD` urls in core for example because those aren't being CI checked and flagged as problematic. Not such an issue because really, the number of users ever using HEAD is limited.

However, there is very infrequently if ever a problem with HTTPS SVN checkouts using Homebrew's SVN, or even a custom SVN linked to a newish OpenSSL with an up-to-date knowledge of the most common CA root certs. Using/abusing this knowledge, we can silently rewrite insecure URLs to use HTTPS when someone has Homebrew's own SVN installed/has specified using an alternative SVN with the `HOMEBREW_SVN` env variable.

My personal temptation given Homebrew is slowly abandoning the system tools as time goes on is just to abandon the system SVN, but I don't hold hope of nudging that through remaining maintainers so this seems a reasonable alternative which remaining maintainers might not actually completely dislike.

This can likely be expanded to other URLs but Sourceforge is the most common use case and is used for some actual stable/devel URLs rather than just HEAD and consequently has more user exposure and more user exposure to plaintext connections.

Realistically, until Homebrew reaches a point where it moves away from the system SVN, which may very well be never, this is a problem that cannot be accurately audited, but can be caught like this and "fixed".

Tested with Homebrew's vanilla SVN installed against variety of stable/head formulae:
```
==> Cloning https://svn.code.sf.net/p/netpbm/code/advanced
==> Checking out 2825
==> Cloning https://svn.code.sf.net/p/astyle/code/trunk/AStyle
==> Cloning https://svn.code.sf.net/p/ctags/code/trunk
==> Cloning https://svn.code.sf.net/p/dasm-dillon/code/trunk
==> Cloning https://svn.code.sf.net/p/gpsim/code/trunk
==> Cloning https://svn.code.sf.net/p/syntaxhighlight/code/highlight/
==> Cloning https://svn.code.sf.net/p/netpbm/code/trunk
==> Cloning https://svn.code.sf.net/p/quex/code/trunk
```

And without:
```
==> Cloning http://svn.code.sf.net/p/netpbm/code/advanced
==> Checking out 2825
==> Cloning svn://svn.code.sf.net/p/astyle/code/trunk/AStyle
==> Cloning http://svn.code.sf.net/p/ctags/code/trunk
==> Cloning svn://svn.code.sf.net/p/dasm-dillon/code/trunk
==> Cloning svn://svn.code.sf.net/p/gpsim/code/trunk
==> Cloning svn://svn.code.sf.net/p/syntaxhighlight/code/highlight/
==> Cloning http://svn.code.sf.net/p/netpbm/code/trunk
==> Cloning http://svn.code.sf.net/p/quex/code/trunk
```

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -640,6 +640,14 @@ class SubversionDownloadStrategy < VCSDownloadStrategy
   def initialize(name, resource)
     super
     @url = @url.sub("svn+http://", "")
+
+    # System SVN tends to barf on https, but brewed SVN should consistently
+    # work due to being linked to brewed OpenSSL & generated PEM, so we can
+    # provide automatic upgrades to https where possible.
+    if !ENV["HOMEBREW_SVN"].nil? || which("svn").to_s.include?(HOMEBREW_PREFIX) &&
+                                    @url =~ /svn\.code\.sf\.net/
+      @url = @url.sub(%r{^(http://|svn://)}, "https://")
+    end
   end
 
   def fetch


DomT4:
I haven't submitted an annoyingly controversial PR for a while, so, here ya go 😄.

MikeMcQuaid:
> This is complicated by the system SVN, out-of-the-box, being garbage at handling secure connections, ending up more often than not in situations like this

Can you explain more why this is the case?

DomT4:
As much as I can, I'll take a swing at it, but having looked for answers on this repeatedly over the years most of the stuff available [seems](https://serverfault.com/a/229684) to be a vaguely [confused](https://apple.stackexchange.com/questions/94573/server-certificates-are-not-properly-validated-by-subversion-openssl) shrug and I don't use SVN enough to offer an awful lot more than that.

Not helpful is that Apple is [sort of mysterious](https://opensource.apple.com/source/subversion/subversion-71/) about how it goes about building SVN compared to the extensive detail Apple provides about system tools not considered purely "developer" such as `openssh`. Even the patches in `files` aren't wildly useful in terms of configuration detail.

Poking about into how Apple builds the `serf` module which is what `svn` leans on for SSL/TLS reveals some interesting things, namely that at least on Sierra they've transitioned `serf` to building using `libressl` instead of `openssl`, presumably `libressl 2.5.0` since that's what `/usr/bin/ssh` is linked to, and `/Library/Developer/CommandLineTools/usr/lib/libserf-1.0.dylib` is linked to the same libraries:

```
/Library/Developer/CommandLineTools/usr/lib/libserf-1.0.dylib:
	@rpath/lib/libserf-1.0.dylib (compatibility version 1.3.0, current version 1.3.0)
	/usr/lib/libapr-1.0.dylib (compatibility version 5.0.0, current version 5.8.0)
	/usr/lib/libaprutil-1.0.dylib (compatibility version 4.0.0, current version 4.0.0)
	/usr/lib/libcrypto.35.dylib (compatibility version 36.0.0, current version 36.0.0)
	/usr/lib/libssl.35.dylib (compatibility version 36.0.0, current version 36.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.0.0)
```
Homebrew uses `openssl` obviously, and so the linkage is slightly different, particularly in terms of using our libraries rather than the system ones, but there's nothing glaringly obvious in terms of explaining the functional differences:
```
/usr/local/opt/subversion/libexec/serf/lib/libserf-1.dylib:
	/usr/local/opt/subversion/libexec/serf/lib/libserf-1.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/opt/openssl/lib/libssl.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
	/usr/local/opt/apr/libexec/lib/libapr-1.0.dylib (compatibility version 6.0.0, current version 6.2.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.50.2)
	/usr/local/opt/apr-util/libexec/lib/libaprutil-1.0.dylib (compatibility version 6.0.0, current version 6.4.0)
	/usr/lib/libexpat.1.dylib (compatibility version 7.0.0, current version 8.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/System/Library/Frameworks/Kerberos.framework/Versions/A/Kerberos (compatibility version 5.0.0, current version 6.0.0)
```

SVN is/was configured-by-default to be cautious about trusting system default CAs, instead falling back on requiring users to verify fingerprints, but Homebrew isn't configuring subversion to behave differently in this aspect and I haven't tweaked my local configuration to encourage it to do anything different by default, so I'm unsure why Homebrew's SVN is happier to trust that the certs it finds are legit & good to use.

The one remaining avenue for clarity, at least with my limited knowledge of SVN's inner workings here, is that it relies on being able to actually, well, find a bundle of root certificates. When it was linked to OpenSSL in the past it would have expected to find those certificates in `/System/Library/OpenSSL`, and Apple didn't install any by default there.

However Apple recently started dumping what I presume is LibreSSL's default cert directory in `/etc/ssl` which does actually contain a set of default certs, although I'm unsure whether that bundle comes from Apple or LibreSSL. Yet the system `svn` remains incapable of handling most SSL/TLS connections without demanding a fingerprint comparison, which makes me wonder if subversion either isn't finding that bundle or the bundle is rather limited. I suspect the former because moving a PEM generated by `openssl`'s `postinstall` step into `/etc/ssl/cert.pem` doesn't change the situation.

Lastly, theoretically you can tweak your subversion configuration to trust default system ca files, and teach it where to find those, by doing something like this:
```
[global]
ssl-authority-files = /usr/local/etc/openssl/cert.pem
ssl-trust-default-ca = yes
```
and then:
```
/usr/bin/svn --config-dir ~/.subversion checkout https://svn.code.sf.net/p/gpsim/code/trunk
```
But that still fails. At this point I run out of ideas. The best I can do is suggest that our `subversion` knows where to actually _find_ our default cert bundle and _for some reason_ inherently trusts it and is happy to execute SSL/TLS connections. It obviously knows where to find our generated PEM because if you `mv /usr/local/etc/openssl/cert.pem ~/Downloads` and then:
```
/usr/local/bin/svn --config-dir ~/.subversion checkout https://svn.code.sf.net/p/gpsim/code/trunk
Error validating server certificate for 'https://svn.code.sf.net:443':
 - The certificate is not issued by a trusted authority. Use the
   fingerprint to validate the certificate manually!
Certificate information:
 - Hostname: *.code.sf.net
 - Valid: from Apr  6 00:00:00 2016 GMT until Jun  5 23:59:59 2017 GMT
 - Issuer: GeoTrust SSL CA - G3, GeoTrust Inc., US
 - Fingerprint: 9C:63:DE:5C:24:C4:86:ED:D2:53:26:E1:26:6B:52:0D:DD:B0:95:A3
(R)eject, accept (t)emporarily or accept (p)ermanently?
```
It falls back to the fingerprint check, because it failed to find that cert bundle. You put the cert bundle back, re-execute the command, and all is rosy again.

DomT4:
Lord did I overuse the word "actually" in that post.

MikeMcQuaid:
This is cool but personally given it adds code complexity to add security to building from source of an almost deprecated protocol I wonder if it's worth the potential bugs. I'd rather we still relied on system versions and worked around their breakages rather than just relying on dupes.

DomT4:
Alright, I'll add it to my increasing pile of local Homebrew hacks 😄.

PR #1967, opened 2017-02-07
Use `Open3` for integration tests instead of manually forking.

reitermarkus:

-----

Simplify `IntegrationCommandTestCase` by using `Open3`.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -217,6 +217,10 @@ class Keg
     true
   end
 
+  def require_relocation?
+    # no-op
+  end
+
   def linked?
     linked_keg_record.symlink? &&
       linked_keg_record.directory? &&
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -218,7 +218,7 @@ class Keg
   end
 
   def require_relocation?
-    # no-op
+    false
   end
 
   def linked?
--- a/Library/Homebrew/test/support/helper/integration_command_test_case.rb
+++ b/Library/Homebrew/test/support/helper/integration_command_test_case.rb
@@ -3,6 +3,7 @@ require "fileutils"
 require "pathname"
 require "formula"
 require "test/support/helper/test_case"
+require "open3"
 
 class IntegrationCommandTestCase < Homebrew::TestCase
   def setup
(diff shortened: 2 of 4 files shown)


Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -254,7 +254,7 @@ def bottle_formula(f)
             relocatable = false if keg_contain?("#{prefix}/etc", keg, ignores)
             relocatable = false if keg_contain?("#{prefix}/var", keg, ignores)
           end
-          skip_relocation = relocatable && !keg.require_relocation?
+          skip_relocation = relocatable && keg.respond_to?(:require_relocation?) && !keg.require_relocation?
```
MikeMcQuaid: When does it not respond to this?

Library/Homebrew/test/support/helper/integration_command_test_case.rb
```diff
@@ -27,68 +28,42 @@ def cmd_id_from_args(args)
...
-  def cmd_output(*args)
-    # 1.8-compatible way of writing def cmd_output(*args, **env)
-    env = args.last.is_a?(Hash) ? args.pop : {}
-    cmd_args = %W[
-      -W0
-      -I#{HOMEBREW_LIBRARY_PATH}/test/support/lib
-      -I#{HOMEBREW_LIBRARY_PATH}
-      -rconfig
+  def cmd_run(*args)
+    env = (args.first.is_a?(Hash) ? args.shift : {}).merge(
```
MikeMcQuaid: Let's keep this as the last argument and keep the function name the same. Let's also split `.merge` onto a new line/call rather than nest it.

Library/Homebrew/dev-cmd/bottle.rb
reitermarkus: One of the `--generic` tests failed because `require_relocation?` is only loaded on Mac. This wasn't happening before because we only checked the output of that command, but not the exit status.

Library/Homebrew/test/support/helper/integration_command_test_case.rb
reitermarkus: I think making the environment the first argument is more in line with the standard library (i.e. `Open3`, which in turn calls `spawn`).

Library/Homebrew/dev-cmd/bottle.rb
MikeMcQuaid: Ok, thanks. Instead of `respond_to?` could you add a generic, no-op version of that method? Thanks!

Library/Homebrew/test/support/helper/integration_command_test_case.rb
MikeMcQuaid: It's more consistent with those but less with idiomatic Ruby as a whole. I don't think it's more readable so let's leave as-is, thanks.

Library/Homebrew/keg.rb
```diff
@@ -217,6 +217,10 @@ def empty_installation?
     true
   end
 
+  def require_relocation?
+    # no-op
```
MikeMcQuaid: Make this `false` rather than `no-op`, thanks.

MikeMcQuaid: (and can drop the comment) but thanks for the change ❤️

reitermarkus: Actually had `false` in there before but thought a comment would be even *no-op*-ier. 😂

ilovezfs:
It seems this PR has broken bottling.

See
https://bot.brew.sh/job/Homebrew%20Core/16215/version=sierra/console
https://bot.brew.sh/job/Homebrew%20Science%20Pull%20Requests/7203/version=sierra/console

Probably the
```
+  def require_relocation?
+    false
+  end
```
since it now seems to be unconditionally assigning
```
"cellar":"any_skip_relocation"
```
in the bottle json files.

ilovezfs:
I've opened https://github.com/Homebrew/brew/pull/1985 to revert this.

PR #1968, opened 2017-02-08
Separate Minitest assertions and add spec support.

reitermarkus:

-----

Move Minitest assertions to separate file and add support for spec syntax.

### DIFF
--- a/Library/Homebrew/test/support/helper/test_case.rb
+++ b/Library/Homebrew/test/support/helper/test_case.rb
@@ -1,6 +1,8 @@
 require "formulary"
 require "tap"
 
+require "test/support/helper/assertions"
+
 module Homebrew
   class TestCase < ::Minitest::Test
     require "test/support/helper/fs_leak_logger"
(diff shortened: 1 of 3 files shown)


Library/Homebrew/test/support/helper/minitest/assertions.rb
```diff
@@ -0,0 +1,19 @@
+module MiniTest
+  module Assertions
```
MikeMcQuaid: What's the advantage of monkey patching this module compared to the previous approach?

Library/Homebrew/test/support/helper/minitest/spec.rb
```diff
@@ -0,0 +1,9 @@
+require "minitest/spec"
+
+module MiniTest
+  class Spec
+    module DSL
```
MikeMcQuaid: How can we avoid monkey patching this existing class?

Library/Homebrew/test/support/helper/test_case.rb
```diff
@@ -1,8 +1,15 @@
 require "formulary"
 require "tap"
+require "minitest/spec"
 
 module Homebrew
   class TestCase < ::Minitest::Test
+    extend MiniTest::Spec::DSL
+
+    register_spec_type(self) do |desc|
```
MikeMcQuaid: What's this do? It's not obvious.

Library/Homebrew/test/ARGV_test.rb
```diff
@@ -1,79 +1,147 @@
 require "testing_env"
 require "extend/ARGV"
 
-class ArgvExtensionTests < Homebrew::TestCase
-  def setup
-    super
-    @argv = [].extend(HomebrewArgvExtension)
-  end
+describe HomebrewArgvExtension do
```
MikeMcQuaid: I don't understand the motivation for rewriting this test. How does this make things better for our users?

PR #1969, opened 2017-02-08
Create custom matchers for “valid symlink”.

reitermarkus:

-----

Adds a custom matcher to check if a file is a symlink and its target exists.

### DIFF
--- a/Library/Homebrew/cask/spec/cask/artifact/binary_spec.rb
+++ b/Library/Homebrew/cask/spec/cask/artifact/binary_spec.rb
@@ -20,7 +20,8 @@ describe Hbc::Artifact::Binary do
     shutup do
       Hbc::Artifact::Binary.new(cask).install_phase
     end
-    expect(expected_path).to be_a_valid_symlink
+    expect(expected_path).to be_a_symlink
+    expect(expected_path.readlink).to exist
   end
 
   it "avoids clobbering an existing binary by linking over it" do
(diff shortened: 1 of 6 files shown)


Library/Homebrew/cask/spec/cask/artifact/binary_spec.rb
```diff
@@ -20,7 +20,7 @@
     shutup do
       Hbc::Artifact::Binary.new(cask).install_phase
     end
-    expect(FileHelper.valid_alias?(expected_path)).to be true
+    expect(expected_path).to be_a_valid_symlink
```
MikeMcQuaid: Personally this seems harder to follow not given it's only used in two places.

reitermarkus: I think this is more idiomatic in this case, and you are still able to look it up by name as you would be for `FileHelper.valid_alias?`.

MikeMcQuaid: I'd argue for https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming) but don't feel strongly either way.

MikeMcQuaid: Personally: yes. This PR is a refactoring for slightly shortening two lines of code which ends up adding more lines than it removes.

reitermarkus: Ok, that makes sense.

PR #1972, opened 2017-02-08
add `brew formula` command to show location of a formula

timotheecour:

Full rationale for this PR is given https://github.com/Homebrew/homebrew-core/issues/8935 and @ilovezfs said (https://github.com/Homebrew/homebrew-core/issues/8935#issuecomment-272778026)
> We'd consider a PR for such a command but no one is actively working on it.

* https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md was not saying much on `adding a new command`, it would've helped (if anything just showing the simplest PR sha for adding a command)


Library/Homebrew/cmd/help.rb
```diff
@@ -15,7 +15,7 @@
 
 Developers:
   brew create [URL [--no-fetch]]
-  brew edit [FORMULA...]
+  brew (where|edit) [FORMULA...]
```
ilovezfs: I think I'd prefer these remain separate in the docs.

ilovezfs:
The equivalent command for commands is `brew command` so I wonder if this should be `brew formula`. If it is `brew where`, then I wonder if it should handle tap arguments as well. Currently you can do
```
iMac-TMP:~ joe$ brew --repo homebrew/games
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-games
```
But this seems a little weird to me:
```
iMac-TMP:~ joe$ brew where homebrew/games
Error: No available formula with the name "homebrew/games" 
iMac-TMP:~ joe$ 
```

Library/Homebrew/cmd/help.rb
timotheecour: sure, i did that because:
* `brew (info|home|options)` does it above
* i remember seeing somewhere to keep vertical size of --help small
* as i described in https://github.com/Homebrew/homebrew-core/issues/8935 `brew where` seems to supersede `brew edit` , `brew cat`
shall i go ahead w ur change?

ilovezfs: Let's see what Mike thinks before I send you off doing more work :)

timotheecour:
i like `brew formula`, will change to that

Library/Homebrew/cmd/help.rb
timotheecour: actually here it is in plain site in help.rb:
```
# NOTE Keep the lenth of vanilla --help less than 25 lines!
```
so how about:
* option A: brew (formula|edit) [FORMULA...]
* option B: brew formula [FORMULA...]
and remove brew edit (which, again, is superseded by `brew formula`, whereas the other direction is not the case)

ilovezfs: Neither `brew cat` nor `brew edit` would be superseded by this command.

timotheecour:
ok @ilovezfs  PTAL, all comments addressed; i left off `brew formula` from --help so as not to block this PR; we can always edit --help in subsequent PR if needed

completions/zsh/_brew
```diff
@@ -793,6 +794,12 @@ _brew_vendor_install() {
     '--target'
 }
 
+
+# brew where formulae:
+_brew_where() {
```
MikeMcQuaid: Is it worth renaming this? Also, is it worth adding to the bash completion?

Library/Homebrew/dev-cmd/formula.rb
```diff
@@ -0,0 +1,15 @@
...
+
+require "formula"
+
+module Homebrew
+  module_function
+
+  def formula
+    raise FormulaUnspecifiedError if ARGV.named.empty?
+    ARGV.resolved_formulae.each do |f|
+      puts "#{f.path}\n"
```
MikeMcQuaid: `puts` should append a newline so the `\n` is probably unnecessary.

Library/Homebrew/dev-cmd/formula.rb
```diff
@@ -0,0 +1,15 @@
...
+#:    echo location of the specified <formulae> to stdout
+
+require "formula"
+
+module Homebrew
+  module_function
+
+  def formula
+    raise FormulaUnspecifiedError if ARGV.named.empty?
+    ARGV.resolved_formulae.each do |f|
```
MikeMcQuaid: `ARGV.resolved_formulae.each { |f| puts f.path }`?

Library/Homebrew/dev-cmd/formula.rb
```diff
@@ -0,0 +1,15 @@
+#:  * `formula` <formulae>:
```
MikeMcQuaid: I'd just go for `<formula` for consistency with other commands.

Library/Homebrew/dev-cmd/formula.rb
```diff
@@ -0,0 +1,15 @@
...
+#:    echo location of the specified <formulae> to stdout
```
MikeMcQuaid: `Display the path where <formula> is`

completions/zsh/_brew
```diff
@@ -88,6 +88,7 @@ __brew_common_commands() {
     'doctor:audits your installation for common issues'
     'edit:edit a formula'
     'fetch:download formula resources to the cache'
+    'formula:location of formulae'
```
MikeMcQuaid: `formula: the path for a formula`

timotheecour:
@MikeMcQuaid PTAL

MikeMcQuaid:
Thanks @timotheecour! Last thing: could you add a small integration test to test the output of this command?

timotheecour:
@MikeMcQuaid  PTAL : `brew tests` passes.
btw: it would lower barrier of entry for external contributors to add docs on specific things (eg steps involved in adding a command: where to add tests, how to update man-pages, commands to run eg brew man, brew tests, etc). Unfortunately https://github.com/Homebrew/brew/pull/1973 (Update CONTRIBUTING.md) was marked as wontfix. This PR would've been easier for me if I didn't have to discover those by myself. Unfortunately also wiki was disabled. how about `CONTRIBUTING-advanced.md` or whatever other name if you don't  want to bloat `CONTRIBUTING.md` ?

MikeMcQuaid:
> btw: it would lower barrier of entry for external contributors to add docs on specific things (eg steps involved in adding a command: where to add tests, how to update man-pages, commands to run eg brew man, brew tests, etc). Unfortunately #1973 (Update CONTRIBUTING.md) was marked as wontfix. This PR would've been easier for me if I didn't have to discover those by myself. Unfortunately also wiki was disabled. how about CONTRIBUTING-advanced.md or whatever other name if you don't want to bloat CONTRIBUTING.md ?

We'll consider documenting these things if we see multiple users struggling with them. Homebrew has thousands of contributors and millions of users; it doesn't scale to try and document every single thing an individual doesn't know.

> What command do i run to only run the test I'm adding?
> tried a few variants of brew tests --only=formula_cmd_test but didn't work (Pass files or folders to run error); an example in manpage would help

It's `--only=formula_cmd` I believe. Agreed adding to manpage would be useful 👍 

--

Thanks for your first contribution to Homebrew/brew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #1985, opened 2017-02-11
Revert "Use `Open3` for integration tests instead of manually forking."

ilovezfs:
Reverts Homebrew/brew#1967


ilovezfs:
CC @reitermarkus

reitermarkus:
Ok, so `keg` is required after `os/mac/keg`, hence overriding `require_relocation?`.

PR #1987, opened 2017-02-11
Convert Homebrew::Hooks::Bottles test to spec.

reitermarkus:

-----

Convert tests to RSpec.


Library/Homebrew/test/bottle_hooks_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "formula_installer"
+require "hooks/bottles"
+
+describe Homebrew::Hooks::Bottles do
+  subject { formula_installer }
+
+  let(:formula_installer) {
```
MikeMcQuaid: How about doing `let(:formula_installer) { FormulaInstaller.new formula }` for one-liners < 80 characters?

Library/Homebrew/test/bottle_hooks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    FormulaInstaller.new formula
+  }
+
+  let(:formula) {
+    double(bottle_disabled?: false, some_random_method: true, bottle: nil, local_bottle_path: nil)
```
MikeMcQuaid: Probably worth splitting out the hash.

Library/Homebrew/test/bottle_hooks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  describe "has bottle" do
+    let(:block) { :some_random_method }
+    it { is_expected.to be_pour_bottle }
+  end
+
+  describe "has no bottle" do
+    let(:block) {
+      ->(f) { !f.some_random_method }
```
MikeMcQuaid: I have no idea what this is doing 😁. Can you explain?

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,70 @@
+require "build_environment"
+
+RSpec::Matchers.define :use_userpaths do
+  match do |actual|
```
MikeMcQuaid: `match {|actual| actual.userpaths?`?

Library/Homebrew/test/bottle_hooks_spec.rb
MikeMcQuaid: Also, why not `let do`?

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+  subject {
+    Object.new.extend(BuildEnvironmentDSL).tap { |o| o.instance_eval(&block) }
+  }
+
+  context "single argument" do
+    let(:block) {
+      proc do
+        env :userpaths
+      end
+    }
```
alyssais: This proc-inside-a-block thing is weird, especially with the `do`/`end` inside the `{`/`}`.

Maybe just define it as a method?:

```ruby
def block
  env :userpaths
end
```

and evaluate it with

```ruby
o.instance_eval(&method(:block))
```

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    it "returns false if the environment does not contain :userpaths" do
+      expect(env).not_to use_userpaths
+    end
+  end
+end
+
+describe BuildEnvironmentDSL do
+  subject {
+    Object.new.extend(BuildEnvironmentDSL).tap { |o| o.instance_eval(&block) }
```
alyssais: I think this should be split up into multiple lines.

The subject line should just construct the subject object, and the block should either be evaluated inside a `before` or inside the tests themselves.

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    actual.userpaths?
+  end
+end
+
+describe BuildEnvironment do
+  let(:env) { described_class.new }
+
+  describe "#<<" do
+    it "returns itself" do
+      expect(env << :foo).to eq env
```
alyssais: You've changed the behaviour of these tests. The Minitest equivalents used `assert_same`, which checks instance equality using `#equal?`. The equivalent construction in RSpec is `be` (instead of `eq`).

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,70 @@
+require "build_environment"
+
+RSpec::Matchers.define :use_userpaths do
+  match do |actual|
+    actual.userpaths?
+  end
+end
+
+describe BuildEnvironment do
+  let(:env) { described_class.new }
```
alyssais: Should this not be `subject`?

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  describe "#merge" do
+    it "returns itself" do
+      expect(env.merge([])).to eq env
+    end
+  end
+
+  describe "#std?" do
+    it "returns true if the environment contains :std" do
+      expect(env << :std).to be_std
```
alyssais: Could this and similar tests be split up into multiple lines? I don't think mutation/setup should happen _inside_ an expectation.

These are specs — we want them to be as simple as possible to read and understand.

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,70 @@
+require "build_environment"
+
+RSpec::Matchers.define :use_userpaths do
```
alyssais: There should actually be no need to define a new matcher here at all:

```ruby
RSpec::Matchers.alias_matcher :use_userpaths, :be_userpaths
```

Library/Homebrew/.rubocop.yml
```diff
@@ -11,8 +11,8 @@ AllCops:
 
 Style/BlockDelimiters:
   Exclude:
-    - '**/cask/spec/**/*'
-    - '**/cask/test/**/*'
+    - '**/spec/**/*'
+    - '**/test/**/*'
```
alyssais: Can you explain what this does and why it was needed?

Library/Homebrew/test/build_environment_spec.rb
reitermarkus: I think in this it's nicer to have a named variable since there is no good way to use the short-hand `it { is_expected }` syntax.

Library/Homebrew/test/build_environment_spec.rb
reitermarkus: I searched for something like this but couldn't find it, thanks.

Library/Homebrew/test/build_environment_spec.rb
reitermarkus: I don't like defining methods inside specs, that is specifically what procs and lambda are for. Anyway, in this case it really only works with procs, not with lambdas or methods.

Library/Homebrew/test/build_environment_spec.rb
alyssais: That's true, but there's also some semantic meaning attached to `subject` and I think consistency is desirable here — it's not just a hack you use to get nice syntax. When I glanced at these tests I assumed `env` to be some kind of helper object.

alyssais:
Additionally, could the conversion be done across multiple PRs, maybe one per file? If the whole test suite is rewritten it's going to be really hard to keep up with and review effectively (and it'll take forever to get any of it merged!)

Library/Homebrew/test/bottle_hooks_spec.rb
reitermarkus: It's passing the lambda to `setup_formula_has_bottle`.

Library/Homebrew/.rubocop.yml
MikeMcQuaid: Agreed.

Library/Homebrew/test/build_environment_spec.rb
MikeMcQuaid: I agree. It may be what procs and lambda are for but it makes the tests significantly more opaque for beginning or non-Rubyists.

Library/Homebrew/test/bottle_hooks_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+require "hooks/bottles"
+
+RSpec::Matchers.alias_matcher :pour_bottle, :be_pour_bottle
+
+describe Homebrew::Hooks::Bottles do
+  subject { formula_installer }
+
+  let(:formula_installer) { FormulaInstaller.new formula }
+
+  let(:formula) {
```
MikeMcQuaid: Can we use `let do` on multiline lets?

Library/Homebrew/test/bottle_hooks_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    double(
+      bottle: nil,
+      local_bottle_path: nil,
+      bottle_disabled?: false,
+      some_random_method: true
```
MikeMcQuaid: May as well add a trailing comma here.

MikeMcQuaid:
I think the heavy use of procs, lambdas and subject makes this significantly less readable than it would be using methods and duplicating function calls.

Library/Homebrew/.rubocop.yml
reitermarkus: This is so RuboCop doesn't complain about multi-line `expect`s, e.g.:

```ruby
expect {
  # something
}.not_to raise_error
```

Library/Homebrew/test/bottle_hooks_spec.rb
reitermarkus: RuboCop complains about commas after the last parameter of a method call.

MikeMcQuaid: Boo. Ignore me.

Library/Homebrew/.rubocop.yml
MikeMcQuaid: Boo. Ignore me.

MikeMcQuaid:
> I think the heavy use of procs, lambdas and subject makes this significantly less readable than it would be using methods and duplicating function calls.

And to clarify on this: I think making it less readable for newer Rubyists will make us less likely to have good, new tests implemented by new contributors.

reitermarkus:
Removed the `proc`s and put `instance_eval` inside `before` blocks instead.

Library/Homebrew/test/bottle_hooks_spec.rb
alyssais: Perhaps that's something we should re-visit? Trailing commas are very handy when argument lists are on multiple lines IMO.

MikeMcQuaid: If it's a rule that can be tweaked 👍

Library/Homebrew/test/bottle_hooks_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    context "given a block which evaluates to false" do
+      let(:block) { ->(f) { !f.some_random_method } }
+
+      it { is_expected.not_to pour_bottle }
+    end
+  end
+
+  describe "#setup_pour_formula_bottle" do
+    subject { -> { formula_installer.pour } }
```
MikeMcQuaid: What would this look like without a proc/lambda?

Library/Homebrew/test/bottle_hooks_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    let(:block) { ->(_) { true } }
```
MikeMcQuaid: What would this look like without a proc/lambda?

Library/Homebrew/test/build_environment_spec.rb
```diff
@@ -0,0 +1,66 @@
+require "build_environment"
```
MikeMcQuaid: I'm 👍 on this test file being merged if you split into a separate PR.

alyssais:
Very happy with how this turned out 🙂

PR #1990, opened 2017-02-12
Convert X11Requirement test to spec.

reitermarkus:


reitermarkus:
@MikeMcQuaid, what is the `netlify` check used for?

MikeMcQuaid:
@reitermarkus It's something I'm experimenting with for docs.brew.sh hosting. Going to look how to turn it off now...

Library/Homebrew/test/x11_requirement_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "requirements/x11_requirement"
+
+describe X11Requirement do
+  subject { described_class.new }
+
+  describe "#eql?" do
```
alyssais: This is testing more than just `#eql?`. Maybe `describe "equality" do`?

Library/Homebrew/test/x11_requirement_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "requirements/x11_requirement"
+
+describe X11Requirement do
+  subject { described_class.new }
```
alyssais: This line is unnecessary, I think.

Library/Homebrew/test/x11_requirement_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  describe "#modify_build_environment" do
+    it "calls ENV#x11" do
+      allow(subject).to receive(:satisfied?).and_return(true)
+
+      expect(ENV).to receive(:x11)
+
+      subject.modify_build_environment
```
alyssais: There's a lot of whitespace here that could probably be condensed…

Library/Homebrew/test/x11_requirement_spec.rb
reitermarkus: Thanks, didn't know `subject` was implicitly initialized.

Library/Homebrew/test/x11_requirement_spec.rb
reitermarkus: Isn't this test basically checking if `eql?` is using `hash` to compare the two items? It isn't, by the way, which is odd.

alyssais: If that's what it's checking, it should say that in the description!

reitermarkus: Well, I don't know what was originally intended to be checked, but that's what I'm guessing should be checked.

Library/Homebrew/requirement.rb
```diff
@@ -109,7 +109,7 @@ def env_proc
   end
 
   def ==(other)
-    instance_of?(other.class) && name == other.name && tags == other.tags
+    instance_of?(other.class) && hash == other.hash
```
alyssais: What's the reason for this change?

reitermarkus: The test is now expecting `eql?` to call `hash`, which it wasn't before. Anyway, I think it would be better to move the `hash` check to a separate test, so ignore this.

reitermarkus:
I removed the `hash` equality check from this test and added a separate test for it to `requirement_spec.rb`.

PR #1991, opened 2017-02-12
Convert Requirement test to spec.

reitermarkus:


Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+describe Requirement do
+  subject { klass.new }
+
+  describe "#tags" do
+    subject { described_class.new(tags) }
+
+    context "single tag" do
+      let(:tags) { ["bar"] }
+
+      its(:tags) { are_expected.to eq(tags) }
```
MikeMcQuaid: Any reason to not just hardcode `tags` given it's only used once?

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    context "multiple tags" do
+      let(:tags) { ["bar", "baz"] }
+
+      its(:tags) { are_expected.to eq(tags) }
```
MikeMcQuaid: Any reason to not just hardcode `tags` given it's only used once?

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    context "symbol tags" do
+      let(:tags) { [:build] }
+
+      its(:tags) { are_expected.to eq(tags) }
```
MikeMcQuaid: Any reason to not just hardcode `tags` given it's only used once?

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+    end
+
+    context "symbol and string tags" do
+      let(:tags) { [:build, "bar"] }
+
+      its(:tags) { are_expected.to eq(tags) }
```
MikeMcQuaid: Any reason to not just hardcode `tags` given it's only used once?

Library/Homebrew/test/requirement_spec.rb
reitermarkus: It's not just used once. The subject is created with it, and the check uses it.

MikeMcQuaid: Course, sorry, missed that.

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+        end
+      end
+
+      it "sets up build environment by default" do
+        expect(ENV).to receive(:with_build_environment) do
+          yield true if block_given?
+          true
+        end
+
+        expect(subject).to be_satisfied
```
alyssais: This is really two different tests, isn't it? One testing the return value as above, and another testing the ENV message.

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+    context "#satisfy with block and build_env returns true" do
+      let(:klass) do
+        Class.new(described_class) do
+          satisfy(build_env: false) do
+            true
+          end
+        end
+      end
+
+      it "build environment can be disabled" do
```
alyssais: Can we use `specify` instead of `it` when the `it` doesn't make sense in the sentence?

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+
+    it "infers path from #satisfy result" do
+      expect(ENV).to receive(:with_build_environment) do
+        yield path if block_given?
+        path
+      end
+      expect(ENV).to receive(:append_path).with("PATH", path.parent)
+
+      expect(subject).to be_satisfied
+      subject.modify_build_environment
```
alyssais: As above, this looks like it should be multiple tests.

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+  end
+
+  describe "#default_formula?" do
+    let(:klass) do
+      Class.new(described_class) do
+        default_formula "foo"
+      end
+    end
+
+    it { is_expected.to have_a_default_formula }
```
alyssais: Should there be a test for the negative case too?

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+  end
+
+  describe "#build?" do
+    let(:klass) do
+      Class.new(described_class) do
+        build true
+      end
+    end
+
+    it { is_expected.to be_a_build }
```
alyssais: Should there be a test for the negative case too?

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+      let(:tags) { [:build, "bar"] }
+
+      its(:tags) { are_expected.to eq(tags) }
+    end
+  end
+
+  describe "#fatal?" do
+    subject { Class.new(described_class) { fatal true }.new }
+
+    it { is_expected.to be_fatal }
```
alyssais: Should there be a test for the negative case too?

Library/Homebrew/test/requirement_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+    context "without env proc" do
+      let(:klass) { Class.new(described_class) }
+
+      it "returns nil" do
+        expect(subject.modify_build_environment).to be nil
+      end
+    end
+  end
+
+  describe "#eql?" do
```
alyssais: This is testing more than just `#eql?`. Maybe `describe "equality" do`?

Library/Homebrew/test/requirement_spec.rb
reitermarkus: How would you split this?

alyssais: I would extract `it { is_expected.to be_satisfied }`.

reitermarkus: `modify_build_environment` depends on `satisfied_result_parent` returning a path, which is set in `satisfied?`, so it has to be called in the same test.

PR #1999, opened 2017-02-12
dependency: avoid expanding same dep twice

vladshablinsky:

## description

Consider the case when with these formulae:
`a`, `b`, `c`, `d`, `e`.

* `a` dependencies: `b`
* `b` dependencies: `c, e`
* `c` dependencies: `d`
* `e` dependencies: `c`

If we're expanding dependencies using post-order traversal

* expand left subtree
* expand right subtree
* display current node

we're doing some unnecessary computations for `c` and `d` which are
displayed in the parantheses:

```
         a
        /
       b
      / \
     c   e
    /     \
   d      (c)
          /
        (d)
```

Let's expand a's dependencies:

1. `a`, call recursively for `b`
2. `b`, call recursively for `c`
3. `c`, call recursively for `d`
4. `d`, push `d` to dependencies list, which now gets `[d]`, go up to
`c`
5. `c`, push `c` to dependencies list, which now gets `[d, c]`, go up to `b`
6. `b`, call recursively for `e`
7. _`e`, call recursively for `c` <-- unnecessary call_
8. _`c`, call recursively for `d` <-- unnecessary call_
9. `d`, dependencies list already has `[d, c]`, go up to `c`
10. `c`, go up to `e`, push `e` to dependencies list, which now gets `[d, c,
e]`, go up to `b`
11. `b`, push `b` to dependencies list, which now gets `[d, c, e, b]`, go up
to a
12. `a`, push `a` to dependencies list, which now gets `[d, c, e, b, a]`

## tl;dr

Let's skip expanding subtrees of dependencies we have already expanded
once, since they are the same as we have already expanded once.

### DIFF
--- a/Library/Homebrew/dependency.rb
+++ b/Library/Homebrew/dependency.rb
@@ -78,6 +78,8 @@ class Dependency
       # Keep track dependencies to avoid infinite cyclic dependency recursion.
       @expand_stack ||= []
       @expand_stack.push dependent.name
+      # Keep track of already expanded dependencies.
+      @ready ||= []
 
       expanded_deps = []
 
@@ -88,14 +90,15 @@ class Dependency
         when :prune
           next
         when :skip
-          next if @expand_stack.include? dep.name
+          next if @expand_stack.include?(dep.name) || @ready.include?(dep.name)
           expanded_deps.concat(expand(dep.to_formula, &block))
         when :keep_but_prune_recursive_deps
           expanded_deps << dep
         else
-          next if @expand_stack.include? dep.name
+          next if @expand_stack.include?(dep.name) || @ready.include?(dep.name)
           expanded_deps.concat(expand(dep.to_formula, &block))
           expanded_deps << dep
+          @ready << dep
         end
       end
 


MikeMcQuaid:
> Let's skip expanding subtrees of dependencies we have already expanded
> once, since they are the same as we have already expanded once.

Just to check the obvious thing that popped into my mind: how does this handle e.g. dependencies with options? Similarly, do you have any benchmarks that show a noticeable performance increase? I tend to favour only caching for measurably performance impact.

vladshablinsky:
> Just to check the obvious thing that popped into my mind: how does this handle e.g. dependencies with options?

Well, the results of `expand(dep.to_formula, &block)`can differ if we pass same `&block` and same `dep` but with different options. Will need to think about it a bit more.

> I tend to favour only caching for measurably performance impact.

Good point. I think if we cache it properly, we'll able to reuse it if we pass a list of formulae to install to `brew`.

> Similarly, do you have any benchmarks that show a noticeable performance increase? 

It's not that difficult to create an artificial example where it should be e.g. twice more efficient, however I don't have benchmarks.

MikeMcQuaid:
> It's not that difficult to create an artificial example where it should be e.g. twice more efficient, however I don't have benchmarks on some real data.

I'd like to gently suggest that caching will introduce edge cases and without user-visible performance increases I'm probably 👎 on it. Performance tuning/caching is valuable, though, but I'd do so using `ruby-prof --printer=call_stack` or [`brew prof`](https://github.com/Homebrew/homebrew-dev-tools/blob/master/cmd/brew-prof.rb) on a particular command we know has low performance i.e. `brew uses --recursive`.

MikeMcQuaid:
Passing on this for now, sorry @vladshablinsky 😭

vladshablinsky:
@MikeMcQuaid no worries. It really requires some testing performed and probably rethinking.

PR #2014, opened 2017-02-14
Convert Checksum test to spec.

reitermarkus:


Library/Homebrew/test/checksum_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "checksum"
+
+describe Checksum do
+  describe "#empty?" do
+    subject { described_class.new(:sha256, "") }
+    it { is_expected.to be_empty }
```
alyssais: I'm late to the party here, I know, but I think named examples should be used here, as RSpec will currently interpret this test as being named "Checksum #empty? is expected to be empty", which is not particularly clear.

PR #2026, opened 2017-02-15
Convert `brew analytics` test to spec.

reitermarkus:


Library/Homebrew/test/support/helper/spec/integration_command.rb
```diff
@@ -0,0 +1,92 @@
+require "open3"
+
+RSpec::Matchers.define_negated_matcher :not_to_output, :output
+RSpec::Matchers.define_negated_matcher :be_a_failure, :be_a_success
+
+RSpec::Matchers.define :be_a_success do
```
alyssais: Since matchers are scoped globally, I think these need more specific names, like e.g. `exist_successfully?`

Library/Homebrew/test/support/helper/spec/integration_command.rb
```diff
@@ -0,0 +1,92 @@
...
+  match do |actual|
+    if actual.is_a?(Proc)
+      status = actual.call
+    else
+      status = actual
+    end
```
alyssais: I think this would be cleaner as a ternary but won't push too hard if you or @MikeMcQuaid have strong opinions to the contrary.

Library/Homebrew/test/support/helper/spec/integration_command.rb
```diff
@@ -0,0 +1,92 @@
...
+  config.include(
+    Module.new do
+      def command_id_from_args(args)
+        @cmd_id_index ||= 0
+        pretty_args = args.join(" ").gsub(TEST_TMPDIR, "@TMPDIR@")
+        description = @example.full_description
+        "[#{description}][#{@cmd_id_index += 1}] brew #{pretty_args}"
+      end
+
+      def brew(*args)
```
alyssais: I think this method is sufficiently complex that it could use a comment explaining what it does.

Library/Homebrew/test/support/helper/spec/integration_command.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  config.after(:each) do
+    FileUtils.rm HOMEBREW_PREFIX/"bin/brew"
+    FileUtils.rmdir HOMEBREW_PREFIX/"bin"
+  end
+
+  config.include(
+    Module.new do
+      def command_id_from_args(args)
```
alyssais: I think this method is sufficiently complex that it could use a comment explaining what it does.

Library/Homebrew/test/support/helper/spec/integration_command.rb
```diff
@@ -0,0 +1,92 @@
...
+  def expects_call_stack_jump?
+    true
+  end
+end
+
+RSpec.configure do |config|
+  config.before(:each) do |example|
+    (HOMEBREW_PREFIX/"bin").mkpath
+    FileUtils.touch HOMEBREW_PREFIX/"bin/brew"
+    @example = example
```
alyssais: Does RSpec not expose `#full_description` in a test body? Storing the example as an instance variable seems pretty hacky.

Library/Homebrew/test/support/helper/spec/integration_command.rb
```diff
@@ -0,0 +1,92 @@
...
+  #       # command
+  #     }.to output(something).to_stdout
+  #   }.to be_a_success
+  #
+  def expects_call_stack_jump?
+    true
+  end
+end
+
+RSpec.configure do |config|
```
alyssais: This should be an RSpec shared context, I think.

Library/Homebrew/test/support/helper/spec/integration_command.rb
reitermarkus: I don't see why, since this will still work without a block.

Library/Homebrew/test/support/helper/spec/integration_command.rb
reitermarkus: I wonder why RuboCop didn't warn about this. It should actually suggest this:

```ruby
status = if actual.is_a?(Proc)
  actual.call
else
  actual
end
```

But yes, ternary it is.

Library/Homebrew/test/support/helper/spec/integration_command.rb
reitermarkus: It doesn't, only works if you run it with a block argument.

```ruby
specify do |example|
  example.full_description
end
```

Library/Homebrew/test/support/helper/spec/integration_command.rb
reitermarkus: How did I not know about shared contexts? 👍

Library/Homebrew/test/support/helper/spec/integration_command.rb
reitermarkus: We could probably just generate a random number instead of all of this.

Library/Homebrew/test/support/helper/spec/integration_command.rb
alyssais: Shared contexts are literally the reason I use RSpec.

Library/Homebrew/test/support/helper/spec/integration_command.rb
alyssais: In that case, can we move it into a shared location (and other marchers created in spec files)? I don't like creating global state in a single spec file.

PR #2028, opened 2017-02-16
Convert Migrator test to spec.

reitermarkus:


Library/Homebrew/test/migrator_spec.rb
```diff
@@ -0,0 +1,269 @@
+require "migrator"
+require "test/support/fixtures/testball"
+require "tab"
+require "keg"
+
+class Migrator
+  describe self do
```
alyssais: What's with these two lines over `describe Migrator`?

Library/Homebrew/test/migrator_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+      tab.source["tap"] = "homebrew/core"
+      tab.write
+
+      expect {
+        described_class.new(new_formula)
+      }.to raise_error(MigratorDifferentTapsError)
+    end
+  end
+
+  describe self do
```
alyssais: Why are there multiple `describe`s about the same thing?

Library/Homebrew/test/migrator_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+      old_tab.tabfile = HOMEBREW_CELLAR/"oldname/0.1/INSTALL_RECEIPT.json"
+      old_tab.source["path"] = "/oldname"
+      old_tab.write
+
+      keg.link
+      keg.optlink
+
+      old_pin.make_relative_symlink old_keg_record
+
+      subject # needs to be evaluated eagerly
```
alyssais: Does this depend on previous lines in this block? If not, could declare it with `subject!`.

Library/Homebrew/test/migrator_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+    after(:each) do
+      if !old_keg_record.parent.symlink? && old_keg_record.directory?
+        keg.unlink
+      end
+
+      if new_keg_record.directory?
+        new_keg = Keg.new(new_keg_record)
+        new_keg.unlink
+      end
+    end
```
alyssais: These ~40 lines of setup code make it pretty difficult to understand what's going on. Is there anything we could do to simplify them?

Library/Homebrew/test/migrator_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+
+    specify "test_move_cellar" do
+      keg.unlink
+      shutup do
+        subject.move_to_new_directory
+      end
```
alyssais: This could be one line, right?

Library/Homebrew/test/migrator_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+        keg.unlink
+      end
+
+      if new_keg_record.directory?
+        new_keg = Keg.new(new_keg_record)
+        new_keg.unlink
+      end
+    end
+
+    specify "test_move_cellar" do
```
alyssais: I think these tests need more descriptive names. It's very difficult for me to tell what's going on.

Library/Homebrew/test/migrator_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+      after(:each) do
+        expect(old_keg_record.parent).to be_a_directory
+        expect(old_keg_record.parent.subdirs).not_to be_empty
+        expect(HOMEBREW_LINKED_KEGS/"oldname").to exist
+        expect(HOMEBREW_PREFIX/"opt/oldname").to exist
+        expect(HOMEBREW_PINNED_KEGS/"oldname").to be_a_symlink
+        expect(keg).to be_linked
+      end
+
+      context "when cellar exists" do
```
alyssais: I'm always skeptical of contexts that contain nothing but a single example. I think these could be refactored to just be, e.g. `it "backs up the old name when cellar exists"`

Library/Homebrew/test/migrator_spec.rb
reitermarkus: This is so that `MigratorNoOldnameError` and `MigratorNoOldpathError` are scoped inside `Migrator`.

Library/Homebrew/test/migrator_spec.rb
reitermarkus: It depends on the order, I tried `subject!`, sadly doesn't work here.

Library/Homebrew/test/migrator_spec.rb
reitermarkus: I can't think of a better way to make sure the test files aren't created for the error tests.

Library/Homebrew/test/migrator_spec.rb
alyssais: I'd much rather you just referred to them as `Migrator::MigratorNoOldnameError` — every line it's used on is already pretty short, and I think this pattern is confusing.

MikeMcQuaid:
@alyssais You 🆒 with this now?

PR #2029, opened 2017-02-16
Convert Formulary test to spec.

reitermarkus:


Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+          prefix.install "bin"
+          prefix.install "libexec"
+        end
+      end
+    EOS
+  }
+  let(:bottle_dir) { Pathname.new("#{TEST_FIXTURE_DIR}/bottles") }
+  let(:bottle) { bottle_dir/"testball_bottle-0.1.#{Utils::Bottles.tag}.bottle.tar.gz" }
+
+  specify "::class_s" do
```
alyssais: I think this should be split into multiple examples with descriptive names, e.g. `it "replaces '+' with 'x'"`.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
+require "formula"
+require "formula_installer"
+require "utils/bottles"
+
+describe Formulary do
+  let(:name) { "testball_bottle" }
+  let(:path) { CoreTap.new.formula_dir/"#{name}.rb" }
+  let(:content) {
```
alyssais: I think these could have slightly more descriptive names.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+    end
+
+    it "raises an error if the Formula cannot be found" do
+      expect {
+        described_class.factory("not_existed_formula")
+      }.to raise_error(FormulaUnavailableError)
+    end
+
+    context "if the Formula has the wrong class" do
+      let(:name) { "giraffe" }
```
alyssais: Correct be if I'm wrong, but I don't think `name` needs to be overridden here.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+
+    it "returns a Formula when given a path" do
+      expect(described_class.factory(path)).to be_kind_of(Formula)
+    end
+
+    it "returns a Formula when given a URL" do
+      begin
+       formula = shutup do
+         described_class.factory("file://#{path}")
+       end
```
alyssais: Could this be a single line?

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+
+       expect(formula).to be_kind_of(Formula)
+     ensure
+       formula.path.unlink
+     end
+    end
```
alyssais: Indentation's gone a bit weird here.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+
+    it "returns a Formula when given a URL" do
+      begin
+       formula = shutup do
+         described_class.factory("file://#{path}")
+       end
+
+       expect(formula).to be_kind_of(Formula)
+     ensure
+       formula.path.unlink
```
alyssais: I'm confused about why this is needed.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+     end
+    end
+
+    it "returns a Formula when given a bottle" do
```
alyssais: I think these names could be improved. We're testing a lot more than the class of it's return value.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+
+      it "raises an error if a Formula is in multiple Taps" do
+        begin
+          another_tap = Tap.new("homebrew", "bar")
+          (another_tap.path/"#{name}.rb").write content
+          expect {
+            described_class.factory(name)
+          }.to raise_error(TapFormulaAmbiguityError)
+        ensure
+          another_tap.path.rmtree
```
alyssais: Will this not automatically be removed by all the paths we delete between test?

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
+require "formula"
+require "formula_installer"
+require "utils/bottles"
+
+describe Formulary do
+  let(:name) { "testball_bottle" }
+  let(:path) { CoreTap.new.formula_dir/"#{name}.rb" }
+  let(:content) {
+    <<-EOS.undent
+      class #{described_class.class_s(name)} < Formula
```
alyssais: https://github.com/Homebrew/brew/pull/2044#discussion_r101815103 also applies here.

alyssais: https://github.com/Homebrew/brew/pull/2044#discussion_r101815103 also applies here.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,215 @@
...
+    it "prioritizes Formulae from pinned Taps" do
+      begin
+        tap.pin
+        formula = shutup do
+          described_class.find_with_priority(name)
+        end
+        expect(formula).to be_kind_of(Formula)
+        expect(formula.path).to eq(tap_path.realpath)
+      ensure
+        tap.pinned_symlink_path.parent.parent.rmtree
```
alyssais: Will this not automatically be removed when we delete a bunch of paths between tests?

Library/Homebrew/test/formulary_spec.rb
reitermarkus: At first I thought you must be right, but since the class is evaluated before with the first name, we need to use a separate name for this specific test.

Library/Homebrew/test/formulary_spec.rb
reitermarkus: It's not, good catch.

Library/Homebrew/test/formulary_spec.rb
reitermarkus: Taps aren't covered by this it seems.

Library/Homebrew/formulary.rb
```diff
@@ -4,7 +4,7 @@
 # The Formulary is responsible for creating instances of Formula.
 # It is not meant to be used directly from formulae.
 
-class Formulary
+module Formulary
```
alyssais: I think this should be a seperate PR. It's pretty hidden here and could have unintended side effects, so I think others should have a chance to be aware of this change.

Library/Homebrew/test/formulary_spec.rb
```diff
@@ -0,0 +1,219 @@
+require "formula"
+require "formula_installer"
+require "utils/bottles"
+
+describe Formulary do
+  let(:formula_name) { "testball_bottle" }
+  let(:formula_path) { CoreTap.new.formula_dir/"#{formula_name}.rb" }
+  let(:formula_content) {
```
alyssais: `do…end`?

Library/Homebrew/test/formulary_spec.rb
alyssais: Huh, okay. Leave it for now then, and maybe we can revisit that.

Library/Homebrew/formulary.rb
MikeMcQuaid: Agreed 👍

Library/Homebrew/test/formulary_spec.rb
MikeMcQuaid: Agreed 👍

PR #2030, opened 2017-02-16
Convert String test to spec.

reitermarkus:


Library/Homebrew/test/string_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "extend/string"
+
+describe String do
+  describe "#undent" do
+    it "removes leading whitespace, taking the first line as reference" do
+      expect(
+        <<-EOS.undent
```
alyssais: Very minor nit, but I'd combine this with the previous line.

Library/Homebrew/test/string_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      ).to eq("hello\ngoodbye\n\n")
+    end
+  end
+end
+
+describe StringInreplaceExtension do
+  subject { string.extend(described_class) }
+  let(:string) { "foobar" }
+
+  describe "#errors" do
```
alyssais: I'd say this is more testing the behaviour of `#sub!` than `#errors`.

Library/Homebrew/test/string_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    it "returns errors from #sub!" do
```
alyssais: it returns errors from `#sub!` <ins>when the string is not found</ins>.

Maybe we should have a test for the positive case too? Although since there currently isn't one I don't really mind either way since it's kind of out of this PR's scope.

Library/Homebrew/test/string_spec.rb
reitermarkus: Then `EOS` won't line up with anything.

alyssais: Homebrew convention seems to be to line it up with the indentation of its contents, e.g. https://github.com/Homebrew/homebrew-core/blob/fa4f85a5712dabdcc89522a8fd93dcb3797d1e57/Formula/awscli.rb

MikeMcQuaid: Agreed with @alyssais

PR #2034, opened 2017-02-16
Convert Gpg test to spec.

reitermarkus:


Library/Homebrew/test/gpg_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "gpg"
+
+describe Gpg do
```
alyssais: I think you should set a subject here.

Library/Homebrew/test/gpg_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  let(:dir) { Pathname.new(Dir.mktmpdir) }
```
alyssais: This should be deleted somewhere.

Library/Homebrew/test/gpg_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  describe "#create_test_key" do
+    it "creates a test key" do
```
alyssais: "it creates a test key <ins>in the home directory</ins>"?

PR #2035, opened 2017-02-16
Convert PkgVersion test to spec.

reitermarkus:


Library/Homebrew/test/pkg_version_spec.rb
```diff
@@ -0,0 +1,65 @@
+require "pkg_version"
+
+describe PkgVersion do
+  specify "#parse" do
```
alyssais: I feel like this could be split into named examples, but I can't think of good names off the top of my head. Thoughts?

Library/Homebrew/test/pkg_version_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  specify "#==" do
+    expect(PkgVersion.parse("1.0_0")).to be == PkgVersion.parse("1.0")
+    expect(PkgVersion.parse("1.0_1")).to be == PkgVersion.parse("1.0_1")
+  end
+
+  describe "#>" do
+    it "returns true if the left version is bigger than the right" do
+      expect(PkgVersion.parse("HEAD")).to be > PkgVersion.parse("1.0")
```
alyssais: I think this should be in its own named example.

Library/Homebrew/test/pkg_version_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      expect(PkgVersion.parse("1.1")).to be > PkgVersion.parse("1.0_1")
+    end
+
+    it "raises an error if the other side isn't of the same class" do
+      expect {
+        PkgVersion.new(Version.create("1.0"), 0) > Object.new
+      }.to raise_error(ArgumentError)
+
+      expect {
```
alyssais: I think this should be in its own named example to highlight that `PkgVersion` isn't compatible with `Version`.

Library/Homebrew/test/pkg_version_spec.rb
reitermarkus: Can't think of any good names either.

alyssais: Fair enough. Just leave it then. :)

PR #2037, opened 2017-02-16
Convert Utils test to spec.

reitermarkus:


Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
+require "utils"
+require "tempfile"
+require "utils/shell"
+
+RSpec::Matchers.alias_matcher :have_failed, :be_failed
```
alyssais: I think this should go somewhere global, as it's likely to be useful in more places than this (and it's modifying the global scope anyway which is not clear from its present location).

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+
+    specify "#ofail" do
+      begin
+        shutup do
+          ofail "foo"
+        end
+
+        expect(Homebrew).to have_failed
+      ensure
+        Homebrew.failed = false
```
alyssais: This should probably be done globally somewhere.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+
+module Utils
+  describe self do
+    let(:dir) { Pathname.new(Dir.mktmpdir) }
+
+    specify "#ofail" do
+      begin
+        shutup do
+          ofail "foo"
+        end
```
alyssais: Could this be one line?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+
+def esc(code)
+  /(\e\[\d+m)*\e\[#{code}m/
+end
+
+module Utils
+  describe self do
+    let(:dir) { Pathname.new(Dir.mktmpdir) }
+
+    specify "#ofail" do
```
alyssais: This could have a better name, e.g. `it "sets Homebrew.failed"`

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+
+    specify "#odie" do
+      expect(self).to receive(:exit).and_return(1)
+      shutup do
+        odie "foo"
+      end
+    end
+
+    describe "#pretty_installed" do
+      context "with HOMEBREW_NO_EMOJI unset and when $stdout is a TTY" do
```
alyssais: "when $stdout is a tty" could be its own context and the stub would only have to be specified once.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+          ENV.delete("HOMEBREW_NO_EMOJI")
+          expect(pretty_uninstalled("foo")).to match(/#{esc 1}foo #{esc 31}✘#{esc 0}/)
+        end
+      end
+
+      context "with HOMEBREW_NO_EMOJI set and when $stdout is a TTY" do
+        it "returns a string with colored info" do
+          allow($stdout).to receive(:tty?).and_return(true)
+          ENV["HOMEBREW_NO_EMOJI"] = "1"
+          expect(pretty_uninstalled("foo")).to match(/#{esc 1}foo \(uninstalled\)#{esc 0}/)
```
alyssais: This line is pretty long. Maybe break like e.g.

```ruby
expect(pretty_uninstalled("foo")).
  to match(/#{esc 1}foo \(uninstalled\)#{esc 0}/)
```

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
+require "utils"
+require "tempfile"
+require "utils/shell"
+
+RSpec::Matchers.alias_matcher :have_failed, :be_failed
+
+def esc(code)
```
alyssais: This is currently declaring this method globally on `Kernel`. It should be moved inside a `describe` block so it's scoped.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+        FileUtils.chmod 0755, shell
+
+        ENV["SHELL"] = shell
+
+        expect { interactive_shell }.not_to raise_error
+        expect(dir/"called").to exist
+      end
+    end
+
+    specify "#with_custom_locale" do
```
alyssais: This could have a better name.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+        # ~~ will fail because ~foo resolves to foo's home and there is no '~' user
+        path = ["~~", File.dirname(cmd)].join(File::PATH_SEPARATOR)
+        expect(which(File.basename(cmd), path)).to eq(cmd)
+      end
+    end
+
+    describe "#which_all" do
+      let(:cmd1) { dir/"foo" }
+      let(:cmd2) { dir/"bar/foo" }
+      let(:cmd3) { dir/"bar/baz/foo" }
```
alyssais: I think these could be named better based on what distinguishes them.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+        end
+      end
+    end
+
+    describe "#interactive_shell" do
+      let(:shell) { dir/"myshell" }
+
+      after(:each) do
+        dir.rmtree
+      end
```
alyssais: I think this should be done at the same scope `dir` is declared on. You can prevent accidentally initialising `dir` in the `after` block if it hasn't been used in the test like this:

```ruby
let(:dir) { @dir = Dir.mktmpdir }
after { @dir.unlink if @dir }
```

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+    specify "#capture_stderr" do
+      expect(
+        capture_stderr do
+          $stderr.print "test"
+        end,
+      ).to eq("test")
+    end
+
+    describe Shell do
+      subject { described_class }
```
alyssais: Is this line necessary?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+
+      describe "#shell_profile" do
+        before(:each) do
+          ENV["SHELL"] = shell
+        end
+
+        context "default" do
```
alyssais: I think these contexts add unnecessary indirection, and these tests should just be single examples that set `ENV["SHELL"]`.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+
+        context "Ksh" do
+          let(:shell) { "/bin/ksh" }
+          its(:shell_profile) { is_expected.to eq("~/.kshrc") }
+        end
+      end
+    end
+
+    describe "#popen_read" do
+      specify "without block" do
```
alyssais: These tests could use better names. Without block it does what?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+    describe "#popen_write" do
+      specify "with block" do
+        Utils.popen_write("grep", "-q", "success") do |pipe|
+          pipe.write("success\n")
+        end
+        expect($?).to be_a_success
+      end
+    end
+
+    specify "#pretty_duration" do
```
alyssais: Split these tests up if you can think of decent example names, but don't sweat it if not.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+      expect(number_readable(1_000)).to eq("1,000")
+      expect(number_readable(1_000_000)).to eq("1,000,000")
+    end
+
+    specify "#truncate_text_to_approximate_size" do
+      glue = "\n[...snip...]\n" # hard-coded copy from truncate_text_to_approximate_size
+      n = 20
+      long_s = "x" * 40
+
+      s = truncate_text_to_approximate_size(long_s, n)
```
alyssais: I think this should be multiple named examples because it's a little unclear what's going on.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,340 @@
...
+
+  class Bottles
+    describe Bintray do
+      specify "#package" do
+        expect(described_class.package("openssl@1.1")).to eq("openssl:1.1")
+        expect(described_class.package("gtk+")).to eq("gtkx")
+        expect(described_class.package("llvm")).to eq("llvm")
+
+        tap = Tap.new("homebrew", "bintray-test")
+        expect(described_class.repository(tap)).to eq("bottles-bintray-test")
```
alyssais: This seems like it's testing something other than `#package`?

Library/Homebrew/test/utils_spec.rb
reitermarkus: > it's a little unclear what's going on

Well, I have no idea what this method is actually doing. This was purely copy-pasted.

reitermarkus:
Moved the `Utils::Shell` specs into https://github.com/Homebrew/brew/pull/2044.

Library/Homebrew/test/spec_helper.rb
```diff
@@ -67,11 +69,15 @@
 
       files_after_test = Find.find(TEST_TMPDIR).map { |f| f.sub(TEST_TMPDIR, "") }
 
-      diff = Set.new(@__files_before_test).difference(Set.new(files_after_test))
+      diff = Set.new(@__files_before_test) ^ Set.new(files_after_test)
```
alyssais: Can you explain this change? :)

Library/Homebrew/test/spec_helper.rb
```diff
@@ -67,11 +69,15 @@
...
         file leak detected:
         #{diff.map { |f| "  #{f}" }.join("\n")}
       EOS
+
+      Homebrew.failed = @__homebrew_failed
     end
   end
 end
+
+RSpec::Matchers.alias_matcher :have_failed, :be_failed
```
alyssais: I think we should probably have a file just for matched definitions, since there are going to be a lot of them.

Library/Homebrew/test/utils/bottles/bintray_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "utils/bottles"
+
+describe Utils::Bottles::Bintray do
```
alyssais: Nice one breaking this into multiple files.

Library/Homebrew/test/utils/popen_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "utils/popen"
+
+describe Utils do
+  describe "::popen_read" do
+    it "reads the standard output of a given command" do
+      expect(subject.popen_read("sh", "-c", "echo success").chomp).to eq("success")
```
alyssais: This might just be because I'm on my phone (and therefore can't check exactly) but this line seems pretty long.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
+require "utils"
+
+describe self do
```
alyssais: I'm still really confused by `describe self`. How is this set?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+  let(:dir) { @dir = Pathname.new(Dir.mktmpdir) }
+
+  after(:each) do
+    dir.rmtree unless @dir.nil?
+  end
```
alyssais: Could this be one line? I think it's better for little things like this because it avoid drawing attention to them.

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+
+  def esc(code)
+    /(\e\[\d+m)*\e\[#{code}m/
+  end
+
+  describe "#ofail" do
+    it "sets Homebrew.failed to true" do
+      shutup do
+        ofail "foo"
+      end
```
alyssais: Could this be one line?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+      expect(Homebrew).to have_failed
+    end
+  end
+
+  describe "#odie" do
+    it "exits with 1" do
+      expect(self).to receive(:exit).and_return(1)
+      shutup do
+        odie "foo"
+      end
```
alyssais: Could this be one line?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+  describe "#pretty_installed" do
+    context "when $stdout is a TTY" do
+      before(:each) do
+        allow($stdout).to receive(:tty?).and_return(true)
+      end
+
+      context "with HOMEBREW_NO_EMOJI unset" do
+        before(:each) do
+          ENV.delete("HOMEBREW_NO_EMOJI")
+        end
```
alyssais: Could this be one line?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+
+        it "returns a string with a colored checkmark" do
+          expect(pretty_installed("foo"))
```
alyssais: Perhaps we should set a subject here, since we're always testing the same value?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+            .to match(/#{esc 1}foo #{esc 32}✔#{esc 0}/)
+        end
+      end
+
+      context "with HOMEBREW_NO_EMOJI set" do
+        before(:each) do
+          ENV["HOMEBREW_NO_EMOJI"] = "1"
+        end
```
alyssais: Could this be one line?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+  describe "#pretty_uninstalled" do
+    context "when $stdout is a TTY" do
+      before(:each) do
+        allow($stdout).to receive(:tty?).and_return(true)
+      end
+
+      context "with HOMEBREW_NO_EMOJI unset" do
+        before(:each) do
+          ENV.delete("HOMEBREW_NO_EMOJI")
+        end
```
alyssais: Could this be one line?

alyssais: Could this be one line?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+
+        it "returns a string with a colored checkmark" do
+          expect(pretty_uninstalled("foo"))
```
alyssais: Maybe set a subject here too?

Library/Homebrew/test/utils_spec.rb
alyssais: Could this be one line?

Library/Homebrew/test/utils_spec.rb
```diff
@@ -0,0 +1,293 @@
...
+      expect(gzip(somefile)[0].to_s).to eq("#{somefile}.gz")
+      expect(Pathname.new("#{somefile}.gz")).to exist
+    end
+  end
+
+  specify "#capture_stderr" do
+    expect(
+      capture_stderr do
+        $stderr.print "test"
+      end,
```
alyssais: A `do…end` inside parentheses looks pretty strange to me. Could this be done with braces in one or two lines?

Library/Homebrew/test/spec_helper.rb
reitermarkus: This is already in a separate PR.

Library/Homebrew/test/utils/popen_spec.rb
reitermarkus: Doesn't seem that long, actually.

alyssais: Yeah, you're right — it's 84 characters. Leave it as is. :)

alyssais:
@reitermarkus almost there! Just let me know what you think of https://github.com/Homebrew/brew/pull/2037#discussion_r101889288 and then this should be good to merge.

Library/Homebrew/test/spec_helper.rb
reitermarkus: Since most of these custom matchers are only used in one file, except for this specific one, I'd like to leave them in their respective specs, and after https://github.com/rspec/rspec-expectations/pull/974 is hopefully accepted, make them scoped. Otherwise, I think, it's easy to forget that they exist.

MikeMcQuaid: I'm fine with keeping them in the same file if they aren't repeated.

Library/Homebrew/test/utils/popen_spec.rb
MikeMcQuaid: This isn't ideal:
screen shot 2017-02-20 at 10 59 22

We do generally prefer <= 80 characters

reitermarkus: > This isn't ideal:

80 would still not fit into that comment view. I'd say 84 is still acceptable.

PR #2040, opened 2017-02-16
Convert Tty test to spec.

reitermarkus:


Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "utils"
+
+describe Tty do
+  specify "::strip_ansi" do
```
alyssais: Can we use spec names that describe the behaviour rather than just naming a method where possible? These cases seem pretty straightforward, e.g. "::strip_ansi removes ANSI escapes from a String".

Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    expect(described_class.strip_ansi("\033\[36;7mhello\033\[0m")).to eq("hello")
+  end
+
+  specify "::width" do
+    expect(described_class.width).to be_kind_of(Integer)
+    expect(described_class.width).to be >= 0
```
alyssais: I think these should be different tests, like before.

Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "utils"
+
+describe Tty do
```
alyssais: https://github.com/Homebrew/brew/pull/2044#discussion_r101815103 also applies here.

Library/Homebrew/test/utils/tty_spec.rb
alyssais: https://github.com/Homebrew/brew/pull/2044#discussion_r101814565 also applies here.

Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    it "doesn't truncate the text if the terminal is unsupported" do
+      allow(described_class).to receive(:width).and_return(0)
+      expect(described_class.truncate("foobar something very long")).to eq("foobar something very long")
+    end
+  end
+
+  context "$stdout is not a TTY" do
+    it "returns an empty string for all colors" do
+      allow($stdout).to receive(:tty?).and_return(false)
+      expect(described_class.to_s).to eq("")
```
alyssais: I think we should use `be_empty` here.

Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    it "doesn't truncate the text if the terminal is unsupported" do
+      allow(described_class).to receive(:width).and_return(0)
+      expect(described_class.truncate("foobar something very long")).to eq("foobar something very long")
+    end
+  end
+
+  context "$stdout is not a TTY" do
+    it "returns an empty string for all colors" do
+      allow($stdout).to receive(:tty?).and_return(false)
```
alyssais: This should be in a `before` block since it's doing what's described in the context.

Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect(described_class.blue.to_s).to eq("")
+      expect(described_class.magenta.to_s).to eq("")
+      expect(described_class.cyan.to_s).to eq("")
+      expect(described_class.default.to_s).to eq("")
+    end
+  end
+
+  context "$stdout is not a TTY" do
+    it "returns an empty string for all colors" do
+      allow($stdout).to receive(:tty?).and_return(true)
```
alyssais: This should be in a before block since it's doing what's described in the context.

Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      expect(described_class.to_s).to eq("")
```
alyssais: I think we should use `be_empty` here.

Library/Homebrew/test/utils/tty_spec.rb
reitermarkus: Not sure if that's better here, since `eq("")` tests if it is an empty string, not just if it's empty.

alyssais: It seems pretty unlikely to me that `#to_s` would return, e.g. an Array, but it's up to you.

Library/Homebrew/test/utils/tty_spec.rb
```diff
@@ -0,0 +1,69 @@
+require "utils"
+
+describe Tty do
+  subject { described_class }
```
alyssais: I don't _think_ this line is necessary. Haven't checked, but I think RSpec automatically sets `subject` like this when `described_class` is a module.

reitermarkus: Nice. 👍

Library/Homebrew/test/utils/tty_spec.rb
reitermarkus: I am aware that this is pretty unlikely, but I like getting two tests for the price of one. 😄

PR #2044, opened 2017-02-17
Convert Utils::Shell test to spec.

reitermarkus:


Library/Homebrew/test/utils/shell_spec.rb
```diff
@@ -0,0 +1,68 @@
+require "utils/shell"
+
+describe Utils::Shell do
+  describe "::path_to_shell" do
```
alyssais: I've never seen `describe "::method"` used for class methods before — it's usually `describe ".method" in my experience. It looks a bit strange to me since methods aren't typically referred to with `::`.

Thoughts?

Library/Homebrew/test/utils/shell_spec.rb
```diff
@@ -0,0 +1,68 @@
+require "utils/shell"
+
+describe Utils::Shell do
```
alyssais: I think it would be nicer to set a subject rather than calling all the methods on `described_class`:

```ruby
subject { Object.new.extend(described_class) }
```

Library/Homebrew/test/utils/shell_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  describe "::prepend_path_in_shell_profile" do
+    subject { described_class.prepend_path_in_shell_profile(path) }
+    let(:path) { "/path" }
+
+    before(:each) do
+      ENV["SHELL"] = shell
+    end
+
+    context "tcsh" do
```
alyssais: Splitting all these into contexts with a single example inside adds quite a bit of indirection, and I'm not sure what benefit it provides over just having tests like this:

```ruby
it "supports tcsh" do
  expect(described_class.prepend_path_in_shell_profile("/bin/tcsh")).
    to_start_with("echo 'setenv PATH /path")
end
```

Library/Homebrew/test/utils/shell_spec.rb
reitermarkus: On the contrary, I have never seen `.method`. RubyDoc uses `::` to describe class methods, e.g. https://ruby-doc.org/core-2.0.0/File.html.

Library/Homebrew/test/utils/shell_spec.rb
reitermarkus: Yes, that makes sense, since the subject in this case is actually the module.

Library/Homebrew/test/utils/shell_spec.rb
alyssais: Huh, okay, leave it as-is then.

I think I picked up my style from [here](https://github.com/thoughtbot/guides/tree/master/style/testing#unit-tests), way back in the day.

PR #2046, opened 2017-02-17
Convert FormulaPin test to spec.

reitermarkus:


Library/Homebrew/test/formula_pin_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "formula_pin"
+
+describe FormulaPin do
+  subject { described_class.new(formula) }
+  let(:formula) { formula_double_class.new }
```
alyssais: Couldn't we just use a `double`? This seems like a very complicated way to do a very simple thing.

reitermarkus: The thing with using a double is that the `installed_prefixes` and `installed_kegs` methods have to be evaluated at runtime, not before the test. No idea if/how to make doubles do that.

alyssais: I believe it's something like `allow(the_double).to receive(:installed_prefixes) { … }`

PR #2051, opened 2017-02-18
Convert Language::Python test to spec.

reitermarkus:


Library/Homebrew/test/language/python_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+  subject { described_class.new(formula, dir, "python") }
+
+  let(:dir) { Pathname.new(Dir.mktmpdir) }
+
+  let(:resource) { double("resource", stage: true) }
+  let(:formula_bin) { dir/"formula_bin" }
+  let(:formula) { double("formula", resource: resource, bin: formula_bin) }
+
+  after(:each) do
+    dir.rmtree
```
alyssais: If `dir` hasn't been used in a test (unlikely but possible in future), this will create the directory just to immediately destroy it.

Library/Homebrew/test/language/python_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      FileUtils.rm dir/"bin/python"
+    end
+  end
+
+  describe "#pip_install" do
+    it "accepts a string" do
+      expect(formula).to receive(:system) do |*args|
+        expect(args.first).to eq(dir/"bin/pip")
+        expect(args).to include("install")
+        expect(args.last).to eq("foo")
```
alyssais: Could we use `receive(…).with(…).and_return(…)` here? I think block syntax for stubs should be a last resort where RSpec doesn't have built-in methods to cover our use case.

Library/Homebrew/test/language/python_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+        true
+      end
+
+      subject.pip_install res
+    end
+  end
+
+  describe "#pip_install_and_link" do
+    let(:bin) { dir/"bin" }
+    let(:dest) { formula.bin }
```
alyssais: These names are a little confusing since they're both `bin`s but only one is called that.

Library/Homebrew/test/language/python_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      expect(bin/"kilroy").not_to exist
+      expect(dest/"kilroy").not_to exist
+
+      FileUtils.touch bin/"irrelevant"
+      bin_before = Dir.glob(bin/"*")
+      FileUtils.touch bin/"kilroy"
+      bin_after = Dir.glob(bin/"*")
+
+      expect(subject).to receive(:pip_install).with("foo")
+      expect(Dir).to receive(:[]).twice.and_return(bin_before, bin_after)
```
alyssais: I know it was there before, but I'm uneasy about this stub. Could we differentiate using arguments or something?

reitermarkus: Not really since the arguments are the same both times.

PR #2054, opened 2017-02-18
bump-formula-pr: improve duplicate detection

zmwangx:

-----

Reduce the chance of false flagging by making sure that the existing pr surfaced by `GitHub.issues_for_formula` actually contains the exact formula name in its title.

This for instance solves a problem I encountered just now where a pr for `imagemagick` blocks a new one for `imagemagick@6` (<s>unfortunately not the other way round</s>).

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -80,7 +80,8 @@ module Homebrew
 
   def fetch_pull_requests(formula)
     GitHub.issues_for_formula(formula.name, tap: formula.tap).select do |pr|
-      pr["html_url"].include?("/pull/")
+      pr["html_url"].include?("/pull/") &&
+        /(^|\s)#{Regexp.quote(formula.name)}(:|\s|$)/i =~ pr["title"]
     end
   rescue GitHub::RateLimitExceededError => e
     opoo e.message


MikeMcQuaid:
> (unfortunately not the other way round).

Would be good to fix that too. Could look for the name with either whitespace or an end/beginning of line immediately after/before. Thoughts?

ilovezfs:
It also should be case insensitive ideally but I'm not sure the Github API supports that.

zmwangx:
@MikeMcQuaid @ilovezfs Both are good advice, I'm now matching the title against `/(^|\s)#{Regexp.quote(formula.name)}(:|\s|$)/i`.

> but I'm not sure the Github API supports that.

GitHub API is case insensitive to begin with, isn't it? I fail to see the problem.

ilovezfs:
@zmwangx yup looks like you're right and it's case-insensitive
```
$ prtitles spidermonkey
Rename SpiderMonkey binary
```
so you can test against https://github.com/Homebrew/homebrew-core/pull/9606

zmwangx:
Sure,

```
$ brew bump-formula-pr --url=https://archive.mozilla.org/pub/mozilla.org/js/js185-1.0.1.tar.gz --sha256=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 spidermonkey
Error: These open pull requests may be duplicates:
Rename SpiderMonkey binary https://github.com/Homebrew/homebrew-core/pull/9606
Duplicate PRs should not be opened. Use --force to override this error.
Error: Kernel.exit
```

ilovezfs:
:tada:

MikeMcQuaid:
Thanks again @zmwangx!

PR #2055, opened 2017-02-18
Convert CompilerFailure test to spec.

reitermarkus:


Library/Homebrew/test/compiler_failure_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "compilers"
+
+describe CompilerFailure do
+  matcher :fail_with do |expected|
+    match do |actual|
+      actual.fails_with?(expected)
+    end
+  end
```
alyssais: Could this be an alias?

reitermarkus: Didn't think about it. Seems like `alias_matcher` can only handle predicates without arguments.

alyssais: That's… odd

reitermarkus: Ah, of course. I did `alias_matcher :fail_with, :fails_with` instead of `alias_matcher :fail_with, :be_fails_with.

PR #2057, opened 2017-02-19
ARGV: add missing "--force" arg to formula_install_option_names

### LINKED ISSUE #2053: `install --force` generates unnecessary warning
- [x] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

Pick an arbitrary formula, e.g., `wallpaper`, and forcibly install it: `brew install --force wallpaper`.  Observe: `Warning: wallpaper: this formula has no --force option so it will be ignored!`

This warning should not appear.  Rather, "--force" should be scrubbed out of the list of "options" passed to the formula (or should be whitelisted, or whatever).

This warning is especially jarring (and potentially confusing) when a user has just tried `brew install foo` (without the "--force") and been told by `brew` to "try again but with --force".  (E.g., perhaps `foo` is keg-only and and older version is already installed.)

JCount:

-----
This fixes an erroneous warning when running `brew install --force foo`. This would currently trigger a warning about the foo formula not having a `--force` option.
e.x. `Warning: foo: this formula has no --force option so it will be ignored!`

This fixes #2053

### DIFF
--- a/Library/Homebrew/extend/ARGV.rb
+++ b/Library/Homebrew/extend/ARGV.rb
@@ -16,11 +16,13 @@ module HomebrewArgvExtension
       --build-bottle
       --force-bottle
       --verbose
+      --force
       -i
       -v
       -d
       -g
       -s
+      -f
     ].freeze
   end
 


MikeMcQuaid:
Thanks @JCount 🎉

PR #1722, opened 2016-12-23
Added check for insecure mirror URLs

broder:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Should cover #564.

This change is similar to #1637 so this will need to be updated when that PR gets merged.

I haven't implemented the diff-ing done in https://github.com/vszakats/https-check/blob/master/chkhttps.sh - would that be something worth adding? I didn't think it would be likely that an http and https URL would serve different files...

PR #2067, opened 2017-02-20
audit: fix insecure mirror check when stdout is empty

ilovezfs:

-----

See https://github.com/Homebrew/brew/pull/1722#issuecomment-281096988.

curl_output's Open3.popen3 is returning

>["", "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (60) SSL certificate problem: Invalid certificate chain\nMore details here: http://curl.haxx.se/docs/sslcerts.html\n\ncurl performs SSL certificate verification by default, using a \"bundle\"\n of Certificate Authority (CA) public keys (CA certs). If the default\n bundle file isn't adequate, you can specify an alternate file\n using the --cacert option.\nIf this HTTPS server uses a certificate signed by a CA represented in\n the bundle, the certificate verification probably failed due to a\n problem with the certificate (it might be expired, or the name might\n not match the domain name in the URL).\nIf you'd like to turn off curl's verification of the certificate, use\n the -k (or --insecure) option.\n", #<Process::Status: pid 42968 exit 60>]

so we need to not die when the first argument is the empty string.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -1499,7 +1499,7 @@ class ResourceAuditor
     secure_url = url.sub "http", "https"
     secure_details = get_content_details(secure_url)
 
-    return if !details[:status].start_with?("2") || !secure_details[:status].start_with?("2")
+    return if details[:status].nil? || secure_details[:status].nil? || !details[:status].start_with?("2") || !secure_details[:status].start_with?("2")
 
     etag_match = details[:etag] && details[:etag] == secure_details[:etag]
     content_length_match = details[:content_length] && details[:content_length] == secure_details[:content_length]


broder:
Looks good - sorry for the bug!

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -1499,7 +1499,7 @@ def check_insecure_mirror(url)
     secure_url = url.sub "http", "https"
     secure_details = get_content_details(secure_url)
 
-    return if !details[:status].start_with?("2") || !secure_details[:status].start_with?("2")
+    return if details[:status].nil? || secure_details[:status].nil? || !details[:status].start_with?("2") || !secure_details[:status].start_with?("2")
```
MikeMcQuaid: Alternate version would be to use `.to_s.start_with?` but 🚢 either way.

MikeMcQuaid:
Thanks @ilovezfs.

ilovezfs:
@broder that's all right. Thank you for the apology.

Issue #2010, opened 2017-02-14
Ruby version dependency is ignored when rbenv is using system ruby


kke:
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [X] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

#### Summary
Ruby version dependency is ignored when rbenv is using "system" ruby.

#### Details

Having a formula that depends on a ruby version:

```ruby
class Foo < Formula
  depends_on :ruby => '2.4'
  def install
    system(which("ruby"), "--version")
  end
end
````

Using the "system" ruby with rbenv and trying to install the formula:

```sh
$ rbenv local system
$ brew install --verbose foo
==> /usr/bin/ruby --version
ruby 2.0.0p648 (2015-12-16 revision 53162) [universal.x86_64-darwin16]
```

It ignores the ruby dependency and points to the system ruby.

If you change to an rbenv ruby and try again:

```sh
$ rbenv local 2.3.1 # <-- switch to a non system ruby
$ brew install --verbose foo
==> /usr/local/opt/ruby/bin/ruby --version
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin16]  # <-- a non rbenv ruby
```

It correctly uses the version defined in `depends_on` (and it's not following the current rbenv ruby, as long as it's not "system")

### brew config
```sh
$ brew config
HOMEBREW_VERSION: 1.1.9-165-ge116ba196
ORIGIN: https://github.com/Homebrew/brew
HEAD: e116ba196e2834025c713a2a566e0e150907d6d0
Last commit: 23 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: a73d496eae43d1c015631d54243e6cf2266855ea
Core tap last commit: 16 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.0 build 800
Git: 2.11.1 => /usr/local/bin/git
Perl: /usr/bin/perl
[...]

PR #2083, opened 2017-02-21
Prepend suitable ruby to path

kke:
Fixes https://github.com/Homebrew/brew/issues/2010
Supports https://github.com/Homebrew/homebrew-core/pull/10031

Prepend satisfying ruby to PATH in RubyRequirement, as is done in [python](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/python_requirement.rb) and [java](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/java_requirement.rb) requirements.

### DIFF
--- a/Library/Homebrew/requirements/ruby_requirement.rb
+++ b/Library/Homebrew/requirements/ruby_requirement.rb
@@ -8,7 +8,7 @@ class RubyRequirement < Requirement
     super
   end
 
-  satisfy build_env: false { suitable_ruby }
+  satisfy(build_env: false) { suitable_ruby }
 
   env do
     ENV.prepend_path "PATH", suitable_ruby
--- a/Library/Homebrew/requirements/ruby_requirement.rb
+++ b/Library/Homebrew/requirements/ruby_requirement.rb
@@ -8,10 +8,10 @@ class RubyRequirement < Requirement
     super
   end
 
-  satisfy(build_env: false) { suitable_ruby }
+  satisfy(build_env: false) { new_enough_ruby }
 
   env do
-    ENV.prepend_path "PATH", suitable_ruby
+    ENV.prepend_path "PATH", new_enough_ruby
   end
 
   def message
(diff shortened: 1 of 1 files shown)


Library/Homebrew/requirements/ruby_requirement.rb
```diff
@@ -33,4 +31,34 @@ def display_s
...
+
+  def rubies
+    rubies = which_all("ruby")
+    if ruby_formula && ruby_formula.installed?
+      rubies.unshift Pathname.new(ruby_formula.bin/"ruby")
+    end
+    rubies.uniq
+  end
+
+  def suitable?(ruby)
```
MikeMcQuaid: Inline this method (as a variable if preferred) as it's only used once.

Library/Homebrew/requirements/ruby_requirement.rb
```diff
@@ -33,4 +31,34 @@ def display_s
...
+
+  private
+
+  def suitable_ruby
+    rubies.detect { |ruby| suitable?(ruby) }
+  end
+
+  def rubies
+    rubies = which_all("ruby")
+    if ruby_formula && ruby_formula.installed?
```
MikeMcQuaid: Turn `ruby_formula` into a variable and just use `Formula["ruby"]`; you can assume it'll never throw a `FormulaUnavailableError` exception.

Library/Homebrew/requirements/ruby_requirement.rb
```diff
@@ -33,4 +31,34 @@ def display_s
...
+      rubies.unshift Pathname.new(ruby_formula.bin/"ruby")
```
MikeMcQuaid: `rubies <<` is our preferable style, thanks. Also `ruby_formula.bin/"ruby"` is already a `Pathname` so don't need `Pathname.new`

Library/Homebrew/requirements/ruby_requirement.rb
```diff
@@ -33,4 +31,34 @@ def display_s
...
+    end
+    rubies.uniq
+  end
+
+  def suitable?(ruby)
+    version = Utils.popen_read(ruby, "-e", "print RUBY_VERSION").strip
+    version =~ /^\d+\.\d+/ && Version.create(version) >= min_version
+  end
+
+  def min_version
```
MikeMcQuaid: Inline this method (as a variable if preferred) as it's only used once.

Library/Homebrew/requirements/ruby_requirement.rb
```diff
@@ -33,4 +31,34 @@ def display_s
       name
     end
   end
+
+  private
+
+  def suitable_ruby
```
MikeMcQuaid: Maybe `new_enough_ruby`?

Library/Homebrew/requirements/ruby_requirement.rb
kke: `<<` is append. `unshift` is prepend.

Library/Homebrew/requirements/ruby_requirement.rb
kke: It's used for every ruby in the loop. The previous implementation does `Version.create` for the minimum version on every iteration. Of course there's usually maybe 1..5 rubies so it's not a big deal.

MikeMcQuaid:
@kke Thanks for your first contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #2104, opened 2017-02-22
Convert `brew desc` test to spec and add integration test helper methods.

reitermarkus:


Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
```diff
@@ -87,6 +87,80 @@ def brew(*args)
       status
     end
   end
+
+  def setup_test_formula(name, content = nil)
```
MikeMcQuaid: Where are these used?

Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -82,6 +82,7 @@ def tests
         "--format", "ParallelTests::RSpec::RuntimeLogger",
         "--out", "tmp/parallel_runtime_rspec.log"
       ]
+      spec_args << "--tag" << "~needs_official_cmd_taps" if ENV["HOMEBREW_TEST_OFFICIAL_CMD_TAPS"].nil?
```
MikeMcQuaid: What does this do? Avoid them being skipped or skip them with nicer syntax? If the latter: I'm not sure I agree. If the prior: I think it's actually worth them being noted as skipped so people are aware they aren't running a full test.

Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
reitermarkus: In many of the integration tests.

MikeMcQuaid: Would be 🆒 to migrate a single integration test at the same time to see usage.

Library/Homebrew/dev-cmd/tests.rb
reitermarkus: So still show then as skipped when not running with `--official-cmd-taps`, and not exclude them?

MikeMcQuaid: Yeh, I think so. I think this is different to the macOS ones because on Linux you're never going to be hitting that code.

Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
reitermarkus: Ok, will cherry-pick one in here.

PR #2105, opened 2017-02-23
Move superenv? into the ENV namespace.

woodruffw:

-----

I'm not sure if exposing `superenv?` outside of the `ENV` namespace was intentional, but moving it into `ENV` makes things slightly clearer. 

cc @sjackman for thoughts/confirmation that this fixes your problem.

### DIFF
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -78,7 +78,7 @@ class Build
 
     ENV.activate_extensions!
 
-    if superenv?
+    if ENV.superenv?
       ENV.keg_only_deps = keg_only_deps
       ENV.deps = formula_deps
       ENV.x11 = reqs.any? { |rq| rq.is_a?(X11Requirement) }
--- a/Library/Homebrew/cmd/--env.rb
+++ b/Library/Homebrew/cmd/--env.rb
@@ -10,7 +10,7 @@ module Homebrew
 
   def __env
     ENV.activate_extensions!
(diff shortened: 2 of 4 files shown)


ilovezfs:
Not thrilled with this since we're deprecating stdenv anyway.

sjackman:
My use case was a formula (`scipy`) that breaks if `FFLAGS` is set to any value at all, so my intended fix was:
```ruby
ENV["FFLAGS"] = nil unless ENV.superenv?
```

Bikeshedding: I think the name `ENV.super?` is perhaps a better name than the somewhat redundant `ENV.superenv?`

ilovezfs:
>My use case was a formula (scipy) that breaks with --env=std if FFLAGS is set to any value at all

That's not a supported use case ...

woodruffw:
> Not thrilled with this since we're deprecating stdenv anyway.

That's fair. 

@sjackman Can `FFLAGS` just be cleared out without adverse effect on superenv? If so, it might make sense to just unconditionally clear it.

MikeMcQuaid:
Yeh, I'm 👎 on this personally; I don't think it should be information available in formulae and https://github.com/Homebrew/brew/pull/1753 is going to help address this. Until then I'd say unconditional `ENV.delete("FFLAGS")` is your best bet.

woodruffw:
Got it, closing this.

sjackman:
Linuxbrew has superenv now (not default yet, but coming) so my use case isn't going to be around long.

PR #2106, opened 2017-02-23
Exclude executables from metafiles

naoty:

-----
Exclude executables in `#empty_installation?` to avoid `Empty Installation` error when only executable which name is the same as one of metafiles is installed. For example, if I define a formula for a package named `todo` like below, `brew install todo` will cause the `Empty Installation` error because the `todo` executable file will be handled as a metafile in `#empty_installation?`.

```rb
class Todo < Formula
  desc "A todo management tool just for myself"
  homepage "https://github.com/naoty/todo"
  url "https://github.com/naoty/todo/releases/download/0.2.0/todo.tar.gz"
  sha256 "be20e4069c0ae49998dfc00a010ca8f5d49d26193bd0c3e8611a4bf53cac469d"

  def install
    bin.install "todo"
  end
end
```

I also want to run a test for added changes but I don't know how to run tests.


Library/Homebrew/keg.rb
```diff
@@ -208,6 +208,7 @@ def ==(other)
 
   def empty_installation?
     Pathname.glob("#{path}/**/*") do |file|
+      return false if file.executable?
       next if file.directory?
       basename = file.basename.to_s
       next if Metafiles.copy?(basename)
```
MikeMcQuaid: I think this check (and perhaps also for `INSTALL_RECEIPT.json` below but not `.DS_Store`) could instead be adjusted to look only in the `prefix` rather than in all subdirectories instead of checking whether things are executable. The reasoning for this is that some software doesn't set permissions properly so files may be set as executable unintentionally and subvert this check.

MikeMcQuaid:
Thanks for the PR!

> I also want to run a test for added changes but I don't know how to run tests.

Run `brew tests`. You can see tests under `Library/Homebrew/test/`. Apologies that we're midway between a Minitest to RSpec conversion.

Library/Homebrew/keg.rb
naoty: Sorry, I could not understand "look only in the `prefix`". What should I look in the `prefix` instead of checking whether files are executable?

MikeMcQuaid: Sorry! I mean instead of using `Pathname.glob("#{path}/**/*")` above make check for metafiles just the root directory of the keg (which is the prefix) rather than in every subdirectory (i.e. `bin` in your case).

naoty: Oh, I see! Metafiles such as `readme` are under root directory not subdirectories. I will fix `#empty_installation?` to check metafiles under only root directory and add tests for it.

MikeMcQuaid: Great, thanks!

naoty:
@MikeMcQuaid I fixed to check for metafiles only in the root directory and added tests.

naoty:
@MikeMcQuaid Thanks! I merged `master` and moved tests into `keg_spec` and `pathname_spec`.

MikeMcQuaid:
@naoty Fantastic work here, really great. I hope to see more contributions from you!

Issue #2172, opened 2017-02-25
feature request: `brew ls --json|grep '"installed_on_request":true'`


timotheecour:
@MikeMcQuaid @ilovezfs 

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

couldn't find in docs or recent issues or PR's a way to list attributes for installed formulas; eg: (among many other use cases): I want to list all the formulas I've installed directly (via brew install X Y Z as opposed to via a dependency). I see https://github.com/Homebrew/brew/pull/1813/files stores the info (installed_as_dependency) but don't see a way to retrieve that info via cmd line. In fact `man brew|grep installed_on_request ` is empty.

* Option 1: `brew ls --installed_on_request ` to only show formulas tagged with `installed_on_request`
This fixes this particular issue but doesn't scale (eg what if i want to see unpinned formulas that are installed_on_request, etc)

So I prefer option 2 below: it's more flexible / scalable (simpler api: just a single `--json`). 

* Option 2: more generally, provide instead `brew ls --json` which would list all formulas along with their attributes formatted as json (1 per line). Then a user can call simply`brew ls --json|grep '"installed_on_request":true'`.  Note: other formats than json could be used if you don't like json, but would be nice to have a standard format to make it easier to integrate with other tools.
It could also accept an optional formula to only show attributes for this formula. 

Eg:
```
brew ls --json
{"name":"curses","full-name":"homebrew/dupes/ncurses","installed_on_request":true,"pinned":false,...}
{"name":"dmd","full-name":"dmd","installed_on_request":true,"pinned":false,...}
```

json fields: we could start with the most basic ones (eg as extracted from https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb ) and add more as needed, eg:
```
name
full-name
tap
version
installed_on_request
pinned
date-installed
install-time-seconds #how long did it take to install
number-of-files-generated
disk-usage # du -sh $(brew --cellar $formula)
binaries-generated # json array containing all binaries generated
install-failed #`brew install X` failed; could keep track of that
#etc
```

This would be extremely useful to have and expose the relevant data with very little added complexity for end-user (just a single `--json` flag).

## follow-up: mongo queries

[...]

PR #1973, opened 2017-02-08
Update CONTRIBUTING.md

timotheecour:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?

@ilovezfs would've made my PR https://github.com/Homebrew/brew/pull/1972/files just a bit easier

PR #2174, opened 2017-02-25
brew info --json=[v1|pretty|lines]

timotheecour:
@MikeMcQuaid @ilovezfs

- [x] Have you followed the guidelines in our [Contributing](https://github.com/Linuxbrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Linuxbrew/brew/pulls) for the same change?
-----
- [ ] follow steps in https://github.com/Homebrew/brew/pull/1973/files (Update CONTRIBUTING.md)

* Rationale: given in https://github.com/Homebrew/brew/issues/2172 (closed it even though current functionality was lacking: --json=v1 only dumps in a single line, which doesn't play well with common line oriented tools eg grep etc; and is slower (not line oriented...)

```
#show all installed_on_request formulas:
$ brew info --json=lines --installed|grep '"installed_on_request":true'

#show caveats:
$ brew info --json=lines --installed | grep -v '"caveats":null'

# show non-relocatable formulas:
brew info --json=lines --all | grep -v '"cellar":":any"' | rdmd --loop='line.parseJSON["name"].str.writeln'

# more readable output:
$ brew info --json=pretty dmd
[
  {
    "name": "dmd",
    "full_name": "dmd",
    "desc": "D programming language compiler for macOS",
    "homepage": "https://dlang.org/",
    "oldname": null,
    "aliases": [

    ],
    "versions": {
      "stable": "2.073.1",
      "bottle": false,
      "devel": null,
      "head": "HEAD"

```
etc.

This simple change gives quite powerful possibilities

* it also fixes a bug where `--json=whatever` will silently call print_info instead of saying you passed a wrong arg to json

* if this will be accepted, i can try to finish the TODO's

### DIFF
--- a/Library/Homebrew/cmd/info.rb
+++ b/Library/Homebrew/cmd/info.rb
@@ -7,8 +7,8 @@
 #:    To view formula history locally: `brew log -p <formula>`.
 #:
 #:  * `info` `--json=`<version> (`--all`|`--installed`|<formulae>):
-#:    Print a JSON representation of <formulae>. Currently the only accepted value
-#:    for <version> is `v1`.
+#:    Print a JSON representation of <formulae>.
+#:    Current options for <version>: `v1|pretty|lines`.
 #:
 #:    Pass `--all` to get information on all formulae, or `--installed` to get
 #:    information on all installed formulae.
@@ -28,10 +28,8 @@ module Homebrew
   module_function
 
   def info
-    # eventually we'll solidify an API, but we'll keep old versions
-    # awhile around for compatibility
-    if ARGV.json == "v1"
-      print_json
+    if ARGV.json
+      print_json ARGV.json
     elsif ARGV.flag? "--github"
       exec_browser(*ARGV.formulae.map { |f| github_info(f) })
     else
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/info.rb
```diff
@@ -7,8 +7,8 @@
 #:    To view formula history locally: `brew log -p <formula>`.
 #:
 #:  * `info` `--json=`<version> (`--all`|`--installed`|<formulae>):
-#:    Print a JSON representation of <formulae>. Currently the only accepted value
-#:    for <version> is `v1`.
+#:    Print a JSON representation of <formulae>.
+#:    Current options for <version>: `v1|pretty|lines`.
```
alyssais: I think this would have to include the version in its name somewhere. If we eventually add `--json=v2`, what does `pretty` return then?

MikeMcQuaid:
> This seems out of Homebrew's scope to me. There are tools like jq, or python -m json.tool (the latter comes with macOS) that you can pipe Homebrew's output into and get exactly what you're looking for.

I agree with this. Thanks for the PR but passing for now.

PR #2180, opened 2017-02-25
Add 'B' for bytes to cleanup command output

Tenzer:
When `brew cleanup` is run it prints out a message like the following:

> This operation has freed approximately 222M of disk space.

The '222M' refers to megabytes but the normal acronym for megabytes would be
'MB'. The 'B' is also missing from kilobytes and gigabytes in the output, so
that's what this commit adds.

-----


MikeMcQuaid:
Thanks for your first contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #2183, opened 2017-02-25
ARGV.include? should return boolean

vladshablinsky:

-----

Wouldn't it be better if `include?` returned `true` or `false`?


Library/Homebrew/extend/ARGV.rb
```diff
@@ -133,7 +133,7 @@ def kegs
 
   # self documenting perhaps?
```
reitermarkus: Maybe also remove this comment, while you're at it.

Issue #668, opened 2016-08-09
devel should auto-upgrade if needed, like stable does


ilovezfs:
Unfortunately the devel spec won't auto-upgrade itself for some reason when it's fulfilling the go dependency, so the Sierra failure came back as soon as core upgraded rc5 -> rc6. (simulated on El Capitan by setting :yosemite)

```
iMac-TMP:homebrew-core joe$ brew install -s godep
go: This formula either does not compile or function as expected on OS X
versions newer than Yosemite due to an upstream incompatibility.
Error: An unsatisfied requirement failed this build.
iMac-TMP:homebrew-core joe$ brew outdated go
go (1.7rc5) < 1.7rc6
iMac-TMP:homebrew-core joe$ brew upgrade go
==> Upgrading 1 outdated package, with result:
go 1.7rc6
==> Upgrading go
==> Downloading https://storage.googleapis.com/golang/go1.7rc6.src.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/go-1.7rc6.tar.gz
==> Downloading https://storage.googleapis.com/golang/go1.7rc6.darwin-amd64.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/go--gobootstrap-1.7rc6.tar.gz
==> ./make.bash --no-clean
==> /usr/local/Cellar/go/1.7rc6/bin/go install -race std
==> Cloning https://go.googlesource.com/tools.git
Updating /Users/joe/Library/Caches/Homebrew/go--gotools--git
==> Checking out branch release-branch.go1.7
==> go build
==> Caveats
As of go 1.2, a valid GOPATH is required to use the `go get` command:
  https://golang.org/doc/code.html#GOPATH

You may wish to add the GOROOT-based install location to your PATH:
  export PATH=$PATH:/usr/local/opt/go/libexec/bin
==> Summary
🍺  /usr/local/Cellar/go/1.7rc6: 6,435 files, 250.6M, built in 36 seconds
iMac-TMP:homebrew-core joe$ brew install -s godep
==> Downloading https://github.com/tools/godep/archive/v74.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/godep-74.tar.gz
==> go build -o /usr/local/Cellar/godep/74_1/bin/godep
🍺  /usr/local/Cellar/godep/74_1: 2 files, 8.6M, built in 1 second
```

ilovezfs:
bump.

[...]

PR #2191, opened 2017-02-25
Check the specs of dependencies if they're installed but not satisfied

vladshablinsky:

-----

# Description

Fixes https://github.com/Homebrew/brew/issues/668

About 18 tests fail now and 15 of them fail due to `FormulaUnavailableError`since `Dependency#use_best_spec!` uses `Formulary.factory` and tests don't expect that. Let's just skip it for now both with implementing new tests, since it's the approach which is important at this point. This PR is a possible solution of https://github.com/Homebrew/brew/issues/668. Can be pretty raw though.

Basically the idea is to mind specs of installed formulae for dependencies and requirements unless the `--ignore-installed-specs` flag is passed. The logic to determine the spec to use for dependency is implemented in `Formula#best_upgrade_spec_sym`.

Any feedback is apprecieated.

CC @ilovezfs @MikeMcQuaid

### DIFF
--- a/Library/Homebrew/dependency.rb
+++ b/Library/Homebrew/dependency.rb
@@ -1,4 +1,5 @@
 require "dependable"
+require "tab"
 
 # A dependency on another Homebrew formula.
 class Dependency
@@ -29,8 +30,48 @@ class Dependency
     name.hash ^ tags.hash
   end
 
+  def tap
+    CoreTap.instance
+  end
+
+  # The best spec we can upgrade the {Formula} to so it can satisfy dependency.
+  def closest_spec_for_dependency_upgrade_sym
+    rack = HOMEBREW_CELLAR.join(name.split("/").last)
+    return :stable unless rack.exist?
+
+    formula = Formulary.from_rack(rack)
+    return :stable if formula.tap != tap
+
+    active_spec_sym = formula.active_spec_sym
+    return active_spec_sym if active_spec_sym == :stable
+
+    installed_keg = formula.prefix
+    return active_spec_sym unless installed_keg.join(Tab::FILENAME).exist?
+
+    installed_tab = Tab.for_keg(installed_keg)
+    return :stable if installed_tab.version_scheme < formula.version_scheme
+
+    if active_spec_sym == :devel && formula.stable.version > installed_tab.devel_version
+      return :stable
+    end
+
+    active_spec_sym
+  end
+
+  def use_closest_for_dependency_upgrade_spec!
+    @spec = closest_spec_for_dependency_upgrade_sym
+  end
+
+  def use_spec!(spec)
+    @spec = spec
+  end
+
+  def spec
+    @spec || :stable
+  end
+
   def to_formula
-    formula = Formulary.factory(name)
+    formula = Formulary.factory(name, spec)
     formula.build = BuildOptions.new(options, formula.options)
     formula
   end
@@ -82,6 +123,8 @@ class Dependency
 
       deps.each do |dep|
         next if dependent.name == dep.name
+
+        dep.use_closest_for_dependency_upgrade_spec!
 
         case action(dependent, dep, &block)
         when :prune
@@ -136,7 +179,9 @@ class Dependency
         dep  = deps.first
         tags = merge_tags(deps)
         option_names = deps.flat_map(&:option_names).uniq
-        dep.class.new(name, tags, dep.env_proc, option_names)
+        result = dep.class.new(name, tags, dep.env_proc, option_names)
+        result.use_spec!(dep.spec)
+        result
       end
     end
 
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -437,6 +437,7 @@ class FormulaInstaller
         elsif req.build? && install_bottle_for_dependent
           Requirement.prune
         elsif install_requirement_formula?(req_dependency, req, install_bottle_for_dependent)
+          req_dependency.use_closest_for_dependency_upgrade_spec!
           deps.unshift(req_dependency)
           formulae.unshift(req_dependency.to_formula)
           Requirement.prune
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -111,11 +111,15 @@ class FormulaInstaller
       return false
     end
 
+    stable_deps_needed = formula.deps.reject do |dep|
+      dep.closest_spec_for_dependency_upgrade_sym == :stable
+    end.map(&:name)
+
     if formula.deps.any? { |dep| dep.closest_spec_for_dependency_upgrade_sym != :stable }
       if install_bottle_options[:warn]
         opoo <<-EOS.undent
           Building #{formula.full_name} from source:
-            The bottle needs stable dependencies which are not installed.
+            The bottle needs stable dependencies which are not installed (#{stable_deps_needed.join(", ")}).
         EOS
       end
       return false
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -111,6 +111,16 @@ class FormulaInstaller
       return false
     end
 
+    if formula.deps.any? { |dep| dep.closest_spec_for_dependency_upgrade_sym != :stable }
+      if install_bottle_options[:warn]
+        opoo <<-EOS.undent
+          Building #{formula.full_name} from source:
+            The bottle needs stable dependencies which are not installed.
+        EOS
+      end
+      return false
+    end
+
     true
   end
 
--- a/Library/Homebrew/test/dependency_expansion_spec.rb
+++ b/Library/Homebrew/test/dependency_expansion_spec.rb
@@ -118,6 +118,15 @@ describe Dependency do
     expect(described_class.expand(formula, [bar, baz])).to eq([bar, baz])
   end
 
+  it "resolves returns devel dep if closest spec is devel" do
+    foo = build_dep(:foo, [:build])
+    allow(foo).to receive(:closest_spec_for_dependency_upgrade_sym).and_return(:devel)
+
+    f = double(name: "f", deps: [foo])
+    expect(described_class.expand(f)).to eq([foo])
+    expect(foo.spec).to eq(:devel)
+  end
+
   it "doesn't raise an error when a dependency is cyclic" do
     foo = build_dep(:foo)
     bar = build_dep(:bar, [], [foo])
--- a/Library/Homebrew/test/dependency_spec.rb
+++ b/Library/Homebrew/test/dependency_spec.rb
@@ -92,6 +92,70 @@ describe Dependency do
     end
   end
 
+  it "finds closest spec to upgrade to" do
+    dep = described_class.new("foo", [:build])
+    f = formula("foo") { url "foo-1.0" }
+    stub_formula_loader f
+    allow(Formulary).to receive(:loader_for).with("homebrew/core/foo", from: :keg).and_return(double(get_formula: f))
+
+    HOMEBREW_CELLAR.join("foo/1.0").mkpath
+    compiler = DevelopmentTools.default_compiler
+    stdlib = :libcxx
+    tab = Tab.create(f, compiler, stdlib)
+    tab.write
+
+    expect(dep.closest_spec_for_dependency_upgrade_sym).to eq(:stable)
+  end
+
+  it "returns stable closest spec to upgrade to since taps differ" do
+    dep = described_class.new("foo", [:build])
+    f = formula("foo") do
+      url "foo-1.0"
+      version "1.0"
+      devel do
+        version "1.2"
+      end
+    end
+    stub_formula_loader f
+    f_tap = formula("foo") do
+      url "foo-1.0"
+    end
+    allow(f_tap).to receive(:tap).and_return(Tap.fetch("user/repo"))
+    allow(Formulary).to receive(:loader_for).with("user/repo/foo", from: :keg).and_return(double(get_formula: f_tap))
+
+    HOMEBREW_CELLAR.join("foo/1.1").mkpath
+    tab = Tab.empty
+    tab.tabfile = HOMEBREW_CELLAR/"foo/1.1/INSTALL_RECEIPT.json"
+    tab["source"]["tap"] = "user/repo"
+    tab["spec"] = "devel"
+    tab["versions"] = { "stable" => "1.0", "devel" => "1.1", "head" => "HEAD", "version_scheme" => 0 }
+    tab.write
+
+    expect(dep.closest_spec_for_dependency_upgrade_sym).to eq(:stable)
+  end
+
+  it "returns devel closest spec to upgrade" do
+    dep = described_class.new("foo", [:build])
+    f = formula("foo") do
+      url "foo-1.0"
+      version "1.0"
+      devel do
+        url "foo-1.2"
+        version "1.2"
+      end
+    end
+    stub_formula_loader f
+
+    HOMEBREW_CELLAR.join("foo/1.1").mkpath
+    tab = Tab.empty
+    tab.tabfile = HOMEBREW_CELLAR/"foo/1.1/INSTALL_RECEIPT.json"
+    tab["source"]["spec"] = "devel"
+    tab["versions"] = { "stable" => "1.0", "devel" => "1.1", "head" => "HEAD", "version_scheme" => 0 }
+    tab.write
+
+    expect(dep.closest_spec_for_dependency_upgrade_sym).to eq(:devel)
+  end
+
   specify "equality" do
     foo1 = described_class.new("foo")
     foo2 = described_class.new("foo")
--- a/Library/Homebrew/test/formula_spec.rb
+++ b/Library/Homebrew/test/formula_spec.rb
@@ -107,6 +107,22 @@ describe Formula do
     end
   end
 
+  describe "#determine_active_spec" do
+    let(:f) do
+      formula do
+        url "foo-1.0"
+      end
+    end
+
+    it "returns stable if no head defined" do
+      expect(f.send(:determine_active_spec, :devel)).to eq(f.stable)
+    end
+
+    it "returns stable if no head defined" do
+      expect(f.send(:determine_active_spec, :head)).to eq(f.stable)
+    end
+  end
+
   example "installed alias with core" do
     f = formula do
       url "foo-1.0"
--- a/Library/Homebrew/test/support/helper/formula.rb
+++ b/Library/Homebrew/test/support/helper/formula.rb
@@ -10,9 +10,35 @@ module Test
       # Use a stubbed {Formulary::FormulaLoader} to make a given formula be found
       # when loading from {Formulary} with `ref`.
       def stub_formula_loader(formula, ref = formula.full_name)
-        loader = double(get_formula: formula)
+        loader = double
+        allow(loader).to receive(:get_formula).and_return(formula)
+
+        allow(loader).to receive(:get_formula).with(:stable, alias_path: anything) do
+          begin
+            formula.active_spec = :stable
+          rescue FormulaSpecificationError
+            nil
+          end
+        end.and_return(formula)
+
+        allow(loader).to receive(:get_formula).with(:devel, alias_path: anything) do
+          begin
+            formula.active_spec = :devel
+          rescue FormulaSpecificationError
+            nil
+          end
+        end.and_return(formula)
+
+        allow(loader).to receive(:get_formula).with(:head, alias_path: anything) do
+          begin
+            formula.active_spec = :head
+          rescue FormulaSpecificationError
+            nil
+          end
+        end.and_return(formula)
         allow(Formulary).to receive(:loader_for).with(ref, from: :keg).and_return(loader)
         allow(Formulary).to receive(:loader_for).with(ref, from: nil).and_return(loader)
+        allow(Formulary).to receive(:loader_for).with(ref, from: :rack).and_return(loader)
       end
     end
   end


Library/Homebrew/dependency.rb
```diff
@@ -28,8 +30,26 @@ def hash
     name.hash ^ tags.hash
   end
 
+  def use_best_spec!
```
MikeMcQuaid: Better to use `newest_spec` or similar rather than `best_spec`.

Library/Homebrew/dependency.rb
```diff
@@ -28,8 +30,26 @@ def hash
...
+    @spec = Formulary.from_rack(HOMEBREW_CELLAR.join(name)).best_upgrade_spec_sym
+    @best_spec_used = true
+  end
+
+  def best_spec_used?
+    return @best_spec_used
```
MikeMcQuaid: No need for `return`

Library/Homebrew/extend/ARGV.rb
```diff
@@ -268,6 +268,10 @@ def fetch_head?
     include? "--fetch-HEAD"
   end
 
+  def ignore_installed_specs?
+    include? "--ignore-installed-specs"
```
MikeMcQuaid: It's not clear from the name what this does; can you elaborate and we might be able to think of a better one? When would you want to use this?

Library/Homebrew/formula.rb
```diff
@@ -476,6 +476,29 @@ def compiler_failures
     active_spec.compiler_failures
   end
 
+  # The best spec we can upgrade the {Formula} to so it can satisfy dependency.
+  def best_upgrade_spec_sym
```
MikeMcQuaid: It feels like this could/should maybe live in `upgrade` or `outdated`?

Library/Homebrew/formula.rb
```diff
@@ -476,6 +476,29 @@ def compiler_failures
...
+    require "tab"
```
MikeMcQuaid: I'd require this at the top of the file instead.

Library/Homebrew/extend/ARGV.rb
MikeMcQuaid: Also all the method calls for this are pretty invasive. I wonder if we can live without it?

vladshablinsky: Before this PR we just ignored the specs of recursive dependencies -- they all were resolved as stable. This PR introduces the logic where we don't ignore the specs for dependencies that are installed before we install some formula with that dependency. However if we pass this option -- we'll ignore installed specs and to satisfy dependencies we'll install stable versions for all unsatisfied dependencies. (The dependency is satisfied if its formula is `Formula#installed?`).

Library/Homebrew/formula.rb
vladshablinsky: Could you elaborate? I don't know the best example of similar method but let's discuss it. The method is primarily used in `Dependency`, but I thought for some formula it's formula that may know it's next/best/newest spec.

Library/Homebrew/dependency.rb
vladshablinsky: Newest is not what we actually mean here, see the comment to `Formula#best_upgrade_spec`

Library/Homebrew/formula.rb
MikeMcQuaid: Ok, I understand this a bit more. I think it should probably live on `Dependency` and be called something like `def oldest_spec_for_dependency_upgrade`

Maybe add this as a comment above this method (if I've understood correctly)
- If the dependency is installed as a `stable` version but outdated: `brew upgrade` to the latest stable version
- If the dependency is installed as a `devel` version but outdated:
  - if the latest `stable` version is newer than the installed `devel` version: `brew upgrade` to the latest stable version
  - if the latest `devel` version is newer than the installed `devel` version and latest `stable` version: `brew upgrade` to the latest devel version

I think HEAD shouldn't ever be upgraded to `stable`/`devel` or perhaps even the latest `HEAD` on `install` or `upgrade` of a reverse dependency.

I also think linking doesn't matter but opt-linking does.

Library/Homebrew/extend/ARGV.rb
MikeMcQuaid: Hmm. I guess I'm still not unsure when you would want to pass this option. I'd consider making this PR without it. CC @ilovezfs for thoughts.

Library/Homebrew/formula.rb
vladshablinsky: I think `HEAD` should be updated to stable when we can detect it's too outdated i.e. when stable have changed since we last installed `HEAD`. Thoughts? Can be easily modified to upgrading to the latest `HEAD` though.

> Maybe add this as a comment above this method (if I've understood correctly)

👍 I will, when we're happy with the approach.

> I also think linking doesn't matter but opt-linking does.

What do you mean?

> I think it should probably live on Dependency and be called something like def oldest_spec_for_dependency_upgrade

Initially the motivation behind placing this method into `Formula` was that we don't need to change `Dependency` in any way if we decide that we need some other logic to resolve specs for dependencies. That said, the same way we can just modify that single method inside `Dependency` and get that new behaviour. Maybe will think about a shorter name and will put it to `Dependency`.

MikeMcQuaid: > I think HEAD should be updated to stable when we can detect it's too outdated i.e. when stable have changed since we last installed HEAD

It's tough; I'm not sure how many people will want to always follow HEAD vs. just upgrade to stable. Would appreciate some other @Homebrew/maintainers thoughts here.

> What do you mean?

Don't care about `LinkedKegs` and just the `opt/foo` link i.e. don't care about `brew link` status.



> well, maybe this speaks for itself that indeed we need to put the method into Dependency. 😃

😁 I agree!

vladshablinsky: > Don't care about LinkedKegs and just the opt/foo link i.e. don't care about brew link status.

You mean for detecting the current spec? `Formulary#from_rack` should use `LinkedKegs` symlink if it's available and if it's not, then use optlink. But if they both exist, they are the same.

alyssais: > > I think HEAD should be updated to stable when we can detect it's too outdated i.e. when stable have changed since we last installed HEAD
>
> It's tough; I'm not sure how many people will want to always follow HEAD vs. just upgrade to stable.

When I install a HEAD formula, it's almost always because I need a specific thing that hasn't been released yet — so when a new stable version is released, I'd be happy to be switched to tracking stable.

However, there's also the use-case where you always want to use HEAD because you want to help find issues before they're released or whatever.

So I think maybe both options should be available somehow — both use-cases seem like something worth covering. I'd make the "update HEAD to stable" use-case the default, but maybe there could be some kind of `pin`-like functionality for forcing upgrades to a specific formula to track HEAD.

ilovezfs: I think if it's HEAD, we should just stay on HEAD.

tdsmith: I'd leave HEAD as HEAD; I think the most predictable things to do with HEAD are upgrade always or never, and probably "never" because I've installed llvm with --HEAD before :) I think it's reasonable for brew to abandon responsibility for upgrading formulas installed from head and rely on the judgement of the user.

ilovezfs: The main potentially ambiguous case I can think of is what to do when the HEAD spec has subsequently been removed from the formula.

vladshablinsky: > The main potentially ambiguous case I can think of is what to do when the HEAD spec has subsequently been removed from the formula.

Supposing the current implementation we have `HEAD` resolved to `HEAD` and if we don't have `HEAD` anymore, then `Formulary.factory` knows how to handle it and resolve everything properly even with `:head` spec passed.

> pin-like functionality for forcing upgrades to a specific formula to track HEAD.
@alyssais I think if a user wants to use these options then he's pretty much aware what's going on with his Cellar. So in my opinion it's better to satisfy whatever spec of dependencies he wants and then install some formula.

Library/Homebrew/extend/ARGV.rb
ilovezfs: I think the idea of the option is basically --use-old-behavior-before-pr-2191.

I think one issue we have with allowing, for example, a devel spec to fulfill a dependency is that we don't then force any reverse dependencies to be built from source, and will still pour bottles that may be specifically linked to the stable spec. So such an option would be a --get-me-out-of-linkage-hell-please-by-using-the-ones-that-actually-go-with-the-bottles-you're-pouring. Of course, the right answer is not to pour such bottles in the first place, and trigger the appropriate source builds.

Library/Homebrew/formula.rb
alyssais: Fair enough. 👍

MikeMcQuaid: Cool then I think stay on HEAD unless it's removed and then handle as expected. Also, remove the option as it adds a lot of code complexity for a use case I'm not sure we even want to support.

JCount: Just to give my opinion as well, I believe the decision to stay on the HEAD spec is the correct one.

MikeMcQuaid:
@vladshablinsky Any thoughts here?

vladshablinsky:
@MikeMcQuaid keeping in mind this PR, will get back as soon as I'm finished with my university stuff in the middle of which I am now.

MikeMcQuaid:
@vladshablinsky Cool, thanks

vladshablinsky:
Sorry for this taking so long. Getting back to it.

Library/Homebrew/extend/ARGV.rb
vladshablinsky: @ilovezfs so, both with supporting the specs we need to think of the way we trigger builds when dependency is satisfied, but not stable, right?

E.g.

`go-1.7rc5` installed and upgraded to `go-1.7rc6` while installing `godep`, but `godep` need to be built from source here, right?

vladshablinsky:
@ilovezfs how does this messaging look?

```bash
➜ vlad:Homebrew$ brew install godep                                            allow_spec_deps_reqs
Error: godep contains not stable dependencies:
  go is not stable, stable is 1.6.3
You can either downgrade them to stables using `brew install` or you can pass:
  --build-from-source to build formula against non-stable dependencies
  --ignore-installed-specs to use stable specs
  --force-bottle to install formula with non-stable dependencies
```

Library/Homebrew/extend/ARGV.rb
vladshablinsky: This have been discussed with @ilovezfs and we agreed that we need some warnings when we try to pour bottle with non-stable dependencies, since the bottle is actually built against stables.

MikeMcQuaid: Warnings need reasonable but I still don't see why we need this option?

monouser7dig:
to me it would be clearer if it said 
--build-from-source to build **godep** formula against non-stable dependencies

otherwise it seems as if you're still referring to go 
(as I understand it, 
> You can either downgrade them to stables using `brew install`
refers to go)

Library/Homebrew/extend/ARGV.rb
vladshablinsky: I can explain. The bottles we have for formulae are built against stable versions of their dependencies, right? `Formula#installed?` is used to see if a dependency is satisfied. The method can return true even if dependency is not stable potentially breaking the bottle we're pouring. But let's skip it for now. Let's skip it for now and consider #668. To allow this we need to mind dependencies specs that are installed and upgrade them properly. Doing so can break the bottle we pour. So, here is how we can solve the problems:
* We used to resolve dependencies as stable and install them as stable unless they were `satisfied?`. Which caused the #668.
* Instead we decided to mind the specs of installed dependencies and upgrade them properly
* If we pour a bottle and its dependencies are not stable, we should warn and ask user to specify some option:
1. build formula against dependency
2. pour bottle regardless it being broken
3. don't mind installed specs of dependencies and install stable dependencies against which the bottle we provide is built.

ilovezfs: The option replaces a multi-step procedure with a single-step procedure.

```
brew install --ignore-installed-specs godep
```

vs.

```
brew unlink go
brew install go
brew install godep
```

So we don't _need_ the option if the multi-step procedure is an acceptable UX.

MikeMcQuaid: > we should warn and ask user to specify some option:

I don't think we should. 2. is a bad option. I think we should always do 3. unless `--build-from-source` is passed. I may well be missing something here.

ilovezfs: brew install --devel go
brew install godep

will immediately downgrade go to stable. That is terrible default behavior.

MikeMcQuaid: Then build `godep` from source by default if a `devel` is spec is installed if we know (not speculate) that in the majority of cases this is known to be broken using a `devel` go with stable `go`. If we do that we should also probably ensure users fall back to `stable` as soon as there's a newer `stable` than the `devel` when they first `brew install --devel`d.

Alternatively if we don't actually know this is broken: just always use whatever version is installed (`stable`, `devel` or `head`) for dependencies and not care until we get user reported breakages and then figure out a solution for those cases. I don't think speculative bug fixing is useful.

ilovezfs: The typical case if we don't trigger source builds will be the bottles pours, and you get some broken dylib warnings.

vladshablinsky: So summing up which options do we leave?

monouser7dig: I think your three options offer the most flexibility, given what @MikeMcQuaid just added, it would probably be good to add hints to each like "recommended" "not recommended" to make it easier for novices (like me).
```
--build-from-source to build formula against non-stable dependencies (recommended)
--force-bottle to install formula with non-stable dependencie (overrides devel version)
--ignore-installed-specs to use stable specs (not recommended)
```

The fallback behaviour to *stable* seems like a very comfortable approach to me as well.
> ensure users fall back to stable as soon as there's a newer stable than the devel

situation is:
1a godep depends on go
1b go is installed as devel
1c godep is e.g. installed -s as a recommended option
2a asa go has a new version, that version is installed by regular upgrade process (right?)
2b godep is upgraded to the home-brew bottle build (would have to be implemented on your side I guess)

One thing that came ti my mind, how are HEAD versions handled in this situation? Just the same as devel? maybe this is off topic but yeah...

vladshablinsky: > The fallback behaviour to stable seems like a very comfortable approach to me as well. 

This is what we do now.

> 1a godep depends on go
1b go is currently installed as devel
1c godep gets e.g. installed -s as a recommended option
2a asa go has a new version, that version is installed by regular upgrade process (right?)
2b godep is upgraded to the home-brew bottle build (would have to be implemented on your side I guess)

At `2b` brew upgrade `godep` should leave us with stable go and poured bottle.


> One thing that came to my mind, how are HEAD versions handled in this situation? Just the same as devel? maybe this is off topic but yeah...

HEADs have been already discussed in the outdated diff right after this one. HEAD stays HEAD.

MikeMcQuaid: @vladshablinsky Sounds like to avoid breakage and prioritise safety: always build from source and don't allow an override to use the bottles (or just continue to ensure `--force-bottle` works) so we don't need an option.

vladshablinsky:
Updated.

Now we build from source formula if it has non-stable dependencies installed (allowing to use `--force-bottle`) and take into account installed specs (fixes #668).

Any guidance on how to approach fixing the tests is appreciated. There are plenty of them broken for now since `Dependency#expand` now uses `to_formula`. What is the best way to fix this? The problem is in `Dependency#closest_for_dependency_upgrade_spec!` method and in it's trace.

Library/Homebrew/formula_installer.rb
```diff
@@ -110,6 +110,16 @@ def pour_bottle?(install_bottle_options = { warn: false })
       return false
     end
 
+    if formula.deps.any? { |dep| dep.closest_spec_for_dependency_upgrade_sym != :stable }
+      if install_bottle_options[:warn]
+        opoo <<-EOS.undent
+          Building #{formula.full_name} from source:
+            The bottle needs stable dependencies while installed are not stable.
```
MikeMcQuaid: `The bottle needs stable dependencies which are not installed.`?

Library/Homebrew/dependency.rb
```diff
@@ -28,8 +30,45 @@ def hash
     name.hash ^ tags.hash
   end
 
+  # The best spec we can upgrade the {Formula} to so it can satisfy dependency.
+  def closest_spec_for_dependency_upgrade_sym
+    formula = Formulary.from_rack(HOMEBREW_CELLAR.join(name))
```
MikeMcQuaid: `HOMEBREW_CELLAR/name`

Library/Homebrew/dependency.rb
```diff
@@ -28,8 +30,45 @@ def hash
...
+    active_spec_sym = formula.active_spec_sym
+
+    return :stable if active_spec_sym == :stable
+    return formula.active_spec_sym unless formula.any_version_installed?
+
+    installed_keg = formula.prefix
+    return formula.active_spec_sym unless installed_keg.join(Tab::FILENAME).file?
```
MikeMcQuaid: `(installed_keg/Tab::FILENAME).exist?`

MikeMcQuaid:
> Any guidance on how to approach fixing the tests is appreciated. There are plenty of them broken for now since Dependency#expand now uses to_formula. What is the best way to fix this? The problem is in Dependency#closest_for_dependency_upgrade_spec! method and in its trace.

I don't see the use of `to_formula`? I think the methods could be adapted to handle the formulae not existing or, probably better to do as well, the tests adjusted so the formulae do exist. There's also `brew style` issues and just methods with differently expected inputs. Can you try to fix up those you can and I'll help with the rest?

vladshablinsky:
> I don't see the use of to_formula?

@MikeMcQuaid I meant this one:

https://github.com/vladshablinsky/brew/blob/178fd0a46ca03fe48fed32514dac8e682094eb3f/Library/Homebrew/dependency.rb#L35

> Can you try to fix up those you can and I'll help with the rest?

Yeah, sure. Just forgot about `brew style`.

vladshablinsky:
Less tests failing now. Will try to figure out this myself and ask for help if I can't.

vladshablinsky:
Updated.

Library/Homebrew/dependency.rb
```diff
@@ -28,8 +30,49 @@ def hash
...
+    CoreTap.instance
+  end
+
+  # The best spec we can upgrade the {Formula} to so it can satisfy dependency.
+  def closest_spec_for_dependency_upgrade_sym
+    rack = HOMEBREW_CELLAR.join(name.split("/").last)
+    return :stable unless rack.exist?
+
+    formula = Formulary.from_rack(rack)
+    return :stable unless formula.tap == tap
```
MikeMcQuaid: `if formula.tap != tap`

Library/Homebrew/dependency.rb
```diff
@@ -172,13 +219,15 @@ def merge_temporality(deps)
...
-  attr_reader :tap
-
   def initialize(name, tags = [], env_proc = DEFAULT_ENV_PROC, option_names = [name.split("/").last])
     @tap = Tap.fetch(name.rpartition("/").first)
     super(name, tags, env_proc, option_names)
   end
 
+  def tap
+    @tap
+  end
```
MikeMcQuaid: Why is this needed? The `attr_reader` should do exactly this.

Library/Homebrew/dependency.rb
```diff
@@ -10,6 +11,7 @@ class Dependency
 
   def initialize(name, tags = [], env_proc = DEFAULT_ENV_PROC, option_names = [name])
     @name = name
+    @spec = nil
```
MikeMcQuaid: This is implicit; instance variables will always be `nil` on `initialize`.

MikeMcQuaid:
@vladshablinsky Check out `brew style`, there's failures there. You may want to rebase this, too.

vladshablinsky:
> Check out brew style, there's failures there. You may want to rebase this, too.

Can't we ignore these? 

 ```
test/support/helper/formula.rb:19:11: W: Do not suppress exceptions.
          rescue FormulaSpecificationError
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test/support/helper/formula.rb:26:11: W: Do not suppress exceptions.
          rescue FormulaSpecificationError
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
test/support/helper/formula.rb:33:11: W: Do not suppress exceptions.
          rescue FormulaSpecificationError
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

**Comment by MikeMcQuaid** [comment]
> Can't we ignore these?

Add a `nil`/`false`/whatever into that block to make it explicit what you're wanting to pass back when the exception is being suppressed and think about if it's appropriate to e.g. output a message for Homebrew developers in each case.

**Comment by vladshablinsky** [comment]
> think about if it's appropriate to e.g. output a message for Homebrew developers in each case.

Initially I thought that I didn't need to surround the code with begin/end/rescue, but I then I thought that we needed. As about messaging -- I'm not sure here. Maybe we either not rescuing or just don't output anything?

**Comment by MikeMcQuaid** [comment]
> As about messaging -- I'm not sure here. Maybe we either not rescuing or just don't output anything?

Yeh, I think either those and/or a comment are reasonable.

File: Library/Homebrew/dependency.rb [hunk 12]
```diff
@@ -28,8 +29,49 @@ def hash
     name.hash ^ tags.hash
   end
 
+  def tap
+    CoreTap.instance
```
MikeMcQuaid: Presumably this is because `TapDependency` sets this to the relevant tap?

MikeMcQuaid:
@vladshablinsky This now looks good to me, thanks for all your work on this. Can you provide some `brew install` commands you've been using to test this?

Library/Homebrew/dependency.rb
```diff
@@ -28,8 +29,49 @@ def hash
...
+  # The best spec we can upgrade the {Formula} to so it can satisfy dependency.
+  def closest_spec_for_dependency_upgrade_sym
+    rack = HOMEBREW_CELLAR.join(name.split("/").last)
+    return :stable unless rack.exist?
+
+    formula = Formulary.from_rack(rack)
+    return :stable if formula.tap != tap
+
+    active_spec_sym = formula.active_spec_sym
+    return :stable if active_spec_sym == :stable
```
MikeMcQuaid: `return active_spec_sym if active_spec_sym == :stable` perhaps?

Library/Homebrew/dependency.rb
```diff
@@ -28,8 +29,49 @@ def hash
...
+    return formula.active_spec_sym unless formula.any_version_installed?
```
MikeMcQuaid: `return active_spec_sym unless formula.any_version_installed?`? Also, not sure I understand how this would be hit if we have `return :stable unless rack.exist?` above?

Library/Homebrew/dependency.rb
```diff
@@ -28,8 +29,49 @@ def hash
...
+
+    installed_keg = formula.prefix
+    return formula.active_spec_sym unless installed_keg.join(Tab::FILENAME).exist?
```
MikeMcQuaid: `return active_spec_sym unless installed_keg.join(Tab::FILENAME).exist?`?

ilovezfs:
@vladshablinsky is `closest_spec_for_dependency_upgrade_sym` handling the case where the user currently has a head version installed, but the head spec has subsequently been removed from the formula?

Also, the same question applies to any spec really. If the user has a devel version installed, but the devel spec has subsequently been removed, or (unlikely) the user has a stable version installed, but the stable spec has subsequently been removed, is `closest_spec_for_dependency_upgrade_sym` handling those cases?

Library/Homebrew/formula_installer.rb
```diff
@@ -110,6 +110,16 @@ def pour_bottle?(install_bottle_options = { warn: false })
       return false
     end
 
+    if formula.deps.any? { |dep| dep.closest_spec_for_dependency_upgrade_sym != :stable }
+      if install_bottle_options[:warn]
+        opoo <<-EOS.undent
+          Building #{formula.full_name} from source:
+            The bottle needs stable dependencies which are not installed.
```
ilovezfs: Should we say which ones? (e.g. in case the user wants to control+C and uninstall the offending formula)

vladshablinsky:
> Can you provide some brew install commands you've been using to test this?

I'll return from vacation next week and will finish this up. As about the commands -- not sure about the particular commands we can use. I used some git checkouts to specific commits and patching formulae to test this. Will provide with them as soon as I'm back, but basically it's installing some devel version of the formula and then updating `version` for formula file. Thanks.

MikeMcQuaid:
> As about the commands -- not sure about the particular commands we can use. I used some git checkouts to specific commits and patching formulae to test this. Will provide with them as soon as I'm back, but basically it's installing some devel version of the formula and then updating version for formula file.

Yeh, even PRs or Gists of modified formulae for testing and the commands you use would be useful, thanks!

vladshablinsky:
> Yeh, even PRs or Gists of modified formulae for testing and the commands you use would be useful, thanks!

https://github.com/Homebrew/brew/issues/668#issuecomment-278685135

```
bash-4.4$ pwd
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core
bash-4.4$ git reset --hard origin/master
HEAD is now at e420e60034 gtk-vnc: update 0.7.0 bottle.
bash-4.4$ brew uninstall go godep --ignore-dependencies -f
bash-4.4$ git checkout 9476215 -- Formula/go.rb Formula/godep.rb
bash-4.4$ perl -p -i -e 's/(MaximumMacOSRequirement => ):el_capitan/\1:yosemite/' Formula/go.rb
bash-4.4$ brew instal -s godep
go: This formula either does not compile or function as expected on macOS
versions newer than Yosemite due to an upstream incompatibility.
Error: An unsatisfied requirement failed this build.
bash-4.4$ brew instal -s go
go: This formula either does not compile or function as expected on macOS
versions newer than Yosemite due to an upstream incompatibility.
Error: An unsatisfied requirement failed this build.
bash-4.4$ brew instal --devel go
==> Using the sandbox
==> Downloading https://storage.googleapis.com/golang/go1.7rc5.src.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/go-1.7rc5.tar.gz
==> Downloading https://storage.googleapis.com/golang/go1.7rc5.darwin-amd64.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/go--gobootstrap-1.7rc5.tar.gz
==> ./make.bash --no-clean
==> /usr/local/Cellar/go/1.7rc5/bin/go install -race std
==> Cloning https://go.googlesource.com/tools.git
Updating /Users/joe/Library/Caches/Homebrew/go--gotools--git
==> Checking out branch release-branch.go1.7
==> go build
==> Caveats
As of go 1.2, a valid GOPATH is required to use the `go get` command:
  https://golang.org/doc/code.html#GOPATH

You may wish to add the GOROOT-based install location to your PATH:
  export PATH=$PATH:/usr/local/opt/go/libexec/bin
==> Summary
🍺  /usr/local/Cellar/go/1.7rc5: 6,435 files, 250.6M, built in 37 seconds
bash-4.4$ brew instal -s godep
==> Using the sandbox
==> Downloading https://github.com/tools/godep/archive/v74.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/godep-74.tar.gz
==> go build -o /usr/local/Cellar/godep/74_1/bin/godep
🍺  /usr/local/Cellar/godep/74_1: 3 files, 8.6M, built in 1 second
bash-4.4$ brew uninstall godep
Uninstalling /usr/local/Cellar/godep/74_1... (3 files, 8.6M)
bash-4.4$ git checkout f98619c -- Formula/go.rb Formula/godep.rb
bash-4.4$ perl -p -i -e 's/(MaximumMacOSRequirement => ):el_capitan/\1:yosemite/' Formula/go.rb
bash-4.4$ brew instal -s godep
go: This formula either does not compile or function as expected on macOS
versions newer than Yosemite due to an upstream incompatibility.
Error: An unsatisfied requirement failed this build.
```

Error should not be displayed now.


> is closest_spec_for_dependency_upgrade_sym handling the case where the user currently has a head version installed, but the head spec has subsequently been removed from the formula?

Yes, it is. We pass the spec to instantiate the formula and it's handled properly by `Formula` itself.

Library/Homebrew/dependency.rb
MikeMcQuaid: @vladshablinsky Ping on this comment, thanks.

MikeMcQuaid:
@vladshablinsky Once conflicts are addressed, tests are 💚 and you're confident in this change (and able to hang around for a little bit afterwards and available in the following days to address any bugs/regressions quickly): 👍 to merge. Thanks for the work here!

MikeMcQuaid:
@vladshablinsky Any news here?

PR #324, opened 2016-06-06
tests: bypass local GPG signing defaults

DomT4:
- [x] Have you followed the guidelines in our [Contributing](httxps://github.com/Homebrew/brew/blob/master/.github/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you successfully run `brew tests` with your changes locally?

-----

This might be a mildly controversial change, but here's a PR. Explained in the commit summaries, but if you have:

```
[commit]
	gpgsign = true
```

Set in your `.gitconfig` either `brew tests` will hang waiting for your key password, or if you leave `gpg-agent` running with a timeout it'll sign the commits automatically, which isn't particularly desired behaviour.

It's a minor edge case but since it's easy enough to stamp out the change seems harmless. This flag has been supported since 2011, git 1.7.9, so should be safe to adopt unconditionally.

`test_tap` actually fails hard with that config setting in place due to the actual hash deviating from the expected hash.

PR #2197, opened 2017-02-26
Tests: prevent `git commit` from GPG signing

mistydemeo:

-----

If the user has GPG signing on by default, git will try to get the user to type their passphrase when signing commits on every test that runs `git commit`. That obviously sucks and it prevents the tests from passing.

### DIFF
--- a/Library/Homebrew/test/cmd/bundle_spec.rb
+++ b/Library/Homebrew/test/cmd/bundle_spec.rb
@@ -6,7 +6,7 @@ describe "brew bundle", :integration_test, :needs_test_cmd_taps do
       HOMEBREW_REPOSITORY.cd do
         shutup do
           system "git", "init"
-          system "git", "commit", "--allow-empty", "-m", "This is a test commit"
+          system "git", "commit", "--no-gpg-sign", "--allow-empty", "-m", "This is a test commit"
         end
       end
 
--- a/Library/Homebrew/test/cmd/install_spec.rb
+++ b/Library/Homebrew/test/cmd/install_spec.rb
@@ -162,7 +162,7 @@ describe "brew install", :integration_test do
         FileUtils.touch "bin/something.bin"
         FileUtils.touch "README"
         system "git", "add", "--all"
-        system "git", "commit", "-m", "Initial repo commit"
+        system "git", "commit", "--no-gpg-sign", "-m", "Initial repo commit"
       end
     end
 
--- a/Library/Homebrew/test/cmd/log_spec.rb
+++ b/Library/Homebrew/test/cmd/log_spec.rb
@@ -3,7 +3,7 @@ describe "brew log", :integration_test do
     HOMEBREW_REPOSITORY.cd do
       shutup do
         system "git", "init"
-        system "git", "commit", "--allow-empty", "-m", "This is a test commit"
+        system "git", "commit", "--no-gpg-sign", "--allow-empty", "-m", "This is a test commit"
       end
     end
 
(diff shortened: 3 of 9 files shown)


mistydemeo:
> Is there an environment variable we can set for this instead?

I've been reading up on this, but it looks like the answer is no.

> Alternatively, it'd be good to override the configuration file used so it doesn't rely on user parameters.

We can set `GIT_CONFIG` to do this - I'll do some testing to make sure it doesn't have unexpected effects.

MikeMcQuaid:
> We can set GIT_CONFIG to do this - I'll do some testing to make sure it doesn't have unexpected effects.

Thanks!

DomT4:
> Is there an environment variable we can set for this instead?

See previous: https://github.com/Homebrew/brew/pull/332, https://github.com/Homebrew/brew/pull/324.

MikeMcQuaid:
Closing this for now just because it's been a while with no changes. Feel free to reopen when it's changing.

alyssais:
This has been annoying me forever too, and I'd love to put some work into finally getting it fixed. My understanding is that `GIT_CONFIG` is only supported by the `git config` command, not Git as a whole (see https://github.com/git/git/commit/dc87183), so the options I can think of are as follows:

- Add `--no-gpgsign` to every `git commit` call in the tests
- Add `-c commit.gpgsign=false` to every `git commit` call in the tests (might be preferable to the above because it could be applied to all `git` calls at once if we created a helper method for that or something)
- Override `HOME` in the tests so Git can't find a user's own configuration
- Set `PATH` in the tests so that `git` calls are intercepted and GPG signing can be disabled
- Use pre-made Git commits for tests so users' configuration can't get in the way

I think my preference would be overriding `HOME`, but I'm aware that that could have other unintended consequences, so I'm curious what others think?

DomT4:
To be honest I thought this PR here was a good shout in terms of being less broad than my prior attempt. It's a hassle to remember to pass `--no-gpgsign` in every test but it does resolve the issue in a simple & effective manner whilst also being something unlikely to ever introduce any unexpected side effects.

alyssais:
My concern about that would be that it would be easy to forget when creating new tests if the author didn't use GPG – it wouldn't be caught by CI either. Might still be worth it as a quick fix though…

MikeMcQuaid:
> Override HOME in the tests so Git can't find a user's own configuration

This seems like the best solution by far.

DomT4:
We went close to that in https://github.com/Homebrew/brew/pull/332 but it was dropped in favour of env sanitisation. I can pick it back up or someone else can run with what I wrote at the time or go an entirely different route.

@alyssais expressed the interest & I don't want to step on that so, whatever they're happy with.

alyssais:
@DomT4 I'd be happy for you to have a go if you're up for it! Let me know if you need/want me to jump in on anything.

DomT4:
I'm not around much beyond Wednesday but I'll try and file something by the end of the weekend. If you don't see anything by the start of next week please feel free to sigh at me & step over my procrastination 🙈.

mistydemeo:
I tried out redefining `HOME`, but it looks like it means we potentially get some file leakage from, for example, `svn` writing files to `${TEST_TMPDIR}/.subversion`.

DomT4:
Our test-suite is considerably more comprehensive than when I first took a run at this, so it's possible there are new issues we hadn't encountered previously.

DomT4:
(One day I'll remember to not refer to Homebrew using possessive terms any more 😸)

PR #2198, opened 2017-02-26
emoji tests: make sure to unset install badge env vars

mistydemeo:

-----

Otherwise the user's specified emoji will be used and the test will fail. For example:

```
Randomized with seed 13929
.F

Failures:

  1) Emoji#install_badge returns 🍺 by default
     Failure/Error: expect(subject).to eq "🍺"

       expected: "🍺"
            got: "🎑"

       (compared using ==)
     # ./test/emoji_spec.rb:8:in `block (3 levels) in <top (required)>'
     # ./test/spec_helper.rb:61:in `block (2 levels) in <top (required)>'
     # ./vendor/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block (2 levels) in <top (required)>'
```

### DIFF
--- a/Library/Homebrew/test/emoji_spec.rb
+++ b/Library/Homebrew/test/emoji_spec.rb
@@ -4,6 +4,11 @@ describe Emoji do
   describe "#install_badge" do
     subject { described_class.install_badge }
 
+    before(:each) do
+      ENV.delete("HOMEBREW_NO_EMOJI")
+      ENV.delete("HOMEBREW_INSTALL_BADGE")
+    end
+
     it "returns 🍺 by default" do
       expect(subject).to eq "🍺"
     end
--- a/Library/Homebrew/test/emoji_spec.rb
+++ b/Library/Homebrew/test/emoji_spec.rb
@@ -9,7 +9,7 @@ describe Emoji do
       ENV.delete("HOMEBREW_INSTALL_BADGE")
     end
 
-    it "returns 🍺 by default" do
+    it "returns 🍺  by default" do
       expect(subject).to eq "🍺"
     end
 


Library/Homebrew/test/emoji_spec.rb
```diff
@@ -4,8 +4,15 @@
   describe "#install_badge" do
     subject { described_class.install_badge }
 
-    it "returns 🍺 by default" do
-      expect(subject).to eq "🍺"
+    context "with emoji environment variables unset" do
```
MikeMcQuaid: I think you can skip the `context` here but CC @reitermarkus @alyssais for RSpec thoughts. Maybe worth adding another test to verify that `HOMEBREW_NO_EMOJI` and `HOMEBREW_INSTALL_BADGE` work as expected too? Feel free to punt on that though.

reitermarkus: Yes, as this is the default, this doesn't really have a context. The `before` block should run before each test, setting this default.

Library/Homebrew/test/emoji_spec.rb
```diff
@@ -4,8 +4,15 @@
...
+      before(:each) do
+        ENV.delete("HOMEBREW_NO_EMOJI")
+        ENV.delete("HOMEBREW_INSTALL_BADGE")
+      end
+
+      it "returns 🍺 by default" do
```
reitermarkus: Also add two spaces after the emoji here, as the terminal seems to need two for an emoji.

Library/Homebrew/test/emoji_spec.rb
alyssais: Agree the context isn't necessary.

@MikeMcQuaid there is a test for `HOMEBREW_INSTALL_BADGE` already, but not `HOMEBREW_NO_EMOJI`, I think because this method shouldn't even be called when that's set.

Library/Homebrew/test/emoji_spec.rb
mistydemeo: Good idea, changed.

Library/Homebrew/test/emoji_spec.rb
mistydemeo: Removed the context.

Library/Homebrew/test/emoji_spec.rb
```diff
@@ -3,8 +3,12 @@
 describe Emoji do
   describe "#install_badge" do
     subject { described_class.install_badge }
+    before(:each) do
```
reitermarkus: One minor nit: Add a newline before `before`.

mistydemeo: Done!

Library/Homebrew/test/emoji_spec.rb
```diff
@@ -4,7 +4,12 @@
   describe "#install_badge" do
     subject { described_class.install_badge }
 
-    it "returns 🍺 by default" do
+    before(:each) do
+      ENV.delete("HOMEBREW_NO_EMOJI")
+      ENV.delete("HOMEBREW_INSTALL_BADGE")
+    end
+
+    it "returns 🍺  by default" do
```
MikeMcQuaid: Pedantry: looks like an added space here 😭. 👍 to merge after that.

reitermarkus: You probably missed my comment above.

PR #2199, opened 2017-02-26
xcodebuild audit: match xcodebuild with no args

mistydemeo:

-----

The old regex only matched `xcodebuild` if passed arguments, but we want to pass `SYMROOT` in all circumstances. For example, without this patch, `brew audit eject` returns no problems, but with this patch it reports `* xcodebuild should be passed an explicit "SYMROOT"`.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -924,7 +924,7 @@ class FormulaAuditor
       end
     end
 
-    if text =~ /xcodebuild[ (]["'*]/ && !text.include?("SYMROOT=")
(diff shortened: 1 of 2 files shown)


mistydemeo:
@BrewTestBot test this please

MikeMcQuaid:
Nice work @mistydemeo!

fxcoudert:
This created two false positives in `homebrew-core`:

```
xctool:
  * xcodebuild should be passed an explicit "SYMROOT"
qt@5.5:
  * xcodebuild should be passed an explicit "SYMROOT"
```

In these formulas `xcodebuild` is mentioned in the `desc` or a comment.

Issue #2184, opened 2017-02-25
Ruby 2.4.0 tests issue.


DomT4:
# Please follow the general troubleshooting steps first:

- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

`brew tests` run against a `bundler` installed with `ruby` 2.4.0 throws this error:

```
......................dyld: lazy symbol binding failed: Symbol not found: _rb_str_new_static
  Referenced from: /usr/local/Homebrew/Library/Homebrew/vendor/bundle/gems/json-2.0.3/lib/json/ext/generator.bundle
  Expected in: flat namespace

dyld: Symbol not found: _rb_str_new_static
  Referenced from: /usr/local/Homebrew/Library/Homebrew/vendor/bundle/gems/json-2.0.3/lib/json/ext/generator.bundle
  Expected in: flat namespace
```

Not really sure what to suggest, barring forcing the system gem or something. Reproduction instructions at least for me:

```
rm -rf ~/.bundle
rm -rf /usr/local/Homebrew/Library/Homebrew/vendor/bundle
rbenv install 2.4.0
rbenv global 2.4.0
gem install bundler
brew tests
```

Tested with a clean Homebrew installation. Flipping back to the system version of Ruby works fine.

MikeMcQuaid:
@reitermarkus @alyssais Looks like we shouldn't be sharing these gems between Ruby versions, right?

alyssais:
This is likely [my fault](https://github.com/Homebrew/brew/pull/1028).

MikeMcQuaid:
@alyssais Thanks for figuring that out. I think hardcoding the directory somewhere under e.g. vendor for everyone would be good (even if that means a bit of duplication).

alyssais:
@MikeMcQuaid that sounds like the right approach to me. 👍

DomT4:
CC @reitermarkus re https://github.com/Homebrew/brew/pull/2203#issuecomment-282719357. Not sure there's a lot you can do about it still, but, that's the origin of the problem.

reitermarkus:
[...]

PR #2203, opened 2017-02-27
brew tests: filter out vendor/

mistydemeo:

-----

Not sure why this doesn't consistently repro for me, but I noticed that `brew tests` was failing for me because the test glob was picking up test files from the bundled gems installed in `test/vendor/bundle`.

### DIFF
--- a/Library/Homebrew/dev-cmd/tests.rb
+++ b/Library/Homebrew/dev-cmd/tests.rb
@@ -57,6 +57,7 @@ module Homebrew
       ENV["SEED"] = ARGV.next if ARGV.include? "--seed"
 
       files = Dir.glob("test/**/*_{spec,test}.rb")
+                 .reject { |p| p =~ %r{^test/vendor/bundle} }
                  .reject { |p| !OS.mac? && p =~ %r{^test/(os/mac|cask)(/.*|_(test|spec)\.rb)$} }
 
       test_args = []
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@
 !/CONTRIBUTING.md
 !/LICENSE.txt
 !/README.md
+
+# Unignore tests' bundle config
+!/Library/Homebrew/test/.bundle
--- a/Library/Homebrew/test/.bundle/config
+++ b/Library/Homebrew/test/.bundle/config
@@ -1,0 +1,3 @@
+---
+BUNDLE_PATH: "../vendor/bundle"
+BUNDLE_DISABLE_SHARED_GEMS: '1'


reitermarkus:
Remove `test/vendor` and `test/.bundle`. The bundled Gems have been in `vendor/bundle` for a few months now, but it seems that the `test/.bundle/config` still contains the `BUNDLE_PATH` for some people.

Alternatively, we could add `test/.bundle/config` to version control, to make sure it contains the same for everyone.

mistydemeo:
Adding it to version control or explicitly running `bundle config` to set the path makes sense to me. But I still think we should exclude these files, because even if the vendor bundle gets installed in the right place, this will fail if a user happens to have this old directory left around.

MikeMcQuaid:
> We could add test/.bundle/config to version control, to make sure it contains the same for everyone.

Makes sense to me 👍

MikeMcQuaid:
@DomT4 Does this fix https://github.com/Homebrew/brew/issues/2184 for you?

DomT4:
@MikeMcQuaid No, same issue. To be honest I initially pinned this yesterday on an upstream problem at some level, because `json` have had identical issues reported to them in the past with semi-new Ruby x.Y.x releases. I noted it [upstream](https://github.com/flori/json/issues/229#issuecomment-282494403) on the long-running issue but nobody else has reported it as far as I can see, so I'm not sure what's going on entirely.

DomT4:
If I `gem install json -V` outside of Bundle masking the output there's a noted failure on an element, but it's not the element I was seeing blow up `tests`.

```
linking shared-object json/ext/parser.bundle
current directory: /usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/json-2.0.3/ext/json/ext/parser
make "DESTDIR=" install
/usr/local/bin/ginstall -c -m 0755 parser.bundle ./.gem.20170227-65081-2nbc27/json/ext
To see why this extension failed to compile, please check the mkmf.log which can be found here:

  /usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/extensions/x86_64-darwin-16/2.4.0-static/json-2.0.3/mkmf.log
```
...
```
"clang -o conftest -I/usr/local/var/rbenv/versions/2.4.0/include/ruby-2.4.0/x86_64-darwin16 -I/usr/local/var/rbenv/versions/2.4.0/include/ruby-2.4.0/ruby/backward -I/usr/local/var/rbenv/versions/2.4.0/include/ruby-2.4.0 -I. -I/usr/local/var/rbenv/versions/2.4.0/include  -D_XOPEN_SOURCE -D_DARWIN_C_SOURCE -D_DARWIN_UNLIMITED_SELECT -D_REENTRANT    -O3 -Wno-error=shorten-64-to-32  -pipe conftest.c  -L. -L/usr/local/var/rbenv/versions/2.4.0/lib -L. -L/usr/local/var/rbenv/versions/2.4.0/lib  -fstack-protector -L/usr/local/lib     -lruby-static -framework CoreFoundation  -lpthread -lgmp -ldl -lobjc "
conftest.c:15:57: error: use of undeclared identifier 'rb_enc_raise'
int t(void) { void ((*volatile p)()); p = (void ((*)()))rb_enc_raise; return !p; }
                                                        ^
1 error generated.
```

ilovezfs:
@DomT4 does it happen if you checkout a tag? 1.1.10, 1.1.9, 1.1.8, ... wondering if this is a bisectable problem.

DomT4:
`json 1.8.6` builds without errors on the `gem install` at least. None of 2.0.0, 2.0.1, 2.0.2 or the latest 2.0.3 compile cleanly. On the brew side, no error thrown at 9cd5a21b473f0271b162bbe7f77f7d1468c0cfa1, 619791e83d2781dca1b675e20249a8aebe085c7c, 9cce3414fbd39d146d97911dbb3bf07fffaa1080, e59ada508727f11464595893783beb914c26f60b, 664d0c67d5947605c914c4c56ebcfaa80cb6eca0.

Had appeared by the time 69d1cedd12d37d6a9535fbba17d2129b165c08f8 was cut.

ilovezfs:
OK. So you should be able to do
```
git checkout master
git bisect start
git checkout 664d0c6
git bisect good
git checkout 69d1ced
git bisect bad
```
I then tell it `git bisect good` or `git bisect bad` for each commit it asks you to test until it spits out the first bad commit.

DomT4:
Was already working on it 😉. e40c63f77c83254970c0e98f69ea76685536a905 is where it was introduced, merged as part of https://github.com/Homebrew/brew/pull/1913.

ilovezfs:
:tada:

Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -57,6 +57,7 @@ def tests
       ENV["SEED"] = ARGV.next if ARGV.include? "--seed"
 
       files = Dir.glob("test/**/*_{spec,test}.rb")
+                 .reject { |p| p =~ %r{^test/vendor} }
```
reitermarkus: Make this `test/vendor/bundle`, since there is at least one test actually testing vendored code (`plist`, currently still in `cask-tests`).

mistydemeo: Done!

DomT4:
Was it intentional here that the repo becomes "dirty" after `brew tests` is run?

```
cd $(brew --repo)
git fetch && git reset --hard origin/master
brew tests
Homebrew 1.1.10-432-ge02223960-dirty
```
The change is:
```diff
diff --git a/Library/Homebrew/test/.bundle/config b/Library/Homebrew/test/.bundle/config
index dcf3d1c4c..e451829e9 100644
--- a/Library/Homebrew/test/.bundle/config
+++ b/Library/Homebrew/test/.bundle/config
@@ -1,3 +1,3 @@
 ---
 BUNDLE_PATH: "../vendor/bundle"
-BUNDLE_DISABLE_SHARED_GEMS: '1'
+BUNDLE_DISABLE_SHARED_GEMS: "true"
```

PR #2211, opened 2017-02-27
Revert "Revert "cc: always filter flags on deps.""

MikeMcQuaid:
https://github.com/Homebrew/homebrew-core/issues/8528 was addressed with https://github.com/Homebrew/brew/pull/1815 so that can be Homebrew/brew#1796 can be reverted (unreverting https://github.com/Homebrew/brew/pull/1746).

Reverts Homebrew/brew#1796

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -206,10 +206,6 @@ class Cmd
   end
 
   def keep?(path)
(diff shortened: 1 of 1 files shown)


ilovezfs:
PR to re-revert https://github.com/Homebrew/brew/pull/2302.

PR #2233, opened 2017-02-28
Remove Minitest.

reitermarkus:

-----

After *only* 12504 lines of code, all tests are now converted. 🎉


Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -1,4 +1,4 @@
-#:  * `tests` [`-v`] [`--coverage`] [`--generic`] [`--no-compat`] [`--only=`<test_script>`:`<test_method>] [`--seed` <seed>] [`--trace`] [`--online`] [`--official-cmd-taps`]:
+#:  * `tests` [`-v`] [`--coverage`] [`--generic`] [`--no-compat`] [`--only=`<test_name>] [`--seed` <seed>] [`--online`] [`--official-cmd-taps`]:
```
MikeMcQuaid: No longer possible to run single tests?

reitermarkus: Doesn't seem possible when using `parallel_rspec`. Maybe @alyssais knows more?

A workaround would be to run default `rspec` when only a single test should be run.

MikeMcQuaid: > A workaround would be to run default rspec when only a single test should be run.

👍 from me.

alyssais: Wouldn't be surprised if that wasn't possible with parallel_rspec since it likes to treat files as being the lowest division of tests (though I haven't checked). Workaround is probably best.

PR #2234, opened 2017-02-28
Add brew LLVM libOMP/omp.h to header/lib search path

aw1621107:

-----

Finally a "fix" for the issue @ilovezfs found in #1130. At least as far as I could test, this works, but I wouldn't call it clean enough to merge. Sorry for the delay, for what it's worth.

I can't seem to run `brew tests`, either; tests fail with LoadErrors or NameErrors. I'll be looking into that in the meantime...

### DIFF
--- a/Library/Homebrew/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/ENV/super.rb
@@ -164,7 +164,11 @@ module Superenv
   end
 
   def determine_include_paths
-    keg_only_deps.map { |d| d.opt_include.to_s }.to_path_s
+    paths = keg_only_deps.map { |d| d.opt_include.to_s }
+    if compiler == :llvm_clang && Formula["llvm"].installed?
+      paths << "#{Formula["llvm"].opt_lib}/libomp"
+    end
+    paths.to_path_s
   end
 
   def homebrew_extra_library_paths
@@ -174,6 +178,9 @@ module Superenv
   def determine_library_paths
     paths = keg_only_deps.map { |d| d.opt_lib.to_s }
     paths << "#{HOMEBREW_PREFIX}/lib"
+    if compiler == :llvm_clang && Formula["llvm"].installed?
+      paths << "#{Formula["llvm"].opt_lib}/libomp"
+    end
     paths += homebrew_extra_library_paths
     paths.to_path_s
   end


Library/Homebrew/extend/ENV/super.rb
```diff
@@ -39,6 +39,14 @@ def setup_build_environment(formula = nil)
     super
     send(compiler)
 
+    if compiler == :llvm_clang && Formula["llvm"].installed?
+      FileUtils::mkdir "#{HOMEBREW_TEMP.to_s}/brew-llvm-openmp"
+      FileUtils::ln_sf "#{Formula["llvm"].lib}/clang/#{Formula["llvm"].version.to_s}/include/omp.h",
+                       "#{HOMEBREW_TEMP.to_s}/brew-llvm-openmp/omp.h"
+      FileUtils::ln_sf "#{Formula["llvm"].lib}/libomp.dylib",
+                       "#{HOMEBREW_TEMP.to_s}/brew-llvm-openmp/libomp.dylib"
```
MikeMcQuaid: Why can't `#{Formula["llvm"].lib}` be added to the link path and `#{Formula["llvm"].lib}/clang/#{Formula["llvm"].version.to_s}/include/` to the include path?

aw1621107: I was concerned that that would cause libc++ and friends to be pulled in instead of the system versions

ilovezfs:
It might be a good idea to decide what we're doing about the fallout from https://github.com/Homebrew/brew/pull/2036#issuecomment-283081542 before merging anything over here.

Library/Homebrew/extend/ENV/super.rb
aw1621107: Sounds like a good idea. I'll give it a shot.

aw1621107: Would something like this in the `install` section of `llvm.rb` work?

```
mkdir_p lib/"libomp"
  [libgomp.dylib libiomp5.dylib libomp.dylib].each {|libomp_dylib|
  (lib/"libomp").install_symlink lib/libomp_dylib
}
```

MikeMcQuaid: Yes.

aw1621107: Alright, done!

Library/Homebrew/extend/ENV/super.rb
```diff
@@ -164,7 +164,11 @@ def determine_isystem_paths
   end
 
   def determine_include_paths
-    keg_only_deps.map { |d| d.opt_include.to_s }.to_path_s
+    paths = keg_only_deps.map { |d| d.opt_include.to_s }
+    if compiler == :llvm_clang && Formula["llvm"].installed?
```
MikeMcQuaid: Why check if it is installed? If it's set as the compiler it's going to be used regardless, no? Also, should this be dependent on OpenMP being used? No worries if not, genuinely asking.

aw1621107: Good question; that would be poor reasoning on my part, so that will be removed.

I guess in theory it depends on OpenMP being used, but I wasn't sure how to check `needs :openmp` and I figured since the directory only contains omp.h and libomp.dylib there wouldn't be any harm in including it for a formula that doesn't use OpenMP.

MikeMcQuaid: I reckon this should be conditional to `needs :openmp` being used but CC @mistydemeo @ilovezfs for thoughts around this stuff

MikeMcQuaid:
@aw1621107 Sorry that I've lost track: what's the current state of this?

aw1621107:
Waiting on me to make another change to implement what ilovezfs suggested [here](https://github.com/Homebrew/homebrew-core/pull/10456#issuecomment-283830370)

aw1621107:
Still need to get to it... I'll try for this weekend

PR #2236, opened 2017-02-28
Move Formatter and Locale specs.

reitermarkus:


MikeMcQuaid:
Is this code shared between Homebrew and Homebrew Cask (and used in both)?

reitermarkus:
At least `Formatter` is shared, but why does this matter? I thought we wanted to have `brew tests` also run Cask tests?

MikeMcQuaid:
> At least Formatter is shared, but why does this matter? 

👍 on Formatter then.

> I thought we wanted to have brew tests also run Cask tests?

Yep, I think so but it should probably still live in e.g. `Library/Homebrew/test/cask/locale_spec.rb` if it's not shared just to make it more obvious.

reitermarkus:
Ok, seems like `Locale` is used for the `os/mac` test.

MikeMcQuaid:
> Yep, I think so but it should probably still live in e.g. Library/Homebrew/test/cask/locale_spec.rb if it's not shared just to make it more obvious.

Just to repeat myself: I'm cool with moving all these tests around but I think still maintaining a `test/cask` subdirectory probably makes sense (and it is better than `cask/test`). Thanks @reitermarkus !

reitermarkus:
> I think still maintaining a `test/cask` subdirectory probably makes sense

Yes, these were the only ones that weren't actually in `test/cask`.

PR #1873, opened 2017-01-18
Custom Cops for `brew audit`

GauthamGoli:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----
This PR implements Custom Cops for a simple rule in `brew audit` command as discussed in #569 

>  `revision` in a `bottle do` block should be renamed to `rebuild` and Rubocop should autocorrect it to `rebuild`

More rules can be added/migrated as cops, and RuboCop can be leveraged to check for style violations instead of regular expressions as being done currently when `brew audit` is ran.

Using `--strict` would check for CustomCop/style violations.
The newly added `--fix` option would automatically correct the source code if any CustomCop/style violations are found.

Issue #569, opened 2016-07-21
Use Rubocop for some `brew audit`s


MikeMcQuaid:
We should investigate using Rubocop for some (maybe all) `brew audit`s rather than just regex as Rubocop can actually parse Ruby.

maxnordlund:
I might be able to take a stab at this, seems doable. But I would like some pointers though. May I please have some? (Also maybe you could ping the right persons @MikeMcQuaid)

jawshooah:
Take a look at [rubocop-cask](https://github.com/caskroom/rubocop-cask) as an example of custom cops and parsing logic.

maxnordlund:
Nice, will do 👍 . But should I then point a PR to that repo or put inside here somewhere (and then what is a good place)?

maxnordlund:
Or make a new repo, but that seems like overkill.

MikeMcQuaid:
@maxnordlund That would be so great! I agree that probably having it inside this repo would be ideal, if that's possible (similarly maybe with rubocop-cask one day). @jawshooah any thoughts on why you made it a separate repo to avoid hitting any issues you may have hit?

Regardless, I wouldn't worry too much about location or even completeness for now; I'd pick a single rule from `brew audit` that looks like a good candidate for Rubocop and try and get the flow working end-to-end and then we can look at porting other rules across.

maxnordlund:
Sounds like a plan, I'll make a new repo and steal/borrow the cask rubocop structure and do as you say and implement a single rule. We can always move the code around if needed I guess.

MikeMcQuaid:
A suggestion for an easy single rule that requires understanding Ruby more than `audit` really can currently: `revision` in a `bottle do` block should be renamed to `rebuild` and Rubocop should autocorrect it to `rebuild`.

maxnordlund:
I haven't been able to get this rolling because a lot of stuff have cropped up at work. I still will try to do this, but it may take quite some time.

MikeMcQuaid:
@maxnordlund No worries or rush.

GauthamGoli:
@MikeMcQuaid I could work on implementing this feature. I implemented a basic cop that you mentioned above https://github.com/GauthamGoli/rubocop-brew

> `revision` in a `bottle do` block should be renamed to `rebuild` and Rubocop should autocorrect it to `rebuild`

I tested it on a formula and its working fine. Here is the link to the code. I'm still trying to understand how to write custom cops, and also am a ruby novice.

[...]

PR #2242, opened 2017-03-01
Port audit_desc audit rules to a cop

GauthamGoli:

-----
This is in continuation of the PR #1873 and issue #569 
This PR ports `audit_desc` audit rules into a rubocop custom cop. The rules are 

- Check if `desc` exists in a Formula
- Check if `desc` size is > 80
- Check if `desc` begins with an article
- Check for proper usage of `command-line` in `desc`
- Check if `desc` contains the formula name

The previous tests for the ported audit rules in `audit_spec.rb` have been removed. ~~I am working on adding tests for the Custom Cops. I have the basic code, but I'm currently trying to figure out where to place the code for the tests to get executed.~~
New tests for the RuboCop custom cops have been added and `Gemfile` has been updated.

### DIFF
--- a/Library/.rubocop.yml
+++ b/Library/.rubocop.yml
@@ -8,17 +8,10 @@ AllCops:
 
 require: ./Homebrew/rubocops.rb
 
-Homebrew/FormulaCop:
-  Enabled: false
-
 Homebrew/CorrectBottleBlock:
-  Include:
-    - '**/Taps/homebrew/**/*.rb'
   Enabled: true
 
 Homebrew/FormulaDesc:
-  Include:
-    - '**/Taps/homebrew/**/*.rb'
   Enabled: true
 
 Metrics/AbcSize:
--- a/Library/.rubocop.yml
+++ b/Library/.rubocop.yml
@@ -8,7 +8,17 @@ AllCops:
 
 require: ./Homebrew/rubocops.rb
 
+Homebrew/FormulaCop:
+  Enabled: false
+
 Homebrew/CorrectBottleBlock:
+  Include:
+    - '**/Taps/homebrew/**/*.rb'
+  Enabled: true
+
+Homebrew/FormulaDesc:
+  Include:
+    - '**/Taps/homebrew/**/*.rb'
   Enabled: true
 
 Metrics/AbcSize:
--- a/Library/Homebrew/rubocops/bottle_block_cop.rb
+++ b/Library/Homebrew/rubocops/bottle_block_cop.rb
@@ -11,34 +11,18 @@ module RuboCop
         MSG = "Use rebuild instead of revision in bottle block".freeze
 
         def audit_formula(_node, _class_node, _parent_class_node, formula_class_body_node)
-          check(formula_class_body_node)
+          bottle = find_block(formula_class_body_node, :bottle)
+          return if bottle.nil? || block_size(bottle).zero?
+          problem "Use rebuild instead of revision in bottle block" if method_called?(bottle, :revision)
         end
 
         private
-
-        def check(formula_class_body_node)
-          formula_class_body_node.each_child_node(:block) do |block_node|
-            next if block_length(block_node).zero?
-            method, _args, block_body = *block_node
-            _keyword, method_name = *method
-            next unless method_name == :bottle
-            check_revision?(block_body)
-          end
-        end
 
         def autocorrect(node)
           lambda do |corrector|
             correction = node.source.sub("revision", "rebuild")
             corrector.insert_before(node.source_range, correction)
             corrector.remove(node.source_range)
-          end
-        end
-
-        def check_revision?(body)
-          body.children.each do |method_call_node|
-            _receiver, method_name, _args = *method_call_node
-            next unless method_name == :revision
-            add_offense(method_call_node, :expression)
           end
         end
       end
--- a/Library/Homebrew/rubocops/bottle_block_cop.rb
+++ b/Library/Homebrew/rubocops/bottle_block_cop.rb
@@ -1,19 +1,30 @@
+require_relative "./extend/formula_cop"
+
 module RuboCop
   module Cop
     module Homebrew
-      class CorrectBottleBlock < Cop
+      # This cop audits `bottle` block in Formulae
+      #
+      # - `rebuild` should be used instead of `revision` in `bottle` block
+
+      class CorrectBottleBlock < FormulaCop
         MSG = "Use rebuild instead of revision in bottle block".freeze
 
-        def on_block(node)
-          return if block_length(node).zero?
-          method, _args, body = *node
-          _keyword, method_name = *method
-
-          return unless method_name == :bottle
-          check_revision?(body)
+        def audit_formula(_node, _class_node, _parent_class_node, formula_class_body_node)
+          check(formula_class_body_node)
         end
 
         private
+
+        def check(formula_class_body_node)
+          formula_class_body_node.each_child_node(:block) do |block_node|
+            next if block_length(block_node).zero?
+            method, _args, block_body = *block_node
+            _keyword, method_name = *method
+            next unless method_name == :bottle
+            check_revision?(block_body)
+          end
+        end
 
         def autocorrect(node)
           lambda do |corrector|
--- a/Library/Homebrew/rubocops/extend/formula_cop.rb
+++ b/Library/Homebrew/rubocops/extend/formula_cop.rb
@@ -1,0 +1,22 @@
+module RuboCop
+  module Cop
+    module Homebrew
+      class FormulaCop < Cop
+        @registry = Cop.registry
+
+        def on_class(node)
+          # This method is called by RuboCop and is the main entry point
+          class_node, parent_class_node, body = *node
+          return unless a_formula_class?(parent_class_node)
+          audit_formula(node, class_node, parent_class_node, body)
+        end
+
+        private
+
+        def a_formula_class?(parent_class_node)
+          parent_class_node && parent_class_node.const_name == "Formula"
+        end
+      end
+    end
+  end
+end
--- a/Library/Homebrew/rubocops/extend/formula_cop.rb
+++ b/Library/Homebrew/rubocops/extend/formula_cop.rb
@@ -6,15 +6,137 @@ module RuboCop
 
         def on_class(node)
           # This method is called by RuboCop and is the main entry point
+          file_path = processed_source.buffer.name
+          return unless file_path_allowed?(file_path)
           class_node, parent_class_node, body = *node
-          return unless a_formula_class?(parent_class_node)
+          return unless formula_class?(parent_class_node)
+          return unless respond_to?(:audit_formula)
+          @formula_name = class_name(class_node)
           audit_formula(node, class_node, parent_class_node, body)
+        end
+
+        def regex_match_group(node, pattern)
+          # Checks for regex match of pattern in the node and
+          # Sets the appropriate instance variables to report the match
+          string_repr = string_content(node)
+          match_object = string_repr.match(pattern)
+          return unless match_object
+          node_begin_pos = start_column(node)
+          line_begin_pos = line_start_column(node)
+          @column = node_begin_pos + match_object.begin(0) - line_begin_pos + 1
+          @length = match_object.to_s.length
+          @line_no = line_number(node)
+          @source_buf = source_buffer(node)
+          @offense_source_range = source_range(@source_buf, @line_no, @column, @length)
+          @offensive_node = node
+          match_object
+        end
+
+        def find_node_method_by_name(node, method_name)
+          # Returns method_node matching method_name
+          return if node.nil?
+          node.each_child_node(:send) do |method_node|
+            next unless method_node.method_name == method_name
+            @offensive_node = method_node
+            @offense_source_range = method_node.source_range
+            return method_node
+          end
+          # If not found then, parent node becomes the offensive node
+          @offensive_node = node.parent
+          @offense_source_range = node.parent.source_range
+          nil
+        end
+
+        def find_block(node, block_name)
+          # Returns a block named block_name inside node
+          return if node.nil?
+          node.each_child_node(:block) do |block_node|
+            next if block_node.method_name != block_name
+            @offensive_node = block_node
+            @offense_source_range = block_node.source_range
+            return block_node
+          end
+          # If not found then, parent node becomes the offensive node
+          @offensive_node = node.parent
+          @offense_source_range = node.parent.source_range
+          nil
+        end
+
+        def method_called?(node, method_name)
+          # Check if a method is called inside a block
+          block_body = node.children[2]
+          block_body.each_child_node(:send) do |call_node|
+            next unless call_node.method_name == method_name
+            @offensive_node = call_node
+            @offense_source_range = call_node.source_range
+            return true
+          end
+          false
+        end
+
+        def parameters(method_node)
+          # Returns the array of arguments of the method_node
+          return unless method_node.send_type?
+          method_node.method_args
+        end
+
+        def line_start_column(node)
+          # Returns the begin position of the node's line in source code
+          node.source_range.source_buffer.line_range(node.loc.line).begin_pos
+        end
+
+        def start_column(node)
+          # Returns the begin position of the node in source code
+          node.source_range.begin_pos
+        end
+
+        def line_number(node)
+          # Returns the line number of the node
+          node.loc.line
+        end
+
+        def class_name(node)
+          # Returns the class node's name, nil if not a class node
+          @offensive_node = node
+          @offense_source_range = node.source_range
+          node.const_name
+        end
+
+        def size(node)
+          # Returns the node size in the source code
+          node.source_range.size
+        end
+
+        def block_size(block)
+          # Returns the block length of the block node
+          block_length(block)
+        end
+
+        def source_buffer(node)
+          # Source buffer is required as an argument to report style violations
+          node.source_range.source_buffer
+        end
+
+        def string_content(node)
+          # Returns the string representation if node is of type str
+          node.str_content if node.type == :str
+        end
+
+        def problem(msg)
+          add_offense(@offensive_node, @offense_source_range, msg)
         end
 
         private
 
-        def a_formula_class?(parent_class_node)
+        def formula_class?(parent_class_node)
           parent_class_node && parent_class_node.const_name == "Formula"
+        end
+
+        def file_path_allowed?(file_path)
+          paths_to_exclude = [%r{/Library/Homebrew/compat/},
+                              %r{/Library/Homebrew/test/}]
+          return true if file_path.nil? # file_path is nil when source is directly passed to the cop eg., in specs
+          file_path !~ Regexp.union(paths_to_exclude)
         end
       end
     end
--- a/Library/Homebrew/rubocops/formula_desc_cop.rb
+++ b/Library/Homebrew/rubocops/formula_desc_cop.rb
@@ -1,0 +1,67 @@
+module RuboCop
+  module Cop
+    module Homebrew
+      class FormulaDesc < Cop
+        def on_class(node)
+          class_node, parent_class_node, body = *node
+          formula_name = class_node.const_name
+          return unless parent_class_node && parent_class_node.const_name == "Formula" && body
+          check(node, body, formula_name)
+        end
+
+        private
+
+        def check(node, body, formula_name)
+          body.each_child_node(:send) do |call_node|
+            _receiver, call_name, args = *call_node
+            next unless call_name == :desc && !args.children[0].empty?
+            description = args.children[0]
+
+            source_buffer = call_node.source_range.source_buffer
+            line_number = call_node.loc.line
+            line_begin_pos = call_node.source_range.source_buffer.line_range(call_node.loc.line).begin_pos
+            desc_begin_pos = call_node.children[2].source_range.begin_pos
+
+            linelength = "#{formula_name}: #{description}".length
+            if linelength > 80
+              column = desc_begin_pos - line_begin_pos
+              length = call_node.children[2].source_range.size
+              sourcerange = source_range(source_buffer, line_number, column, length)
+              message = <<-EOS.strip_indent
+                        Description is too long. \"name: desc\" should be less than 80 characters.
+                        Length is calculated as #{formula_name} + desc. (currently #{linelength})
+                        EOS
+              add_offense(call_node, sourcerange, message)
+            end
+
+            match_object = description.match(/([Cc]ommand ?line)/)
+            if match_object
+              column = desc_begin_pos+match_object.begin(0)-line_begin_pos+1
+              length = match_object.to_s.length
+              sourcerange = source_range(source_buffer, line_number, column, length)
+              add_offense(call_node, sourcerange, "Description should use \"command-line\" instead of \"#{match_object}\"")
+            end
+
+            match_object = description.match(/^([Aa]n?)\s/)
+            if match_object
+              column = desc_begin_pos+match_object.begin(0)-line_begin_pos+1
+              length = match_object.to_s.length
+              sourcerange = source_range(source_buffer, line_number, column, length)
+              add_offense(call_node, sourcerange, "Description shouldn't start with an indefinite article (#{match_object})")
+            end
+
+            match_object = description.match(/^#{formula_name}/i)
+            if match_object
+              column = desc_begin_pos+match_object.begin(0)-line_begin_pos+1
+              length = match_object.to_s.length
+              sourcerange = source_range(source_buffer, line_number, column, length)
+              add_offense(call_node, sourcerange, "Description shouldn't include the formula name")
+            end
+            return
+          end
+          add_offense(node, node.source_range, "Formula should have a desc (Description).")
+        end
+      end
+    end
+  end
+end
--- a/Library/Homebrew/rubocops/formula_desc_cop.rb
+++ b/Library/Homebrew/rubocops/formula_desc_cop.rb
@@ -13,6 +13,9 @@ module RuboCop
       # - Checks if `desc` contains the formula name
 
       class FormulaDesc < FormulaCop
+        attr_accessor :formula_name, :description, :source_buffer, :line_number, :line_begin_pos,
+                      :desc_begin_pos, :call_node
+
         def audit_formula(node, class_node, _parent_class_node, body)
           check(node, body, class_node.const_name)
         end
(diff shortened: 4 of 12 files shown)


GauthamGoli:
`brew style` without any args runs style checks on whole Homebrew `Library` which is why the build is failing as custom cops are running on files outside `Library/Taps/homebrew/`.

Instead we want Custom Cops to be run only on formulae files. Can I modify `style.rb` to achieve this? (To exclude custom cop checks on core code and include them on formulae)

reitermarkus:
`rubocop-cask` is checking checking if the file is a Cask,

https://github.com/caskroom/rubocop-cask/blob/eca041681dc554996a926173ec1d0c378efaecd3/lib/rubocop/cask/extend/node.rb#L14-L20


and skips it if it isn't.

https://github.com/caskroom/rubocop-cask/blob/cd020379b3a38d4dc20a23eb6c596b5f39414565/lib/rubocop/cop/cask/mixin/cask_help.rb#L9

GauthamGoli:
@reitermarkus I too added a check for the same https://github.com/Homebrew/brew/pull/2242/commits/19f693d25ba514a6b604128474763e83de55b548#diff-166fafc45d1d65494760d2fb578d6e46R8

But the problem is there are formulae defined outside `Library/Taps/homebrew/` https://travis-ci.org/Homebrew/brew/jobs/206899446#L266-L279

reitermarkus:
> I too added a check for the same

Ok, so there is a `on_class` method, do you know if there is a `on_file` method? This way we could simply skip all files under `Library/Homebrew`.

reitermarkus:
This looks like it could do what we want: https://github.com/bbatsov/rubocop/blob/9754e06f468631e9145b72f30e23e4cc0883bdb8/lib/rubocop/cop/style/file_name.rb#L20-L27

GauthamGoli:
@reitermarkus Yes, just found that too. Thanks! :)

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+  module Cop
+    module Homebrew
+      class FormulaDesc < Cop
+        def on_class(node)
+          return unless processed_source.path =~ %r{#{ENV["HOMEBREW_LIBRARY"]}\/Taps\/homebrew}
+          class_node, parent_class_node, body = *node
+          formula_name = class_node.const_name
+          return unless parent_class_node && parent_class_node.const_name == "Formula" && body
+          check(node, body, formula_name)
+        end
```
reitermarkus: Can you move this to a helper method? We'll probably need this for every Formula cop.

Library/Homebrew/rubocops/mixins/on_formula_class.rb
```diff
@@ -0,0 +1,18 @@
+module RuboCop
+  module Cop
+    # Helper module for checking formula class
+    module FormulaDef
```
MikeMcQuaid: To me it makes more sense for there to be a `HomebrewCop` base class that other cops inherit from as all will want to use this code.

Library/Homebrew/rubocops/mixins/on_formula_class.rb
```diff
@@ -0,0 +1,18 @@
...
+      def file_inside_formula_root?
+        return false unless processed_source.path =~ %r{#{ENV["HOMEBREW_LIBRARY"]}\/Taps\/homebrew}
```
MikeMcQuaid: Don't think `HOMEBREW_LIBRARY` will be set when using `rubocop` outside of `brew style`?

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,76 @@
...
+            line_begin_pos = call_node.source_range.source_buffer.line_range(call_node.loc.line).begin_pos
+            desc_begin_pos = call_node.children[2].source_range.begin_pos
+
+            linelength = "#{formula_name}: #{description}".length
+            if linelength > 80
+              column = desc_begin_pos - line_begin_pos
+              length = call_node.children[2].source_range.size
+              sourcerange = source_range(source_buffer, line_number, column, length)
+              message = <<-EOS.strip_indent
+                        Description is too long. \"name: desc\" should be less than 80 characters.
```
MikeMcQuaid: Indent this two characters after `message` and use `EOS.undent`. Are you sure you need to escape `\"`?

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,76 @@
...
+        def on_formula(node, class_node, _parent_class_node, body)
+          check(node, body, class_node.const_name)
+        end
+
+        private
+
+        def check(node, body, formula_name)
+          body.each_child_node(:send) do |call_node|
+            _receiver, call_name, args = *call_node
+            next unless call_name == :desc && !args.children[0].empty?
```
MikeMcQuaid: `unless &&` is generally hard to parse; can you flip this to an `if`?

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,76 @@
...
+              length = call_node.children[2].source_range.size
+              sourcerange = source_range(source_buffer, line_number, column, length)
+              message = <<-EOS.strip_indent
+                        Description is too long. \"name: desc\" should be less than 80 characters.
+                        Length is calculated as #{formula_name} + desc. (currently #{linelength})
+                        EOS
+              add_offense(call_node, sourcerange, message)
+            end
+
+            match_object = description.match(/([Cc]ommand ?line)/)
```
MikeMcQuaid: Could make this regex case-insensitive?

Library/Homebrew/rubocops/mixins/on_formula_class.rb
GauthamGoli: Yeah, this makes more sense, but I was not able to inherit `HomebrewCop` due to some error. Found a fix. Will push the changes.

Library/Homebrew/rubocops/mixins/on_formula_class.rb
GauthamGoli: Should I use `%r{\/Taps\/homebrew}` without `HOMEBREW_LIBRARY` ?

Library/Homebrew/rubocops/extend/Homebrewcop.rb
```diff
@@ -0,0 +1,19 @@
+module RuboCop
+  module Cop
+    class HomebrewCop < Cop
```
reitermarkus: I'd call this `FormulaCop`. Also, use `snake_case` for the filename.

GauthamGoli: Thanks.

Library/Homebrew/rubocops/mixins/on_formula_class.rb
MikeMcQuaid: Yeh, that sounds like a sensible default. Don't need `\/` inside a `%r`, though!

Library/Homebrew/rubocops/bottle_block_cop.rb
```diff
@@ -1,20 +1,31 @@
...
 
-          return unless method_name == :bottle
-          check_revision?(body)
+        def on_formula(_node, _class_node, _parent_class_node, body)
+          check(body)
         end
 
         private
 
+        def check(body)
```
MikeMcQuaid: Not totally clear to me what `body` is here; could you explain so we can maybe pick a better name?

Library/Homebrew/rubocops/bottle_block_cop.rb
```diff
@@ -1,20 +1,31 @@
...
+          body.each_child_node(:block) do |block_node|
```
MikeMcQuaid: Not totally clear to me what `block_node ` is here; could you explain so we can maybe pick a better name?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,19 @@
+module RuboCop
+  module Cop
+    class FormulaCop < Cop
+      @registry = Cop.registry
+
+      def file_inside_formula_root?
+        return false unless processed_source.path =~ %r{/Taps/homebrew}
+        true
```
MikeMcQuaid: `!(processed_source.path =~ %r{/Taps/homebrew}).nil?` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+
+      def on_class(node)
```
MikeMcQuaid: Not sure what `node` is here; could you explain so we can figure out a better name?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,19 @@
...
+        return unless file_inside_formula_root?
```
MikeMcQuaid: I wonder if this could be done in a shared initializer?

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,75 @@
...
+      # This cop audits `desc` in Formulae
+      #
+      # - Checks for existence of `desc`
+      # - Checks if size of `desc` > 80
+      # - Checks if `desc` begins with an article
+      # - Checks for correct usage of `command-line` in `desc`
+      # - Checks if `desc` contains the formula name
+
+      class FormulaDesc < FormulaCop
+        def on_formula(node, class_node, _parent_class_node, body)
```
MikeMcQuaid: Not super clear what `on_formula` means or the various `node`s?

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,75 @@
...
+          check(node, body, class_node.const_name)
+        end
+
+        private
+
+        def check(node, body, formula_name)
```
MikeMcQuaid: Would be good to figure out a way to extract more of the Rubocop/Ruby parse tree logic here from the audit logic.

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,75 @@
...
+              EOS
+              add_offense(call_node, sourcerange, message)
+            end
+
+            match_object = description.match(/(command ?line)/i)
+            if match_object
+              column = desc_begin_pos+match_object.begin(0)-line_begin_pos+1
+              length = match_object.to_s.length
+              sourcerange = source_range(source_buffer, line_number, column, length)
+              add_offense(call_node, sourcerange, "Description should use \"command-line\" instead of \"#{match_object}\"")
```
MikeMcQuaid: A linebreak here would be good to make this a little shorter

Library/Homebrew/rubocops/formula_desc_cop.rb
GauthamGoli: Inside `FormulaCop` there's a method `on_class` which does the following sequentially

-  Checks if the file being processed is inside /Taps/Homebrew/
-  If so, check if current class is being inherited from `Formula` ( Thus making sure that its a Homebrew Formula)
-  Call `on_formula` with arguments below. (`on_formula` is not yet defined as we will define it inside the CustomCop which inherits from `FormulaCop`)
-  `node` is the RuboCop::AST::node instance of whole formula
-  `class_node` is the RuboCop::AST::node instance of formula's name
-  `_parent_class_node` is the RuboCop::AST::node instance of `Formula` 
-  `body` is the RubCop::AST::node for ever thing thats inside the Formula

`class_node`, `_parent_class_node`, `body` are children of `node` and may be we can omit passing `node` as the argument to `on_formula` as its redundant.

Library/Homebrew/rubocops/extend/formula_cop.rb
GauthamGoli: RuboCop processes the source code and creates an Abstract Syntax Tree(AST) representation, and traverses the AST. As it encounters a `node` of some `type`, a message is sent to cop instance class hierarchy `send(:"on_#{node.type}", node)`
and we override it by defining `on_class` to check and add violations.

Library/Homebrew/rubocops/extend/formula_cop.rb
GauthamGoli: Yes, This is more elegant 👍

Library/Homebrew/rubocops/bottle_block_cop.rb
GauthamGoli: `body.each_child_node(:block)` calls the given block for each `child` node which has `child.type==:block`. (We want to iterate only over blocks and find `bottle` block). 
The given block for `body.each_child_node(:block)` takes `block_node` as the argument.

Library/Homebrew/rubocops/bottle_block_cop.rb
GauthamGoli: `body` is the `RuboCop::AST::Node` instance of everything that's inside the formula's class.
The child nodes of `body` are also instances of `RuboCop::AST::Node`

Library/Homebrew/rubocops/extend/formula_cop.rb
GauthamGoli: Can you please explain what's a shared initializer and why would we want to use it?

Library/Homebrew/rubocops/formula_desc_cop.rb
MikeMcQuaid: I'm wondering if it's worth being called `audit_formula`; the `on_` naming scheme is better if people are familiar with RuboCop but I'd think the majority won't be.

Library/Homebrew/rubocops/formula_desc_cop.rb
MikeMcQuaid: Perhaps by generally splitting this into a bunch of smaller methods.

Library/Homebrew/rubocops/extend/formula_cop.rb
reitermarkus: This will exclude Formulae in all Taps other than `homebrew/*`.

GauthamGoli: Actually, this may not be necessary. We can use `Include` and `Exclude` options for individual cops in `.rubocops.yml` to pass file/directory patterns to decide whether cops will run or not. I will update the PR. 
@reitermarkus Where else are the Formulae located?

reitermarkus: > Where else are the Formulae located?

In personal/custom Taps, and I think you can also pass Ruby files directly to `brew style`.

GauthamGoli: Yeah you're right. Thanks. I'll think of some way to handle this.

Library/Homebrew/rubocops/bottle_block_cop.rb
```diff
@@ -1,20 +1,31 @@
...
+            next if block_length(block_node).zero?
+            method, _args, block_body = *block_node
```
GauthamGoli: If there are multiple such unused variables in the same block, instead of prefixing `_` for all of them, using just `_` , would it be okay?

MikeMcQuaid: Yeh, I think using `_` instead of all of them would be good 👍

Library/Homebrew/rubocops/formula_desc_cop.rb
GauthamGoli: While trying to split them into smaller methods, `line_number`, `line_begin_pos`, `desc_begin_pos`, `call_node` are required as arguments in every method. Can I make them instance variables?

reitermarkus: Can you push the changes? Otherwise it's hard to tell if instance variables make sense.

MikeMcQuaid: Agreed; go for it.

GauthamGoli: I've made an additional commit and also some style checks won't be passing. I will squash the commits later once code gets reviewed.

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,84 @@
...
+                      :desc_begin_pos, :call_node
+
+        def audit_formula(node, class_node, _parent_class_node, body)
+          check(node, body, class_node.const_name)
+        end
+
+        private
+
+        def check(node, body, formula_name)
+          body.each_child_node(:send) do |call_node|
```
reitermarkus: You can make this `do |(_receiver, call_name, args)|`.

Library/.rubocop.yml
```diff
@@ -8,7 +8,17 @@ AllCops:
 
 require: ./Homebrew/rubocops.rb
 
+Homebrew/FormulaCop:
```
MikeMcQuaid: Maybe worth putting this in another directory so it doesn't need to be disabled like this?

Library/.rubocop.yml
```diff
@@ -8,7 +8,17 @@ AllCops:
...
+  Enabled: false
+
 Homebrew/CorrectBottleBlock:
+  Include:
+    - '**/Taps/homebrew/**/*.rb'
```
MikeMcQuaid: If this will need to be repeated for every cop I think the existing approach was superior.

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,84 @@
...
+            _receiver, call_name, args = *call_node
+            next if call_name != :desc || args.children[0].empty?
+            @formula_name = formula_name
+            @description = args.children[0]
+            @source_buffer = call_node.source_range.source_buffer
+            @line_number = call_node.loc.line
+            @line_begin_pos = call_node.source_range.source_buffer.line_range(call_node.loc.line).begin_pos
+            @desc_begin_pos = call_node.children[2].source_range.begin_pos
+            @call_node = call_node
```
MikeMcQuaid: I'm thinking some of this could be extracted into a method named `find_method_by_name(:desc)` which then returns the node or perhaps just the parameters? That's the type of thing I'm thinking with splitting this into smaller methods; have generic functions like `find_method_by_name` and later when needed `find_block_by_name` etc. which could take method `name` and optional `block_name` arguments and return e.g. the parameters or information as a hash. This should all live in `FormulaCop` so that it's not necessary to really understand RuboCop to implement a `FormulaCop` subclass and you can just call these documented, public methods. What do you think?

Library/.rubocop.yml
reitermarkus: Also, as I said before, this will exclude every Formula in personal/custom Taps and Ruby files that are passed directly to `brew style`.

We need to exclude `HOMEBREW_LIBRARY_PATH` rather than include `Taps/**`.

MikeMcQuaid: I think just using `'**/Taps/**/*.rb'` may be sufficient? We don't want these applying to non-`HOMEBREW_LIBRARY_PATH` Ruby files.

reitermarkus: They won't, they will only match files with `class ***** < Formula` anyways.

Library/Homebrew/rubocops/formula_desc_cop.rb
GauthamGoli: Yes, this makes sense. Wrapping rubocop specific code into documented methods would make it easier for other maintainers to add custom cops. But, I think, (as of now) its not possible to make writing custom cops to be completely rubocop agnostic as every possible requirement cannot be known beforehand, although we can try to cover most of the common cases,
And I think closures would be a good fit in this use case. I am pushing the code, please let me know further comments.
I apologize for the delay, as I was on a vacation.

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,67 @@
...
+  module Cop
+    module Homebrew
+      class FormulaCop < Cop
+        @registry = Cop.registry
+
+        def on_class(node)
+          # This method is called by RuboCop and is the main entry point
+          class_node, parent_class_node, body = *node
+          return unless a_formula_class?(parent_class_node)
+          audit_formula(node, class_node, parent_class_node, body)
```
reitermarkus: Add a `return unless respond_to?(:audit_formula)` here, so you can remove `Homebrew/FormulaCop` from the RuboCop config.

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,67 @@
+module RuboCop
+  module Cop
+    module Homebrew
+      class FormulaCop < Cop
+        @registry = Cop.registry
```
reitermarkus: What is this used for?

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,78 @@
...
+            # Check if formula's description size exceeds 80
+            desc_length = "#{formula_name}: #{description}".length
+            max_desc_length = 80
+            if desc_length > max_desc_length
+              column = desc_begin_pos - line_begin_pos
+              length = desc_size
+              offense_source_range = source_range(source_buffer, line_number, column, length)
+              desc_length_offense_msg = <<-EOS.undent
+              Description is too long. "name: desc" should be less than #{max_desc_length} characters.
+              Length is calculated as #{formula_name} + desc. (currently #{"#{formula_name}: #{description}".length})
```
reitermarkus: Indent these lines by 2 spaces.

Library/Homebrew/rubocops/extend/formula_cop.rb
GauthamGoli: Ah! That's a hack. I hadn't found a solution for it yet. Without this line, there's an error ```undefined method `enlist' for nil:NilClass``` being thrown.

Library/Homebrew/rubocops/extend/formula_cop.rb
GauthamGoli: Thanks! Added it.

Library/.rubocop.yml
```diff
@@ -8,7 +8,15 @@ AllCops:
 
 require: ./Homebrew/rubocops.rb
 
+
 Homebrew/CorrectBottleBlock:
+  Include:
+    - '**/Taps/homebrew/**/*.rb'
```
MikeMcQuaid: Let's remove these includes and leave the logic in the cops themselves unless there's a way to do this that avoids duplicating this line for every cop.

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+
+        def on_class(node)
+          # This method is called by RuboCop and is the main entry point
+          class_node, parent_class_node, body = *node
+          return unless a_formula_class?(parent_class_node)
+          return unless respond_to?(:audit_formula)
+          audit_formula(node, class_node, parent_class_node, body)
+        end
+
+        def find_method(node, method_name)
```
MikeMcQuaid: `def find_node_method_by_name(node, method_name)` maybe?

Library/Homebrew/rubocops/bottle_block_cop.rb
```diff
@@ -1,20 +1,31 @@
...
-          return if block_length(node).zero?
-          method, _args, body = *node
-          _keyword, method_name = *method
+      class CorrectBottleBlock < FormulaCop
+        MSG = "Use rebuild instead of revision in bottle block".freeze
 
-          return unless method_name == :bottle
-          check_revision?(body)
+        def audit_formula(_node, _class_node, _parent_class_node, formula_class_body_node)
+          check(formula_class_body_node)
```
MikeMcQuaid: Ideally this method would contain code like the following:
```ruby
def audit_formula
  bottle = find_block(:bottle)
  return unless bottle
  if method_called?(bottle, :revision)
    problem "Use rebuild instead of revision in a bottle block"
  end
end
```

The `find_block` and `method_called?` methods should be defined in `formula_cop.rb` and the various instance variables set by `on_class` or something it calls.

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Executes a block for method named `method_name` called inside `node`
+          # Style violations are checked in the block and offenses can be reported
+          node.each_child_node(:send) do |call_node|
+            receiver, call_name, args = *call_node
+            next if call_name != method_name
+            yield receiver, call_name, args, call_node if block_given?
+          end
+        end
+
+        def node_line_begin_pos(node)
```
MikeMcQuaid: `def line_start_column(node)` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Returns the begin position of the node's line in source code
+          node.source_range.source_buffer.line_range(node.loc.line).begin_pos
+        end
+
+        def node_begin_pos(node)
```
MikeMcQuaid: `def start_column(node)` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Returns the begin position of the node in source code
+          node.source_range.begin_pos
+        end
+
+        def node_line_number(node)
```
MikeMcQuaid: `def line_number(node)` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Returns the line number of the node
+          node.loc.line
+        end
+
+        def class_node_name(node)
```
MikeMcQuaid: `def name(node)` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Returns the class node's name, nil if not a class node
+          node.const_name
+        end
+
+        def node_size(node)
```
MikeMcQuaid: `def size(node)` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Returns the node size in the source code
+          node.source_range.size
+        end
+
+        def get_source_buffer(node)
```
MikeMcQuaid: `source_buffer(node)` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Source buffer is required as an argument to report style violations
+          node.source_range.source_buffer
+        end
+
+        def get_str_content(node)
```
MikeMcQuaid: `string_content(node)` maybe?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -0,0 +1,68 @@
...
+          # Returns the string representation if node is of type str
+          node.str_content if node.type == :str
+        end
+
+        private
+
+        def a_formula_class?(parent_class_node)
```
MikeMcQuaid: `is_formula_class?(node)` maybe?

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -0,0 +1,78 @@
...
+    module Homebrew
+      # This cop audits `desc` in Formulae
+      #
+      # - Checks for existence of `desc`
+      # - Checks if size of `desc` > 80
+      # - Checks if `desc` begins with an article
+      # - Checks for correct usage of `command-line` in `desc`
+      # - Checks if `desc` contains the formula name
+      class FormulaDesc < FormulaCop
+        def audit_formula(node, class_node, _parent_class_node, body)
```
MikeMcQuaid: Ideally this method would contain code like the following:
```ruby
def audit_formula
  desc_call = find_method_call(:desc)
  return unless desc_call
  desc = parameters(desc_call).first

  max_length = 80
  if desc.length > max_length
    problem <<-EOS.undent
      Description is too long. "name: desc" should be less than #{max_length} characters.
      Length is calculated as #{formula_name} + desc. (currently #{"#{formula_name}: #{desc}".length})
    EOS

  if desc =~ /(command ?line)/i
    problem "Description should use \"command-line\" instead of \"#{$1}\""
  end

 ... 

end
```

The `find_method_call` and `parameters?` methods should be defined in `formula_cop.rb` and the various instance variables set by `on_class` or something it calls and by e.g. `find_method_call` and `parameters` so `problem` can access e.g. the `source_range`.

Obviously this can (and will) vary but just giving you a heads up for the level of complexity I think would be ideal for these cops to hide the implementation details.

Library/Homebrew/rubocops/extend/formula_cop.rb
reitermarkus: Without the `is_`, I'd say.

Library/Homebrew/rubocops/formula_desc_cop.rb
GauthamGoli: `desc_call = find_method_call(:desc)` we are assuming that there would be only one method call. There might be multiple calls to the same method.  Should it return an array instead?

MikeMcQuaid: I think having `find_method_call` which always returns the `first` would be a sensible first step. If it's needed later we can have `find_method_calls` for e.g. checking use of things like `inreplace` and then loop over them.

GauthamGoli: While trying to implement the above I faced an issue. Just by having 
```ruby
if desc =~ /(command ?line)/i
  problem "Description should use \"command-line\" instead of \"#{$1}\""
end
```
its difficult to set all the instance variables required for pinpointing the `source_range` of the style violation. So I added a method `check_regex_match`. Now the code looks like 
```ruby
match_object = check_regex_match(desc, /(command ?line)/i)
if match_object
  problem "Description should use \"command-line\" instead of \"#{match_object}\""
end
```

MikeMcQuaid: How about:

```ruby
if match = regex_match_group(desc, /(command ?line)/i)
  problem "Description should use \"command-line\" instead of \"#{match_object}\""
end
```

Library/.rubocop.yml
GauthamGoli: As @reitermarkus pointed out, due to `Include` in the `.rubocop.yml` the cops won't be run on personal/custom Taps and the files directly passed as arguments which do not match the pattern under `Include`.
 To resolve this issue I need some clarification.  `brew style` command, runs on all the files inside `HOMEBREW_LIBRARY_PATH` while `brew audit` runs only on the forumlae files. 
Do we want(or is it okay for) the custom cops to get executed under `brew audit` and `brew style` also?

MikeMcQuaid: > As @reitermarkus pointed out, due to Include in the .rubocop.yml the cops won't be run on personal/custom Taps and the files directly passed as arguments which do not match the pattern under Include.

I think it's OK for the taps to run on unnecessary files as they are still a pretty quick no-op.

> Do we want(or is it okay for) the custom cops to get executed under brew audit and brew style also?

Yeh, they should be executed by all of them I think (including `rubocop` obviously, just to be explicit)

GauthamGoli: There are classes in files
`compat/formula_specialties.rb`
`test/exceptions_spec.rb`
`test/support/fixtures/testball.rb`
`test/support/fixtures/testball_bottle.rb`
 inheriting from `Formula` in these files and are returning cop violations as they don't have formula `desc`, causing the build to fail. If `desc` can be added in the above files, then we don't have to repeat `Include` in `.rubocop.yml`

reitermarkus: > If desc can be added in the above files, then we don't have to repeat Include in .rubocop.yml

Why not just check if the file path contains `/Library/Homebrew/`, and exclude those? Didn't we have this approach already before?

GauthamGoli: Well, I was trying to keep the cop code and config code separate in their respective files. But turns out, keeping the `Include` logic in `FormulaCop` is more elegant. Thanks @reitermarkus and @MikeMcQuaid, now Personal Taps can be passed as arguments. 
I added the logic in Formula Cop and pushed the new code. I also used rebase to add another test in bottle_block_cop_spec in a past commit to pass coverage test.

MikeMcQuaid:
This is looking pretty good to merge for me now. @reitermarkus, any thoughts?

MikeMcQuaid:
Great work again @GauthamGoli! My suggestion for the next rules would be https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L251-L309; these are the ones I personally tend to wish to have in my editor the most.

PR #2246, opened 2017-03-03
README: Switch to https:// links

leonklingele:


MikeMcQuaid:
Thanks for all these @leonklingele.

PR #2250, opened 2017-03-04
Add brew cask outdated command

joshka:
Addresses the first part of https://github.com/caskroom/homebrew-cask/issues/29301
Does not (yet) have tests

I'm submitting this PR as a WIP for review. The absence of tests is intentional right now. I'll add tests before it gets merged.

-----


Library/Homebrew/cask/lib/hbc/cli/outdated.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      def self.help
+        "Show casks that have an updated version available"
+      end
+
+      def self.check_installed
+        casks = Hbc.installed
+        casks.each do |cask|
+          odebug "Checking outdated for Cask #{cask}"
```
reitermarkus: Move this into the `check_outdated` method.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,23 @@ def installed?
...
+        versions.each do |version|
+          if version.to_s.casecmp @latest_version
+            @latest_version=version.to_s
+          end
+        end
+      end
+      @latest_version
+    end
+
+    def outdated?(greedy)
```
reitermarkus: `def outdated?(greedy: false)`

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,23 @@ def installed?
       !versions.empty?
     end
 
+    def latest_installed_version
```
reitermarkus: `latest_installed_version` is actually just `versions.last`, since they are sorted by timestamp.

joshka:
@reitermarkus done

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,15 @@ def installed?
...
+      versions.last
+    end
+
+    def outdated?(greedy: false)
+      (greedy || !auto_updates && !version == :latest) &&
+      !version.nil? && (version.casecmp latest_installed_version) == 1
```
reitermarkus: I don't think `version` can be `nil`, so just check for `version == latest_installed_version`.

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
```diff
@@ -0,0 +1,49 @@
...
+          rescue CaskUnavailableError => e
+            onoe e
+          end
+        end
+      end
+
+      def self.check_outdated(cask)
+        odebug "Checking outdated for Cask #{cask.token}"
+        if cask.outdated?(@greedy)
+          puts "#{cask.token}: (#{cask.latest_installed_version}) < #{cask.version}"
```
reitermarkus:
```ruby
if $stdout.tty?
  puts "#{cask.token}: (#{cask.latest_installed_version}) < #{cask.version}"
else
  puts cask.token
end
```

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
```diff
@@ -0,0 +1,49 @@
...
+        cask_tokens = cask_tokens_from(args)
+        if cask_tokens.empty?
+          check_installed
+        else
+          check_casks(cask_tokens)
+        end
+      end
+
+      def self.help
+        "Show casks that have an updated version available"
```
reitermarkus: Make `Casks` uppercase.

Library/Homebrew/cask/lib/hbc/cask.rb
vitorgalvao: A non-existing `version` would fail `audit`, if I’m not mistaken.

reitermarkus: Yes, and `version == latest_installed_version` would still work even if it was `nil`.

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def self.check_installed
+        casks = Hbc.installed
+        casks.each do |cask|
+          check_outdated(cask)
+        end
```
reitermarkus: `Hbc.installed.each(&method(:check_outdated))`

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: Version is nil when you've installed casks that aren't yet in a tap. E.g I have a couple of casks that I wrote that I submitted PRs for but never followed up to complete (sorry). They show up with a ! in brew cask list. This check is here as the call to casecmp fails otherwise. We need the casecmp for a later than check rather than an equality check.

reitermarkus: > We need the casecmp for a later than check rather than an equality check.

How is a case comparison helping with deciding whether a version is later than another given version? I think it's safe to assume that any other version other than the installed one is “later“, since I have never seen a Cask get downgraded.

So just make this `version != latest_version_installed`.

joshka: I agree. More the point is outdated should mean "does not match the version in HBC"

vitorgalvao: > I have never seen a Cask get downgraded.

It has happened, but it’s mostly when the new version has a major bug that stops the app from running/installing, so even in those cases the downgrade can be considered an upgraded (you’d want the earlier version).

Either way, it has been discussed in the past we can’t trust any comparison other than “`version` is different” because our versions can be weird strings of characters. You can’t compare a hash to another and determine which was done later.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,16 @@ def installed?
       !versions.empty?
     end
 
+    def latest_installed_version
+      versions.last
+    end
+
+    def outdated?(greedy)
+      (greedy || !auto_updates && !version == :latest) &&
+      !version.nil? &&  # ignore Casks without sources
```
reitermarkus: I'd make this the first line, like this: `return false if version.nil?`

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,16 @@ def installed?
       !versions.empty?
     end
 
+    def latest_installed_version
+      versions.last
+    end
+
+    def outdated?(greedy)
```
reitermarkus: `def outdated?(greedy: false)`

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
```diff
@@ -0,0 +1,50 @@
...
+            onoe e
+          end
+        end
+      end
+
+      def self.check_outdated(cask)
+        odebug "Checking outdated for Cask #{cask.token}"
+        if cask.outdated?(@greedy)
+          if $stdout.tty?
+            puts "#{cask.token}: (#{cask.latest_installed_version}) < #{cask.version}"
```
reitermarkus: Make this `!=` instead of `<`.

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: Is there much difference semantically to instead swapping the two lines? Happy to put this first, but I don't understand what the `return false if` adds.

If anything, perhaps pulling this out to a small method e.g. `has_source` would be better from a self documenting perspective.

Happy to go whichever way on this as it seems mostly preferential.

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: Woops, I added this the other day, but deleted it in a bug fix

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: This is fixed, but the comment is on the wrong line to pick up the change.

Library/Homebrew/cask/lib/hbc/cask.rb
reitermarkus: This is mostly just to make it more readable.

`(greedy || !auto_updates && !version == :latest)` is also pretty hard to follow. I'd probably write something like this: 

```ruby
unless greedy
  return false if auto_updates
  return false if version.latest?
end
```

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,20 @@ def installed?
       !versions.empty?
     end
 
+    def latest_installed_version
+      versions.last
+    end
+
+    def has_source?()
+      !version.nil?
+    end
```
reitermarkus: I think this was a good idea in theory, but now that I see it, this is even more unclear, since there is no obvious relationship between missing source and version being `nil`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,20 @@ def installed?
...
+
+    def outdated?(greedy = false)
+      has_source? &&
+      (greedy || !auto_updates && !version == :latest) &&
```
reitermarkus: `!version == :latest` ≠ `version != :latest`

Anyways, you should use `version.latest?`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,20 @@ def installed?
...
+      version != latest_installed_version.to_s
```
reitermarkus: You can remove `to_s` here after https://github.com/Homebrew/brew/pull/2282 is merged.

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: I agree

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: Lol yep

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: I've subscribed to that PR and will update this when it's merged.

ilovezfs:
Is it the case that there is no `brew cask upgrade` commabd? If that's true, it seems like putting the cart before the horse to have a `brew cask outdated` command.

joshka:
> Is it the case that there is no brew cask upgrade command?

No, a brew cask upgrade command does not yet exist. It's been on the drawing board for a very long time. See https://github.com/caskroom/homebrew-cask/issues/29301

> If that's true, it seems like putting the cart before the horse to have a brew cask outdated command.

Not really, the first step to knowing how to upgrade is knowing what to upgrade. I can use `brew cask outdated` to list Casks that require a manual upgrade already, which has some immediate utility despite not being able to do anything other than manually reinstall those. In actual fact `brew cask reinstall $(brew cask outdated)` does exactly what you'd think it should (but doesn't handle the rollback stuff.

See https://github.com/caskroom/homebrew-cask/issues/29301 for much more detail on the spec and rationale for this.

Library/Homebrew/cask/lib/hbc/cask.rb
joshka: I removed `to_s` and rebased

joshka:
I''ll add `--verbose` and `--quiet` as well as tests before this is ready to merge

joshka:
TODO: show all versions instead of just the latest version in the outdated list

vitorgalvao:
> TODO: show all versions instead of just the latest version in the outdated list

I wouldn’t argue it that feature did not exist. [As mentioned](https://github.com/caskroom/homebrew-cask/issues/29301#issuecomment-285622306), we only support one version at a time. When you’re moving things to `/Applications`, there’s really no way you actually have two versions installed at the same time.

The goal is even to [stop supporting multiple versions of the same cask](https://github.com/caskroom/homebrew-cask/issues/13201) (item **5**), so I don’t see it as necessary.

joshka:
Closing as https://github.com/Homebrew/brew/pull/2309 is a better version of this.

PR #2254, opened 2017-03-04
Lowercase `caskroom` and add missing Taps.

reitermarkus:

---

It's actually `caskroom`, not `Caskroom`. Also added newly created Taps.

### DIFF
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -100,10 +100,8 @@ module Homebrew
     raise SEARCH_ERROR_QUEUE.pop unless SEARCH_ERROR_QUEUE.empty?
   end
 
-  SEARCHABLE_TAPS = OFFICIAL_TAPS.map { |tap| ["Homebrew", tap] } + [
-    %w[caskroom cask],
-    %w[caskroom versions],
-  ]
+  SEARCHABLE_TAPS = OFFICIAL_TAPS.map { |tap| ["Homebrew", tap] } +
+                    OFFICIAL_CASK_TAPS.map { |tap| ["caskroom", tap] }
 
   def query_regexp(query)
     case query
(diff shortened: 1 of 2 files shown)


Library/Homebrew/official_taps.rb
```diff
@@ -8,6 +8,14 @@
   tex
 ].freeze
 
+OFFICIAL_CASK_TAPS = %w[
```
MikeMcQuaid: I'm hesitant to add any more of these as it slows down every search for each additional one we add in here (and I'm currently actively trying to remove Homebrew's ones)

Library/Homebrew/cmd/search.rb
```diff
@@ -127,7 +125,7 @@ def search_tap(user, repo, regex_or_string)
     regex = regex_or_string.is_a?(String) ? /^#{Regexp.escape(regex_or_string)}$/ : regex_or_string
 
     if (HOMEBREW_LIBRARY/"Taps/#{user.downcase}/homebrew-#{repo.downcase}").directory? && \
-       user != "Caskroom"
+       user != "caskroom"
```
MikeMcQuaid: Out of interest: what difference does the casing make or just aesthetic? I have no preference on this.

Library/Homebrew/official_taps.rb
MikeMcQuaid: There may also be a way to use a smarter GitHub search that involves hitting one endpoint for multiple taps rather than multiple endpoints.

Library/Homebrew/cmd/search.rb
reitermarkus: It's just aesthetic. The GitHub username is lowercase, so it doesn't make sense to capitalize it when on the other hand we are lowercasing `homebrew`, which actually is capitalized.

MikeMcQuaid: 🆒 👍

Library/Homebrew/official_taps.rb
MikeMcQuaid: @reitermarkus Thoughts?

reitermarkus: > There may also be a way

I thought this implied that you know of another way to do this, so I haven't looked into it.

PR #2261, opened 2017-03-05
audit: whitelist node@* to use conflicts_with

ilovezfs:

-----

Add exceptions for node@* versioned formulae to use conflicts_with instead of keg_only :versioned_formula since they, and the main node formula, all currently overwrite npm during postinstall.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -519,14 +519,6 @@ class FormulaAuditor
   end
 
   def audit_conflicts
-    if formula.conflicts.any? && formula.versioned_formula?
-      problem <<-EOS
-        Versioned formulae should not use `conflicts_with`.
-        Use `keg_only :versioned_formula` instead.
-      EOS
-      return
-    end
-
     formula.conflicts.each do |c|
       begin
         Formulary.factory(c.name)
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -519,6 +519,8 @@ def audit_deps
   end
 
   def audit_conflicts
+    conflicts_whitelist = %w[node@0.12 node@0.10]
```
MikeMcQuaid: I reckon stick this inside the `formula.versioned_formula?` `if`.

ilovezfs: do you mean in formula.rb?

MikeMcQuaid: Sorry, I mean move the addition down inside the `if formula.conflicts.any? && formula.versioned_formula?` block as I think the other conflicts changes should be done. Alternatively, move this whole lot under the loop below.

ilovezfs: PR refreshed.

ilovezfs:
As in, `return if formula.name.start_with? "node@"` instead of a generic whitelist?

MikeMcQuaid:
> As in, return if formula.name.start_with? "node@" instead of a generic whitelist?

Either is fine with me but just all the versions going under the 🔪

ilovezfs:
@MikeMcQuaid PR refreshed.

PR #2269, opened 2017-03-05
audit: whitelist bash-completion@* to use conflicts_with

JCount:

-----
this is necessary because they install conflicting scripts into HOMEBREW_PREFIX/etc/profile.d

I changed how the whitelisted formulae are stored, both to improve readability and to make it easier to potentially add more formulae in the future if necessary. I tried to make the return conditional compact, but I am a bit uncertain if it conforms to the style of the rest of the code, so any feedback there would be welcome.

bash-completion@2 PR this is meant to address: https://github.com/Homebrew/homebrew-core/pull/10594

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -532,7 +532,7 @@ class FormulaAuditor
       end
     end
 
-    versioned_conflicts_whitelist = %w[node@ bash-completion@]
+    versioned_conflicts_whitelist = %w[node@ bash-completion@].freeze
 
     return unless formula.conflicts.any? && formula.versioned_formula?
     return if formula.name.start_with?(*versioned_conflicts_whitelist)
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -532,8 +532,11 @@ class FormulaAuditor
       end
     end
 
+    versioned_conflicts_whitelist = %w[node@ bash-completion@]
+
     return unless formula.conflicts.any? && formula.versioned_formula?
-    return if formula.name.start_with? "node@"
+    versioned_conflicts_whitelist.each { |c|
+      return if formula.name.start_with? c }
     problem <<-EOS
       Versioned formulae should not use `conflicts_with`.
       Use `keg_only :versioned_formula` instead.
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -535,8 +535,7 @@ class FormulaAuditor
     versioned_conflicts_whitelist = %w[node@ bash-completion@]
 
     return unless formula.conflicts.any? && formula.versioned_formula?
-    versioned_conflicts_whitelist.each { |c|
-      return if formula.name.start_with? c }
+    return if formula.name.start_with?(*versioned_conflicts_whitelist)
     problem <<-EOS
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -532,8 +532,11 @@ def audit_conflicts
       end
     end
 
+    versioned_conflicts_whitelist = %w[node@ bash-completion@]
+
     return unless formula.conflicts.any? && formula.versioned_formula?
-    return if formula.name.start_with? "node@"
+    versioned_conflicts_whitelist.each { |c|
+      return if formula.name.start_with? c }
```
reitermarkus: You can do this:

```ruby
return if formula.name.start_with?(*versioned_conflicts_whitelist)
```

JCount: I wasn't certain if homebrew used "splat" much, so I wrote a longer version

reitermarkus: > I wasn't certain if homebrew used "splat" much

I think it is, but even if it wasn't, I would say in this case it is certainly more readable.

JCount: That is a very good point. 👍

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -532,8 +532,11 @@ def audit_conflicts
       end
     end
 
+    versioned_conflicts_whitelist = %w[node@ bash-completion@]
```
reitermarkus: Maybe make this into a constant?

JCount: I assume you mean freezing it to make it an immutable array, not making it a true constant.

reitermarkus: I actually meant making it a true constant, since a whitelist seems to me like it should be a constant.

JCount: Ok. I can change it to that, but the issue then is that it would need to be moved out method, and into the FormulaAuditor class scope to prevent a 'dynamic constant assignment' error. Also, there are other whitelists that aren't made into constants, https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L768-L790, for example. However, it may make sense to make it a true constant. I will create a comment on the current code so that it is more visible.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -532,8 +532,10 @@ def audit_conflicts
       end
     end
 
+    versioned_conflicts_whitelist = %w[node@ bash-completion@].freeze
```
JCount: Is there any compelling reason to move this out of the `audit_conflicts` method and turn it into a true constant?

reitermarkus: I think it would make sense to move all of these whitelists outside, since they really are supposed to be constants, and it makes the methods easier to read if there aren't a billion lines of constants in-between “actual“ code.

JCount: That sounds like a perfectly valid argument to me. When I get a chance, I will transform this into a true constant. The other lists should probably be done in a separate PR I think.

reitermarkus: Yes, I'd leave this as is for now, and convert all of them in one go in a separate PR.

PR #2274, opened 2017-03-06
Port OsxfuseRequirement to Linux

rwhogg:

-----

Next round of Linuxbrew upstreaming changes.

We've been getting a lot of requests recently for FUSE support, so it made sense to tackle this one next.

### DIFF
--- a/Library/Homebrew/dev-cmd/tests.rb
+++ b/Library/Homebrew/dev-cmd/tests.rb
@@ -97,6 +97,10 @@ module Homebrew
         files = files.reject { |p| p =~ %r{^test/(os/mac|cask)(/.*|_spec\.rb)$} }
       end
 
+      unless OS.linux?
+        files = files.reject { |p| p =~ %r{^test/os/linux(/.*|_spec\.rb)$} }
+      end
+
       if parallel
         system "bundle", "exec", "parallel_rspec", *opts, "--", *args, "--", *files
       else
--- a/Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
+++ b/Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
@@ -1,0 +1,39 @@
+require "requirement"
+
+class OsxfuseRequirement < Requirement
+  download "https://github.com/libfuse/libfuse"
+
+  satisfy(build_env: false) do
+    next true if libfuse_formula_exists? && Formula["libfuse"].installed?
+    includedirs = %w[
+      /usr/include
+      /usr/local/include
+    ]
+    next true if (includedirs.map do |dir|
+      File.exist? "#{dir}/fuse.h"
+    end).any?
+    false
+  end
+
+  def message
+    msg = "libfuse is required to install this formula.\n"
+    if libfuse_formula_exists?
+      msg + <<-EOS.undent
+        Run "brew install libfuse" to install it.
+      EOS
+    else
+      msg + super
+    end
+  end
+
+  private
+
+  def libfuse_formula_exists?
+    begin
+      Formula["libfuse"]
+    rescue FormulaUnavailableError
+      return false
+    end
+    true
+  end
+end
--- a/Library/Homebrew/extend/os/mac/requirements/osxfuse_requirement.rb
+++ b/Library/Homebrew/extend/os/mac/requirements/osxfuse_requirement.rb
@@ -1,0 +1,34 @@
+require "requirement"
+
+class OsxfuseRequirement < Requirement
+  cask "osxfuse"
+  download "https://osxfuse.github.io/"
+
+  satisfy(build_env: false) { self.class.binary_osxfuse_installed? }
+
+  def self.binary_osxfuse_installed?
+    File.exist?("/usr/local/include/osxfuse/fuse.h") &&
+      !File.symlink?("/usr/local/include/osxfuse")
+  end
+
+  env do
+    ENV.append_path "PKG_CONFIG_PATH", HOMEBREW_LIBRARY/"Homebrew/os/mac/pkgconfig/fuse"
+
+    unless HOMEBREW_PREFIX.to_s == "/usr/local"
+      ENV.append_path "HOMEBREW_LIBRARY_PATHS", "/usr/local/lib"
+      ENV.append_path "HOMEBREW_INCLUDE_PATHS", "/usr/local/include/osxfuse"
+    end
+  end
+end
+
+class NonBinaryOsxfuseRequirement < Requirement
+  fatal true
+  satisfy(build_env: false) { HOMEBREW_PREFIX.to_s != "/usr/local" || !OsxfuseRequirement.binary_osxfuse_installed? }
+
+  def message
+    <<-EOS.undent
+      osxfuse is already installed from the binary distribution and
+      conflicts with this formula.
+    EOS
+  end
+end
--- a/Library/Homebrew/requirements/osxfuse_requirement.rb
+++ b/Library/Homebrew/requirements/osxfuse_requirement.rb
@@ -2,34 +2,10 @@ require "requirement"
 
 class OsxfuseRequirement < Requirement
   fatal true
-  cask "osxfuse"
-  download "https://osxfuse.github.io/"
-
-  satisfy(build_env: false) { self.class.binary_osxfuse_installed? }
-
-  def self.binary_osxfuse_installed?
-    File.exist?("/usr/local/include/osxfuse/fuse.h") &&
-      !File.symlink?("/usr/local/include/osxfuse")
-  end
-
-  env do
-    ENV.append_path "PKG_CONFIG_PATH", HOMEBREW_LIBRARY/"Homebrew/os/mac/pkgconfig/fuse"
-
-    unless HOMEBREW_PREFIX.to_s == "/usr/local"
-      ENV.append_path "HOMEBREW_LIBRARY_PATHS", "/usr/local/lib"
-      ENV.append_path "HOMEBREW_INCLUDE_PATHS", "/usr/local/include/osxfuse"
-    end
-  end
 end
 
 class NonBinaryOsxfuseRequirement < Requirement
-  fatal true
-  satisfy(build_env: false) { HOMEBREW_PREFIX.to_s != "/usr/local" || !OsxfuseRequirement.binary_osxfuse_installed? }
-
-  def message
-    <<-EOS.undent
-      osxfuse is already installed from the binary distribution and
-      conflicts with this formula.
-    EOS
-  end
+  fatal false
 end
+
+require "extend/os/requirements/osxfuse_requirement"
--- a/Library/Homebrew/test/os/mac/osxfuse_requirement_spec.rb
+++ b/Library/Homebrew/test/os/mac/osxfuse_requirement_spec.rb
@@ -1,0 +1,36 @@
+require "requirements/osxfuse_requirement"
+
+describe OsxfuseRequirement do
+  subject { described_class.new([]) }
+
+  describe "::binary_osxfuse_installed?" do
+    it "returns false if fuse.h does not exist" do
+      allow(File).to receive(:exist?).and_return(false)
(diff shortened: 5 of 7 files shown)


Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -80,6 +80,10 @@ def tests
         files = files.reject { |p| p =~ %r{^test/(os/mac|cask)(/.*|_spec\.rb)$} }
       end
 
+      unless OS.linux?
+        files = files.reject { |p| p =~ %r{^test/os/linux(/.*|_spec\.rb)$} }
```
rwhogg: This prevents the Linux-specific tests from running on (and consequently breaking) macOS.

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
```diff
@@ -0,0 +1,11 @@
+require "requirement"
+
+class OsxfuseRequirement < Requirement
+  default_formula "libfuse"
```
rwhogg: This is a Linuxbrew formula.

Since, unlike with the Java requirement, the relevant formula only supports Linux, this is placed here rather than in the OS-agnostic file.

I've made the assumption that if you're using brew and it's NOT on a Mac, you're using Linuxbrew's homebrew-core. Is that a reasonable assumption to make?

Library/Homebrew/extend/os/mac/requirements/osxfuse_requirement.rb
```diff
@@ -0,0 +1,34 @@
+require "requirement"
```
rwhogg: This file should have basically no substantive changes from the original, it's just been moved into a Mac-specific place.

Library/Homebrew/requirements/osxfuse_requirement.rb
```diff
@@ -2,34 +2,10 @@
...
-  fatal true
-  satisfy(build_env: false) { HOMEBREW_PREFIX.to_s != "/usr/local" || !OsxfuseRequirement.binary_osxfuse_installed? }
-
-  def message
-    <<-EOS.undent
-      osxfuse is already installed from the binary distribution and
-      conflicts with this formula.
-    EOS
-  end
+  fatal false
```
rwhogg: An explanation, since this probably seems a little weird:

The notion of a "binary osxfuse distribution" conflicting with a brew-provided one is particular to macOS. However, for backwards compatibility reasons, this class has to remain defined here.

Library/Homebrew/test/os/mac/osxfuse_requirement_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      allow(File).to receive(:symlink?).and_return(true)
+      expect(described_class).not_to be_binary_osxfuse_installed
+    end
+  end
+
+  describe "environment" do
+    it "adds the fuse directories to the appropriate paths" do
+      expect(ENV).to receive(:append_path).with("PKG_CONFIG_PATH", any_args)
+      expect(ENV).to receive(:append_path).with("HOMEBREW_LIBRARY_PATHS", any_args)
+      expect(ENV).to receive(:append_path).with("HOMEBREW_INCLUDE_PATHS", any_args)
```
rwhogg: Note that this test exploits the fact that `HOMEBREW_PREFIX` is set to a value other than /usr/local for the test environment. I'm not 100% pleased with this and will change it if requested.

Library/Homebrew/dev-cmd/tests.rb
MikeMcQuaid: Can you flip the two `unless`s to be an `if OS.mac?` `elsif OS.linux?` `else`?

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
MikeMcQuaid: > I've made the assumption that if you're using brew and it's NOT on a Mac, you're using Linuxbrew's homebrew-core. Is that a reasonable assumption to make?

I think that's not a safe assumption yet so let's just omit `default_formula` for now.

Library/Homebrew/requirements/osxfuse_requirement.rb
MikeMcQuaid: Makes sense, thanks 👍

Library/Homebrew/test/os/mac/osxfuse_requirement_spec.rb
MikeMcQuaid: I think that's a reasonable assumption 👍

Library/Homebrew/dev-cmd/tests.rb
reitermarkus: @MikeMcQuaid, how?

`unless OS.mac?` = `if OS.linux? || ENV["HOMEBREW_TEST_GENERIC_OS"]`
`unless OS.linux?` = `if OS.mac? || ENV["HOMEBREW_TEST_GENERIC_OS"]`

MikeMcQuaid: Ignore me then 😁

reitermarkus:
By the way, does it make sense to also call it `OsxfuseRequirement` on Linux?

rwhogg:
> By the way, does it make sense to also call it OsxfuseRequirement on Linux?

Not at all, but I deliberately left any name changes out of this PR so it could focus on functionality and behaviour.

The ideal resolution (from my POV, not necessarily Linuxbrew's) would be to deprecate the old name and rename it to just FuseRequirement, but I'm not sure how difficult that is to accomplish or if it would have any negative impact on macOS users.

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
rwhogg: Ok. That will cause issues building FUSE formulae on CI, but since nothing in core has a hard dependency on it that shouldn't come up any time soon, so IMO that's not a problem.

This will require a change to the `satisfy` block too, to check for formula-unavailable errors.

reitermarkus:
I think you could do something similar to: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/os/mac.rb#L14

```ruby
class FuseRequirement < Requirement
  ::OsxfuseRequirement = self
end
```

MikeMcQuaid:
> By the way, does it make sense to also call it OsxfuseRequirement on Linux?

It can be renamed and added to `Library/Homebrew/compat/requirements.rb`; you can see this happened before, too:
```
OsxfuseDependency          = OsxfuseRequirement
ConflictsWithBinaryOsxfuse = NonBinaryOsxfuseRequirement
```

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
MikeMcQuaid: 🆒

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
```diff
@@ -0,0 +1,37 @@
+require "requirement"
+
+class OsxfuseRequirement < Requirement
+  download "https://github.com/libfuse/libfuse"
```
MikeMcQuaid: Does this want to be `default_formula "libfuse"` instead?

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  satisfy(build_env: false) do
+    next true if libfuse_formula_exists? && Formula["libfuse"].installed?
+    includedirs = %w[
+      /usr/include
+      /usr/local/include
+    ]
+    next true if includedirs.map { |dir| File.exist?(File.join(dir, "fuse.h")) }.any?
```
MikeMcQuaid: Wrap this line if it exceeds 80 chars.

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
rwhogg: Very much so, but my understanding was you preferred we not include that. Did you change your mind or am I just misunderstanding your previous comments?

MikeMcQuaid: If this doesn't 💥 if this formula is missing (i.e. try it with `libfuse2`) I'm fine for the `default_formula` to be set.

MikeMcQuaid: (apologies to contradict myself)

rwhogg: Darn, it does blow up :boom: 

```
Error: No available formula with the name "libfuse2" (dependency of fossil)
==> Searching for a previously deleted formula...
Error: No previously deleted formula found.
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

rwhogg:
@MikeMcQuaid fixed the style issue from before, thanks for pointing it out.

Side question:
Since the coverage report doesn't run on Linux, should we exclude files under extend/os/linux from the codecov CI requirements?

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  satisfy(build_env: false) do
+    next true if libfuse_formula_exists? && Formula["libfuse"].installed?
+    includedirs = %w[
+      /usr/include
+      /usr/local/include
+    ]
+    next true if (includedirs.map do |dir|
+      File.exist?(File.join(dir, "fuse.h"))
+    end).any?
```
rwhogg: Admittedly, I couldn't find a good way to write this that kept it below 120 chars. Sorry if this is kind of ugly, feel free to suggest something better or point me to the styleguide (I couldn't find it for the life of me...)

MikeMcQuaid:
> Since the coverage report doesn't run on Linux, should we exclude files under extend/os/linux from the codecov CI requirements?

Yeh, good idea 👍

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
MikeMcQuaid: It's fine, thanks.

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
```diff
@@ -0,0 +1,39 @@
...
+  download "https://github.com/libfuse/libfuse"
+
+  satisfy(build_env: false) do
+    next true if libfuse_formula_exists? && Formula["libfuse"].installed?
+    includedirs = %w[
+      /usr/include
+      /usr/local/include
+    ]
+    next true if (includedirs.map do |dir|
+      File.exist?(File.join(dir, "fuse.h"))
```
MikeMcQuaid: I don't think you need to `File.join` here; just interpolated string

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
MikeMcQuaid: Yeh, let's pass then. Perhaps a potential Homebrew/brew PR is in your future to handle that 😀

Library/Homebrew/extend/os/linux/requirements/osxfuse_requirement.rb
rwhogg: Fixed, thanks

rwhogg:
@MikeMcQuaid this is ready for another review. Fixed the `File.join` issue you mentioned.

MikeMcQuaid:
LGTM, thanks @rwhogg!

PR #2297, opened 2017-03-08
linkage_checker: simplify logic for declared deps and include recursive deps

zmwangx:

-----

Use `Formula#runtime_dependencies` to collect dependencies.

This inevitably pulls in all recursive deps, which might be controversial (I think this was discussed during the `:linked` dependency days, and no conclusion was reached; the discussion died off after `:linked` was dropped, but dependency resolution has improved since then, so it's worth taking another look), but IMO is a good idea.

Concrete example: in the case of `ffmpeg`, depending on either `theora` or `libvorbis` (both of which has a hard dependency on `libogg`) would introduce a `libogg` linkage, which is currently listed under "possible undeclared dependencies".

### DIFF
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -54,18 +54,10 @@ class LinkageChecker
   end
 
   def check_undeclared_deps
-    filter_out = proc do |dep|
-      next true if dep.build?
-      next false unless dep.optional? || dep.recommended?
-      formula.build.without?(dep)
-    end
-    declared_deps = formula.deps.reject { |dep| filter_out.call(dep) }.map(&:name)
-    declared_requirement_deps = formula.requirements.reject { |req| filter_out.call(req) }.map(&:default_formula).compact
-    declared_dep_names = (declared_deps + declared_requirement_deps).map { |dep| dep.split("/").last }
+    declared_deps = keg.to_formula.runtime_dependencies.map { |dep| dep.to_formula.full_name }
     undeclared_deps = @brewed_dylibs.keys.select do |full_name|
-      name = full_name.split("/").last
-      next false if name == formula.name
-      !declared_dep_names.include?(name)
+      next false if full_name == formula.full_name
+      !declared_deps.include?(full_name)
     end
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Code looks sound to me but wondering if @ilovezfs @alyssais have thoughts on the results.

ilovezfs:
This will mean we have no easy way of telling which linkages are or are not explicitly included in the formula.

zmwangx:
> This will mean we have no easy way of telling which linkages are or are not explicitly included in the formula.

Yes. But keep in mind that recursive dependencies are explicitly exposed in LDFLAGS. Complaining about it later feels inconsistent.

ilovezfs:
Printing them as possibly undeclared doesn't necessarily amount to complaining.

Ideally I think we should print out the indirect, undeclared ones in one list and the flat-out opportunistic ones in another list rather than lumping them together in a generic undeclared pile. But just not printing the former makes it harder to answer the question: what's this linked to that's not noted explicitly with a depends_on?

zmwangx:
> what's this linked to that's not noted explicitly with a depends_on?

Aw, smells like `:linked`...

ilovezfs:
Yes, by definition. One hopes `=> :linked` was never intended for the opportunistic ones.

MikeMcQuaid:
Not sure on the state of this. @ilovezfs? Feel free to point out what would need done from your perspective and/or close if we're at the end of this road.

MikeMcQuaid:
> Ideally I think we should print out the indirect, undeclared ones in one list and the flat-out opportunistic ones in another list rather than lumping them together in a generic undeclared pile.

This seems like the best idea. It won't result in code simplification but printing out two different lists here feels like the right thing for now.

ilovezfs:
>This seems like the best idea. It won't result in code simplification but printing out two different lists here feels like the right thing for now.

Agreed. @zmwangx are you willing to implement it as two lists?

zmwangx:
> are you willing to implement it as two lists?

I may take a look this weekend, but I'm kind of short on time recently and is juggling free time between several things, so feel free.

zmwangx:
Sorry guys, I forgot about this. I'll close this since the improvement would be very minor anyway.

Issue #2237, opened 2017-02-28
Invoking non-llvm_clang compiler when building with brew LLVM OpenMP results in xcrun script failing


aw1621107:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [ ] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

I'm fairly certain this is my fault...

Normally, when `llvm_clang` is invoked, Cmd#tool replaces `llvm_clang` with the path to the clang executable installed as part of brew LLVM. However, when `clang` is invoked, brew replaces `clang` with `llvm_clang`, which causes the `xcrun` script to fail:

```
bash-3.2$ llvm_clang
Undefined symbols for architecture x86_64:
  "_main", referenced from:
     implicit entry/start for main executable
ld: symbol(s) not found for architecture x86_64
clang-3.9: error: linker command failed with exit code 1 (use -v to see invocation)
bash-3.2$ clang

Failed to execute llvm_clang -isystem/usr/local/include -isystem/usr/include/libxml2 -isystem/opt/X11/include -isystem/opt/X11/include/freetype2 -isystem/System/Library/Frameworks/OpenGL.framework/Versions/Current/Headers -I/usr/local/opt/tcl-tk/include -I/usr/local/opt/ncurses/include -I/tmp/brew-llvm-openmp -L/usr/local/opt/tcl-tk/lib -L/usr/local/opt/ncurses/lib -L/usr/local/lib -L/tmp/brew-llvm-openmp -L/opt/X11/lib -L/System/Library/Frameworks/OpenGL.framework/Versions/Current/Libraries -Wl,-headerpad_max_install_names

Xcode and/or the CLT appear to be misconfigured. Try one or both of the following:
  xcodebuild -license
  sudo xcode-select -switch /path/to/Xcode.app

bash-3.2$
```

This occurs when using brew LLVM to compile `fftw --with-mpi --with-openmp`. `./configure` tests the OpenMPI installation by invoking `/usr/local/bin/mpicc`, which in turn invokes `clang`, which causes configuration to fail with the above error.

A dirty fix would be to replace the `HOMEBREW_CC` environment variable with the full path to the clang executable installed as part of brew LLVM, but obviously a cleaner fix would be preferred.

Thoughts?

[...]

PR #2298, opened 2017-03-09
super/cc: Fix --cc=llvm_clang invoked as clang

sjackman:

-----

Fix the error:
```
brew sh --cc=llvm_clang <<<'clang --version'
Failed to execute llvm_clang --version
```

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -72,7 +72,11 @@ class Cmd
     else
       # Note that this is a universal fallback, so that we'll always invoke
       # HOMEBREW_CC regardless of what name under which the tool was invoked.
-      ENV["HOMEBREW_CC"]
+      if ENV["HOMEBREW_CC"] == "llvm_clang"
+        "#{ENV["HOMEBREW_PREFIX"]}/opt/llvm/bin/clang"
+      else
+        ENV["HOMEBREW_CC"]
+      end
     end
   end
 


Library/Homebrew/shims/super/cc
```diff
@@ -72,7 +72,11 @@ class Cmd
     else
       # Note that this is a universal fallback, so that we'll always invoke
       # HOMEBREW_CC regardless of what name under which the tool was invoked.
-      ENV["HOMEBREW_CC"]
+      if ENV["HOMEBREW_CC"] == "llvm_clang"
+        "#{ENV["HOMEBREW_PREFIX"]}/opt/llvm/bin/clang"
```
MikeMcQuaid: This looks like it should be handled by line 61; any idea why it isn't?

sjackman: Yep. This patch addresses the case when `HOMEBREW_CC == llvm_clang` and `super/cc` is invoked as any compiler other than `llvm_clang`, for example `clang` or `gcc`. Line 61 tests whether `arg0 == llvm_clang` but not `HOMEBREW_CC == llvm_clang`.

MikeMcQuaid: Gotcha, thanks.

MikeMcQuaid:
Thanks again @sjackman!

aw1621107:
This might fix #2237

PR #2302, opened 2017-03-10
Revert "Revert "Revert "cc: always filter flags on deps."""

ilovezfs:
Reverts Homebrew/brew#2211

This is still causing problems. For example, vtk cannot compile --with-python3.

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -206,6 +206,10 @@ class Cmd
   end
 
   def keep?(path)
+    # The logic in this method will eventually become the default,
+    # but is currently opt-in.
+    return keep_orig?(path) unless ENV["HOMEBREW_EXPERIMENTAL_FILTER_FLAGS_ON_DEPS"]
+
     # Allow references to self
     if formula_prefix && path.start_with?("#{formula_prefix}/")
       true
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
This time I want to try and fix these cases as they come up. Will try to do so today.

ilovezfs:
Unfortunately since this shipped in the tag, we now have users out of commission who depend on vtk with python3 support.

MikeMcQuaid:
Yes but if I can get a fix today I'm not sure a tag is going to be any slower. Other options worth considering beyond a fix:
- whitelist individual taps
- a formula `env` DSL to enable loosening this behaviour

MikeMcQuaid:
Also:
- disable this if options are used

ilovezfs:
>a formula env DSL to enable loosening this behaviour

that plus a new tag seems like an appropriate stop-gap so we can stop the revert-revert-revert-revert-revert ping pong

MikeMcQuaid:
Holding off on this until we've got reproducible test cases.

ilovezfs:
https://github.com/Homebrew/brew/issues/2333

iMichka:
So the problem I had, and the other user had, is that `brew install vtk --without-python --with-python3` did a `brew update`, but did not `brew upgrade`. So we had Python 3.6.0 installed, whereas Python 3.6.0_1 should have been there. Probably the reason why the path got removed. This problem may occur with other formulas I guess.

MikeMcQuaid:
> but did not brew upgrade

Yeh, this is the bit that needs fixed 👍

MikeMcQuaid:
Closing in favour of https://github.com/Homebrew/brew/issues/2348

ilovezfs:
@MikeMcQuaid I've re-opened this because #2348 looks like it needs to be reverted.

PR #2308, opened 2017-03-10
linux: Use -march=native for optimization_flags

sjackman:

-----

Linuxbrew uses `-march=native` by default and `-march=core2` for 64-bit Linux bottles as on Homebrew.

See https://github.com/Linuxbrew/brew#257

### DIFF
--- a/Library/Homebrew/extend/os/linux/hardware/cpu.rb
+++ b/Library/Homebrew/extend/os/linux/hardware/cpu.rb
@@ -1,6 +1,16 @@
 module Hardware
   class CPU
     class << self
+      LINUX_OPTIMIZATION_FLAGS = {
+        core2: OPTIMIZATION_FLAGS[:core2],
+        core: OPTIMIZATION_FLAGS[:core],
+        dunno: "-march=native",
+      }.freeze
+
+      def optimization_flags
+        LINUX_OPTIMIZATION_FLAGS
+      end
+
       def universal_archs
         [].extend ArchitectureListExtension
       end


Library/Homebrew/extend/os/linux/hardware/cpu.rb
```diff
@@ -1,6 +1,16 @@
 module Hardware
   class CPU
     class << self
+      OPTIMIZATION_FLAGS_LINUX = {
```
MikeMcQuaid: `LINUX_OPTIMIZATION_FLAGS` would be more consistent with `PPC_OPTIMIZATION_FLAGS`. I wonder if it's worth just overriding the `dunno` in OPTIMIZATION_FLAGS instead? The other values look good?

sjackman: I'd like to remove `:penryn` from `OPTIMIZATION_FLAGS`. I'd prefer that it use `-march=native` like the others rather than `-march=core2 -msse4.1`.

sjackman: > LINUX_OPTIMIZATION_FLAGS would be more consistent with PPC_OPTIMIZATION_FLAGS.

:+1:

> I wonder if it's worth just overriding the dunno in OPTIMIZATION_FLAGS instead? The other values look good?

I've updated the PR to get the values for `LINUX_OPTIMIZATION_FLAGS` from `OPTIMIZATION_FLAGS`.

MikeMcQuaid:
@sjackman Just for future reference I meant `OPTIMIZATION_FLAGS.merge({dunno:"-march=native"})`. I've actually just set this on the generic class instead in https://github.com/Homebrew/brew/pull/2312 as I think it makes sense as a global default. Just FYI if you make your branches pushable I'd have pushed it there for future reference.

sjackman:
> Just for future reference I meant `OPTIMIZATION_FLAGS.merge({dunno:"-march=native"})`

Yes, I thought so. You may have missed this reply from me: https://github.com/Homebrew/brew/pull/2308#discussion_r105475178

> I'd like to remove `:penryn` from `OPTIMIZATION_FLAGS`. I'd prefer that it use `-march=native` like the others rather than `-march=core2 -msse4.1`.

I've opened PR https://github.com/Homebrew/brew/pull/2336 to this effect.

> Just FYI if you make your branches pushable I'd have pushed it there for future reference.

Will do. Thanks for the tip, Mike.

PR #2309, opened 2017-03-10
add command 'cask outdated'

axiac:

-----

Disclaimer for the unchecked item
----------------------------------
A couple of weeks ago I was missing the `brew cask outdated` command. I checked the Issues and PRs on the Homebrew/brew repo and I couldn't find anything related so I decided to give it a try.
When I was ready to make a pull request I discovered PR #2250 was already created two days before and I learned about issue caskroom/homebrew-cask#29301. I went back to coding, implemented the `--greedy` option too and now I have a complete PR for the `cask outdated` feature.

How it works:
-------------
When the `--greedy` options is not present in the command line, a Cask that has `auto_updates true` or `version 'latest'` is ignored, as required.

When `--greedy` is present, the value of `auto_updates` is ignored and the Cask is processed as explained below; the casks that have `version 'latest'` are always displayed as outdated.

The version of the Cask from tap is compared against the currently installed version of the application (`versions.last`). The Cask is outdated if the currently installed version is different than the version from the tap. In this case, all the local versions of the application are compared to the tap version and all that are different are reported as "outdated".

If the current installed version is the same as the tap version, the Cask is not reported as outdated, even if there are other (different) local versions.

Output formatting
------------------
The format of the command is the same as the output of `brew outdated`, including the processing of `--verbose` and `--quiet` command line options.

Other points of interest
-----------------------
The new functionality is accompanied by tests. The documentation (the `man` page) was updated.


Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,47 @@ def installed?
...
+        return [] if auto_updates
+        return [] if version.latest?
+      end
+
+      # the current installed version
+      installed = versions
+      current   = installed.last
+
+      # special case: tap version is 'latest' or current version is 'latest'
+      # changing from :latest to versioned or vice-versa probably never happens
```
vitorgalvao: It does happen occasionally.

vitorgalvao:
> The Cask is outdated if the currently installed version is smaller than the version from the tap.

You can’t compare “size”, because versions are non-standard, for us. A version can just as well be a number or an intelligible string. The only comparison that can be made is “is it different?”.

ilovezfs:
I think sortable versions might also be a prerequisite for this making any sense.

axiac:
You both are right regarding the version comparison. I had semantic versioning in mind but I overlooked that, as strings, "10" is smaller than "9". And semantic versioning is a foolish assumption in this case, anyway.

Simplified the version comparison in a new commit.

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
```diff
@@ -0,0 +1,39 @@
+module Hbc
+  class CLI
+    class Outdated < Base
+      def self.run(*args)
+        greedy  = args.include?("--greedy")
+        verbose = ($stdout.tty? || ARGV.verbose?) && !args.include?("--quiet")
```
reitermarkus: `ARGV.verbose?` -> `verbose?`

axiac: Are you sure? It says:  ```Error: undefined method `verbose?' for Hbc::CLI::Outdated:Class```

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,45 @@ def installed?
...
+
+      # if version is "latest" then all installed version are outdated (greedy)
+      return installed if version.latest?
+
+      # not outdated unless there is a different version on tap
+      return [] unless current != version
+
+      # current installed version different than tap version
+      # collect all installed versions that are different than tap version and return them
+      installed.map { |v| v if v != version.to_s }.compact
```
reitermarkus: Use `select` here.

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
reitermarkus: Ah, then it's `CLI.verbose?`, I thought it would use the `CLI` namespace.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,45 @@ def installed?
       !versions.empty?
     end
 
+    def outdated?(greedy = false)
```
reitermarkus: This whole method is basically `!outdated_versions.empty?`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,45 @@ def installed?
...
+      greedy = true if greedy
+      @outdated ||= {}
+      @outdated[greedy] ||= _check_outdated(greedy)
+      !@outdated[greedy].empty?
+    end
+
+    def outdated_versions(greedy = false)
+      greedy = true if greedy
+      @outdated ||= {}
+      @outdated[greedy] ||= _check_outdated(greedy)
```
reitermarkus: Let's not cache this. At the moment we are only using it once anyways, but even if we were to use it multiple times, something will probably have changed between these calls.

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+
+      def self.list_if_outdated(cask, greedy, verbose)
+        return unless cask.outdated?(greedy)
+
+        if verbose
+          outdated_versions = cask.outdated_versions(greedy)
+          outdated_info   = "#{cask.token} (#{outdated_versions.join(", ")})"
+          current_version = cask.version.to_s
+          puts "#{outdated_info} < #{current_version}"
```
reitermarkus: Use `!=` here instead of `<`.

axiac: This is the output format used by `brew outdated` and it is also the one suggested in caskroom/homebrew-cask#29301

Library/Homebrew/cask/lib/hbc/cask.rb
axiac: `#select` is the function I was looking for (but it seems I was too lazy to read the complete list of `Array` methods and I concluded too early). Thank you!

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
reitermarkus: See the discussion here: https://github.com/Homebrew/brew/pull/2250#discussion_r104305304

Library/Homebrew/cask/lib/hbc/cask.rb
axiac: `#outdated?` and `outdated_versions` are both called when `--verbose`. This is why the information is cached (`versions` is expensive).

Is it ok to (drop caching and the `#outdated_versions` method and) let `#outdated?` return a boolean **and** the list of outdated versions?

Library/Homebrew/cask/lib/hbc/cli/outdated.rb
axiac: I understood the reasons about using `!=` to compare the versions and I changed the code. 

Please remark the `puts` in front of that line. It is not a version comparison, `"<"` is part of the text that is displayed to the user:

    $ brew cask outdated
    cyberduck (5.3.7.23440, 5.3.8.23611) < 5.3.9.23624
    virtualbox (5.1.14-112924) < 5.1.16-113841
    virtualbox-extension-pack (5.1.14-112924) < 5.1.16-113841

reitermarkus: Yes, that is exactly the reason. Displaying `<` when we are only comparing with `!=` doesn't make sense.

axiac: I got your point now. But I still wonder what kind of magic allows `brew outdated` compare the Formula versions using `<`.

reitermarkus: This is because Formulae use the comparable `PkgVersion` class, and `Hbc::DSL::Version` is a subclass of `String`, so it only performs a String comparison.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,35 @@ def installed?
...
+
+      # keep the installed versions in a local variable because a call to :versions is expensive
+      installed = versions
+      current   = installed.last
+
+      # if version is "latest" then all installed version are outdated (greedy)
+      return installed if version.latest?
+
+      # not outdated unless there is a different version on tap
+      return [] unless current != version
```
reitermarkus: Make this `if version == current`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,35 @@ def installed?
...
+    def outdated?(greedy = false)
+      !outdated_versions(greedy).empty?
+    end
+
+    def outdated_versions(greedy = false)
+      # special case: tap version is not available
+      return [] if version.nil?
+
+      # if --greedy is not present in the command line then skip when detection is not safe
+      unless greedy
```
reitermarkus: This should be:

```ruby
      if greedy
        return versions if version.latest?
      else
        return [] if auto_updates
      end
```

Otherwise, when I have the `:latest` version installed, but there is an updated cask with version `1.0.0`, it will not show up as outdated.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,35 @@ def installed?
...
+        return [] if auto_updates
+        return [] if version.latest?
+      end
+
+      # keep the installed versions in a local variable because a call to :versions is expensive
+      installed = versions
+      current   = installed.last
+
+      # if version is "latest" then all installed version are outdated (greedy)
+      return installed if version.latest?
```
reitermarkus: This must be removed after changing the above to `if greedy`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,35 @@ def installed?
...
+      # special case: tap version is not available
+      return [] if version.nil?
+
+      # if --greedy is not present in the command line then skip when detection is not safe
+      unless greedy
+        return [] if auto_updates
+        return [] if version.latest?
+      end
+
+      # keep the installed versions in a local variable because a call to :versions is expensive
```
reitermarkus: Remove this comment.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,35 @@ def installed?
...
+      installed = versions
+      current   = installed.last
+
+      # if version is "latest" then all installed version are outdated (greedy)
+      return installed if version.latest?
+
+      # not outdated unless there is a different version on tap
+      return [] unless current != version
+
+      # current installed version different than tap version
```
reitermarkus: Remove this comment.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -85,6 +85,35 @@ def installed?
...
+      # collect all installed versions that are different than tap version and return them
+      installed.select { |v| v if v != version.to_s }
```
reitermarkus: No need for the `if` here, `version != v` is enough.

reitermarkus: Or even `select(&version.method(:!=))`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -84,6 +84,30 @@ def installed?
...
+      end
+
+      installed = versions
+      current   = installed.last
+
+      # not outdated unless there is a different version on tap
+      return [] if current == version
+
+      # collect all installed versions that are different than tap version and return them
+      installed.select(&version.method(:!=))
```
reitermarkus: Ok, I changed my mind, I think `select { |v| version != v }` is clearer.

Library/Homebrew/test/cask/cask_spec.rb
```diff
@@ -89,4 +89,89 @@
       expect(c.metadata_versioned_container_path.to_s).to eq(metadata_path.to_s)
     end
   end
+
+  describe "outdated" do
+    it "ignores the Casks that have auto_updates true (without --greedy)" do
+      c = Hbc.load("auto-updates")
+      expect(c.outdated?).to be_falsey
```
reitermarkus: Make this `expect(c).to be_outdated`.

Library/Homebrew/test/cask/cli/outdated_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  before do
+    shutup do
+      cask = Hbc::CaskLoader.load_from_file(TEST_FIXTURE_DIR/"cask/Casks/basic-cask.rb")
+      InstallHelper.install_with_caskfile(cask)
+      cask = Hbc::CaskLoader.load_from_file(TEST_FIXTURE_DIR/"cask/Casks/outdated/local-caffeine.rb")
+      InstallHelper.install_with_caskfile(cask)
+      cask = Hbc::CaskLoader.load_from_file(TEST_FIXTURE_DIR/"cask/Casks/outdated/local-transmission.rb")
+      InstallHelper.install_with_caskfile(cask)
+    end
+    Hbc::CLI.verbose = true
```
reitermarkus: Instead of this, use `allow(Hbc::CLI).to receive(:verbose?).and_return(true)`.

Library/Homebrew/test/cask/cli/outdated_spec.rb
```diff
@@ -0,0 +1,88 @@
+describe Hbc::CLI::Outdated, :cask do
+  before do
+    shutup do
+      cask = Hbc::CaskLoader.load_from_file(TEST_FIXTURE_DIR/"cask/Casks/basic-cask.rb")
```
reitermarkus: Define these like `let(:basic_cask) { Hbc::CaskLoader.load_from_file(TEST_FIXTURE_DIR/"cask/Casks/basic-cask.rb") }` above the `before` block.

Library/Homebrew/test/cask/cli/outdated_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      InstallHelper.install_with_caskfile(cask)
+      cask = Hbc::CaskLoader.load_from_file(TEST_FIXTURE_DIR/"cask/Casks/outdated/local-transmission.rb")
+      InstallHelper.install_with_caskfile(cask)
+    end
+    Hbc::CLI.verbose = true
+  end
+
+  after do
+    shutup do
+      Hbc::CLI::Uninstall.run("basic-cask", "local-caffeine", "local-transmission")
```
reitermarkus: I think these are automatically removed after each test.

Library/Homebrew/cask/lib/hbc/cask.rb
axiac: It is clearer, indeed.

Library/Homebrew/test/cask/cask_spec.rb
axiac: It's `expect(c).to_not be_outdated`. Got it.

Library/Homebrew/test/cask/cli/outdated_spec.rb
axiac: This eliminates the need to reset `CLI.verbose` in `after`, isn't it?

Library/Homebrew/test/cask/cask_spec.rb
reitermarkus: `.not_to`

Library/Homebrew/test/cask/cli/outdated_spec.rb
reitermarkus: Exactly.

joshka:
Heya, I'm disappointed that we ended up duplicating effort, but this looks much neater than my PR (#2250). I also learnt a bunch of ruby and rspec reading your version. Thanks :)

Library/Homebrew/test/cask/cli/outdated_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        Hbc::CLI::Outdated.run("--greedy")
+      }.to output(<<-EOS.undent).to_stdout
+        local-caffeine (1.2.2) != 1.2.3
+        local-transmission (2.60) != 2.61
+        version-latest-string (latest) != latest
+      EOS
+    end
+  end
+
+  it "provides help" do
```
reitermarkus: No need to test this, since this is basically the same for every command and I think this is already tested elsewhere.

axiac:
@joshka Thank you for your appreciation. I'm sorry my PR rendered your work useless, it was not my intention. You can still implement `cask upgrade` (the second part of issue caskroom/homebrew-cask#29301), I promise I won't do it. I don't feel the need for it; `cask reinstall` does (almost) the same thing for outdated casks.

Library/Homebrew/test/cask/cli/outdated_spec.rb
reitermarkus: > @reitermarkus you want to spoil my perfect code coverage ;-)

Fixed that for you, by adding a test for the `help` methods. 😄

reitermarkus:
Thanks @axiac and @joshka for working on this!

PR #2321, opened 2017-03-11
Add Python 3 path in brew config

iMichka:
This is useful when debugging user issues for
formulae using Python 3.

-----

### DIFF
--- a/Library/Homebrew/system_config.rb
+++ b/Library/Homebrew/system_config.rb
@@ -76,8 +76,8 @@ class SystemConfig
       describe_path(which("perl"))
     end
 
-    def describe_python
-      python = which "python"
+    def describe_python(version)
+      python = which "python#{version}"
       return "N/A" if python.nil?
       python_binary = Utils.popen_read python, "-c", "import sys; sys.stdout.write(sys.executable)"
       python_binary = Pathname.new(python_binary).realpath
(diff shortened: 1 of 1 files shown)


tdsmith:
This won't work for system Python since Apple doesn't ship a `python2` executable. It does ship a `python2.7`. Understanding where the naked `python` executable is pointing is also interesting, though.

MikeMcQuaid:
I think we could end up adding a bunch of stuff in here. I'd rather see a more generic solution so am 👎 on this I'm afraid.

ilovezfs:
Can we print the path to COBOL?

MikeMcQuaid:
Sorry, passing on this for now.

iMichka:
No problem. My ruby skills / knowledge of homebrew's internals are not yet good enough to provide a more generic solution here. And I understand that that list could grow, which may or may not be a good idea. So I'm OK on passing on this :).

Still, there is a mix between system specific stuff and other things that seem unrelated: Python2 , java, x11? So I thought: why not add Python3. @tdsmith is right that this output's only a part of the picture at the moment, at least for Python.

Maybe `brew doctor` is better at warning us when weird stuff is going on with the user's setup, so extending the `brew config` list is not the best idea.

MikeMcQuaid:
> Still, there is a mix between system specific stuff and other things that seem unrelated: Python2 , java, x11?

All that stuff was at one time provided by macOS FWIW 😁

PR #2324, opened 2017-03-12
keg: Skip relocation when dylib ID/install name doesn't change.

woodruffw:

-----

Our current relocation behavior is harmless, but also useless when the load command being rewritten doesn't actually change. 

cc @ilovezfs

### DIFF
--- a/Library/Homebrew/os/mac/keg.rb
+++ b/Library/Homebrew/os/mac/keg.rb
@@ -1,5 +1,6 @@
 class Keg
   def change_dylib_id(id, file)
+    return if file.dylib_id == id
     @require_relocation = true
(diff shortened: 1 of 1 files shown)


Library/Homebrew/os/mac/keg.rb
```diff
@@ -1,5 +1,6 @@
 class Keg
   def change_dylib_id(id, file)
+    return if file.dylib_id == id
```
ilovezfs: I assume we know these are strings not pathnames? Ditto below.

woodruffw: Yep, we do. `dylib_id` comes from `macho.dylib_id`, which is a string, and `id` comes from `dylib_id_for`, which converts a Pathname into a string.

PR #2334, opened 2017-03-13
enforce *.sourceforge.io and more for urls

vszakats:

-----

Add more URL checks to audit, following recent series of formula updates.

Alert will now go off for these URLs:
```
(http|git)://git.(savannah|sv).gnu.org/
(http|svn)://svn.(savannah|sv).gnu.org/
http://hg.(savannah|sv).gnu.org/
(http|git)://git.code.sf.net/
(http|svn)://svn.code.sf.net/
(http|git)://*.git.sourceforge.net/gitroot/
(http|svn)://*.svn.sourceforge.net/svnroot/
```

Plus, existing `*.sourceforge.io` homepage rule has been extended to `url` as well.
These will be redirected to insecure `.net` page unless the SourceForge project
maintainer enables HTTPS for these pages explicitly.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -1431,7 +1431,9 @@ class ResourceAuditor
            %r{^http://hackage\.haskell\.org/},
            %r{^http://(?:[^/]*\.)?archive\.org},
            %r{^http://(?:[^/]*\.)?freedesktop\.org},
-           %r{^http://(?:[^/]*\.)?mirrorservice\.org/}
+           %r{^http://(?:[^/]*\.)?mirrorservice\.org/},
+           %r{^(http|git)://git\.code\.sf\.net/},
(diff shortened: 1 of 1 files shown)


ilovezfs:
The svn https ones should be disabled for < Sierra so we don't get false positives. Alternatively, we could automatically degrade svn https to http on < Sierra such that those URLs can lie on < Sierra and claim to be https when they're not. The latter is arguably unethical since we'd be claiming in the formula to be using secure URLs and then actually not doing so.

vszakats:
You're right, I forgot about it. How to translate that rule to an audit check? To avoid all this messy ambiguity throughout each formula and audit rules, I'd prefer the downgrade approach. If this fact is clearly articulated towards the user, I think it should be fine.

ilovezfs:
Yeah, "warning: degrading https to http since blah" would take care of that concern, I think.

vszakats:
Closing this, probably won't have the required time/knowledge to implement it properly.

PR #2335, opened 2017-03-13
add audit rule to enforce slash after domain name

vszakats:

-----

Currently there is random variations of the same URLs (or the same group of URLs) in this regard, some of them have an ending slash added, some of them don't. While it's certainly not the end of the world, it doesn't look neat and presents an extra hurdle when searching/grepping/parsing this field. It's also general good practice to add it and may help servers to return the correct page on the first request.

Example:
```
$ grep 'sourceforge.io\"' *.rb | wc -l
      73
$ grep 'sourceforge.io/\"' *.rb | wc -l
     286
```

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -613,6 +613,10 @@ class FormulaAuditor
       problem "The homepage should start with http or https (URL is #{homepage})."
     end
 
+    unless homepage =~ %r{^(.+)://(.*)/(.*)}
+      problem "#{homepage} must have a slash after the domain name part."
+    end
+
     # Check for http:// GitHub homepage urls, https:// is preferred.
     # Note: only check homepages that are repo pages, not *.github.com hosts
     if homepage.start_with? "http://github.com/"
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -613,7 +613,7 @@ class FormulaAuditor
       problem "The homepage should start with http or https (URL is #{homepage})."
     end
 
-    unless homepage =~ %r{^(.+)://(.*)/(.*)}
+    if homepage =~ %r{^.+://[^/]+$}
       problem "#{homepage} must have a slash after the domain name part."
     end
 
--- a/Library/Homebrew/test/dev-cmd/audit_spec.rb
+++ b/Library/Homebrew/test/dev-cmd/audit_spec.rb
@@ -465,6 +465,8 @@ describe FormulaAuditor do
         "sf3" => "http://foo.sf.net/",
         "sf4" => "http://foo.sourceforge.io/",
         "waldo" => "http://www.gnu.org/waldo",
+        "sl1" => "http://example.com",
+        "sl2" => "http://example.com#section",
       }
 
       formula_homepages.each do |name, homepage|
@@ -476,7 +478,10 @@ describe FormulaAuditor do
         EOS
 
         fa.audit_homepage
-        if homepage =~ %r{http:\/\/www\.freedesktop\.org}
+        if name =~ /(sf2|sl)/
+          expect(fa.problems.first)
+            .to match("#{homepage} must have a slash after the domain name part.")
+        elsif homepage =~ %r{http:\/\/www\.freedesktop\.org}
           if homepage =~ /Software/
             expect(fa.problems.first).to match(
               "#{homepage} should be styled " \


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -613,6 +613,10 @@ def audit_homepage
       problem "The homepage should start with http or https (URL is #{homepage})."
     end
 
+    unless homepage =~ %r{^(.+)://(.*)/(.*)}
+      problem "#{homepage} must have a slash after the domain name part."
```
reitermarkus: Since this only applies to URLs without a path component, I think the message could be a bit simpler, like `#{homepage} has to end with a slash.`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -613,6 +613,10 @@ def audit_homepage
       problem "The homepage should start with http or https (URL is #{homepage})."
     end
 
+    unless homepage =~ %r{^(.+)://(.*)/(.*)}
```
reitermarkus: `if homepage =~ %r{^(.+)://[^/]+}`, maybe?

Library/Homebrew/dev-cmd/audit.rb
vszakats: I agree, but I had URLs like `https://example.com#anchor` or `https://exampe.com?page=100` in mind. These are valid and missing a slash, but not from the end.

Library/Homebrew/dev-cmd/audit.rb
vszakats: Looks good!

vszakats: Hm, it fails with this basic case: `http://example.com/`

Library/Homebrew/dev-cmd/audit.rb
reitermarkus: You're right, I didn't think of these possibilities.

Library/Homebrew/dev-cmd/audit.rb
reitermarkus: Ah, right, it should still be `unless`, or `!~`.

vszakats: Or maybe this?: `if homepage =~ %r{^.+://[^/]+$}`

vszakats: It's was the missing negation that's causing problems, but the basic case don't have extra characters after the slash, so it fails to match.
Ah okay, the edit version should work.

woodruffw:
Personally :-1: on this; I'm not sure why it's important for us to enforce this level of uniformity. Checking for HTTPS is good, but after that I don't think we should really care if it resolves. 

As for parsing/grepping: Lots of the `brew` commands have the ability to emit JSON, and we have a public Ruby API (e.g., `Formula["gcc"]`). Being able to `grep` for something quickly is nice, but it'll never be as rigorous as using `jq` or another tool to extract the relevant information from, say, `brew info --all --json=v1`.

vszakats:
It's good to know Brew can emit JSON, didn't about know it. Anyhow it was not my major concern, after either `jq` or `grep`, the URL itself will need to be sanitized to be able to work with, unless you're lucky to have a proper URL parser at hand (atypical in shell/bash scripts).

All in all though, it's one small thing that helps things *look* and *work* great, while also following known best practice. I think it also fits nicely into the evolution of formulae, where they became gradually more and more organized.

It also looks like this is already enforced/endorsed for [Casks](https://github.com/caskroom/homebrew-cask/commit/50a3c4aaa2ef854d0b3cf16d5514adaf6310ed38) [already](https://github.com/caskroom/homebrew-cask/commit/c575df576c03b37d7fc44caa1eceba053ab64caf).

Some further read about this: https://stackoverflow.com/questions/5948659/when-should-i-use-a-trailing-slash-in-my-url

[ Please note I'm talking about **bare domains** in this context. It's not the goal to add an ending slash to every URL, which would altogether be incorrect. ]

woodruffw:
> Anyhow it was not my major concern, after either jq or grep, the URL itself will need to be sanitized to be able to work with, unless you're lucky to have a proper URL parser at hand (atypical in shell/bash scripts)

In principle, the URL is already santitized in the sense that `audit` has checked it and resolved it. There are other checks that we can perform, but they're mostly periodic (checking to see if the project's website is still online, checking whether their HTTPS cert is going to expire soon, etc).

> it's one small thing that helps things look and work great, while also following known best practice

I agree that consistency is good. However, I'm not convinced that there *is* a best practice here (at the domain level, there definitely is a semantic difference at the path level). 

Ultimately, my fear is that this sort of change will lead to a lot of busywork in formulae PRs - we already spend a lot of time helping new contributors fix simple `audit` mistakes, and having to remind users to remove a single character just adds to that busywork.

vszakats:
Maybe the busywork-effect can be reduced by outputting the correct URL and/or by ignoring the issue if the bare domain contains any extra anchor/parameter (which - unless abused - appears to be a rather theoretical corner-case anyway.)

I'm not arguing that current URLs are incorrect, but to parse them, they need to be doctored. And they hurt the eye of course :)

[As a moon-shot solution, there could be a "formula-doctor-bot" that cycles through formulae and commits such fixes automatically, but this would probably take a larger collective time/effort than the current audit-time corrections spread over contributors. And it would create undesired commit-churn.]

Well anyhow, I'd appreciate if maintainers could reach a conclusion to either close this quickly or massage it enough to be acceptable. I'd be quite happy if URLs wouldn't just work great, but look/parse great too.

woodruffw:
Yeah, there are definitely steps that could be taken towards automating away a lot of the busywork. I think automatic reviews/corrections have been proposed before, although I'm not sure if anybody actually started work on them.

> I'm not arguing that current URLs are incorrect, but to parse them, they need to be doctored.

I'm not 100% sure what you mean by this, since a URL should be parseable in either format:

```ruby
URI("https://example.com").host # => "example.com"
URI("https://example.com/").host # => "example.com"
```

The only real difference is that the first one has a `nil` path, while the second has a `/` path. That might lead to a bit of checking when consuming URLs, but both are valid in every normal sense (and shouldn't be any semantically different in terms of the WWW).

vszakats:
I meant pretty much what you described. There _is_ a required step to "tame" the URL to work with it. It's doable and as I started the PR, it's not an end-of-the-world issue, just one small nicety, like the fixed order of stanzas (if I get the terminology right) and several other existing similar niceties.

I understand this is one more such, but I feel it fits the line of existing ones. Of course like all existing ones this should be weighted and accepted/discarded. My point is that it'd make formulae overall nicer with the occasional price of an extra keypress to make audit happy.

reitermarkus:
> Maybe the busywork-effect can be reduced by outputting the correct URL

If you're up for it, you could try converting this to a RuboCop check with `--auto-correct` support, which would be even better than just outputting it.

vszakats:
@reitermarkus Sounds like a great idea. I completely missed the `--auto-correct` feature so far, so I'm going to do a little bit of reading on that.

(In the meantime any hint/advice is welcome.)

woodruffw:
> There is a required step to "tame" the URL to work with it.

What do you mean by "work with it"? Like I said before, no (sane) HTTP client makes a semantic distinction between being told to fetch `domain.tld` and `domain.tld/` - both result in a `GET` on the root, which the HTTPd needs to interpret and redirect.

Could you give me an example of a use case?

vszakats:
Here's one example: https://github.com/vszakats/https-check/blob/master/chkhttps.sh
(I don't mean that as a argument for clean URLs, because writing tools like above is admittedly on the extreme niche. But an example nevertheless.)

reitermarkus:
> It also looks like this is already enforced/endorsed for Casks already.

Since this wasn't already enforced for Casks, I implemented it: https://github.com/caskroom/rubocop-cask/pull/48

> (In the meantime any hint/advice is welcome.)

This should help as a starting point.

vszakats:
@reitermarkus That's most cool. Admittedly it's way over my Ruby/RuboCop skill level, but just exactly the same would be superb for formulae. I'll explore this tomorrow with a fresher mind (= :coffee:).

woodruffw:
> Here's one example: https://github.com/vszakats/https-check/blob/master/chkhttps.sh

I think that's a reasonable use case, but that it would be *much* better suited to the JSON interface we provide + a URI/URL parsing module + your actual HTTPS testing code. As it currently stands, I don't think it makes much sense to shoulder this in `audit`/increased maintainer time (although that could change if we ever get automated reviews working :smile:) in exchange for some relative convenience in a relatively rare use case.

cc @Homebrew/maintainers for thoughts on this, since I don't want to reject it unilaterally.

ilovezfs:
>Ultimately, my fear is that this sort of change will lead to a lot of busywork in formulae PRs - we already spend a lot of time helping new contributors fix simple audit mistakes, and having to remind users to remove a single character just adds to that busywork.

This.

MikeMcQuaid:
> If you're up for it, you could try converting this to a RuboCop check with --auto-correct support, which would be even better than just outputting it.

Agreed that this is the way to go for these checks in future. Semantic pedantry should be autofixable.

Thanks anyway for the PR @vszakats!

PR #2337, opened 2017-03-14
Add extend/os/linux/system_config.rb

sjackman:
Report the Linux kernel, OS, and gcc versions.

### DIFF
--- a/Library/Homebrew/extend/os/linux/system_config.rb
+++ b/Library/Homebrew/extend/os/linux/system_config.rb
@@ -1,0 +1,42 @@
+require "formula"
+
+class SystemConfig
+  class << self
+    def host_os_version
+      if which("lsb_release")
+        description = `lsb_release -d`.chomp.sub("Description:\t", "")
+        codename = `lsb_release -c`.chomp.sub("Codename:\t", "")
+        "#{description} (#{codename})"
+      elsif (redhat_release = Pathname.new("/etc/redhat-release")).readable?
+        redhat_release.read.chomp
+      else
+        "N/A"
+      end
+    end
+
+    def host_gcc_version
+      gcc = Pathname.new "/usr/bin/gcc"
+      return "N/A" unless gcc.executable?
+      `#{gcc} --version 2>/dev/null`[/ (\d+\.\d+\.\d+)/, 1]
+    end
+
+    def formula_version(formula)
+      return "N/A" unless CoreTap.instance.installed?
+      f = Formulary.factory formula
+      return "N/A" unless f.installed?
+      f.version
+    rescue FormulaUnavailableError
+      return "N/A"
+    end
+
+    def dump_verbose_config(out = $stdout)
+      dump_generic_verbose_config(out)
+      out.puts "Kernel: #{`uname -mors`.chomp}"
+      out.puts "OS: #{host_os_version}"
+      out.puts "/usr/bin/gcc: #{host_gcc_version}"
+      ["glibc", "gcc", "xorg"].each do |f|
+        out.puts "#{f}: #{formula_version f}"
+      end
+    end
+  end
+end
--- a/Library/Homebrew/extend/os/system_config.rb
+++ b/Library/Homebrew/extend/os/system_config.rb
@@ -1,2 +1,7 @@
 require "system_config"
-require "extend/os/mac/system_config" if OS.mac?
+
+if OS.mac?
+  require "extend/os/mac/system_config"
+elsif OS.linux?
+  require "extend/os/linux/system_config"
+end


Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def host_gcc_version
+      gcc = Pathname.new "/usr/bin/gcc"
+      `#{gcc} --version 2>/dev/null`[/ (\d+\.\d+\.\d+)/, 1] if gcc.exist?
+    end
+
+    def formula_version(formula)
+      f = Formula[formula]
+      f.installed? ? f.version : "N/A"
```
MikeMcQuaid:
```
return "N/A" unless f.installed?
f.version
```

Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -0,0 +1,39 @@
...
+    rescue FormulaUnavailableError
+      return "N/A"
+    end
+
+    def dump_verbose_config(f = $stdout)
+      dump_generic_verbose_config(f)
+      f.puts "Kernel: #{`uname -mors`.chomp}"
+      f.puts "OS: #{host_os_version}"
+      f.puts "OS gcc: #{host_gcc_version}"
```
MikeMcQuaid: Maybe `/usr/bin/gcc` would be more obvious?

Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -0,0 +1,39 @@
...
+      f = Formula[formula]
+      f.installed? ? f.version : "N/A"
+    rescue FormulaUnavailableError
+      return "N/A"
+    end
+
+    def dump_verbose_config(f = $stdout)
+      dump_generic_verbose_config(f)
+      f.puts "Kernel: #{`uname -mors`.chomp}"
+      f.puts "OS: #{host_os_version}"
```
MikeMcQuaid: `OS version`?

Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -0,0 +1,39 @@
...
+      f.puts "OS gcc: #{host_gcc_version}"
+      f.puts "Linuxbrew glibc: #{formula_version "glibc"}"
```
MikeMcQuaid: `glibc installed formula` or something?

MikeMcQuaid: Also I believe it's `Glibc`

Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -0,0 +1,39 @@
...
+      f.puts "Linuxbrew gcc: #{formula_version "gcc"}"
```
MikeMcQuaid: `GCC`?

Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -0,0 +1,39 @@
...
+      f.puts "Linuxbrew xorg: #{formula_version "linuxbrew/xorg/xorg"}"
```
MikeMcQuaid: `Xorg`?

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: I'm mimicking the output of `brew config` on macOS where it reads:
```
macOS: 10.11.6-x86_64
```
so I went for
```
OS: Ubuntu 14.04.5 LTS (trusty)
```
I feel the "version" is implied.

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: Agreed.

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: I prefer using the name of the formula `gcc` as opposed to the upstream name.

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: I prefer to use the name of the formula `glibc` as opposed to the upstream name.

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: How about `homebrew/core/gcc` ?

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: How about `How about `linuxbrew/xorg/xorg` ?

Library/Homebrew/extend/os/linux/system_config.rb
MikeMcQuaid: How about just `glibc`?

Library/Homebrew/extend/os/linux/system_config.rb
MikeMcQuaid: How about just `xorg`?

Library/Homebrew/extend/os/linux/system_config.rb
MikeMcQuaid: How about just `gcc`?

MikeMcQuaid:
Trying to think of a way of not using `Linuxbrew` in these names as Homebrew on Linux doesn't imply Linuxbrew (yet, at least) and these technically may use any tap's formula.

sjackman:
Thanks for the suggestions, Mike. I've updated the PR.

MikeMcQuaid:
CI failing; may need a more aggressive `rescue`: https://travis-ci.org/Homebrew/brew/jobs/211025276#L1036

sjackman:
The error is
```
Error: Invalid formula: /home/travis/build/Homebrew/brew/Library/Taps/homebrew/homebrew-core/Formula/media-info.rb
uninitialized constant Formulary::FormulaNamespacea2b7870269c6d817b9989c0f5561f49f::GobjectIntrospection::MacOS
```
https://travis-ci.org/Homebrew/brew/jobs/211025276#L296
caused by for example
```ruby
depends_on "curl" if MacOS.version < :snow_leopard
```
because the constant MacOS is not defined when Homebrew is run on Linux.

The failing command is `brew config`. I'm guessing that `Formula["gcc"]` is causing `brew` to tap `Homebrew/core` and check the syntax validity of all those formula. If my understanding above is correct, options are…

- We could load the actual `MacOS` constant on Linux.
- We could load a stubbed out version of the `MacOS` constant that is just for Linux.
- We could somehow prevent `homebrew/core` from being loaded on Linux, by for example forcing the use of `NullLoader` for commands other than just `brew tests`.
- We could catch and ignore the `Invalid formula` exception in `brew config` alongside ignoring the `FormulaUnavailableError` exception.

Thoughts on which of these possible solutions is preferable?

sjackman:
I'm overcomplicating the situation I think. I only need to use an equivalent of `Formulary.factory` that does not auto-tap Homebrew/core if it's not yet tapped.

sjackman:
Fixed it (hopefully) with
```ruby
return "N/A" unless CoreTap.instance.installed?
```

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #2338, opened 2017-03-14
audit: online - limit download time based on HTTPS determination

JCount:

-----
Previously, the http_content_headers_and_checksum method always
downloaded the entire url, including headers. These downloads make
the audit more likely to hang or output false positives. The only
time the entire file is actually used is when verifying that the url
can use the HTTPS protocol instead of HTTP. Therefore, pass curl the
-I flag to only fetch headers unless the protocol is HTTP. If it is,
maintain the old behavior by passing --include to download both the
headers and the file.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -177,8 +177,9 @@ class FormulaAuditor
 
     details = nil
     user_agent = nil
+    hash_needed = url.start_with?("http:")
     user_agents.each do |ua|
-      details = http_content_headers_and_checksum(url, user_agent: ua)
+      details = http_content_headers_and_checksum(url, hash_needed: hash_needed, user_agent: ua)
       user_agent = ua
       break if details[:status].to_s.start_with?("2")
     end
@@ -188,11 +189,11 @@ class FormulaAuditor
       return "The URL #{url} is not reachable (HTTP status code #{details[:status]})"
     end
 
-    return unless url.start_with? "http:"
+    return unless hash_needed
 
     secure_url = url.sub "http", "https"
     secure_details =
-      http_content_headers_and_checksum(secure_url, user_agent: user_agent)
+      http_content_headers_and_checksum(secure_url, hash_needed: true, user_agent: user_agent)
 
     if !details[:status].to_s.start_with?("2") ||
        !secure_details[:status].to_s.start_with?("2")
@@ -210,9 +211,10 @@ class FormulaAuditor
     "The URL #{url} could use HTTPS rather than HTTP"
   end
 
-  def self.http_content_headers_and_checksum(url, user_agent: :default)
+  def self.http_content_headers_and_checksum(url, hash_needed: false, user_agent: :default)
+    max_time = hash_needed ? "600" : "25"
     args = curl_args(
-      extra_args: ["--connect-timeout", "15", "--include", url],
+      extra_args: ["--connect-timeout", "15", "--include", "--max-time", max_time, url],
       show_output: true,
       user_agent: user_agent,
     )
@@ -224,11 +226,13 @@ class FormulaAuditor
       status_code = headers[%r{HTTP\/.* (\d+)}, 1]
     end
 
+    output_hash = Digest::SHA256.digest(output) if hash_needed
+
     {
       status: status_code,
       etag: headers[%r{ETag: ([wW]\/)?"(([^"]|\\")*)"}, 2],
       content_length: headers[/Content-Length: (\d+)/, 1],
-      file_hash: Digest::SHA256.digest(output),
+      file_hash: output_hash,
     }
   end
 


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -210,9 +211,9 @@ def self.check_http_content(url, user_agents: [:default])
     "The URL #{url} could use HTTPS rather than HTTP"
   end
 
-  def self.http_content_headers_and_checksum(url, user_agent: :default)
+  def self.http_content_headers_and_checksum(url, hash_needed, user_agent: :default)
```
MikeMcQuaid: `hash_needed: false`; boolean arguments are a bit evil.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -210,9 +211,9 @@ def self.check_http_content(url, user_agents: [:default])
...
     args = curl_args(
-      extra_args: ["--connect-timeout", "15", "--include", url],
+      extra_args: ["--connect-timeout", "15", hash_needed ? "--include" : "-I", url],
```
MikeMcQuaid: Split this into a variable; the inline ternary operator is a bit hard to read.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -177,8 +177,9 @@ def self.check_http_content(url, user_agents: [:default])
 
     details = nil
     user_agent = nil
+    hash_needed = url.start_with?("http:") ? true : false
```
MikeMcQuaid: `hash_needed = url.start_with?("http:")` will do the same thing.

MikeMcQuaid:
> Previously, the http_content_headers_and_checksum method always
> downloaded the entire url, including headers. These downloads make
> the audit more likely to hang or output false positives. The only
> time the entire file is actually used is when verifying that the url
> can use the HTTPS protocol instead of HTTP. Therefore, pass curl the
> -I flag to only fetch headers unless the protocol is HTTP. If it is,
> maintain the old behavior by passing --include to download both the
> headers and the file.

🆒, makes sense. To avoid confusion I think the `details[:file_hash]` field shouldn't be set if the hash hasn't been queried.

Another option if you're feeling smart (for a future PR) would be to check the ETag rather than download hash if available. That would avoid the need for a full download.

Library/Homebrew/dev-cmd/audit.rb
JCount: Yes, I am not quite certain why I wrote it the way I did. 🤔

JCount:
We are in fact already parsing the ETag and checking them in addition to the hashes. Do you think that we should change the default behavior to simply comparing the ETags?

MikeMcQuaid:
> We are in fact already parsing the ETags and checking them in addition to the hashes. Do you think that we should change the default behavior to simply comparing the ETags?

I'm not sure. Perhaps if the ETag comparison either fails or an ETag isn't present? What do you think? What does @broder think?

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -210,9 +211,10 @@ def self.check_http_content(url, user_agents: [:default])
     "The URL #{url} could use HTTPS rather than HTTP"
   end
 
-  def self.http_content_headers_and_checksum(url, user_agent: :default)
+  def self.http_content_headers_and_checksum(url, hash_needed = false, user_agent: :default)
```
MikeMcQuaid: `hash_needed: false`

MikeMcQuaid:
> Adding another condition to further prevent the need to download the entire file seems reasonable, but it may be best addressed in a separate PR. That way, we can 🚢 this PR, hopefully improving matters, and then iterate if need be.

@JCount 👍 from me; 👎 to scope creep.

Library/Homebrew/dev-cmd/audit.rb
JCount: @MikeMcQuaid I originally tried that, but it causes `brew audit --online foo` to spit out:
```
Error: wrong number of arguments (2 for 1)
Please report this bug:
  http://docs.brew.sh/Troubleshooting.html
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:214:in `http_content_headers_and_checksum'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:182:in `block in check_http_content'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:181:in `each'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:181:in `check_http_content'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:1573:in `block in audit_urls'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:1565:in `each'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:1565:in `audit_urls'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:1314:in `audit'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:740:in `block in audit_specs'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:737:in `each'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:737:in `audit_specs'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:1259:in `audit'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:78:in `block in audit'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:74:in `each'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/audit.rb:74:in `audit'
/usr/local/Homebrew/Library/Homebrew/brew.rb:91:in `<main>'
```
Accidentally overlooking the scope creep, using `hash_needed = false` seemed to be the solution. However, I think I am overlooking something and cannot find an actual solution. Any guidance would be most welcome.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -210,9 +211,10 @@ def self.check_http_content(url, user_agents: [:default])
...
+    header_strategy = hash_needed ? "--include" : "-I"
```
broder: We should probably make "-I" into "--head" for clarity. Also, I'm not sure HEAD requests will work 100% of the time (e.g. on github). IIRC, the reason we're doing the full download every time is because it's more reliable.

broder:
> Perhaps if the ETag comparison either fails or an ETag isn't present?

The fallback to the hash of the file was mainly for the case when the ETag isn't present.

Library/Homebrew/dev-cmd/audit.rb
broder: You can also use `--range` but, again, I don't think it was reliable as some hosts didn't support it.

JCount: @broder Do you think it would be acceptable to simply use `--include --max-time 25` in the place of `--head` to provide more reliability? This would still eliminate the possibility of `audit --online` hanging on HTTPS urls, yet also address your concerns.

broder: It would almost certainly stop the hanging but wouldn't it change the results of `brew audit --online` (e.g. in the case of large files)? Is the hanging due to large files being downloaded or something else?

Library/Homebrew/dev-cmd/audit.rb
MikeMcQuaid: You'll need to use `hash_needed: false` in both the method definition and the callers of that method, too. Passing boolean arguments unless it's an accessor is generally not great because it's not as readable.

Library/Homebrew/dev-cmd/audit.rb
MikeMcQuaid: > We should probably make -I into --head for clarity.

Agreed.

MikeMcQuaid:
> The fallback to the hash of the file was mainly for the case when the ETag isn't present.

Ok, cool. In that case I think we can (in the next PR if necessary) check just for the ETag first and then fall back to a hash only if it isn't available (or perhaps it doesn't match).

ilovezfs:
Note that the hanging is not just about large files. It can get stuck and hang permanently on small or large files alike.

MikeMcQuaid:
> Note that the hanging is not just about large files. It can get stuck and hang permanently on small or large files alike.

Ok. A max time/timeout seems sensible regardless then.

Library/Homebrew/dev-cmd/audit.rb
JCount: Also note that this change in behavior only applies in situations where we were already only examining the headers and ignoring the rest of the output already.

Library/Homebrew/dev-cmd/audit.rb
JCount: In that case, how about I switch to simply passing a string object?

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -177,8 +177,9 @@ def self.check_http_content(url, user_agents: [:default])
 
     details = nil
     user_agent = nil
+    hash_needed = url.start_with?("http:")
     user_agents.each do |ua|
-      details = http_content_headers_and_checksum(url, user_agent: ua)
+      details = http_content_headers_and_checksum(url, hash_needed, user_agent: ua)
```
MikeMcQuaid: `details = http_content_headers_and_checksum(url, hash_needed: hash_needed, user_agent: ua)`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -188,11 +189,11 @@ def self.check_http_content(url, user_agents: [:default])
       return "The URL #{url} is not reachable (HTTP status code #{details[:status]})"
     end
 
-    return unless url.start_with? "http:"
+    return unless hash_needed
 
     secure_url = url.sub "http", "https"
     secure_details =
-      http_content_headers_and_checksum(secure_url, user_agent: user_agent)
+      http_content_headers_and_checksum(secure_url, true, user_agent: user_agent)
```
MikeMcQuaid: `http_content_headers_and_checksum(secure_url, hash_needed: true, user_agent: user_agent)`

Library/Homebrew/dev-cmd/audit.rb
MikeMcQuaid: `def self.http_content_headers_and_checksum(url, hash_needed: false, user_agent: :default)`

Library/Homebrew/dev-cmd/audit.rb
MikeMcQuaid: `header_strategy = hash_needed ? "--include" : "--head"`

PR #2343, opened 2017-03-14
gpg: add GnuPG series 2.1.x to supported versions

JCount:

-----
Now that the formula for GnuPG 2.1.x (`gnupg@2.1`) has been migrated to `homebrew/core`, extend the `gpg2_requirement` to support this branch as well. The actual substantive change is made in `gpg.rb` with the addition of "2.1" as a valid version.

CC @DomT4

### DIFF
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -5,7 +5,9 @@ class Gpg
     which_all(executable).detect do |gpg|
       gpg_short_version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
       next unless gpg_short_version
-      Version.create(gpg_short_version.to_s) == Version.create("2.0")
+      gpg_version = Version.create(gpg_short_version.to_s)
+      gpg_version == Version.create("2.0") ||
+        gpg_version == Version.create("2.1")
     end
   end
 
--- a/Library/Homebrew/requirements/gpg2_requirement.rb
+++ b/Library/Homebrew/requirements/gpg2_requirement.rb
@@ -7,6 +7,6 @@ class GPG2Requirement < Requirement
 
   # MacGPG2/GPGTools installs GnuPG 2.0.x as a vanilla `gpg` symlink
   # pointing to `gpg2`, as do we. Ensure we're actually using a 2.0 `gpg`.
-  # Temporarily, only support 2.0.x rather than the 2.1.x "modern" series.
+  # Support both the 2.0.x "stable" and 2.1.x "modern" series.
   satisfy(build_env: false) { Gpg.gpg2 || Gpg.gpg }
 end
--- a/Library/Homebrew/requirements/gpg2_requirement.rb
+++ b/Library/Homebrew/requirements/gpg2_requirement.rb
@@ -3,7 +3,7 @@ require "gpg"
 
 class GPG2Requirement < Requirement
   fatal true
-  default_formula "gnupg2"
+  default_formula "gnupg"
 
   # MacGPG2/GPGTools installs GnuPG 2.0.x as a vanilla `gpg` symlink
   # pointing to `gpg2`, as do we. Ensure we're actually using a 2.0 `gpg`.


DomT4:
Is everything in core today compatible with 2.1? If so the default formula should be moved to `gnupg` and things potentially rebuilt to check that and enforce the new requirement. If not the default formula needs to be moved to `gnupg@2.0`.

I worry that in future as things stand this part of my comment from the other PR is going to become a very real problem if Homebrew tries to support new & old at the same level of priority:
> Longer-term this also makes it necessary to rethink Homebrew's (incomplete because I'm a doofus) GPG verification, and Cask's (incomplete because nobody finished it) GPG verification if the plan there is to allow using existing keychains so people can still brew install stuff that needs/wants GPG verification offline, in the same way Homebrew technically supports you brew fetch-ing things in advance now and then brew install xyz whilst you're sat on a bus in the middle of nowhere.

But formula support for 2.1 won't immediately speed up the risk there.

JCount:
@DomT4 There is a reason that I left `default_formula "gnupg2"` as it was. I ran into the quite thorny problem you presented and could not commit to a course of action. This is partially due to the somewhat uncertain state of things in general. 

However, in the end, since both versions are now in core, I think that formulae should only use the requirement if they are compatible with both 2.0 and 2.1. If not, then it is more appropriate for them to specify the pertinent formula. 

Unfortunately, this reasoning and approach ignores a current use case, namely using the requirement to allow for the use of non-brew installations of gpg 2.0.x by formula. I think this is an issue that needs to be addressed before this (or something like it) was to be merged.

Any insight, opinions, commentary, criticism, derision, etc?

DomT4:
Let me think about it a bit. I'll try to pace out some thoughts this afternoon around this. My initial _(tentative)_ long-term thoughts around when `2.1.x` became `gnupg2` or `gnupg` was for Homebrew to drop all requirement/verification support for anything not the `2.1.x` branch but I expected this issue less soon than it has happened, which complicates that a touch.

Library/Homebrew/gpg.rb
```diff
@@ -5,7 +5,7 @@ def self.find_gpg(executable)
     which_all(executable).detect do |gpg|
       gpg_short_version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
       next unless gpg_short_version
-      Version.create(gpg_short_version.to_s) == Version.create("2.0")
+      Version.create(gpg_short_version.to_s) == Version.create("2.0") || Version.create("2.1")
```
MikeMcQuaid: I'd stick parentheses around this as the operator precedence is ambiguous. I'd also consider just doing `>= Version.create("2.0")`

JCount: Personally, I think it might be better to leave it hardcoded to check for the two specific version branches. Currently, and by deliberate design, this finds any executable of the specified version(s) in the PATH, not just ones installed by Homebrew. This is behavior we probably want to keep.

However, if it were simplified to `>= Version.create("2.0")`, once the new 2.2 branch is released in the near future it would be be picked up as well. This would probably be problematic because, for one, I doubt that all the formulae that use the gpg2 requirement would support the new version immediately. This would of course lead to issues installing and using those formulae.

Therefore, here I'd rather take the more conservative approach and not simplify it to a single comparison. This way we have more control over when and how the `GnuPG 2.2.x` branch is introduced into the Homebrew ecosystem.

MikeMcQuaid: Cool, seems fine. Parens would be good, then. Thanks 👍

DomT4:
> There is a reason why I left default_formula "gnupg2" as it was. I ran into the quite thorny problem you presented and could not commit to a course of action. This is partially due to the somewhat uncertain state of things in general.

It's mildly annoying that the `default_formula` usage is as inflexible as it is. However, folks will need to settle on some sort of change here because under @ilovezfs' current core PR `gnupg2` is being rerouted to `gnupg` which will be the 2.1.x version, so the `default_formula` is getting silently changed or explicitly changed. Might as well settle on what to do and do it explicitly, IMO.

One other problem with this PR and ILZ's core PR is that the latter will stop the `Gpg` here ever finding a 1.4.x `gpg` without fiddling with the `$PATH`. That's already kind of the case since 1.4.x started being installed as `gpg1` and such, but `keg_only` makes it a formal impossibility to find a 1.4.x along the default `$PATH`.

> Unfortunately, this reasoning and approach ignores a current use case, namely using the requirement to allow for the use of non-brew installations of gpg 2.0.x by formula. I think this is an issue that needs to be addressed before this (or something like it) was to be merged.

I think `GPG Suite` is popular enough that Homebrew is doing a "good thing" by trying to support it, but I don't think Homebrew has to be bound by a decision I made on that 7+ months ago. If there's a way to continue supporting it I think that's worthwhile, but sooner or later 2.1.x will become the default for both sides of this support dance.

DomT4:
I think this is alright as-is, although possibly still worth doing this:
> Theoretically everything in core currently using gpg might need testing afresh given we're forcing a new major gpg version onto everyone (which I think is the "right thing to do" ™️ under this model), but I'll leave that call to you.

Ideally here once everything's ready, you'll want to merge this, cut a new tag, and then merge https://github.com/Homebrew/homebrew-core/pull/11083 immediately after. That should keep things fairly smooth for everyone.

JCount:
@DomT4 Thanks for the feedback! At this point it seems necessary to keep support for gpg 2.0.x in brew, but trying to maintain support for too long will just cause issues later on. One possibility could be to remove this support once GnuPG 2.2.x is released, and we have a formula for it in core. That would ideally allow for this final phaseout of 2.0.x to happen in concert with the phasing in of 2.2.x. I haven't given this idea much real thought, but it seems like a logical progression of events. 

(However, as you know, sometimes things aren't quite logical here. 😺)

DomT4:
> At this point it seems necessary to keep support for gpg 2.0.x in brew, but trying to maintain support for too long will just cause issues later on.

Agree on this. I'd support it at least until `GPG Suite` drops it, which might come with macOS 10.13, presuming Apple sticks with the 10.x version theme. I'm not sure what the upstream GnuPG plan is around versioning but it took a long, long time between 2.0 and 2.1 going stable, like close to a decade a long long time, heh.

MikeMcQuaid:
@JCount 🚢 this so it ends up in the next tag? GPG stuff will be pulled shortly afterwards I suspect.

PR #2358, opened 2017-03-17
upgrade: keg_only before non-keg_only formulae

ilovezfs:

-----

Avoids unnecessary conflicts when a previously non-keg_only formula is
in the way by prioritizing keg_only before non-keg_only formulae.

This change is motivated by the upgrade of gnupg 2.0 to 2.1, since the
latter no longer depends on the gpg-agent formula, which, even if made
keg_only, still causes the link step to fail for 2.1, as gpg-agent's
non-keg version won't have been upgraded to the keg_only version at that
point (alphabetically gnupg precedes gpg-agent).

### DIFF
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -75,6 +75,18 @@ module Homebrew
       puts pinned.map { |f| "#{f.full_specified_name} #{f.pkg_version}" } * ", "
     end
 
+    # Sort keg_only before non-keg_only formulae to avoid any needless conflicts
+    # with outdated, non-keg_only versions of formulae being upgraded.
+    formulae_to_install.sort! do |a, b|
+      if !a.keg_only? && b.keg_only?
+        1
+      elsif a.keg_only? && !b.keg_only?
+        -1
+      else
+        0
+      end
+    end
+
     formulae_to_install.each do |f|
       upgrade_formula(f)
       next unless ARGV.include?("--cleanup")


Library/Homebrew/cmd/upgrade.rb
```diff
@@ -75,6 +75,15 @@ def upgrade
...
 
+    formulae_to_install.sort! do |a, b|
+      if !a.keg_only? && b.keg_only?
+        1
+      elsif a.keg_only? && !b.keg_only?
+        -1
+      else
+        0
+      end
+    end
```
MikeMcQuaid: Newline after this.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -75,6 +75,15 @@ def upgrade
       puts pinned.map { |f| "#{f.full_specified_name} #{f.pkg_version}" } * ", "
     end
 
+    formulae_to_install.sort! do |a, b|
```
MikeMcQuaid: Comment above this explaining what the motivation is and what it's doing.

Library/Homebrew/cmd/upgrade.rb
ilovezfs: done.

Library/Homebrew/cmd/upgrade.rb
ilovezfs: done.

ilovezfs:
PR refreshed.

ilovezfs:
Shipped!

PR #2359, opened 2017-03-17
migrator: allow new cellar to exist already

ilovezfs:

-----

Lets us migrate a formula to a name that may have previously been used.

If gnupg 1.x is installed as "gnupg" and gnupg 2.x is installed as
"gnupg2," it's currently not possible to rename gnupg2 -> gnupg, since
the 1.4 keg will already be installed in the gnupg cellar, so in order
to reclaim the name "gnupg" to be used for 2.1, either 1.x must be
manually uninstalled, or the new cellar needs to be allowed to exist
already.

The other choice here would be to continue using the name "gnupg2" for
the 2.1 series, and not reclaim the name "gnupg" from 1.x, which,
incidentally, would open up the possibility of a gnupg -> gnupg@1.4
rename.

### DIFF
--- a/Library/Homebrew/migrator.rb
+++ b/Library/Homebrew/migrator.rb
@@ -147,15 +147,25 @@ class Migrator
   end
 
   def migrate
-    if new_cellar.exist?
-      onoe "#{new_cellar} already exists; remove it manually and run brew migrate #{oldname}."
-      return
+    if old_cellar.exist? && new_cellar.exist?
+      conflicted = false
+      old_cellar.each_child do |c|
+        if (new_cellar/c.basename).exist?
+          conflicted = true
+          onoe "#{new_cellar/c.basename} already exists."
+        end
+      end
+      if conflicted
+        onoe "Remove #{new_cellar} manually and run brew migrate #{oldname}."
+        return
+      end
     end
 
     begin
       oh1 "Migrating #{Formatter.identifier(oldname)} to #{Formatter.identifier(newname)}"
       lock
       unlink_oldname
+      unlink_newname if new_cellar.exist?
       move_to_new_directory
       repin
       link_oldname_cellar
@@ -178,7 +188,11 @@ class Migrator
   # move everything from Cellar/oldname to Cellar/newname
   def move_to_new_directory
     puts "Moving to: #{new_cellar}"
-    FileUtils.mv(old_cellar, new_cellar)
+    if new_cellar.exist?
+      FileUtils.mv(old_cellar.children, new_cellar)
+    else
+      FileUtils.mv(old_cellar, new_cellar)
+    end
   end
 
   def repin
@@ -202,6 +216,14 @@ class Migrator
   def unlink_oldname
     oh1 "Unlinking #{Formatter.identifier(oldname)}"
     old_cellar.subdirs.each do |d|
+      keg = Keg.new(d)
+      keg.unlink
+    end
+  end
+
+  def unlink_newname
+    oh1 "Unlinking #{Formatter.identifier(newname)}"
+    new_cellar.subdirs.each do |d|
       keg = Keg.new(d)
       keg.unlink
     end


MikeMcQuaid:
CC @vladshablinsky for review and thoughts here.

@ilovezfs in cases like this I'd have thought the best thing would be to create a new `gnupg@1` formula and upgrade the existing `gnupg` formula to be 2.X. Thoughts?

ilovezfs:
That's what we're doing.

MikeMcQuaid:
Ok, I reread and I think I understand now, thanks.

vladshablinsky:
@ilovezfs do we really need that `gnupg2 -> gnupg` rename? I think I need some time to revise how all that things with versions work but something doesn't feel right for now. I think we can eventually end up with a rename that will probably break thins, but I'm not sure. Will answer as soon as I can.

Also, I hope we'll allow any types of renames and deletes sooner or later so we're able to rename anything to anything if that't name is free at specific moment of time. So probably we need to be careful with complicating renames logic. Now when versioned formulae arrived it got really more complicated than it used to be.

ilovezfs:
>@ilovezfs do we really need that gnupg2 -> gnupg rename?

If we want to upgrade all gnupg 1.0.x and 2.0.x users to 2.1.x, then yes.

Note that we also have been sitting on https://github.com/Homebrew/homebrew-core/pull/8334 and not shipping it for three months for the same reason.

MikeMcQuaid:
> do we really need that gnupg2 -> gnupg rename?

I think if there are any known problems we should fix them but if there's none specifically known then it's fine to 🚢 this. 


> Note that we also have been sitting on Homebrew/homebrew-core#8334 and not shipping it for three months for the same reason.

Not quite right; it's blocked on https://github.com/Homebrew/brew/issues/1770.

ilovezfs:
>Not quite right; it's blocked on #1770.

So does 1770 affect this gpg too then? Not sure I see a difference between the situations really.

ilovezfs:
>Not quite right; it's blocked on #1770.

Also, let me re-phrase then. Regardless of the disposition of #1770, which I did see existed, this PR _also_ seems to be a requirement for the qt5 -> qt rename to proceed. Whether that had been realized heretofore or not isn't especially relevant.

MikeMcQuaid:
> So does 1770 affect the gpg upgrades too then?

I have no idea. I've still not been able to debug that problem yet.

ilovezfs:
@vladshablinsky https://github.com/Homebrew/brew/issues/1770#issuecomment-287700693

vladshablinsky:
@ilovezfs https://github.com/Homebrew/brew/pull/1772#issuecomment-287930580

The only concern I have is what if somehow we got a different formula installed under `newname` cellar? I think it can be possible if we allow multiple renames.

ilovezfs:
@vladshablinsky I think it's `link_newname`'s decision which of the kegs to link.

vladshablinsky:
>  I think it's link_newname's decision which of the kegs to link.

Also, let's suppose we have a tap installed under `newname` and `oldname` -> `newname` core rename. What happens if we migrate `oldname` to `newname` allowing `newname` to exist while performing migration? Well, we'll have two different formula sharing same name under the same Cellar. I can remember that when formula renames were implemented we only supposed that only one formula can be installed into one cellar subdirectory. So we at least need an extra check that installed formula is from the same tap as the formula we're trying to migrate. Thoughts?

ilovezfs:
The tap migrations from homebrew/versions to homebrew/core all had a rename to the @ naming scheme, so I'm not sure we can require the source and destination taps to be the same.

vladshablinsky:
> The tap migrations from homebrew/versions to homebrew/core all had a rename to the @ naming scheme, so I'm not sure we can require the source and destination taps to be the same.

I might have confused you, what I meant were formula renames and migrations of renamed formula.   

Let's migrate `oldname2` from `homebrew/version` to `homebrew/core`, then we have
`oldname2` installed and it's tap is `homebrew/core` due to migration being performed.
Now we want to rename it to `oldname@2` while `Cellar/oldname@2` already exists. No problem, but if we want to perform a migration for this rename we need to be sure `Cellar/oldname@2`'s kegs are `homebrew/core`'s and not from some other taps that also provides `oldname@2`.
Am I wrong?

ilovezfs:
Yes I think that makes sense. Even though there never was a formula named gcc48 in homebrew/core, I think a lookup of the tap of a not-yet-renamed-to-gcc@4.8 gcc48 keg originally installed from homebrew/versions would return homebrew/core.

ilovezfs:
@vladshablinsky Yes, you are right. It looks like the `from_same_taps?` test doesn't consider the tap of a pre-existing new-name cellar. I'm wonder if we should tolerate that situation and defer the migration until the conflicting cellar goes away, or print a manual tap migration error of some kind, or something else.

MikeMcQuaid:
> defer the migration until the conflicting cellar goes away

When would that happen?

vladshablinsky:
> > defer the migration until the conflicting cellar goes away

>When would that happen?

Probably when user eventually uninstalls it.


> it looks like the `from_same_taps?` test doesn't consider the tap of a pre-existing new-name cellar.

Even though what you say is true, please notice that this is not the case we're discussing.

`from_same_taps?` checks if `oldname` installed is from the same tap as `Formula` for `oldname` however the case we're discussing is if Formula for `newname` (for which we migrate its old Cellar) is from the same tap as `newname` installed.

MikeMcQuaid:
> Probably when user eventually uninstalls it.

I'm wondering if we should uninstall it for the user in a case like this (or at least prompt them to do so before migration?).

vladshablinsky:
> I'm wondering if we should uninstall it for the user in a case like this (or at least prompt them to do so before migration?).

IMO, migration must be impossible unless `--force` given. With force we should either uninstall it and perform migration or merge several formulae's kegs under `Cellar/newname`. :+1: for prompting.

ilovezfs:
PR refreshed.

PR #2374, opened 2017-03-21
Cleanup: Continue on error removing keg

### LINKED ISSUE #2355: cleanup command stops on permission denied
### Bug reports:
```
%  brew doctor

Warning: A .pydistutils.cfg file was found in $HOME, which may cause Python
builds to fail. See:
  https://bugs.python.org/issue6138
  https://bugs.python.org/issue4655

Warning: Some installed formula are missing dependencies.
You should `brew install` the missing dependencies:
  brew install shared-mime-info

Run `brew missing` for more details.

% brew config
HOMEBREW_VERSION: 1.1.11
HEAD: 394f9fa0aaa5854aa52bc589708a079665dcf462
Last commit: 10 days ago
Core tap HEAD: 58d65f62dc3416ec1f3c4c63b3a9aa1b72b31f3e
Core tap last commit: 29 minutes ago
HOMEBREW_PREFIX: /usr/local
CPU: quad-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
GCC-4.2: build 5666
Clang: 8.0 build 800
Git: 2.12.0 => /usr/local/bin/git
Perl: /usr/bin/perl
Ruby: /usr/local/bin/ruby => /usr/local/Cellar/ruby/2.4.0/bin/ruby
Java: 1.7.0_71
macOS: 10.11.6-x86_64
Xcode: 8.2.1
CLT: 8.2.0.0.1.1480973914
X11: 2.7.9 => /opt/X11
```

When running the `brew cleanup` command, Homebrew seems to stop on the first error it encounters.

Meaning that if a single formulae can't be cleaned up, the following won't be either.

The behaviour I'm expecting would be to skip the said formulae, but continue with the others

Example in my case (bind having root-owned files):
```
% brew cleanup
Warning: Skipping ack: most recent version 2.16 not installed
Warning: Skipping asciidoc: most recent version 8.6.9_1 not installed
Removing: /usr/local/Cellar/bind/9.11.0-P2... (420 files, 50.5MB)
Error: Permission denied - /usr/local/Cellar/bind/9.11.0-P2/var/run/named
… cleanup stopped here

% sudo rm -r /usr/local/Cellar/bind/9.11.0-P2/var
% brew cleanup
Warning: Skipping asciidoc: most recent version 8.6.9_1 not installed
Removing: /usr/local/Cellar/bind/9.11.0-P2... (68B)
Warning: Skipping cairo: most recent version 1.14.8 not installed
… cleanup continues normally

```

joshka:

-----
Fixes #2355

### DIFF
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -6,7 +6,9 @@ module Homebrew
   module Cleanup
     @disk_cleanup_size = 0
 
-    def self.cleanup
+    module_function
+
+    def cleanup
       cleanup_cellar
       cleanup_cache
       cleanup_logs
@@ -15,34 +17,41 @@ module Homebrew
       rm_ds_store
     end
 
-    def self.update_disk_cleanup_size(path_size)
+    def update_disk_cleanup_size(path_size)
       @disk_cleanup_size += path_size
     end
 
-    def self.disk_cleanup_size
+    def disk_cleanup_size
       @disk_cleanup_size
     end
 
-    def self.cleanup_formula(formula)
-      formula.eligible_kegs_for_cleanup.each do |keg|
-        cleanup_path(keg) { keg.uninstall }
-      end
+    def unremovable_kegs
+      @unremovable_kegs ||= []
     end
 
-    def self.cleanup_logs
+    def cleanup_cellar(formulae = Formula.installed)
+      formulae.each(&method(:cleanup_formula))
+    end
+
+    def cleanup_formula(formula)
+      formula.eligible_kegs_for_cleanup.each(&method(:cleanup_keg))
+    end
+
+    def cleanup_keg(keg)
+      cleanup_path(keg) { keg.uninstall }
+    rescue Errno::EACCES => e
+      opoo e.message
+      unremovable_kegs << keg
+    end
+
+    def cleanup_logs
       return unless HOMEBREW_LOGS.directory?
       HOMEBREW_LOGS.subdirs.each do |dir|
         cleanup_path(dir) { dir.rmtree } if prune?(dir, days_default: 14)
       end
     end
 
-    def self.cleanup_cellar
-      Formula.installed.each do |formula|
-        cleanup_formula formula
-      end
-    end
-
-    def self.cleanup_cache(cache = HOMEBREW_CACHE)
+    def cleanup_cache(cache = HOMEBREW_CACHE)
       return unless cache.directory?
       cache.children.each do |path|
         if path.to_s.end_with? ".incomplete"
@@ -97,7 +106,7 @@ module Homebrew
       end
     end
 
-    def self.cleanup_path(path)
+    def cleanup_path(path)
       if ARGV.dry_run?
         puts "Would remove: #{path} (#{path.abv})"
       else
@@ -108,7 +117,7 @@ module Homebrew
       update_disk_cleanup_size(path.disk_usage)
     end
 
-    def self.cleanup_lockfiles
+    def cleanup_lockfiles
       return unless HOMEBREW_LOCK_DIR.directory?
       candidates = HOMEBREW_LOCK_DIR.children
       lockfiles  = candidates.select(&:file?)
@@ -118,7 +127,7 @@ module Homebrew
       end
     end
 
-    def self.rm_ds_store
+    def rm_ds_store
       paths = Queue.new
       %w[Cellar Frameworks Library bin etc include lib opt sbin share var]
         .map { |p| HOMEBREW_PREFIX/p }.each { |p| paths << p if p.exist? }
@@ -136,7 +145,7 @@ module Homebrew
       workers.map(&:join)
     end
 
-    def self.prune?(path, options = {})
+    def prune?(path, options = {})
       @time ||= Time.now
 
       path_modified_time = path.mtime
--- a/Library/Homebrew/cmd/cleanup.rb
+++ b/Library/Homebrew/cmd/cleanup.rb
@@ -21,16 +21,26 @@ module Homebrew
     if ARGV.named.empty?
       Cleanup.cleanup
     else
-      ARGV.resolved_formulae.each { |f| Cleanup.cleanup_formula f }
+      Cleanup.cleanup_cellar(ARGV.resolved_formulae)
     end
 
-    return if Cleanup.disk_cleanup_size.zero?
+    report_disk_usage unless Cleanup.disk_cleanup_size.zero?
+    report_unremovable_kegs unless Cleanup.unremovable_kegs.empty?
+  end
 
+  def report_disk_usage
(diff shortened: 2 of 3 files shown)


Library/Homebrew/cleanup.rb
```diff
@@ -36,10 +42,18 @@ def self.cleanup_logs
       end
     end
 
-    def self.cleanup_cellar
-      Formula.installed.each do |formula|
-        cleanup_formula formula
+    def self.cleanup_cellar(formulae = Formula.installed)
```
MikeMcQuaid: I don't think we need a default argument here.

Library/Homebrew/cleanup.rb
```diff
@@ -36,10 +42,18 @@ def self.cleanup_logs
...
+      unremovable_kegs = []
+      formulae.each do |formula|
+        cleanup_formula formula, unremovable_kegs
```
MikeMcQuaid: Feels weird to pass `unremovable_kegs` as an argument; what about an instance variable instead?

Library/Homebrew/cleanup.rb
```diff
@@ -36,10 +42,18 @@ def self.cleanup_logs
...
       end
+
+      return if unremovable_kegs.empty?
+      ofail <<-EOS.undent
+        Could not cleanup old kegs
```
MikeMcQuaid: `Could not cleanup old kegs! Fix your permissions on:`

Library/Homebrew/cleanup.rb
```diff
@@ -36,10 +42,18 @@ def self.cleanup_logs
...
+        Check you have permission to remove the following folders in the cellar:
```
MikeMcQuaid: Can roll this line into the previous one.

Library/Homebrew/cleanup.rb
```diff
@@ -36,10 +42,18 @@ def self.cleanup_logs
...
+          #{unremovable_kegs.join " "}
```
MikeMcQuaid: `.join "\n  "`

Library/Homebrew/cleanup.rb
joshka: The use case of this is to have the command not duplicate the continue on fail logic. See https://github.com/Homebrew/brew/pull/2374/files#diff-8cc42f61c29a8f960250394cbc3319c4L24

Library/Homebrew/cleanup.rb
joshka: I agree, I wasn't sure whether we needed to maintain the previous behavior of failing when cleanup_formula fails for individual calls, so I did it this way. I'll change this to use an instance variable.

joshka: Using an instance variable also allows us to move the failure messaging to the command rather than the library functions. We can also continue onto the cache, logs, lockfiles etc when formula cleanup fails.

joshka:
Broken test seems like it's a connection issue. Not sure if this will work for me or if it's locked to maintainers, but here goes nothing:
@BrewTestBot test this please

joshka:
Thanks @ilovezfs

ilovezfs:
yw

Library/Homebrew/cleanup.rb
```diff
@@ -23,10 +24,27 @@ def self.disk_cleanup_size
       @disk_cleanup_size
     end
 
+    def self.add_unremovable_keg(keg)
+      @unremovable_kegs << keg
+    end
+
+    def self.unremovable_kegs
+      @unremovable_kegs
+    end
```
MikeMcQuaid: I don't think these helper methods are needed; just write directly to the instance variable.

Library/Homebrew/cleanup.rb
MikeMcQuaid: Gotcha, I misread, thanks.

Library/Homebrew/cmd/cleanup.rb
```diff
@@ -21,16 +21,26 @@ def cleanup
...
       ohai "This operation would free approximately #{disk_space} of disk space."
     else
       ohai "This operation has freed approximately #{disk_space} of disk space."
     end
   end
+
+  def report_unremovable_kegs
+    ofail <<-EOS.undent
+      Could not cleanup old kegs! Fix your permissions on:
+        #{unremovable_kegs.join "\n  "}
```
joshka: This needs a Cleanup. in front

Library/Homebrew/cleanup.rb
joshka: It's used externally from the Cleanup command. I wrote it this way for consistency with the two `disk_cleanup_size` methods rather than a knowledge of idiomatic ruby (I'm a ruby n00b coming from a mostly C# background). I could reasonably drop the `add_unremovable_keg` method, but I think the accessor needs to remain.

Is the following more idiomatic in ruby?

```ruby
def self.unremovable_kegs
  @unremovable_kegs ||= []
end

def cleanup_keg(keg)
  # ...
  unremovable_kegs << keg
```

or:

```ruby
@unremovable_kegs = []
attr_accessor :unremovable_kegs

def cleanup_keg(keg)
  # ...
  unremovable_kegs << keg
```

joshka:
A couple of quick questions:
Do you prefer commits to be rebased prior to merge?
Is there a style guide that you have a use in this project. I'm fairly new to ruby (though not software development).
It might be best to update https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md with the answers to these questions as it would help others, instead of just me.

Library/Homebrew/cleanup.rb
joshka: I went with the first option based on not finding any instances of attr_reader through HBC where the class only has class methods.

joshka:
One more question. This class only has class methods, should it be a module? Some of the ruby style guides I'm reading suggest this approach, others do not.

MikeMcQuaid:
> Do you prefer commits to be rebased prior to merge?

We don't mind either way on Homebrew/brew as we use merge commits anyway.

> Is your preferred style guide the github one?

Yeh although our Rubocop does most of the enforcing too.

> It might be best to update https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md with the answers to these questions as it would help others, instead of just me.

We generally rely on tooling and CI rather than documentation as it's easier to keep up to date.



> This class only has class methods, should it be a module?

Yeh, that may make sense and could use `module_function` on it. Feel free to add to this PR?

MikeMcQuaid:
@joshka Travis CI got upset. Can you force push here to kick it off again? Thanks!

joshka:
I wasn't sure if there was an easy way to force push without making a change, so instead I squashed the commits and rewrote the commit messages. I made no other changes other than this.

MikeMcQuaid:
Great work here (again) @joshka!

PR #1732, opened 2016-12-26
Hint at new location of migrated formulae

zmwangx:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

This is a partial implementation of https://github.com/Homebrew/brew-evolution/pull/15. Please read [the proposal](https://github.com/zmwangx/brew-evolution/blob/hint-at-new-location-of-migrated-formulae/015-hint-at-new-location-of-migrated-formulae.md), which was basically agreed upon. (Note that #1371 was also a step towards the goal.)

With this change:

```
$ brew info aeskeyfind
Error: No available formula with the name "aeskeyfind"
==> Searching for migrated formulae...
A deprecated formula named "aeskeyfind" has been migrated from homebrew/core to homebrew/boneyard.
A formula named "aeskeyfind" has been migrated from zmwangx/old to zmwangx/new.
```

(`zmwangx/old` has a placeholder `tap_migrations.json` with `{"aeskeyfind": "zmwangx/new"}`.)

```
$ brew install aeskeyfind
Error: No available formula with the name "aeskeyfind"
==> Searching for migrated formulae...
A deprecated formula named "aeskeyfind" has been migrated from homebrew/core to homebrew/boneyard.
A formula named "aeskeyfind" has been migrated from zmwangx/old to zmwangx/new.
```

```
$ brew info r
Error: No available formula with the name "r"
==> Searching for migrated formulae...
A formula named "r" has been migrated from caskroom/cask to homebrew/science.
```

```
$ brew install r
Error: No available formula with the name "r"
==> Searching for migrated formulae...
A formula named "r" has been migrated from caskroom/cask to homebrew/science.
```

```
$ brew info aeskeyfin
Error: No available formula with the name "aeskeyfin"
==> Searching for migrated formulae...
```

```
$ brew install aeskeyfin
Error: No available formula with the name "aeskeyfin"
==> Searching for migrated formulae...
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

PR #2377, opened 2017-03-22
missing_formula: warn when git-log takes very long

zmwangx:

-----

Since #1732, trying to install or info a nonexistent formula that's neither blacklisted nor recorded in a tap migration database triggers a search through the entire git history, which could take very long (~10s on my machine) in a non-shallow core tap:

```zsh
$ time (brew info foo)  # nothing happens for ten seconds
Error: No available formula with the name "foo"
( brew info foo; )  8.97s user 0.65s system 96% cpu 9.990 total
```

This is confusing to users (users shouldn't encounter the output of `deleted_reason` until we drop homebrew/boneyard). Therefore, we should let users know what's going on behind the scenes if `git-log` goes on for too long. I'm setting the threshold to two seconds. After this PR:

```zsh
$ time (brew info foo)
Warning: Searching through git history. This may take a while...
Error: No available formula with the name "foo"
( brew info foo; )  9.01s user 0.66s system 98% cpu 9.859 total
```

An iTerm2 screenshot with timestamps on each output line:

screen shot 2017-03-22 at 12 40 16 am

### DIFF
--- a/Library/Homebrew/missing_formula.rb
+++ b/Library/Homebrew/missing_formula.rb
@@ -125,11 +125,23 @@ module Homebrew
         relative_path = path.relative_path_from tap.path
 
         tap.path.cd do
-          # We know this may return incomplete results for shallow clones but
-          # we don't want to nag everyone with a shallow clone to unshallow it.
-          log_command = "git log --name-only --max-count=1 --format=%H\\\\n%h\\\\n%B -- #{relative_path}"
-          hash, short_hash, *commit_message, relative_path =
-            Utils.popen_read(log_command).gsub("\\n", "\n").lines.map(&:chomp)
+          begin
+            timer_pid = fork do
+              # Let the user know what's going on when the search goes on for
+              # more than two seconds.
+              sleep 2
+              opoo "Searching through git history. This may take a while..."
+            end
+
+            # We know this may return incomplete results for shallow clones but
+            # we don't want to nag everyone with a shallow clone to unshallow it.
+            log_command = "git log --name-only --max-count=1 --format=%H\\\\n%h\\\\n%B -- #{relative_path}"
+            hash, short_hash, *commit_message, relative_path =
+              Utils.popen_read(log_command).gsub("\\n", "\n").lines.map(&:chomp)
+          ensure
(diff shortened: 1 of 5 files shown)


ilovezfs:
Does git log let us improve performance by setting a maxdepth on the search in a non-shallow clone?

zmwangx:
> Does git log let us improve performance by setting a maxdepth on the search in a non-shallow clone?

Sure, you can use a commit range. That's rather arbitrary though.

zmwangx:
By the way, tests don't seem to like `fork`, for some reason.

ilovezfs:
Yeah I'm just wondering if we really want spelling errors triggering 10s searches but :man_shrugging:

Library/Homebrew/missing_formula.rb
```diff
@@ -125,11 +125,23 @@ def deleted_reason(name)
...
-          # we don't want to nag everyone with a shallow clone to unshallow it.
-          log_command = "git log --name-only --max-count=1 --format=%H\\\\n%h\\\\n%B -- #{relative_path}"
-          hash, short_hash, *commit_message, relative_path =
-            Utils.popen_read(log_command).gsub("\\n", "\n").lines.map(&:chomp)
+          begin
+            timer_pid = fork do
+              # Let the user know what's going on when the search goes on for
+              # more than two seconds.
+              sleep 2
+              opoo "Searching through git history. This may take a while..."
```
MikeMcQuaid: `Searching previously deleted formulae...` ?

JCount: That seems a bit more accurate, yet at the same time it is somewhat strange. This is given, at least for me, that what triggers this is usually a mistaken key stroke. From that perspective it may be better to just leave it completely generic?...

MikeMcQuaid:
Pushed some tweaks here. Made `install` and `info` immediately print the formula unavailable exception and then the search message. Should make it obvious it can be Ctrl-Cd if necessary. `search` is silent because it's already a long-running command and matches the existing behaviour (and I removed another search header that looks out of place).

zmwangx:
Looks reasonable to me.

PR #2401, opened 2017-03-27
Include the pinned version in brew outdated output for pinned formulae

wjlroe:

-----

This pull request adds pinned versions to the output of the `brew outdated` command. The format of the output is similar to the output from `brew info` in that it appends `[pinned at __version__]` to pinned formulae. For the JSON output, 2 fields are added: `pinned` (`true`|`false`) and `pinned_version` ("the_version" or `null`).

This is mainly useful for the peace of mind when running `brew outdated` before a `brew upgrade` that Homebrew has definitely pinned a formula and it won't get upgraded, even though it's outdated.

In addition to the added output, I've added tests for the previous output (effectively verbose output)  for unpinned formulae. I've divided these tests into their own contexts which should hopefully aid readability of the tests.

Here's an example of the output:

```bash
 % brew outdated
terraform (0.8.8) < 0.9.1_1 [pinned at 0.8.8]
```

### DIFF
--- a/Library/Homebrew/cmd/outdated.rb
+++ b/Library/Homebrew/cmd/outdated.rb
@@ -64,7 +64,9 @@ module Homebrew
           "#{full_name} (#{kegs.map(&:version).join(", ")})"
         end.join(", ")
 
-        puts "#{outdated_versions} < #{current_version}"
+        pinned_version = " [pinned at #{f.pinned_version}]" if f.pinned?
+
+        puts "#{outdated_versions} < #{current_version}#{pinned_version}"
       else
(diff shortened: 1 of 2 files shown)


wjlroe:
The build error I'm seeing on Travis seems to be on the master branch, rather than introduced here, so I will rebase this once that's gone away.

MikeMcQuaid:
Have rerun the tests. Looking good so far.

MikeMcQuaid:
Thanks for your (great!) first contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #2403, opened 2017-03-27
ruby_requirement: fix path prepend

carlorosati:

-----

### DIFF
--- a/Library/Homebrew/requirements/ruby_requirement.rb
+++ b/Library/Homebrew/requirements/ruby_requirement.rb
@@ -11,7 +11,7 @@ class RubyRequirement < Requirement
   satisfy(build_env: false) { new_enough_ruby }
 
   env do
-    ENV.prepend_path "PATH", new_enough_ruby
+    ENV.prepend_path "PATH", new_enough_ruby.dirname
   end
 
   def message


MikeMcQuaid:
What is this fixing?

To help us understand this issue can you explain:
- What happened
- What you expected to happen
- Step-by-step reproduction instructions (with as minimal input data as possible)
- How this fixes your problem

carlorosati:
prepend_path [on line 95 of ENV/shared.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/ENV/shared.rb#L95) will ignore path parameters that are not directories.

[new_enough_ruby](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/ruby_requirement.rb#L38) returns the path to a ruby binary.

Thus, the prepend_path will never affect the PATH.

I changed ruby_requirement to return the directory that the ruby binary is located in so that it may be added to the PATH as @kke most likely intended.

You can reproduce the behavior by adding a print statement before and after the prepend_path to verify that PATH is not modified.

MikeMcQuaid:
Makes sense! Thanks for your first contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #2406, opened 2017-03-28
Install a bottle from an URL

sjackman:

-----

This PR makes it possible to install a bottle from an URL. For example
```sh
brew install https://4574-56942570-gh.circle-artifacts.com/0/tmp/circle-artifacts.PaJXYOU/git-2.12.2.x86_64_linux.bottle.tar.gz
```

### DIFF
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -105,7 +105,18 @@ module Formulary
   # Loads formulae from bottles.
   class BottleLoader < FormulaLoader
     def initialize(bottle_name)
-      @bottle_filename = Pathname(bottle_name).realpath
+      case bottle_name
+      when %r{(https?|ftp|file)://}
+        # The name of the formula is found between the last slash and the last hyphen.
+        resource = Resource.new bottle_name[%r{([^/]+)-}, 1] { url bottle_name }
+        downloader = CurlBottleDownloadStrategy.new resource.name, resource
+        @bottle_filename = downloader.cached_location
+        cached = @bottle_filename.exist?
+        downloader.fetch
+        ohai "Pouring the cached bottle" if cached
+      else
+        @bottle_filename = Pathname(bottle_name).realpath
+      end
       name, full_name = Utils::Bottles.resolve_formula_names @bottle_filename
       super name, Formulary.path(full_name)
     end
@@ -335,10 +346,10 @@ module Formulary
 
   def self.loader_for(ref, from: nil)
     case ref
-    when %r{(https?|ftp|file)://}
-      return FromUrlLoader.new(ref)
     when Pathname::BOTTLE_EXTNAME_RX
       return BottleLoader.new(ref)
+    when %r{(https?|ftp|file)://}
+      return FromUrlLoader.new(ref)
     when HOMEBREW_TAP_FORMULA_REGEX
       return TapLoader.new(ref, from: from)
     end


chdiza:
This sounds great, particularly if the URL can be of the form `file:///blah/bloo/blee`.

MikeMcQuaid:
@chdiza We already support installing bottles from file paths, FYI.

chdiza:
> We already support installing bottles from file paths, FYI.

I haven't tried in a while, but I found this to be pretty buggy last time (at least when the bottles were "bottled" locally, rather than being cached from Bintray).

sjackman:
> I haven't tried in a while, but I found this to be pretty buggy last time (at least when the bottles were "bottled" locally, rather than being cached from Bintray).

It works for me with two caveats.
1. The bottle version has to match the formula in the repo.
2. The formula in the repo must have a `sha256` for this OS.

1. is by design, but could be relaxed now that that the bottle includes the formula inside the bottle.
2. is a bug in my opinion.

sjackman:
TravisCI on macOS failed. It passed on Linux.
```
==> brew update-test --to-tag
Error: undefined method `chomp' for nil:NilClass
Please report this bug:
  http://docs.brew.sh/Troubleshooting.html
/Users/travis/build/Homebrew/brew/Library/Homebrew/dev-cmd/update-test.rb:36:in `update_test'
/Users/travis/build/Homebrew/brew/Library/Homebrew/brew.rb:91:in `<main>'
==> FAILED
```
https://travis-ci.org/Homebrew/brew/jobs/215813460#L177

This bug was hopefully (fingers crossed) resolved by https://github.com/Homebrew/brew/pull/2404
I'll rebase onto `master`.

sjackman:
The TravisCI failure is resolved after rebasing onto `master`.

Library/Homebrew/formulary.rb
```diff
@@ -105,7 +105,16 @@ def load_file
   # Loads formulae from bottles.
   class BottleLoader < FormulaLoader
     def initialize(bottle_name)
-      @bottle_filename = Pathname(bottle_name).realpath
+      case bottle_name
+      when %r{(https?|ftp|file)://}
+        @bottle_filename = HOMEBREW_CACHE/File.basename(bottle_name)
+        unless @bottle_filename.exist?
+          HOMEBREW_CACHE.mkpath
+          curl bottle_name, "-o", @bottle_filename
```
MikeMcQuaid: I think the `CurlBottleDownloadStrategy` should be used here instead to handle e.g. caching and resuming.

chdiza:
> We already support installing bottles from file paths

Tried this again just now, to pour a locally-held bottle that I myself built from a custom formula in my own tap.  (I.e., using `--build-bottle`, it is not just the bintray bottle held locally).  I get:

```
$ brew install file:///Users/chdiza/gnu-tar-1.29.sierra.bottle.tar.gz
[curl is doing its thing]
Error: gnu-tar-1.29.sierra.bottle.tar.gz: /stuff/brache/Formula/gnu-tar-1.29.sierra.bottle.tar.gz:1: Invalid char `\x1F' in expression
/stuff/brache/Formula/gnu-tar-1.29.sierra.bottle.tar.gz:1: invalid multibyte char (UTF-8)
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```
I get that same error when trying other packages too.

sjackman:
With this PR, the above command should work. Was your test with or without this PR applied?

Without this PR applied, this works:
```sh
brew install /Users/chdiza/gnu-tar-1.29.sierra.bottle.tar.gz
```
and this does not work:
```sh
brew install file:///Users/chdiza/gnu-tar-1.29.sierra.bottle.tar.gz
```

Library/Homebrew/formulary.rb
sjackman: Fixed. Thanks for the helpful suggestion, Mike.

chdiza:
> Was your test with or without this PR applied?

It was without.

> brew install /Users/chdiza/gnu-tar-1.29.sierra.bottle.tar.gz

That does NOT work sans the PR for me.  It merely tries building the package from source.

I will get to testing the PR momentarily.

sjackman:
> That does NOT work sans the PR for me. It merely tries building the package from source.

I'm not sure that this PR will help that. All it does is download the bottle to cache, and then install that file.

> I will get to testing the PR momentarily.

Thanks!

chdiza:
It also doesn't work even with the PR.  It still tries to build from source no matter what:
```
brew install --force-bottle  file:///users/chdiza/gzip-1.8.sierra.bottle.tar.gz
==> Downloading file:///users/chdiza/gzip-1.8.sierra.bottle.tar.gz
Already downloaded: /stuff/brache/gzip-1.8.sierra.bottle.tar.gz
Warning: Pouring the cached bottle.
==> Installing gzip from chdiza/dizaone
==> Downloading file:///stuff/bb/unix/gzip-1.8.tar.xz
Already downloaded: /stuff/brache/gzip-1.8.tar.xz
==> Verifying gzip-1.8.tar.xz checksum
tar xJf /stuff/brache/gzip-1.8.tar.xz
==> ./configure --prefix=/usr/local/Cellar/gzip/1.8
```

However, *with* the PR at least I'm not getting the "invalid multibyte char" error from above.

sjackman:
The formula for `gzip` doesn't have a bottle for `sierra`. See my comment above
https://github.com/Homebrew/brew/pull/2406#issuecomment-289825278

chdiza:
> The formula for gzip doesn't have a bottle sha256 for sierra. See my comment above
#2406 (comment)

OK.  So pouring local bottles that are made from one's custom tap are actually not supported.  I had thought that the only thing preventing this was that the CLI didn't seem to like a URL or filepath-to-the-gzipped-bottle as the argument.  So I thought that this PR would fix that bit, and then I could pour custom bottles.  I was wrong.

sjackman:
It's a bug in my opinion, and shouldn't be too terribly hard to fix. When pouring a local bottle or URL, it needs to skip the check for a `sha256` in the formula.

sjackman:
Note that the `sha256` in the formula doesn't have to match that of the local bottle, it just has to exist at all. `sha256 "" => :sierra` is good enough.

chdiza:
> Note that the bottle sha256 in the formula doesn't have to match that of the local bottle, it just has to exist. sha256 "" => :sierra is good enough.

I don't think this is true.  `gnu-tar` has a sha256 for sierra, but I can't pour it from a path or URL either, with or without this PR.  (Same result: `brew` just tries to build it from source.)

sjackman:
```
❯❯❯ brew install ~/Library/Caches/Homebrew/gnu-tar-1.29_1.el_capitan.bottle.tar.gz
==> Pouring gnu-tar-1.29_1.el_capitan.bottle.tar.gz
```
I'm not able to replicate that behaviour on `:el_capitan`.

chdiza:
> I'm not able to replicate that behaviour on :el_capitan.

Is the bottle you're trying to install just a cached version of the Bintray bottle?  I'm trying to install a bottle built from my own `gnu-tar` formula in my own tap.

sjackman:
Yes. Ah, I see. Bottling works for our custom taps `Linuxbrew/extra` and friends, but you may have some debugging to do. To eliminate one source of possible confusion, you may want to first try it on a formula name that's not also found in `Homebrew/core`.

sjackman:
@MikeMcQuaid Good to merge?

Library/Homebrew/formulary.rb
```diff
@@ -105,7 +105,17 @@ def load_file
   # Loads formulae from bottles.
   class BottleLoader < FormulaLoader
     def initialize(bottle_name)
-      @bottle_filename = Pathname(bottle_name).realpath
+      case bottle_name
+      when %r{(https?|ftp|file)://}
+        resource = Resource.new bottle_name[%r{([^/]+)-}, 1] { url bottle_name }
```
MikeMcQuaid: Not sure I understand the regex here, maybe warrants a comment?

Library/Homebrew/formulary.rb
```diff
@@ -105,7 +105,17 @@ def load_file
...
+        downloader = CurlBottleDownloadStrategy.new resource.name, resource
+        @bottle_filename = downloader.cached_location
+        cached = @bottle_filename.exist?
+        downloader.fetch
+        opoo "Pouring the cached bottle." if cached
```
MikeMcQuaid: `ohai` and skip the trailing `.`

MikeMcQuaid:
@chdiza Does this work for you now?

chdiza:
> @chdiza Does this work for you now?

I won't be in a position to check whether it works for the next several hours, at least.  Probably just merge it and I'll open a separate issue if needed.

Library/Homebrew/formulary.rb
sjackman: Done. The name of the formula is found between the last slash and the last hyphen.

Library/Homebrew/formulary.rb
sjackman: Done.

sjackman:
Thanks for the feedback, Mike. Updated.

MikeMcQuaid:
Thanks @sjackman!

sjackman:
My pleasure. Thanks for merging, Mike.

PR #2414, opened 2017-03-28
update-test: Use git fetch --tags --depth=1

sjackman:
Use `git fetch --tags --depth=1` to fetch fewer commits.

-----

`git fetch --tags --depth=1` works equally well as `git fetch --tags` for the purpose of `update-test --to-tag` and fetches fewer unnecessary historical commits.

There's a second somewhat related error that `brew --version` may not work if the repo is shallow, if there are any commits missing between `previous_tag` and `master`. Those commits will not be fetched by `git fetch --tags`, so it's unaffected by whether the tags above are fetched shallow or not.

I haven't found a `git` command yet to fetch just those commits between `previous_tag` and `master`, unless you know exactly how many of them there are and use `git fetch --depth=N`, but theres'a 🐔 and :egg: problem there. That makes `git fetch --unshallow` necessary to ensure that `brew --version` works.

### DIFF
--- a/Library/Homebrew/dev-cmd/update-test.rb
+++ b/Library/Homebrew/dev-cmd/update-test.rb
@@ -36,7 +36,7 @@ module Homebrew
       previous_tag =
         Utils.popen_read("git", "tag", "--list", "--sort=-version:refname").lines[1]
       unless previous_tag
-        safe_system "git", "fetch", "--tags"
+        safe_system "git", "fetch", "--tags", "--depth=1"
         previous_tag =
           Utils.popen_read("git", "tag", "--list", "--sort=-version:refname").lines[1]
       end


MikeMcQuaid:
What's the speed difference between this and the previous version on Travis CI?

sjackman:
I haven't tested on TravisCI. On my local machine, here's the timing / space data.

| Command                      | Time (s) | Space (MiB) | Commits fetched |
| ---------------------------- | -------- | ----------- | --------------- |
| `git fetch --tags`           | 25       | 29          | 12604           |
| `git fetch --tags --depth=1` | 4        | 10          | 41              |

Commits fetched calculted by `git rev-list --all | wc -l`
Note that `brew` has 40 stable tags.

sjackman:
+ `git fetch --tags` fetches all the commits from day 1 until the most recent tag. It does not fetch the commits from the most recent tag until `master`.
+ `git fetch --tags --depth=1` fetches one commit per tag.

MikeMcQuaid:
Thanks for the explanation @sjackman!

PR #2426, opened 2017-03-31
Remove osmium from blacklisted formulas

Zverik:
The author of the Osmium library says:

> That is from about 5 years back when Homebrew wanted to package the then existing osmium c++ lib. And they said they only can do that if i do proper released versions, which i didn't want to do at that time. Somehow they translated this into that sentence above. But i have proper releases for years now, and, as you mention, libosmium has been packaged.

So I guess we should unblacklist the Osmium library.


MikeMcQuaid:
Do you have a URL for that quote? Thanks!

joto:
@MikeMcQuaid No URL, but, yes, I said that in an IRC channel.

MikeMcQuaid:
@joto There's a URL for it now 😆.

Thanks @Zverik.

PR #2434, opened 2017-04-01
gpg_spec test add 2.1.x compatibility

JCount:

-----

This is meant to update the tests so that they do not begin failing once `GnuPG 2.1.x` is the default. Backwards compatibility for `2.0.x` has been maintained for now in `gpg_spec` because the binary is fetched directly from the path.

### DIFF
--- a/Library/Homebrew/test/gpg_spec.rb
+++ b/Library/Homebrew/test/gpg_spec.rb
@@ -13,7 +13,12 @@ describe Gpg do
         shutup do
           subject.create_test_key(dir)
         end
-        expect(dir/".gnupg/secring.gpg").to exist
+
+        begin
+          expect(dir/".gnupg/pubring.kbx").to be_file
+        rescue RSpec::Expectations::ExpectationNotMetError
+          expect(dir/".gnupg/secring.gpg").to be_file
+        end
       end
     end
   end
--- a/Library/Homebrew/test/gpg_spec.rb
+++ b/Library/Homebrew/test/gpg_spec.rb
@@ -12,14 +12,12 @@ describe Gpg do
 
         shutup do
           subject.create_test_key(dir)
-          gpg = subject::GPG_EXECUTABLE
-          @version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
         end
 
-        if @version.to_s.start_with?("2.1")
-          expect(dir/".gnupg/pubring.kbx").to be_file
+        if subject.version == Version.create("2.0")
+          expect(dir/".gnupg/secring.gpg").to be_a_file
         else
-          expect(dir/".gnupg/secring.gpg").to be_file
+          expect(dir/".gnupg/pubring.kbx").to be_a_file
         end
       end
     end
--- a/Library/Homebrew/test/gpg_spec.rb
+++ b/Library/Homebrew/test/gpg_spec.rb
@@ -12,11 +12,13 @@ describe Gpg do
 
         shutup do
           subject.create_test_key(dir)
+          gpg = subject::GPG_EXECUTABLE
+          @version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
         end
 
-        begin
+        if @version.to_s.start_with?("2.1")
           expect(dir/".gnupg/pubring.kbx").to be_file
-        rescue RSpec::Expectations::ExpectationNotMetError
+        else
           expect(dir/".gnupg/secring.gpg").to be_file
         end
       end
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -6,6 +6,7 @@ class Gpg
       gpg_short_version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
       next unless gpg_short_version
       gpg_version = Version.create(gpg_short_version.to_s)
+      @version = gpg_version
       gpg_version == Version.create("2.0") ||
         gpg_version == Version.create("2.1")
     end
@@ -23,6 +24,10 @@ class Gpg
 
   def self.available?
     File.executable?(GPG_EXECUTABLE.to_s)
+  end
+
+  def self.version
+    @version if available?
   end
 
   def self.create_test_key(path)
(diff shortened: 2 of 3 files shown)


JCount:
cc @MikeMcQuaid

Library/Homebrew/test/gpg_spec.rb
```diff
@@ -13,7 +13,12 @@
         shutup do
           subject.create_test_key(dir)
         end
-        expect(dir/".gnupg/secring.gpg").to exist
+
+        begin
+          expect(dir/".gnupg/pubring.kbx").to exist
+        rescue RSpec::Expectations::ExpectationNotMetError
```
ilovezfs: is this really idiomatic for `||` in rspec?

JCount: Probably not, since this is currently quite rough, but it does essentially the same thing functionally. Feedback from someone more versed in RSpec is probably badly needed.

ilovezfs: CC @reitermarkus

JCount:
ping @MikeMcQuaid 

Currently `brew tests` appears to be failing because of this test, so it probably would be good to get this in a shippable state. I mainly could use some help/advice on the RSpec aspect of this because functionally it appears sound. (I know very little about RSpec so this really is a hack job currently.)

Library/Homebrew/test/gpg_spec.rb
```diff
@@ -13,7 +13,12 @@
         shutup do
           subject.create_test_key(dir)
         end
-        expect(dir/".gnupg/secring.gpg").to exist
+
+        begin
+          expect(dir/".gnupg/pubring.kbx").to be_file
+        rescue RSpec::Expectations::ExpectationNotMetError
+          expect(dir/".gnupg/secring.gpg").to be_file
+        end
```
MikeMcQuaid: Can you explain in prose what this is trying to do? I don't really understand.

JCount: Ok, the core issue here is that `subject.create_test_key(dir)` can potentially invoke either gnupg 2.1.x or more rarely gnupg 2.0.x.

This is because `Gpg::find_gpg` will set `GPG_EXECUTABLE` to whichever version it finds first in the path. In turn, `Gpg::create_test_key` invokes this executable to generate the keystore. This is a problem, because 2.1.x will create a `pubring.kbx` file but not a `secring.gpg` file while the reverse is true for 2.0.x. 

Most likely it will find gnupg 2.1.x, since that is now the default formula, and the first expectation will pass. However, if gnupg 2.0.x is found first, the test will still be run and it is guaranteed the first expectation will fail. Therefore, that failure needs to be handled and the expectation designed for 2.0.x evaluated instead.

Hopefully that describes what I am trying to do with that code sufficiently.

MikeMcQuaid: Ok, in that case this should use an `if`/`else` to set the expectation accordingly rather than `rescue`.

JCount: Alright, I first tried to do that but got hung up on the RSpec side of things. Also, it will require a bit of thinking to figure out how to extract the version....

MikeMcQuaid: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/gpg.rb#L22 should help

JCount: thanks

Library/Homebrew/test/gpg_spec.rb
```diff
@@ -12,8 +12,15 @@
 
         shutup do
           subject.create_test_key(dir)
+          gpg = subject::GPG_EXECUTABLE
+          @version = Utils.popen_read(gpg, "--version")[/\d\.\d/, 0]
```
JCount: I am a bit unsatisfied with these two lines, the implementation is not very good.

MikeMcQuaid: Alternatively now I'm awake and thinking clearer: mock/stub the methods accordingly so it only ever hits one path and then add two tests. Shout if you need more help with that.

JCount: Ok, I am officially out of ideas. I now know a lot more about mocks/stubs, class_doubles, partial doubles, RSpec scope, etc. Unfortunately, though it was quite an educational experience, I have yet to produce a passing test and am essentially stuck. Aid with this would be greatly appreciated.

reitermarkus: Maybe adding a `version` method to the `Gpg` class would make this clearer and more robust than a `start_with?` check. Otherwise this seems fine, I don't see anything that could even be stubbed here, since it comes down to the version of the installed `gpg` executable.

JCount:
Is this generally what you had in mind, @reitermarkus? As always, I welcome feedback. 😄

reitermarkus:
> Is this generally what you had in mind, @reitermarkus?

Yes, this is definitely much nicer to look at. 👍

Library/Homebrew/test/gpg_spec.rb
```diff
@@ -13,7 +13,12 @@
         shutup do
           subject.create_test_key(dir)
         end
-        expect(dir/".gnupg/secring.gpg").to exist
+
+        if subject.version == Version.create("2.1")
```
reitermarkus: Maybe switch this around, so it falls back to the newer version in the `else` branch?

Library/Homebrew/test/gpg_spec.rb
```diff
@@ -13,7 +13,12 @@
...
+          expect(dir/".gnupg/pubring.kbx").to be_file
```
reitermarkus: `be_a_file`

JCount: Why use the longer form when they are equivalent? Is this just a general style element?

reitermarkus: The reason is that `be_file` is not grammatically correct.

reitermarkus:
Thanks!

PR #2442, opened 2017-04-02
Virtualenv relocation fixes

tdsmith:
Fix a couple of issues discovered in https://github.com/Homebrew/homebrew-core/issues/9424 that affect the relocatability of virtualenv formulae.

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -16,7 +16,9 @@ class Keg
       link = file.readlink
       # Don't fix relative symlinks
       next unless link.absolute?
-      next unless link.to_s.start_with?(HOMEBREW_CELLAR.to_s) || link.to_s.start_with?(HOMEBREW_PREFIX.to_s)
+      link_starts_cellar = link.to_s.start_with?(HOMEBREW_CELLAR.to_s)
+      link_starts_prefix = link.to_s.start_with?(HOMEBREW_PREFIX.to_s)
+      next if !link_starts_cellar && !link_starts_prefix
       new_src = link.relative_path_from(file.parent)
       file.unlink
       FileUtils.ln_s(new_src, file)
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -97,7 +97,12 @@ class Keg
   end
 
   def each_unique_file_matching(string)
-    Utils.popen_read("/usr/bin/fgrep", "-lr", string, to_s) do |io|
+    bsd = `/usr/bin/fgrep -V`.include?("BSD grep")
+    grep_args = "-lr"
+    # Don't recurse into symlinks; the man page says this is the default, but
+    # it's wrong.
+    grep_args += "O" if bsd
+    Utils.popen_read("/usr/bin/fgrep", grep_args, string, to_s) do |io|
       hardlinks = Set.new
 
       until io.eof?
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -98,13 +98,14 @@ class Keg
     []
   end
 
+  def recursive_fgrep_args
+    # for GNU grep; overridden for BSD grep on OS X
+    "-lr"
+  end
+  alias generic_recursive_fgrep_args recursive_fgrep_args
+
   def each_unique_file_matching(string)
-    bsd = `/usr/bin/fgrep -V`.include?("BSD grep")
-    grep_args = "-lr"
-    # Don't recurse into symlinks; the man page says this is the default, but
-    # it's wrong.
-    grep_args += "O" if bsd
-    Utils.popen_read("/usr/bin/fgrep", grep_args, string, to_s) do |io|
+    Utils.popen_read("/usr/bin/fgrep", recursive_fgrep_args, string, to_s) do |io|
       hardlinks = Set.new
 
       until io.eof?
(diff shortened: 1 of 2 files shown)


Library/Homebrew/keg_relocate.rb
```diff
@@ -97,7 +99,12 @@ def detect_cxx_stdlibs(_options = {})
   end
 
   def each_unique_file_matching(string)
-    Utils.popen_read("/usr/bin/fgrep", "-lr", string, to_s) do |io|
+    bsd = `/usr/bin/fgrep -V`.include?("BSD grep")
```
MikeMcQuaid: Favour `Utils.popen_read` over backticks.

MikeMcQuaid: Also I reckon BSD `grep` logic should live in https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/os/mac/keg_relocate.rb or the converse in os/linux

Library/Homebrew/keg_relocate.rb
```diff
@@ -17,7 +17,9 @@ def fix_dynamic_linkage
       # Don't fix relative symlinks
       next unless link.absolute?
       if link.to_s.start_with?(HOMEBREW_CELLAR.to_s) || link.to_s.start_with?(HOMEBREW_PREFIX.to_s)
-        FileUtils.ln_sf(link.relative_path_from(file.parent), file)
+        new_src = link.relative_path_from(file.parent)
+        file.unlink
+        FileUtils.ln_s(new_src, file)
```
MikeMcQuaid: How does this differ from before? Not disagreeing, just wanting to understand.

tdsmith: Oh, I mentioned in the commit message; shoulda copied it into the PR body:
```
ln_sf does the right thing when `dest` is a symlink pointing to a file:
the symlink gets overwritten with a link pointing to the new src. But
when dest points to a directory, we create a new symlink inside the
folder dest points to, which doesn't help us at all.
```

Library/Homebrew/keg_relocate.rb
tdsmith: I think I'd have to copy the body of the method into both os/mac/keg_relocate.rb and os/linux/keg_relocate.rb and just change the arguments that are being passed to grep. Is that right? It feels repetitive.

Why not backticks, out of curiosity? Consistency or something more subtle?

tdsmith: (or do I want to write e.g. a private each_unique_file_matching_grep_args method for each OS?)

Library/Homebrew/keg_relocate.rb
MikeMcQuaid: Interesting, cheers 👍

tdsmith:
Need to look at the tests:

```
Failures:

  1) brew bottle builds a bottle for the given Formula
     Failure/Error:
       expect { brew "bottle", "--no-rebuild", "testball" }
         .to output(/testball-0\.1.*\.bottle\.tar\.gz/).to_stdout
         .and not_to_output.to_stderr
         .and be_a_success

          expected block to not output to stderr, but output "Error: No such file or directory - /usr/bin/fgrep\n"
```

Library/Homebrew/keg_relocate.rb
MikeMcQuaid: > Why not backticks, out of curiosity? Consistency or something more subtle?

I dunno, it's just a pattern we've been following here for a while. I think we handle them slightly differently. I don't care enormously, to be honest.

> I think I'd have to copy the body of the method into both os/mac/keg_relocate.rb and os/linux/keg_relocate.rb and just change the arguments that are being passed to grep. Is that right? It feels repetitive.

Ideally use a separate `grep_arguments` method for just getting the arguments which should make it pretty small. Up to you to decide what the `generic_grep_arguments` version should be but I think there's only a need for a single override e.g. assume BSD here and override in `os/linux` or assume GNU here and override in `os/mac`.

Library/Homebrew/keg_relocate.rb
```diff
@@ -16,9 +16,10 @@ def fix_dynamic_linkage
       link = file.readlink
       # Don't fix relative symlinks
       next unless link.absolute?
-      if link.to_s.start_with?(HOMEBREW_CELLAR.to_s) || link.to_s.start_with?(HOMEBREW_PREFIX.to_s)
-        FileUtils.ln_sf(link.relative_path_from(file.parent), file)
-      end
+      next unless link.to_s.start_with?(HOMEBREW_CELLAR.to_s) || link.to_s.start_with?(HOMEBREW_PREFIX.to_s)
```
MikeMcQuaid: Can you make this two `next unless`s for line length and for stupid Mike who can't handle `unless ||`.

Last thing and otherwise looks 👍

tdsmith: I don't think I can; `next if a || b` can be written `next if a; next if b` and `next unless a && b` can be written `next unless a; next unless b` but I don't see what to do here. My boolean logic or my imagination might be failing me.

monouser7dig: next if !a && !b would work and is maybe easier

MikeMcQuaid: > My boolean logic or my imagination might be failing me.

MikeMcQuaid: > My boolean logic or my imagination might be failing me.

No but mine was which is why I had `unless ||` 😁 

Yeh, `next if !a && !b` and perhaps putting them into variables for length is a little easier to follow.

MikeMcQuaid:
Thanks @tdsmith!

PR #2462, opened 2017-04-07
Rescue from multiple installation attempts for dependencies

scpeters:
This fixes an early termination of brew install when a dependency tree includes a formula twice, such as the original formula and an alias.

-----

I maintain a tap that has a formula that depends on `qt5`. I noticed that the `qt5` (alias `qt`) was renamed to `qt` (alias `qt5`) in https://github.com/Homebrew/homebrew-core/pull/8334. There's now an audit warning to change our dependency to the new name `qt` instead of the alias, which is a helpful notice. I'll do that soon and rebuild our bottles.

There is, however, some weird behavior when installing a formula like `osrf/simulation/gazebo8` that hasn't yet been updated to use the new dependency name. This is because `gazebo8` also depends on `qwt` in core, which depends on `qt`. So `brew deps` shows `gazebo8` depending on both `qt` and `qt5`:

~~~
$ brew deps osrf/simulation/gazebo8 | grep ^q
qt
qt5
qwt
~~~

Then when I `brew install gazebo8` on a fresh system, it installs the dependencies up until `qt`, then just stops, without an error code. So it installs `qt5` first, then it attempts to install `qt` and a `FormulaInstallationAlreadyAttemptedError` exception is raised. That exception is explicitly rescued in [cmd/install.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/install.rb#L321-L323) (see 5c799ef8c8720306a9dfb4db1e47f647e06f9857 and https://github.com/Homebrew/legacy-homebrew/issues/16957), so installation of `gazebo8` is terminated and considered successful at that point. Invoking `brew install gazebo8` once more will resume where it left off and successfully complete the installation.

This PR adds a rescue for that same exception to `install_dependency` so that installation can continue the first time. It currently allows it to just proceed. We could consider printing an appropriate console message and raising a different exception that wouldn't be rescued by `cmd/install.rb` if you'd rather not sweep this behavior under the rug.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -531,6 +531,9 @@ class FormulaInstaller
     oh1 "Installing #{formula.full_name} dependency: #{Formatter.identifier(dep.name)}"
     fi.install
     fi.finish
+  rescue FormulaInstallationAlreadyAttemptedError
+    # We already attempted to install df as part of the dependency tree of
+    # another formula. In that case, don't generate an error, just move on.
   rescue Exception
     ignore_interrupts do
       tmp_keg.rename(installed_keg) if tmp_keg && !installed_keg.directory?
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -534,6 +534,7 @@ class FormulaInstaller
   rescue FormulaInstallationAlreadyAttemptedError
     # We already attempted to install df as part of the dependency tree of
     # another formula. In that case, don't generate an error, just move on.
+    nil
   rescue Exception
     ignore_interrupts do
       tmp_keg.rename(installed_keg) if tmp_keg && !installed_keg.directory?


scpeters:
Apparently `brew style` does not agree with my proposal to ignore this exception:

~~~
Library/Homebrew/formula_installer.rb:534:3: W: Do not suppress exceptions.
  rescue FormulaInstallationAlreadyAttemptedError
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
~~~

I can switch to a console message and raising a different exception that won't be rescued by `cmd/install.rb` if that is preferred.

Library/Homebrew/formula_installer.rb
```diff
@@ -531,6 +531,9 @@ def install_dependency(dep, inherited_options)
     oh1 "Installing #{formula.full_name} dependency: #{Formatter.identifier(dep.name)}"
     fi.install
     fi.finish
+  rescue FormulaInstallationAlreadyAttemptedError
+    # We already attempted to install df as part of the dependency tree of
+    # another formula. In that case, don't generate an error, just move on.
```
MikeMcQuaid: Add a `nil` on the line below this.

MikeMcQuaid:
> This PR adds a rescue for that same exception to install_dependency so that installation can continue the first time. It currently allows it to just proceed. We could consider printing an appropriate console message and raising a different exception that wouldn't be rescued by cmd/install.rb if you'd rather not sweep this behavior under the rug.

This seems like the right behaviour for your case but I'm not sure about just always suppressing this exception rather than e.g. detecting an alias was used. It feels like the right solution for this may be wherever the dependency tree is expanded to fix that these compute as two dependencies rather than one.

Library/Homebrew/formula_installer.rb
scpeters: 41cd60c95c35610dacc5243b331150a0b2205376

scpeters:
The `merge_repeats` function in dependency.rb seems like the right place to eliminate duplicate Aliases, but I'm not sure when the Aliases get resolved.

* https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dependency.rb#L132

scpeters:
I think the issue is that you can't know if dependencies are aliases of each other until you instantiate their formulae. I see both `qt` and `qt5` with `brew deps gazebo8` that I don't see with `brew deps gazebo8 --full-name`, since the `--full-name` argument [converts all the dependencies to formulae and then gets their name](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/deps.rb#L76).

Maybe we could avoid this problem if the [compute_dependencies function in formula_installer.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula_installer.rb#L367-L375) converted all the dependencies to formulae then called `uniq` on the `:name`?

MikeMcQuaid:
I appreciate the work on this but I think unfortunately for now this should be considered a formulae bug rather than Homebrew/brew bug until we've got an approach that doesn't cause too much overhead and can specifically target this issue. Sorry 😭

scpeters:
No problem, thanks for the review

PR #2465, opened 2017-04-08
audit: audit_components method to rubocops and tests

GauthamGoli:

-----
This is in continuation of the PR #2242 and issue #569 .
This PR ports `audit_components` https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L251-L309 
 method to custom cops.


Library/Homebrew/rubocops/components_order_cop.rb
```diff
@@ -0,0 +1,71 @@
...
+  module Cop
+    module Homebrew
+      # This cop checks for correct order of components in a Formula
+      #
+      # - component_precedence_list has component hierarchy in a nested list
+      #   where each sub array contains components' details which are at same precedence level
+      class FormulaComponentsOrder < FormulaCop
+        def audit_formula(_node, _class_node, _parent_class_node, formula_class_body_node)
+          component_precedence_list = [
+            [{ "name" => :include,  "type" => "method" }],
```
MikeMcQuaid: Random comment that symbols for both keys and values here might be nice if it's not overly complex? Also, what about `method_call` vs. `method_definition` vs. `block_call` or something?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -74,6 +102,35 @@ def method_called?(node, method_name)
           false
         end
 
+        def method_called?(node, method_name)
+          # Check if method_name is called among the direct children nodes in the given node
```
MikeMcQuaid: I'd put these comments above rather than below the method definitions.

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -74,6 +102,35 @@ def method_called?(node, method_name)
...
+            return true
+          end
+          false
+        end
+
+        def check_precedence(first_nodes, next_nodes)
+          # Checks for precedence, returns the first pair of precedence violating nodes
+          next_nodes.each do |each_next_node|
+            first_nodes.each do |each_first_node|
+              return [each_first_node, each_next_node] if component_precedes?(each_first_node, each_next_node)
```
MikeMcQuaid: Generally lines over 80 characters (or thereabouts) are to be avoided when possible.

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -74,6 +102,35 @@ def method_called?(node, method_name)
...
+            end
+          end
+          nil
+        end
+
+        def component_precedes?(first_node, next_node)
+          # If first node does not precede next_node, sets appropriate instance variables for reporting
+          return false unless line_number(first_node) > line_number(next_node)
```
MikeMcQuaid: Flip the `>` to make it an `if`

Library/Homebrew/rubocops/components_order_cop.rb
GauthamGoli: `method_call`, `block_call` and `method_definition` nodes are of type `method`, `block` and `def` respectively in Rubocop. Do you mean, we should have something more explicit for `type` values in `component_precedence_list`?

MikeMcQuaid: Oh, that's fine then, wasn't clear they were internal types. Still think symbols (at least for keys) would be great.

GauthamGoli: I updated them with symbols and also used `method_call` ,`block_call` and `method_definition` instead, as we want to hide the rubocop specific details in `formula_cop.rb` itself.

MikeMcQuaid:
Thanks again @GauthamGoli!

GauthamGoli:
@MikeMcQuaid The violating `block`/`method`/ `def` s in the editor are just highlighted by their at the names. Do we want the whole block of code to get highlighted instead?

MikeMcQuaid:
@GauthamGoli My editor doesn't support the highlighting so I think the current version is fine, personally. Up to you, though!

PR #2485, opened 2017-04-11
sandbox: implement some read protections

DomT4:

-----

Way way way too many things still require you to enter passwords into the terminal & not do anything to mask that, which then gets stashed away in your shell history, as well as any servers you `ssh` into, etc. There's a lot of sensitive information in there, basically.

Since Homebrew doesn't police network usage during builds better to deny read completely & close the small loophole. In reality almost everything in `HOME` should be blocked off for reading during builds, but there's a bunch of people who prefer to put Homebrew in `HOME` which this would screw with if we went that broad. We should get there eventually but compatibility baby steps.

Also gets rid of this exceptionally annoying message when you debug builds:
```
zsh: locking failed for /Users/xyz/.zsh_history: operation not permitted: reading anyway
```

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -669,7 +669,15 @@ class FormulaInstaller
         sandbox = Sandbox.new
         formula.logs.mkpath
         sandbox.record_log(formula.logs/"build.sandbox.log")
-        sandbox.allow_write_path(ENV["HOME"]) if ARGV.interactive?
+
+        if ARGV.interactive?
+          # The allow_read here is superfluous for now.
+          sandbox.allow_read_path(ENV["HOME"])
+          sandbox.allow_write_path(ENV["HOME"])
+        else
+          sandbox.deny_read_broad_home
+        end
+
         sandbox.allow_write_temp_and_cache
         sandbox.allow_write_log(formula)
         sandbox.allow_write_xcode
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -105,6 +105,34 @@ class Sandbox
     end
   end
 
+  def deny_read_shell_files
+    # Configuration.
+    # deny_read "/Users/#{ENV["USER"]}/.bash_profile"
+    # deny_read "/Users/#{ENV["USER"]}/.cshrc"
+    # deny_read "/Users/#{ENV["USER"]}/.config/fish/config.fish"
+    # deny_read "/Users/#{ENV["USER"]}/.kshrc"
+    # deny_read "/Users/#{ENV["USER"]}/.tcshrc"
+    # deny_read "/Users/#{ENV["USER"]}/.zshrc"
+
+    # Maybe this way is cleaner, although less explicit.
+    # Should this use a "/[^/]+/" regex instead of explicit USER to ensure
+    # we're not just protecting the active user on the machine but all of them?
+    %w[bash_profile cshrc config/fish/config.fish kshrc tcshrc zshrc].each do |s|
+      deny_read "/Users/#{ENV["USER"]}/.#{s}"
+    end
+
+    # Same as above. Seek review on preference.
+    # History.
+    # deny_read "/Users/#{ENV["USER"]}/.bash_history"
+    # deny_read "/Users/#{ENV["USER"]}/.zsh_history"
+    # deny_read "/Users/#{ENV["USER"]}/.config/fish/fish_history"
+    # deny_read "/Users/#{ENV["USER"]}/.sh_history"
+    # deny_read "/Users/#{ENV["USER"]}/.history"
+    %w[bash_history zsh_history config/fish/fish_history sh_history history].each do |s|
+      deny_read "/Users/#{ENV["USER"]}/.#{s}"
+    end
+  end
+
   def exec(*args)
     seatbelt = Tempfile.new(["homebrew", ".sb"], HOMEBREW_TEMP)
     seatbelt.write(@profile.dump)
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -134,8 +134,9 @@ class Sandbox
   end
 
   def deny_read_broad_home
-    # Limit this to protecting non-HOME installations for now.
-    if HOMEBREW_PREFIX.to_s != "/Users/#{ENV["USER"]}"
+    # Limit this to protecting non-HOME & human installations for now.
+    if HOMEBREW_PREFIX.to_s != "/Users/#{ENV["USER"]}" &&
+       HOMEBREW_REPOSITORY.to_s != "/Users/#{ENV["USER"]}" && !ENV["CI"]
       # Deny reading to everything under /Users/xyz.
       deny_read "^/Users/#{ENV["USER"]}/[^/]+", type: :regex
       # Don't deny read access to /Users/xyz itself.
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -134,12 +134,21 @@ class Sandbox
   end
 
   def deny_read_broad_home
+    # Limit this to protecting non-HOME installations for now.
     if HOMEBREW_PREFIX.to_s != "/Users/#{ENV["USER"]}"
-      deny_read_path "/Users/#{ENV["USER"]}"
+      # Deny reading to everything under /Users/xyz.
+      deny_read "^/Users/#{ENV["USER"]}/[^/]+", type: :regex
+      # Don't deny read access to /Users/xyz itself.
+      allow_read "/Users/#{ENV["USER"]}"
+      # Need to test this more, but Apple dumps things in here that may need reading.
       allow_read_path "/Users/#{ENV["USER"]}/Library"
-      allow_read_path "/Users/#{ENV["USER"]}/.gem" # Used by Homebrew
+      # Used by Homebrew.
+      allow_read_path "/Users/#{ENV["USER"]}/.gem"
+      # Used by the HEAD install_spec tests.
+      allow_read "/Users/#{ENV["USER"]}/.gitconfig"
+    else
+      deny_read_shell_files
     end
-    deny_read_shell_files
   end
 
   def exec(*args)
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -137,6 +137,7 @@ class Sandbox
     if HOMEBREW_PREFIX.to_s != "/Users/#{ENV["USER"]}"
       deny_read_path "/Users/#{ENV["USER"]}"
       allow_read_path "/Users/#{ENV["USER"]}/Library"
+      allow_read_path "/Users/#{ENV["USER"]}/.gem" # Used by Homebrew
     end
     deny_read_shell_files
   end
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -133,6 +133,14 @@ class Sandbox
     end
   end
 
+  def deny_read_broad_home
+    if HOMEBREW_PREFIX.to_s != "/Users/#{ENV["USER"]}"
+      deny_read_path "/Users/#{ENV["USER"]}"
+      allow_read_path "/Users/#{ENV["USER"]}/Library"
+    end
+    deny_read_shell_files
+  end
+
   def exec(*args)
     seatbelt = Tempfile.new(["homebrew", ".sb"], HOMEBREW_TEMP)
     seatbelt.write(@profile.dump)
--- a/Library/Homebrew/sandbox.rb
+++ b/Library/Homebrew/sandbox.rb
@@ -56,6 +56,22 @@ class Sandbox
     deny_write path, type: :subpath
   end
 
+  def allow_read(path, options = {})
+    add_rule allow: true, operation: "file-read*", filter: path_filter(path, options[:type])
+  end
+
+  def allow_read_path(path)
+    allow_read path, type: :subpath
+  end
+
+  def deny_read(path, options = {})
+    add_rule allow: false, operation: "file-read*", filter: path_filter(path, options[:type])
+  end
+
+  def deny_read_path(path)
+    deny_read path, type: :subpath
+  end
+
   def allow_write_temp_and_cache
     allow_write_path "/private/tmp"
     allow_write_path "/private/var/tmp"
--- a/Library/Homebrew/cmd/postinstall.rb
+++ b/Library/Homebrew/cmd/postinstall.rb
@@ -40,6 +40,7 @@ module Homebrew
         sandbox.allow_write_log(formula)
         sandbox.allow_write_xcode
         sandbox.deny_write_homebrew_repository
+        sandbox.deny_read_broad_home
         sandbox.allow_write_cellar(formula)
         Keg::TOP_LEVEL_DIRECTORIES.each do |dir|
           sandbox.allow_write_path "#{HOMEBREW_PREFIX}/#{dir}"
--- a/Library/Homebrew/dev-cmd/test.rb
+++ b/Library/Homebrew/dev-cmd/test.rb
@@ -72,6 +72,7 @@ module Homebrew
             sandbox.allow_write_path(HOMEBREW_PREFIX/"var/cache")
             sandbox.allow_write_path(HOMEBREW_PREFIX/"var/log")
             sandbox.allow_write_path(HOMEBREW_PREFIX/"var/run")
+            sandbox.deny_read_broad_home
             sandbox.exec(*args)
           else
             exec(*args)


Library/Homebrew/sandbox.rb
```diff
@@ -165,6 +162,10 @@ class SandboxProfile
           (regex #"^/dev/ttys?[0-9]*$")
           )
       (deny file-write*) ; deny non-whitelist file write operations
+      (deny file-read*
+          (regex #"^/Users/[a-zA-Z0-9_]+/.*_history$")
+          (regex #"^/Users/[a-zA-Z0-9_]+/.config/fish/fish_history")
```
ilovezfs: does the first regex not already match fish_history?

DomT4: Yes, it does. This is a leftover from when the first regex was more specific.

Library/Homebrew/sandbox.rb
```diff
@@ -165,6 +162,9 @@ class SandboxProfile
           (regex #"^/dev/ttys?[0-9]*$")
           )
       (deny file-write*) ; deny non-whitelist file write operations
+      (deny file-read*
+          (regex #"^/Users/[a-zA-Z0-9_]+/.*_history$")
+          )
```
xu-cheng: Hi, here are some advice from me:
* Using `[a-zA-Z0-9_]+` to match user name is bad. Because in fact, user name can have many other arbitrary characters, e.g. Unicode chars in other language. It is recommend to use `/[^/]+/` to match a directory.
* Using `.*_history$` is way too general. It can even match path like `~/Homebrew/Cellar/foo/1.0/.../some_script_end_with_history`. You get the idea. It would be better to specific the common shell history files verbosely without using `.*` regex.
* I would not recommend to add this rule directly in `SandboxProfile` base template. Instead, can you create `Sandbox#deny_read_shell_history` method? This can also be helpful, say not enforcing such rule under `--interactive` mode.

DomT4: Hey Xu, good to see you again. Thanks for the advice.

1 & 2 are perfectly fair points. I did actually have them explicitly written at first but I wanted to avoid introducing large clumps of code, reasoning the more I wrote the harder it would be to get through review 😅.

> I would not recommend to add this rule directly in SandboxProfile base template. Instead, can you create Sandbox#deny_read_shell_history method? This can also be helpful, say not enforcing such rule under --interactive mode.

My broader plan was to introduce a `deny_read` & related method, add shell history as well as user downloads/pictures/documents/etc whilst not blocking off the whole of `$HOME` entirely as obviously there are Homebrew users who live there. But I was wary about getting that through review. But you're right, I guess if I'm going to do it I might as well do it properly.

MikeMcQuaid: > It is recommend to use /[^/]+/ to match a directory.

👍 



> It would be better to specific the common shell history files verbosely without using .* regex.

👍 

> I would not recommend to add this rule directly in SandboxProfile base template. Instead, can you create Sandbox#deny_read_shell_history method? This can also be helpful, say not enforcing such rule under --interactive mode.

👍

DomT4:
@MikeMcQuaid @ilovezfs Can someone with the access to do so reopen this please? It was incorrectly closed by Mike's commit message from https://github.com/Homebrew/brew/issues/2502 😄.

DomT4:
@ilovezfs Much appreciated ❤️

MikeMcQuaid:
Apologies!

DomT4:
No worries! It happens 😄. Will try to push some commits here later today.

DomT4:
This is the next big thing on my to-do list now the caveats PR is done with. Likely won't be today but should see some action tomorrow. Mostly posting this so the bot that prowls around & tells people off for leaving Issues/PRs inactive doesn't flag this one 😄.

MikeMcQuaid:
@DomT4 You may also be interested in https://github.com/Homebrew/brew/pull/2617

MikeMcQuaid:
@DomT4 in case you didn't see this is probably vaguely interesting: https://github.com/Homebrew/brew/pull/2617

DomT4:
Yeah noticed that earlier & thought about it a bit in this context. I don't think it'll directly impact this too much, but I'll remember to test against your PR if it doesn't get merged by the time I tweak this.

DomT4:
Made changes locally, but still testing through & will need to write up some tests to minimise the risk of breaking things weirdly. Hoping to be happy enough to open it up to review later today 😅

DomT4:
@probot-stale[bot] Thanks for the reminder, I'm a bad human.

DomT4:
This still needs a bunch of review, not least because I've sprinkled it with code comments requesting precisely that heh, and it needs `brew` tests added, and I want to install a bunch more formulae locally under it first, but thoughts welcome on direction.

Never let it be said that I won't _eventually_ continue work on PRs 😅.

Library/Homebrew/sandbox.rb
```diff
@@ -89,6 +105,43 @@ def deny_write_homebrew_repository
...
+      deny_read "/Users/#{ENV["USER"]}/.#{s}"
+    end
+  end
+
+  def deny_read_broad_home
+    if HOMEBREW_PREFIX.to_s != "/Users/#{ENV["USER"]}"
+      deny_read_path "/Users/#{ENV["USER"]}"
+      allow_read_path "/Users/#{ENV["USER"]}/Library"
+      allow_read_path "/Users/#{ENV["USER"]}/.gem" # Used by Homebrew
+    end
```
DomT4: This should probably be an `else` instead of an `end` since otherwise the above `deny_read_path "/Users/#{ENV["USER"]}"` should make it redundant.

DomT4:
I'm not really sure what Homebrew is doing with `/Users/xyz` that causes the tests to erupt when `deny_read_path "/Users/#{ENV["USER"]}"` is in effect. Trying to work out what it wants to read that I've not already whitelisted for one reason or another.

MikeMcQuaid:
@DomT4 If I were in your 👞 I'd be getting one of the tests that's displaying `Operation not permitted` in the `stderr` output and fiddle with it so you can read the whole `stderr` output and that'll show you what it's not letting you read.

DomT4:
I had a hunch overnight that looks like it may be correct, in terms of the tests wanting to read `/Users/xyz` itself, which the output looks like it's confirming.

I should be able to change it to deny reading to everything _under_ `/Users/xyz` rather than `/Users/xyz` itself which I'm hoping will resolve the issue. There's a ton of different ways to discover username on macOS & there's no reasonable reason to deny things the ability to check `HOME` or `USER` exist.

DomT4:
Still very much a WIP but latest push should fix the existing tests.

DomT4:
Apparently not on CI. I'm getting consistently clean runs locally now. Hmm.

Library/Homebrew/formula_installer.rb
```diff
@@ -669,7 +669,15 @@ def build
         sandbox = Sandbox.new
         formula.logs.mkpath
         sandbox.record_log(formula.logs/"build.sandbox.log")
-        sandbox.allow_write_path(ENV["HOME"]) if ARGV.interactive?
+
+        if ARGV.interactive?
+          # The allow_read here is superfluous for now.
```
MikeMcQuaid: Can you elaborate on this comment? Not sure I understand what it means.

Library/Homebrew/sandbox.rb
```diff
@@ -89,6 +105,53 @@ def deny_write_homebrew_repository
     end
   end
 
+  def deny_read_shell_files
```
MikeMcQuaid: What is there that's sensitive in these files?

Library/Homebrew/sandbox.rb
```diff
@@ -89,6 +105,53 @@ def deny_write_homebrew_repository
...
+    # deny_read "/Users/#{ENV["USER"]}/.cshrc"
+    # deny_read "/Users/#{ENV["USER"]}/.config/fish/config.fish"
+    # deny_read "/Users/#{ENV["USER"]}/.kshrc"
+    # deny_read "/Users/#{ENV["USER"]}/.tcshrc"
+    # deny_read "/Users/#{ENV["USER"]}/.zshrc"
+
+    # Maybe this way is cleaner, although less explicit.
+    # Should this use a "/[^/]+/" regex instead of explicit USER to ensure
+    # we're not just protecting the active user on the machine but all of them?
+    %w[bash_profile cshrc config/fish/config.fish kshrc tcshrc zshrc].each do |s|
```
MikeMcQuaid: What is there that's sensitive in these files?

Library/Homebrew/sandbox.rb
```diff
@@ -89,6 +105,53 @@ def deny_write_homebrew_repository
...
+    # deny_read "/Users/#{ENV["USER"]}/.zsh_history"
+    # deny_read "/Users/#{ENV["USER"]}/.config/fish/fish_history"
+    # deny_read "/Users/#{ENV["USER"]}/.sh_history"
+    # deny_read "/Users/#{ENV["USER"]}/.history"
+    %w[bash_history zsh_history config/fish/fish_history sh_history history].each do |s|
+      deny_read "/Users/#{ENV["USER"]}/.#{s}"
+    end
+  end
+
+  def deny_read_broad_home
```
MikeMcQuaid: I don't understand what the `broad` here means

Library/Homebrew/sandbox.rb
```diff
@@ -89,6 +105,53 @@ def deny_write_homebrew_repository
...
+    # Limit this to protecting non-HOME & human installations for now.
+    if HOMEBREW_PREFIX.to_s != "/Users/#{ENV["USER"]}" &&
+       HOMEBREW_REPOSITORY.to_s != "/Users/#{ENV["USER"]}" && !ENV["CI"]
```
MikeMcQuaid: 👎 to different behaviour for CI here.

Library/Homebrew/sandbox.rb
```diff
@@ -89,6 +105,53 @@ def deny_write_homebrew_repository
...
+    end
+
+    # Same as above. Seek review on preference.
+    # History.
+    # deny_read "/Users/#{ENV["USER"]}/.bash_history"
+    # deny_read "/Users/#{ENV["USER"]}/.zsh_history"
+    # deny_read "/Users/#{ENV["USER"]}/.config/fish/fish_history"
+    # deny_read "/Users/#{ENV["USER"]}/.sh_history"
+    # deny_read "/Users/#{ENV["USER"]}/.history"
+    %w[bash_history zsh_history config/fish/fish_history sh_history history].each do |s|
```
MikeMcQuaid: Given the PR description I think these are the only files we need to handle, here.

Library/Homebrew/formula_installer.rb
DomT4: It's an outdated code comment & should've been wiped, apologies.

Library/Homebrew/sandbox.rb
DomT4: I talked about various things further up. I have mixed feelings about the shell configuration files & fairly strong feelings about the history files.

In my `zshrc` because I'm lazy I have functions that flip my active GPG key around depending on whether I'm signing personal stuff or signing code stuff. That's not a super novel idea & there's certainly a chunk of folk who have keys they keep non-public & given Homebrew doesn't police network calls during build it seems kinder to be overzealous on presuming everyone is potentially compromised rather than assuming everyone has good intentions & perfect security. Fundamentally we don't _lose_ anything by protecting those files, so. As long as functionality isn't impeded I think a sandbox potentially too broad is a better option than a sandbox potentially too narrow.

On the history files that is a significantly more important one though, for pretty obvious reasons. Not everything masks passwords on the command line, you can discover the software people use, where they spend the most time working around their system from the CLI, the servers they `ssh` into, the things they `wget`, `curl`, `youtube-dl`, etc off the internet, the `gpg` keys they request and by extension some of the people they potentially communicate with, etc.

It also handles this annoying UX issue every time you debug a build in `zsh`, although I suspect the same applies to most history files given the nature of the error:
```
zsh: locking failed for /Users/xyz/.zsh_history: operation not permitted: reading anyway
```

Library/Homebrew/sandbox.rb
DomT4: Discussed above.

Library/Homebrew/sandbox.rb
DomT4: The PR description was an unhappy compromise in the first place, as I think I mentioned before. I wanted to more broadly protect `$HOME` but I didn't think I'd easily get it passed through review given the number of maintainers & contributors interested in this topic, but as it turned out it didn't pass through review anyway, so I took Xu's suggestions & expanded this PR back to its original intention. The PR description is now misleading, but it was always a compromise between me & the reality of actually getting a PR merged.

Library/Homebrew/sandbox.rb
DomT4: Yeah, difficult one to name. Essentially designed to block access to the vast majority of the directories and files under `$HOME` aside from certain exemptions. `broad_home` was working shorthand for `most_of_home`, which isn't ideal either really. If you're not fussed about deviating from the `deny_` and `allow_` language this could be something like `read_protect_user_exclusive_home_elements` or something. I think we're gonna fail for a short tile that is actually accurate & descriptive here though 😅.

Library/Homebrew/sandbox.rb
DomT4: Temporary dance. Need to look into what CI is actually using around `$HOME`. Homebrew's CI setup is relatively complex in structure, from memory, which makes working from that by-now-outdated memory somewhat tricky.

DomT4:
Sorry for the delay in replying to review, have had a fair bit going on lately. Appreciate the comments.

Library/Homebrew/sandbox.rb
```diff
@@ -89,6 +105,53 @@ def deny_write_homebrew_repository
...
+      deny_read "/Users/#{ENV["USER"]}/.#{s}"
```
apjanke: Should we use `${ENV["HOME"]}` or `Etc.getpwuid.dir` instead of `"/Users/#{ENV["USER"]}"` here, in case a user has a home dir in a non-default location?

Library/Homebrew/sandbox.rb
apjanke: The location of the history files for most of these shells is configurable; we should probably respect that.

Should probably check `$HISTFILE`, in case the user has a custom history file location. And `$ZDOTDIR` for `zsh` to cover the case of custom startup file location. (I think just `$HISTFILE` might cover it for `bash` and `zsh`.

As of `fish` 2.3.0, its history file is in `~/.local/share/fish` or under `$XDG_DATA_HOME` instead of under `~/.config` or `$XDG_CONFIG_HOME`. I don't see a `fish` variable to get the location of the history file directly.

apjanke: If we're going to be conservative about what we deny, perhaps we should deny `$HOME/.ssh` as well as the shell history files? That's pretty sure to be sensitive.

MikeMcQuaid:
> In my zshrc because I'm lazy I have functions that flip my active GPG key around depending on whether I'm signing personal stuff or signing code stuff. That's not a super novel idea & there's certainly a chunk of folk who have keys they keep non-public & given Homebrew doesn't police network calls during build it seems kinder to be overzealous on presuming everyone is potentially compromised rather than assuming everyone has good intentions & perfect security. Fundamentally we don't lose anything by protecting those files, so. As long as functionality isn't impeded I think a sandbox potentially too broad is a better option than a sandbox potentially too narrow.

> The location of the history files for most of these shells is configurable; we should probably respect that.

> If we're going to be conservative about what we deny, perhaps we should deny $HOME/.ssh as well as the shell history files? That's pretty sure to be sensitive.

Given all this I'm increasingly 👎 on the whole idea of this PR. The sandbox doesn't exist to stop the buildsystem from reading random files on disk, it exists to partly stop writing files in random locations and primarily to enforce our prefix/keg system.

If you don't trust software you're installing to not read your configuration files and send them over the network to some malicious source at build-time: why on earth would you trust running the compiled code that can do exactly the same thing without you noticing?

I'm concerned at this point this becomes pretty arbitrary whack-a-mole.

DomT4:
As I have outlined to you in repeated discussions over the years, not just on this topic but on others, my point of view has always been that Homebrew has a strong responsibility to protect within its boundaries.

Yes, we do not protect users from what happens the first time they run a tool, which in some ways is a disappointing side of package management whilst also being a deeply realistic position. Nonetheless we do attempt to protect users with "sane defaults" across the board, whether that's in choosing to pick up upstream/Debian/etc security patches and applying them out-of-band or choosing not to ship known-insecure cryptographic protocols before upstreams formally toss them.

Heck, if my memory doesn't completely fail me, you even consented to turning off `ssh` password authentication for the CI machines that produced bottles out of minor concerns `ssh` passwords have some well-known pitfalls. It's not like Homebrew has never proactively taken steps to close potential holes, and that has never been limited to any specific element of the project at least for as long as I've been lingering.

What happens once the `brew` process and its children have finished executing is beyond Homebrew's remit, I fully appreciate & agree with that point. What happens under the Homebrew umbrella however feels like fair game for considering whether or not things can be tightened up.

DomT4:
Ultimately though, above comment aside, I'm gonna let it be here. I don't foresee a ton of positive feelings on either side if this discussion continues so what's the point in us both ending up there if it achieves nothing. Thanks for the review folks.

MikeMcQuaid:
> It's not like Homebrew has never proactively taken steps to close potential holes, and that has never been limited to any specific element of the project at least for as long as I've been lingering.

There's a pretty big difference between closing known attacks that haven't been used against us and closing unknown attacks that don't seem to have been used against anyone.

In this case, to make this attack you'd need to compromise upstream's source code. If you were going to do that why would you sabotage e.g. the Makefile where this could be dramatically easier to detect and stop than the runtime of the application?

> What happens under the Homebrew umbrella however feels like fair game for considering whether or not things can be tightened up.

The point worth remembering (that itself may be worth changing and making a PR for): as-is homebrew/core is the only tap that has the sandbox enabled and it also has bottles for pretty much anything that requires a build-system.

MikeMcQuaid:
> I'm gonna let it be here.

That's fine. I felt it was necessary to respond to the points you'd made in this PR should others want to do something similar in future so they have an understanding of my position.

Issue #2439, opened 2017-04-02
[wish] date-based versioning in the RCF 3339 format


retokromer:
Currently date based versioning using RFC 3339 (e.g. `v2017-04-02`) is cut (in the example to `04-02` i.e. `/usr/local/Cellar/my_formula/04-02` instead of `/usr/local/Cellar/my_formula/2017-04-02`). Only the less readable writing without separation (`v20170402` i.e. `/usr/local/Cellar/my_formula/20170402`) or with the "wrong" separation (`v2017.04.02` i.e. `/usr/local/Cellar/my_formula/2017.04.02`) seem to work. I suggest to allow also the "correct" separation character.

MikeMcQuaid:
We interpret everything internally using semantic versioning, essentially so this isn't really possible. You can set `version` to whatever you want, though, it's just comparisons may not work as expected.

retokromer:
Thank you for the feedback!

retokromer:
@MikeMcQuaid If I try to add the regex for this in https://github.com/Homebrew/brew/blob/master/Library/Homebrew/version.rb and open a PR, will this be considered or I am loosing my time? I am fine either way, just asking before starting.

MikeMcQuaid:
@retokromer If you can add a regex and tests then it may be accepted, yeh.

PR #2490, opened 2017-04-14
add date-based versioning

retokromer:

As suggested in https://github.com/Homebrew/brew/issues/2439 adds date-based versioning (RFC 3339).

### DIFF
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -344,6 +344,11 @@ class Version
     m = /[-v]((?:\d+\.)*\d+)$/.match(spec_s)
     return m.captures.first unless m.nil?
 
+    # date-based versioning
+    # e.g. ltopers-v2017-04-14.tar.gz
+    m = /-v?((?:\d{4}-\d{2}-\d{2}))/.match(stem)
+    return m.captures.first unless m.nil?
+
     # e.g. lame-398-1
     m = /-((?:\d)+-\d+)/.match(stem)
     return m.captures.first unless m.nil?
@@ -407,11 +412,6 @@ class Version
 
     # e.g. http://www.ijg.org/files/jpegsrc.v8d.tar.gz
     m = /\.v(\d+[a-z]?)/.match(stem)
-    return m.captures.first unless m.nil?
-
-    # date-based versioning
-    # e.g. ltopers-v2017-04-14.tar.gz
-    m = /-v?((?:\d{4}-\d{2}-\d{2}))/.match(stem)
     return m.captures.first unless m.nil?
   end
 end
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -408,6 +408,11 @@ class Version
     # e.g. http://www.ijg.org/files/jpegsrc.v8d.tar.gz
     m = /\.v(\d+[a-z]?)/.match(stem)
     return m.captures.first unless m.nil?
+
+    # date-based versioning
+    # e.g. ltopers-v2017-04-14.tar.gz
+    m = /-v?((?:\d{4}-\d{2}-\d{2}))/.match(stem)
+    return m.captures.first unless m.nil?
   end
 end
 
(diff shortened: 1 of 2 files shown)


Library/Homebrew/version.rb
```diff
@@ -408,6 +408,11 @@ def self._parse(spec)
     # e.g. http://www.ijg.org/files/jpegsrc.v8d.tar.gz
     m = /\.v(\d+[a-z]?)/.match(stem)
     return m.captures.first unless m.nil?
+
+    # date-based versioning
+    # e.g. ltopers-v2017-04-14.tar.gz
```
MikeMcQuaid: Please add a test for this case.

MikeMcQuaid: (in https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/version_spec.rb)

retokromer: I’ve to delve a little deeper into that, but cannot do this immediately.

MikeMcQuaid: It's because the ordering means this will be caught by another regex first.

retokromer: Thank you for the hint!

Library/Homebrew/version.rb
```diff
@@ -344,6 +344,11 @@ def self._parse(spec)
     m = /[-v]((?:\d+\.)*\d+)$/.match(spec_s)
     return m.captures.first unless m.nil?
 
+    # date-based versioning
+    # e.g. ltopers-v2017-04-14.tar.gz
+    m = /-v?((?:\d{4}-\d{2}-\d{2}))/.match(stem)
```
reitermarkus: You can remove the inner uncaptured group.

Library/Homebrew/version.rb
```diff
@@ -344,6 +344,11 @@ def self._parse(spec)
     m = /[-v]((?:\d+\.)*\d+)$/.match(spec_s)
     return m.captures.first unless m.nil?
 
+    # date-based versioning
+    # e.g. ltopers-v2017-04-14.tar.gz
+    m = /-v?(?:\d{4}-\d{2}-\d{2})/.match(stem)
```
reitermarkus: Now you removed the outer group, it should be `/-v?(\d{4}-\d{2}-\d{2})/`

retokromer: Sorry!

MikeMcQuaid:
Looks good, thanks @retokromer!

ilovezfs:
Seems like this cannot handle `re2`'s date-based versions (https://github.com/google/re2/releases). For instance `2017-05-01` parses as `05-01`.

retokromer:
Hmm… That was the behavior before the change! Possibly the regex needs to be moved before https://github.com/retokromer/brew/blob/3585dfd9bd84005ba20e4a97802dea13fb82c63e/Library/Homebrew/version.rb#L335 or so. Sorry, I’m travelling today, but will delve into this later.

ilovezfs:
@retokromer cool, thanks. Also, I notice that
```
irb(main):001:0> Version.new("2017-05-01") > Version.new("20170401`")
=> false
```
It might be useful to have that compare as true to avoid having to bump `version_scheme` every time we switch from a `20170401` style version to a `2017-05-01` style version.

retokromer:
@ilovezfs Moving the test at the beginning of the tests seems to work as expected. Now I’ll try to find and understand the comparison code.

MikeMcQuaid:
Problem with moving it to the beginning is the potential to break versions on other formulae unless the regex is very specific.

retokromer:
In my opinion, it is indeed very specific: `-v?(\d{4}-\d{2}-\d{2})`, i.e. the hard coded date as in RFC 3339.
The problem is that the current first regex matches the second dash of the date…
I’m still travelling, but I’ll try to propose a solution that solves also the remark above https://github.com/Homebrew/brew/pull/2490#issuecomment-298304347.

retokromer:
For the moment, I made the PR https://github.com/Homebrew/brew/pull/2591 with just moving this regex at the beginning, as I don’t see other regexes with two dashes that could conflict.

PR #2495, opened 2017-04-15
FileUtils.make: don't bypass superenv

mistydemeo:
The intention here was to make sure that `make` or `gmake` is called as appropriate; as long as paths are set up right, superenv shims should call through to the Homebrew copy as appropriate. This should also work in stdenv (e.g., Tiger).

Fixes Homebrew/homebrew-portable#31.

-----

cc @xu-cheng

### DIFF
--- a/Library/Homebrew/extend/fileutils.rb
+++ b/Library/Homebrew/extend/fileutils.rb
@@ -108,13 +108,13 @@ module FileUtils
 
   # Run `make` 3.81 or newer.
   # Uses the system make on Leopard and newer, and the
-  # path to the actually-installed make on Tiger or older.
+  # name of the actually-installed make on Tiger or older.
   def make(*args)
     if Utils.popen_read("/usr/bin/make", "--version").match(/Make (\d\.\d+)/)[1] > "3.80"
-      system "/usr/bin/make", *args
+      system "make", *args
     else
       make = Formula["make"].opt_bin/"make"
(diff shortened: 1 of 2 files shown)


xu-cheng:
:+1:

xu-cheng:
I think you need to update https://github.com/Homebrew/brew/blob/master/Library/Homebrew/shims/super/make#L4 to:
```
exec xcrun "$0 "$@"
```

xu-cheng:
I think this may not work. As calling `gmake` who is symlinked to `make` in superenv is equivalent  to calling normal `make`

mistydemeo:
> I think you need to update https://github.com/Homebrew/brew/blob/master/Library/Homebrew/shims/super/make#L4

Good catch, changed.

xu-cheng:
> > I think you need to update https://github.com/Homebrew/brew/blob/master/Library/Homebrew/shims/super/make#L4
>
> Good catch, changed.

Unfortunately, I think this is incorrect. As it will break normal formulae. If a formula is calling `gmake` and `bsdmake` on modern macOS, superenv will change it to `make`, since both `gmake` and `bsdmake` do not exist on macOS. However, this change will break such behaviour.

mistydemeo:
Hm, yeah. That's not great for existing things that call `gmake` - do we know if that's something we regularly encounter?

mistydemeo:
Hm okay, yes it does; redirecting just `gmake` breaks cmake test executables, which do call `gmake`.

PR #2496, opened 2017-04-15
Hardware cpu can run arch helper

mistydemeo:

-----

This adds a helper that returns whether or not the active CPU is capable of running an executable of a given architecture. This is useful for cross-compiling; for example, in homebrew-portable, we can end up building a library for an architecture the host CPU can't run, which prevents `make test` and similar things from failing. A practical example is building a universal portable-openssl on a PowerPC host: it can build Intel binaries, but `make test` will fail since it can't run them.

cc @xu-cheng

### DIFF
--- a/Library/Homebrew/extend/os/mac/hardware/cpu.rb
+++ b/Library/Homebrew/extend/os/mac/hardware/cpu.rb
@@ -107,6 +107,20 @@ module Hardware
         end
       end
 
+      # Determines whether the current CPU and macOS combination
+      # can run an executable of the specified architecture.
+      # `arch` is a symbol in the same format returned by
+      # Hardware::CPU.family
+      def can_run?(arch)
+        if Hardware::CPU.intel?
+          intel_can_run? arch
+        elsif Hardware::CPU.ppc?
+          ppc_can_run? arch
+        else
+          false
+        end
+      end
+
       def features
         @features ||= sysctl_n(
           "machdep.cpu.features",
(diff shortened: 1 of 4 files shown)


Library/Homebrew/extend/os/mac/hardware/cpu.rb
```diff
@@ -162,6 +174,33 @@ def sysctl_n(*keys)
...
+          # Rosetta is still available
+          MacOS.version < :lion
+        when :ppc64
+          # Rosetta never supported PPC64
+          false
+        when :x86_64
+          Hardware::CPU.is_64_bit?
+        else # i386
+          true
+        end
```
xu-cheng: Would it be better to have:
```
when :i386
  true
else # :dunno
  false
end
```

Library/Homebrew/extend/os/mac/hardware/cpu.rb
```diff
@@ -107,6 +107,18 @@ def universal_archs
...
+      # Determines whether the current CPU and macOS combination
+      # can run an executable of the specified architecture.
+      # `arch` is a symbol in the same format returned by
+      # Hardware::CPU.family
+      def can_run? arch
+        if Hardware::CPU.intel?
+          intel_can_run? arch
+        else
+          ppc_can_run? arch
+        end
```
xu-cheng: Maybe:
```
elsif Hardware::CPU.ppc?
  ppc_can_run? arch
else
  false
end
```

The idea is to avoid unnecessary (and maybe incorrect) assumption that it is either only intel or ppc?

Library/Homebrew/extend/os/mac/hardware/cpu.rb
mistydemeo: Done, and added a test for that case.

Library/Homebrew/extend/os/mac/hardware/cpu.rb
mistydemeo: Good idea, best not to hardcode it. Changed, and added a test for that case.

PR #2531, opened 2017-04-23
audit: Allow skipping/selective running of cops and cops refactor

GauthamGoli:

-----
As discussed in #2478 , this PR implements `--only` , `--except` kind of functionality for RuboCop Cops and also refactors existing cops into two "departments" viz., `FormulaAudit` and `FormulaAuditStrict` to separate analogous audit rules.
Please suggest better names if the current naming conventions are not acceptable. I could also probably write integration tests for the new options? 

Example commands:
`brew audit --only-cops=Desc,CompentsOrder nrpe`
`brew audit --except-cops=BottleBlock`

### DIFF
--- a/Library/Homebrew/cmd/style.rb
+++ b/Library/Homebrew/cmd/style.rb
@@ -15,6 +15,8 @@
 
 require "utils"
 require "json"
+require "rubocop"
+require_relative "../rubocops"
 
 module Homebrew
   module_function
@@ -53,6 +55,10 @@ module Homebrew
       --force-exclusion
     ]
     args << "--auto-correct" if fix
+
+    if options[:only]
+      args << "--only" << RuboCop::Cop::Cop.registry.with_department(options[:only]).names.join(" ")
+    end
 
     if files.nil?
       args << "--config" << HOMEBREW_LIBRARY_PATH/".rubocop.yml"
--- a/Library/Homebrew/cmd/style.rb
+++ b/Library/Homebrew/cmd/style.rb
@@ -1,4 +1,4 @@
-#:  * `style` [`--fix`] [`--display-cop-names`] [<files>|<taps>|<formulae>]:
+#:  * `style` [`--fix`] [`--display-cop-names`] [`--only-cops=`[COP1,COP2..]|`--except-cops=`[COP1,COP2..]] [<files>|<taps>|<formulae>]:
 #:    Check formulae or files for conformance to Homebrew style guidelines.
 #:
 #:    <formulae> and <files> may not be combined. If both are omitted, style will run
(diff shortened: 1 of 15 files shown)


docs/Manpage.md
```diff
@@ -635,6 +635,10 @@ With `--verbose` or `-v`, many commands print extra debugging information. Note
 
     If `--except` is passed, the methods named `audit_`method`` will not be run.
 
+    If `--only-cops` is passed, only the mentioned cops' violations would be checked.
```
MikeMcQuaid: `Rubocop cops'`

Library/Homebrew/cmd/style.rb
```diff
@@ -54,6 +56,20 @@ def check_style_impl(files, output_type, options = {})
     ]
     args << "--auto-correct" if fix
 
+    if options[:exclude].eql?(:FormulaAuditStrict) && !(options.key?(:except) || options.key?(:only))
```
MikeMcQuaid: Probably want to add these to the header documentation, too.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -69,15 +73,26 @@ def audit
       files = ARGV.resolved_formulae.map(&:path)
     end
 
+    only_cops = ARGV.value("only-cops").to_s.split(",")
```
MikeMcQuaid: What would you think about parsing values from `--only` instead? Feel free to decline, I'm not sure if I like it.

MikeMcQuaid:
> Conflicting files
> Library/Homebrew/rubocops/extend/formula_cop.rb

Now that https://github.com/Homebrew/brew/pull/2465 was merged this is conflicting. I'm sure you knew that but just FYI.

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: Initially I was trying to do that. The code would remain almost the same, but for starters I pushed this code for review. That would also conflict with the current convention that values of `--only` would be translated to `audit_<method_name>` . I mean, it may or may not look **clean**  :-|

MikeMcQuaid: Cool, well see what you think. @ilovezfs, any preferences welcome here.

Library/Homebrew/cmd/style.rb
GauthamGoli: `except` and `only` are just options in code. Should `brew style` accept them as options?

MikeMcQuaid: Yeh, may as well.

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: @MikeMcQuaid I updated the code. Simplified cops names. The command syntax now is
`$ brew audit --only-cops=Desc,ComponentsOrder nrpe`
`$ brew audit --except-cops=BottleBlock`
similarly for `brew style` 

I'm thinking unless ilovezfs has another preference we'll keep it this way and later on when most of the audit rules are translated to cops we might think of parsing values from `--only`|`--except`

MikeMcQuaid:
Thanks again @GauthamGoli!

GauthamGoli:
Thank you too for the review @MikeMcQuaid :D

PR #2537, opened 2017-04-24
Remove `to_s` from some `Pathname`s.

reitermarkus:

-----

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -24,7 +24,7 @@ def require?(path)
   require path
 rescue LoadError => e
   # we should raise on syntax errors but not if the file doesn't exist.
-  raise unless e.to_s.include? path
+  raise unless e.message.include?(path)
 end
 
 begin
--- a/Library/Homebrew/cask/lib/hbc/artifact/relocated.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/relocated.rb
@@ -35,10 +35,10 @@ module Hbc
         altnames = "(#{altnames})"
 
         # Some packges are shipped as u=rx (e.g. Bitcoin Core)
-        @command.run!("/bin/chmod", args: ["--", "u+rw", file.to_s, file.realpath.to_s])
+        @command.run!("/bin/chmod", args: ["--", "u+rw", file, file.realpath])
 
         @command.run!("/usr/bin/xattr",
-                      args:         ["-w", ALT_NAME_ATTRIBUTE, altnames, file.to_s],
+                      args:         ["-w", ALT_NAME_ATTRIBUTE, altnames, file],
                       print_stderr: false)
       end
 
--- a/Library/Homebrew/cask/lib/hbc/cli.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli.rb
@@ -113,7 +113,7 @@ module Hbc
       if command.respond_to?(:run)
         # usual case: built-in command verb
         command.run(*rest)
-      elsif require? which("brewcask-#{command}.rb").to_s
+      elsif require?(which("brewcask-#{command}.rb"))
         # external command as Ruby library on PATH, Homebrew-style
       elsif command.to_s.include?("/") && require?(command.to_s)
         # external command as Ruby library with literal path, useful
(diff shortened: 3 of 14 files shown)


MikeMcQuaid:
Have you tested this with many formulae?

Library/Homebrew/extend/ENV.rb
```diff
@@ -3,6 +3,12 @@
 require "extend/ENV/std"
 require "extend/ENV/super"
 
+class Array
+  def to_path_s
+    map(&:to_s).uniq.select { |s| File.directory?(s) }.join(File::PATH_SEPARATOR).chuzzle
```
alyssais: It wasn't clear to me what this method did until I looked at several usages of it and worked it out — the name a bit like it creates a single file path string using a list of directories, or something. Could we consider naming it `#to_PATH`, maybe? I think that would be clearer since this kind of "PATH" is conventionally uppercase.

I'm also not sure that this belongs in ENV.rb? It's not particularly related to the ENV hash. Maybe a new extend/Array.rb file.

Actually, to be honest, I'm not sure this method belongs on Array at all — converting to a PATH string isn't super related to a generic collection of elements. It feels like there's a separate responsibility here, which suggests to me it's time for a new object.

What would you think of something like this?

```ruby
class PATH # or title case, whatever. BinPath?
  attr_reader :entries

  def initialize(entries)
    @entries = entries.to_ary
  end

  def to_str
    @entries.join(File::PATH_SEPERATOR)
  end
end
```

(You'd want to expand to add stuff like checking `directory?` and calling `chuzzle`, etc., of course.)

Library/Homebrew/utils/popen.rb
```diff
@@ -12,6 +12,8 @@ def self.popen_write(*args, &block)
   end
 
   def self.popen(args, mode)
+    args = args.map { |a| a.respond_to?(:to_path) ? a.to_path : a }
```
alyssais: Are you sure about calling `#to_path` on _all_ of the arguments? It's only the first argument that should always be an object representing a path, IMO — generic shell arguments are just strings with no special meaning, so I wonder if those should just be converted straight to strings. Very minor point, I know, but I'm a bit of a conversion protocol enthusiast so this kind of thought appeals to me. :P

Library/Homebrew/utils/curl.rb
```diff
@@ -26,8 +26,13 @@ def curl_args(options = {})
     args << "--silent" if !$stdout.tty? || ENV["TRAVIS"]
   end
 
-  args += options[:extra_args] if options[:extra_args]
-  args
+  extra_args = [*options[:extra_args]].map do |arg|
+    next arg.to_s if arg.is_a?(URI)
```
alyssais: Slightly uneasy about the explicit type check here — shouldn't all arguments be converted to strings for a shell command?

Library/Homebrew/extend/ENV.rb
reitermarkus: > What would you think of something like this?

This seems like a good idea.

MikeMcQuaid:
@reitermarkus @alyssais I just want to call this out:

> Have you tested this with many formulae?

Unless we're very confident this will break zero existing formula the existing method will need to go into `compat/` and use `odeprecated`. Even in that case, I'm unsure what advantages we get from removing it at this point.

alyssais:
fwiw I also am not sure I see the advantages of removing the method, but some of the other cleanups in this PR are nice.

reitermarkus:
> Have you tested this with many formulae?

Yes, that's why I didn't actually remove the alias, because there are a bunch of Formulae using it.

MikeMcQuaid:
Ok, what's the general purpose of this PR? I'm not sure I understand then.

reitermarkus:
> I'm not sure I understand then.

That makes two of us. 😂 I started, trying to remove the dependencies on `Pathname#to_str`, but I'm beginning to think that it's not that bad, so we can probably just remove my first commit here.

The rest is refactoring of some things related to `PATH`s.

MikeMcQuaid:
@reitermarkus Cool. Might be worth opening a dedicated PR with just PATH refactoring and providing more context there. Thanks ❤️

MikeMcQuaid:
@reitermarkus Good to 🚢 when you are.

PR #2545, opened 2017-04-24
xcode: update Sierra CLT version to 802.0.42

JCount:
update Sierra CLT version to 802.0.42


JCount:
Hopefully this update will fix some of the issues that CLT only users have been running across.

MikeMcQuaid:
Thanks @JCount!

PR #2560, opened 2017-04-27
Refactor PATH generation.

reitermarkus:

-----

This add a `PATH` class which is used to create `PATH`-strings, replacing various other methods.


Library/Homebrew/cmd/sh.rb
```diff
@@ -23,7 +23,7 @@ def sh
     ENV.setup_build_environment
     if superenv?
       # superenv stopped adding brew's bin but generally users will want it
-      ENV["PATH"] = ENV["PATH"].split(File::PATH_SEPARATOR).insert(1, "#{HOMEBREW_PREFIX}/bin").join(File::PATH_SEPARATOR)
+      ENV["PATH"] = PATH.new(PATH.new(ENV["PATH"]).to_a.insert(1, HOMEBREW_PREFIX/"bin"))
```
alyssais: Could we make this a bit nicer somehow?

Library/Homebrew/PATH.rb
```diff
@@ -0,0 +1,61 @@
...
+  private
+
+  def parse(*paths)
+    paths
+      .flatten
+      .flat_map { |p| p.respond_to?(:to_str) ? p.to_str.split(File::PATH_SEPARATOR): p }
+      .compact
+      .map { |p| p.respond_to?(:to_path) ? p.to_path : p.to_str }
+      .uniq
+  end
```
alyssais: Rather than doing this complicated conversion stuff, we should use Ruby's built-in method for exactly this purpose — `Kernel#Pathname` from the `pathname` library.

Library/Homebrew/PATH.rb
```diff
@@ -0,0 +1,61 @@
...
+    @paths
+  end
+  alias to_a to_ary
+
+  def to_str
+    @paths.join(File::PATH_SEPARATOR)
+  end
+  alias to_s to_str
+
+  def eql?(other)
```
alyssais: If you implement `#eql?`, you need to implement `#hash` as well. Something like this would do:

```ruby
def hash
  [self.class, @paths].hash
end
```

Library/Homebrew/PATH.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  private
+
+  def parse(*paths)
+    paths
+      .flatten
+      .flat_map { |p| p.respond_to?(:to_str) ? p.to_str.split(File::PATH_SEPARATOR): p }
+      .compact
+      .map { |p| p.respond_to?(:to_path) ? p.to_path : p.to_str }
+      .uniq
```
MikeMcQuaid: I'm maybe misreading this but it doesn't look like this will avoid prepend/append adding duplicate entries (which would be good).

Library/Homebrew/cmd/sh.rb
MikeMcQuaid: Would be good to figure out a way to avoid multiple `PATH.new` here.

Library/Homebrew/diagnostic.rb
```diff
@@ -100,7 +100,7 @@ def check_for_installed_developer_tools
 
       # See https://github.com/Homebrew/legacy-homebrew/pull/9986
       def check_path_for_trailing_slashes
-        all_paths = ENV["PATH"].split(File::PATH_SEPARATOR)
+        all_paths = PATH.new(ENV["PATH"]).to_a
```
MikeMcQuaid: Not sure I understand why you'd want to create it and make it an array immediately. Would be nice to perhaps have a solution like @alyssais's or a new method that takes a string and returns an array.

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -196,7 +197,7 @@ def ncurses_define
 
   # @private
   def userpaths!
-    paths = self["PATH"].split(File::PATH_SEPARATOR)
+    paths = PATH.new(self["PATH"]).to_a
```
MikeMcQuaid: That this is here again smells of a needed utility method.

Library/Homebrew/extend/ENV/std.rb
```diff
@@ -57,12 +57,12 @@ def homebrew_extra_pkg_config_paths
...
-    paths << "#{HOMEBREW_PREFIX}/share/pkgconfig"
-    paths += homebrew_extra_pkg_config_paths
-    paths << "/usr/lib/pkgconfig"
-    paths.select { |d| File.directory? d }.join(File::PATH_SEPARATOR)
+    PATH.new(
+      HOMEBREW_PREFIX/"lib/pkgconfig",
+      HOMEBREW_PREFIX/"share/pkgconfig",
+      homebrew_extra_pkg_config_paths,
+      "/usr/lib/pkgconfig",
+    ).validate
```
MikeMcQuaid: A better name than `validate` would be good. `existing` or something?

Library/Homebrew/extend/ENV/shared.rb
alyssais: This is one of the cases that implementing `Enumerable` would get rid of.

alyssais: (So should be removable now)

Library/Homebrew/PATH.rb
reitermarkus: This doesn't take into account when strings are given that already are a PATH, which have to be split, or do you only mean the `map` line?

Library/Homebrew/extend/ENV/shared.rb
reitermarkus: @alyssais, I have to take a look at this method again, since in this case it is used twice.

But yes, `Array += PATH` should be an `Array` due to `to_ary`.

Library/Homebrew/PATH.rb
alyssais: Yeah, I think all we have to do here is, given an array of paths

1. Split any string-likes by path separator
2. Convert every array element to a pathname
3. Unique the paths

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -211,7 +212,7 @@ def userpaths!
         nil
       end
     end - %w[/usr/X11/bin /opt/X11/bin]
-    self["PATH"] = new_paths.uniq.join(File::PATH_SEPARATOR)
+    self["PATH"] = PATH.new(new_paths.uniq)
```
alyssais: Is the `#uniq` call here necessary?

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -196,22 +197,24 @@ def ncurses_define
...
+               .append(
+                 # user paths
+                 ORIGINAL_PATHS.map do |p|
+                   begin
+                     p.realpath.to_s
+                   rescue
+                     nil
+                   end
+                 end - %w[/usr/X11/bin /opt/X11/bin],
+               )
```
alyssais: This big long method chain is quite difficult for me to follow — is there a reason for not just writing `path.append`, etc. at the start of each line?

reitermarkus: Nope, fixed.

Library/Homebrew/PATH.rb
reitermarkus: > 2. Convert every array element to a pathname

These should be strings (for the uses of `select` with `start_with?`), so you would still have to call `to_path`/`to_s` on the pathname.

I don't know which one is better, `p.respond_to?(:to_path) ? p.to_path : p.to_str` or `Pathname.new(p).to_path`?

Library/Homebrew/PATH.rb
```diff
@@ -0,0 +1,73 @@
...
+    @paths = parse(*paths, *@paths)
+    self
+  end
+
+  def append(*paths)
+    @paths = parse(*@paths, *paths)
+    self
+  end
+
+  def insert(idx, *paths)
```
MikeMcQuaid: `idx` -> `index`

Library/Homebrew/PATH.rb
```diff
@@ -0,0 +1,73 @@
...
+    false
+  end
+
+  def empty?
+    @paths.empty?
+  end
+
+  def existing
+    existing_path = select(&File.method(:directory?))
+    existing_path unless existing_path.empty?
```
MikeMcQuaid: I'm tempted to say returning an empty array is nicer here because it avoids a bunch of `nil` checks.

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -80,7 +81,8 @@ def prepend(keys, value, separator = " ")
   end
 
   def append_path(key, path)
-    append key, path, File::PATH_SEPARATOR if File.directory? path
+    puts "'#{path}'"
```
MikeMcQuaid: This a debugging thing that was left in?

Library/Homebrew/PATH.rb
reitermarkus: The usage for this is more like this:

```ruby
ENV["MY_PATH"] = mypath.existing
```

So, by returning `nil`, this will unset `MY_PATH`, instead of setting it to an empty string.

Library/Homebrew/extend/ENV/shared.rb
reitermarkus: Ah, yes!

Library/Homebrew/PATH.rb
reitermarkus: Done.

Library/Homebrew/PATH.rb
MikeMcQuaid: Ah, I see. Ok, that makes sense. Maybe worth a comment to avoid that being changed in future.

Library/Homebrew/PATH.rb
```diff
@@ -0,0 +1,74 @@
...
+    # return nil instead of empty PATH, to unset environment variables
+    existing_path unless existing_path.empty?
+  end
+
+  private
+
+  def parse(*paths)
+    paths.flatten
+         .compact
+         .flat_map { |p| Pathname.new(p).to_path.split(File::PATH_SEPARATOR) }
```
alyssais: Minor nit: I think `Pathname(p)` is more semantically appropriate, even though they both (currently) do the same thing, since `Pathname()` is a conversion function rather than a constructor.

reitermarkus: > `Pathname()` is a conversion function rather than a constructor.

Huh, I never thought about it this way, but I'll leave it as is for now.

alyssais: I highly recommend [Confident Ruby](http://www.confidentruby.com/) if you want to go super in-depth into this stuff.

PR #2562, opened 2017-04-28
audit: formally ban go get usage

DomT4:

-----

This has finally bothered me one too many times.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -885,6 +885,10 @@ class FormulaAuditor
       problem "Formulae using virtualenvs do not need a `setuptools` resource."
     end
 
+    if text =~ /system\s+['"]go['"],\s+['"]get['"]/
+      problem "Formulae should not use `go get`. If non-vendored resources are required use `go_resource`s."
+    end
+
     return unless text.include?('require "language/go"') && !text.include?("go_resource")
     problem "require \"language/go\" is unnecessary unless using `go_resource`s"
   end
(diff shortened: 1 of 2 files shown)


ilovezfs:
It's a pity that make files using go get will escape this check.

DomT4:
It may be fairly expensive in terms of speed, but we could probably run a `File.read` on `Makefile` if one exists and the project is using `go` as a dependency to check that, if there's the desire to do so.

bfontaine:
@DomT4 It would be overly complicated, I think. What if the `Makefile` has a target that runs `go get` but the formula doesn’t use it?

ilovezfs:
I think if anything it would need to be a superenv shim for `go` that blows up if called with `get`, but I was mostly musing.

DomT4:
Yeah, I was mostly throwing an idea out there 😺. I think this `audit` tweak will catch the most likely problem which is people newish to Homebrew trying to fix `go` problems in formulae. I'm happy with this as-is at least for now, unless y'all desire changes.

>  What if the Makefile has a target that runs go get but the formula doesn’t use it?

This is actually pretty common. In some cases IIRC we even `inreplace` out those lines from the darn `Makefile` so we can run `make` without it fetching things we've already fetched 😓.

As always the theme comes back to upstreams needing to move to the `vendor` system, which is a beautiful thing that almost always _"just works"_ ™️, but adoption on that hasn't been as strong as I hoped it would be.

bfontaine:
That’s a great branch name, by the way 😄

MikeMcQuaid:
Great work here @DomT4!

DomT4:
Thanks folks ❤️

PR #2565, opened 2017-04-29
Increment system command timeout to avoid problems installing some casks

danielmartin:
Depending on the computer speed and the cask size, the former timeout
caused some problems installing some programs like Google Chrome,
Spotify, or LibreOffice.

-----

This fixes https://github.com/caskroom/homebrew-cask/issues/33116 and https://github.com/caskroom/homebrew-cask/issues/32830

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/system_command.rb
+++ b/Library/Homebrew/cask/lib/hbc/system_command.rb
@@ -92,7 +92,7 @@ module Hbc
 
     def each_line_from(sources)
       loop do
-        selected_sources = IO.select(sources, [], [], 10)
+        selected_sources = IO.select(sources, [], [], 50)
 
         break if selected_sources.nil?
 


mislav:
I can confirm that this makes `brew cask install spotify` work again, whereas before the `hdiutil mount` command would consistently fail without anything printed on stderr nor stdout, and the `*.dmg` file would somehow still remain "in-use" by the diskimage process:

```
$ lsof ~/Library/Caches/Homebrew/Cask/spotify--latest.dmg
COMMAND     PID   USER   FD   TYPE DEVICE SIZE/OFF    NODE NAME
diskimage 49861 mislav    3r   REG    1,4 97553538 4346868 /Users/mislav/Library/Caches/Homebrew/Cask/spotify--latest.dmg
```

reitermarkus:
Might be better to fully revert https://github.com/Homebrew/brew/pull/2517 since this seems to cause more problems than it solves, and could still fail with bigger programs, even with 50 seconds timeout.

@vitorgalvao?

vitorgalvao:
@reitermarkus Agreed on reverting.

danielmartin:
OK, revert PR is here: https://github.com/Homebrew/brew/pull/2571

PR #2568, opened 2017-04-30
Fix url extension issue when it contains query strings

sepehr:

-----

I usually contribute to caskroom/homebrew-cask repo by adding/updating casks. I was in middle of writing a new cask for an application that is hosting its binary on Dropbox! The direct download URL on the website is something like this:  

```
https://dropbox.com/s/someHash/eqMac.dmg?dl=1
```

When brew cask downloads the dmg file, it gets saved as `appName.dmg?dl=1` and so it throws an error and exists without installing:  

```
==> Installing Cask bitgapp-eqmac
Error: Uh oh, could not figure out how to unpack '/Users/Sepehr/Library/Caches/Homebrew/Cask/bitgapp-eqmac--latest.dmg?dl=1'
Error: nothing to install
```

I went through the codebase and I think that I found the issue. But I'm not sure if I got this right since I'm neither a ruby developer nor familiar with the codebase. I'd be grateful if the maintainers review this PR and give me some feedbacks :)

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/download_strategy.rb
+++ b/Library/Homebrew/cask/lib/hbc/download_strategy.rb
@@ -181,7 +181,7 @@ module Hbc
     end
 
     def ext
-      Pathname.new(@url).extname.sub(/\?.*/, '')
+      Pathname.new(@uri_object.path).extname
     end
   end
 


Library/Homebrew/cask/lib/hbc/download_strategy.rb
```diff
@@ -181,7 +181,7 @@ def referer_args
     end
 
     def ext
-      Pathname.new(@url).extname
+      Pathname.new(@url).extname.sub(/\?.*/, '')
```
reitermarkus: Instead of this, let's use the `@uri_object` from above, to get the path without the query string.

```ruby
@uri_object.path
```

Then, get the extension from this.

```ruby
Pathname.new(@uri_object.path).extname
```

sepehr: Thank you @reitermarkus for the review. I've updated the PR. Please have a look.

reitermarkus:
I just realized that this will break URLs that actually use the query string to specify the filename.

`https://example.org/download?file=application.zip`

sepehr:
Good catch @reitermarkus, thanks. I'll update the pull request asap.

reitermarkus:
Using a Regex instead of `extname` here will probably work.

sepehr:
FYI, it's on my todo list and I'll get to it as soon as I can find some free time.

PR #2571, opened 2017-05-01
Revert "Fix `IO#select` blocking."

danielmartin:
Reverts Homebrew/brew#2517

This may cause some timeout problems on slow machines or big casks. See caskroom/homebrew-cask#33116 and caskroom/homebrew-cask#32830


reitermarkus:
Thanks!

PR #2598, opened 2017-05-06
Relocate virtualenv orig-prefix

tdsmith:
Virtualenvs remember the path to the stdlib in a file named
orig_prefix.txt. This file wasn't being relocated because Homebrew skips
files that look like they're intended for human consumption by matching
against the list of metafile extensions. This led to the bug described
in https://github.com/Homebrew/homebrew-core/issues/12869. This fixes
relocation by creating an exception for orig-prefix.txt.

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -133,6 +133,7 @@ class Keg
       files = Set.new path.find.reject { |pn|
         next true if pn.symlink?
         next true if pn.directory?
+        next false if pn.basename.to_s == "orig-prefix.txt" # for python virtualenvs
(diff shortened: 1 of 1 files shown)


ilovezfs:
Any reason not to drop the file extension and even possibly make it a dot file?

tdsmith:
I wouldn't mind but orig-prefix.txt is written by the python virtualenv package, and so is the code that reads it, so the change would have to be made there.

ilovezfs:
Oh this is an upstream virtualenv file name?

tdsmith:
Yes.

ilovezfs:
Then this seems fine.

ilovezfs:
I tested this PR, and it works as expected after rebottling `ansible`.

PR #2601, opened 2017-05-07
Add audit check to see if both version and checksum changed.

reitermarkus:

---

Addresses https://github.com/caskroom/homebrew-cask/issues/33536.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -752,7 +752,7 @@ class FormulaAuditor
       next unless spec = formula.send(spec_sym)
       next unless previous_version_and_checksum[spec_sym][:version] == spec.version
       next if previous_version_and_checksum[spec_sym][:checksum] == spec.checksum
-      problem "#{spec_sym}: only sha256 changed; needs to be confirmed by the developer"
+      problem "#{spec_sym}: sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed."
     end
 
     attributes = [:revision, :version_scheme]
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -31,6 +31,9 @@
 #:
 #:    If `--except-cops` is passed, the given Rubocop cop(s)' checks would be skipped.
 #:
+#:    If `--commit-range` is is passed, the audited Formula will be compared to the
+#:    last revision before the `<commit_range>`.
+#:
 #:    `audit` exits with a non-zero status if any errors are found. This is useful,
 #:    for instance, for implementing pre-commit hooks.
 
@@ -648,8 +651,24 @@ class FormulaAuditor
       problem "Devel-only (no stable download)"
     end
 
+    previous_formula_contents = unless formula.tap.nil?
+      commit_range = ARGV.value("commit-range")
+      Git.last_revision_of_file(formula.tap.path, formula.path, before_commit: commit_range)
+    end
+    previous_formula = unless (previous_formula_contents || "").empty?
+      Formulary.from_contents(formula.name, formula.path, previous_formula_contents)
+    end
+
     %w[Stable Devel HEAD].each do |name|
       next unless spec = formula.send(name.downcase)
+
+      unless previous_formula.nil?
+        previous_spec = previous_formula.send(name.downcase)
+
+        if previous_spec.version == spec.version && previous_spec.checksum != spec.checksum
+          problem "#{name}: only sha256 changed; needs to be confirmed by the developer"
+        end
+      end
 
       ra = ResourceAuditor.new(spec, online: @online, strict: @strict).audit
       problems.concat ra.problems.map { |problem| "#{name}: #{problem}" }
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -31,9 +31,6 @@
 #:
 #:    If `--except-cops` is passed, the given Rubocop cop(s)' checks would be skipped.
 #:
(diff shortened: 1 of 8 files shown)


ilovezfs:
Should this check apply to formulae, too?

reitermarkus:
@ilovezfs, added the same check for Formulae, but it's probably at the wrong location at the moment. Also, not sure if this should check both `devel` and `stable` or just `stable`.

ilovezfs:
Both.

vitorgalvao:
But if `audit` fails unless both `version` and `sha256` change, then how do we deal with the legitimate cases where `sha256` changes, but not the `version`?

reitermarkus:
@vitorgalvao, we still merge it, after it is verfied. This is basically just a fail-safe so we don't have/forget to check it.

vitorgalvao:
@reitermarkus My main gripe was with the previous message, then, since it would be confusing to users. The new one (`only sha256 changed; needs to be confirmed by the developer`) is better, but I still feel it’ll be confusing to users running `audit` locally.

Maybe something that reassures them they should still submit the PR anyway: `only sha256 changed, so it will need to be confirmed by the developer (https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md#updating-the-sha256)`. I feel the URL to the full instructions is important, so users know what we’re talking about. Either that or we make this check run only on Travis (like the one that checks for cask/formula name collisions).

reitermarkus:
> or we make this check run only on Travis

I'd opt for that.

MikeMcQuaid:
> I'd opt for that.

I'll review the code shortly (at a conference, sorry) but I'd suggest that you should always get the same results locally as you do on CI to avoid confusion.

vitorgalvao:
> but I'd suggest that you should always get the same results locally as you do on CI to avoid confusion.

In this particular case I’d argue the confusion will occur if this message is shown locally. Like the aforementioned case where we check for name collisions between HB and HBC (so we know to verify if we’re not reintroducing duplicates), this is an audit that _can benefit from user help but ultimately requires maintainer decision._

We already teach users about this verification in [our `PULL_REQUEST_TEMPLATE`](https://github.com/caskroom/homebrew-cask/blob/b1269b5025fcaaf176ae16d6f4ce895814dd99a1/.github/PULL_REQUEST_TEMPLATE.md) (see “Additionally, if updating a cask:”). I fear that by showing this locally we’ll discourage submissions from users that don’t understand the message (and will just abandon the fix, instead of asking).

In sum, my view is:

+ Local audits should be the limited to the things you can fix locally.
+ Travis-only audits are the ones that have mandatory maintainer checks.
+ Travis should also run all local audits.

MikeMcQuaid:
> In this particular case I’d argue the confusion will occur if this message is shown locally. Like the aforementioned case where we check for name collisions between HB and HBC (so we know to verify if we’re not reintroducing duplicates), this is an audit that can benefit from user help but ultimately requires maintainer decision.
> 
> We already teach users about this verification in our PULL_REQUEST_TEMPLATE (see “Additionally, if updating a cask:”). I fear that by showing this locally we’ll discourage submissions from users that don’t understand the message (and will just abandon the fix, instead of asking).

I think it's right/good that people who don't understand this case don't file these PRs as we should ask them to submit upstream. What `brew cask audit` does it up to you, though, but with `brew audit --new-formula` we've specifically followed this pattern previous of matching local with CI so will want to continue doing so here.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -795,6 +795,15 @@ def audit_revision_and_version_scheme
     return if @new_formula
 
     fv = FormulaVersions.new(formula)
+
+    previous_version_and_checksum = fv.previous_version_and_checksum("origin/master")
```
MikeMcQuaid: You should be able to pass `:sha256` as an extra argument to `version_attributes_map` instead.

reitermarkus: This doesn't work, since `sha256` always forwards to `stable.sha256`.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -742,6 +742,15 @@ def audit_revision_and_version_scheme
...
+    [:stable, :devel].each do |spec_sym|
+      next unless spec = formula.send(spec_sym)
+      next unless previous_version_and_checksum[spec_sym][:version] == spec.version
+      next if previous_version_and_checksum[spec_sym][:checksum] == spec.checksum
+      problem "#{spec_sym}: only sha256 changed; needs to be confirmed by the developer"
```
MikeMcQuaid: @ilovezfs thoughts on wording here?

ilovezfs: Maybe

"sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed."

vitorgalvao: I still say we should link to the relevant documentation. Whatever the message, there will always be people that do not understand the wording. I argue it’s best to keep the message short and explain the why the error, not the reasoning behind it. That should be left to the longer documentation, where people are linked to. That way we’re not always hammering people with the long explanation and don’t need to keep updating the message to better clarify it (because it can be done at length in the documentation).

MikeMcQuaid: The long explanation here will be really rare. There currently is no out-of-band audit documentation and I don't particular want to create it. Our goal with our audits should be that people know what they need to do to get past the issue ideally before creating the PR.

vitorgalvao: Oh, right, I was thinking this was HBC-only (we have the documentation for this case). Never mind, then.

zmwangx:
`brew audit` seems to be ridiculously slow for me after this. Please take a look at #2665 if you wish.

PR #2610, opened 2017-05-08
Add autocorrect method for ComponentsOrder rubocop and tests

GauthamGoli:

-----
Can fix components' order using to `brew audit --fix`, 

But please note that `autocorrect` is called just after `problem` is called, ~~and `autocorrect` works by swapping those two components in the source~~

Edit: `autocorrect` doesn't swap anymore, rather reorders the components.  

 During the re order (only for the components out of precedence order ) 
 -  linebreaks, and grouping of components are taken into account. ( A Multi line 
 component is surrounded by line breaks, while components of same type are placed together) 
 - First 8 components i.e., upto `version_scheme` are grouped together.

### DIFF
--- a/Library/Homebrew/rubocops/components_order_cop.rb
+++ b/Library/Homebrew/rubocops/components_order_cop.rb
@@ -36,7 +36,7 @@ module RuboCop
             [{ name: :test, type: :block_call }],
           ]
 
-          present_components = component_precedence_list.map do |components|
+          @present_components = component_precedence_list.map do |components|
             relevant_components = []
             components.each do |component|
               case component[:type]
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
> But please note that autocorrect is called just after problem is called

Might be worth noting that in the code as it makes it more obvious to understand why instance variables are used? Otherwise: looks great.

GauthamGoli:
Please add do not merge label. Working on the issues pointed out by ilovezfs https://github.com/Homebrew/homebrew-core/pull/13375#discussion_r115444463

GauthamGoli:
The PR description is updated to reflect the new changes. `do not merge` may be removed now.

MikeMcQuaid:
Thanks again @GauthamGoli !

PR #2624, opened 2017-05-11
Use Python's flock instead of Ruby's

mistydemeo:

-----

I'm currently in the process of adding vendored Ruby support to Tigerbrew (https://github.com/mistydemeo/tigerbrew/pull/510). One of the roadblocks I ran into was that the flock shell script doesn't work. Ruby first gained flock in 1.8.7, which is a problem since we're using this lock utility in `vendor-install` in order to install a newer Ruby. Catch-22!

Fortunately, Python 2.3(!) has flock support. Should work anywhere from ancient Pythons to modern OS X.

### DIFF
--- a/Library/Homebrew/utils/lock.sh
+++ b/Library/Homebrew/utils/lock.sh
@@ -42,13 +42,13 @@ _create_lock() {
   [[ -x "$ruby" ]] || ruby="$(which ruby 2>/dev/null)"
   [[ -x "$python" ]] || python="$(which python 2>/dev/null)"
 
-  if [[ -n "$ruby" && $("$ruby" -e "puts RUBY_VERSION >= '1.8.7' ? 0 : 1") = 0 ]]
+  if [[ -x "$ruby" && $("$ruby" -e "puts RUBY_VERSION >= '1.8.7' ? 0 : 1") = 0 ]]
   then
     "$ruby" -e "File.new($lock_fd).flock(File::LOCK_EX | File::LOCK_NB) || exit(1)"
-  elif [[ -n "$(which flock)" ]]
+  elif [[ -x "$(which flock)" ]]
   then
     flock -n "$lock_fd"
-  elif [[ -n "$python" ]]
+  elif [[ -x "$python" ]]
   then
     "$python" -c "import fcntl; fcntl.flock($lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)"
   else
--- a/Library/Homebrew/utils/lock.sh
+++ b/Library/Homebrew/utils/lock.sh
@@ -38,14 +38,19 @@ _create_lock() {
   local lock_fd="$1"
   local name="$2"
   local ruby="/usr/bin/ruby"
+  local python="/usr/bin/python"
   [[ -x "$ruby" ]] || ruby="$(which ruby 2>/dev/null)"
+  [[ -x "$python" ]] || python="$(which python 2>/dev/null)"
 
-  if [[ -n "$ruby" ]]
+  if [[ -n "$ruby" && $("$ruby" -e "puts RUBY_VERSION >= '1.8.7' ? 0 : 1") = 0 ]]
   then
     "$ruby" -e "File.new($lock_fd).flock(File::LOCK_EX | File::LOCK_NB) || exit(1)"
   elif [[ -n "$(which flock)" ]]
   then
     flock -n "$lock_fd"
+  elif [[ -n "$python" ]]
+  then
+    "$python" -c "import fcntl; fcntl.flock($lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)"
   else
     onoe "Cannot create $name lock, please avoid running Homebrew in parallel."
   fi
--- a/Library/Homebrew/utils/lock.sh
+++ b/Library/Homebrew/utils/lock.sh
@@ -42,7 +42,7 @@ _create_lock() {
   [[ -x "$ruby" ]] || ruby="$(which ruby 2>/dev/null)"
   [[ -x "$python" ]] || python="$(which python 2>/dev/null)"
 
-  if [[ -x "$ruby" && $("$ruby" -e "puts RUBY_VERSION >= '1.8.7' ? 0 : 1") = 0 ]]
+  if [[ -x "$ruby" ]] && "$ruby" -e "exit(RUBY_VERSION >= '1.8.7')"
   then
     "$ruby" -e "File.new($lock_fd).flock(File::LOCK_EX | File::LOCK_NB) || exit(1)"
   elif [[ -x "$(which flock)" ]]


sjackman:
This PR would make Python a required dependency of `brew`, which creates trouble for `brew` on Linux, where `python` is not guaranteed to be available. Could this PR use `ruby` if available and new enough, and `python` otherwise?

mistydemeo:
Does Linux have the `flock` utility installed? This falls back to using the CLI `flock` tool if `python`'s not available.

mistydemeo:
I also tried a perl version but for the life of me couldn't get it to work. 😰

sjackman:
Yes, Linux typically has `flock`. It's provided by the package [util-linux](https://github.com/karelzak/util-linux/blob/master/sys-utils/flock.c).

MikeMcQuaid:
> This PR would make Python a required dependency of brew, which creates trouble for brew on Linux, where python is not guaranteed to be available. Could this PR use ruby if available and new enough, and python otherwise?

Yeh, I'm afraid I'd really rather avoid adding a Python dependency to Homebrew/brew. Are there any other options?

mistydemeo:
> Yeh, I'm afraid I'd really rather avoid adding a Python dependency to Homebrew/brew. Are there any other options?

As mentioned above, the script already uses the `flock` CLI utility as a fallback when installed, which as @sjackman mentioned is typically installed on Linux systems.

sjackman:
macOS has the syscall `flock(2)`. You could compile [flock.c](https://github.com/karelzak/util-linux/blob/master/sys-utils/flock.c) for macOS. Perhaps not a great option, but I thought I'd mention it as an option. You could create a formula for this stolen version of `flock(1)`.

mistydemeo:
I don't think compiling a utility's really feasible; this is implemented in shell (and using system utilities) so it has minimal deps, since it's called automatically when `brew` is installed if the system Ruby isn't suitable.

sjackman:
How about compiling a binary executable of this `flock(1)` that will run any version of macOS (assuming that's possible), and shipping the precompiled executable with `brew` or fetching it with `curl`?

woodruffw:
Can't we use [`Fcntl`](http://ruby-doc.org/stdlib-2.4.0/libdoc/fcntl/rdoc/Fcntl.html) from the Ruby stdlib for this? It's been available since at least 1.8.6.

It wraps `fcntl(2)`, and should provide `F_SETLK`, `F_SETLKW`, and `F_GETLK` for advisory locks a la `flock(2)`. It's also POSIX, unlike `flock(2)`.

mistydemeo:
Looks like that's available in 1.8.2; I'll look at using it.

mistydemeo:
It turns out I'm very bad at fcntl 😅 Something like this:

```ruby
File.new(200, "w").fcntl(Fcntl::F_SETLK, Fcntl::F_WRLCK)
```

Fails with a `Errno::EFAULT: Bad address`.

yaakov-h:
Saw this mentioned on Twitter and got curious, sorry if this seems strange and out-of-the-blue.

Something like this?

```
ruby -e "require 'fcntl'; File.new(200).fcntl(Fcntl::F_SETLKW, [ 0, 0, Process.pid, Fcntl::F_WRLCK, IO::SEEK_SET ].pack('qqiss'));
```

The struct packed there is:

```
struct flock {
    off_t       l_start;    /* starting offset */
    off_t       l_len;      /* len = 0 means until end of file */
    pid_t       l_pid;      /* lock owner */
    short       l_type;     /* lock type: read/write, etc. */
    short       l_whence;   /* type of l_start */
};
```

mistydemeo:
🙇‍♀️ Thank you, that's great! I'll look at switching to that version.

I also got the Perl version going, on the basis that Unix being 50% Perl by volume, it would always be available:

```perl
"$perl" -e "use Fcntl qw(:flock); open(my \$fh, '>>&=', $lock_fd); flock(\$fh, LOCK_EX | LOCK_NB) or exit(1)"
```

MikeMcQuaid:
> As mentioned above, the script already uses the flock CLI utility as a fallback when installed, which as @sjackman mentioned is typically installed on Linux systems.

Sorry, to clarify, I mean I'd like to avoid adding it on macOS systems. Could we use the existing code and fall back to `/usr/bin/python` only if it's not available in `/usr/bin/ruby`?

MikeMcQuaid:
> I also got the Perl version going, on the basis that Unix being 50% Perl by volume, it would always be available:

I'm cool with Perl or Python if we can use Ruby first. On macOS it's just reducing the chances that someone monkeying with their system will :boom: at this stage down to a single interpreter, ideally.

mistydemeo:
Okay, after some futzing around with @yaakov-h's solution, I found that `Fcntl::F_SETLKW` wasn't working reliably, and `fcntl(2)`'s flock has some subtle differences from `flock(2)`, which makes me nervous about using something different.

I've updated this PR to use the Python version, while also keeping the Ruby version. The order of preference is Ruby (if 1.8.7 or greater is installed) > `flock(1)` > Python. Please excuse my horrid shell and let me know if there's a better way to handle that conditional.

Library/Homebrew/utils/lock.sh
```diff
@@ -38,14 +38,19 @@ _create_lock() {
   local lock_fd="$1"
   local name="$2"
   local ruby="/usr/bin/ruby"
+  local python="/usr/bin/python"
   [[ -x "$ruby" ]] || ruby="$(which ruby 2>/dev/null)"
+  [[ -x "$python" ]] || python="$(which python 2>/dev/null)"
 
-  if [[ -n "$ruby" ]]
+  if [[ -x "$ruby" && $("$ruby" -e "puts RUBY_VERSION >= '1.8.7' ? 0 : 1") = 0 ]]
```
MikeMcQuaid: I'd be tempted to just make this something like:
```bash
if [[ -x "$ruby" && "$ruby" -e "exit(RUBY_VERSION >= '1.8.7')" ]]
```

That'll just use the exit code correctly.

Library/Homebrew/utils/lock.sh
```diff
@@ -38,14 +38,19 @@ _create_lock() {
...
   then
     "$ruby" -e "File.new($lock_fd).flock(File::LOCK_EX | File::LOCK_NB) || exit(1)"
-  elif [[ -n "$(which flock)" ]]
+  elif [[ -x "$(which flock)" ]]
   then
     flock -n "$lock_fd"
+  elif [[ -x "$python" ]]
+  then
+    "$python" -c "import fcntl; fcntl.flock($lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)"
```
MikeMcQuaid: Presumably this is available in very old Pythons?

mistydemeo: Yeah, it's available since at least 2.3 (which was shipped on Tiger), possibly older.

Library/Homebrew/utils/lock.sh
mistydemeo: Pushed a commit to simplify the conditional. Realized what I was doing wrong was testing the exit status within `[[` - bash doesn't allow testing exit status like that there. But testing `if [[ -x "$ruby" ]] && $expression` works.

PR #2626, opened 2017-05-13
Added MonoRequirement for using installations from official Mono packages.

alexreg:


MikeMcQuaid:
@alexreg Can you detail which of our current Mono packages in homebrew/core you've tested this with? Thanks!

alexreg:
@MikeMcQuaid Sure. I tested on the *cake* formula already. Would you like me to test on some others?

MikeMcQuaid:
Yes, please. As many as possible. Note that this will result in these formulae being essentially `depends_on "mono"` when poured from a bottle.

alexreg:
Sure, will do. And when you say "test", I suppose you mean that I should verify they install correctly, but also that `brew test` works on them?

I could also submit a PR for *hombrew-core* with their updated formulas, if you like.

MikeMcQuaid:
Verify they install correctly, `brew test` works with them, that when built from source with the Homebrew Mono they still work at runtime with the non-Homebrew Mono and that they seem to not have any obvious issues.

MikeMcQuaid:
@alexreg Additionally can you explain why you need this and why this is useful to you?

alexreg:
It's useful because I like to use the official installation of Mono, included with Xamarin (well, Visual Studio for Mac these days). I think many others will too.

Okay, tested all the formulae that depend on mono, and they all seem to install and test correctly. :-) Well, one slight exception: the *gitversion* test fails, but it was already doing that for me before these changes.

```
fatal: unable to auto-detect email address (got '_homebrew@Alex-MacBookPro.(none)')
```

Is that all good with you?

MikeMcQuaid:
> It's useful because I like to use the official installation of Mono, included with Xamarin (well, Visual Studio for Mac these days). I think many others will too.

That explains what you want to do but not why. Can you elaborate on why specifically you want to use it at runtime for Homebrew applications?

alexreg:
It's what and a why. The only alternative is to install a Homebrew version of Mono too now, which is far from ideal.

MikeMcQuaid:
Sorry, there's still no "why" here except "because @alexreg wants to". So it's saving disk space/saving download time or something else?

alexreg:
Yes, that's right. I thought that was implicit, sorry. It's saving effective duplication of a dependency (hence downloading and installation time, not to mention space).

alexreg:
Also, having two versions of Mono on one's system can mess things up in terms of resolving paths...

MikeMcQuaid:
> Also, having two versions of Mono on one's system can mess things up in terms of resolving paths...

Can you elaborate on this?

alexreg:
> Also, having two versions of Mono on one's system can mess things up in terms of resolving paths...
> Can you elaborate on this?

The resolution for DLLs (libraries), with regards to the .NET Framework (core libraries) can get muddled, as far as I know.  It can be done, but care needs to be taken, as hinted by this page: http://www.mono-project.com/docs/compiling-mono/parallel-mono-environments/

Regardless, I think that not having two copies installed is a good enough reason, especially given the size of a Mono installation. :)

DomT4:
Doesn't Homebrew basically already provide this in `core` at least by the regular practice of `mono` dependencies being semi-consistently `=> :recommended`. Seems like that could just be made wholly consistent rather than adding _yet another_ special dependency.

DomT4:
Unless there's significant use of things detecting the path to `mono` at buildtime and hardcoding that as a hashbang, but the way I've seen people talk about the `=> :recommended` deps suggests that's uncommon?

MikeMcQuaid:
> The resolution for DLLs (libraries), with regards to the .NET Framework (core libraries) can get muddled, as far as I know. It can be done, but care needs to be taken, as hinted by this page: http://www.mono-project.com/docs/compiling-mono/parallel-mono-environments/

This doesn't seem to be a problem for you personally, though, so I'm going to say this is speculative.

> Regardless, I think that not having two copies installed is a good enough reason, especially given the size of a Mono installation. :)

I'm not sure, I don't think that's a good reason or precedent, personally. Will leave this open for a bit for others to think about this.

alexreg:
Okay, @MikeMcQuaid.

Another (small) benefit I thought of, is verifying the version of Mono installed is sufficient, when installing Homebrew formulae depending on Mono. Also, I don't like the `:optional`/`:recommended` approach, since that's not what it's meant for. Very hacky.

MikeMcQuaid:
Passing on this, sorry.

alexreg:
Any particular reason?

MikeMcQuaid:
No-one else has chimed in in support of this, it wouldn't see widespread usage and there's no real disadvantage beyond disk-space/one-time bandwidth usage.

alexreg:
I think just no-one else has seen it, to be honest! I doubt many people trawl through issues just to give thumbs up/down... but okay. I see disk space as a big point in its favour, as well as other minor ones.

PR #2628, opened 2017-05-13
audit: Detect multiline and interpolated strings in formula desc cop

### LINKED ISSUE #2611: `brew audit` on description length isn't strict enough
- [x] Confirmed this is a problem with Homebrew/brew and not one or more formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

-----

`brew audit` on description length is not restrictive enough. See https://github.com/Homebrew/homebrew-science/pull/5608

```
 $ brew desc https://raw.githubusercontent.com/MmgTools/homebrew-science/30777e8/mmg.rb | wc -c
     113
```

yet `brew audit --strict --online` for the same formula doesn't raise an error. The issue is that the string is constructed weirdly:

```
  desc "Simplicial remeshers: mesh improvement,"\
  "isotropic or anisotropic mesh adaptation and isovalue"\
  "discretization."
```

I don't know how to debug rubocops so if someone could point me in that direction that would be great.

GauthamGoli:

-----
Fixes #2611 
*Note:* It assumes there is no expression/variables in the multiline string. Ex: `'#{var1} is a formula'` Although var1 can be substituted it doesn't make sense, as rubocop is a static analyser, so I'm ignoring that part of string .
Also does not support here document , let me know if it's necessary for formula `desc`

### DIFF
--- a/Library/Homebrew/rubocops/extend/formula_cop.rb
+++ b/Library/Homebrew/rubocops/extend/formula_cop.rb
@@ -180,9 +180,10 @@ module RuboCop
         node.source_range.source_buffer
       end
 
-      # Returns the string representation if node is of type str
+      # Returns the string representation if node is of type str(plain) or dstr(interpolated)
       def string_content(node)
-        node.str_content if node.type == :str
+        return node.str_content if node.type == :str
+        node.each_child_node(:str).map(&:str_content).join("") if node.type == :dstr
       end
 
       # Returns printable component name
--- a/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
+++ b/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
@@ -51,6 +51,31 @@ describe RuboCop::Cop::FormulaAuditStrict::Desc do
       end
     end
 
+    it "When desc is multiline string" do
+      source = <<-EOS.undent
+        class Foo < Formula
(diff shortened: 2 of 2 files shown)


Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -180,9 +180,10 @@ def source_buffer(node)
         node.source_range.source_buffer
       end
 
-      # Returns the string representation if node is of type str
+      # Returns the string representation if node is of type str(plain) or dstr(interpolated)
       def string_content(node)
-        node.str_content if node.type == :str
+        return node.str_content if node.type == :str
+        node.each_child_node(:str).map(&:str_content).join("") if node.type == :dstr
```
MikeMcQuaid: Are there any other types of strings beyond these two?

GauthamGoli: There is execute-string https://github.com/whitequark/parser/blob/master/doc/AST_FORMAT.md#execute-string
I'm not sure if we should consider that here. Can you please take a look and confirm that?

MikeMcQuaid: Yeh, I think we can ignore that, thanks!

MikeMcQuaid:
Thanks again @GauthamGoli!

Issue #2651, opened 2017-05-18
Error running `brew cask` with no arguments


ewang12:
# Please always follow these steps:
- [x] Confirmed this is a problem with running a `brew` command and just `brew install`ing one or more formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed all issues and retried your prior step?
- [x] Ran `brew config` and `brew doctor` and included their output with your issue?

**Please note we may immediately close your issue without comment if you delete or do not fill out the issue checklist and provide ALL the requested information.**

To help us debug your issue please explain:
- What you were trying to do (and why)
I ran `brew cask` with no arguments.
- What happened (include command output)
The error I got was:
```
Error: no implicit conversion of nil into String
Follow the instructions here:
  https://github.com/caskroom/homebrew-cask#reporting-bugs
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/usr/local/Homebrew/Library/Homebrew/brew.rb:24:in `require?'
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/cli.rb:116:in `run_command'
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/cli.rb:158:in `process'
/usr/local/Homebrew/Library/Homebrew/cmd/cask.rb:8:in `cask'
/usr/local/Homebrew/Library/Homebrew/brew.rb:92:in `<main>'
```
- What you expected to happen
I expected the `brew cask` help message to be printed.
- Step-by-step reproduction instructions (by running `brew` commands)
Just run `brew cask` with the latest version of brew from master.

I identified the problem as having been introduced in https://github.com/Homebrew/brew/commit/f0dc1d96c50551be69ea418148a9bc2d72cfaf16 (line 116 of `cli.rb`). I'll be opening a PR to fix this issue.

PR #2652, opened 2017-05-18
Fix implicit conversion of nil into string error

ewang12:

-----
This fixes https://github.com/Homebrew/brew/issues/2651 by explicitly converting `nil` values to an empty string.

cc @reitermarkus

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -21,6 +21,7 @@ if ARGV == %w[--version] || ARGV == %w[-v]
 end
 
 def require?(path)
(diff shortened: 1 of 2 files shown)


Library/Homebrew/brew.rb
```diff
@@ -21,6 +21,7 @@
 end
 
 def require?(path)
+  path ||= ""
```
reitermarkus: Let's make this `return false if path.nil?`, which makes this a bit more clear.

reitermarkus:
Thanks!

PR #2644, opened 2017-05-17
audit: fix full versioned alias name resolution in taps

JCount:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----
This fixes an issue where audit would prompt for the creation
of an already existing versioned alias while at the same time
declaring the existing alias invalid.

This can be done in one line, but it seemed less readable. I am happy to change it, however.

cc @MikeMcQuaid as this is a slight fix to his PR: https://github.com/Homebrew/brew/pull/2619.

Fixes #2650.

PR #2658, opened 2017-05-21
audit: fix versioned alias errors in homebrew/core introduced in a7c6a57

### LINKED ISSUE #2657: Wrong audit failure for versioned alias
The audit for versioned aliases is misbehaving, diagnosing an inexistent problem with the `imagemagick` formula:

```
$ brew audit --strict imagemagick
imagemagick:
  * Formula has other versions so create a versioned alias:
      cd /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Aliases
      ln -s ../Formula/imagemagick.rb imagemagick@7
  * Formula has invalid versioned aliases:
      imagemagick@7
Error: 2 problems in 1 formula
```

The audit is invalid, as the versioned alias already exists, and rightly so.

(Possibly related to #2644 ? poke @JCount @MikeMcQuaid)

JCount:

-----
This fixes #2657 which in turn were caused by an oversight in https://github.com/Homebrew/brew/pull/2644.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -323,7 +323,7 @@ class FormulaAuditor
       end
       valid_alias_names = [alias_name_major, alias_name_major_minor]
 
-      if formula.tap
+      if formula.tap && !formula.tap.core_tap?
         valid_alias_names.map! { |a| "#{formula.tap}/#{a}" }
       end
 


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -323,7 +323,7 @@ def audit_file
       end
       valid_alias_names = [alias_name_major, alias_name_major_minor]
 
-      if formula.tap
+      unless ["homebrew/core", ""].include? formula.tap.to_s
```
MikeMcQuaid: I think the version suggested in the PR was easier to follow but I'd suggest using `formula.tap.core_tap?` instead.

JCount: Ahh well, I guess no esoteric/eccentric styling that doesn't match that of the rest of the code. 😆

PR #2663, opened 2017-05-22
Convert `FormulaLock` to more generic `LockFile`.

reitermarkus:

-----

Converts `FormulaLock` to `LockFile`, which is then again subclassed as `CaskLock` and `FormulaLock`.


Library/Homebrew/lock_file.rb
```diff
@@ -35,3 +35,15 @@ def create_lockfile
     @lockfile.fcntl(Fcntl::F_SETFD, Fcntl::FD_CLOEXEC)
   end
 end
+
+class FormulaLock < LockFile
+  def initialize(name)
+    super("#{name}.formula")
```
MikeMcQuaid: I'm not sure I can see/understand why this change is needed? Can you explain a little more?

Library/Homebrew/cask/lib/hbc/cli/cleanup.rb
```diff
@@ -75,14 +75,18 @@ def delete_paths(paths)
         paths.each do |item|
```
MikeMcQuaid: Any changes needed to Homebrew's cleanup code?

Library/Homebrew/lock_file.rb
```diff
@@ -35,3 +35,15 @@ def create_lockfile
...
+  end
+end
+
+class CaskLock < LockFile
```
MikeMcQuaid: Should this maybe live in the cask directory?

Library/Homebrew/lock_file.rb
reitermarkus: This is just a cosmetic change, so we can easily see whether a lock file is for a Formula or for a Cask.

Library/Homebrew/cask/lib/hbc/cli/cleanup.rb
reitermarkus: No, it is still using the `FormulaLock` as before.

MikeMcQuaid: It's still cleaning up both old and new despite the name change? If so, 👍

Library/Homebrew/lock_file.rb
MikeMcQuaid: I think it would be nice to keep the formulae ones as-is and consider just using the new naming format for those in casks. I don't see the costmetic upsides outweighing the entropy and potential breakage of existing uses.

Library/Homebrew/cask/lib/hbc/cli/cleanup.rb
reitermarkus: Yes, basically it just loops through each file in `HOMEBREW_LOCK_DIR`, tries to lock it, and removes it if it can lock it, so the suffix doesn't matter.

MikeMcQuaid:
@reitermarkus Merge when you're ready and able to be around for a bit in case the cosmetic change goes 💥

PR #2673, opened 2017-05-23
formulary: don't redeclare namespace modules

alyssais:

-----

This has been happening since #2601, and generates a warning. I don't _think_ there's any reason to worry about using the same module, but someone taking a second look would be nice.

@reitermarkus any idea why 330307b01 would have caused this to suddenly start happening?

### DIFF
--- a/Library/Homebrew/formulary.rb
+++ b/Library/Homebrew/formulary.rb
@@ -15,13 +15,20 @@ module Formulary
     FORMULAE.fetch(path)
   end
 
+  def self.namespace_module(namespace)
+    if const_defined?(namespace)
+      const_get(namespace)
+    else
+      Module.new.tap { |m| const_set(namespace, m) }
+    end
+  end
+
   def self.load_formula(name, path, contents, namespace)
     if ENV["HOMEBREW_DISABLE_LOAD_FORMULA"]
       raise "Formula loading disabled by HOMEBREW_DISABLE_LOAD_FORMULA!"
     end
 
-    mod = Module.new
-    const_set(namespace, mod)
+    mod = namespace_module(namespace)
     begin
       mod.module_eval(contents, path)
     rescue ScriptError => e


reitermarkus:
No idea why this is happening. What is the warning, exactly?

MikeMcQuaid:
How can you reproduce the warning here?

alyssais:
`brew tests`

MikeMcQuaid:
@alyssais Can you try to pin down which tests are producing that message? Presumably you identified that PR with `git bisect` or similar. I'm just confused because I don't see anything there that looks like it should cause this change 😢

alyssais:
Does this mean the rest of you aren't getting this message?

MikeMcQuaid:
I see it also I'm just confused by the change that's caused it.

MikeMcQuaid:
CC @reitermarkus to check if you can reproduce this (I can and I agree something needs fixed here).

reitermarkus:
I also see it when running `brew tests`.

MikeMcQuaid:
@reitermarkus Could you see if you can figure out why https://github.com/Homebrew/brew/pull/2601 caused this and if the change can me made without causing that message.

reitermarkus:
https://github.com/Homebrew/brew/pull/2601 doesn't use anything new in terms of loading Formulae, so I think this problem existed before, but we never saw it because we were never loading the same Formula twice. So this is probably the best fix.

MikeMcQuaid:
Could we instead modify the tests in #2601 so we're not loading the same formulae twice? Thanks!

MikeMcQuaid:
Thanks for the change @alyssais but believe this is addressed in https://github.com/Homebrew/brew/issues/2727

reitermarkus:
Well, https://github.com/Homebrew/brew/pull/2727 doesn't actually fix this, it just hides it by not loading the same Formula twice.

MikeMcQuaid:
> it just hides it by not loading the same Formula twice.

I think that's a good thing, I've never seen a situation where loading the same formula twice was the right thing to do. It doesn't quite imply loading the same formula twice; it implies loading the same formula twice in the same Homebrew process which has changed between those load times. There's assumptions all over the codebase on that not happening such that I think this warning is desirable.

alyssais:
Should we perhaps give a better warning messor or raise an exception, then?

MikeMcQuaid:
> raise an exception, then?

👍

MikeMcQuaid:
Passing on this as-is but a better warning/exception would be 👍

PR #2675, opened 2017-05-24
Vendor compatibility fixes

mistydemeo:

-----

This fixes a few compatibility issues with very old versions of OS X.

* `xcode-select` was first shipped in 10.5; scope the `xcode-select` and `xcrun` checks to only run if `/usr/bin/xcode-select` exists.
* While bash arrays exist in bash 2.x, the `+=` syntax doesn't. Arrays need to be appended via `[]` indexing.
* A few `-n` checks on the output of `which` were changed to `-x` checks, since older versions of `which` return a string when no results are found.

### DIFF
--- a/Library/Homebrew/cmd/vendor-install.sh
+++ b/Library/Homebrew/cmd/vendor-install.sh
@@ -76,10 +76,10 @@ fetch() {
     trap - SIGINT
   fi
 
-  if [[ -n "$(which shasum)" ]]
+  if [[ -x "$(which shasum)" ]]
   then
     sha="$(shasum -a 256 "$CACHED_LOCATION" | cut -d' ' -f1)"
-  elif [[ -n "$(which sha256sum)" ]]
+  elif [[ -x "$(which sha256sum)" ]]
   then
     sha="$(sha256sum "$CACHED_LOCATION" | cut -d' ' -f1)"
   else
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/vendor-install.sh
```diff
@@ -39,10 +39,10 @@ fetch() {
 
   if [[ -n "$HOMEBREW_QUIET" ]]
   then
-    curl_args+=(--silent)
+    curl_args[${#curl_args[*]}]="--silent"
```
MikeMcQuaid: This syntax is awful (Bash's fault not yours). Is it worth just avoiding Bash arrays here in favour of space separated strings?

mistydemeo: Yeah, that might be prettier all things considered.

mistydemeo: Looked into it, and I'm not sure that's straightforward to do while keeping the quoted properties of `$HOMEBREW_USER_AGENT_CURL` unfortunately. Even when single-quoting that string within a `curl_args` string, it's interpreted as multiple arguments. Think I may leave it like this, ugly as it is.

MikeMcQuaid: @mistydemeo No worries, thanks for checking ❤️

PR #2683, opened 2017-05-27
Add vendored packages for older versions of OS X

mistydemeo:

-----

This adds portable packages for Tigerbrew's dependencies - Ruby and curl. These packages are built for 10.4+, and are universal builds with i386 and PPC slices. Unfortunately, for technical reasons, I was unable to build the same version of Ruby; an [upstream bug](https://github.com/Homebrew/homebrew-portable/commit/744323ab1195cf088ba782e7a095ce54b9e7bddc) prevents 2.0 from being built universally, and a [different bug](https://bugs.ruby-lang.org/issues/13413) prevents 2.3.0 and newer from being built universally. I chose 2.2.7 as the newest Ruby I could build universally.

I've adjusted the logic for the existing `ruby_URL` so the new packages will be used on 10.4 Intel along with all PPC platforms; 10.5 Intel and newer will use the current Ruby binaries.

I've removed the current logic we use for setting `HOMEBREW_CURL` and pushed that into `util/curl.sh`, like Ruby. On 10.8 and older, we'll install the portable curl. Unfortunately, thanks to the chicken-and-egg problem, this necessitates downloading the prebuilt packages via curl's `--insecure` option.

Since PPC machines and 10.4 Intel are downloading a different Ruby, I've added the ability to specify what version to install on the commandline in order to avoid dropping a second `portable-ruby-version` file. This is awkward and I hope to be able to work around it at some point in the future when we can package the same version of Ruby on all platforms.

I'll be following this PR with one other that adds a falback `sha256` tool for platforms that don't include one bundled; without that, older platforms can't yet use the packages downloaded in this PR.

cc @xu-cheng


xu-cheng:
I'm not sure why do you need vendor openssl? For SHA, is not the vendor Ruby already efficient to do the job?

Also FYI, vendor curl bundles its own SSL cert file, so there is no need to set SSL_CERT_FILE

mistydemeo:
> For SHA, is not the vendor Ruby already efficient to do the job?

You know, that slipped my mind. It should be capable of that, in which case shelling out should no longer be necessary.

> Also FYI, vendor curl bundles its own SSL cert file, so there is no need to set SSL_CERT_FILE

I totally missed that! Thanks.

I'll remove openssl.

MikeMcQuaid:
I think for `curl` given we're having to use the system version to do an initial insecure download it's worth building the formula rather than vendoring it. Vendoring Ruby allows us to use the massive amount of existing Ruby logic and migrate it all to Ruby 2.0 but I'm not seeing the same level of advantage for a vendored `curl` vs. using the `curl` formula. The latter also means we're more likely to be on top of security updates.

mistydemeo:
I've been doing that for awhile, but it runs into a notable issue that upgrading curl itself is a pain - Homebrew unlinks the existing curl during the upgrade, for example, which breaks the fetch. Even with `--insecure`, the older curl is also unable to fetch the source package for curl itself, which makes getting it a real pain. This is going to massively simplify things for Tigerbrew users.

MikeMcQuaid:
> Even with --insecure, the older curl is also unable to fetch the source package for curl itself, which makes getting it a real pain. This is going to massively simplify things for Tigerbrew users.

Can you elaborate on this a bit?

With both this and #2684 I'm wondering if there's any way we can push any of this stuff into e.g. an external command that lives in Tigerbrew/homebrew-core. I'd really like to have Tigerbrew be able to run Homebrew/brew unpatched eventually but want to avoid large chunks of relatively complex, untested code that are Tigerbrew (or Linuxbrew) specific.

MikeMcQuaid:
> With both this and #2684 I'm wondering if there's any way we can push any of this stuff into e.g. an external command that lives in Tigerbrew/homebrew-core.

Another thought: this stuff could be setup by the install script instead because that's never going to be fully shared between Homebrew and Tigerbrew.



> Even with --insecure, the older curl is also unable to fetch the source package for curl itself, which makes getting it a real pain.

Another thought: `HOMEBREW_CURL` is used on <10.9 for this already so it would be good to improve the existing flow so it works better, perhaps?

DomT4:
> Another thought: HOMEBREW_CURL is used on <10.9 for this already so it would be good to improve the existing flow so it works better, perhaps?

Since you mentioned it, ideally would be <10.11 at this point. Yosemite keeps blowing up CI on various secure URLs lately & if it's doing that for us presumably it does for users too.

mistydemeo:
> Another thought: HOMEBREW_CURL is used on <10.9 for this already so it would be good to improve the existing flow so it works better, perhaps?

> Since you mentioned it, ideally would be <10.11 at this point. Yosemite keeps blowing up CI on various secure URLs lately & if it's doing that for us presumably it does for users too.

In that case, maybe a way we can incorporate this with the flow of Homebrew is to use the vendored curl on 10.10 and older, and make use of it in both Homebrew and Tigerbrew?

> Can you elaborate on this a bit?

Curl's source tarball is served up via TLS 1.2 only; it doesn't allow any older protocols. Meanwhile the system curl on older versions of OS X does TLS 1.0 at the newest. The `--insecure` still uses SSL/TLS, it just doesn't verify the server's certificate; doesn't help if the client and server don't speak a common protocol.

MikeMcQuaid:
> if it's doing that for us presumably it does for users too.

I don't think it's worth presuming that without any reported user issues. I'd like to keep us on the system `curl` there until 10.13 is released.

> In that case, maybe a way we can incorporate this with the flow of Homebrew is to use the vendored curl on 10.10 and older, and make use of it in both Homebrew and Tigerbrew?

I'd really like to avoid keeping a separate `curl` up to date with the `curl` formulae. `ruby` is a very different beast as we're bundling an intentionally old version that matches the macOS system one. If there was some way this vendored `curl` install could e.g. use the bottle or similar I'd be less against it but I'd rather fix the formula flow for all non-`ruby` things to work better in this way. I don't think the e.g. unlinking issues are insurmountable or the system `curl` is unusable for e.g. upgrading itself.

> Curl's source tarball is served up via TLS 1.2 only; it doesn't allow any older protocols.

There are HTTP mirrors available for `curl` e.g. https://curl.haxx.se/download/curl-7.54.0.tar.gz. If we have a SHA256 hash for it the threat model requires breaking that hash. Given this would only apply on older versions of macOS that are not receiving security updates anyway I think that's the least of their/our worries.

MikeMcQuaid:
@ilovezfs has confirmed with adding http://curl.askapache.com/download/curl-7.54.0.tar.bz2 as a mirror `curl` installs as expected with a system `curl` that refuses to HTTPS to the primary mirror. It also displays `curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version` which is a nice signalling mechanism to the user.

This definitely seems like the right approach to me; use the system `curl` to download the Homebrew `curl` package over HTTP and then use the Homebrew `curl` from there onwards. We can add an `audit` to ensure that `curl` always has an HTTP mirror.

ilovezfs:
In bootstrapping brew on 10.8 without any code changes, and immediately trying to `brew install --with-openssl curl`, the first and only failure I encountered is
```
==> Installing curl --with-openssl
==> Downloading https://curl.haxx.se/download/curl-7.54.0.tar.bz2

curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
Error: Failed to download resource "curl"
Download failed: https://curl.haxx.se/download/curl-7.54.0.tar.bz2
```
which is resolved by adding an http mirror to the curl formula.

`brew reinstall` works as well if an http mirror is available.

So it seems to me that the only thing the vendoring of curl achieves is being able to use the canonical URL as opposed to one of the many http mirrors, of which our own Bintray mirror could be one.

mistydemeo:
Okay, how about this version? It checks to see if a core tap is present and has a `cmd/vendor-curl.sh`; if it is, it sources that command and runs `setup-curl-path`.

ilovezfs:
I think this should just use the curl formula and an http url.

MikeMcQuaid:
> I think this should just use the curl formula and an http url.

Agreed.

Library/Homebrew/cmd/vendor-install.sh
```diff
@@ -11,13 +11,22 @@ VENDOR_DIR="$HOMEBREW_LIBRARY/Homebrew/vendor"
 # Built from https://github.com/Homebrew/homebrew-portable.
 if [[ -n "$HOMEBREW_MACOS" ]]
 then
-  if [[ "$HOMEBREW_PROCESSOR" = "Intel" ]]
+  # Universal 10.4 build
+  if [[ "$HOMEBREW_MACOS_VERSION_NUMERIC" -lt "100500" || "$HOMEBREW_PROCESSOR" != "Intel" ]]
   then
+    ruby_URL="https://archive.org/download/tigerbrew/portable-ruby-2.2.7.tiger.bottle.tar.gz"
```
MikeMcQuaid: Pretty sure you told me already but remind me: why 2.2 rather than 2.0 or even 2.4?

mistydemeo: 2.0 has a bug which prevents it from being built as a universal binary, and 2.4 has a bug which prevents it from being built statically.

mistydemeo:
> I think this should just use the curl formula and an http url.

This has been the status quo for Tigerbrew for a few years, and the general consensus is that the user experience sucks. I'm quite set on vendoring a curl being the better path forward.

ilovezfs:
>the general consensus is that the user experience sucks

Can you elaborate? I'm sure I don't understand since `brew install curl` seems like it shouldn't be a bad UX unless something else is broken.

Library/Homebrew/cmd/vendor-install.sh
MikeMcQuaid: Ok, thanks!

MikeMcQuaid:
> This has been the status quo for Tigerbrew for a few years, and the general consensus is that the user experience sucks. I'm quite set on vendoring a curl being the better path forward.

To elaborate on my position: I'd be OK with vendoring a `curl` if a PR that upgrades `curl` in homebrew/core also builds, bottles and uploads the vendored `curl` requiring no more interaction from a Homebrew maintainer. Otherwise we have a small number of maintainers (possibly one) who we need to wait on if there is a `curl` CVE. Obviously Tigerbrew can take whatever approach makes sense there but I'd really like to have Homebrew/brew be usable for Tigerbrew (and Linuxbrew) unmodified eventually but this means the code needs to meet the same standards and maintenance of it be shared.

To be clear, I'm not saying the current `HOMEBREW_CURL` use is perfect but I'd like to at least try and detail all the problems with it and improve it before we give up on it. I'm willing to do work to make that happen.

MikeMcQuaid:
> which is resolved by adding an http mirror to the curl formula.

We've done this and adding an audit method to check this in future.

MikeMcQuaid:
@mistydemeo Can this be scaled back to be just the Ruby version addition? At this point it may be worth having one version matching High Sierra's Ruby and use that everywhere, right?

MikeMcQuaid:
Closing this as there's been no updates for over a month.

MikeMcQuaid:
Will consider a new PR scoped only to the Ruby addition (preferably the High Sierra version: `ruby 2.3.3p222`).

PR #2684, opened 2017-05-27
Add vendored sha256

mistydemeo:

-----

This adds a pure-Perl sha256 implementation, which is compatible with essentially any version of Perl released in the last 15 years. It's suitable for use on older versions of OS X, or Linuces which lack a `shasum` or `sha256sum`; this makes it very useful for installing a vendored Ruby.

This script was originally contributed to Tigerbrew by @geoff-codes.

cc @xu-cheng, @sjackman

### DIFF
--- a/Library/Homebrew/cmd/vendor-install.sh
+++ b/Library/Homebrew/cmd/vendor-install.sh
@@ -82,6 +82,15 @@ fetch() {
   elif [[ -x "$(which sha256sum)" ]]
   then
     sha="$(sha256sum "$CACHED_LOCATION" | cut -d' ' -f1)"
+  elif [[ -x "$(which ruby)" ]]
+  then
+    sha="$(ruby <<EOSCRIPT
+            require 'digest/sha2'
+            digest = Digest::SHA256.new
+            File.open('$CACHED_LOCATION', 'rb') { |f| digest.update(f.read) }
+            puts digest.hexdigest
+EOSCRIPT
+)"
   else
     odie "Cannot verify the checksum ('shasum' or 'sha256sum' not found)!"
   fi
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -322,7 +322,7 @@ class Pathname
 
   def sha256
     require "digest/sha2"
-    incremental_hash(Digest::SHA2)
+    incremental_hash(Digest::SHA256)
   end
 
   def verify_checksum(expected)


MikeMcQuaid:
Is there nothing in e.g. old Rubies or other macOS system libraries on Tigerbrew versions that could calculate a SHA256? I begrudingly think upstreaming Tigerbrew code to Homebrew/brew is worth it despite the lower user base but I'm much less convinced for language modules and particularly non-Ruby modules.

mistydemeo:
> Is there nothing in e.g. old Rubies or other macOS system libraries on Tigerbrew versions that could calculate a SHA256?

Absolutely nothing. Perl doesn't include this, OpenSSL is missing it, there's no CommonCrypto, and there are no other tools with support. Without this, there's no way to calculate sha256 checksums at all.

MikeMcQuaid:
With both this and #2683 I'm wondering if there's any way we can push any of this stuff into e.g. an external command that lives in Tigerbrew/homebrew-core. I'd really like to have Tigerbrew be able to run Homebrew/brew unpatched eventually but want to avoid large chunks of relatively complex, untested code that are Tigerbrew (or Linuxbrew) specific.

DomT4:
Is the vendored Ruby the default on older macOS versions yet?

If so you could simply drop all of the `require "digest"` in Homebrew in favour of `require "openssl"` and gain universal SHA256 support that way no? I'm simplifying the changes required here obviously, but, still seems less drastic than explicitly vendoring another implementation.

DomT4:
My comment was stupid, I apologise 😄.

mistydemeo:
Having just made a ridiculous discovery, I'm rethinking this. It turns out Ruby 1.8.2 *does* include sha2 - not via openssl, but via a vendored copy of Aaron Gifford's sha2 tool. The constant name is different, but otherwise the API is compatible. Fortunately, newer Rubies include the `Digest::SHA256` name for compatibility, so a very small patch resolves that. I've made this much less invasive, and added a small clause to let `vendor-install` use Ruby as a fallback if shasum/sha256sum aren't present.

>  I'm wondering if there's any way we can push any of this stuff into e.g. an external command that lives in Tigerbrew/homebrew-core.

I'm not sure there's an easy way to do that with curl, unfortunately. :/ Setting up curl happens so early in the process of bootstrapping Homebrew, it doesn't really feel like there's a nice way to bootstrap a curl without using the same mechanism that's already being used for Ruby.

Library/Homebrew/cmd/vendor-install.sh
```diff
@@ -82,6 +82,9 @@ fetch() {
   elif [[ -x "$(which sha256sum)" ]]
   then
     sha="$(sha256sum "$CACHED_LOCATION" | cut -d' ' -f1)"
+  elif [[ -x "$(which ruby)" ]]
+  then
+    sha="$(ruby -e "require 'digest/sha2'; digest = Digest::SHA256.new; File.open('$CACHED_LOCATION', 'rb') { |f| digest.update(f.read) }; puts digest.hexdigest")"
```
MikeMcQuaid: Is it worth using a heredoc here so this can live on multiple lines?

MikeMcQuaid:
One suggestion but once green (I suspect after a rebase) I'm 👍 to ship. Nice work finding the SHA256 thing 👏

Library/Homebrew/cmd/vendor-install.sh
mistydemeo: Good idea, done! Idented the Ruby code in slightly past the `ruby` line.

Library/Homebrew/cmd/vendor-install.sh
```diff
@@ -82,6 +82,15 @@ fetch() {
   elif [[ -x "$(which sha256sum)" ]]
   then
     sha="$(sha256sum "$CACHED_LOCATION" | cut -d' ' -f1)"
+  elif [[ -x "$(which ruby)" ]]
+  then
+    sha="$(ruby <<EOSCRIPT
+            require 'digest/sha2'
+            digest = Digest::SHA256.new
+            File.open('$CACHED_LOCATION', 'rb') { |f| digest.update(f.read) }
+            puts digest.hexdigest
```
MikeMcQuaid: Nice!

PR #2687, opened 2017-05-27
Match the "formula name in description" on word boundaries

issyl0:

-----

- The regexp for the "check if formula name is used in formula's
  description" cop matches every instance of the formula name if it
  exists, whether it's in a word or not.
- For example, the formula `mon` has the description "Monitor
  hosts/services/whatever and alert about problems". This makes
  `brew audit --strict` complain because it matches "Monitor",
  which isn't the formula name! The formula `pass` has the description
  "Password manager".  Again, the strict audit matches "Password",
  which isn't an issue.
- Instead, this change matches on a word boundary, so it will match
  `mon:`, or `mon `, but not "Monitor", or, for example, "harmony".
- I've changed the tests to account for this change.

### DIFF
--- a/Library/Homebrew/rubocops/formula_desc_cop.rb
+++ b/Library/Homebrew/rubocops/formula_desc_cop.rb
@@ -40,7 +40,7 @@ module RuboCop
           end
 
           # Check if formula's name is used in formula's desc
-          problem "Description shouldn't include the formula name" if regex_match_group(desc, /^#{@formula_name}/i)
+          problem "Description shouldn't include the formula name" if regex_match_group(desc, /^#{@formula_name}\b/i)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


GauthamGoli:
Thanks @issyl0 for finding this bug and the PR!

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -40,7 +40,7 @@ def audit_formula(_node, _class_node, _parent_class_node, body)
           end
 
           # Check if formula's name is used in formula's desc
-          problem "Description shouldn't include the formula name" if regex_match_group(desc, /^#{@formula_name}/i)
+          problem "Description shouldn't include the formula name" if regex_match_group(desc, /^\b#{@formula_name}\b/i)
```
bfontaine: You can drop the first `\b`; `^` already ensures it’s the first word.

issyl0: Thanks. I've amended my commit.

MikeMcQuaid:
Thanks for your first contribution to Homebrew, @issyl0! Without people like you submitting PRs we couldn't run this project. You rock!

ilovezfs:
This still seems to be an issue. The `rack` formula's description is, "CLI for Rackspace" and the audit is now complaining, `* C: 2: col 12: Description shouldn't include the formula name.`

PR #2691, opened 2017-05-29
Make `--official-cmd-taps` imply `--online`.

reitermarkus:

-----

Tests that need a network connection aren't currently run on CI, but tests that use official command taps (i.e. also require a network connection) are, which is inconsistent.


MikeMcQuaid:
Heh, the one test fails here. Manpage would need to be updated, too.

In general I think it's worth considering installing more stuff that's needed for pending tests or removing them. They are just noise at this point, otherwise.

MikeMcQuaid:
> I think it's worth considering installing more stuff that's needed for pending tests or removing them. They are just noise at this point, otherwise.

@reitermarkus Any thoughts on this?

PR #2696, opened 2017-05-29
check_non_libraries: fix false positive subdirectory reports

mistydemeo:

-----

This fixes a bug introduced in 7a38bab333c6808022fe53aac2be9ca2e329cd53. Since `check_non_libraries` now uses `reject` instead of `select`, the value returned for the `g.directory?` check has to be inverted too; otherwise, non-directories stay in the array and will be reported as inappropriately-installed files.

Fixes warnings like this one:

```
Non-libraries were installed to "/usr/local/opt/imagemagick/lib"
Installing non-libraries to "lib" is discouraged.
The offending files are:
  /usr/local/opt/imagemagick/lib/ImageMagick
  /usr/local/opt/imagemagick/lib/pkgconfig
```

### DIFF
--- a/Library/Homebrew/formula_cellar_checks.rb
+++ b/Library/Homebrew/formula_cellar_checks.rb
@@ -62,7 +62,7 @@ module FormulaCellarChecks
     valid_extensions = %w[.a .dylib .framework .jnilib .la .o .so
                           .jar .prl .pm .sh]
     non_libraries = formula.lib.children.reject do |g|
-      next if g.directory?
+      next true if g.directory?
       valid_extensions.include? g.extname
     end
     return if non_libraries.empty?


mistydemeo:
> Have you written new tests for your changes? Here's an example.

There currently aren't tests for this module, which I'd like to see added at this point, but I don't want that to delay this fix.

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -62,8 +62,11 @@ def check_non_libraries
     valid_extensions = %w[.a .dylib .framework .jnilib .la .o .so
                           .jar .prl .pm .sh]
     non_libraries = formula.lib.children.reject do |g|
-      next if g.directory?
-      valid_extensions.include? g.extname
+      if g.directory?
```
MikeMcQuaid: `next true if g.directory?`

mistydemeo: I considered it, but I don't find `next true` to be very clear, especially for people who are new to Ruby. Glad to change it if you have a strong opinion though.

mistydemeo: I realized we do use that syntax elsewhere, after reviewing a few other `select/reject` calls, so updated to that.

MikeMcQuaid:
Thanks for this @mistydemeo!

PR #2697, opened 2017-05-29
cabextract: WARNING lines aren't fatal

### LINKED ISSUE #2689: Nested CAB containers fail on Warnings
- [X] Confirmed this is a problem with running a `brew` command and just `brew install`ing one or more formulae?
- [X] Ran `brew doctor`, fixed all issues and retried your prior step?
- [X] Ran `brew config` and `brew doctor` and included their output with your issue?

I'm trying to install [this font cask formula](https://github.com/joeyhoer/homebrew-extras/blob/master/Casks/font-microsoft-office.rb), but homebrew cannot properly detect the nested container. *Note:* I've confirmed this is an issue with homebrew, not cask.

I've determined that the issue is a result of `cabextract` writing warnings (in addition to errors) to stderr. [Homebrew attempts to validate the cab file with `.stderr.empty?`](https://github.com/Homebrew/brew/blob/19347f3b87917368d7157fe5007d506a4db0c7db/Library/Homebrew/cask/lib/hbc/container/cab.rb#L13) which prevents the correct container type from being detected, and thus throws an error.

When the test is run:

    # Write only stderr to stdout
    $ cabextract -t -- ~/Library/Caches/Homebrew/Cask/font-microsoft-office--latest/ppviewer.cab 2>&1 >/dev/null
    /Library/Caches/Homebrew/Cask/font-microsoft-office--latest/ppviewer.cab: WARNING; possible 5968 extra bytes at end of file.

but the file is actually valid, with no errors, and returns a clean exit code:

    $ cabextract -t -- ~/Library/Caches/Homebrew/Cask/font-microsoft-office--latest/ppviewer.cab
     …
     All done, no errors.

mistydemeo:
Fixes #2689.

-----

As reported in #2689, it's possible for `cabextract` to issue non-fatal `WARNING`s to stderr. Right now cask's cab support treats that as an error, but it seems safe to filter that out. I confirmed the formula @joeyhoer was trying to install works with this patch.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/container/cab.rb
+++ b/Library/Homebrew/cask/lib/hbc/container/cab.rb
@@ -6,11 +6,7 @@ module Hbc
   class Container
     class Cab < Base
       def self.me?(criteria)
-        cabextract = which("cabextract")
-
-        criteria.magic_number(/^(MSCF|MZ)/n) &&
-          !cabextract.nil? &&
-          criteria.command.run(cabextract, args: ["-t", "--", criteria.path.to_s]).stderr.empty?
+        criteria.magic_number(/^(MSCF|MZ)/n)
       end
 
       def extract


Library/Homebrew/cask/lib/hbc/container/cab.rb
```diff
@@ -8,9 +8,10 @@ class Cab < Base
       def self.me?(criteria)
         cabextract = which("cabextract")
 
+        # Ignores lines containing WARNING, but fails on other errors.
         criteria.magic_number(/^(MSCF|MZ)/n) &&
           !cabextract.nil? &&
-          criteria.command.run(cabextract, args: ["-t", "--", criteria.path.to_s]).stderr.empty?
+          criteria.command.run(cabextract, args: ["-t", "--", criteria.path.to_s]).stderr.split("\n").reject { |l| l.include? "WARNING" }.empty?
```
reitermarkus: The exit status is still `0` when there are warnings, right? If so, I think we can rely on the `exit_status` instead of checking the `stderr` output.

Also – highly unlikely to happen – this would fail if the file path contained “WARNING” and the exit code was non-zero.

mistydemeo: Hah, good point! I'll change it to an exit status check.

Just to make sure, is this an `stderr` check because errors can exit 0?

reitermarkus: > Just to make sure, is this an `stderr` check because errors can exit 0?

I really don't know, sorry.

Then again, I think we could remove this whole check, since it is also unlikely that the magic number matches, but it isn't a `cab` container. In the “worst” case, it will then fail when extracting it.

mistydemeo: Good idea, removed the check.

reitermarkus:
Actually I meant removing everything except for the magic number check.

mistydemeo:
Oh, of course! Removed.

Library/Homebrew/cask/lib/hbc/container/cab.rb
```diff
@@ -8,9 +8,7 @@ class Cab < Base
       def self.me?(criteria)
         cabextract = which("cabextract")
```
reitermarkus: This is not needed anymore.

mistydemeo: Removed.

PR #2699, opened 2017-05-29
Fix boolean types

mistydemeo:

-----

This fixes a few return types, which changed from `Boolean` to `~Fixnum` in 7a38bab333c6808022fe53aac2be9ca2e329cd53. `===` is not my favourite bit of Ruby syntax, but Rubocop likes it more than it likes `!!`.

### DIFF
--- a/Library/Homebrew/extend/string.rb
+++ b/Library/Homebrew/extend/string.rb
@@ -1,3 +1,6 @@
+# Contains backports from newer versions of Ruby
+require_relative "../vendor/backports/string"
+
 class String
   def undent
     gsub(/^[ \t]{#{(slice(/^[ \t]+/) || '').length}}/, "")
--- a/Library/Homebrew/vendor/backports/string.rb
+++ b/Library/Homebrew/vendor/backports/string.rb
@@ -1,0 +1,12 @@
+# Taken from https://github.com/marcandre/backports/blob/v3.8.0/lib/backports/2.4.0/string/match.rb
+unless String.method_defined? :match?
+  class String
+    def match?(*args)
+      # Fiber to avoid setting $~
+      f = Fiber.new do
+        !match(*args).nil?
+      end
+      f.resume
+    end
+  end
+end
--- a/Library/Homebrew/cask/lib/hbc/cask_loader.rb
+++ b/Library/Homebrew/cask/lib/hbc/cask_loader.rb
@@ -54,7 +54,7 @@ module Hbc
 
     class FromURILoader < FromPathLoader
       def self.can_load?(ref)
-        ref.to_s =~ ::URI.regexp
+        ref.to_s.match?(::URI.regexp)
       end
 
       def initialize(url)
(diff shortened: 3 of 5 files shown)


reitermarkus:
> `===` is not my favourite bit of Ruby syntax

I actually had in mind backporting Ruby 2.4.0's [`match?`](https://ruby-doc.org/core-2.4.0/Regexp.html#method-i-match-3F) for a while, this might be the time for it.

mistydemeo:
> I actually had in mind backporting Ruby 2.4.0's match? for a while, this might be the time for it.

That would be perfect, yeah.

reitermarkus:
Take a look at https://github.com/marcandre/backports/tree/master/lib/backports/2.4.0.

mistydemeo:
Backported the `String` version.

Library/Homebrew/extend/string.rb
```diff
@@ -79,3 +79,36 @@ def get_make_var(flag)
     self[/^#{Regexp.escape(flag)}[ \t]*=[ \t]*(.*)$/, 1]
   end
 end
+
+# Taken from https://github.com/marcandre/backports
```
MikeMcQuaid: Can we stick this in some sort of `vendor` directory and link directly to https://github.com/marcandre/backports/blob/master/lib/backports/2.4.0/string/match.rb?

mistydemeo: Done!

Library/Homebrew/global.rb
```diff
@@ -5,6 +5,7 @@
 require "extend/ARGV"
 require "PATH"
 require "extend/string"
+require "vendor/backports/string"
```
MikeMcQuaid: Maybe worth requiring this inside `require "extend/string"` and adding a comment in there noting we're extending string in another place?

Library/Homebrew/vendor/backports/string.rb
```diff
@@ -0,0 +1,32 @@
+# Taken from https://github.com/marcandre/backports/blob/v3.8.0/lib/backports/2.4.0/string/match.rb
```
MikeMcQuaid: Can the license and other metadata go in `Library/Homebrew/vendor/README.md` to be consistent with the other vendored stuff?

mistydemeo: Done!

Library/Homebrew/global.rb
mistydemeo: Done!

MikeMcQuaid:
Thanks @mistydemeo, looks good now!

Library/Homebrew/global.rb
mistydemeo: I realized afterwards that `extend/string` is required by non-Homebrew code (like our rubocop extensions), and those don't have the right `$LOAD_PATH` to access the vendor directory. Going to move this back to `global`.

MikeMcQuaid: I'd rather we figure out how to fix that or just avoid the backport at this point, to be honest.

mistydemeo:
Should be fixed now, will merge when CI goes green.

PR #2700, opened 2017-05-30
Stdenv: Add ENV.libxml2 and ENV.x11 for Linux

sjackman:

-----

Add `ENV.libxml2` primarily for the use of `test do` blocks. Also add a dummy `ENV.x11` function.

### DIFF
--- a/Library/Homebrew/extend/os/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/os/extend/ENV/std.rb
@@ -1,2 +1,6 @@
 require "extend/ENV/std"
-require "extend/os/mac/extend/ENV/std" if OS.mac?
+if OS.mac?
+  require "extend/os/mac/extend/ENV/std"
+elsif OS.linux?
+  require "extend/os/linux/extend/ENV/std"
+end
--- a/Library/Homebrew/extend/os/linux/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/os/linux/extend/ENV/std.rb
@@ -1,0 +1,7 @@
+module Stdenv
+  def libxml2
+    append "CPPFLAGS", "-I#{Formula["libxml2"].include/"libxml2"}"
+  rescue FormulaUnavailableError
+    nil
+  end
+end
--- a/Library/Homebrew/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/ENV/std.rb
@@ -210,6 +210,8 @@ module Stdenv
   end
   alias generic_set_cpu_flags set_cpu_flags
 
+  def x11; end
+
   # @private
   def effective_arch
     if ARGV.build_bottle?


Library/Homebrew/extend/os/linux/extend/ENV/std.rb
```diff
@@ -0,0 +1,10 @@
+module Stdenv
+  # Add libxml2 to CPPFLAGS
```
MikeMcQuaid: Can remove this comment; the code is readable enough for it to be self-explanatory.

Library/Homebrew/extend/os/linux/extend/ENV/std.rb
```diff
@@ -0,0 +1,10 @@
...
+  def libxml2
+    append "CPPFLAGS", "-I#{Formula["libxml2"].include/"libxml2"}"
+  rescue FormulaUnavailableError
+    nil
+  end
+
+  def x11; end
```
MikeMcQuaid: I'd declare this in the base class as there's nothing Linux-specific here.

Library/Homebrew/extend/os/extend/ENV/std.rb
```diff
@@ -1,2 +1,3 @@
 require "extend/ENV/std"
 require "extend/os/mac/extend/ENV/std" if OS.mac?
```
MikeMcQuaid: `if`/`elsif` might be nice here if it doesn't contradict what we do elsewhere?

sjackman:
Thanks for the review, Mike! Updated.

sjackman:
Using `Formula["libxml2"]` causes `Homebrew/core` to auto-tap. The auto-tap checks that the syntax of all tapped formulae is valid. That fails because the constant `MacOS` is undefined on Linux.

Here's a few options:
1. Prevent auto-tapping `Homebrew/core` on Linux.
2. Add a stubbed-out implementation of `MacOS` on Linux that returns `false`/`Version::NULL` for all tests.
3. Use `-I#{HOMEBREW_PREFIX}/include/libxml2` rather than `-I#{Formula["libxml2"].include/"libxml2"}`

MikeMcQuaid:
@sjackman Looks like it may be working now.

sjackman:
Well that's mysterious. Does Jenkins keep the tapped Homebrew/core from between runs of CI?

MikeMcQuaid:
> Well that's mysterious. Does Jenkins keep the tapped Homebrew/core between runs of CI?

It does not but I have a filthy hack which makes Linux use the single `testbottest` formula and pretend that it's all of Homebrew/core.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #2706, opened 2017-05-30
Added tests for disk_cleanup and unremovable kegs

mansimarkaur:

-----

### DIFF
--- a/Library/Homebrew/test/cleanup_spec.rb
+++ b/Library/Homebrew/test/cleanup_spec.rb
@@ -63,11 +63,6 @@ describe Homebrew::Cleanup do
     end
   end
 
-  specify "::disk_cleanup_size" do
-    disk_cleanup_size = described_class.instance_variable_get(:@disk_cleanup_size)
-    expect(described_class.disk_cleanup_size).to eq(disk_cleanup_size)
-  end
-
   specify "::cleanup_formula" do
     f1 = Class.new(Testball) do
       version "1.0"
@@ -109,14 +104,30 @@ describe Homebrew::Cleanup do
     expect(f4).to be_installed
   end
 
-  specify "::cleanup_logs" do
-    path = (HOMEBREW_LOGS/"delete_me")
-    path.mkpath
-    ARGV << "--prune=all"
+  describe "::cleanup_logs" do
+    before do
+      path.mkpath
+    end
 
-    described_class.cleanup_logs
+    let(:path) { (HOMEBREW_LOGS/"delete_me") }
 
-    expect(path).not_to exist
+    it "cleans all logs if prune all" do
+      ARGV << "--prune=all"
+      described_class.cleanup_logs
+      expect(path).not_to exist
+    end
+
+    it "cleans up logs if older than 14 days" do
+      allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 15)
+      described_class.cleanup_logs
+      expect(path).not_to exist
+    end
+
+    it "does not clean up logs less than 14 days old" do
+      allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
+      described_class.cleanup_logs
+      expect(path).to exist
+    end
   end
 
   describe "::cleanup_cache" do
@@ -127,6 +138,15 @@ describe Homebrew::Cleanup do
       described_class.cleanup_cache
 
       expect(incomplete).not_to exist
+    end
+
+    it "cleans up 'glide_home'" do
+      glide_home = (HOMEBREW_CACHE/"glide_home")
+      glide_home.mkpath
+
+      described_class.cleanup_cache
+
+      expect(glide_home).not_to exist
     end
 
     it "cleans up 'java_cache'" do
@@ -151,20 +171,28 @@ describe Homebrew::Cleanup do
       git = (HOMEBREW_CACHE/"gist--git")
       gist = (HOMEBREW_CACHE/"gist")
       svn = (HOMEBREW_CACHE/"gist--svn")
+
       git.mkpath
       gist.mkpath
       FileUtils.touch svn
+
       allow(ARGV).to receive(:value).with("prune").and_return("all")
-      begin
-        described_class.cleanup_cache
-        expect(git).not_to exist
-        expect(gist).to exist
-        expect(svn).not_to exist
-      ensure
-        FileUtils.rm_rf(git)
-        FileUtils.rm_rf(gist)
-        FileUtils.rm_rf(svn)
-      end
+
+      described_class.cleanup_cache
+
+      expect(git).not_to exist
+      expect(gist).to exist
+      expect(svn).not_to exist
+    end
+
+    it "does not clean up directories that are not VCS checkouts" do
+      git = (HOMEBREW_CACHE/"git")
+      git.mkpath
+      allow(ARGV).to receive(:value).with("prune").and_return("all")
+
+      described_class.cleanup_cache
+
+      expect(git).to exist
     end
 
     it "cleans up VCS checkout directories with modified time < prune time" do
@@ -172,12 +200,16 @@ describe Homebrew::Cleanup do
       foo.mkpath
       allow(ARGV).to receive(:value).with("prune").and_return("1")
       allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
-      begin
-        described_class.cleanup_cache
-        expect(foo).not_to exist
-      ensure
-        FileUtils.rm_rf(foo)
-      end
+      described_class.cleanup_cache
+      expect(foo).not_to exist
+    end
+
+    it "does not clean up VCS checkout directories with modified time >= prune time" do
+      foo = (HOMEBREW_CACHE/"--foo")
+      foo.mkpath
+      allow(ARGV).to receive(:value).with("prune").and_return("1")
+      described_class.cleanup_cache
+      expect(foo).to exist
     end
 
     context "cleans old files in HOMEBREW_CACHE" do
@@ -213,21 +245,27 @@ describe Homebrew::Cleanup do
       end
 
       it "cleans up file if stale" do
-        puts described_class.cleanup_cache
+        described_class.cleanup_cache
         expect(bottle).not_to exist
         expect(testball).not_to exist
       end
     end
   end
 
-  specify "::prune?" do
-    foo = mktmpdir/"foo.rb"
-    foo.mkpath
-    allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
-    begin
+  describe "::prune?" do
+    before do
+      foo.mkpath
+    end
+
+    let(:foo) { mktmpdir/"foo.rb" }
+
+    it "returns true when path_modified_time < days_default" do
+      allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
       expect(described_class.prune?(foo, days_default: "1")).to be_truthy
-    ensure
-      FileUtils.rm_rf(foo)
+    end
+
+    it "returns true when path_modified_time >= days_default" do
+      expect(described_class.prune?(foo, days_default: "2")).to be_falsey
     end
   end
 end
--- a/Library/Homebrew/test/cleanup_spec.rb
+++ b/Library/Homebrew/test/cleanup_spec.rb
@@ -63,20 +63,9 @@ describe Homebrew::Cleanup do
     end
   end
 
-  specify "::update_disk_cleanup_size" do
-    described_class.instance_eval("@disk_cleanup_size = 0")
-    described_class.update_disk_cleanup_size(128)
-    expect(described_class.instance_variable_get("@disk_cleanup_size")).to eq(128)
-  end
-
   specify "::disk_cleanup_size" do
     described_class.instance_eval("@disk_cleanup_size = 0")
     expect(described_class.disk_cleanup_size).to eq(described_class.instance_variable_get("@disk_cleanup_size"))
-  end
-
-  specify "::unremovable_kegs" do
-    described_class.unremovable_kegs
-    expect(described_class.instance_variable_get("@unremovable_kegs")).to eq([])
   end
 
   specify "::cleanup_formula" do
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
> codecov/project — 64.41% (-0.01%) compared to 90c9716

@mansimarkaur It's worth noting that this doesn't actually improve the overall code coverage of the project, unfortunately. Take a look through to that URL and check out the areas that need coverage. Thanks!

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -69,6 +69,28 @@
...
+  specify "::update_disk_cleanup_size" do
+    shutup do
+      described_class.instance_eval("@disk_cleanup_size = 0")
+      described_class.update_disk_cleanup_size(128)
+    end
+    expect(described_class.instance_variable_get("@disk_cleanup_size")).to eq(128)
+  end
+
+  specify "::disk_cleanup_size" do
+    shutup do
```
MikeMcQuaid: I don't think this should be required here?

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -69,6 +69,28 @@
...
+      described_class.instance_eval("@disk_cleanup_size = 0")
+    end
+    expect(described_class.disk_cleanup_size).to eq(described_class.instance_variable_get("@disk_cleanup_size"))
```
MikeMcQuaid: I'm not sure it's worth verifying internal class variables are set; test the externally visible functionality instead (which may mean not having tests for every method)

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -69,6 +69,28 @@
     end
   end
 
+  specify "::update_disk_cleanup_size" do
+    shutup do
+      described_class.instance_eval("@disk_cleanup_size = 0")
+      described_class.update_disk_cleanup_size(128)
+    end
+    expect(described_class.instance_variable_get("@disk_cleanup_size")).to eq(128)
```
MikeMcQuaid: I'm not sure it's worth verifying internal class variables are set; test the externally visible functionality instead (which may mean not having tests for every method)

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -69,6 +69,28 @@
...
+      described_class.instance_eval("@disk_cleanup_size = 0")
+    end
+    expect(described_class.disk_cleanup_size).to eq(described_class.instance_variable_get("@disk_cleanup_size"))
+  end
+
+  specify "::unremovable_kegs" do
+    shutup do
+      described_class.unremovable_kegs
+    end
+    expect(described_class.instance_variable_get("@unremovable_kegs")).to eq([])
```
MikeMcQuaid: I'm not sure it's worth verifying internal class variables are set; test the externally visible functionality instead (which may mean not having tests for every method)

MikeMcQuaid:
@mansimarkaur https://codecov.io/gh/Homebrew/brew gives you a nice overview of coverage and where new tests are needed.

MikeMcQuaid:
@mansimarkaur What are your current thoughts here?

MikeMcQuaid:
@mansimarkaur Gentle ping on this again, thanks.

mansimarkaur:
@MikeMcQuaid I'm working on this.

MikeMcQuaid:
@mansimarkaur Any news here?

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -181,5 +181,21 @@
 
       expect(npm_cache).not_to exist
     end
+
+    it "cleans up files and directories with name containing -- if ARGV prune is all" do
```
mistydemeo: The `--prune=all` option is intended to be date-based - it doesn't just arbitrarily clean up files containing `--`. Since these files were made just now with current timestamps, the default time values wouldn't catch them, but `--all` causes the directory and file to get picked up. We probably want to update this to make it clear that we're cleaning up every eligible file in this case, and also add other time-based prune tests.

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -181,5 +181,21 @@
...
+      a = (HOMEBREW_CACHE/"--a")
```
mistydemeo: The filename is meant to catch git (and other VCS) checkouts; for example, `~/Library/Caches/Homebrew/gist--git`. It'd be good to make the directory name more similar to the pattern we're trying to check for.

MikeMcQuaid:
@mansimarkaur Take a look at this PR's comments next time you're blocked on https://github.com/Homebrew/brew/pull/2804, thanks!

mansimarkaur:
@mistydemeo  How should I try to dodge the https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cleanup.rb#L93 (TapFormulaWithOldnameAmbiguityError) ?

mansimarkaur:
I see a 94.9% coverage for `cleanup.rb` on my system while codecov shows a mere 20%.
Could you please take a look what's going on? @mistydemeo

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -63,6 +63,11 @@
     end
   end
 
+  specify "::disk_cleanup_size" do
+    described_class.instance_eval("@disk_cleanup_size = 0")
```
MikeMcQuaid: Can we do this without setting an instance variable?

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -141,5 +146,87 @@
...
+        described_class.cleanup_cache
+        expect(foo).not_to exist
+      ensure
+        FileUtils.rm_rf(foo)
+      end
+    end
+
+    context "cleans old files in HOMEBREW_CACHE" do
+      before(:each) do
+        @bottle = (HOMEBREW_CACHE/"testball-0.0.1.bottle.tar.gz")
```
MikeMcQuaid: Favour `let` over instance variables

Library/Homebrew/test/cleanup_spec.rb
mansimarkaur: We can test this by writing an integration test in `test/cmd/cleanup_spec.rb` for `report_disk_usage`.

mansimarkaur:
@mistydemeo Could you please review and see if this is ready to be merged or not. Thanks!

mistydemeo:
While I think it might be helpful to see some tests to see that files get preserved, not just cleaned up, what we have here looks good to me.

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -63,6 +63,11 @@
     end
   end
 
+  specify "::disk_cleanup_size" do
+    disk_cleanup_size = described_class.instance_variable_get(:@disk_cleanup_size)
+    expect(described_class.disk_cleanup_size).to eq(disk_cleanup_size)
+  end
```
MikeMcQuaid: Instead of this test could just use `attr_reader :disk_cleanup_size` and replace the `def disk_cleanup_size` method.

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -141,5 +146,88 @@
...
+      git.mkpath
+      gist.mkpath
+      FileUtils.touch svn
+      allow(ARGV).to receive(:value).with("prune").and_return("all")
+      begin
+        described_class.cleanup_cache
+        expect(git).not_to exist
+        expect(gist).to exist
+        expect(svn).not_to exist
+      ensure
```
MikeMcQuaid: Use `after` instead of `ensure`.

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -141,5 +146,88 @@
...
+
+    it "cleans up VCS checkout directories with modified time < prune time" do
+      foo = (HOMEBREW_CACHE/"--foo")
+      foo.mkpath
+      allow(ARGV).to receive(:value).with("prune").and_return("1")
+      allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
+      begin
+        described_class.cleanup_cache
+        expect(foo).not_to exist
+      ensure
```
MikeMcQuaid: Use `after` instead of `ensure`.

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -141,5 +146,88 @@
...
+    end
+  end
+
+  specify "::prune?" do
+    foo = mktmpdir/"foo.rb"
+    foo.mkpath
+    allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
+    begin
+      expect(described_class.prune?(foo, days_default: "1")).to be_truthy
+    ensure
```
MikeMcQuaid: Use `after` instead of `ensure`.

Library/Homebrew/test/cleanup_spec.rb
reitermarkus: Generally, you should never (have to) use `instance_variable_get` in tests.

Library/Homebrew/test/cleanup_spec.rb
reitermarkus: `HOMEBREW_CACHE` is cleaned up after each spec already.

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -141,5 +166,106 @@
...
+    end
+
+    let(:foo) { mktmpdir/"foo.rb" }
+
+    it "returns true when path_modified_time < days_default" do
+      allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
+      expect(described_class.prune?(foo, days_default: "1")).to be_truthy
+    end
+
+    it "returns true when path_modified_time >= days_default" do
```
mistydemeo: Is this test name incorrect?

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -104,14 +104,30 @@
     expect(f4).to be_installed
   end
 
-  specify "::cleanup_logs" do
-    path = (HOMEBREW_LOGS/"delete_me")
-    path.mkpath
-    ARGV << "--prune=all"
+  describe "::cleanup_logs" do
+    before do
```
mistydemeo: How about putting the `let` that defines this variable before the `before`?

mansimarkaur: Yeah. That would be better. :)

Library/Homebrew/test/cleanup_spec.rb
mansimarkaur: Yes, that was a mistake. Thanks!

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -141,5 +166,106 @@
...
+        FileUtils.touch(testball)
+        (HOMEBREW_CELLAR/"testball"/"0.0.1").mkpath
+        FileUtils.touch(CoreTap.instance.formula_dir/"testball.rb")
+      end
+
+      after(:each) do
+        FileUtils.rm_rf(bottle)
+        FileUtils.rm_rf(testball)
+        FileUtils.rm_rf(HOMEBREW_CELLAR/"testball")
+        FileUtils.rm_rf(CoreTap.instance.formula_dir/"testball.rb")
```
MikeMcQuaid: Are you sure all these are needed and aren't deleted by default for all tests? Thanks!

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -104,14 +104,30 @@
...
 
-    expect(path).not_to exist
+    it "cleans up logs if older than 14 days" do
+      allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 15)
+      described_class.cleanup_logs
+      expect(path).not_to exist
+    end
+
+    it "does not clean up logs less than 14 days old" do
+      allow_any_instance_of(Pathname).to receive(:mtime).and_return(Time.now - 60 * 60 * 24 * 2)
```
MikeMcQuaid: May be worth having a `let` for the `Time.now - 60 * 60 * 24 * 2` as it's repeated three times. Also may as well do it for the `Time.now - 60 * 60 * 24 * 15` just to be consistent.

Library/Homebrew/test/cleanup_spec.rb
```diff
@@ -141,5 +166,106 @@
...
+      end
+    end
+  end
+
+  describe "::prune?" do
+    before do
+      foo.mkpath
+    end
+
+    let(:foo) { mktmpdir/"foo.rb" }
```
MikeMcQuaid: Why does this need to be in a `mktmpdir`? Can it live in some tap or other already created temporary directory instead?

Library/Homebrew/test/cleanup_spec.rb
MikeMcQuaid: For example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/cleanup_spec.rb
MikeMcQuaid: For example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/cleanup_spec.rb
mansimarkaur: I'm doing a `let` for `Time.now - 60 * 60 * 24 ` and call it num_of_days.

Library/Homebrew/test/cleanup_spec.rb
mansimarkaur: These are deleted automatically at the end of each test. Removing the `rm_rf` for these. Thanks!

Library/Homebrew/test/cleanup_spec.rb
mansimarkaur: I've created `foo.rb` in `HOMEBREW_CACHE` now.

mistydemeo:
Thanks @mansimarkaur!

MikeMcQuaid:
Nice work here @mansimarkaur!

PR #2708, opened 2017-06-01
.codecov.yml: move root so files are tracked

JCount:

-----
Currently, all files appear to be untracked on Codecov. This attempts to rectify this by correctly pointing to the root location of the `.codecov.yml` file and while also adjusting the path to be one directory up, in `Library/Homebrew/`.


Library/Homebrew/test/.codecov.yml
```diff
@@ -1,7 +0,0 @@
-comment: false
-coverage:
-  fixes:
-    - "::Library/Homebrew/"
-  parsers:
-    v1:
-      include_full_missed_files: yes
```
MikeMcQuaid: Do we not need/want this any more?

JCount: That appears to only apply to [v1.0 of Codecov](https://docs.codecov.io/v1.0/docs/language-configuration), and the current version is 4.3.6.

reitermarkus:
Do not merge because this doesn't seem to change anything.

Library/Homebrew/test/.codecov.yml
reitermarkus: It says in the docs that this is deprecated. Not sure if this was fixed in some way.

JCount:
I wan't going to, I was just rebasing to remove the first commit.

reitermarkus:
Merging this now, maybe it will work once it is in `master`. Otherwise I'll contact support again.

PR #2712, opened 2017-06-01
Fix curl invocation in Hbc::DSL::Appcast

### LINKED ISSUE #2592: [cask] Hbc::DSL::Appcast doesn't check HTTP status codes
[This curl invocation](https://github.com/Homebrew/brew/blob/93d4fdf938779a5cccb4aabed7d82cb533f9506e/Library/Homebrew/cask/lib/hbc/dsl/appcast.rb#L15) can silently fail if the final response is not a successful status code.

Discovered this by recently [trying to create a cask](https://github.com/caskroom/homebrew-cask/pull/33441#issuecomment-299487969) and using appcast.

From that cask example, you can see that the appcast URL (`http://changelog.sodaplayer.com`) inadvertently returns a 404 response when improperly requested (it requires a `Accept: text/html` HTTP header. This is actually a web server idiosyncrasy specific to that app), but `curl` does not exit with an error, it exits with success (`curl` will only exit with an error for non-successful HTTP statuses if using the `--fail`/`-f` flag).

Not only the HTTP status is ignored, but also appcast checkpoints are wrongly generated based on, possibly, error messages in the response body, or even empty responses for that matter. Once this happens (request always returns non-200 responses with the same error message), all checkpoints will be the same for every check.

On the other hand, `Hbc::Audit`, [implements this check correctly](https://github.com/Homebrew/brew/blob/93d4fdf938779a5cccb4aabed7d82cb533f9506e/Library/Homebrew/cask/lib/hbc/audit.rb#L123-L132).

13k:

Explained in #2592 

`cask/lib/hbc/dsl/appcast.rb` didn't have any specs, so I created.

`appcast_spec.rb` passes, but I got errors from gpg specs:

```
rspec ./test/gpg_spec.rb:7 # Gpg::create_test_key creates a test key in the home directory
```

-----

Fixes #2592

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/dsl/appcast.rb
+++ b/Library/Homebrew/cask/lib/hbc/dsl/appcast.rb
@@ -12,7 +12,7 @@ module Hbc
       end
 
       def calculate_checkpoint
-        result = SystemCommand.run("/usr/bin/curl", args: ["--compressed", "--location", "--user-agent", URL::FAKE_USER_AGENT, @uri], print_stderr: false)
+        result = SystemCommand.run("/usr/bin/curl", args: ["--compressed", "--location", "--user-agent", URL::FAKE_USER_AGENT, "--fail", @uri], print_stderr: false)
 
         checkpoint = if result.success?
           processed_appcast_text = result.stdout.gsub(%r{<pubDate>[^<]*</pubDate>}m, "")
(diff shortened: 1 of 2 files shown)


reitermarkus:
Thanks, especially for the very thorough test! 👍

reitermarkus:
> I got errors from gpg specs:

Is this still happening?

PR #2718, opened 2017-06-02
Fix `rubocop` in `brew audit`.

reitermarkus:

-----

cc @ilovezfs

### DIFF
--- a/Library/Homebrew/cmd/style.rb
+++ b/Library/Homebrew/cmd/style.rb
@@ -120,7 +120,7 @@ module Homebrew
       # exitstatus can also be nil if RuboCop process crashes, e.g. due to
       # native extension problems.
       # JSON needs to be at least 2 characters.
-      if !status.success? || json.to_s.length < 2
+      if !(0..1).cover?(status.exitstatus) || json.to_s.length < 2
         raise "Error running `rubocop --format json #{args.join " "}`"
       end
       RubocopResults.new(JSON.parse(json))
--- a/Library/Homebrew/test/cmd/style_spec.rb
+++ b/Library/Homebrew/test/cmd/style_spec.rb
@@ -1,0 +1,34 @@
+require "cmd/style"
+
+describe "brew style" do
+  around(:each) do |example|
+    begin
+      FileUtils.ln_s HOMEBREW_LIBRARY_PATH, HOMEBREW_LIBRARY/"Homebrew"
+      FileUtils.ln_s HOMEBREW_LIBRARY_PATH.parent/".rubocop.yml", HOMEBREW_LIBRARY/".rubocop.yml"
+
+      example.run
+    ensure
+      FileUtils.rm_f HOMEBREW_LIBRARY/"Homebrew"
+      FileUtils.rm_f HOMEBREW_LIBRARY/".rubocop.yml"
+    end
+  end
+
+  describe "Homebrew::check_style_json" do
+    let(:dir) { mktmpdir }
+
+    it "returns RubocopResults when RuboCop reports offenses" do
+      formula = dir/"my-formula.rb"
+
+      formula.write <<-'EOS'.undent
+        class MyFormula < Formula
+
+        end
+      EOS
+
+      rubocop_result = Homebrew.check_style_json([formula])
+
+      expect(rubocop_result.file_offenses(formula.realpath.to_s).map(&:message))
+        .to include("Extra empty line detected at class body beginning.")
+    end
+  end
+end


ilovezfs:
Any idea why this wasn't caught by `brew tests` originally?

Library/Homebrew/cmd/style.rb
```diff
@@ -120,7 +120,7 @@ def check_style_impl(files, output_type, options = {})
       # exitstatus can also be nil if RuboCop process crashes, e.g. due to
       # native extension problems.
       # JSON needs to be at least 2 characters.
-      if !status.success? || json.to_s.length < 2
+      if status.exitstatus > 1|| json.to_s.length < 2
```
MikeMcQuaid: Can you explain the difference here?

MikeMcQuaid: @reitermarkus As-is it looks like a non-zero exit code is being considered a success which seems like a pretty big Ruby bug.

JCount: Rubocop exits with a 1 status code if offenses are found and a 0 exit code if there are no offenses. A convenient way of handling this in compact code might be to use:
```
if !status.exitstatus === 1..0 || json.to_s.length < 2
```

ilovezfs: It's worth noting what the old code was I think: https://github.com/Homebrew/brew/pull/2690/files

Why is `if $?.exitstatus.nil? || $?.exitstatus > 1 || json.to_s.length < 2` no loner appropriate?

MikeMcQuaid: Ah, I see, thanks @reitermarkus. Agreed with @ilovezfs it may be worth adding the nil check.

Library/Homebrew/cmd/style.rb
```diff
@@ -101,7 +101,7 @@ def check_style_impl(files, output_type, options = {})
       args << "--config" << HOMEBREW_LIBRARY_PATH/".rubocop.yml"
       args << HOMEBREW_LIBRARY_PATH
     else
-      args << "--config" << HOMEBREW_LIBRARY/".rubocop.yml"
+      args << "--config" << HOMEBREW_LIBRARY_PATH.parent/".rubocop.yml"
```
MikeMcQuaid: Why this change?

Library/Homebrew/cmd/style.rb
```diff
@@ -120,7 +120,7 @@ def check_style_impl(files, output_type, options = {})
       # exitstatus can also be nil if RuboCop process crashes, e.g. due to
       # native extension problems.
       # JSON needs to be at least 2 characters.
-      if !status.success? || json.to_s.length < 2
+      if !(0..1).cover?(status.exitstatus) || json.to_s.length < 2
```
MikeMcQuaid: Does this handle the `nil` case?

reitermarkus: Yep.

Library/.rubocop.yml
```diff
@@ -1,4 +1,5 @@
 AllCops:
+  AllowSymlinksInCacheRootDirectory: true
```
MikeMcQuaid: What does this do?

Library/Homebrew/cmd/style.rb
reitermarkus: `HOMEBREW_LIBRARY` is set to a temporary directory in tests.

Library/.rubocop.yml
reitermarkus: Weird, now it is working even without it. I had to set it because RuboCop complained about `/tmp` being a symlink since the cache directory while running tests is inside `/tmp`.

Library/Homebrew/cmd/style.rb
MikeMcQuaid: Feels a little weird to rely on this without at least a comment?

reitermarkus: We are already relying on this in the same way in the `.rubocop.yml` inside `HOMEBREW_LIBRARY_PATH`:

```yaml
inherit_from:
  - ../.rubocop.yml
```

MikeMcQuaid: Sure, I just mean a comment explaining why we're doing this rather than `HOMEBREW_LIBRARY` would be good.

reitermarkus: Changed the test instead.

ilovezfs:
@reitermarkus thanks for adding a test!

JCount:
@reitermarkus Nice fix!

PR #2727, opened 2017-06-02
Refactor `FormulaVersions`.

reitermarkus:

-----

Do not load the same Formula twice.

https://github.com/Homebrew/brew/pull/2673

### DIFF
--- a/Library/Homebrew/formula_versions.rb
+++ b/Library/Homebrew/formula_versions.rb
@@ -17,6 +17,7 @@ class FormulaVersions
     @repository = formula.tap.path
     @entry_name = @path.relative_path_from(repository).to_s
     @current_formula = formula
+    @formula_at_revision = {}
   end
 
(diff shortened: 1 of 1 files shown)


Library/Homebrew/formula_versions.rb
```diff
@@ -32,11 +33,12 @@ def file_contents_at_revision(rev)
   end
 
   def formula_at_revision(rev)
+    yield @formula_at_revision[rev] if @formula_at_revision[rev]
     contents = file_contents_at_revision(rev)
```
MikeMcQuaid: It's early so I'm not sure I quite understand how/why an extra `yield` is in here and it's not passing the contents. Can you elaborate?

reitermarkus: You're right, it should `yield` and then `return`.

PR #2734, opened 2017-06-06
cmd/reinstall: Fix reinstallation of formulae without bottles

rwhogg:
Reinstallation of non-bottled formulae was failing because it
used a deprecated method.

Signed-off-by: Bob W. Hogg <rwhogg@linux.com>

-----

Running `brew reinstall` on a formula which
a) was not already installed AND
b) does not have a bottle
would fail because it used a deprecated method.

Note that this issue is reproducible on macOS, but it's way easier to reproduce on Linux because so few formulae lack Mac bottles.


MikeMcQuaid:
Nice spot, thanks @rwhogg!

PR #2744, opened 2017-06-07
High Sierra: more initial support

DomT4:
- [x] Have you successfully run `brew tests` with your changes locally? (On macOS 10.12, not 10.13)

-----

One small note:

I trimmed `Libs.private` in `libcurl.pc` because I'm pretty sure it's wildly outdated & has been for quite some time. In the 10.12 `libcurl.pc` for example we have things like `-lssl -lcrypto` which is almost certainly not true these days since `curl` links to macOS' `SecureTransport`.

Reading the `libcurl.pc` file from Homebrew's standard, non-option, bottle-poured `curl` I get `Libs.private: -lldap -lz` so that's what I've trimmed the line to 🤷‍♂️.

### DIFF
--- a/Library/Homebrew/os/mac/pkgconfig/10.13/libcurl.pc
+++ b/Library/Homebrew/os/mac/pkgconfig/10.13/libcurl.pc
@@ -1,0 +1,39 @@
+#***************************************************************************
+#                                  _   _ ____  _
+#  Project                     ___| | | |  _ \| |
+#                             / __| | | | |_) | |
+#                            | (__| |_| |  _ <| |___
+#                             \___|\___/|_| \_\_____|
+#
+# Copyright (C) 2004 - 2012, Daniel Stenberg, <daniel@haxx.se>, et al.
+#
+# This software is licensed as described in the file COPYING, which
+# you should have received as part of this distribution. The terms
+# are also available at http://curl.haxx.se/docs/copyright.html.
+#
+# You may opt to use, copy, modify, merge, publish, distribute and/or sell
+# copies of the Software, and permit persons to whom the Software is
+# furnished to do so, under the terms of the COPYING file.
+#
+# This software is distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY
+# KIND, either express or implied.
+#
+###########################################################################
+
+# This should most probably benefit from getting a "Requires:" field added
+# dynamically by configure.
+#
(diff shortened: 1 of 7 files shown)


MikeMcQuaid:
Thanks again @DomT4!

MikeMcQuaid:
A random aside: have you upgraded a VM or your actual machine and if the latter: do you consider it usable for day-to-day use?

DomT4:
I have two High Sierra installs running locally currently 😅. I sliced a partition out of my SSD to test APFS a little bit, because when I tried to upgrade to High Sierra inside a VMware box with APFS transition enabled it borked the whole thing & refused to boot, which [they're aware of](https://twitter.com/patlee/status/872244863214927873).

I use the VMware box for testing primarily because it means I'm not having to reboot my whole system to check anything. In terms of stability, even running natively on the SSD rather than trapped inside VMware it's not _super_ stable yet, as expected really I guess.

I ended up swapping out Safari for Chrome because Safari kept turning websites completely black, persisting through page refreshes, as an example. There's nothing I've encountered yet that will absolutely ruin your day, but a fair bit that might make you sigh & go hunting for a workaround.

I'm planning to not upgrade my main partition for a little while though; I'm kinda concerned how third-party apps are going to handle APFS.

monouser7dig:
I had HighSierra installed on an external SSD (upgraded from Sierra) and during about half a day not crashed anything, just worked, but I had not enabled APFS, I was not prompted during install or anything, how did you change to APFS in your VM?

thanks

DomT4:
> I was not prompted during install or anything, how did you change to APFS in your VM?

There's a little checkbox & brief explanation "page" in the `Install macOS 10.13 Beta.app` process that asks whether you want to convert your filesystem. The release notes note that APFS has some "unsupported configurations" though so perhaps you hit one of the exclusions.

I think I can talk about those here without angering the Apple gods given everyone else in the world is:

* HDD-only Macs currently are not converted to APFS by default.
* Some iMacs with 3TB Fusion drives and BootCamp may be unsupported for use with
APFS.
* APFS support for Mid 2012 Mac Pro will be included in an upcoming seed.

monouser7dig:
Hm not sure any of those exceptions apply but will see how the next versions behave 

Thanks

MikeMcQuaid:
Thanks @DomT4!

PR #2753, opened 2017-06-08
Adding shims for GCC 7

leedm777:

-----


MikeMcQuaid:
@leedm777 Out of interest: what's the use-case for this? I know it's the same as the others but I'm just interested how you personally have run into it.

leedm777:
Sorry, should have put more details in the request.

I have a tap that only builds with GCC. Without these symlinks, on gcc-7 it fails because it's not running in the superenv, and can't find the keg-only dependencies.

MikeMcQuaid:
Thanks for your first contribution to Homebrew, @leedm777! You rock; we couldn't run this project without contributions from people like you.

PR #2755, opened 2017-06-08
audit: Port audit_checksum method to rubocop and add tests

GauthamGoli:

-----
#569


Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,54 @@
...
+            audit_checksums(spec_body, name)
+            resource_blocks = find_all_blocks(spec_body, :resource)
+            resource_blocks.each do |rb|
+              _, _, resource_body = *rb
+              audit_checksums(resource_body, name, string_content(parameters(rb).first))
+            end
+          end
+        end
+
+        def audit_checksums(node, spec, resource = nil)
```
MikeMcQuaid: Maybe `resource_name`?

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,54 @@
...
+
+          if string_content(checksum).size != 64 && regex_match_group(checksum, /^\w*$/)
+            problem "#{msg_prefix}sha256 should be 64 characters"
+          end
+
+          unless regex_match_group(checksum, /^[a-fA-F0-9]+$/)
+            problem "#{msg_prefix}sha256 contains invalid characters"
+          end
+
+          problem "#{msg_prefix}sha256 should be lowercase" unless regex_match_group(checksum, /^[a-f0-9]+$/)
```
MikeMcQuaid: If this line is >80 characters: use `unless ... do`

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,54 @@
...
+          if string_content(checksum).size.zero?
+            problem "#{msg_prefix}sha256 is empty"
+            return
+          end
+
+          if string_content(checksum).size != 64 && regex_match_group(checksum, /^\w*$/)
+            problem "#{msg_prefix}sha256 should be 64 characters"
+          end
+
+          unless regex_match_group(checksum, /^[a-fA-F0-9]+$/)
```
MikeMcQuaid: Maybe make it a case-sensitive regex (`/i`) and skip`A-F`?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -250,7 +256,7 @@ def caveats_strings
 
       # Returns the array of arguments of the method_node
       def parameters(method_node)
-        return unless method_node.send_type?
+        return unless method_node.send_type? || method_node.block_type?
```
MikeMcQuaid: Turn this into an `if` (or two); `unless .. ||` is hard to parse.

Library/Homebrew/rubocops/checksum_cop.rb
GauthamGoli: Can I make it `return unless` and followed by `problem`?

MikeMcQuaid: Sure!

Library/Homebrew/rubocops/checksum_cop.rb
JCount: Just to clarify for posterity, you do mean `case-insensitive` regex using `/i`, correct?

MikeMcQuaid: Sorry: yes, thanks @JCount

GauthamGoli:
@MikeMcQuaid a side point: if `sha256` gets deprecated / new hashes in future, then this would require refactoring. Should I refactor line 34 - 51 to not hardcode `sha256`?

MikeMcQuaid:
> @MikeMcQuaid a side point: if sha256 gets deprecated / new hashes in future, then this would require refactoring. Should I refactor line 34 - 51 to not hardcode sha256?
> Edit: May not be required as new hash would mean, just adding new conditions below line 51.

Nah, I think it's fine as-is; we'll be on sha256 for quite a while longer.

MikeMcQuaid:
@GauthamGoli Great work again, thanks!

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,55 @@
+require_relative "./extend/formula_cop"
+
+module RuboCop
+  module Cop
+    module FormulaAudit
+      class Checksum < FormulaCop
+        def audit_formula(_node, _class_node, _parent_class_node, body_node)
+          %w[Stable Devel HEAD].each do |name|
```
MikeMcQuaid: Let's just make these lowercase, I don't think `Devel` looks very nice.

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,55 @@
...
+            next unless spec_node = find_block(body_node, name.downcase.to_sym)
+            _, _, spec_body = *spec_node
+            audit_checksums(spec_body, name)
+            resource_blocks = find_all_blocks(spec_body, :resource)
```
MikeMcQuaid: We also need to get all resource blocks that aren't inside a stable/devel/head.

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,55 @@
...
+          end
+          if find_node_method_by_name(node, :md5)
+            problem "#{msg_prefix}MD5 checksums are deprecated, please use SHA256"
+          end
+
+          if find_node_method_by_name(node, :sha1)
+            problem "#{msg_prefix}SHA1 checksums are deprecated, please use SHA256"
+          end
+
+          checksum_node = find_node_method_by_name(node, :sha256)
```
MikeMcQuaid: Need to handle if this is `nil`; I hit this locally. Also, this doesn't seem to be constrained to the current `node` as far as I can tell.

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,55 @@
...
+
+          if string_content(checksum).size != 64 && regex_match_group(checksum, /^\w*$/)
+            problem "#{msg_prefix}sha256 should be 64 characters"
+          end
+
+          unless regex_match_group(checksum, /^[a-f0-9]+$/i)
+            problem "#{msg_prefix}sha256 contains invalid characters"
+          end
+
+          return unless regex_match_group(checksum, /^[a-f0-9]+$/)
```
MikeMcQuaid: `return unless regex_match_group(checksum, /[A-F]/)` seems correct for here.

Library/Homebrew/rubocops/checksum_cop.rb
GauthamGoli: This is not currently being done in `audit_specs`, right?

Library/Homebrew/rubocops/checksum_cop.rb
GauthamGoli: Should an error be raised, if there is no `sha256` specified?

Library/Homebrew/rubocops/checksum_cop.rb
GauthamGoli: Ah. This is embarrassing :/

MikeMcQuaid: We all make mistakes; it's embarrassing that I didn't spot it in review!

Library/Homebrew/rubocops/checksum_cop.rb
MikeMcQuaid: A RuboCop error: yes, I think so. I was seeing this on formulae that had them though so it may be a bug, too.

Library/Homebrew/rubocops/checksum_cop.rb
JCount: I believe that the resource blocks are currently being audited as part of running `audit_specs`, but the code structure relies on the ResourceAuditor class, and the `audit_checksum` it contains. I think that gap might be part of the cause of this issue.

GauthamGoli: I couldn't find any existing method/code in `audit.rb` which audits `resource` blocks that aren't inside stable/devel/head blocks, is this something we want to add?

Library/Homebrew/rubocops/checksum_cop.rb
GauthamGoli: Do you mean, `find_node_method_by_name` returns `nil` even when there is `sha256` in node?

Library/Homebrew/rubocops/checksum_cop.rb
MikeMcQuaid: Auditing the stable spec audits everything which is outside those blocks (as they are implicitly stable). The existing code is therefore fine, it's just this code that needs adjusted for that implicit stable resource handling. Make sense?

Library/Homebrew/rubocops/checksum_cop.rb
MikeMcQuaid: I can't remember now helpfully but I think it's just that again we can have a stable/devel/head block that doesn't have a URL therefore doesn't need a checksum (although head urls don't need checksums either).

Library/Homebrew/rubocops/checksum_cop.rb
GauthamGoli: Yeah. Got it. Thanks.

PR #2763, opened 2017-06-09
Use Ruby 2.3’s `URI`.

reitermarkus:

-----

RFC 2396 is almost 20 years old and was replaced by RFC 3986 in 2005.


Library/Homebrew/global.rb
```diff
@@ -13,6 +13,13 @@
 require "official_taps"
 require "pp"
 
+ruby_version_major, ruby_version_minor, = RUBY_VERSION.split(".").map(&:to_i)
+if ruby_version_major == 2 && ruby_version_minor < 3
+  # Use Ruby 2.3's URI class to support RFC 3986.
+  $LOAD_PATH.unshift("#{ENV["HOMEBREW_LIBRARY"]}/Homebrew/vendor/ruby/2.3.0")
```
mistydemeo: Rather than do this here in `global`, how about making the vendored `uri` define its features conditionally based on the `RUBY_VERSION`?

woodruffw:
Sorry if this is a stupid question: What are we getting from vendoring 2.3's `URI`, as opposed to just using the `URI` available in the system Ruby (2.0.0)? From a cursory glance, it looks like it hasn't changed much.

MikeMcQuaid:
> (Piggybacking off of what Mike said, sorry if this is a stupid question).
> 
> Why are we vendoring 2.3's URI, as opposed to just using the URI available in the system Ruby (2.0.0)? From a cursory glance, it looks like it hasn't changed much besides adding RFC 3986.

Not a stupid question, I'm interested in it too. Given we were generally :-1: on vendoring code we already need and use from gems I'm not sure I can see a justification for this. If we really want to start using Ruby 2.3 features we can vendor a Ruby 2.3 and start using the system one when High Sierra is released.

reitermarkus:
You're right, it's not worth it.

This is such an edgy edge case, it might as well be a myth, and will be fixed when High Sierra ships with Ruby 2.3.

PR #2764, opened 2017-06-09
keg_relocate: Don't relocate duplicated Rust dylibs.

woodruffw:
This is a workaround for https://github.com/rust-lang/rust/issues/39870.

We could make the filename match even more restrictive, but I think that `/lib/rustlib/` will probably be exclusive enough.

cc @ilovezfs 

-----

### DIFF
--- a/Library/Homebrew/extend/os/mac/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/mac/keg_relocate.rb
@@ -84,7 +84,16 @@ class Keg
     dylibs.each(&block)
   end
 
+  # Rust ships two copies of its libraries, one in lib/ and the other in lib/rustlib/.
+  # We need to avoid rewriting the second copy until
+  # https://github.com/rust-lang/rust/issues/39870 is fixed.
+  def rust_duplicate?(fn)
+    fn =~ %r{/lib/rustlib/}
+  end
+
   def dylib_id_for(file)
+    return file.dylib_id if rust_duplicate?(file.to_s)
+
     # The new dylib ID should have the same basename as the old dylib ID, not
     # the basename of the file itself.
     basename = File.basename(file.dylib_id)


MikeMcQuaid:
To add to my previous comment: what are the options for making a change in Homebrew/homebrew-core for this?

woodruffw:
> How does this break things for Homebrew users today?

My understanding is that the `rust` package is currently half-working, with certain things causing a crash along the lines of https://github.com/rust-lang/rust/issues/39870 due to multiple copies of the same library being loaded and clobbering each other's state.

> Has work started upstream on fixing their issue?

They've gotten an issue open at https://github.com/rust-lang/rust/issues/39875, but I don't think any work has been started.

> what are the options for making a change in Homebrew/homebrew-core for this?

I think the simplest equivalent solution would be to glob all dylibs under `lib/rustlib` and re-rewrite them to contain `@rpath` metavariables in their `LC_DYLIB_ID`s.

I went with this because it avoids a second rewrite, but it is awfully specific to be included in core. The in-formula solution isn't that much more complex, I'll have it ready in a day or two.

ilovezfs:
I'm :-1: on basically permanently (I don't think upstream is going to "fix" something that isn't broken) carrying an in-formula "solution" for this since it's unwinding changes that brew just made.

woodruffw:
> general solution in brew

This PR is pretty specific right now, but I could definitely morph it into a more general solution for excluding files or patterns from relocation. That would avoid making a rust-specific change to `brew` *and* undoing our own relocations in postinstall.

ilovezfs:
Maybe take a look at `skip_clean` for inspiration?

woodruffw:
> Maybe take a look at skip_clean for inspiration?

Thanks. Yeah, I was thinking something along the lines of `skip_relocation "some/path/*.glob"`.

One potential problem is that we *don't* need to skip install name relocation in rust's case - only dylib ID relocation. I could split those two up, but you can see how that would get ugly/complicated fast.

ilovezfs:
>One potential problem is that we don't need to skip install name relocation in rust's case

Is it a problem though? Does rust break if we don't do the install name relocation either?

woodruffw:
> Does rust break if we don't do the install name relocation either?

Actually, I'm not sure. I wouldn't expect it to break with the unrelocated paths, we just haven't *needed* to do it. I'll test this evening.

ilovezfs:
FYI, I've shipped 1.18.0 (https://github.com/Homebrew/homebrew-core/pull/14375) in its currently "broken" state so no need to worry about that waiting on this or anything like that.

woodruffw:
I think we should go ahead and merge https://github.com/Homebrew/homebrew-core/pull/14490, since this is indeed pretty specific to one language/formula.

I'm going to test how the `rustlib/` dylibs behave with no install name relocation later today, and I'll make another PR along the lines of `skip_relocation "some/path"` if it works out.

PR #2776, opened 2017-06-12
audit: Fix audit_checksum method's rubocop and add more tests

GauthamGoli:

-----
Fixes the bugs #2755 had, as described below:
1. Auditing checksums in implicitly stable `resource` blocks called outside `stable` block.
2. Handling edge case where `sha256` may not be present in stable/devel/head blocks.
3. Bug where violation is raised for lowercase checksum rather than uppercase checksum.
Also added more tests to cover new corner cases.


Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,61 @@
+require_relative "./extend/formula_cop"
+
+module RuboCop
+  module Cop
+    module FormulaAudit
+      class Checksum < FormulaCop
+        def audit_formula(_node, _class_node, _parent_class_node, body_node)
+          %w[stable devel head].each do |name|
+            next unless spec_node = find_block(body_node, name.to_sym)
```
MikeMcQuaid: Looks like this will`next` without an explicit `stable do` block whereas we want to ensure a root `sha256` is checked in that case. I think having a separate section of this method for `stable` outside of `stable do` would make sense; you can then put `find_blocks(body_node, :resource)` in there too to avoid the `if` below.

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,61 @@
...
+            _, _, spec_body = *spec_node
+            audit_checksums(spec_body, name)
+            if name == "stable"
+              resource_blocks = find_blocks(body_node, :resource) +
+                                find_all_blocks(spec_body, :resource)
```
MikeMcQuaid: @I think you also want to always add `body_node` as it's effectively a `resource` block too. Similarly with root level `patch` blocks.

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,61 @@
+require_relative "./extend/formula_cop"
+
+module RuboCop
+  module Cop
+    module FormulaAudit
+      class Checksum < FormulaCop
+        def audit_formula(_node, _class_node, _parent_class_node, body_node)
+          %w[stable devel head].each do |name|
```
MikeMcQuaid: You also want to look at `bottle` and `patch` blocks.

Library/Homebrew/rubocops/checksum_cop.rb
JCount: I think, as MikeMcQuaid said, the `body_node` should always be added. This is because blocks that are outside an explicit spec are by default a part of all specs, and therefore need to be evaluated as such.

GauthamGoli:
I have an idea. Since rubocop reports line number and also has highlighting, we could just do `find_every_method_call_by_name(body_node, :sha256)` and then just audit every checksum.

MikeMcQuaid:
> I have an idea. Since rubocop reports line number and also has highlighting, we could just do find_every_method_call_by_name(body_node, :sha256) and then just audit every checksum.

That seems sensible. In theory there could be an edge case with some other method called `sha256` in the formula but I think that's unlikely enough to be worth ignoring.

GauthamGoli:
@MikeMcQuaid You were right about the edge case, it did occur. That case is handled now, I believe.

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,54 @@
...
+            sha256_node = get_checksum_node(sha256_call)
+            audit_sha256(sha256_node)
+          end
+        end
+
+        def get_checksum_node(call)
+          return if parameters(call).empty? || parameters(call).nil?
+          if parameters(call).first.str_type?
+            parameters(call).first
+          elsif parameters(call).first.hash_type?
```
MikeMcQuaid: Might want to document with a comment that this is the case for `bottle do` blocks.

Library/Homebrew/rubocops/checksum_cop.rb
```diff
@@ -0,0 +1,54 @@
...
+          if string_content(checksum).size != 64 && regex_match_group(checksum, /^\w*$/)
+            problem "sha256 should be 64 characters"
+          end
+
+          if regex_match_group(checksum, /[^a-f0-9]+/i)
+            problem "sha256 contains invalid characters"
+          end
+
+          return unless regex_match_group(checksum, /[A-F]+/)
+          problem "sha256 should be lowercase"
```
MikeMcQuaid: This is maybe worth having an autocorrect for?

GauthamGoli: Yeah. Will add it.

MikeMcQuaid:
Tested locally and looks good. Nice work @GauthamGoli!

JCount:
@GauthamGoli 🎉

PR #2777, opened 2017-06-12
Added tests for language/node.rb

mansimarkaur:

-----
@mistydemeo

### DIFF
--- a/Library/Homebrew/test/language/node_spec.rb
+++ b/Library/Homebrew/test/language/node_spec.rb
@@ -31,7 +31,7 @@ describe Language::Node do
 
     it "does not raise error with a zero exitstatus" do
       allow(Utils).to receive(:popen_read).with("npm pack").and_return("pack")
-      allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
+      allow($?).to receive(:exitstatus).and_return(0)
       resp = subject.std_npm_install_args(npm_install_arg)
       expect(resp).to include("--prefix=#{npm_install_arg}", "#{Dir.pwd}/pack")
     end
--- a/Library/Homebrew/test/language/node_spec.rb
+++ b/Library/Homebrew/test/language/node_spec.rb
@@ -1,56 +1,36 @@
 require "language/node"
 
 describe Language::Node do
-  specify "#npm_cache_config" do
-    shutup do
-      ret_val = described_class.npm_cache_config
-      expect(ret_val).to eq("cache=#{HOMEBREW_CACHE}/npm_cache\n")
+  describe "#setup_npm_environment" do
+    it "does nothing when npmrc exists" do
+      expect(subject.setup_npm_environment).to be_nil
+    end
+
+    it "calls prepend_path when npmrc does not exist" do
+      allow(Formula).to receive(:[]).with("node").and_return(formula { url "foo-1.0" })
+      allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
+      allow(ENV).to receive(:prepend_path).with("PATH", Formula["node"].opt_libexec/"bin").and_return("prepend_path called")
+      expect(subject.setup_npm_environment).to eq("prepend_path called")
     end
   end
 
-  describe "#pack_for_installation" do
+  describe "#std_npm_install_args" do
+    npm_install_arg = "libexec"
+
     it "raises error with non zero exitstatus" do
-      shutup do
-        expect { described_class.pack_for_installation }.to raise_error
-      end
+      expect { subject.std_npm_install_args(npm_install_arg) }.to raise_error("npm failed to pack #{Dir.pwd}")
     end
 
     it "does not raise error with a zero exitstatus" do
-      shutup do
-        allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
-        expect { described_class.pack_for_installation }.not_to raise_error
-      end
-    end
-  end
-
-  describe "#setup_npm_environment" do
-    it "npmrc exists" do
-      shutup do
-        expect(described_class.setup_npm_environment).to be_nil
-      end
-    end
-
-    it "npmrc does not exist" do
-      shutup do
-        allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
-        described_class.setup_npm_environment
-      end
-    end
-  end
-
-  specify "#std_npm_install_args" do
-    shutup do
-      npm_install_arg = "libexec"
+      allow(Utils).to receive(:popen_read).with("npm pack").and_return("pack")
       allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
-      resp = described_class.std_npm_install_args npm_install_arg
-      expect(resp).to eq(["--verbose", "--global", "--prefix=#{npm_install_arg}", "#{Dir.pwd}/"])
+      resp = subject.std_npm_install_args(npm_install_arg)
+      expect(resp).to include("--prefix=#{npm_install_arg}", "#{Dir.pwd}/pack")
     end
   end
 
   specify "#local_npm_install_args" do
-    shutup do
-      resp = described_class.local_npm_install_args
-      expect(resp).to eq(["--verbose"])
-    end
+    resp = subject.local_npm_install_args
+    expect(resp).to include("--verbose")
   end
 end
--- a/Library/Homebrew/test/language/node_spec.rb
+++ b/Library/Homebrew/test/language/node_spec.rb
@@ -7,10 +7,14 @@ describe Language::Node do
     end
 
     it "calls prepend_path when npmrc does not exist" do
-      allow(Formula).to receive(:[]).with("node").and_return(formula { url "foo-1.0" })
+      node = formula "node" do
+        url "node-test"
+      end
+      stub_formula_loader(node)
       allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
-      allow(ENV).to receive(:prepend_path).with("PATH", Formula["node"].opt_libexec/"bin").and_return("prepend_path called")
-      expect(subject.setup_npm_environment).to eq("prepend_path called")
+
+      subject.setup_npm_environment
+      expect(ENV["PATH"]).to include(Formula["node"].opt_libexec/"bin")
     end
   end
 
--- a/Library/Homebrew/test/language/node_spec.rb
+++ b/Library/Homebrew/test/language/node_spec.rb
@@ -18,7 +18,7 @@ describe Language::Node do
 
     it "does not call prepend_path when node formula does not exist but npmrc exists" do
       allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
-      expect(subject.setup_npm_environment).to eq(nil)
+      expect(subject.setup_npm_environment).to be_nil
     end
   end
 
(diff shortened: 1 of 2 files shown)


mistydemeo:
Just to check, did all of these tests output stdout/stderr? I notice they're all wrapped in `shutup do`.

.gitignore
```diff
@@ -11,6 +11,7 @@
 /Library/Homebrew/**/.bundle
 /Library/Homebrew/**/vendor/bundle
 /Library/Homebrew/doc
+/Library/Homebrew/bin
```
mistydemeo: Mind extracting 67ec3b9a0271c3ca49ea3aca78c15a9417c8f54b into a separate pull request?

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it "does not raise error with a zero exitstatus" do
+      shutup do
+        allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
+        expect { described_class.pack_for_installation }.not_to raise_error
+      end
+    end
+  end
+
+  describe "#setup_npm_environment" do
+    it "npmrc exists" do
```
mistydemeo: When writing your test names, would you able to word them to refer to what you expect them to do? For example, this test could be "#setup_npm_environment does nothing when npmrc exists".

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "language/node"
+
+describe Language::Node do
+  specify "#npm_cache_config" do
```
MikeMcQuaid: This method is called by `#setup_npm_environment` so I don't think a separate test that's longer than the actual method is warranted.

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    shutup do
```
MikeMcQuaid: `shutup` is unneeded here; please only use it when there's output that needs suppressed.

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      ret_val = described_class.npm_cache_config
+      expect(ret_val).to eq("cache=#{HOMEBREW_CACHE}/npm_cache\n")
+    end
+  end
+
+  describe "#pack_for_installation" do
```
MikeMcQuaid: This method is called by `#setup_npm_environment` so I don't think a separate test that's longer than the actual method is warranted.

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it "raises error with non zero exitstatus" do
+      shutup do
+        expect { described_class.pack_for_installation }.to raise_error
```
MikeMcQuaid: Just an aside given I don't think this test specifically is needed: you should also use `allow_any_instance_of(Process::Status)` here or it'll be a false-negative if a user has `npm` in their `PATH`.

Library/Homebrew/test/language/node_spec.rb
MikeMcQuaid: Just to be more explicit: the `it` would be `"does nothing when npmrc exists"`

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      shutup do
+        allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
+        expect { described_class.pack_for_installation }.not_to raise_error
+      end
+    end
+  end
+
+  describe "#setup_npm_environment" do
+    it "npmrc exists" do
+      shutup do
```
MikeMcQuaid: Remove `shutup` unless output is created. Also add a `allow_any_instance_of(Pathname).to receive(:exist?).and_return(true)` to not cause a false-negative if the user does have that file when running `brew tests`

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        expect(described_class.setup_npm_environment).to be_nil
+      end
+    end
+
+    it "npmrc does not exist" do
+      shutup do
+        allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
+        described_class.setup_npm_environment
```
MikeMcQuaid: I think it might be good to verify that `ENV["PATH"]` changes?

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    end
+  end
+
+  specify "#std_npm_install_args" do
+    shutup do
+      npm_install_arg = "libexec"
+      allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
+      resp = described_class.std_npm_install_args npm_install_arg
+      expect(resp).to eq(["--verbose", "--global", "--prefix=#{npm_install_arg}", "#{Dir.pwd}/"])
```
MikeMcQuaid: I think it would be good to stub `pack_for_installation` and verify not every element in this array but just that the `pack` and `npm_install_arg` are present in values.

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  specify "#local_npm_install_args" do
+    shutup do
+      resp = described_class.local_npm_install_args
+      expect(resp).to eq(["--verbose"])
```
MikeMcQuaid: I'd make this check for inclusion rather than equality.

MikeMcQuaid:
@mansimarkaur Nice work so far! I'd recommend reading some posts from my talented coworker @jasonrudolph on testing that I think you'd find helpful:
- http://jasonrudolph.com/blog/2008/07/30/testing-anti-patterns-the-ugly-mirror/
- http://jasonrudolph.com/blog/2008/07/01/testing-anti-patterns-overspecification/

Library/Homebrew/test/language/node_spec.rb
mansimarkaur: So, should I add the test for `#npm_cache_config` within the test for `#setup_npm_environment` by checking that `"cache=#{HOMEBREW_CACHE}/npm_cache\n"` is written to the .npmrc file?

Library/Homebrew/test/language/node_spec.rb
mansimarkaur: Why should we make a check for inclusion rather than equality?

Library/Homebrew/test/language/node_spec.rb
MikeMcQuaid: I don't think that behaviour needs tested at all; it's an implementation detail.

Library/Homebrew/test/language/node_spec.rb
MikeMcQuaid: It makes the test less brittle to future changes/additions.

mansimarkaur:
Thanks for reviewing my PR, @mistydemeo and @MikeMcQuaid. Also, thanks for the links you gave.
I've made the changes you requested.

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+describe Language::Node do
+  describe "#setup_npm_environment" do
+    it "does nothing when npmrc exists" do
+      expect(subject.setup_npm_environment).to be_nil
+    end
+
+    it "calls prepend_path when npmrc does not exist" do
+      allow(Formula).to receive(:[]).with("node").and_return(formula { url "foo-1.0" })
+      allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
+      allow(ENV).to receive(:prepend_path).with("PATH", Formula["node"].opt_libexec/"bin").and_return("prepend_path called")
```
MikeMcQuaid: I think it would be good to avoid stubbing this method and just verify that the `ENV["PATH"]` includes the expected `node` path. The less stubbing we can do, the better. We want to use stubs to control only for variation in the users who run the tests.

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it "does not raise error with a zero exitstatus" do
+      allow(Utils).to receive(:popen_read).with("npm pack").and_return("pack")
+      allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
+      resp = subject.std_npm_install_args(npm_install_arg)
+      expect(resp).to include("--prefix=#{npm_install_arg}", "#{Dir.pwd}/pack")
+    end
+  end
+
+  specify "#local_npm_install_args" do
+    resp = subject.local_npm_install_args
```
MikeMcQuaid: Well done, this has exposed poor error-handling in `language/node`: https://travis-ci.org/Homebrew/brew/jobs/243010332#L280

`setup_npm_environment` should have a `rescue FormulaUnavailableError` which returns `nil` if the `node` formula doesn't exist. Alternatively, if you think (which is a reasonable view) that this should throw an error if the `node` formula doesn't exist: you may need to stub a `Formulary` method to pretend this formula exists or use `formula do` to create a formula.

mansimarkaur: @MikeMcQuaid I stubbed the existence of `Formula["node"]` but now the error is: ```NoMethodError:
       undefined method `prepend_path' for #<Object:0x007fe2288cd248>```

MikeMcQuaid: `expect(ENV).to receive(:prepend_path)` or `ENV.extend(SharedEnvExtension)`. If you're blocked on stuff like that try to `grep` (or `ripgrep` with the `rg` command) through the source to see if you can find other tests that use it. e.g. in this case I saw `Library/Homebrew/test/os/mac/java_requirement_spec.rb` using this.

mansimarkaur: Thanks @MikeMcQuaid

mansimarkaur:
Please review, @mistydemeo

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+      stub_formula_loader(node)
+      allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
+      expect(ENV).to receive(:prepend_path)
+      subject.setup_npm_environment
+    end
+
+    it "does not call prepend_path when node formula does not exist but npmrc exists" do
+      allow_any_instance_of(Pathname).to receive(:exist?).and_return(false)
+      expect(subject.setup_npm_environment).to eq(nil)
```
MikeMcQuaid: `.to be_nil`

Library/Homebrew/test/language/node_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+  end
+
+  describe "#std_npm_install_args" do
+    npm_install_arg = "libexec"
+
+    it "raises error with non zero exitstatus" do
+      expect { subject.std_npm_install_args(npm_install_arg) }.to raise_error("npm failed to pack #{Dir.pwd}")
```
MikeMcQuaid: Split onto two lines.

mansimarkaur: I've made the changes, @MikeMcQuaid. Thanks for reviewing.

MikeMcQuaid:
Thanks @mansimarkaur, great work!

PR #2784, opened 2017-06-14
diagnostic: Fix suggested git remote commands

bfontaine:

-----

`git remote add origin ...` doesn’t work because there’s already a remote.


MikeMcQuaid:
Nice! Does this work both with non-existing and already existing remotes?

bfontaine:
No, but the `origin.nil?` case is already taken care of in both… cases.

MikeMcQuaid:
Nice work @bfontaine!

PR #2790, opened 2017-06-16
audit: Port audit_legacy_patches method to rubocop and add tests

GauthamGoli:

-----
#569 
Instead of normalizing patches ( as patches could be in a hash, array or a nested array ) and could be two levels deep with an `if` guard clause, I'm fetching all the string nodes inside `def patches` and checking against the url patterns to report rubocop offenses, also went to ~2014 commits to test the cop on legacy patches 🤓, but I feel porting this is an overkill as we don't have legacy patches anymore? 
`patch_problems` would be moved to `formula_cop.rb` once `audit_specs` porting is finished.

### DIFF
--- a/Library/Homebrew/rubocops/legacy_patches_cop.rb
+++ b/Library/Homebrew/rubocops/legacy_patches_cop.rb
@@ -1,0 +1,53 @@
+require_relative "./extend/formula_cop"
+require_relative "../extend/string"
+
+module RuboCop
+  module Cop
+    module FormulaAudit
+      # This cop checks for and audits legacy patches in Formulae
+      class LegacyPatches < FormulaCop
+        def audit_formula(_node, _class_node, _parent_class_node, body)
+          patches_node = find_method_def(body, :patches)
+          return if patches_node.nil?
+          legacy_patches = find_strings(patches_node)
+          problem "Use the patch DSL instead of defining a 'patches' method"
+          legacy_patches.each { |p| patch_problems(p) }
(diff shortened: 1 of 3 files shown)


Library/Homebrew/rubocops/legacy_patches_cop.rb
```diff
@@ -0,0 +1,53 @@
...
+      class LegacyPatches < FormulaCop
+        def audit_formula(_node, _class_node, _parent_class_node, body)
+          patches_node = find_method_def(body, :patches)
+          return if patches_node.nil?
+          legacy_patches = find_strings(patches_node)
+          problem "Use the patch DSL instead of defining a 'patches' method"
+          legacy_patches.each { |p| patch_problems(p) }
+        end
+
+        def patch_problems(p)
```
MikeMcQuaid: `def patch_problems(patch)`

MikeMcQuaid:
Looks good so far! Can you give an example of some formulae you tested this on? One tiny nit and then I think good to 🚢.

GauthamGoli:
@MikeMcQuaid  In `homebrew-core` I ran a command to find all commits with legacy patches. and then tested it. For example checkout `587658a94b5ee` and there is a formula `cardpeek` with legacy patches.

MikeMcQuaid:
Thanks again @GauthamGoli!

GauthamGoli:
Thanks for the review @MikeMcQuaid @JCount

PR #2793, opened 2017-06-16
Add support for `uninstall/zap :trash`.

reitermarkus:

-----

This makes `uninstall :trash` and `zap :trash` actually trash files instead of straight out deleting them.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/uninstall_base.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/uninstall_base.rb
@@ -224,9 +224,14 @@ module Hbc
       end
 
       def uninstall_trash(*paths)
-        # :trash functionality is stubbed as a synonym for :delete
-        # TODO: make :trash work differently, moving files to the Trash
-        uninstall_delete(*paths)
+        return if paths.empty?
+
+        ohai "Trashing files:"
+        each_resolved_path(:trash, paths) do |path, resolved_paths|
+          puts path
+          resolved_paths.each { |resolved_path| Utils.gain_permissions(resolved_path, ["-R"], @command) }
+          @command.run!("/usr/bin/xargs", args: ["-0", "--", HOMEBREW_LIBRARY_PATH/"utils/trash.swift"], input: resolved_paths.join("\0"))
+        end
       end
 
       def uninstall_rmdir(*directories)
--- a/Library/Homebrew/test/utils/trash_spec.rb
+++ b/Library/Homebrew/test/utils/trash_spec.rb
@@ -1,0 +1,32 @@
+require "open3"
+
+describe "trash", :needs_macos do
+  let(:executable) { HOMEBREW_LIBRARY_PATH/"utils/trash.swift" }
+  let(:dir) { mktmpdir }
+  let(:file) { dir/"new_file" }
+
+  it "moves existing files to the trash" do
+    FileUtils.touch file
+
+    expect(file).to exist
+
+    out, err, status = Open3.capture3(executable, file)
+
+    expect(out).to match %r{moved #{file} to .*/\.Trash/\.*}
+    expect(err).to be_empty
+    expect(status).to be_a_success
+
+    expect(file).not_to exist
+
+    trashed_path = out.sub(/^moved #{Regexp.escape(file.to_s)} to (.*)\n$/, '\1')
+    FileUtils.rm_f trashed_path
+  end
+
+  it "fails when files don't exist" do
+    out, err, status = Open3.capture3(executable, file)
+
+    expect(out).to be_empty
+    expect(err).to eq "could not move #{file} to trash\n"
+    expect(status).to be_a_failure
+  end
+end
--- a/Library/Homebrew/utils/trash.swift
+++ b/Library/Homebrew/utils/trash.swift
@@ -1,0 +1,43 @@
+#!/usr/bin/swift
+
+import Cocoa
+
+DispatchQueue.main.async {
+  let arguments = CommandLine.arguments.dropFirst().filter { !$0.isEmpty }
+  let URLs = arguments.map { URL(fileURLWithPath: $0) }
+
+  #if swift(>=4.0)
+  let workspace = NSWorkspace.shared
+  #else
+  let workspace = NSWorkspace.shared()
+  #endif
+
+  workspace.recycle(URLs) { (dict, error) in
+    dict.forEach {
+      #if swift(>=4.0)
+      let originalPath = $0.0.path
+      let trashPath    = $0.1.path
+      #else
+      let originalPath = $0.path
+      let trashPath    = $1.path
+      #endif
+
+      print("moved \(originalPath) to \(trashPath)")
+    }
+
+    if error == nil {
+      exit(0)
+    }
+
+    let trashedURLs = dict.keys
+    let untrashedURLs = URLs.filter { !trashedURLs.contains($0) }
+
+    untrashedURLs.forEach {
+      fputs("could not move \($0.path) to trash\n", stderr)
+    }
+
+    exit(1)
+  }
+}
+
+RunLoop.main.run()


chdiza:
Out of curiosity: why is the trash util in swift?  Can moving things to the trash not be done in ruby?

reitermarkus:
The system API for this is only available in Objective-C and Swift, as far as I know.

chdiza:
But I mean, isn't it just moving files to the trash?  Why is some system API needed for this, rather than ruby's `FileUtils.mv`?

reitermarkus:
> isn't it just moving files to the trash

The trash can either be at `~/.Trash`, or `/Volumes/External/.Trashes/{user_id}`, depending on which file is being trashed. Also, `mv` will not handle the case where a file with the same name is already in the trash. The system API handles all of this automatically.

vitorgalvao:
Asking the Finder to trash files is even better. It’s supposedly slower (not enough to make a difference in this case) but it makes the trashing sound and we get the “Put Back” feature.

MikeMcQuaid:
I think it would have been good to have a wider discussion amongst @Homebrew/maintainers before we added Swift code to this repository.

vitorgalvao:
We can use [`trash`](http://hasseg.org/trash/) instead. Not only is it ruby, it’s included as a formula in the core.

reitermarkus:
> Not only is it ruby, it’s included as a formula in the core.

It is not Ruby, or did I miss something? Anyways, I wanted to avoid having a Formula dependency for Homebrew itself.

MikeMcQuaid:
> It is not Ruby, or did I miss something?

Yeh, doesn't seem to be the case for me.

> Anyways, I wanted to avoid having a Formula dependency for Homebrew itself.

I don't think reinventing the wheel here is a great approach, either. What's the bus factor of this code? Who else who is a Homebrew maintainer has ever written any Swift?

woodruffw:
If the goal is to trash things, we could probably use AppleScript and invoke it from Ruby. That would avoid the need for a Swift program (which I'm personally :-1: on in the core).

chdiza:
> The trash can either be at ~/.Trash, or /Volumes/External/.Trashes/{user_id}, depending on which file is being trashed. Also, mv will not handle the case where a file with the same name is already in the trash.

All of this can be done in Ruby.

woodruffw:
> All of this can be done in Ruby.

I think @reitermarkus was saying that the decision between the two folders isn't clear, although I could be misunderstanding. Similarly, I don't think we should reimplement OS X's "trashing" logic - it's probably not formally specified anywhere, and can change whenever Apple pleases. It's much better, IMO, to rely on an external tool provided by the system itself.

reitermarkus:
> I don't think we should reimplement OS X's "trashing" logic

That was basically my intention here.

vitorgalvao:
> It is not Ruby, or did I miss something?

You’re right; I’m mistaking it for something else. I’m certain there’s a capable ruby trash gem, though, as I’ve used it in an Alfred Workflow in the past.

> I don't think we should reimplement OS X's "trashing" logic

Agreed.

alyssais:
We could use whatever the equivalent C APIs are using Fiddle in the Ruby standard library. http://ruby-doc.org/stdlib-2.0.0/libdoc/fiddle/rdoc/Fiddle.html Would be a bit of a pain, though.

alyssais:
> Who else who is a Homebrew maintainer has ever written any Swift?

Me, fwiw.

alyssais:
The Ruby trash gems I could find just seem to always use ~/.Trash, and implement unique file names themselves.

ilovezfs:
The feature proposed here doesn't itself seem sufficiently compelling to start reaching for Swift whenever there's a  Swift/Objective C only API that is "better."

ilovezfs:
PR to revert https://github.com/Homebrew/brew/pull/2808

reitermarkus:
@ilovezfs, instead of simply reverting this, what do you propose as an alternative?

ilovezfs:
I think an acceptable alternative is no trash feature.

MikeMcQuaid:
I'd suggest `mv ~/.Trash`

ilovezfs:
When a user then later clicks the `restore` button in the trash, is brew going to somehow restore the metadata for the cask, too?

vitorgalvao:
> I'd suggest `mv ~/.Trash`

Making a `system` call to AppleScript would be preferable, for the reasons already outlined by @reitermarkus.

> When a user then later clicks the `restore` button in the trash, is brew going to somehow restore the metadata for the cask, too?

The `trash` feature isn’t applied to everything, just to specific preference files. There’d be no point to restoring cask metadata, because the app itself wouldn’t be restored.

ilovezfs:
>Making a system call to AppleScript would be preferable

I think that would require the user to change security settings, and regardless of the security settings it would be pretty undesirable to mix AppleScript into brew.

Also, I don't think this Swift code will even work on < Yosemite.

vitorgalvao:
> I think that would require the user to change security settings

I don’t think it would. When asking the Finder trash files, it asks for elevated permissions if trying to delete files the user does not own. Either way, this feature would only be used on user preference files, so that’s not an issue. I’ve been using AppleScript (and now JXA) in ruby code for years (in the context of Alfred Workflows) and it never posed an issue to me or my users, and I have quite a few that are considerably less technical than the HB users (never touched a Terminal).

> Also, I don't think this Swift code will even work on < Yosemite.

That could be solved by a simple “if < Yosemite, use the old behaviour”. But the policy on HBC is to officially support two versions down and freqeuntly phase out old app versions from casks.

ilovezfs:
I don't think we should start intermixing non-Ruby code into brew just because there are nice Swift or AppleScript APIs. That is a precedent I'd rather not see us set, and there is no natural limit on it. What's the next feature that will justify bringing in more Swift or AppleScript? Can you really say there isn't going to be another one, and another one after that?

> But the policy on HBC is to officially support two versions down and freqeuntly phase out old app versions from casks.

App versions and the backend code are not even close to being the same thing.

vitorgalvao:
> What's the next feature that will justify bringing in more Swift or AppleScript? Can you really say there isn't going to be another one, and another one after that?

That’s a slippery slope. `trash` isn’t something we just came up with; it’s a feature that has been stubbed for years. Like it, many others exist, and no other would benefit from calling the system API, like this one does.

No user is asking for other features we don’t yet have that would benefit from AppleScript/Swift so I don’t think there’s a risk we will start adding that left and right, no.

> App versions and the backend code are not even close to being the same thing.

We’ve already specifically dropped support in the backend code for old app versions (PPC arch). The backend code serves to support the installation of apps, so they are closely related.

ilovezfs:
@vitorgalvao I think we're going to simply have to agree to disagree, as I am fairly certain at this point neither of our minds are going to change by further debate.

alyssais:
I think that

- This feature probably isn't worth the effort it's going to take to do it in a way we're all satisfied with
- If we _are_ going to have this feature, we should do it properly, through Apple's APIs, in whatever language that turns out to be.
- Swift is worse than doing this in either Ruby or AppleScript because of how often it changes in a backwards-compatible way.

apjanke:
I agree with @alyssais and @ilovezfs here. Adding another language to `brew` is a pretty big dependency to take for a small feature.

alyssais:
fwiw I'd be fine with AppleScript as long it was a very small amount

ilovezfs:
Maybe we can use Automator while we're at it.

ilovezfs:
Well it looks like we're already a bit of the way down the proverbial slippery slope:
https://github.com/Homebrew/brew/search?utf8=%E2%9C%93&q=osascript&type=

apjanke:
> fwiw I'd be fine with AppleScript as long it was a very small amount

Same. That's more like a system/utility call than a new language.

ilovezfs:
`osascript -e 'say "No trash for you."'`

ilovezfs:
>fwiw I'd be fine with AppleScript as long it was a very small amount

I still think it's :-1: when we can just use `mv`.

vitorgalvao:
> Well it looks like we're already a bit of the way down the proverbial slippery slope:
https://github.com/Homebrew/brew/search?utf8=%E2%9C%93&q=osascript&type=

Which is also something I can reply to and justify why it‘s not part of a slippery slope. But you [asked we stop trying to convince each other](https://github.com/Homebrew/brew/pull/2793#issuecomment-310337151) and I obliged without complaint. It’s my understanding from such a situation that you’d do the same. If we agree to cease our discussion but you continue to pile on arguments, that’s not exactly fair.

I only quoted that post because it’s the only one that pertains to our conversation.

MikeMcQuaid:
A reasonable compromise seems to be to use `osascript` here and remove the Swift code.

ilovezfs:
So we're all agreed on `osascript -e 'say "No trash for you."'` then?

PR #2795, opened 2017-06-17
formula_desc_cop: ensure no full stops at the end of desc

issyl0:

-----

- This was a suggestion by Mike McQuaid in my `homebrew-core` audit
  description PR. Currently, no formulae do have full stops at the
  end of descriptions (based on `ack "desc\b" *.rb | ack "\.$" *.rb`),
  but in the interests of future-proofing and consistency, we should
  check for this anyway?

### DIFF
--- a/Library/Homebrew/rubocops/formula_desc_cop.rb
+++ b/Library/Homebrew/rubocops/formula_desc_cop.rb
@@ -40,6 +40,7 @@ module RuboCop
       # - Checks for correct usage of `command-line` in `desc`
       # - Checks description starts with a capital letter
       # - Checks if `desc` contains the formula name
+      # - Checks if `desc` ends with a full stop
       class Desc < FormulaCop
         VALID_LOWERCASE_WORDS = %w[
           ex
@@ -78,8 +79,13 @@ module RuboCop
           end
 
           # Check if formula's desc starts with formula's name
-          return unless regex_match_group(desc, /^#{@formula_name} /i)
-          problem "Description shouldn't start with the formula name"
+          if regex_match_group(desc, /^#{@formula_name} /i)
+            problem "Description shouldn't start with the formula name"
+          end
+
+          # Check if a full stop is used at the end of a formula's desc
+          return unless regex_match_group(desc, /\.$/)
+          problem "Description shouldn't end with a full stop"
         end
 
         private
@@ -97,6 +103,7 @@ module RuboCop
             correction.gsub!(/(^|[^a-z])#{@formula_name}([^a-z]|$)/i, "\\1\\2")
             correction.gsub!(/^(['"]?)\s+/, "\\1")
             correction.gsub!(/\s+(['"]?)$/, "\\1")
+            correction.gsub!(/\.$/, "")
             corrector.insert_before(node.source_range, correction)
             corrector.remove(node.source_range)
           end
--- a/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
+++ b/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
@@ -91,6 +91,16 @@ describe RuboCop::Cop::FormulaAuditStrict::Desc do
       RUBY
     end
 
+    it "When the description ends with a full stop" do
+      expect_offense(<<~RUBY, "/homebrew-core/Formula/foo.rb")
+        class Foo < Formula
+          url 'http://example.com/foo-1.0.tgz'
+          desc 'Description with a full stop at the end.'
+                                                       ^ Description shouldn\'t end with a full stop
+        end
+      RUBY
+    end
+
     it "autocorrects all rules" do
       source = <<~EOS
         class Foo < Formula


Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
```diff
@@ -162,6 +162,26 @@ class Foo < Formula
       end
     end
 
+    when "desc ends in a full stop" do
```
GauthamGoli: `it "When ..." do`

Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
```diff
@@ -162,6 +162,26 @@ class Foo < Formula
...
+      expected_offenses = [{ message: "Description shouldn't end with a full stop",
+                             severity: :convention,
+                             line: 3,
+                             column: 8,
+                             source: source }]
+      inspect_source(cop, source)
+      expected_offenses.zip(cop.offenses).each do |expected, actual|
+        expect_offense(expected, actual)
+      end
+
```
GauthamGoli: Can remove this blank line

issyl0:
It turns out I got my regex wrong for this (7:30am isn't the best time to be thinking about that), so this isn't entirely speculative. Thanks to @ilovezfs for pointing that out in https://github.com/Homebrew/homebrew-core/pull/14229#discussion_r122562397. I've changed the regex to be correct now, as `desc` strings end in `"`, not *just* full stops.

Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
issyl0: 😥

reitermarkus: Still not quite.

```ruby
context "when ..." do
  it "does something" do
    ...
  end
end
```

issyl0: Thanks for pointing this out. I should (and do) know this (and I'm frustrated with myself)? 💦 

However, every other `it "When ...` block in this file is the same, there are no `context`s. I'm loathe to change them all in this PR, but I agree that "When" should be a `context`. What should I do?

reitermarkus: You could also use `specify`.

```ruby
specify "description shouldn't end with a full stop" do
  ...
end
```

issyl0: Thanks - I just made this change.

MikeMcQuaid:
@issyl0 Check out the failing `brew tests` and `brew style`.

issyl0:
@MikeMcQuaid Will do. On Tuesday when I'm back from holiday, if that's OK? Sorry that this is taking so long. I can close and reopen at a later date if you'd like?

MikeMcQuaid:
> On Tuesday when I'm back from holiday, if that's OK?

NO YOU MUST HOMEBREW ALL HOLIDAY

😉 

(Yes, of course, totally fine. Just had this in my queue to try and help so figured I'd jump in and offer some advice. Enjoy the holiday!)

issyl0:
I found a bit of downtime. 👍 

I don't know why my editor inserted a weird type of space (it looked like a space, but it wasn't a space), but that's fixed now.

As for the failing test, I don't understand it. No matter how many `puts` I put around the test case (for `expected`, `actual`, etc), I can't see where it's going wrong. I even added `'` as well as `"` to the regex in case it was that, but `actual` is still `nil` as it's not detecting a problem. It should be working. I wonder what silly thing I've done wrong or differently this time? It's probably something painfully obvious. Do you know, @MikeMcQuaid?

GauthamGoli:
@issyl0 You don't have to add `'` or `"` as `regex_match_group` method matches only with the string part.  You din't do anything wrong, except that this has to do with the ruby coding guidelines. 
Convert line 77-78 into an `if .. ` block, and also update your regex pattern for full stop matching( remove match for `'`/`"` ) . Tests should then pass 👍

issyl0:
Aha! Thanks, @GauthamGoli - I must have got confused with the syntax of the test above as that used `return unless` yet still worked. Thanks for setting me on the right track again with `regex_match_group`, too - I was overthinking the quotes again!

It doesn't seem to be working again yet though, and I've made the correct changes, I think.

issyl0:
Sorry this is taking so long and turning complicated!

MikeMcQuaid:
@issyl0 Never apologise for 1) trying when others don't and 2) learning publicly 👍 👏 ❤️

GauthamGoli:
@issyl0 Probably I told it wrong, but I meant, [Line 77 - 78](https://github.com/issyl0/brew/blob/63ebcfd5f97b1d462f1979cfba39ebccb066dbef/Library/Homebrew/rubocops/formula_desc_cop.rb#L77-L78) needs to be in `if-end` block while [Line 81 -83](https://github.com/issyl0/brew/blob/63ebcfd5f97b1d462f1979cfba39ebccb066dbef/Library/Homebrew/rubocops/formula_desc_cop.rb#L81-L83) has be `return unless ` guard clause, not vice versa. Otherwise Line 81 -83 may not get executed because there is a `return unless` statement on line 77.

JCount:
@issyl0 Feel free to rebase my commit, I thought I would just resolve the conflict since I was looking at this PR anyway. 😄

MikeMcQuaid:
@issyl0 Shout if we can help here at all but: no rush (and hope you had a good holiday)!

issyl0:
@JCount Thanks for that, but I did just rebase out your merge commit. :-)

issyl0:
@GauthamGoli Thanks! That makes a lot more sense, and now I can see the problem. I've just made that change and tests pass locally.

@MikeMcQuaid I did have a good holiday, thanks! The tests pass for this now, apart from `brew style` which is complaining about `if`s instead of `return`s again: `Use a guard clause instead of wrapping the code inside a conditional expression.`

DomT4:
What happened here? The PR looks to be in pretty good shape but it was closed as stale in the end. Was it just the guard clause hiccup?

issyl0:
@DomT4 That was it. And then I disappeared off the face of the planet, apparently. :-)

ilovezfs:
@issyl0 should we reopen it?

issyl0:
@ilovezfs If you'd like. I have time to work on it again I think (this week), and it's literally one thing that's wrong. But I don't remember if I ever knew how to fix it.

ilovezfs:
@issyl0 it will also need to be made consistent with https://github.com/Homebrew/brew/pull/3349 but that shouldn't be a huge deal.

ilovezfs:
@issyl0 #3349 has now been merged. Please rebase this PR on master when you have a chance! Thanks :)

ilovezfs:
@issyl0 
```
diff --git a/Library/Homebrew/rubocops/formula_desc_cop.rb b/Library/Homebrew/rubocops/formula_desc_cop.rb
index 4233c3a..bc36330 100644
--- a/Library/Homebrew/rubocops/formula_desc_cop.rb
+++ b/Library/Homebrew/rubocops/formula_desc_cop.rb
@@ -84,9 +84,8 @@ module RuboCop
           end
 
           # Check if a full stop is used at the end of a formula's desc
-          if regex_match_group(desc, /.*\.$/)
-            problem "Description shouldn't end with a full stop"
-          end
+          return unless regex_match_group(desc, /\.$/)
+          problem "Description shouldn't end with a full stop"
         end
 
         private
diff --git a/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb b/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
index c82a1f2..aacb52e 100644
--- a/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
+++ b/Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
@@ -96,9 +96,8 @@ describe RuboCop::Cop::FormulaAuditStrict::Desc do
         class Foo < Formula
           url 'http://example.com/foo-1.0.tgz'
           desc 'Description with a full stop at the end.'
-                                                       ^ Description shouldn\'t end with a full stop.
-          end
-      end
+                                                       ^ Description shouldn\'t end with a full stop
+        end
       RUBY
     end

**Comment by issyl0** [comment]
Me ending my Rubocop error message with a full stop was unfortunate given this PR. But is that the fix for the guard clause error, the `return unless`? Thank you so much!

**Comment by ilovezfs** [comment]
>Me ending my Rubocop error message with a full stop was unfortunate given this PR.

LOL. The same thing occurred to me :)

Yes the `return unless` is the guard clause. Unfortunately rubocop forces a different format for the last one in the function as opposed to all of the others. :see_no_evil:

**Comment by issyl0** [comment]
Ahhhh, that's why I had such trouble in the beginning with adding a new test and had to switch the old one around! It totally should have occurred to me. But thanks again. I just pushed the changes and fingers crossed it's green now! And now I know for next time!

File: Library/Homebrew/rubocops/formula_desc_cop.rb [hunk 3]
```diff
@@ -78,8 +79,13 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
...
 
           # Check if formula's desc starts with formula's name
-          return unless regex_match_group(desc, /^#{@formula_name} /i)
-          problem "Description shouldn't start with the formula name"
+          if regex_match_group(desc, /^#{@formula_name} /i)
+            problem "Description shouldn't start with the formula name"
+          end
+
+          # Check if a full stop is used at the end of a formula's desc
+          return unless regex_match_group(desc, /.*\.$/)
```
ilovezfs: You'll want to remove the .* here too otherwise the rubocop error message will underline the entire description with ^^^^^^^^^^^^^^^^^^^^^^^

issyl0: Thanks - done!

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -78,8 +79,13 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
...
 
           # Check if formula's desc starts with formula's name
-          return unless regex_match_group(desc, /^#{@formula_name} /i)
-          problem "Description shouldn't start with the formula name"
+          if regex_match_group(desc, /^#{@formula_name} /i)
+            problem "Description shouldn't start with the formula name"
+          end
+
+          # Check if a full stop is used at the end of a formula's desc
+          return unless regex_match_group(desc, /.\.$/)
```
ilovezfs: `/\.$/` should be fine.

`desc "."` would be bogus too, after all.

issyl0: Of course. I'm going to blame my post-conference illness for this lack of attention! 🤒

issyl0:
Yay, the build is green. Thanks very much for your help, everyone.

Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
```diff
@@ -91,6 +91,16 @@ class Foo < Formula
       RUBY
     end
 
+    it "When the description ends with a full stop" do
+      expect_offense(<<~RUBY, "/homebrew-core/Formula/foo.rb")
+        class Foo < Formula
+          url 'http://example.com/foo-1.0.tgz'
+          desc 'Description with a full stop at the end.'
+                                                       ^ Description shouldn\'t end with a full stop
```
MikeMcQuaid: Are you sure you need to escape the `\'`? If so: no worries.

Library/Homebrew/rubocops/formula_desc_cop.rb
```diff
@@ -78,8 +79,13 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
...
           # Check if formula's desc starts with formula's name
-          return unless regex_match_group(desc, /^#{@formula_name} /i)
-          problem "Description shouldn't start with the formula name"
+          if regex_match_group(desc, /^#{@formula_name} /i)
+            problem "Description shouldn't start with the formula name"
+          end
+
+          # Check if a full stop is used at the end of a formula's desc
+          return unless regex_match_group(desc, /\.$/)
+          problem "Description shouldn't end with a full stop"
```
MikeMcQuaid: Could you add an `autocorrect` for this on line 105 below? Something like this would work: `correction.gsub!(/\.$/, "")`

Library/Homebrew/test/rubocops/formula_desc_cop_spec.rb
issyl0: The other occurrences of "shouldn't" in this file escape the apostrophe.

MikeMcQuaid:
Thanks again for your contribution to Homebrew, @issyl0! Without people like you submitting PRs we couldn't run this project. You rock!

ilovezfs:
@issyl0 @MikeMcQuaid the autocorrect here doesn't actually work because the descriptions actually end in `'` or `"` not in `.` so I think we need something like
```
-            correction.gsub!(/\.$/, "")
+            correction.gsub!(/\.(['"]?)$/, "\\1")
```

However that can lead to some undesirable behavior such as
```
-  desc "Encrypts your files, so you can safely store them in Dropbox, iCloud, ..."
+  desc "Encrypts your files, so you can safely store them in Dropbox, iCloud, "
```

I've replaced that description with
```
  desc "Encrypts your files, so you can safely store them in Dropbox, iCloud, etc."
```

on the assumption that we also need to white list descriptions ending in `etc.`

issyl0:
@ilovezfs That's true. I'll do a PR for whitelisting "etc", and also fix up the regex when I get a chance probably tomorrow. I also had a local branch for fixing the formulae affected by this new audit, but you beat me to it, so thanks!

PR #2801, opened 2017-06-19
rubocops: refactor to use consistent _body_node parameter name

apjanke:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). *No new tests needed because no functionality change.*

-----

The various FormulaCop classes use different names for their fourth parameter (`_body_node`/`body_node`/`body`/`formula_class_body_node`). This PR changes them all to `_body_node` for consistency. I chose the `_body_node` variant for consistency with the other parameter names, which all have leading underscores and include `_node`.

No functional changes.

<strike>If this is approved, then, #2776 should also switch to `_body_node`.</strike>

Also changes a hash to hash rocket syntax to quiet `brew style rubocops/*.rb`.

### DIFF
--- a/Library/Homebrew/rubocops/bottle_block_cop.rb
+++ b/Library/Homebrew/rubocops/bottle_block_cop.rb
@@ -10,8 +10,8 @@ module RuboCop
       class BottleBlock < FormulaCop
         MSG = "Use rebuild instead of revision in bottle block".freeze
 
-        def audit_formula(_node, _class_node, _parent_class_node, formula_class_body_node)
-          bottle = find_block(formula_class_body_node, :bottle)
+        def audit_formula(_node, _class_node, _parent_class_node, body_node)
+          bottle = find_block(body_node, :bottle)
           return if bottle.nil? || block_size(bottle).zero?
           problem "Use rebuild instead of revision in bottle block" if method_called_in_block?(bottle, :revision)
         end
--- a/Library/Homebrew/rubocops/components_order_cop.rb
+++ b/Library/Homebrew/rubocops/components_order_cop.rb
@@ -8,7 +8,7 @@ module RuboCop
       # - component_precedence_list has component hierarchy in a nested list
       #   where each sub array contains components' details which are at same precedence level
       class ComponentsOrder < FormulaCop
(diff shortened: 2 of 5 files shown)


GauthamGoli:
When an argument is prefixed with `_` means we don't use it in the method. 
So we can't use `_body_node` when we are using that argument inside the method.

apjanke:
Oh, gotcha. I forgot about that convention. Will fix.

apjanke:
Fixed to use `body_node` or `_body_node` as appropriate. Also fixed a `brew style` complaint about hash syntax. `brew style` now comes back clean on all the cop files.

MikeMcQuaid:
@apjanke What do you think about using `formula_class_body_node` consistently?

apjanke:
Hmm. Seems a little wordy to me. And maybe redundant: all the parameters are sort of relative to the formula class definition, right? I.e. `node` is the formula class itself, `class_node` is the formula class's class name, `parent_class_node` is the formula class's parent class name. I don't feel strongly either way though; just as long as they're consistent.

MikeMcQuaid:
👍 when RuboCop is happy.

apjanke:
That's odd. RuboCop was happy when I ran it on just that file. (In fact, I made that change to satisfy its complaints about the old syntax.) But a full `brew tests` run complains about it.

```
[/usr/local/Homebrew/Library/Homebrew/rubocops on ⇄ formula-cop-uniform-_body_node-param]
$ brew style ./components_order_cop.rb
[/usr/local/Homebrew/Library/Homebrew/rubocops on ⇄ formula-cop-uniform-_body_node-param]
$ brew style
Inspecting 634 files
.................................................................................................................................................................................................................................................................................................................................................................................................C........................................................................................................................................................................................................................................................

Offenses:

components_order_cop.rb:13:16: C: Use the new Ruby 1.9 hash syntax.
            [{ :name => :include,  :type => :method_call }],
               ^^^^^^^^
components_order_cop.rb:13:36: C: Use the new Ruby 1.9 hash syntax.
            [{ :name => :include,  :type => :method_call }],
                                   ^^^^^^^^
```

I'll look in to why the `brew style` behavior is different there. In the mean time, reverting to satisfy `brew style`.

PR #2804, opened 2017-06-20
Increasing test coverage for caveats.rb

mansimarkaur:

-----
@mistydemeo

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -164,20 +164,8 @@ class Caveats
   def plist_caveats
     s = []
     if f.plist || (keg && keg.plist_installed?)
-      destination = if f.plist_startup
-        "/Library/LaunchDaemons"
-      else
-        "~/Library/LaunchAgents"
-      end
-
-      plist_filename = if f.plist
-        f.plist_path.basename
-      else
-        File.basename Dir["#{keg}/*.plist"].first
-      end
       plist_domain = f.plist_path.basename(".plist")
-      destination_path = Pathname.new File.expand_path destination
-      plist_path = destination_path/plist_filename
+      plist_path = plist_complete_path
 
       # we readlink because this path probably doesn't exist since caveats
       # occurs before the link step of installation
@@ -209,10 +197,29 @@ class Caveats
         s << "  #{f.plist_manual}"
       end
 
+      # pbpaste is a system clipboard tool for OSX that doesn't work well with tmux
+      # This checks if tmux is being used and warns about brew failure
       if ENV["TMUX"] && !quiet_system("/usr/bin/pbpaste")
         s << "" << "WARNING: brew services will fail when run under tmux."
       end
     end
     s.join("\n") + "\n" unless s.empty?
+  end
+
+  def plist_complete_path
+    destination = if f.plist_startup
+      "/Library/LaunchDaemons"
+    else
+      "~/Library/LaunchAgents"
+    end
+
+    plist_filename = if f.plist
+      f.plist_path.basename
+    else
+      File.basename Dir["#{keg}/*.plist"].first
+    end
+    destination_path = Pathname.new File.expand_path destination
+
+    return destination_path/plist_filename
   end
 end
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -196,7 +196,7 @@ class Caveats
         s << "  #{f.plist_manual}"
       end
 
-      # pbpaste pastes the system clipboard tool on macOS and fails with `tmux` by default
+      # pbpaste is the system clipboard tool on macOS and fails with `tmux` by default
       # check if this is being run under `tmux` to avoid failing
       if ENV["TMUX"] && !quiet_system("/usr/bin/pbpaste")
         s << "" << "WARNING: brew services will fail when run under tmux."
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -165,7 +165,6 @@ class Caveats
     s = []
     if f.plist || (keg && keg.plist_installed?)
       plist_domain = f.plist_path.basename(".plist")
-      plist_path = plist_complete_path
 
       # we readlink because this path probably doesn't exist since caveats
       # occurs before the link step of installation
@@ -197,8 +196,8 @@ class Caveats
         s << "  #{f.plist_manual}"
       end
 
-      # pbpaste is a system clipboard tool for OSX that doesn't work well with tmux
-      # This checks if tmux is being used and warns about brew failure
+      # pbpaste pastes the system clipboard tool on macOS and fails with `tmux` by default
+      # check if this is being run under `tmux` to avoid failing
       if ENV["TMUX"] && !quiet_system("/usr/bin/pbpaste")
         s << "" << "WARNING: brew services will fail when run under tmux."
       end
@@ -206,7 +205,7 @@ class Caveats
     s.join("\n") + "\n" unless s.empty?
   end
 
-  def plist_complete_path
+  def plist_path
     destination = if f.plist_startup
       "/Library/LaunchDaemons"
     else
@@ -218,8 +217,8 @@ class Caveats
     else
       File.basename Dir["#{keg}/*.plist"].first
     end
-    destination_path = Pathname.new File.expand_path destination
+    destination_path = Pathname.new(File.expand_path(destination))
 
-    return destination_path/plist_filename
+    destination_path/plist_filename
   end
 end
--- a/Library/Homebrew/test/caveats_spec.rb
+++ b/Library/Homebrew/test/caveats_spec.rb
@@ -28,7 +28,7 @@ describe Caveats do
   end
 
   describe "#caveats" do
-    context "when f.plist is not nil" do
+    context "when f.plist is not nil", :needs_macos do
       it "prints plist startup information when f.plist_startup is not nil" do
         f = formula do
           url "foo-1.0"
@@ -199,6 +199,46 @@ describe Caveats do
       it "gives dir where fish completions have been installed" do
         (path/"share/fish/vendor_completions.d").mkpath
         expect(caveats).to include(HOMEBREW_PREFIX/"share/fish/vendor_completions.d")
+      end
+    end
+
+    context "python caveats" do
+      before do
+        (f.prefix.resolved_path/"lib/python2.7/site-packages").mkpath
+      end
+
+      context "when f is not keg_only" do
+        let(:f) {
+          formula do
+            url "foo-1.0"
+          end
+        }
+        let(:caveats) { described_class.new(f).caveats }
+        let(:user_site_packages) { Language::Python.user_site_packages("python") }
+
+        it "give commands to run when Homebrew's site-packages is not in Python sys.path" do
+          expect(caveats).to include("Homebrew's site-packages is not\nin your Python sys.path")
+          expect(caveats).to include(user_site_packages)
+          expect(caveats).to include("import site")
+        end
+
+        it "gives commands to run when python pth files are installed" do
+          allow(Homebrew).to receive(:_system).and_return(true)
+          allow(Dir).to receive(:[]).with(any_args).and_return(["blah.pth"])
+          expect(caveats).to include(".pth files to Homebrew's site-packages and your\nPython isn't configured")
+          expect(caveats).to include(user_site_packages)
+          expect(caveats).to include("import site")
+        end
+      end
+
+      it "gives commands to run when formula is keg_only" do
+        f = formula do
+          url "foo-1.0"
+          keg_only "some reason"
+        end
+        caveats = described_class.new(f).caveats
+        homebrew_site_packages = Language::Python.homebrew_site_packages
+        expect(caveats).to include("echo #{f.opt_prefix}/lib/python2.7/site-packages >> #{homebrew_site_packages/f.name}.pth")
       end
     end
   end
--- a/Library/Homebrew/test/caveats_spec.rb
+++ b/Library/Homebrew/test/caveats_spec.rb
@@ -1,6 +1,5 @@
 require "formula"
 require "caveats"
-require "pathname"
 
 describe Caveats do
   subject { described_class.new(f) }
@@ -49,6 +48,81 @@ describe Caveats do
           end
         end
         expect(described_class.new(f).caveats).to include("login")
+      end
+
+      it "gives information about restarting services after upgrade" do
+        f = formula do
+          url "foo-1.0"
+          def plist
+            "plist_test.plist"
+          end
+          plist_options startup: true
+        end
+        f_obj = described_class.new(f)
+        plist_path = Pathname.new("plist")
+        FileUtils.touch plist_path
+        allow(f_obj).to receive(:plist_path).and_return(plist_path)
+        allow(plist_path).to receive(:symlink?).and_return(true)
+        expect(f_obj.caveats).to include("restart #{f.full_name}")
+        expect(f_obj.caveats).to include("sudo")
+      end
+
+      context "when plist_path is not a file nor symlinked and plist_startup is false" do
+        let(:f) {
+          formula do
+            url "foo-1.0"
+            def plist
+              "plist_test.plist"
+            end
+          end
+        }
+        let(:f_obj) { described_class.new(f) }
+        let(:caveats) { f_obj.caveats }
+        let(:plist_path) { Pathname.new("plist") }
+
+        before do
+          FileUtils.touch plist_path
+          allow(f_obj).to receive(:plist_path).and_return(plist_path)
+          allow(plist_path).to receive(:symlink?).and_return(true)
+        end
+
+        it "tells command to run after upgrade" do
+          allow(Kernel).to receive(:system).with(any_args).and_return(true)
+          expect(caveats).to include("restart #{f.full_name} after an upgrade")
+        end
+
+        it "tells command to run to start formula" do
+          expect(caveats).to include("To start #{f.full_name}:")
+        end
+      end
+
+      it "gives information about plist_manual" do
+        f = formula do
+          url "foo-1.0"
+          def plist
+            "plist_test.plist"
+          end
+          plist_options manual: "foo"
+        end
+        caveats = described_class.new(f).caveats
+
+        expect(caveats).to include("background service")
+        expect(caveats).to include(f.plist_manual)
+      end
+
+      it "warns about brew failing under tmux" do
+        f = formula do
+          url "foo-1.0"
+          def plist
+            "plist_test.plist"
+          end
+        end
+        allow(ENV).to receive(:[]).with("TMUX").and_return(true)
+        allow(Homebrew).to receive(:_system).with("/usr/bin/pbpaste").and_return(false)
+        caveats = described_class.new(f).caveats
+
+        expect(caveats).to include("WARNING:")
+        expect(caveats).to include("tmux")
       end
     end
 
@@ -125,54 +199,6 @@ describe Caveats do
       it "gives dir where fish completions have been installed" do
         (path/"share/fish/vendor_completions.d").mkpath
         expect(caveats).to include(HOMEBREW_PREFIX/"share/fish/vendor_completions.d")
-      end
-    end
-
-    context "when plist_caveats are given" do
-      it "gives information about plist_manual" do
-        f = formula do
-          url "foo-1.0"
-          def plist
-            "plist_test.plist"
-          end
-          plist_options manual: "foo"
-        end
-        caveats = described_class.new(f).caveats
-
-        expect(caveats).to include("background service")
-        expect(caveats).to include(f.plist_manual)
-      end
-
-      it "warns about brew failing under tmux" do
-        f = formula do
-          url "foo-1.0"
-          def plist
-            "plist_test.plist"
-          end
-        end
-        allow(ENV).to receive(:[]).with("TMUX").and_return(true)
-        allow(Homebrew).to receive(:_system).with("/usr/bin/pbpaste").and_return(false)
-        caveats = described_class.new(f).caveats
-
-        expect(caveats).to include("WARNING:")
-        expect(caveats).to include("tmux")
-      end
-
-      it "gives information about restarting services after upgrade" do
-        f = formula do
-          url "foo-1.0"
-          def plist
-            "plist_test.plist"
-          end
-          plist_options startup: true
-        end
-        f_obj = described_class.new(f)
-        plist_path = Pathname.new("plist")
-        FileUtils.touch plist_path
-        allow(f_obj).to receive(:plist_complete_path).and_return(plist_path)
-        allow(plist_path).to receive(:symlink?).and_return(true)
-        expect(f_obj.caveats).to include("restart #{f.full_name}")
-        expect(f_obj.caveats).to include("sudo")
       end
     end
   end
--- a/Library/Homebrew/test/caveats_spec.rb
+++ b/Library/Homebrew/test/caveats_spec.rb
@@ -26,4 +26,51 @@ describe Caveats do
       expect(described_class.new(f)).not_to be_empty
     end
   end
+
+  describe "#caveats" do
+    context "when f.plist is not nil" do
+      it "prints plist startup information when f.plist_startup is not nil" do
+        f = formula do
+          url "foo-1.0"
+          def plist
+            "plist_test.plist"
+          end
+          plist_options startup: true
+        end
+        expect(described_class.new(f).caveats).to include("startup:\n  sudo brew")
+      end
+
+      it "prints plist login information when f.plist_startup is nil" do
+        f = formula do
+          url "foo-1.0"
+          def plist
+            "plist_test.plist"
+          end
+        end
+        expect(described_class.new(f).caveats).to include("login:\n  brew")
+      end
+    end
+
+    context "when f.keg_only is not nil" do
+      it "tells formula is keg_only and gives information about command to be run when f.bin and f.sbin are directories" do
+        Path = Pathname.new("path")
+        f = formula do
+          url "foo-1.0"
+          keg_only "some reason"
+        end
+
+        allow(f).to receive(:bin).and_return(Path)
+        allow(f.bin).to receive(:directory?).and_return(true)
+
+        allow(f).to receive(:sbin).and_return(Path)
+        allow(f.sbin).to receive(:directory?).and_return(true)
+
+        caveats = described_class.new(f).caveats
+
+        expect(caveats).to include("keg-only")
+        expect(caveats).to include(f.opt_bin.to_s)
+        expect(caveats).to include(f.opt_sbin.to_s)
+      end
+    end
+  end
 end
--- a/Library/Homebrew/test/caveats_spec.rb
+++ b/Library/Homebrew/test/caveats_spec.rb
@@ -1,5 +1,6 @@
 require "formula"
 require "caveats"
+require "pathname"
 
 describe Caveats do
   subject { described_class.new(f) }
@@ -37,7 +38,7 @@ describe Caveats do
           end
           plist_options startup: true
         end
-        expect(described_class.new(f).caveats).to include("startup:\n  sudo brew")
+        expect(described_class.new(f).caveats).to include("startup")
       end
 
       it "prints plist login information when f.plist_startup is nil" do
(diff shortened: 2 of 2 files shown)


Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,39 @@ def caveats
       expect(described_class.new(f)).not_to be_empty
     end
   end
+
+  describe "#caveats" do
+    describe "when f.plist is not null" do
+      it "when plist_startup is not null" do
```
mistydemeo: For `it` test names, we want to describe what the test does, not just when it applies. For example, `it "prints plist startup information when plist_startup is not null"`.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,39 @@ def caveats
...
+    describe "when f.keg_only is not null" do
+      it "" do
+        allow(f).to receive(:keg_only?).and_return(true)
+        expect(subject.caveats).to include("This formula is keg-only, which means it was not symlinked into #{HOMEBREW_PREFIX},\nbecause #{f.keg_only_reason.to_s.chomp}.")
+      end
+
+      it "f.bin.directory? is true" do
+        allow(f).to receive(:keg_only?).and_return(true)
+        bin = f.bin
+        allow(bin).to_receive(:directory?).and_return(true)
```
mistydemeo: Are you able to move this and the previous line onto one line, like you do in the `sbin` test?

mistydemeo:
Looks like there's a couple of typoes in the `keg_only` tests:

https://travis-ci.org/Homebrew/brew/jobs/244744936#L287 - `undefined method `to_receive'`
https://travis-ci.org/Homebrew/brew/jobs/244744936#L299 - `undefined method `to_receive'`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,39 @@ def caveats
       expect(described_class.new(f)).not_to be_empty
     end
   end
+
+  describe "#caveats" do
+    describe "when f.plist is not null" do
```
reitermarkus: For blocks beginning with „when“, you should use `context` instead of `describe`.

mansimarkaur: I wanted to stub methods received by `f` and these can't be stubbed inside `context` / `describe` blocks, hence, I chose `it`. I'll use `context` when using `when` from  now on.

reitermarkus: You can also stub in `before`/`after` blocks, which could be useful.

mansimarkaur: Okay. Thanks @reitermarkus

mansimarkaur: Since I'm not using "when" in my code now, should I prefer changing some blocks to use `context` and "when" instead of `it`, @reitermarkus ?

reitermarkus: The `describe "when ..."` blocks are redundant now. Everything else looks fine. 👍

mansimarkaur: I don't have any `describe "when ..."` blocks in my code now. Do you think it's fine now?

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,40 @@ def caveats
...
+      end
+
+      it "prints plist login information when f.plist_startup is null" do
+        allow(f).to receive(:plist).and_return(true)
+        expect(subject.caveats).to include("To have launchd start #{f.full_name} now and restart at login:\n  brew services start #{f.full_name}\n")
+      end
+    end
+
+    context "when f.keg_only is not null" do
+      it "prints formula is keg_only when keg_only is not null" do
```
mistydemeo: Minor nitpick: we don't need `when keg_only is not null` since that's already a part of the `context` here.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,36 @@ def caveats
...
+
+  describe "#caveats" do
+    context "when f.plist is not null" do
+      it "prints plist startup information when plist_startup is not null" do
+        allow(f).to receive(:plist).and_return(true)
+        allow(f).to receive(:plist_startup).and_return(true)
+        expect(subject.caveats).to include("To have launchd start #{f.full_name} now and restart at startup:\n  sudo brew services start #{f.full_name}\n")
+      end
+
+      it "prints plist login information when f.plist_startup is null" do
```
reitermarkus: Two small nits: Above you are using only `plist_startup`, here you are using `f.plist_startup`. Secondly, use `nil` instead of `null`, to be consistent with Ruby.

mansimarkaur: I've made the changes, @reitermarkus. Thanks for reviewing :)

mansimarkaur:
There is something wrong with the build: https://travis-ci.org/Homebrew/brew/jobs/246014857#L99

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,35 @@ def caveats
       expect(described_class.new(f)).not_to be_empty
     end
   end
+
+  describe "#caveats" do
+    context "when f.plist is not nil" do
+      it "prints plist startup information when f.plist_startup is not nil" do
+        allow(f).to receive(:plist).and_return(true)
+        allow(f).to receive(:plist_startup).and_return(true)
+        expect(subject.caveats).to include("To have launchd start #{f.full_name} now and restart at startup:\n  sudo brew services start #{f.full_name}\n")
```
MikeMcQuaid: Shorten the `include` string to match just `sudo brew services start`. That will make this more resistant to language changes. This also applies to the test below.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,35 @@ def caveats
       expect(described_class.new(f)).not_to be_empty
     end
   end
+
+  describe "#caveats" do
+    context "when f.plist is not nil" do
+      it "prints plist startup information when f.plist_startup is not nil" do
+        allow(f).to receive(:plist).and_return(true)
+        allow(f).to receive(:plist_startup).and_return(true)
```
MikeMcQuaid: It would be preferable to use a `formula do` and test this on an actual formula rather than mock these. This also applies to tests below.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,35 @@ def caveats
...
+        expect(subject.caveats).to include("To have launchd start #{f.full_name} now and restart at login:\n  brew services start #{f.full_name}\n")
+      end
+    end
+
+    context "when f.keg_only is not nil" do
+      it "gives information about command to be run when f.bin is a directory" do
+        allow(f).to receive(:keg_only?).and_return(true)
+        allow(f).to receive(:bin).and_return(Pathname.new("bin"))
+        allow(f.bin).to receive(:directory?).and_return(true)
+        expect(subject.caveats).to include("This formula is keg-only, which means it was not symlinked into #{HOMEBREW_PREFIX},\nbecause #{f.keg_only_reason.to_s.chomp}.\n\nIf you need to have this software first in your PATH run:\n  #{Utils::Shell.prepend_path_in_profile(f.opt_bin.to_s)}\n\n")
```
MikeMcQuaid: I'd just check for something like `formula is keg-only` here (and test below)

mansimarkaur: You mean I should not be checking if the string is included in caveats and just check if f.keg_only is true or not?

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: This is how plist is defined in formula.rb. So, that's why I'm mocking `f.plist`.
```
def plist
    nil
end```
For `plist_startup`, I'm trying to set it as true through `plist_options`: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L2361
Do you think a `formula do` will still be needed?

Comment (MikeMcQuaid) [inline Library/Homebrew/test/caveats_spec.rb]: You should be able to do something like:

```ruby
formula do
...
  def plist
    "..."
  end
end
```

It's just preferable to do more of an integration test using an actual formula class than a unit test where mocking is done. We want to avoid mocking whenever we can do so; it aids coverage and testing the code actually works interfacing with other code as it normally would.

Library/Homebrew/test/caveats_spec.rb
MikeMcQuaid: Sorry, no, I mean just string string `formula is keg-only` which means this test is more resistant to rewording of this message. My bad explaining 😆

MikeMcQuaid: Sorry for terrible previous message:

Sorry, no, I mean instead of checking for the long string being `include`d that you currently have just check for `include("formula is keg-only")`. This will mean that this test is more resistant (i.e. less likely to break) if we tweak the wording of this message in future. Does that make sense? Thanks!

MikeMcQuaid:
@mansimarkaur Any news here?

mansimarkaur:
One of the tests is failing. `allow(g.sbin).to receive(:directory?).and_return(true)` isn't working correctly.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,65 @@ def caveats
...
     end
   end
+
+  describe "#caveats" do
+    context "when f.plist is not nil" do
+      it "prints plist startup information when f.plist_startup is not nil" do
+        p = formula do
+              url "foo-1.0"
+              def plist
+                true
```
MikeMcQuaid: Define an actual `plist` here; check formulae for usage.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,65 @@ def caveats
...
+              end
+              def plist_startup
+                true
```
MikeMcQuaid: Should be `plist_options :startup => true`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,65 @@ def caveats
...
+        expect(described_class.new(p).caveats).to include("login:\n  brew")
+      end
+    end
+
+    context "when f.keg_only is not nil" do
+      it "gives information about command to be run when f.bin is a directory" do
+        p = formula do
+              url "foo-1.0"
+              def keg_only?
+                true
```
MikeMcQuaid: Should be `keg_only "some reason"`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,65 @@ def caveats
...
+      end
+
+      it "gives information about command to be run when f.sbin is a directory" do
+        g = formula do
+              url "foo-1.0"
+              def keg_only?
+                true
+              end
+              def sbin
+                Pathname.new("sbin")
```
MikeMcQuaid: There's no need to define this method.

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: Got it. Added an actual plist.

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: Corrected. Thanks

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: Both `p.bin.directory?` and `g.sbin.directory?` give `false` even after mocking them (p and g are formulae).
If I mock values of `p.bin` and `g.sbin`, then the mocking for `p.bin.directory?` and `g.sbin.directory?` works.
Could you please help figure out why this is happening? @mistydemeo @MikeMcQuaid

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+    end
+
+    context "when f.keg_only is not nil" do
+      it "gives information about command to be run when f.bin is a directory" do
+        p = formula do
+              url "foo-1.0"
+              keg_only "some reason"
+            end
+        p_inst = described_class.new(p)
+        allow(p.bin).to receive(:directory?).and_return(true)
```
MikeMcQuaid: This is operating on an instance that will be different every time you (or the Homebrew internals) call `p.bin`. You may need to stub `Pathname` globally instead.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+        expect(p_inst.caveats).to include(Utils::Shell.prepend_path_in_profile(p.opt_bin.to_s))
+      end
+
+      it "gives information about command to be run when f.sbin is a directory" do
+        g = formula do
+              url "foo-1.0"
+              keg_only "some reason"
+            end
+        g_inst = described_class.new(g)
+        allow(g.sbin).to receive(:directory?).and_return(true)
```
MikeMcQuaid: This is operating on an instance that will be different every time you (or the Homebrew internals) call `p.bin`. You may need to stub `Pathname` globally instead.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
       expect(described_class.new(f)).not_to be_empty
     end
   end
+
+  describe "#caveats" do
+    context "when f.plist is not nil" do
+      it "prints plist startup information when f.plist_startup is not nil" do
+        p = formula do
```
MikeMcQuaid: `f = formula do`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+              def plist
+                "plist_test.plist"
+              end
+              plist_options :startup => true
+            end
+        expect(described_class.new(p).caveats).to include("startup:\n  sudo brew")
+      end
+
+      it "prints plist login information when f.plist_startup is nil" do
+        p = formula do
```
MikeMcQuaid: `f = formula do`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+                "plist_test.plist"
+              end
+            end
+        expect(described_class.new(p).caveats).to include("login:\n  brew")
+      end
+    end
+
+    context "when f.keg_only is not nil" do
+      it "gives information about command to be run when f.bin is a directory" do
+        p = formula do
```
MikeMcQuaid: `f = formula do`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+              keg_only "some reason"
+            end
+        p_inst = described_class.new(p)
+        allow(p.bin).to receive(:directory?).and_return(true)
+        expect(p_inst.caveats).to include("This formula is keg-only")
+        expect(p_inst.caveats).to include(Utils::Shell.prepend_path_in_profile(p.opt_bin.to_s))
+      end
+
+      it "gives information about command to be run when f.sbin is a directory" do
+        g = formula do
```
MikeMcQuaid: `f = formula do`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+
+    context "when f.keg_only is not nil" do
+      it "gives information about command to be run when f.bin is a directory" do
+        p = formula do
+              url "foo-1.0"
+              keg_only "some reason"
+            end
+        p_inst = described_class.new(p)
+        allow(p.bin).to receive(:directory?).and_return(true)
+        expect(p_inst.caveats).to include("This formula is keg-only")
```
MikeMcQuaid: `include("keg-only")`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+        expect(p_inst.caveats).to include(Utils::Shell.prepend_path_in_profile(p.opt_bin.to_s))
```
MikeMcQuaid: `include(f.opt_bin.to_s)`

Library/Homebrew/test/caveats_spec.rb
MikeMcQuaid: `caveats = described_class.new(f).caveats` above this line to and then you don't need to repeat `p_inst.caveats` or declare `p_inst`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+      end
+
+      it "gives information about command to be run when f.sbin is a directory" do
+        g = formula do
+              url "foo-1.0"
+              keg_only "some reason"
+            end
+        g_inst = described_class.new(g)
+        allow(g.sbin).to receive(:directory?).and_return(true)
+        expect(g_inst.caveats).to include("This formula is keg-only")
```
MikeMcQuaid: `include("keg-only")`

MikeMcQuaid: `caveats = described_class.new(f).caveats` above this line to and then you don't need to repeat `p_inst.caveats` or declare `p_inst`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +26,53 @@ def caveats
...
+        expect(g_inst.caveats).to include(Utils::Shell.prepend_path_in_profile(g.opt_sbin.to_s))
```
MikeMcQuaid: `include(f.opt_sbin.to_s)`

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: If I use `f = formula do`, https://github.com/mansimarkaur/brew/blob/4b4762f5cba3d183d5347497df1bf8a524ce035e/Library/Homebrew/test/caveats_spec.rb#L6 interferes with this definition.

MikeMcQuaid: Can you elaborate on "interferes"? It seems fine in this test: https://github.com/mansimarkaur/brew/blob/4b4762f5cba3d183d5347497df1bf8a524ce035e/Library/Homebrew/test/caveats_spec.rb#L17

mansimarkaur: Using `f = formula do` works. There was some error earlier which got mistakenly corrected after I used another variable for the formula.

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: Using `caveats = described_class.new(f).caveats` gives the following error:
```
Failure/Error: caveats = described_class.new(f).caveats

     Errno::ENOENT:
       No such file or directory - /tmp/homebrew-tests-20170717-70571-15k7shm/cellar/formula_name
     # ./extend/pathname.rb:341:in `subdirs'
     # ./formula.rb:605:in `installed_prefixes'
     # ./tab.rb:142:in `for_formula'
     # ./caveats.rb:12:in `caveats'
     # ./test/caveats_spec.rb:78:in `block (4 levels) in <top (required)>'
     # ./test/spec_helper.rb:79:in `block (2 levels) in <top (required)>'
```

mansimarkaur: This error is arising because I stubbed all_instances of Pathname to be directories.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+          end
+        end
+        expect(described_class.new(f).caveats).to include("login:\n  brew")
+      end
+    end
+
+    context "when f.keg_only is not nil" do
+      Path = Pathname.new("path")
+
+      ## Refactoring needed:
```
MikeMcQuaid: Use `let(:f) { formula { ... } }` in each of the `context`s to set `f` for all the values in the context.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+          def plist
+            "plist_test.plist"
+          end
+        end
+        expect(described_class.new(f).caveats).to include("login:\n  brew")
+      end
+    end
+
+    context "when f.keg_only is not nil" do
+      Path = Pathname.new("path")
```
MikeMcQuaid: Use `let(:path)` here.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+      end
+
+      it "gives command to be run when f.bin is a directory" do
+        f = formula do
+          url "foo-1.0"
+          keg_only "some reason"
+        end
+
+        allow(f).to receive(:bin).and_return(Path)
+        allow(f.bin).to receive(:directory?).and_return(true)
```
MikeMcQuaid: `allow(path)` rather than `f.bin`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+
+          allow(f).to receive(:include).and_return(Path)
+          allow(f.include).to receive(:directory?).and_return(true)
+
+          caveats = described_class.new(f).caveats
+
+          expect(caveats).to include("-I#{f.opt_include}\n")
+        end
+
+        it "gives PKG_CONFIG_PATH when f.share/\"pkgconfig\" is a directory" do
```
MikeMcQuaid: `f.share/'pkgconfig'`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+          f = formula do
+            url "foo-1.0"
+            keg_only "some reason"
+          end
+
+          Pathname.new(f.share/"pkgconfig").mkpath
```
MikeMcQuaid: I'd consider using `mkpath` in all the above examples, too, instead of stubbing.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+    context "when f.plist is not nil" do
+      it "prints plist startup information when f.plist_startup is not nil" do
+        f = formula do
+          url "foo-1.0"
+          def plist
+            "plist_test.plist"
+          end
+          plist_options startup: true
+        end
+        expect(described_class.new(f).caveats).to include("startup:\n  sudo brew")
```
MikeMcQuaid: `include("startup:")`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+      end
+
+      it "prints plist login information when f.plist_startup is nil" do
+        f = formula do
+          url "foo-1.0"
+          def plist
+            "plist_test.plist"
+          end
+        end
+        expect(described_class.new(f).caveats).to include("login:\n  brew")
```
MikeMcQuaid: `include("login:")`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+            url "foo-1.0"
+            keg_only "some reason"
+          end
+
+          allow(f).to receive(:lib).and_return(Path)
+          allow(f.lib).to receive(:directory?).and_return(true)
+
+          caveats = described_class.new(f).caveats
+
+          expect(caveats).to include("-L#{f.opt_lib}\n")
```
MikeMcQuaid: `include("-L#{f.opt_lib}")`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+            url "foo-1.0"
+            keg_only "some reason"
+          end
+
+          allow(f).to receive(:include).and_return(Path)
+          allow(f.include).to receive(:directory?).and_return(true)
+
+          caveats = described_class.new(f).caveats
+
+          expect(caveats).to include("-I#{f.opt_include}\n")
```
MikeMcQuaid: `include("-I#{f.opt_include}")`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+            url "foo-1.0"
+            keg_only "some reason"
+          end
+
+          Pathname.new(f.share/"pkgconfig").mkpath
+          Pathname.new(f.lib/"pkgconfig").mkpath
+
+          caveats = described_class.new(f).caveats
+
+          expect(caveats).to include("#{f.opt_lib}/pkgconfig\n")
```
MikeMcQuaid: `include("#{f.opt_lib}/pkgconfig")`

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -26,4 +27,123 @@ def caveats
...
+          expect(caveats).to include("#{f.opt_share}/pkgconfig\n")
```
MikeMcQuaid: `include("#{f.opt_share}/pkgconfig")`

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: Why should `''` be preferred over `""`?

MikeMcQuaid: Avoids the `\"`

Library/Homebrew/caveats.rb
```diff
@@ -209,10 +197,29 @@ def plist_caveats
...
+    end
+
+    plist_filename = if f.plist
+      f.plist_path.basename
+    else
+      File.basename Dir["#{keg}/*.plist"].first
+    end
+    destination_path = Pathname.new File.expand_path destination
+
+    return destination_path/plist_filename
```
MikeMcQuaid: No need for `return` here; the last statement in a Ruby function is explicitly returned.

Library/Homebrew/caveats.rb
```diff
@@ -209,10 +197,29 @@ def plist_caveats
...
+    else
+      "~/Library/LaunchAgents"
+    end
+
+    plist_filename = if f.plist
+      f.plist_path.basename
+    else
+      File.basename Dir["#{keg}/*.plist"].first
+    end
+    destination_path = Pathname.new File.expand_path destination
```
MikeMcQuaid: `destination_path = Pathname.new(File.expand_path(destination))` makes this a bit easier to follow

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -1,5 +1,6 @@
 require "formula"
 require "caveats"
+require "pathname"
```
MikeMcQuaid: Do the new tests fail without this? If so, 👍 to leave it, otherwise remove.

Library/Homebrew/caveats.rb
```diff
@@ -209,10 +197,29 @@ def plist_caveats
...
+      # pbpaste is a system clipboard tool for OSX that doesn't work well with tmux
+      # This checks if tmux is being used and warns about brew failure
       if ENV["TMUX"] && !quiet_system("/usr/bin/pbpaste")
         s << "" << "WARNING: brew services will fail when run under tmux."
       end
     end
     s.join("\n") + "\n" unless s.empty?
   end
+
+  def plist_complete_path
```
MikeMcQuaid: I'd just call this `plist_path`

Library/Homebrew/caveats.rb
```diff
@@ -209,10 +197,29 @@ def plist_caveats
         s << "  #{f.plist_manual}"
       end
 
+      # pbpaste is a system clipboard tool for OSX that doesn't work well with tmux
```
MikeMcQuaid:
```
pbpaste pastes the system clipboard tool on macOS and fails with `tmux` by default
```

Library/Homebrew/caveats.rb
```diff
@@ -209,10 +197,29 @@ def plist_caveats
...
+      # This checks if tmux is being used and warns about brew failure
```
MikeMcQuaid:
```
check if this is being run under `tmux` to avoid failing
```

Library/Homebrew/caveats.rb
MikeMcQuaid: `brew style` told you this, FYI.

Library/Homebrew/test/caveats_spec.rb
mansimarkaur: No, they don't fail. I'll remove it.

Library/Homebrew/test/caveats_spec.rb
```diff
@@ -91,7 +91,7 @@ def plist
           expect(caveats).to include("restart #{f.full_name} after an upgrade")
         end
 
-        it "tells command to run to start formula" do
+        it "tells command to run to start formula", :needs_macos do
```
mistydemeo: How about adding this to all the plist tests? You can also add this to `context` and `describe` calls to skip large sets of tests.

mansimarkaur: Yes, that makes sense. I'll make the changes. Thanks :)

Library/Homebrew/caveats.rb
```diff
@@ -209,10 +196,29 @@ def plist_caveats
         s << "  #{f.plist_manual}"
       end
 
+      # pbpaste pastes the system clipboard tool on macOS and fails with `tmux` by default
```
mistydemeo: Nitpick - probably mean to say "pbpaste is the system clipboard tool"?

mansimarkaur: Yes! Made the change.

mistydemeo:
🚢

PR #2805, opened 2017-06-20
linkage_checker: Add harmless dylibs check.

woodruffw:
While the linkage checker should normally report all broken links,
there are a few cases where broken links are harmless. One case
is when a the PPC slice of a program links to a PPC-only library
that no longer exists on x86_64 OS X. Since the PPC slice is never
loaded on modern OS X, inconsistencies within it do not need to
be reported.

This is a fix for https://github.com/Homebrew/homebrew-core/pull/14753,
as well as everything here: https://github.com/Homebrew/homebrew-core/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aclosed%20libgcc_s_ppc64.1.dylib

-----

### DIFF
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -36,6 +36,7 @@ class LinkageChecker
           rescue NotAKegError
             @system_dylibs << dylib
           rescue Errno::ENOENT
+            next if harmless_broken_link?(dylib)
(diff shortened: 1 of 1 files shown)


woodruffw:
N.B. I made this into a method because we might identify more cases in the future, especially as Apple begins to move away from x86-only support.

MikeMcQuaid:
Great change, thanks @woodruffw!

PR #2819, opened 2017-06-24
🗑️ Use AppleScript instead of Swift for trashing.

reitermarkus:


Library/Homebrew/cask/lib/hbc/artifact/uninstall_base.rb
```diff
@@ -226,12 +228,35 @@ def uninstall_delete(*paths)
       def uninstall_trash(*paths)
         return if paths.empty?
 
+        return `say No trash for you!` if Utils.current_user == "ilovezfs"
```
ilovezfs: LOL. :rofl:

dunn: 👍

ilovezfs:
>No easter egg for you!

:sob:

Library/Homebrew/cask/lib/hbc/artifact/uninstall_base.rb
```diff
@@ -226,12 +228,33 @@ def uninstall_delete(*paths)
...
+
+              repeat with i from 1 to (count trashedItems)
+                set item i of trashedItems to POSIX path of (item i of trashedItems as string)
+                set output to output & (item i of trashedItems) & (do shell script "printf \"\\0\"")
+              end repeat
+
+              return output
+            end tell
+          end run
+        EOS
```
alyssais: Could we loop in Ruby instead to keep AppleScript to a minimum?

reitermarkus: No, otherwise it will make the trashing sound for every single file.

ilovezfs: That could be considered a feature.

vitorgalvao: Also note that in addition to having the trashing sound happen multiple times (which can be funny: it’s so fast it sounds like a paper machine gun), the Finder’s undo would only work for one file at a time (and to a limit). This way, doing undo once will recover everything to their right places.

alyssais: Awesome that it hooks into undo!

PR #2820, opened 2017-06-24
[WIP] language/node: fix prepublish scripts

chrmoritz:

-----

Refs: https://github.com/Homebrew/homebrew-core/pull/14827

This PR adds support for installing node modules with a prepublish step, which requires the execution of some of it's (dev)dependencies. This is required for adding the `now-cli` formula to homebrew-core (see: https://github.com/Homebrew/homebrew-core/pull/14827).

The current implementation of `pack_for_installation` is based on the assumption that the module we want to pack is self-contained. Unfortunately this is not always the case and some modules require a prepublish step, which is executed before (=at the beginning of) every `npm pack` run (and it was even executed during our old pre npm@5 way of installing modules, but there we had all dependencies in place).
Unfortunately prepublish scripts are likely trying to execute at least some of their dev(dependencies), so we have to install them a second (or first?) time before executing `npm pack`. I did made this an opt-in, so that we don't have the overhead of installing all dependencies twice in the majority of formulas for node modules without such prepublish steps.

cc @ilovezfs

I was able to successfully install `now-cli` and `coffeescript` with this PR.

### DIFF
--- a/Library/Homebrew/language/node.rb
+++ b/Library/Homebrew/language/node.rb
@@ -1,3 +1,5 @@
+require "json"
+
 module Language
   module Node
     def self.npm_cache_config
@@ -5,11 +7,23 @@ module Language
     end
 
     def self.pack_for_installation(prepare_required: false)
-      # Some packages are requiring (dev)dependencies to be in place when
-      # running the prepublish script (which is run before pack). In this case
-      # we have to install all dependencies a first time already before
-      # executing npm pack (and a second time when doing the actual install).
-      safe_system "npm", "install", *local_npm_install_args if prepare_required
+      # Read https://gist.github.com/chrmoritz/34e4c4d7779d72b549e2fc41f77c365c
+      # for a complete overview of the edge cases this method has to handle.
+      if prepare_required
+        # Rewrites the package.json so that npm pack will bundle all deps
+        # into a self-contained package, so that we avoid installing them a
+        # second time during the final installation of the package to libexec.
+        pkg_json = JSON.parse(IO.read("package.json"))
+        if pkg_json["dependencies"]
+          pkg_json["bundledDependencies"] = pkg_json["dependencies"].keys
+          IO.write("package.json", JSON.pretty_generate(pkg_json))
+        end
+
+        # We have to already install all deps here before npm pack, because the
+        # prepare script could require them. This already executes the prepare
+        # script to, so that we can continue to npm pack with --ignore-scripts.
+        safe_system "npm", "install", *local_npm_install_args
+      end
 
       # Homebrew assumes the buildpath/testpath will always be disposable
       # and from npm 5.0.0 the logic changed so that when a directory is
--- a/Library/Homebrew/language/node.rb
+++ b/Library/Homebrew/language/node.rb
@@ -6,6 +6,16 @@ module Language
       "cache=#{HOMEBREW_CACHE}/npm_cache"
     end
 
+    def self.node_system(cmd, *args)
+      ohai "#{cmd} #{args * " "}"
+      if ARGV.verbose?
+        safe_system(cmd, *args)
+      else
+        quiet_system(cmd, *args)
+        raise ErrorDuringExecution.new(cmd, args) unless $CHILD_STATUS.exitstatus.zero?
+      end
+    end
+
     # Read https://gist.github.com/chrmoritz/34e4c4d7779d72b549e2fc41f77c365c
     # for a complete overview of the edge cases this method has to handle.
     def self.pack_for_installation(prepare_required: false)
@@ -27,7 +37,7 @@ module Language
       # so that we can continue to `npm pack` with `--ignore-scripts`.
       install_args = local_npm_install_args
       install_args << "--production" unless prepare_required
-      safe_system "npm", "install", *install_args
+      node_system "npm", "install", *install_args
 
       # Homebrew assumes the buildpath/testpath will always be disposable
       # and from npm 5.0.0 the logic changed so that when a directory is
--- a/Library/Homebrew/language/node.rb
+++ b/Library/Homebrew/language/node.rb
@@ -6,24 +6,28 @@ module Language
       "cache=#{HOMEBREW_CACHE}/npm_cache"
     end
 
+    # Read https://gist.github.com/chrmoritz/34e4c4d7779d72b549e2fc41f77c365c
+    # for a complete overview of the edge cases this method has to handle.
     def self.pack_for_installation(prepare_required: false)
-      # Read https://gist.github.com/chrmoritz/34e4c4d7779d72b549e2fc41f77c365c
-      # for a complete overview of the edge cases this method has to handle.
-      if prepare_required
-        # Rewrites the package.json so that npm pack will bundle all deps
-        # into a self-contained package, so that we avoid installing them a
-        # second time during the final installation of the package to libexec.
-        pkg_json = JSON.parse(IO.read("package.json"))
-        if pkg_json["dependencies"]
-          pkg_json["bundledDependencies"] = pkg_json["dependencies"].keys
-          IO.write("package.json", JSON.pretty_generate(pkg_json))
-        end
-
-        # We have to already install all deps here before npm pack, because the
-        # prepare script could require them. This already executes the prepare
-        # script to, so that we can continue to npm pack with --ignore-scripts.
-        safe_system "npm", "install", *local_npm_install_args
+      # Rewrites the package.json so that npm pack will bundle all deps
+      # into a self-contained package, so that we avoid installing them a
+      # second time during the final installation of the package to libexec.
+      pkg_json = JSON.parse(IO.read("package.json"))
+      if pkg_json["dependencies"]
+        pkg_json["bundledDependencies"] = pkg_json["dependencies"].keys
+        IO.write("package.json", JSON.pretty_generate(pkg_json))
       end
+
+      # If `prepare_required` is false we pass `--production` to install all
+      # production deps (no devDeps) for bundling them into the pack and also
+      # to prevent the toplevel prepare / prepublish script from being executed
+      # while still executing all lifecycle scripts for the deps.
+      # Otherwise we omit `--production` to install all deps (devDeps might be
+      # required in `prepare`). This already executes the prepare script too,
+      # so that we can continue to `npm pack` with `--ignore-scripts`.
+      install_args = local_npm_install_args
+      install_args << "--production" unless prepare_required
+      safe_system "npm", "install", *install_args
 
       # Homebrew assumes the buildpath/testpath will always be disposable
       # and from npm 5.0.0 the logic changed so that when a directory is
--- a/Library/Homebrew/language/node.rb
+++ b/Library/Homebrew/language/node.rb
@@ -4,7 +4,13 @@ module Language
       "cache=#{HOMEBREW_CACHE}/npm_cache"
     end
 
-    def self.pack_for_installation
+    def self.pack_for_installation(prepare_required: false)
+      # Some packages are requiring (dev)dependencies to be in place when
+      # running the prepublish script (which is run before pack). In this case
+      # we have to install all dependencies a first time already before
+      # executing npm pack (and a second time when doing the actual install).
+      safe_system "npm", "install", *local_npm_install_args if prepare_required
+
       # Homebrew assumes the buildpath/testpath will always be disposable
       # and from npm 5.0.0 the logic changed so that when a directory is
       # fed to `npm install` only symlinks are created linking back to that
@@ -30,13 +36,13 @@ module Language
       end
     end
 
-    def self.std_npm_install_args(libexec)
+    def self.std_npm_install_args(libexec, prepare_required: false)
       setup_npm_environment
       # tell npm to not install .brew_home by adding it to the .npmignore file
       # (or creating a new one if no .npmignore file already exists)
       open(".npmignore", "a") { |f| f.write("\n.brew_home\n") }
 
-      pack = pack_for_installation
+      pack = pack_for_installation(prepare_required: prepare_required)
 
       # npm install args for global style module format installed into libexec
       %W[
--- a/Library/Homebrew/test/language/node_spec.rb
+++ b/Library/Homebrew/test/language/node_spec.rb
@@ -28,7 +28,7 @@ describe Language::Node do
 
     it "raises error with non zero npm pack exitstatus" do
       allow(IO).to receive(:read).and_return("{}")
-      allow(Language::Node).to receive(:safe_system).with("npm", "install", "-ddd", "--build-from-source",
+      allow(Language::Node).to receive(:node_system).with("npm", "install", "-ddd", "--build-from-source",
         "--cache=#{HOMEBREW_CACHE}/npm_cache", "--production") { `true` }
       allow(Utils).to receive(:popen_read).with(npm_pack_cmd) { `false` }
       expect { subject.std_npm_install_args(npm_install_arg) }.to \
@@ -37,7 +37,7 @@ describe Language::Node do
 
     it "raises error with empty npm pack output" do
       allow(IO).to receive(:read).and_return("{}")
-      allow(Language::Node).to receive(:safe_system).with("npm", "install", "-ddd", "--build-from-source",
+      allow(Language::Node).to receive(:node_system).with("npm", "install", "-ddd", "--build-from-source",
         "--cache=#{HOMEBREW_CACHE}/npm_cache", "--production") { `true` }
       allow(Utils).to receive(:popen_read).with(npm_pack_cmd) { `false` }
       expect { subject.std_npm_install_args(npm_install_arg) }.to \
@@ -46,7 +46,7 @@ describe Language::Node do
 
     it "does not raise error with a zero npm pack exitstatus" do
       allow(IO).to receive(:read).and_return("{}")
-      allow(Language::Node).to receive(:safe_system).with("npm", "install", "-ddd", "--build-from-source",
+      allow(Language::Node).to receive(:node_system).with("npm", "install", "-ddd", "--build-from-source",
         "--cache=#{HOMEBREW_CACHE}/npm_cache", "--production") { `true` }
       allow(Utils).to receive(:popen_read).with(npm_pack_cmd) { `echo pack.tgz` }
       resp = subject.std_npm_install_args(npm_install_arg)
@@ -59,7 +59,7 @@ describe Language::Node do
                       '"bundledDependencies":["foo","bar"]}'
       expected_json = JSON.pretty_generate(JSON.parse(expected_json))
       allow(IO).to receive(:write).with("package.json", expected_json) { `true` }
-      allow(Language::Node).to receive(:safe_system)
+      allow(Language::Node).to receive(:node_system)
         .with("npm", "install", "-ddd", "--build-from-source", "--cache=#{HOMEBREW_CACHE}/npm_cache") { `true` }
       allow(Utils).to receive(:popen_read).with("npm pack --ignore-scripts") { `echo pack.tgz` }
       resp = subject.std_npm_install_args(npm_install_arg, prepare_required: true)
@@ -68,7 +68,7 @@ describe Language::Node do
 
     it "raises error with non zero local npm install exitstatus" do
       allow(IO).to receive(:read).and_return("{}")
-      allow(Language::Node).to receive(:safe_system).and_raise("Failure while executing: npm install -ddd")
+      allow(Language::Node).to receive(:node_system).and_raise("Failure while executing: npm install -ddd")
       expect { subject.std_npm_install_args(npm_install_arg, prepare_required: true) }.to \
         raise_error("Failure while executing: npm install -ddd")
     end
--- a/Library/Homebrew/test/language/node_spec.rb
+++ b/Library/Homebrew/test/language/node_spec.rb
@@ -26,19 +26,28 @@ describe Language::Node do
     npm_install_arg = "libexec"
     npm_pack_cmd = "npm pack --ignore-scripts"
 
-    it "raises error with non zero exitstatus" do
+    it "raises error with non zero npm pack exitstatus" do
+      allow(IO).to receive(:read).and_return("{}")
+      allow(Language::Node).to receive(:safe_system).with("npm", "install", "-ddd", "--build-from-source",
+        "--cache=#{HOMEBREW_CACHE}/npm_cache", "--production") { `true` }
       allow(Utils).to receive(:popen_read).with(npm_pack_cmd) { `false` }
       expect { subject.std_npm_install_args(npm_install_arg) }.to \
         raise_error("npm failed to pack #{Dir.pwd}")
     end
 
     it "raises error with empty npm pack output" do
-      allow(Utils).to receive(:popen_read).with(npm_pack_cmd) { `true` }
+      allow(IO).to receive(:read).and_return("{}")
+      allow(Language::Node).to receive(:safe_system).with("npm", "install", "-ddd", "--build-from-source",
+        "--cache=#{HOMEBREW_CACHE}/npm_cache", "--production") { `true` }
+      allow(Utils).to receive(:popen_read).with(npm_pack_cmd) { `false` }
       expect { subject.std_npm_install_args(npm_install_arg) }.to \
         raise_error("npm failed to pack #{Dir.pwd}")
     end
 
-    it "does not raise error with a zero exitstatus" do
+    it "does not raise error with a zero npm pack exitstatus" do
+      allow(IO).to receive(:read).and_return("{}")
+      allow(Language::Node).to receive(:safe_system).with("npm", "install", "-ddd", "--build-from-source",
+        "--cache=#{HOMEBREW_CACHE}/npm_cache", "--production") { `true` }
       allow(Utils).to receive(:popen_read).with(npm_pack_cmd) { `echo pack.tgz` }
       resp = subject.std_npm_install_args(npm_install_arg)
       expect(resp).to include("--prefix=#{npm_install_arg}", "#{Dir.pwd}/pack.tgz")
@@ -57,9 +66,8 @@ describe Language::Node do
       expect(resp).to include("--prefix=#{npm_install_arg}", "#{Dir.pwd}/pack.tgz")
     end
 
-    it "raises error with non zero exitstatus with prepare_required set" do
+    it "raises error with non zero local npm install exitstatus" do
       allow(IO).to receive(:read).and_return("{}")
-      allow(IO).to receive(:write).with("package.json", "{}") { `true` }
       allow(Language::Node).to receive(:safe_system).and_raise("Failure while executing: npm install -ddd")
       expect { subject.std_npm_install_args(npm_install_arg, prepare_required: true) }.to \
         raise_error("Failure while executing: npm install -ddd")
--- a/Library/Homebrew/test/language/node_spec.rb
+++ b/Library/Homebrew/test/language/node_spec.rb
@@ -42,6 +42,20 @@ describe Language::Node do
       resp = subject.std_npm_install_args(npm_install_arg)
       expect(resp).to include("--prefix=#{npm_install_arg}", "#{Dir.pwd}/pack.tgz")
     end
+
+    it "does not raise error with prepublish_required set" do
+      allow(Language::Node).to receive(:safe_system).with("npm", "install", "-ddd")
+      allow(Utils).to receive(:popen_read).with("npm pack").and_return("pack.tgz")
+      resp = subject.std_npm_install_args(npm_install_arg, prepublish_required: true)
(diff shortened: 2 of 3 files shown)


Library/Homebrew/language/node.rb
```diff
@@ -4,13 +4,21 @@ def self.npm_cache_config
...
     end
 
-    def self.pack_for_installation
+    def self.pack_for_installation(prepublishRequiresDeps)
+      # Some packages are requiring (dev)dependencies to be in place when
+      # running the prepublish script (which is run before pack). In this case
+      # we have to install all dependencies a first time already before
+      # executing npm pack (and a second time when doing the actual install).
+      if prepublishRequiresDeps
+        system "npm", "install" # ToDo: check exit status
```
chrmoritz: Can someone give me a hint about how I handle exit codes here? (looks like it doesn't raise automatically here like in formulas)

Library/Homebrew/language/node.rb
```diff
@@ -32,13 +40,13 @@ def self.setup_npm_environment
       end
     end
 
-    def self.std_npm_install_args(libexec)
+    def self.std_npm_install_args(libexec, prepublishRequiresDeps=false)
```
chrmoritz: Should this stay an explicit opt-in or do we want to do this unconditionally? Note that later would add a for most formula unnecessary overhead of installing all dependencies twice (before pack and during install).

Also any suggestions for a better variable name?

Library/Homebrew/language/node.rb
chrmoritz: Is `safe_system "npm", "install"` the recommended way?

Library/Homebrew/language/node.rb
```diff
@@ -4,15 +4,25 @@ def self.npm_cache_config
...
       # directory, consequently breaking that assumption. We require a tarball
       # because npm install creates a "real" installation when fed a tarball.
-      output = Utils.popen_read("npm pack").chomp
-      raise "npm failed to pack #{Dir.pwd}" unless $CHILD_STATUS.exitstatus.zero?
-      output
+      output = Utils.popen_read("npm pack")
+      unless $CHILD_STATUS.exitstatus.zero? and output.lines.length > 0
+        raise "npm failed to pack #{Dir.pwd}"
+      end
+      output.lines.last.chomp
```
chrmoritz: This change is needed because `npm pack` could contain verbose output from a prepublish script and we are only interested in the last line from npm itself containing the name of the created package.

Does someone with more knowledge in ruby has a suggestion how this could be improved?

chrmoritz:
Thinking about this further we may run into another edge case with prepublish scripts in the future (which is still no fixed in this current version of this PR):

Imagine a formula for a node module with a prepublish script, which does not support to be run twice (e.g. you need to run a `npm run clean` before running `prepublish a second time or it will fail because the build artifacts already exist). Now unlike `now-cli` it has a package on npm containing a copy of it with already build artifacts (prepublish is run before something is uploaded to the npm registry). Now if we use it inside the formula and run language/node on it our `npm pack` would run prepublish a second time on the already prebuild package, which would fails (see previous assumption). So we may need to add another parameter to `std_npm_install_args` which if set to true would pass an additional `--ignore-scripts` to `npm pack` to fix this use case.

Should I add support for this edge case too in this PR or do we want to wait until we need it in homebrew-core before supporting it?

Library/Homebrew/language/node.rb
```diff
@@ -32,13 +40,13 @@ def self.setup_npm_environment
       end
     end
 
-    def self.std_npm_install_args(libexec)
+    def self.std_npm_install_args(libexec, prepublish_requires_deps = false)
```
chrmoritz: Should this stay an explicit opt-in or do we want to do this unconditionally? Note that later would add a for most formula unnecessary overhead of installing all dependencies twice (before pack and during install).

Also any suggestions for a better variable name?

Library/Homebrew/language/node.rb
```diff
@@ -4,15 +4,23 @@ def self.npm_cache_config
...
       # directory, consequently breaking that assumption. We require a tarball
       # because npm install creates a "real" installation when fed a tarball.
-      output = Utils.popen_read("npm pack").chomp
-      raise "npm failed to pack #{Dir.pwd}" unless $CHILD_STATUS.exitstatus.zero?
-      output
+      output = Utils.popen_read("npm pack")
+      unless $CHILD_STATUS.exitstatus.zero? && !output.lines.empty?
+        raise "npm failed to pack #{Dir.pwd}"
+      end
+      output.lines.last.chomp
```
chrmoritz: This change is needed because npm pack could contain verbose output from a prepublish script and we are only interested in the last line from npm itself containing the name of the created package.

Does someone with more knowledge in ruby has a suggestion how this could be improved?

ilovezfs:
Thanks for this @chrmoritz!

I'm assuming these warnings are just an upstream issue they'll need to address at some point?

```
npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.

> now@7.0.2 prepublish /private/tmp/now-cli-20170624-10884-uanglh/now-cli-7.0.2
> in-install || (npm run webpack && cp /dev/null download/dist/now)

npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN babel-loader@7.0.0 requires a peer of webpack@2 but none was installed.
npm WARN ajv-keywords@2.1.0 requires a peer of ajv@>=5.0.0 but none was installed.
```

Library/Homebrew/language/node.rb
```diff
@@ -4,15 +4,27 @@ def self.npm_cache_config
...
       # directory, consequently breaking that assumption. We require a tarball
       # because npm install creates a "real" installation when fed a tarball.
-      output = Utils.popen_read("npm pack").chomp
-      raise "npm failed to pack #{Dir.pwd}" unless $CHILD_STATUS.exitstatus.zero?
-      output
+      pack_cmd = "npm pack"
+      # We don't want to run the prepublish script for some node modules in
+      # some formulae, so skip it by using --ignore-scripts.
+      pack_cmd += " --ignore-scripts" if pack_skip_prepublish
+      output = Utils.popen_read(pack_cmd)
```
ilovezfs: One thing I'd noticed while trying to get `now-cli` to work was that stderr for the pack command isn't getting printed, which made it a bit difficult to see what was going wrong. I wonder if there's a reasonable way to get stderr printed as well even though we need to capture stdout for the tarball name, or if we just need to continue to debug those situations by manually running `npm pack` within the `debrew` shell.

Library/Homebrew/language/node.rb
```diff
@@ -32,13 +44,13 @@ def self.setup_npm_environment
       end
     end
 
-    def self.std_npm_install_args(libexec)
+    def self.std_npm_install_args(libexec, prepublish_requires_deps = false, pack_skip_prepublish = false)
```
woodruffw: Minor nit, but let's shorten these arguments down and make them keywords:

```ruby
def self.std_npm_install_args(libexec, req_deps: false, skip_prepub: false)
```

(Or something along those lines).

docs/Node-for-Formula-Authors.md
```diff
@@ -72,6 +72,12 @@ This will install your Node module in npm's global module style with a custom pr
 bin.install_symlink Dir["#{libexec}/bin/*"]
 ```
 
+*Note:* If your module requires a (dev)dependencies to to present at npm prepublish time, you have to invoke `std_npm_install_args` with `prepublish_requires_deps` set to `true` like:
+
+```ruby
+system "npm", "install", *Language::Node.std_npm_install_args(libexec, prepublish_requires_deps = true)
```
woodruffw: You can just do `std_npm_install_args(libexec, true)` here, since it's a default parameter and not a keyword (unless it gets changed above, in which case it will become `req_deps: true`).

Library/Homebrew/language/node.rb
```diff
@@ -4,15 +4,27 @@ def self.npm_cache_config
       "cache=#{HOMEBREW_CACHE}/npm_cache\n"
     end
 
-    def self.pack_for_installation
+    def self.pack_for_installation(prepublish_requires_deps, pack_skip_prepublish)
```
MikeMcQuaid: I think they should be keyword or hash arguments; taking two booleans here means the method calls are a bit opaque.

Library/Homebrew/language/node.rb
```diff
@@ -4,15 +4,27 @@ def self.npm_cache_config
...
       # because npm install creates a "real" installation when fed a tarball.
-      output = Utils.popen_read("npm pack").chomp
-      raise "npm failed to pack #{Dir.pwd}" unless $CHILD_STATUS.exitstatus.zero?
-      output
+      pack_cmd = "npm pack"
+      # We don't want to run the prepublish script for some node modules in
+      # some formulae, so skip it by using --ignore-scripts.
+      pack_cmd += " --ignore-scripts" if pack_skip_prepublish
+      output = Utils.popen_read(pack_cmd)
+      unless $CHILD_STATUS.exitstatus.zero? && !output.lines.empty?
```
MikeMcQuaid: Use `if` if it's for two conditions.

Library/Homebrew/language/node.rb
MikeMcQuaid: I think longer arguments makes these much easier to read but agreed on keywords. `req_` generally means `requirement_` in Homebrew. This is why saving a few characters on variable names is generally a false economy as you need to add a comment to explain your translation.

Library/Homebrew/language/node.rb
chrmoritz: +1 I too had to manually debug `npm pack`, through I don't know enough ruby to fix this.

chrmoritz:
I've addressed some of the review comments in the latest commit.

**Note:** In the [current direction the now-cli is going](https://github.com/Homebrew/homebrew-core/pull/14827#issuecomment-310896994) it looks like running the republish scripts there is no longer needed (because we want to avoid bundling a null node executable into the binary and instead use our owb node formula at runtime for it). This doesn't mean there are no other possible formulas not submitted to core yet, which would benefit from the `prepublish_requires_deps`, through now-cli might only benefit from `pack_skip_prepublish` [as proposed here](https://github.com/Homebrew/homebrew-core/pull/14827#issuecomment-310897963).

* I've changed the arguments to keyword, but I'm still not totally happy with their naming either (as [original mention here](https://github.com/Homebrew/brew/pull/2820#discussion_r123887722)). Through I have no ideas for better names for them.
* I still don't know it it is possible (and if so how) to log out `npm pack` output and at the same time capture stdout to get the pack name from it. See: https://github.com/Homebrew/brew/pull/2820#discussion_r123890729

Library/Homebrew/language/node.rb
chrmoritz: Something like `requires_deps` would be strictly speaking incorrect, because the deps would be required in any case on install time. Having to set `requires_deps: false` to not install deps at pack/prepublish time, but have them installed in the install step would likely create more confusion than the longer names.

Library/Homebrew/language/node.rb
chrmoritz: Looks like
 ```
output = `#{pack_cmd}`
```
does the job.

File: Library/Homebrew/test/language/node_spec.rb [hunk 11]
```diff
@@ -30,17 +30,17 @@
...
 
-    it "does not raise error with a zero exitstatus" do
-      allow(Utils).to receive(:popen_read).with("npm pack").and_return("pack")
-      allow_any_instance_of(Process::Status).to receive(:exitstatus).and_return(0)
-      allow_any_instance_of(nil::NilClass).to receive(:exitstatus).and_return(0)
-      resp = subject.std_npm_install_args(npm_install_arg)
-      expect(resp).to include("--prefix=#{npm_install_arg}", "#{Dir.pwd}/pack")
-    end
+    #it "does not raise error with a zero exitstatus" do
+    #  allow(Utils).to receive(:popen_read).with("npm pack").and_return("pack")
```
chrmoritz: is it possible to do something similar for 
```
output = `#{pack_cmd}`
```
?

MikeMcQuaid:
> Note: In the current direction the now-cli is going it looks like running the republish scripts there is no longer needed (because we want to avoid bundling a null node executable into the binary and instead use our owb node formula at runtime for it). This doesn't mean there are no other possible formulas not submitted to core yet, which would benefit from the prepublish_requires_deps, through now-cli might only benefit from pack_skip_prepublish as proposed here.

I'm pretty 👎 on merging anything that doesn't have at least one formula in core that wants/needs to use it.

chrmoritz:
> I'm pretty 👎 on merging anything that doesn't have at least one formula in core that wants/needs to use it.

Thats what I wanted to know in the exact oposite case in https://github.com/Homebrew/brew/pull/2820#issuecomment-310872807 .

Nevertheless it would still need the `pack_skip_prepublish` part of this PR and the more log output part of it coudn't hurt anyway too.

MikeMcQuaid:
> Nevertheless it would still need the pack_skip_prepublish part of this PR and the more log output part of it coudn't hurt anyway too. Unneeded would only be the prepublish_requires_deps keyword and the part from line 12,

Yep 👍. Thanks @chrmoritz 🙇

chrmoritz:
The `(dev)Deps required in prepublish scripts` part of this PR is blocked by the ongoing discussion in the [`now-cli` PR](https://github.com/Homebrew/homebrew-core/pull/14827) and @MikeMcQuaid comment about not wanting to fix prepublish scripts in language/node before it is actual needed in core.

Because of this I've split the independent improvements from this PR into a separate PR in #2826 (and added some other small improvements which I had already on my radar to get landed over there). Please review this PR in the meantime.

MikeMcQuaid:
> Because of this I've split the improvements independent from this issue from this PR into a separate PR in #2826 (and added some other small improvements which I had already on my radar to get landed over there). Please review this PR in the meantime.

👍

chrmoritz:
I've rebased this on the latest version of #2826 and added 2 tests and a doc update to the `prepublish_required` commit https://github.com/Homebrew/brew/pull/2820/commits/ea6d624a8aa5f4bd21cc0e991aa0ad0d0f658e73.

MikeMcQuaid:
Code looks good here whenever we've got a formula to use it 👍

chrmoritz:
I've rebased this on master, did some cleanup to tests, docs and bundledDeps logic and renamed the option `prepare_required`: In the long term naming the option after the since npm@4 deprecated `prepublish` lifecycle script doesn't felt right, even if it's still commonly used in the npm ecosystem for legacy reasons.

> Since `npm@1.1.71`, the npm CLI has run the `prepublish` script for both `npm publish` and `npm install`, because it's a convenient way to prepare a package for use. It has also turned out to be, in practice, very confusing. As of `npm@4.0.0`, a new event has been introduced, `prepare`, that preserves this existing behavior. A new event, `prepublishOnly` has been added as a transitional strategy to allow users to avoid the confusing behavior of existing npm versions and only run on `npm publish`.

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,40 @@
...
+
 module Language
   module Node
     def self.npm_cache_config
       "cache=#{HOMEBREW_CACHE}/npm_cache"
     end
 
-    def self.pack_for_installation
+    def self.pack_for_installation(prepare_required: false)
+      # Doing this unconditional could break for already prepared npm registry
```
MikeMcQuaid:
```
Using `prepare_required` unconditionally could
```

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,40 @@
...
+      # tarballs, because our local npm install would unecessary call prepare /
+      # prepublish a second time on them which might fail (e.g. calling mkdir
+      # on an already existing folder).
+      # We also can't just always do the local npm install und bundledDeps into
```
MikeMcQuaid: `und` -> `and`

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,40 @@
...
+      # package trick and just pass --ignore-scripts to the local npm install
```
MikeMcQuaid:
```
`--ignore-scripts`
```

and generally do this for all referenced variable/parameter names and other flags.

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,40 @@
...
+        # script. Note that this local install already executes both lifecycle
+        # scripts if present to that we can / must pass --ignore-scripts to
+        # npm pack to prevent them from being executed a second time.
+        safe_system "npm", "install", *local_npm_install_args
+
+        # This adds the module's dependencies as bundledDependencies to the
+        # package.json and tells npm to bundle the dependencies into the
+        # package created during npm pack so that we avoid the overhead of
+        # installing all dependencies a second time during the final step of
+        # installation the package to libexecs.
```
MikeMcQuaid: Can this be more than a single sentence?

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,40 @@
...
+      # on an already existing folder).
+      # We also can't just always do the local npm install und bundledDeps into
+      # package trick and just pass --ignore-scripts to the local npm install
+      # unless prepare_required because doing so would ignore scripts for dependencies
+      # too and would break for example installing of native addons.
+      if prepare_required
+        # Install all the modules dependencies already here and not just at
+        # npm pack time, because they might be used by the prepublish / prepare
+        # script. Note that this local install already executes both lifecycle
+        # scripts if present to that we can / must pass --ignore-scripts to
```
MikeMcQuaid: `can` or `must`? I don't think they are interchangeable.

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,40 @@
...
+      # tarballs, because our local npm install would unecessary call prepare /
+      # prepublish a second time on them which might fail (e.g. calling mkdir
+      # on an already existing folder).
+      # We also can't just always do the local npm install und bundledDeps into
+      # package trick and just pass --ignore-scripts to the local npm install
+      # unless prepare_required because doing so would ignore scripts for dependencies
+      # too and would break for example installing of native addons.
+      if prepare_required
+        # Install all the modules dependencies already here and not just at
+        # npm pack time, because they might be used by the prepublish / prepare
```
MikeMcQuaid: Pick `prepublish` or `prepare` here.

chrmoritz:
I agree that the comments in this PR aren't ideal. But I think even the current amount of comments aren't enough, if anyone should ever be able to understand the quite complex reasoning behind the code to support every possible edge case with our current approve (regarding lifecycle scripts) by just reading the language/node sources.

Maybe it would be better to write a PR comment explaining this and just link to it from a comment, to don't blow up the sources to much?

Could we also agree on that we want to do the `bundledDependencies` `package.json` rewrite (as in the second commit) to avoid the overhead of installing deps twice. If so I could squash these commits.

chrmoritz:
We have to support **5 different cases with lifecycle scripts** (`prepare` and `prepublish`) in node module based formulas. These are sorted by increasing difficulty / complexity to support:

1. arbitrary download location with no `prepare` (or `prepublish`) script: trivial to handle
2. npm registry tarball download (already prepared) with a `prepare` script, which could be run a second time without failure: running `prepare` twice would be unnecessary but wouldn't break anything
3. npm registry tarball download (already prepared) with a `prepare` script, which fails when run a second time (e.g. mkdir fails on already existing directory): `--ignore-scripts` is required for `npm pack` *(and if we want this case to be rooted through the bundledDependencies branch for consistency we would need to delete the `prepare` and `prepublish` script from the `package.json` (because `--ignore-scripts` won't work here, see npm background#1))*
4. source tarball downloads (not already prepared), but we don't want to run `prepare` (e.g. formula implements it's own logic): same a 3.
5. source tarball downloads (not already prepared), we want `prepare` to be executed (`prepare_required` is set): we have to install all deps (because they could be required to execute the `prepare` script) before `npm pack` by doing a local npm installation. Because said installation already executes `prepare` (see npm background#2) we still have to pass `--ignore-scripts` to `npm pack` (because executing `prepare` twice would be unnecessary and could even fail in some cases (e.g mkdir))

Supported cases by `language/node` version:
* Old `npm@<=4` (node <=7) version: 1, 2, 5 (3 + 4 will fail in all `language/node` versions before #2826)
* current version: 1, 2, 3, 4
* this PR: 1, 2, 3, 4, 5 (all)

**npm background:**
1. passing `--ignore-scrips` to a command prevents all lifecycle scripts from being executed, even the lifecycle scripts from deps installed by this step (which is problematic if you want to use it to only prevent the `prepare` / `prepublish` script of the toplevel module from being executed, but accidentally are breaking e.g. the compilation of native addons in the deps with it)
2. `prepare` (and `npm@4+` deprecated `prepublish`) is run for `npm pack` (and `npm publish`). It is also run for local npm installation: `npm install` without argument to install all deps of a module according to `package.json` or `npm install <moduleName>` to install a module to the local `node_modules` folder. (It was also run in `npm@<4` for our old `npm install -g --prefix=<path> <folder>` method).
3. But `prepare` is NOT run for global npm registry installations (`npm install -g <moduleName>`) and installation from a local tarbal (`npm install -g --prefix=<path> <pathToPackedTarball>`), which we do now.
4. The new `prepublishOnly` doesn't affect us, because it's only run before `npm publish` and NOT even before `npm pack`
5. `postinstall` (also other `install` lifecycle scripts) is run AFTER any `npm install` command (local and global from any sources), but it shouldn't hurt us because npm registry tarballs are not already postinstalled and running it twice for our local install and the global prefix install wouldn't hurt, because its run AFTER install in the libexec directory at the 2nd run, which shouldn't conflict.

**different installations method**:
1. `npm pack` only toplevel module (with empty dep tree) and install `npm i -g --prefix=libexec` it to libexec prefix (current version): doesn't support `prepare` (`prepublish`) scripts at all, quite hacky workaround for the npm@5 symlinks issues, `npm pack` is not designed to be run on an empty dep tree, but seems to work as long as no lifecycle scripts are involved
2. install all dependencies beforehand (which executes `prepare`), pack everything in a self-contained package using `npm pack` with the `bundledDependencies` `package.json` rewrite and then install `npm i -g --prefix=libexec` it to libexec, without the need to handle deps in this last step (partly done in this PR for case 5): This matches exactly what the pre `npm@<=4` (node <=7) version of `language/node` had done in npm internally, is more npm supported (running `npm pack` on a complete dep tree) and it supports `prepare` scripts out of the box

<hr>

In theory we could remove the `if prepare_required` block from this PR and do 2. unconditionally to get the cleaner more npm supported way for `language/node` every time. Unfortunately this would bring `language/node` back to a state where case 3 + 4 are not supported (like before #2826), because npm registry tarballs where never meant to be extracted before installation, which breaks npms assumption that only tarballs are already prepared (and extracted stuff is not prepared). To fix this we would need again a flag like `prepare_required` to tell the first local `npm install` call wether to execute `prepare` or not. Unfortunately we can't just pass `--ignore-scripts` to it for this because of npm background#1 and would have to rewrite the `package.json` to remove `scripts.prepare` and `scripts.prepublish` `unless prepare_required` in addition to the `bundledDependencies` rewrite (which could be done in one go by moving the rewrite above the first local `npm install` call). Any opinion on wether we should do this?

chrmoritz:
> In theory we could remove the `if prepare_required` block from this PR and do 2. unconditionally to get the cleaner more npm supported way for `language/node` every time. Unfortunately this would bring `language/node` back to a state where case 3 + 4 are not supported (like before #2826), because npm registry tarballs where never meant to be extracted before installation, which breaks npms assumption that only tarballs are already prepared (and extracted stuff is not prepared). To fix this we would need again a flag like `prepare_required` to tell the first local `npm install` call wether to execute `prepare` or not by passing `--production` to it `unless prepare_required` (see npm background#1+4). Any opinion on wether we should do this?

Edit: Just found out that passing `--production` to a local `npm install` will not only prevent devDeps from being installed, but also prevents `prepare` script from being executed (because they might require devDeps for execution) while still executing (pre/post)install scripts during installing the deps. This makes being consistent with bundling all deps into the package (see method#2 above) much more easier while still supporting case 3+4). I've added a [3rd commit](https://github.com/Homebrew/brew/pull/2820/commits/2ae152144ebf6b1867a05124818868134cd1e0d6) with this approve to the PR (still needs updating / fixing the tests).

@MikeMcQuaid Any opinion which approve (represanted by a commit in this PR) you would prefere?

chrmoritz:
testing https://github.com/Homebrew/brew/pull/2820/commits/25b201eb9ac97e3a36ee33199eafc952db0833a3 always bundle deps into pack:

* [ ] prepare test formula
* [ ] alexjs
* [ ] angular-cli
* [ ] autocode
* [ ] azure-cli
* [ ] babel
* [ ] bower
* [ ] chronograf
* [ ] coffeescript
* [ ] diff-so-fancy
* [ ] firebase-cli
* [ ] generate-json-schema
* [ ] gitter-cli
* [ ] grunt-cli
* [ ] heroku
* [ ] insect
* [ ] ios-sim
* [ ] jhipster
* [ ] jsdoc3
* [ ] jsonlint
* [ ] nativefier
* [ ] svgo
* [ ] typescript
* [ ] webpack
* [ ] write-good

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,44 @@
+require "json"
+
 module Language
   module Node
     def self.npm_cache_config
       "cache=#{HOMEBREW_CACHE}/npm_cache"
     end
 
-    def self.pack_for_installation
+    def self.node_system(cmd, *args)
```
MikeMcQuaid: Would be good to use the formula system method instead of reimplementing it here.

Library/Homebrew/language/node.rb
```diff
@@ -1,10 +1,44 @@
...
+      end
+    end
+
+    # Read https://gist.github.com/chrmoritz/34e4c4d7779d72b549e2fc41f77c365c
+    # for a complete overview of the edge cases this method has to handle.
+    def self.pack_for_installation(prepare_required: false)
+      # Rewrites the package.json so that npm pack will bundle all deps
+      # into a self-contained package, so that we avoid installing them a
+      # second time during the final installation of the package to libexec.
+      pkg_json = JSON.parse(IO.read("package.json"))
```
MikeMcQuaid: May be worth producing a nice error message if this file doesn't exist?

Library/Homebrew/language/node.rb
chrmoritz: How to do so? Using just `system` inside a `language` will call [`Homebrew/utils.rb`system](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils.rb#L185) and not [`Homebrew/formula.rb` one](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L1720).

An alternative implementation of this using util's `_system`(basically mixing the implementation of `safe_system` and `quite_system) would be:
```ruby
def self.node_system(cmd, *args)
  ohai "#{cmd} #{args * " "}"
  Homebrew._system(cmd, *args) do
    unless ARGV.verbose?
      $stdout.reopen("/dev/null")
      $stderr.reopen("/dev/null")
    end
  end
  raise ErrorDuringExecution.new(cmd, args) unless $CHILD_STATUS.success?
end

``

Library/Homebrew/language/node.rb
chrmoritz: What error message do you have in mind here?

A `package.json` should always exist unless someone is calling `language/node` from a directory, which is not a valid node module. The `Error: No such file or directory - package.json` from IO is nearly identical to the error from npm (`ENOENT: no such file or directory, open '/absolute/buildpath/to/package.json'`) someone would get with the current version of `language/node` when calling it on a non node module folder.

chrmoritz: Replacing the error msg with our own custom one would also hide the actual kind of the IO error: `package.json` doesn't exist (see above), permission error (no read/write permissions) ...
Just doing nothing here and throw the original IO error seems to be the best for debugging.

Library/Homebrew/language/node.rb
chrmoritz: And speaking about `formula`: Is it possible to access the `formula`'s download `url` inside `language/node`? With it we could match it against a npm registry url regex and figure the correct way to handle life cycle scripts out on our own without requiring the formula author to set the `prepare_require` parameter.

MikeMcQuaid: The formula object can be passed through as an argument or this module could be included inside the formula itself. That'll handle both of these cases.

Library/Homebrew/language/node.rb
MikeMcQuaid: I was thinking any error message that's not a raw exception but you've convinced me this is fine as-is.

MikeMcQuaid:
@chrmoritz What's the status of this?

MikeMcQuaid:
@chrmoritz Closing for now but happy to reopen when you let us know the status.

PR #2825, opened 2017-06-26
Refactoring using `Forwardable` and `DelegateClass`.

reitermarkus:

-----

Simplify some methods and classes, as well as add an `attr_predicate` helper.

### DIFF
--- a/Library/Homebrew/extend/predicable.rb
+++ b/Library/Homebrew/extend/predicable.rb
@@ -1,0 +1,9 @@
+module Predicable
+  def attr_predicate(*attrs)
+    attrs.each do |attr|
+      define_method attr do
+        instance_variable_get("@#{attr.to_s.sub(/\?$/, "")}") == true
+      end
+    end
+  end
+end
--- a/Library/Homebrew/cask/lib/hbc/artifact/base.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/base.rb
@@ -1,6 +1,8 @@
 module Hbc
   module Artifact
     class Base
+      extend Predicable
+
       def self.artifact_name
         @artifact_name ||= name.sub(/^.*:/, "").gsub(/(.)([A-Z])/, '\1_\2').downcase
       end
@@ -65,13 +67,7 @@ module Hbc
         {}
       end
 
-      def verbose?
-        @verbose
-      end
-
-      def force?
-        @force
-      end
+      attr_predicate :force?, :verbose?
 
       def initialize(cask, command: SystemCommand, force: false, verbose: false)
         @cask = cask
--- a/Library/Homebrew/cask/lib/hbc/cask.rb
+++ b/Library/Homebrew/cask/lib/hbc/cask.rb
@@ -1,5 +1,3 @@
-require "forwardable"
-
 require "hbc/dsl"
 require "hbc/metadata"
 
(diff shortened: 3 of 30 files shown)


Library/Homebrew/brew.sh
```diff
@@ -345,5 +345,5 @@ else
 
   # Unshift command back into argument list (unless argument list was empty).
   [[ "$HOMEBREW_ARG_COUNT" -gt 0 ]] && set -- "$HOMEBREW_COMMAND" "$@"
-  { update-preinstall; exec "$HOMEBREW_RUBY_PATH" -W0 "$HOMEBREW_LIBRARY/Homebrew/brew.rb" "$@"; }
+  { update-preinstall; exec "$HOMEBREW_RUBY_PATH" "$HOMEBREW_LIBRARY/Homebrew/brew.rb" "$@"; }
```
MikeMcQuaid: What's this change for?

Library/Homebrew/extend/refinements.rb
```diff
@@ -0,0 +1,20 @@
+# Disable refinement warning.
```
MikeMcQuaid: Why is there a warning?

Library/Homebrew/extend/refinements.rb
```diff
@@ -0,0 +1,20 @@
...
+  old_verbose = $VERBOSE
+  $VERBOSE = nil
+  using Module.new
+ensure
+  $VERBOSE = old_verbose
+end
+
+module Refinements
+  refine Module do
+    def attr_predicate(*attrs)
```
MikeMcQuaid: What's the advantage of this over e.g. including a module?

alyssais: I'm also not sure a refinement is necessary in this case.

reitermarkus: Well, first I added this to the `extend/module` monkey-patch. Then I thought, let's not add another monkey-patch, let's try refinements! Something as primitive as using a simple Module didn't even cross my mind. 😄

Library/Homebrew/extend/predicable.rb
```diff
@@ -0,0 +1,9 @@
+module Predicable
```
alyssais: Could this name be more descriptive?

Library/Homebrew/extend/predicable.rb
```diff
@@ -0,0 +1,9 @@
...
+  def attr_predicate(*attrs)
+    attrs.each do |attr|
+      define_method attr do
+        instance_variable_get("@#{attr.to_s.sub(/\?$/, "")}") == true
```
alyssais: I think this should take names without the `?` so it's consistent with `attr_writer`.

reitermarkus: I made it this way so it‘s easier to grep for predicates.

Library/Homebrew/extend/predicable.rb
reitermarkus: Suggestions? I think it fits in nicely with other `*able`s.

MikeMcQuaid: `AttrPredicate` 😉

Library/Homebrew/extend/predicable.rb
alyssais: Consistency with Ruby is more important, IMO, but it looks like I'm in the minority.

PR #2827, opened 2017-06-26
version: improve url-only version parsing

JCount:

-----
add support for both styles prefixed with up to two characters,
and styles with four groups of digits, seperated by periods

I believe I made the regex sufficiently strict to prevent unwanted results, but if necessary it could be split into two or more pieces. Some of the regex syntax itself is also more esoteric so it may be desirable to modify them for that reason as well.

Originally begun as a fix for [homebrew/core#14957](https://github.com/Homebrew/homebrew-core/pull/14957#discussion_r124086480).

I added tests, which is good, but did I add too many?

cc @ilovezfs @dunn

### DIFF
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -407,8 +407,13 @@ class Version
 
     # e.g. http://mirrors.jenkins-ci.org/war/1.486/jenkins.war
     # e.g. https://github.com/foo/bar/releases/download/0.10.11/bar.phar
-    m = %r{/(\d\.\d+(\.\d+)?)}.match(spec_s)
-    return m.captures.first unless m.nil?
+    # e.g. https://github.com/clojure/clojurescript/releases/download/r1.9.293/cljs.jar
+    # e.g. https://github.com/fibjs/fibjs/releases/download/v0.6.1/fullsrc.zip
+    # e.g. https://wwwlehre.dhbw-stuttgart.de/~sschulz/WORK/E_DOWNLOAD/V_1.9/E.tgz
+    # e.g. https://github.com/JustArchi/ArchiSteamFarm/releases/download/2.3.2.0/ASF.zip
+    # e.g. https://people.gnome.org/~newren/eg/download/1.7.5.2/eg
+    m = %r{/([rvV]_?)?(\d\.\d+(\.\d+){,2})}.match(spec_s)
+    return m.captures[1] unless m.nil?
 
     # e.g. http://www.ijg.org/files/jpegsrc.v8d.tar.gz
     m = /\.v(\d+[a-z]?)/.match(stem)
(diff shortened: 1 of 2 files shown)


Library/Homebrew/version.rb
```diff
@@ -407,7 +407,12 @@ def self._parse(spec)
 
     # e.g. http://mirrors.jenkins-ci.org/war/1.486/jenkins.war
     # e.g. https://github.com/foo/bar/releases/download/0.10.11/bar.phar
-    m = %r{/(\d\.\d+(\.\d+)?)}.match(spec_s)
+    # e.g. https://github.com/clojure/clojurescript/releases/download/r1.9.293/cljs.jar
+    # e.g. https://github.com/fibjs/fibjs/releases/download/v0.6.1/fullsrc.zip
+    # e.g. https://wwwlehre.dhbw-stuttgart.de/~sschulz/WORK/E_DOWNLOAD/V_1.9/E.tgz
+    # e.g. https://github.com/JustArchi/ArchiSteamFarm/releases/download/2.3.2.0/ASF.zip
+    # e.g. https://people.gnome.org/~newren/eg/download/1.7.5.2/eg
+    m = %r{/[a-z_]{,2}(\d\.\d+(\.\d+){,2})}i.match(spec_s)
```
MikeMcQuaid: What's the `[a-z_]{,2}` and (\.\d+){,2} handling?

MikeMcQuaid:
Nice work @JCount. One question on this.

Library/Homebrew/version.rb
JCount: @MikeMcQuaid My attempted explanation follows:
- The `[a-z_]{,2}`, combined with the case-insensitive modifier, will handle cases where the version substring in the url is prefixed by a combination of zero to two letters or underscores. This will handle, from the examples I've given, `/releases/download/r1.9.293/cljs.jar`, `/download/v0.6.1/fullsrc.zip`, and `/WORK/E_DOWNLOAD/V_1.9/E.tgz`.

- `(\.\d+){,2}` will handle instances where there are, in addition to the two, `.` separated base groups of digits, an additional one to two `.` prefixed digit groupings concatenated onto that base, for a total of up to four groupings of digits separated by `.`s. This will handle, for example, `/download/2.3.2.0/ASF.zip` and `/download/1.7.5.2/eg`, while not breaking the new handling of `/WORK/E_DOWNLOAD/V_1.9/E.tgz`.

I apologize for the awkward language, it is difficult to turn regex into understandable language, 🤔.

MikeMcQuaid: > will handle cases where the version substring in the url is prefixed by a combination of zero to two letters or underscores

I'd make this more specific e.g. `[rvV]_?`



> will handle instances where there are, in addition to the two

This one makes sense 👍

JCount:
@MikeMcQuaid Rebased

JCount:
@MikeMcQuaid Rebased again so all the regex changes are in one commit to make the commit history  a bit less fragmented.

PR #2829, opened 2017-06-27
Fix `bump-formula-pr` when working with a shallow git clone

mislav:

-----

Creating a fork and pushing to it doesn't work when the repository in question is a shallow clone. By default, Homebrew clones all taps in shallow mode unless `--full` was passed or $HOMEBREW_DEVELOPER was set.

/cc @mikemcquaid

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -286,8 +286,11 @@ module Homebrew
 
     formula.path.parent.cd do
       branch = "#{formula.name}-#{new_formula_version}"
+      git_dir = Utils.popen_read("git rev-parse --git-dir")
+      shallow = !git_dir.empty? && File.exist?("#{git_dir}/shallow")
+
       if ARGV.dry_run?
-        ohai "git fetch --unshallow origin"
+        ohai "git fetch --unshallow origin" if shallow
         ohai "git checkout --no-track -b #{branch} origin/master"
         ohai "git commit --no-edit --verbose --message='#{formula.name} #{new_formula_version}#{devel_message}' -- #{formula.path}"
         ohai "hub fork --no-remote"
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -296,6 +297,7 @@ def bump_formula_pr
         ohai "hub pull-request --browse -m '#{formula.name} #{new_formula_version}#{devel_message}'"
         ohai "git checkout -"
       else
+        safe_system "git", "fetch", "--unshallow", "origin"
```
mislav: Optionally, we chould conditionally perform `--unshallow` only if the git repository was shallow to begin with. However, since `brew update` ran just prior to this, this should return fast if there's nothing to do. Let me know what you think is best.

ilovezfs:
```
==> replace /https:\/\/github\.com\/fsprojects\/Paket\/releases\/download\/5\.2\.1\/paket\.exe/ with "https://github.com/fsprojects/Paket/releases/download/5.2.2/paket.exe"
==> replace "c59ea7abc167c3553eb36d7a140ea4218309368263d09b869b5225233eb17a28" with "26676be1f0c3a75fe4aa450f619863604262d9f5d9e063bac433945c907783ca"
fatal: --unshallow on a complete repository does not make sense
Error: Failure while executing: git fetch --unshallow origin
```

You may want to use `brew tap --full` for this.

mislav:
@ilovezfs Thanks for trying this out. Now changed to conditionally unshallow repos only when necessary.

MikeMcQuaid:
Ugh, CI failing on RuboCop stuff I should probably just disable at this point.

MikeMcQuaid:
Thanks @mislav! Will let @ilovezfs merge if he's 👍

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -286,7 +286,11 @@ def bump_formula_pr
 
     formula.path.parent.cd do
       branch = "#{formula.name}-#{new_formula_version}"
+      git_dir = Utils.popen_read("git rev-parse --git-dir")
```
ilovezfs: This needs `.chomp`

mislav: Good catch! Fixed

ilovezfs:
Thanks @mislav!

PR #2837, opened 2017-06-29
audit: ask for full_index patches

mistydemeo:

-----

This configures audit to recommend the new `?full_index=1` option for patch URLs. This option uses unabbreviated git refs, which aren't subject to change like abbreviated refs are. (See https://github.com/Homebrew/homebrew-core/pull/13792#issuecomment-302973700 for an example of large numbers of patches changing at once for this reason.)

This will cause `audit` to start flagging lots of existing formulae. Do we want to mass-migrate these in batches?

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -819,6 +819,13 @@ class FormulaAuditor
 
   def patch_problems(patch)
     case patch.url
+    when %r{https?://github\.com/.+/.+/(?:commit|pull)/[a-fA-F0-9]*.(?:patch|diff)}
+      unless patch.url =~ /\?full_index=\w+$/
+        problem <<-EOS.undent
+          GitHub patches should use the full_index parameter:
+            #{patch.url}?full_index=1
+        EOS
+      end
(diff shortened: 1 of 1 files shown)


ilovezfs:
>This will cause audit to start flagging lots of existing formulae. Do we want to mass-migrate these in batches?

I think we should migrate all existing patches before merging this.

mistydemeo:
> I think we should migrate all existing patches before merging this.

👍

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -819,6 +819,13 @@ def audit_revision_and_version_scheme
 
   def patch_problems(patch)
     case patch.url
+    when %r{https?://github\.com/.+/.+/(?:commit|pull)/[a-fA-F0-9]*.(?:patch|diff)}
+      unless patch.url =~ /\?full_index=\w+$/
```
ilovezfs: why `\w+`?

mistydemeo: I'm being overly pedantic by matching any param, not just a literal `1`. I guess it could just restrict it to that, heh.

ilovezfs: I'm sure this is fine hehe

PR #2841, opened 2017-06-30
Define a custom assertion failure error for cross-version compat

### LINKED ISSUE #2840: `brew test` fails on 10.13 due to different assertion class
- [x] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

10.13 uses Ruby 2.3 as its system Ruby, with some differences in how Test::Unit's assertions work. This causes some troubles with `brew test`, which uses the system Ruby's Test::Unit.

Specifically, the assertions in the system Ruby 2.0 on older OSs raise a `MiniTest::Assertion` on assertion failure, while Ruby 2.3 in 10.13 raises a `Test::Unit::AssertionFailedError`. Since the `MiniTest` constant isn't loaded at all, `brew test` fails with an `uninitialized constant`; even if it were these exceptions can't be rescued via `MiniTest::Assertion`.

Steps to reproduce:

1. Find a formula with a failing assertion (creating one if necessary).
2. Run `brew test formulaname`

Expected:

The failed test is reported as expected.

Actual:

A `Error: uninitialized constant Homebrew::MiniTest` error is printed.

cc @MikeMcQuaid

mistydemeo:

-----

Fixes #2840.

Note that this partially reverts c45cca8e36401f6d95d851befef35879a5629182 - it was unnecessary on Ruby 2.0, but turns out to be necessary for compatibility with the system Ruby 2.3 on OS X 10.13.

### DIFF
--- a/Library/Homebrew/dev-cmd/test.rb
+++ b/Library/Homebrew/dev-cmd/test.rb
@@ -83,7 +83,7 @@ module Homebrew
             exec(*args)
           end
         end
-      rescue MiniTest::Assertion => e
+      rescue Homebrew::Assertions::AssertionFailed => e
         ofail "#{f.full_name}: failed"
         puts e.message
       rescue Exception => e
--- a/Library/Homebrew/dev-cmd/test.rb
+++ b/Library/Homebrew/dev-cmd/test.rb
@@ -83,7 +83,7 @@ module Homebrew
             exec(*args)
           end
         end
-      rescue Homebrew::Assertions::AssertionFailed => e
+      rescue ::Test::Unit::AssertionFailedError => e
         ofail "#{f.full_name}: failed"
         puts e.message
       rescue Exception => e
--- a/Library/Homebrew/formula_assertions.rb
+++ b/Library/Homebrew/formula_assertions.rb
@@ -3,6 +3,19 @@ module Homebrew
     require "test/unit/assertions"
     include ::Test::Unit::Assertions
 
+    # Custom name here for cross-version compatibility.
+    # In Ruby 2.0, Test::Unit::Assertions raise a MiniTest::Assertion,
+    # but they raise Test::Unit::AssertionFailedError in 2.3.
+    # If neither is defined, this might be a completely different
+    # version of Ruby.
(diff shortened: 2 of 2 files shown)


MikeMcQuaid:
> Note that this partially reverts c45cca8 - it was unnecessary on Ruby 2.0, but turns out to be necessary for compatibility with the system Ruby 2.3 on OS X 10.13.

I was happy to leave that code as-is until it broke `brew tests` under Bundler 1.15.1. Can you ensure that's not the case now?

My main issue with the previous code was the lack of documentation: it was unclear in what situation any of those `if`s would occur. I'd like to see that addressed if we merge this.

I think the ideal situation, though, would be using the same module names regardless of the version. This should be possible for `brew tests` at least by ensuring it's using the same version of e.g. `minitest` everywhere. It may not be possible on Ruby 2.3 for formulae so this may be acceptable as a short-term workaround but before High Sierra is released I think I'd like to see us upgrade to a vendored Ruby 2.3.X (matching the High Sierra version). Differing Ruby versions was an absolute nightmare between 1.8 and 2.0 and now we've got a solution I'd love to see us lean into it.

mistydemeo:
> it was unclear in what situation any of those ifs would occur. I'd like to see that addressed if we merge this.

Will do!

> before High Sierra is released I think I'd like to see us upgrade to a vendored Ruby 2.3.X (matching the High Sierra version). 

The static build for Ruby 2.3 is currently broken: https://bugs.ruby-lang.org/issues/13413 I'm going to take a stab at it, but it's a bit arcane.

ilovezfs:
>The static build for Ruby 2.3 is currently broken: https://bugs.ruby-lang.org/issues/13413 I'm going to take a stab at it, but it's a bit arcane.

I wonder if upstream would be willing to add static builds to their CI so that this doesn't happen again, assuming it does get fixed at some point.

MikeMcQuaid:
Thanks @mistydemeo!

PR #2842, opened 2017-06-30
audit: Don't run audit methods when `--only-cops` option is passed

GauthamGoli:

-----
Previously `--only-cops` in addition to the passed cops, would also run audit methods defined in `FormulaAuditor`. This PR would change that to run only the cops passed through `--only-cops` and `audit_style` audit method to report style offenses.

Also while testing I discovered that I forgot to commit `.rubocop.yml` and `rubocops.rb` in #2790 
So adding those files here too so `LegacyPatches` cop would be enabled.


Library/Homebrew/cmd/style.rb
```diff
@@ -94,7 +94,12 @@ def check_style_impl(files, output_type, options = {})
           RuboCop::Cop::Cop.registry.departments.include?(cop.to_sym)
       end
 
-      args << "--only" << cops_to_include.join(",") unless cops_to_include.empty?
+      if cops_to_include.empty?
+        odie "Cop(s): #{options[:only_cops].join(",")} were either not found/enabled"
```
MikeMcQuaid: `RuboCops #{options[:only_cops].join(",")} were not found`

Library/Homebrew/cmd/style.rb
```diff
@@ -94,7 +94,12 @@ def check_style_impl(files, output_type, options = {})
...
+      else
```
MikeMcQuaid: don't need the `else` as `odie` will leave the function.

GauthamGoli:
`brew audit --only-cops=Desc` 
`brew audit --only-cops=ComponentsOrder`
Commands like these used to run other non cop checks too so the command used to take a lot of time. Now the command just runs the cop and no other check.

MikeMcQuaid:
Thanks again @GauthamGoli!

JCount:
@GauthamGoli 👍

PR #2846, opened 2017-07-01
Explain how to reduce auto-updates if they're slow

benkuhn:
I'm in Ethiopia on mobile data most of the time, and my internet can get down to 50kbps with 2+-second pings. Even under the best conditions, a no-op `brew update --preinstall` takes 2.5+ seconds. It would have improved my life a lot if I'd learned about `HOMEBREW_AUTO_UPDATE_SECS` earlier :)

(Also, right now if you Google for "homebrew disable auto update" it takes some poking around to learn how to reduce the frequency instead of the less-safe option of completely disabling it--hopefully this will make that more discoverable.)

- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). _no--not sure how to test brew.sh, sorry!_

-----

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -284,6 +284,8 @@ setup-analytics
 update-preinstall-timer() {
   sleep 3
   echo 'Updating Homebrew...' >&2
+  sleep 3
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Having this in `brew man` is sufficient for us for now, thanks for the suggestion though.

PR #2848, opened 2017-07-01
Refactor SVN and cURL download strategies.

reitermarkus:

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/download_strategy.rb
+++ b/Library/Homebrew/cask/lib/hbc/download_strategy.rb
@@ -90,13 +90,8 @@ module Hbc
       end
     end
 
-    def downloaded_size
-      temporary_path.size? || 0
-    end
-
     def _fetch
-      odebug "Calling curl with args #{cask_curl_args}"
-      curl(*cask_curl_args)
+      curl_download url, *cask_curl_args, to: temporary_path, user_agent: uri_object.user_agent
     end
 
     def fetch
@@ -131,23 +126,7 @@ module Hbc
     private
 
     def cask_curl_args
-      default_curl_args.tap do |args|
-        args.concat(user_agent_args)
-        args.concat(cookies_args)
-        args.concat(referer_args)
-      end
-    end
-
-    def default_curl_args
-      [url, "-C", downloaded_size, "-o", temporary_path]
-    end
-
-    def user_agent_args
-      if uri_object.user_agent
-        ["-A", uri_object.user_agent]
-      else
-        []
-      end
+      cookies_args + referer_args
     end
 
     def cookies_args
@@ -181,8 +160,7 @@ module Hbc
 
   class CurlPostDownloadStrategy < CurlDownloadStrategy
     def cask_curl_args
-      super
-      default_curl_args.concat(post_args)
+      super.concat(post_args)
     end
 
     def post_args
--- a/Library/Homebrew/cask/lib/hbc/download_strategy.rb
+++ b/Library/Homebrew/cask/lib/hbc/download_strategy.rb
@@ -10,7 +10,7 @@ module Hbc
   class AbstractDownloadStrategy
     attr_reader :cask, :name, :url, :uri_object, :version
 
-    def initialize(cask, command = SystemCommand)
+    def initialize(cask, command: SystemCommand)
       @cask       = cask
       @command    = command
       # TODO: this excess of attributes is a function of integrating
@@ -33,8 +33,8 @@ module Hbc
   class HbVCSDownloadStrategy < AbstractDownloadStrategy
     REF_TYPES = [:branch, :revision, :revisions, :tag].freeze
 
-    def initialize(cask, command = SystemCommand)
-      super
+    def initialize(*args, **options)
+      super(*args, **options)
       @ref_type, @ref = extract_ref
       @clone = Hbc.cache.join(cache_filename)
     end
@@ -64,11 +64,6 @@ module Hbc
   end
 
   class CurlDownloadStrategy < AbstractDownloadStrategy
-    # TODO: should be part of url object
-    def mirrors
-      @mirrors ||= []
-    end
-
     def tarball_path
       @tarball_path ||= Hbc.cache.join("#{name}--#{version}#{ext}")
     end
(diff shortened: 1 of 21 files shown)


Library/Homebrew/utils/curl.rb
```diff
@@ -1,42 +1,38 @@
...
+
+def curl_args(*extra_args, show_output: false, user_agent: HOMEBREW_USER_AGENT_CURL)
+  args = [curl_executable.to_s]
 
-  args = [
-    curl.to_s,
-    "--remote-time",
-    "--location",
-  ]
+  args << "--remote-time" unless extra_args.each_cons(2).include?(["--output", "/dev/null"])
```
MikeMcQuaid: Think I'd rather this was a specific argument

Library/Homebrew/utils/curl.rb
```diff
@@ -1,42 +1,38 @@
...
 
-  case options[:user_agent]
-  when :browser
+  if user_agent == :browser
     args << "--user-agent" << HOMEBREW_USER_AGENT_FAKE_SAFARI
   else
-    args << "--user-agent" << HOMEBREW_USER_AGENT_CURL
+    args << "--user-agent" << user_agent
```
MikeMcQuaid: When is this overridden?

Library/Homebrew/utils/curl.rb
```diff
@@ -1,42 +1,38 @@
...
   end
 
-  args += options[:extra_args] if options[:extra_args]
-  args
+  args + extra_args
 end
 
 def curl(*args)
-  safe_system(*curl_args(extra_args: args))
+  safe_system(*curl_args("--location", *args))
```
MikeMcQuaid: Why is `--location` used here and not elsewhere?

Library/Homebrew/utils/curl.rb
```diff
@@ -1,42 +1,38 @@
...
 end
 
 def curl_output(*args)
```
MikeMcQuaid: Why has `Open3.capture3` replaced calls of this function?

Library/Homebrew/cask/lib/hbc/download_strategy.rb
```diff
@@ -64,11 +64,6 @@ def clear_cache
   end
 
   class CurlDownloadStrategy < AbstractDownloadStrategy
```
MikeMcQuaid: It would be good to try and unify this code with Homebrew's considering it was a fork in the past.

Library/Homebrew/utils/curl.rb
reitermarkus: `--location` was previously added inside `curl_args`, but some commands only work without `--location`.

Library/Homebrew/cask/lib/hbc/download_strategy.rb
reitermarkus: I agree, but this is a whole nother project.

Library/Homebrew/utils/curl.rb
MikeMcQuaid: Ok but it's not clear from which which commands and why it's present on `curl` and not `curl_args`.

Library/Homebrew/cask/lib/hbc/download_strategy.rb
MikeMcQuaid: Cool 👍

Library/Homebrew/download_strategy.rb
```diff
@@ -375,35 +375,35 @@ def _fetch
...
-    urls = []
-    curl_args = _curl_opts << "-I" << "-L" << url
-    Utils.popen_read("curl", *curl_args).scan(/^Location: (.+)$/).map do |m|
-      urls << URI.join(urls.last || url, m.first.chomp).to_s
+    return ["--user" << meta.fetch(:user)] if meta.key?(:user)
+    []
+  end
+
+  def resolved_url(url)
+    redirect_url, _, status = Open3.capture3(*curl_args(
```
MikeMcQuaid: This is an example of where I'm not sure why one of the `curl` helper methods isn't used.

reitermarkus:
Added a new helper function, `curl_download`, for those calls which should always use `--location` and `--remote-time`.

Library/Homebrew/utils/curl.rb
```diff
@@ -1,42 +1,46 @@
...
+  raise "#{curl} is not executable"
+end
 
+def curl_args(*extra_args, show_output: false, user_agent: :default)
   args = [
-    curl.to_s,
-    "--remote-time",
-    "--location",
+    curl_executable.to_s,
+    "--fail",
```
MikeMcQuaid: We don't want this to fail silently when output is requested.

Library/Homebrew/utils/curl.rb
```diff
@@ -1,42 +1,46 @@
...
+  unless show_output
     args << "--progress-bar" unless ARGV.verbose?
     args << "--verbose" if ENV["HOMEBREW_CURL_VERBOSE"]
-    args << "--fail"
-    args << "--silent" if !$stdout.tty? || ENV["TRAVIS"]
   end
 
-  args += options[:extra_args] if options[:extra_args]
-  args
+  args << "--silent" if !$stdout.tty? || ENV["TRAVIS"]
```
MikeMcQuaid: This is going to be silent on non-TTY/Travis even when output is requested, no?

reitermarkus: `--silent` doesn't silence the output, it only silences the progress bar/indicator, which would go to `stderr`, so it doesn't have an effect on `curl_output`.

Library/Homebrew/utils/curl.rb
reitermarkus: Added the `--show-error` flag.

Library/Homebrew/utils/curl.rb
MikeMcQuaid: This is adding it unconditionally to `curl_args` though which, even if it makes no difference, does not feel like appropriate behaviour.

reitermarkus:
Did a `brew fetch --build-from-source` and normal `brew fetch` with `brew ls` as input. Also did a `brew fetch --build-from-source` and `brew fetch --build-from-source --HEAD` with some of the Formulae which explicitly specify `:using => ...`.

So this should be ready now.

Issue #2847, opened 2017-07-01
Non-strict audits should not cause CI failure (even if they happen to be implemented as cops)


ilovezfs:
Because [BrewTestBot calls `brew style foo`](https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L832-L836), any audits that have been moved to cops will cause CI to fail, even if they have been explicitly designated as `--strict` audits, which by definition should not cause CI failure (except for new formulae).

So, for example,

```
iMac-TMP:~ joe$ brew audit qemu
iMac-TMP:~ joe$ brew style qemu
== /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/qemu.rb ==
C:  2:  9: Description should start with a capital letter

1 file inspected, 1 offense detected
iMac-TMP:~ joe$ brew audit --strict qemu
qemu:
  * C: 2: col 9: Description should start with a capital letter
Error: 1 problem in 1 formula
iMac-TMP:~ joe$
```

This makes the "description should start with a capital letter" audit fatal for CI even though it was supposed to be a `--strict` failure only.

CC @GauthamGoli @JCount @MikeMcQuaid

GauthamGoli:
I had this in pipeline. Will figure out and push a fix.

ilovezfs:
Thanks @GauthamGoli!

JCount:
@GauthamGoli Great response time! 👍

MikeMcQuaid:
I think ideal behaviour would have a plain 'rubocop' still flag these up but not 'brew style' (so they still show up in editor)

MikeMcQuaid:
Fixed by https://github.com/Homebrew/brew/issues/2853.

PR #2853, opened 2017-07-03
style: Don't run FormulaAuditStrict cops when `brew style foo` cmd is executed

GauthamGoli:

-----
#2847 
cc @ilovezfs @JCount  @MikeMcQuaid 
This basically disables `FormulaAuditStrict` cops when `brew style` is ran on some formula and neither `--only-cops` nor `--except-cops` is passed as cmd line options. 

Also means, `brew style some_new_formula` wouldn't run the strict checks, but I'm assuming `brew audit --strict` would run those? 
Let me know if this code behaviour is acceptable.

### DIFF
--- a/Library/Homebrew/cmd/style.rb
+++ b/Library/Homebrew/cmd/style.rb
@@ -46,6 +46,8 @@ module Homebrew
       options[:only_cops] = only_cops
     elsif !except_cops.empty?
       options[:except_cops] = except_cops
+    elsif only_cops.empty? && except_cops.empty?
+      options[:except_cops] = %w[FormulaAuditStrict FormulaAudit]
     end
 
     Homebrew.failed = check_style_and_print(target, options)


ilovezfs:
I don't think `brew style` should report non-strict audit issues either.

GauthamGoli:
Yeah its kind of redundant given that `brew audit` runs the checks. Should I disable `FormulaAudit` rules too then?

MikeMcQuaid:
> Yeah its kind of redundant given that brew audit runs the checks. Should I disable FormulaAudit rules too then?

Yeh, seems like a good idea 👍. Given this (good) approach in the PR you'll still get those in your editor if you have editor integration setup.

JCount:
This seems to be a good approach. 👍

MikeMcQuaid:
Thanks again @GauthamGoli!

PR #2858, opened 2017-07-05
Call upgrade if something was passed to update

marinintim:

-----

This pull request calls `brew upgrade <formula>` if was mistakenly called as `brew update <formula>`. If there is an obvious and single solution to user's mistake, why wouldn't homebrew fix it itself?

I didn't add any tests, because I'm not sure how to test this change.

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -336,7 +336,7 @@ homebrew-update() {
       *)
         odie <<EOS
 This command updates brew itself, and does not take formula names.
-Use 'brew upgrade <formula>'.
+Use 'brew upgrade $@' instead.
 EOS
         ;;
     esac


reitermarkus:
`upgrade` and `update` are two different things, so as a user I wouldn't expect to be redirected to `upgrade`.

marinintim:
@reitermarkus: when you are aware of difference between `upgrade` и `update`, you would not call `update` with formula as argument anyway, as explained by error message. What's the case where user experiences this message and her next action is not to call brew upgrade with same arguments?

I use npm more often, where `npm update` is an alias to `upgrade`, and therefore I use `update <pkg>` automatically, so I had seen this message at least a few times.

MikeMcQuaid:
> I use npm more often, where npm update is an alias to upgrade, and therefore I use update <pkg> automatically, so I had seen this message at least a few times.

Thanks for the PR! I appreciate this explanation and if others feel the same way: please chime in on this PR but I'd like to hear from a lot more people before we add this.

ilovezfs:
While it might be convenient to be able to be able to use `brew update foo` and `brew up foo`, it seems like it would be surprising for `brew update` (without additional arguments) and `brew upgrade` (without additional arguments) to do wildly different things.

vitorgalvao:
> it seems like it would be surprising for `brew update` (without additional arguments) and `brew upgrade` (without additional arguments) to do wildly different things.

Agreed. It seems like someone would get used to `update` (because it’s always automatically corrected) and always use it, then be surprised when it acts differently from expected when used with no arguments.

Now that I think about it, since `update` is now called automatically, there’s little reason to ever call it on purpose. So if a user gets used to always calling `upgrade` instead, they also do not need to care about the difference and the issue will be solved.

marinintim:
Maybe then not call brew upgrade immediately, but instead print `You probably meant 'brew upgrade whatever'` instead of `<formula>`. Then it's easier to copy and paste.

But I do not see a problem that people would rely on this behaviour constantly and would be surprised when they'd try `brew update` without arguments. The error message is still here, notifying the user that this is not the intended usage of `update`, yet taking the most logical step immediately. This tidbit would improve UX of using brew a little without hurting anyone else.

MikeMcQuaid:
> Maybe then not call brew upgrade immediately, but instead print You probably meant 'brew upgrade whatever' instead of <formula>. Then it's easier to copy and paste.

This is something we'd merge if you change it to output the full arguments for `update.sh` (not `update-reset.sh`). I'm now 👎 on making this run `brew upgrade` given the conversation above.

marinintim:
@MikeMcQuaid: I've updated pull request to only display message without running `brew upgrade`.

Library/Homebrew/cmd/update.sh
```diff
@@ -336,7 +336,8 @@ homebrew-update() {
       *)
         odie <<EOS
 This command updates brew itself, and does not take formula names.
-Use 'brew upgrade <formula>'.
+Use 'brew upgrade <formula>' instead.
+You probably want 'brew upgrade $@'...
```
MikeMcQuaid: `Use 'brew upgrade $@' instead.` I think is sufficient and more concise, thanks.

marinintim: @MikeMcQuaid: done

marinintim:
Rebased on current master.

Summary: after applying this change `brew upgrade` would not be called automagically, it only changes message to copy-paste ready one.

MikeMcQuaid:
Thanks for your first contribution to Homebrew, @marinintim! Without people like you submitting PRs we couldn't run this project. You rock!

PR #2860, opened 2017-07-05
Bump `rubocop-cask` version.

reitermarkus:
Bump `rubocop-cask` version and reword the spec.


MikeMcQuaid:
A suggestion: figure out a way to embed at least part of this into Rubocop files that can be run standalone like the new Homebrew Rubocops so people can get some e.g. editor integration without installing the gem.

PR #2862, opened 2017-07-05
Formula#keg_only should be a boolean

mistydemeo:

-----

Since the left half of this conditional is false if `keg_only_reason` is nil, this was returning `nil` or `true` instead of `false` or `true`. Noticed this via the JSON output, where this value is exposed publicly.

Before:
```json
    "linked_keg": "8.27",
    "pinned": false,
    "outdated": false,
    "keg_only": null,
```

After:

```json
    "linked_keg": "8.27",
    "pinned": false,
    "outdated": false,
    "keg_only": false,
```

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1008,7 +1008,7 @@ class Formula
   # rarely, you don't want your library symlinked into the main prefix
   # see gettext.rb for an example
   def keg_only?
-    keg_only_reason && keg_only_reason.valid?
+    !!keg_only_reason && keg_only_reason.valid?
   end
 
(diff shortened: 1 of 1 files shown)


mistydemeo:
Rubocop doesn't like `!!`, but it feels like the most natural solution in this case. It's not like #2699 where there was a more natural method to use.

```
Library/Homebrew/formula.rb:1011:5: C: Avoid the use of double negation (!!).
    !!keg_only_reason && keg_only_reason.valid?
    ^
```

mistydemeo:
Pushing to `origin` was a mistake. Reopening from my fork.

PR #2863, opened 2017-07-06
Formula#keg_only should be a boolean

mistydemeo:

-----

Since the left half of this conditional is false if `keg_only_reason` is nil, this was returning `nil` or `true` instead of `false` or `true`. Noticed this via the JSON output, where this value is exposed publicly.

Before:
```json
    "linked_keg": "8.27",
    "pinned": false,
    "outdated": false,
    "keg_only": null,
```

After:

```json
    "linked_keg": "8.27",
    "pinned": false,
    "outdated": false,
    "keg_only": false,
```

Rubocop doesn't like `!!`, but it feels like the most natural solution in this case. It's not like #2699 where there was a more natural method to use.

```
Library/Homebrew/formula.rb:1011:5: C: Avoid the use of double negation (!!).
    !!keg_only_reason && keg_only_reason.valid?
    ^
```

-----

This was originally opened at #2862, which was accidentally pushed to `origin` rather than my fork.

### DIFF
--- a/Library/Homebrew/.rubocop.yml
+++ b/Library/Homebrew/.rubocop.yml
@@ -33,6 +33,10 @@ Style/BlockDelimiters:
   Exclude:
     - '**/*_spec.rb'
 
+# used idiomatically to return boolean values
+Style/DoubleNegation:
+  Enabled: false
+
 # so many of these in formulae but none in here
 Style/GuardClause:
   Enabled: true
--- a/Library/Homebrew/.rubocop.yml
+++ b/Library/Homebrew/.rubocop.yml
@@ -33,10 +33,6 @@ Style/BlockDelimiters:
   Exclude:
     - '**/*_spec.rb'
 
-# used idiomatically to return boolean values
-Style/DoubleNegation:
-  Enabled: false
-
 # so many of these in formulae but none in here
 Style/GuardClause:
   Enabled: true
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1008,7 +1008,8 @@ class Formula
   # rarely, you don't want your library symlinked into the main prefix
   # see gettext.rb for an example
   def keg_only?
-    !!keg_only_reason && keg_only_reason.valid?
+    return false unless keg_only_reason
+    keg_only_reason.valid?
   end
 
   # @private
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1008,7 +1008,7 @@ class Formula
   # rarely, you don't want your library symlinked into the main prefix
   # see gettext.rb for an example
   def keg_only?
-    keg_only_reason && keg_only_reason.valid?
+    !!keg_only_reason && keg_only_reason.valid?
   end
 
   # @private


reitermarkus:
> it feels like the most natural solution in this case.

Maybe a guard clause? `return false unless keg_only_reason`
Or duck-typing? `keg_only_reason.respond_to?(:valid?) && keg_only_reason.valid?`

mistydemeo:
Pushed a commit to disable the `Style/DoubleNegation` cop. There's more discussion of this in bbatsov/rubocop#3344 and the Ruby style guide, but there's a reasonable agreement that *returning* values via double negative is just fine.

reitermarkus:
I believe I remember you saying that Ruby 2.3 couldn't be vendored because of some bug, is that still the case?

Would be nice once High Sierra ships, so we can use `keg_only_reason&.valid?`. 😉

mistydemeo:
Yeah, that's right - https://bugs.ruby-lang.org/issues/13413 hasn't been fixed yet. I might take a stab at it sometime.

Library/Homebrew/.rubocop.yml
```diff
@@ -33,6 +33,10 @@ Style/BlockDelimiters:
   Exclude:
     - '**/*_spec.rb'
 
+# used idiomatically to return boolean values
+Style/DoubleNegation:
+  Enabled: false
```
MikeMcQuaid: Let's not disable this for a single case where we don't like it.

Library/Homebrew/formula.rb
```diff
@@ -1008,7 +1008,7 @@ def caveats
   # rarely, you don't want your library symlinked into the main prefix
   # see gettext.rb for an example
   def keg_only?
-    keg_only_reason && keg_only_reason.valid?
+    !!keg_only_reason && keg_only_reason.valid?
```
MikeMcQuaid:
```ruby
return false unless keg_only_reason
keg_only_reason.valid?
```

ilovezfs: why false?

MikeMcQuaid: That matches the original logic but returns `false` instead of `nil`, no?

MikeMcQuaid: I'm open to me being stupid here, though 😁

ilovezfs: Right, but I think that was a bug. If I add the following line to a formula
```
  keg_only
```
with no reason supplied, presumably I intend `keg_only` to be `true` not `false`.

MikeMcQuaid: Yeh, good point. Following that case, though: would you want it to be silently not keg-only if the reason isn't valid?

ilovezfs: My not supplying a reason I think would mean I want it to be `keg_only` unconditionally, just as `fails_with :clang` means it fails with clang, regardless of whether I happen to elaborate in a `do` block with a `build`, or `cause`, or both.

ilovezfs: Alternatively, we could make `keg_only` without a reason a run-time failure with a "please supply a reason, fool" message, though being able to just add `keg_only` for testing purposes is kind of a good thing. Perhaps lack of a reason is best left to audit to complain about.

ilovezfs: Ditto `conflicts_with "foo"` without a `:because =>`

ilovezfs: >would you want it to be silently not keg-only if the reason isn't valid?

Do you mean a non-nil invalid reason, like on >= Mountain Lion

```
  keg_only :provided_pre_mountain_lion
```
Then yes, I think we want that to be silently non-keg-only, right? If I'm on >= mountain_lion, why would I want to be told about the "invalid" reason that has nothing to do with me?

MikeMcQuaid: Yeh, in that case:

```ruby
return true unless keg_only_reason
keg_only_reason.valid?
```

ilovezfs: or do you prefer
```
keg_only_reason.nil? || keg_only_reason.valid?
```

JCount: Wouldn't that also return true any time keg_only is not included in a formula?

reitermarkus: If I specify `keg_only`, wouldn‘t it still call https://github.com/Homebrew/brew/blob/fd5e673d2d367eff03f1411d909ca0bbb2003a2c/Library/Homebrew/formula.rb#L2343?

ilovezfs: >Wouldn't that also return true any time keg_only is not included in a formula?

That would be bad lol

>If I specify keg_only, wouldn‘t it still call https://github.com/Homebrew/brew/blob/fd5e673d2d367eff03f1411d909ca0bbb2003a2c/Library/Homebrew/formula.rb#L2343? So keg_only_reason wouldn‘t be nil.

@reitermarkus ah. Is empty string considered a valid reason?

JCount: Currently, only having `keg_only` throws `Error: wrong number of arguments (0 for 1..2)`.

reitermarkus: I'm confused now, wouldn't `keg_only` without arguments actually fail? The method signature is `keg_only(reason, explanation = "")`.

ilovezfs: Fun times.

JCount: with:
```
keg_only_reason.nil? || keg_only_reason.valid?
```

I get:
```
irb(main):001:0> :bash.f.keg_only?
=> true
```

ilovezfs: I accept that the DSL does not love me and assent to
```
return false unless keg_only_reason
keg_only_reason.valid?
```

reitermarkus: So it should probably be `keg_only(reason = true, explanation = "")`.

ilovezfs: @reitermarkus how would that behave when there is no `keg_only` included in the formula?

reitermarkus: Then `keg_only_reason` would be `nil`.

ilovezfs: Sounds good to me.

MikeMcQuaid:
Thanks again @mistydemeo!

PR #2877, opened 2017-07-07
FileUtils.make: don't bypass superenv

mistydemeo:

-----

Updated version of #2495 - this should resolve the problems with that version. Was reminded when working in homebrew-portable again.

cc @xu-cheng - I believe this fixes the issue you noticed with that branch.

### DIFF
--- a/Library/Homebrew/shims/super/make
+++ b/Library/Homebrew/shims/super/make
@@ -1,4 +1,5 @@
 #!/bin/bash
 
+export MAKE=${MAKE:-make}
 export HOMEBREW_CCCFG="O$HOMEBREW_CCCFG"
-exec xcrun make "$@"
+exec xcrun $MAKE "$@"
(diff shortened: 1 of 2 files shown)


Library/Homebrew/shims/super/make
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 
+export MAKE=${MAKE:-make}
```
xu-cheng: Would it better to use `HOMEBREW_MAKE` as it is a private variable used by Homebrew?

xu-cheng:
:+1:

Library/Homebrew/shims/super/make
mistydemeo: This was in response to https://github.com/Homebrew/brew/pull/2495#commitcomment-21896290, which suggested setting this to make sure the right `MAKE` gets called in all circumstances.

MikeMcQuaid:
Thanks again @mistydemeo!

PR #2878, opened 2017-07-09
List cask full-names

wendorf:
`brew cask list` supports the `--full-name` flag which will include the tap name for casks not part of the core caskroom/cask tap. For example, if cask "foo-beta" is installed from the caskroom/versions cask, `brew cask list --full-name` will report the name as "caskroom/versions/foo-beta".

This will be especially helpful with `brew bundle cleanup`, which currently cannot correctly identify a cask installed from a non-core tap.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cask.rb
+++ b/Library/Homebrew/cask/lib/hbc/cask.rb
@@ -41,6 +41,14 @@ module Hbc
                           .reverse
     end
 
+    def full_name
+      if @tap.nil? || @tap == Hbc.default_tap
+        token
+      else
+        "#{@tap}/#{token}"
+      end
+    end
+
     def installed?
       !versions.empty?
     end
--- a/Library/Homebrew/test/cask/cli/list_spec.rb
+++ b/Library/Homebrew/test/cask/cli/list_spec.rb
@@ -14,6 +14,26 @@ describe Hbc::CLI::List, :cask do
     EOS
   end
 
+  it "lists full names" do
+    casks = %w[
+      local-caffeine
+      third-party/tap/third-party-cask
+      local-transmission
+    ].map { |c| Hbc::CaskLoader.load(c) }
+
+    casks.each do |c|
+      InstallHelper.install_with_caskfile(c)
+    end
+
+    expect {
+      Hbc::CLI::List.run("--full-name")
+    }.to output(<<-EOS.undent).to_stdout
+      local-caffeine
+      local-transmission
+      third-party/tap/third-party-cask
+    EOS
+  end
+
   describe "lists versions" do
     let(:casks) { ["local-caffeine", "local-transmission"] }
     let(:expected_output) {
--- a/Library/Homebrew/test/support/helper/spec/shared_context/homebrew_cask.rb
+++ b/Library/Homebrew/test/support/helper/spec/shared_context/homebrew_cask.rb
@@ -18,6 +18,7 @@ HOMEBREW_CASK_DIRS = [
 
 RSpec.shared_context "Homebrew-Cask" do
   around(:each) do |example|
+    third_party_tap = Tap.fetch("third-party", "tap")
     begin
       dirs = HOMEBREW_CASK_DIRS.map do |dir|
         Pathname.new(TEST_TMPDIR).join("cask-#{dir}").tap do |path|
@@ -31,11 +32,18 @@ RSpec.shared_context "Homebrew-Cask" do
         FileUtils.ln_sf TEST_FIXTURE_DIR.join("cask"), tap.path
       end
 
+      third_party_tap.tap do |tap|
+        FileUtils.mkdir_p tap.path.dirname
+        FileUtils.ln_sf TEST_FIXTURE_DIR.join("third-party"), tap.path
+      end
+
       example.run
     ensure
       FileUtils.rm_rf dirs
       Hbc.default_tap.path.unlink
       FileUtils.rm_rf Hbc.default_tap.path.parent
+      third_party_tap.path.unlink
+      FileUtils.rm_rf third_party_tap.path.parent
     end
   end
 end
--- a/Library/Homebrew/cask/lib/hbc/cli/list.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/list.rb
@@ -3,6 +3,7 @@ module Hbc
     class List < AbstractCommand
       option "-1", :one, false
       option "--versions", :versions, false
+      option "--full-name", :full_name, false
 
       option "-l", (lambda do |*|
         one = true # rubocop:disable Lint/UselessAssignment
@@ -42,6 +43,8 @@ module Hbc
           puts installed_casks.map(&:to_s)
         elsif versions?
           puts installed_casks.map(&self.class.method(:format_versioned))
+        elsif full_name?
+          puts installed_casks.map(&:full_name).sort &tap_and_name_comparison
         elsif !installed_casks.empty?
           puts Formatter.columns(installed_casks.map(&:to_s))
         end
--- a/Library/Homebrew/cmd/list.rb
+++ b/Library/Homebrew/cmd/list.rb
@@ -39,15 +39,7 @@ module Homebrew
       filtered_list
     elsif ARGV.named.empty?
       if ARGV.include? "--full-name"
-        full_names = Formula.installed.map(&:full_name).sort do |a, b|
-          if a.include?("/") && !b.include?("/")
-            1
-          elsif !a.include?("/") && b.include?("/")
-            -1
-          else
-            a <=> b
-          end
-        end
+        full_names = Formula.installed.map(&:full_name).sort &tap_and_name_comparison
         return if full_names.empty?
         puts Formatter.columns(full_names)
       else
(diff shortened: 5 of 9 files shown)


Library/Homebrew/test/cask/cli/list_spec.rb
```diff
@@ -14,6 +14,22 @@
     EOS
   end
 
+  it "lists full names" do
+    casks = %w[local-caffeine external-caskroom/spec/external-cask local-transmission].map { |c| Hbc::CaskLoader.load(c) }
```
reitermarkus: Make this multiple lines.

Library/Homebrew/formula.rb
```diff
@@ -1364,6 +1364,18 @@ def self.clear_installed_formulae_cache
     @installed = nil
   end
 
+  def self.comparison
```
reitermarkus: Since this is used for both Formulae and Casks, this should live somewhere else, and have a self-explanatory name.

Also, maybe add a test for this. This could even be simplified to `->(a) { [a.include?("/"), a] }` I think, but I'm not sure without a test.

Library/Homebrew/test/support/helper/spec/shared_context/homebrew_cask.rb
```diff
@@ -31,11 +32,18 @@
...
+        FileUtils.mkdir_p tap.path.dirname
+        FileUtils.ln_sf TEST_FIXTURE_DIR.join("external-cask"), tap.path
+      end
+
       example.run
     ensure
       FileUtils.rm_rf dirs
       Hbc.default_tap.path.unlink
       FileUtils.rm_rf Hbc.default_tap.path.parent
+      external_caskroom_tap.path.unlink
```
reitermarkus: This is unnecessary when you are already calling `FileUtils.rm_rf` on the parent below.

Library/Homebrew/test/support/helper/spec/shared_context/homebrew_cask.rb
```diff
@@ -18,6 +18,7 @@
 
 RSpec.shared_context "Homebrew-Cask" do
   around(:each) do |example|
+    external_caskroom_tap = Tap.fetch("external-caskroom", "spec")
```
reitermarkus: I'd name this `third-party` instead of `external-caskroom`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -10,6 +10,13 @@ class Cask
     def initialize(token, sourcefile_path: nil, &block)
       @token = token
       @sourcefile_path = sourcefile_path
+
+      @tap = if @sourcefile_path == CaskLoader.default_path(@token)
+        Hbc.default_tap
+      elsif @sourcefile_path.to_s =~ HOMEBREW_TAP_PATH_REGEX
+        Tap.fetch(Regexp.last_match(1), Regexp.last_match(2))
+      end
```
reitermarkus: I don't think this will ever work this way with `brew cask list`. For installed Casks, the `sourcefile_path` will always be the one in the `Caskroom` metadata directory, hence will never contain any information about the Tap.

wendorf: I found this to be required and correct for third-party taps. For example, the `sourcefile_path` for caskroom/homebrew-versions/microsoft-remote-desktop-beta is "/usr/local/Homebrew/Library/Taps/caskroom/homebrew-versions/Casks/microsoft-remote-desktop-beta.rb" in my environment. `Regexp.last_match(1)` is "caskroom" and `Regexp.last_match(2)` is "homebrew-versions".

Library/Homebrew/formula.rb
wendorf: There's no location that jumps out to me as more appropriate. My first thought is to add formula_comparer.rb, but this still suffers from the "this is used by formulae and casks" issue and I'm hesitant to add another file to ./Library/Homebrew. Do you have a location in mind? Is there a common name for "formulae and casks"? Maybe "token_comparer.rb" or "full_token_comparer.rb"?

Library/Homebrew/test/cask/cli/list_spec.rb
wendorf: Addressed in the latest commit

Library/Homebrew/test/support/helper/spec/shared_context/homebrew_cask.rb
wendorf: Addressed in the latest commit

Library/Homebrew/formula.rb
MikeMcQuaid: Agreed this should live elsewhere and additionally a more descriptive name would be good (`tap_and_name_comparison` or similar)

wendorf: Thanks for the thoughts! I've moved this to utils/comparer.rb as `Comparer.tap_and_name` and tested it.

MikeMcQuaid: I think `Comparer` is too generic a class name and it doesn't make sense having a single method there. `tap_and_name_comparison` in `Utils` makes more sense.

wendorf: Sounds good. It's now in utils/comparison.rb, but extends `Utils` instead of creating a new module.

MikeMcQuaid: Sorry to not be clearer: please put it in `utils.rb`, thanks.

wendorf: Whoops, sorry about this misunderstanding. Just pushed out another change to move it there as a global method.

Library/Homebrew/cask/lib/hbc/cask.rb
reitermarkus: Maybe this should be set from `CaskLoader` instead.

wendorf: Sure thing—just pushed that change. It's a little funky to me now that the cask_spec is also testing `CaskLoader` logic, but it was doing that before so I think it's fine.

Library/Homebrew/cask/lib/hbc/cask_loader.rb
```diff
@@ -31,6 +31,12 @@ def initialize(path)
 
         @token = path.basename(".rb").to_s
         @path = path
+
+        @tap ||= if @path == CaskLoader.default_path(@token)
```
reitermarkus: What I meant here was that it should only be set by the `FromTapLoader`, since this is the only case where we can be sure of the Tap. And `Cask#full_name` will still default to `Hbc.default_tap` if it is `nil`.

wendorf: Unfortunately, this would make `--full-name` not work, since the locally-installed casks are not loaded with `FromTapLoader`.

The two conditionals use full paths (`default_path` resolves to something like `/usr/local/Homebrew/Library/Taps/caskroom/homebrew-cask/Casks/foo.rb`, and `HOMEBREW_TAP_PATH_REGEX` to something like `(?-mix:\\/usr\\/local\\/Homebrew\\/Library\\/Taps\\/([\\w-]+)\\/([\\w-]+)\\/(.*))`), so I believe we can be quite confident that this is the correct tap. This is the same logic that is used in `Formula#initialize`, which is why I originally felt comfortable putting it directly in `Hbc::Cask#initialize` instead of `FromPathLoader`.

wendorf: I tried out changing `Hbc::Scopes::installed` (used by the `brew cask list` command) to fetch casks by tap path, thus using `FromTapLoader` (see the [patch](https://gist.github.com/wendorf/10619660bcba7c8450f91bce7df32809)), but failed the `rspec ./test/cask/scopes_spec.rb:21 # Hbc::Scopes installed optimizes performance by resolving to a fully qualified path before calling Hbc::CaskLoader.load` test. It looks like this optimization was put there for good reason, so I'm hesitant to change that behavior.

@reitermarkus Is there another option you think I should consider? Or do you think my existing approach is okay since the path checking is checking explicitly for the Homebrew tap install locations? I think path checking is the only way to do it without making seemingly-undesirable changes to how `brew cask list` gets its casks.

wendorf: @reitermarkus Thanks for your work on #2965 , which makes this part unnecessary. I've rebased and removed the unneeded code, so now this doesn't need to touch `CaskLoader` or the `Cask` constructor at all.

Library/Homebrew/test/support/helper/spec/shared_context/homebrew_cask.rb
```diff
@@ -18,6 +18,7 @@
 
 RSpec.shared_context "Homebrew-Cask" do
   around(:each) do |example|
+    third_party_tap = Tap.fetch("third-party", "spec")
```
reitermarkus: Maybe call this `"tap"` instead of `"spec"`.

reitermarkus:
@wendorf, one small change, and this should be good to go.

wendorf:
@reitermarkus Thanks for the additional feedback. Just did another rebase and push, so hopefully this is ready to be merged!

reitermarkus:
Thanks, @wendorf!

PR #2881, opened 2017-07-10
Follow the XDG Base Directory specification for caching on Linux

rwhogg:
Prefer $XDG_CACHE_HOME if defined, falling back to $HOME/.cache.
The Mac cache location is not affected by this change.

Most Linux users won't have a ~/Library directory, and those who do probably use it for something very different than what macOS uses it for.

-----

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -123,7 +123,15 @@ HOMEBREW_USER_AGENT_CURL="$HOMEBREW_USER_AGENT $HOMEBREW_CURL_VERSION"
 
 if [[ -z "$HOMEBREW_CACHE" ]]
 then
-  HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
+  if [[ -n "$HOMEBREW_MACOS" ]]
+  then
+    HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
+  elif [[ -n "$XDG_CACHE_HOME" ]]
+  then
+    HOMEBREW_CACHE="$XDG_CACHE_HOME/Homebrew"
+  else
+    HOMEBREW_CACHE="$HOME/.cache/Homebrew"
+  fi
 fi
 
 # Declared in bin/brew


Library/Homebrew/brew.sh
```diff
@@ -123,7 +123,15 @@ HOMEBREW_USER_AGENT_CURL="$HOMEBREW_USER_AGENT $HOMEBREW_CURL_VERSION"
 
 if [[ -z "$HOMEBREW_CACHE" ]]
 then
-  HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
+  if [[ -n "$HOMEBREW_OSX" ]]
```
rwhogg: Whoops, need to change this to HOMEBREW_MACOS. Will fix.

rwhogg:
Ping @sjackman because this isn't the same as the current Linuxbrew behaviour, but IMO this is what we should do.

Unrelated note: the exact opposite happens when using Rubocop (on both Mac and Linux) where XDG_CACHE_HOME is set to a subdirectory of HOMEBREW_CACHE. (https://github.com/Homebrew/brew/pull/2690/files) I don't believe this is an issue, but bringing it up in case there's potential for conflict.

Library/Homebrew/brew.sh
```diff
@@ -123,7 +123,15 @@ HOMEBREW_USER_AGENT_CURL="$HOMEBREW_USER_AGENT $HOMEBREW_CURL_VERSION"
...
-  HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
+  if [[ -n "$HOMEBREW_MACOS" ]]
+  then
+    HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
+  elif [[ -n "$XDG_CACHE_HOME" ]]
+  then
+    HOMEBREW_CACHE="$XDG_CACHE_HOME/Homebrew"
+  else
+    HOMEBREW_CACHE="$HOME/.cache/Homebrew"
+  fi
```
rwhogg: Hmm, would probably make sense to move these into the appropriate blocks above, but I can deal with that after everyone's in agreement about the cache location.

sjackman:
> Ping @sjackman because this isn't the same as the current Linuxbrew behaviour, but IMO this is what we should do.

Looks good to me. For background, Linuxbrew currently uses `HOMEBREW_CACHE=~/.cache/Homebrew` on Linux (and `~/Library/Caches/Homebrew` on maOS) and ignores `$XDG_CACHE_HOME`.

MikeMcQuaid:
I think we should consider respecting this path on macOS too. What do you think (CC @reitermarkus who I suspect has feelings).

MikeMcQuaid:
CC @reitermarkus for feelings if you have any 😁

reitermarkus:
No strong feelings. I don't have `XDG_CACHE_HOME` set (does anyone on macOS?), so as long as the default doesn't change, it's fine with me.

MikeMcQuaid:
@reitermarkus I asked because you added it for `brew style`?

reitermarkus:
Ah, ok. Well, I only used `XDG_CACHE_HOME` for RuboCop because there is no other way to change the cache directory.

MikeMcQuaid:
@reitermarkus But just to be super explicit: you wouldn't mind if your setting it meant that the whole Homebrew cache went there?

apjanke:
I've actually been experimenting with the `$XDG_*_HOME` stuff on Mac, and I also think it makes sense for the whole Homebrew cache to go under `$XDG_CACHE_HOME` if it's set.

ilovezfs:
I'm a bit confused how this adds value. Why wouldn't someone just set

```
export HOMEBREW_CACHE=$XDG_CACHE_HOME
```

if they want this behavior? It seems redundant.

rwhogg:
> I'm a bit confused how this adds value.

Do you mean in general, or just for macOS?

ilovezfs:
Both, since setting HOMEBREW_CACHE works on Linuxbrew too.

rwhogg:
For Linux, it's mostly because whenever you're using Homebrew/brew and fetch a formula, it'll create ~/Library and clutter your home directory if you haven't set your cache to something else.

Granted, this should mostly be Linuxbrew developers doing this at the moment, so you could argue that the couple of us who get annoyed by this should just export HOMEBREW_CACHE to the same location used by Linuxbrew. However, once Linuxbrew starts using Homebrew/brew officially, it may annoy users.

For macOS, it's probably a harder sell. XDG Base Directory and other freedesktop.org standards aren't intended for arbitrary operating systems; they're meant to provide consistency between GNOME/KDE/XFCE/etc. I'm happy implementing anything there, so long as everyone's good with the location.

ilovezfs:
Why wouldn't we just set HOMEBREW_CACHE to the current Linuxbrew default if the detected OS is Linux?

rwhogg:
> Why wouldn't we just set HOMEBREW_CACHE to the current Linuxbrew default if the detected OS is Linux?

Mostly because Linuxbrew currently implements *half* of the specification for XDG_CACHE_HOME - the current default is the specified fallback. (See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html, which I apparently forgot to post in the OP. Sorry about that!)

To be clear, while I'd prefer using XDG_CACHE_HOME as the default, I'm not dead-set on it, especially since `brew --cache` exists for anyone who gets confused.

ilovezfs:
So this PR would implement the same XDG handling as Linuxbrew but not the same fallback? That seems a bit weird.

rwhogg:
Sorry, I phrased that poorly. Here's what I meant:

* Currently, *Linuxbrew's default,* `$HOME/.cache/homebrew`, is what XDG Base Directory says is *supposed* to be the *fallback* when $XDG_CACHE_HOME is not set
* This PR implements the logic so that $XDG_CACHE_HOME is honoured by default and *falls back* to `$HOME/.cache/homebrew` if it's not set. This is what the spec says should happen.

Linuxbrew doesn't currently do *any* XDG handling; I decided to change that for this PR precisely because it seemed weird that it didn't.

ilovezfs:
If the macOS fallback is not going to be the XDG default, then it seems weird to honor a user-specified XDG_CACHE_HOME on macOS.

MikeMcQuaid:
> If the macOS fallback is not going to be the XDG default, then it seems weird to honor a user-specified XDG_CACHE_HOME on macOS.

This is a legit criticism so I'm fine with making this Linux-only. I do think we shouldn't bother using `XDG_CACHE_HOME` for `rubocop` either in that case, though.

MikeMcQuaid:
> I do think we shouldn't bother using XDG_CACHE_HOME for rubocop either in that case, though.

We're not, I'm an idiot and misunderstood that.

Sorry to contradict myself but can we move this back to being Linux-only? SORRY 😭

rwhogg:
> Sorry to contradict myself but can we move this back to being Linux-only? SORRY 😭

No problem, I'll try to make the change I mentioned [here](https://github.com/Homebrew/brew/pull/2881/files#r126328205) this weekend.

Just glad we've reached a decision on this 😸

rwhogg:
Closing in favour of #2934. I've implemented the changes suggested, but I had to make a new PR because of Linuxbrew getting deforked.

PR #2883, opened 2017-07-10
Allow passing hash to system

mistydemeo:

-----

This adds support for passing an env hash to `Formula#system` and `Homebrew.system`. This syntax is supported by the language methods we're shadowing. The env hash is combined with the system env, so it allows passing extra environment variables or overriding environment variables in individual `system`/`exec` calls.

I missed this in #2877, which makes use of this syntax.

### DIFF
--- a/Library/Homebrew/shims/super/make
+++ b/Library/Homebrew/shims/super/make
@@ -1,4 +1,5 @@
 #!/bin/bash
 
+export MAKE=${HOMEBREW_MAKE:-make}
 export HOMEBREW_CCCFG="O$HOMEBREW_CCCFG"
-exec xcrun make "$@"
+exec xcrun $MAKE "$@"
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -102,53 +102,52 @@ class Build
       end
     end
 
-    old_tmpdir = ENV["TMPDIR"]
-    old_temp = ENV["TEMP"]
-    old_tmp = ENV["TMP"]
-    ENV["TMPDIR"] = ENV["TEMP"] = ENV["TMP"] = HOMEBREW_TEMP
+    new_env = {
+      "TMPDIR" => HOMEBREW_TEMP,
+      "TEMP" => HOMEBREW_TEMP,
+      "TMP" => HOMEBREW_TEMP,
+    }
 
-    formula.extend(Debrew::Formula) if ARGV.debug?
+    with_env(new_env) do
+      formula.extend(Debrew::Formula) if ARGV.debug?
 
-    formula.brew do |_formula, staging|
-      staging.retain! if ARGV.keep_tmp?
-      formula.patch
-
-      if ARGV.git?
-        system "git", "init"
-        system "git", "add", "-A"
-      end
-      if ARGV.interactive?
-        ohai "Entering interactive mode"
-        puts "Type `exit' to return and finalize the installation"
-        puts "Install to this prefix: #{formula.prefix}"
+      formula.brew do |_formula, staging|
+        staging.retain! if ARGV.keep_tmp?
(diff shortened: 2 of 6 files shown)


ilovezfs:
I'm not thrilled with allowing this in formulae.

MikeMcQuaid:
> I'm not thrilled with allowing this in formulae.

Agreed. I'd rather we continued to use `ENV[]` as that's the pattern we've used in formula for a long time.

mistydemeo:
That's fair. My thought was that, since we're shadowing the name of a core function, it would make sense to support more of its functionality. As well, this saves us a bit of work, namely:

```ruby
def make(*args)
  system({ "MAKE" => make_name }, "make", *args)
end
```

vs

```ruby
def make(*args)
  begin
    old_make = ENV.delete "MAKE"
    ENV["MAKE"] = make_name
    system "make", *args
  rescue
    ENV["MAKE"] = old_make
  end
end
```

Since the pattern already provided in the builtin Ruby function handles this logic already, it avoids us having to write that logic in the places we need to do this (and avoids potential race conditions if we had multithreaded code).

mistydemeo:
> I'm not thrilled with allowing this in formulae.

By the way, for reference, the method I'm aiming to use this fix is in FileUtils. This is only indirectly being called in formulae; the methods included from FileUtils see `Formula#system` instead of `Homebrew.system` or `Kernel.system`.

MikeMcQuaid:
I'd be OK with providing this functionality for use not in formulae (i.e. this should specifically not work in formulae.

> `old_make = ENV.delete "MAKE"` 

We use this pattern enough in the codebase already that a `env "MAKE" => make_name do` refactoring for both formulae and internals would be best, I think.

mistydemeo:
Updated to add an `env` helper, and reimplemented the `make` fix using that.

I briefly tried out putting this in `ENV`, but the fact that `SharedEnvExtension` isn't active everywhere we'd use this made it impractical.

MikeMcQuaid:
> Updated to add an env helper,

Looks great 👍. There's a few places in `build.rb`, `formula.rb` that would use this nicely; I used `grep "] = old_"` to search for them.

> and reimplemented the make fix using that.

Is that a fix for the issue in https://github.com/Homebrew/homebrew-core/issues/15547? What was the cause?

mistydemeo:
> Is that a fix for the issue in Homebrew/homebrew-core#15547? What was the cause?

That's a separate bug, which I've just pushed a fix for. It looks like cmake's configure script, when it's testing whether to use `make` or `gmake`, sets the `MAKE` environment variable. Since the `make`/`gmake` shim scripts respect the value of `MAKE`, it means our shim tries to `xcrun` something that's out of its `PATH` and bails.

I fixed this by adjusting the environment variable to `HOMEBREW_MAKE`, so stuff like cmake won't be able to interfere. It still exports `MAKE` so subprocesses pick the make we want.

> There's a few places in build.rb, formula.rb that would use this nicely

Updated to use this!

mistydemeo:
Renamed the helper to `with_env`; the old name was clashing with `Formula#env`.

Library/Homebrew/shims/super/make
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 
+export MAKE=${HOMEBREW_MAKE:-make}
```
MikeMcQuaid: @mistydemeo Do we need to `export` this here?

mistydemeo: It's probably not strictly necessary, but it was a suggestion in the previous PR to make sure that subprocesses call the correct make.

mistydemeo:
Tested rebuilding cmake from source, along with ~5 things which require `make` including `portable-ruby@2.4`. Think we're good.

PR #2889, opened 2017-07-12
fix link to Interesting-Taps-and-Forks.md

BM5k:
re bb33685


scpeters:
see also https://github.com/Homebrew/brew/blame/master/docs/Acceptable-Formulae.md#L5

BM5k:
@scpeters fixed, thx

PR #2879, opened 2017-07-09
audit: Port audit_options non-strict rules to rubocop and add tests

GauthamGoli:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----
#569 
This PR ports non-strict rules of `audit_options` method to a custom cop. `strict` and `new-formula` rules would be dealt in separate PRs for ease of review.

PR #2901, opened 2017-07-14
audit: Port audit_options strict rules to rubocop and add tests

GauthamGoli:

-----
#569 
Continuation of #2879 
Had to convert the option node to string ( [lines 27-28](https://github.com/GauthamGoli/brew/blob/cb2fb6fce9ef1cbd60f121a85883f9a3e4ad5acc/Library/Homebrew/rubocops/options_cop.rb#L27-L28) )  because there is a negative regex match, hence was not able to reuse `regex_match_group` method which takes the node itself as an argument and does the regex match but the method does not do negative regex match. Please suggest if you think there's a better way to do this.

### DIFF
--- a/Library/.rubocop.yml
+++ b/Library/.rubocop.yml
@@ -24,6 +24,9 @@ FormulaAudit/Conflicts:
 FormulaAudit/Options:
   Enabled: true
 
+FormulaAuditStrict/Options:
+  Enabled: true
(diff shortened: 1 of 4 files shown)


JCount:
@GauthamGoli 🎊 Good work on getting another PR shipped!

PR #2905, opened 2017-07-15
audit: Port audit_options rules for new formulae to rubocop and add test

GauthamGoli:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----
#569 
Continuation of #2879 and #2901 . This PR finishes porting of `audit_options` method completely from `audit.rb` to rubocop. 
This also introduces a new department of cops `NewFormulaAudit` as the name suggests, cops which execute when `--new-formula` flag is passed to audit.

PR #2906, opened 2017-07-15
style: disable NewFormulaAudit cops' execution by default unless specified

GauthamGoli:

-----
The `NewFormulaAudit` cop got enabled for all Formulae because I forgot to disable it for `brew style` ( I disabled for `brew audit` )  in #2905, This PR fixes that mistake.

### DIFF
--- a/Library/.rubocop.yml
+++ b/Library/.rubocop.yml
@@ -6,47 +6,14 @@ AllCops:
 
 require: ./Homebrew/rubocops.rb
 
-FormulaAudit/Text:
+FormulaAudit:
   Enabled: true
 
-FormulaAudit/Caveats:
+FormulaAuditStrict:
   Enabled: true
 
-FormulaAudit/Checksum:
-  Enabled: true
-
-FormulaAudit/ChecksumCase:
-  Enabled: true
-
-FormulaAudit/Conflicts:
-  Enabled: true
-
-FormulaAudit/Options:
-  Enabled: true
-
-FormulaAuditStrict/Options:
-  Enabled: true
-
-NewFormulaAudit/Options:
-  Enabled: true 
-
-FormulaAuditStrict/BottleBlock:
-  Enabled: true
-
-FormulaAuditStrict/Desc:
-  Enabled: true
-
-FormulaAuditStrict/ComponentsOrder:
-  Enabled: true
-
-FormulaAuditStrict/ComponentsRedundancy:
-  Enabled: true
-
-FormulaAudit/Homepage:
-  Enabled: true
-
-FormulaAudit/LegacyPatches:
-  Enabled: true
+NewFormulaAudit:
+  Enabled: false
 
 # `system` is a special case and aligns on second argument
 Layout/AlignParameters:
--- a/Library/.rubocop.yml
+++ b/Library/.rubocop.yml
@@ -28,7 +28,7 @@ FormulaAuditStrict/Options:
   Enabled: true
 
 NewFormulaAudit/Options:
-  Enabled: false
+  Enabled: true 
 
 FormulaAuditStrict/BottleBlock:
   Enabled: true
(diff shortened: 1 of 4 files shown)


Library/.rubocop.yml
```diff
@@ -28,7 +28,7 @@ FormulaAuditStrict/Options:
   Enabled: true
 
 NewFormulaAudit/Options:
-  Enabled: false
+  Enabled: true
```
MikeMcQuaid: This still won't quite work because it'll show up in files when using e.g. `rubocop` in my editor. Need some way to have it be disabled by default and enabled only when requested.

GauthamGoli: Seems tricky to achieve. But I'll try to think of someway.

GauthamGoli: @MikeMcQuaid One way that works involves some redundant configuration. (Redundant because we are doing similar thing in `brew style` , but it has to be redundant because `brew` and `rubocop` run in different processes and aren't as such coupled and editor integration is independent of `brew`)

In sublime, we can add `args` line in Sublime Linter User Settings (found in `Menu->Preferences->Package Settings->SublimeLinter->Settings - User`) so final `rubocop` section should look like below 

```
"rubocop": {
       "@disable": false,
       "args": ["--except", "FormulaAuditStrict, NewFormulaAudit"],
       "excludes": []
}
```
If we are planning to go by this method, It would be better to update our docs to have a section on editor integration.

MikeMcQuaid: @GauthamGoli I think we need to have something that involves zero configuration. Is there any way to have a cop that's only run when specified on the command-line? There may be some hack using a global `ARGV` check or similar in the cop itself.

GauthamGoli: Yes
Example: https://github.com/Homebrew/brew/issues/569#issuecomment-271933527

MikeMcQuaid: @GauthamGoli Worst case we can run the new formulae cops in a separate `rubocop` invocation.

GauthamGoli: The editor plugin finds `.rubocop.yml` file at `HOMEBREW_LIBRARY` and uses it to highlight violations. 
So, in `HOMEBREW_LIBRARY/.rubocop.yml` ,`FormulaAuditStrict` and `NewFormulaAudit` departments can be disabled. This configuration can be inherited and overriden in another configuration file with the former departments(or specific cops) enabled, which will now be used in `style.rb` to handle `brew audit` and `brew style` commands. 

This also achieves the zero configuration goal for end user/developer. I can push it here if you want to see this approach.

MikeMcQuaid: I think `FormulaAuditStrict` are fine to leave enabled by default and we just have a separate configuration file for `NewFormulaAudit`. That could be useful for enforcing things like e.g. 80 character wrapping, too.

GauthamGoli:
@MikeMcQuaid 
If the developer/contributor does not want some cops to get executed in editor, then they can disable them either in the editor config or `HOMEBREW_LIBRARY/.rubocops.yml`

MikeMcQuaid:
> Yeah. This line needs to be included because this department has been disabled in .rubocop.yml. But below this we can disable cops in that namespace.

I think it's worth just enabling/disabling cops based on their namespace rather than enabling each one individually. What do you think?

GauthamGoli:
Yeah, agree.

JCount:
This seems to be a be in general a good approach. We just need to make certain that this won't accidentally trigger unwanted strict/new formula, audits/style checks on the `homebrew/core`'s CI when this eventually gets merged.

To that end, cc @ilovezfs

MikeMcQuaid:
> This seems to be a be in general a good approach. We just need to make certain that this won't accidentally trigger unwanted strict/new formula, audits/style checks on the homebrew/core's CI/test-bot when this eventually gets merged.

Tried that and it's looking 👍 so 🚢!

PR #2908, opened 2017-07-18
[RFC] docs: Allow `:python3 => :recommended`

methane:

-----

I opened [this thread on discourse](https://discourse.brew.sh/t/when-homebrew-move-to-python-3-as-default-python/851) and @MikeMcQuaid opened Homebrew/homebrew-core#14408 which was changing default Python to Python 3.

But the pull request stop trying to change default Python.

So I propose to removing the policy "`:python` dependency **should** be `:recommended`".

It means each Formula can switch to `:python3 => :recommended` and `:python => :optional` when it's Python 3 support is good enough to recommend for new users.

### DIFF
--- a/docs/Python-for-Formula-Authors.md
+++ b/docs/Python-for-Formula-Authors.md
@@ -179,7 +179,7 @@ Sometimes we have to `inreplace` a `Makefile` to use our prefix for the Python b
 
 ### Python declarations
 
-Libraries **should** declare a dependency on `:python` or `:python3` as appropriate, which will respectively cause the formula to use the first python or python3 discovered in `PATH` at install time. If a library supports both Python 2.x and Python 3.x, the `:python` dependency **should** be `:recommended` (i.e. built by default) and the `:python3` dependency should be `:optional`. Python 2.x libraries **must** function when they are installed against either the system Python or Homebrew Python.
+Libraries **should** declare a dependency on `:python` or `:python3` as appropriate, which will respectively cause the formula to use the first python or python3 discovered in `PATH` at install time. Python 2.x libraries **must** function when they are installed against either the system Python or Homebrew Python.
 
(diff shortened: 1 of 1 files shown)


ilovezfs:
This change does not make sense to me because we do not want to switch things to python3 that work fine with the system python, so the existing guidance still looks correct to me.

methane:
> we do not want to switch things to python3 that work fine with the system python,

Bottles are built with brew's Python 2, not system Python 3.
For example, `brew install vim` installs brew's python2.  It doesn't use system Python.

So this proposal is not relating to system Python usage.

ilovezfs:
That is why we will sometimes use an unbound "without-python" option. But even if we're strictly talking about `:python => :recommended`, a source build will use the system python.

methane:
> But even if we're strictly talking about :python => :recommended, a source build will use the system python.

Sorry, I can't get what your point.
I'm strictly talking about allowing each Formula choose `:python => :recommended; :python3 => :optional` or `:python => :optional; :python3 => :recommended`.
I never propose disallowing using system Python.

ilovezfs:
The question is whether this is going to lead to more PRs where someone puts
```
depends_on :python3 => :recommended
```
and in fact neither `depends_on :python => :recommended` nor `depends_on :python3 => :recommended` is needed, since system Python will suffice.

I think the answer is yes.

methane:
As you can see, this pull request removes this sentence: "the `:python` dependency **should** be `:recommended` (i.e. built by default)"

So I don't think this is relating formulas which is not `depends_on :python => :recommended` currently.

For example, vim is currently:

```
  depends_on :python => :recommended
  depends_on :python3 => :optional
```

So all vim user is forced  to install Python 2 even if they always use Python 3, if they want to install from bottle.

MikeMcQuaid:
Thanks for the PR but I think the wording is fine as-is until we start migrating to a newer Python version (which from https://github.com/Homebrew/homebrew-core/pull/14408's reaction is still a while off).

PR #2911, opened 2017-07-18
audit: Port audit_urls partially to rubocop and add corresponding tests

GauthamGoli:

-----
#569 
This PR attempts to partially port `audit_urls` method of `ResourceAuditor` to RuboCop custom cop and adds corresponding tests. 
I tried to reuse `regex_match_group` method of `FormulaCop`. Instead, for the sake of brevity of code in this cop, `string_content` method can be used to extract string part from url `node` and a `case` statement to match regexs. 

Edit: I used a closure to reduce redundant code, please let me know what you think!


Library/Homebrew/rubocops/urls_cop.rb
```diff
@@ -0,0 +1,102 @@
...
+          # GNU urls; doesn't apply to mirrors
+          gnu_pattern = %r{^(?:https?|ftp)://ftpmirror.gnu.org/(.*)}
+
+          # Fossies upstream requests they aren't used as primary URLs
+          # https://github.com/Homebrew/homebrew-core/issues/14486#issuecomment-307753234
+          fossies_pattern = %r{^https?://fossies\.org/}
+
+          # Check a variety of SSL/TLS URLs that don't consistently auto-redirect
+          # or are overly common errors that need to be reduced & fixed over time.
+          http_urls_pattern = Regexp.union([%r{^http://ftp\.gnu\.org/},
```
JCount: Perhaps this could use a name other than `http_urls_pattern`? One that better reflects the fact that the urls which match this pattern should be changed to `https://` might be a better fit . `https_url_id_pattern` is a bad substitute, and but it would probably be good to make this rely a little less on the inline comment. However, if nothing better comes to mind this is still fine. 😄

MikeMcQuaid: `http_to_https_patterns`

MikeMcQuaid: `http_to_https_patterns`

Library/Homebrew/rubocops/urls_cop.rb
```diff
@@ -0,0 +1,102 @@
...
+                                            %r{^http://github\.com/},
+                                            %r{^http://bitbucket\.org/},
+                                            %r{^http://anonscm\.debian\.org/},
+                                            %r{^http://cpan\.metacpan\.org/},
+                                            %r{^http://hackage\.haskell\.org/},
+                                            %r{^http://(?:[^/]*\.)?archive\.org},
+                                            %r{^http://(?:[^/]*\.)?freedesktop\.org},
+                                            %r{^http://(?:[^/]*\.)?mirrorservice\.org/}])
+
+          # CPAN pattern
```
MikeMcQuaid: Don't need comments like this; inferred from the variable name.

Library/Homebrew/rubocops/urls_cop.rb
```diff
@@ -0,0 +1,102 @@
...
+          # GNOME pattern
+          gnome_pattern = %r{^(http|ftp)://ftp\.gnome\.org/pub/gnome/(.*)}i
+
+          # DEBIAN pattern
+          debian_pattern = %r{^git://anonscm\.debian\.org/users/(.*)}i
+
+          urls = find_every_method_call_by_name(body_node, :url)
+          mirrors = find_every_method_call_by_name(body_node, :mirror)
+
+          audit_urls(urls, gnu_pattern) do |match_obj, url|
```
MikeMcQuaid: `do |match, url|`

Library/Homebrew/rubocops/urls_cop.rb
```diff
@@ -0,0 +1,102 @@
...
+          end
+
+          audit_urls(urls, fossies_pattern) do |_, _|
+            problem "Please don't use fossies.org in the url (using as a mirror is fine)"
+          end
+
+          audit_urls(mirrors, /.*/) do |_, mirror|
+            urls.each do |url|
+              url_string = string_content(parameters(url).first)
+              if url_string.eql?(mirror)
```
MikeMcQuaid: `next unless url_string.eql?(mirror)`

Library/Homebrew/rubocops/urls_cop.rb
```diff
@@ -0,0 +1,102 @@
...
+          debian_pattern = %r{^git://anonscm\.debian\.org/users/(.*)}i
+
+          urls = find_every_method_call_by_name(body_node, :url)
+          mirrors = find_every_method_call_by_name(body_node, :mirror)
+
+          audit_urls(urls, gnu_pattern) do |match_obj, url|
+            problem "Please use \"https://ftp.gnu.org/gnu/#{match_obj[1]}\" instead of #{url}."
+          end
+
+          audit_urls(urls, fossies_pattern) do |_, _|
```
MikeMcQuaid: I'm not sure but you may be able to omit `|_, _|` entirely.

Library/Homebrew/rubocops/urls_cop.rb
```diff
@@ -0,0 +1,102 @@
...
+          audit_urls(urls, http_urls_pattern) do |_, url|
+            problem "Please use https:// for #{url}"
+          end
+
+          audit_urls(urls, cpan_pattern) do |match_obj, url|
+            problem "#{url} should be `https://cpan.metacpan.org/#{match_obj[1]}`"
+          end
+
+          audit_urls(urls, gnome_pattern) do |match_obj, url|
+            problem "#{url} should be `https://download.gnome.org/#{match_obj[2]}`"
```
MikeMcQuaid: Let's define e.g. `gnome_pattern` immediately above where it is called so it's obvious what regex group `match[2]` refers to.

Library/Homebrew/test/rubocops/urls_cop_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+require_relative "../../rubocops/urls_cop"
+
+describe RuboCop::Cop::FormulaAudit::Urls do
+  subject(:cop) { described_class.new }
+
+  context "When auditing urls" do
+    it "with offenses" do
+      formulas = [{
+        "url" => "https://ftpmirror.gnu.org/lightning/lightning-2.1.0.tar.gz",
+        "msg" => "Please use \"https://ftp.gnu.org/gnu/lightning/lightning-2.1.0.tar.gz\" instead of https://ftpmirror.gnu.org/lightning/lightning-2.1.0.tar.gz.",
```
MikeMcQuaid: Use `'` where the string includes `"` but doesn't need interpolation.

Library/Homebrew/rubocops/urls_cop.rb
GauthamGoli: Yeah. Makes more sense.

GauthamGoli:
@MikeMcQuaid Pushed the changes and squashed the two commits into one.

MikeMcQuaid:
Thanks again @GauthamGoli!

PR #2912, opened 2017-07-18
docs: update for unshadowed system Python

jonchang:

-----

See https://discourse.brew.sh/t/what-happened-to-the-standard-python-package/935 and https://github.com/Homebrew/homebrew-core/pull/14408. I'm not sure if there needs to be much more explanation for this as the caveats for `brew install python` cover how to shadow system Python if that is desired.


MikeMcQuaid:
Thanks for the quick fix, @jonchang!

PR #2913, opened 2017-07-19
docs: fix fallback fonts

jonchang:

-----

http://docs.brew.sh/ doesn't render well on Windows due to the CSS fallback fonts being quoted. (It will look for a font called "sans-serif" and doesn't find it). 

See below. Note that the browser default font Times New Roman (a serif font) was chosen instead of a sans-serif font for body text and a monospace font for the code block.

Imgur

### DIFF
--- a/docs/css/screen.scss
+++ b/docs/css/screen.scss
@@ -111,7 +111,7 @@ html {
   margin: 0;
   padding: 0;
   font-size: 62.5%;
-  font-family: "-apple-system", "BlinkMacSystemFont", "Helvetica Neue", "Roboto", "sans-serif";
+  font-family: "-apple-system", "BlinkMacSystemFont", "Helvetica Neue", "Roboto", sans-serif;
   height: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


docs/css/screen.scss
```diff
@@ -111,7 +111,7 @@ html {
   margin: 0;
   padding: 0;
   font-size: 62.5%;
-  font-family: "-apple-system", "BlinkMacSystemFont", "Helvetica Neue", "Roboto", "sans-serif";
+  font-family: "-apple-system", "BlinkMacSystemFont", "Helvetica Neue", "Segoe UI", "Roboto", sans-serif;
```
MikeMcQuaid: Let's just unquote `sans-serif` here and not add `Segoe UI`.

apjanke:
👍 on the diagnosis and fix here, though I also agree with leaving out Segoe UI.

jonchang:
Changes made. Also see https://github.com/Homebrew/homebrew.github.io/pull/170

MikeMcQuaid:
Thanks again @jonchang!

PR #2914, opened 2017-07-19
Fix Link to XCode on Apple

SConaway:

-----
This just fixes the link to XCode on Apple's website.


mistydemeo:
The old URL is correct for this usecase; we print this link on older versions of OS X, and the `https://developer.apple.com/download/` URL only provides download links for current versions. Thanks though!

PR #2917, opened 2017-07-20
python_requirement: prefer Homebrew's Python

ilovezfs:

-----

This avoids linkage to both system Python and Homebrew's python.

### DIFF
--- a/Library/Homebrew/requirements/python_requirement.rb
+++ b/Library/Homebrew/requirements/python_requirement.rb
@@ -18,11 +18,11 @@ class PythonRequirement < Requirement
 
     if !system_python? && short_version == Version.create("2.7")
       ENV.prepend_path "PATH", which_python.dirname
-    # Homebrew Python should take precedence over older Pythons in the PATH
-    elsif short_version != Version.create("2.7")
-      ENV.prepend_path "PATH", Formula["python"].opt_bin
-      ENV.prepend_path "PATH", Formula["python"].opt_libexec/"bin"
     end
+
+    # Homebrew Python should take precedence over other Pythons in the PATH
+    ENV.prepend_path "PATH", Formula["python"].opt_bin
+    ENV.prepend_path "PATH", Formula["python"].opt_libexec/"bin"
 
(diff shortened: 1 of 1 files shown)


DomT4:
Think this caused a problem with https://github.com/Homebrew/homebrew-core/pull/15941, FWIW.

ilovezfs:
@DomT4 this is likely to cause all manner of problems but the MacVim bleeding needed to stop.

ilovezfs:
@DomT4 you'll need to elaborate for me since I don't really see how this change impacts PYTHONPATH for python3.

DomT4:
Oh, you are correct, sorry I misread the commit. I'll go hunting for the right breakage 🙈.

ilovezfs:
@DomT4 what this _will_ totally as far as I know: you don't have `depends_on :python` listed as a dependency in a formula, with which you want to use system Python, in a multi-formula PR, but brew python happens to get installed along the way.

DomT4:
Starting to wonder if my PR is an upstream thing, or a local thing, perhaps. Although I've checked my env for `PYTHON` stuff about 50 times, and there's nothing there. But I can checkout as far back as the `1.0.0` tag and still hit the problem:

```
==> python3 -c import setuptools... --no-user-cfg install --prefix=/usr/local/Cellar/diffoscope/84/libexec/vendor --single-version-externally-managed --record=installed.txt
Failed to import the site module
Your PYTHONPATH points to a site-packages dir for Python 2.x but you are running Python 3.x!
     PYTHONPATH is currently: "/usr/local/Cellar/diffoscope/84/libexec/vendor/lib/python3.6/site-packages:/usr/local/lib/python2.7/site-packages"
     You should `unset PYTHONPATH` to fix this.
```

I managed to mislead myself by checking out `stable` and then it built fine, but I forgot I hadn't commented back out the `ENV.delete` initially, so 👏 👏  @ myself.

DomT4:
Still fails with `HOMEBREW_ENV_FILTERING="1"` set as well, so presumably not local. Guess I'll move discussion over to the other PR since this seems completely unrelated then.

PR #2919, opened 2017-07-20
Add check for build from source

baughj:

-----

This is a simple change to add a warning to `brew doctor` if the unsupported environment variable `HOMEBREW_BUILD_FROM_SOURCE` is set. I believe they would be beneficial for older `brew` users (myself included) that have this set (I wasn't aware it was unsupported until recently).

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -99,7 +99,7 @@ module Homebrew
       end
 
       def check_build_from_source
-        return unless ENV.key?("HOMEBREW_BUILD_FROM_SOURCE")
+        return unless ENV("HOMEBREW_BUILD_FROM_SOURCE")
 
         <<-EOS.undent
           You have HOMEBREW_BUILD_FROM_SOURCE set. This environment variable is
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -98,6 +98,17 @@ module Homebrew
         EOS
       end
 
+      def check_build_from_source
+        return if !ENV.has_key?("HOMEBREW_BUILD_FROM_SOURCE")
+
+        <<-EOS.undent
+          You have HOMEBREW_BUILD_FROM_SOURCE set. This environment variable is
+          intended for use by Homebrew developers. If you are encountering errors,
+          please try unsetting this. Please do not file issues if you encounter
+          errors when using this environment variable.
+        EOS
+      end
+
       # See https://github.com/Homebrew/legacy-homebrew/pull/9986
       def check_path_for_trailing_slashes
         bad_paths = PATH.new(ENV["HOMEBREW_PATH"]).select { |p| p.end_with?("/") }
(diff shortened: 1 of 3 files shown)


baughj:
I notice that `docs/FAQ.md` also needs to be updated to make a note that this flag is unsupported. I am happy to add that to this PR, or open a new one.

chdiza:
`HOMEBREW_BUILD_FROM_SOURCE` is unsupported?  When did this happen?

ilovezfs:
https://github.com/Homebrew/brew/commit/b745546943a05bd2aa6d129e9a44e29771bf115a

baughj:
It has been unsupported for a while (indeed, years). I wasn't aware of this myself until I opened an issue, so I am trying to ensure that other people don't go down a rat hole trying to diagnose something when it's an unsupported path to begin with.

chdiza:
It doesn't say there that it's unsupported though.  It just says it's intended for homebrew developers.

I don't see the benefit of this.  `brew --env` will already show someone whether it's set.

baughj:
It also says `Please do not file issues if you encounter errors when using this environment variable` which, to me, is the definition of unsupported. 

If a user is going to be told to not waste time opening a ticket if they encounter errors as a result of setting it - my feeling is that `brew doctor`, the advertised first step when opening an issue, should tell the user this, so that they don't waste their time and the time of others.

docs/FAQ.md
```diff
@@ -89,16 +89,11 @@ following conditions:
 will use a bottled version of the formula, but
 `brew install <formula> --enable-bar` will trigger a source build.
 * The `--build-from-source` option is invoked.
-* The environment variable `HOMEBREW_BUILD_FROM_SOURCE` is set.
```
MikeMcQuaid: Let's leave this line as-is because it's factually correct.

docs/FAQ.md
```diff
@@ -89,16 +89,11 @@ following conditions:
...
 * The machine is not running a supported version of macOS as all
 bottled builds are generated only for supported macOS versions.
 * Homebrew is installed to a prefix other than the standard
 `/usr/local` (although some bottles support this).
 
-In order to completely disable bottled builds, simply add a value for
-the environment variable `HOMEBREW_BUILD_FROM_SOURCE` to
-your profile.
```
MikeMcQuaid: 👍 on deleting this.

Library/Homebrew/diagnostic.rb
```diff
@@ -98,6 +98,17 @@ def check_for_installed_developer_tools
         EOS
       end
 
+      def check_build_from_source
+        return unless ENV.key?("HOMEBREW_BUILD_FROM_SOURCE")
```
MikeMcQuaid: `return unless ENV["HOMEBREW_BUILD_FROM_SOURCE"]` is our usual style here.

docs/FAQ.md
baughj: What about adding something that this is not intended for general use? Like `(for developers only)`.

baughj:
OK, changed as requested and updated language to `(intended for developers only)` which matches the man page.

MikeMcQuaid:
Thanks for your first contribution to Homebrew/brew, @baughj! Without people like you submitting PRs we couldn't run this project. You rock!

PR #2934, opened 2017-07-24
Follow the XDG Base Directory specification for caching on Linux

rwhogg:
This is the same as #2881, but it had to be reopened after Linuxbrew was deforked.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -110,21 +110,31 @@ then
   then
     HOMEBREW_CURL="$HOMEBREW_PREFIX/opt/curl/bin/curl"
   fi
+
+  if [[ -z "$HOMEBREW_CACHE" ]]
+  then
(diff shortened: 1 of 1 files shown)


Library/Homebrew/brew.sh
```diff
@@ -110,22 +110,29 @@ then
   then
     HOMEBREW_CURL="$HOMEBREW_PREFIX/opt/curl/bin/curl"
   fi
+
+  [[ -z "$HOMEBREW_CACHE" ]] && HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
```
MikeMcQuaid: Keep this as an `if`, `else` ta.

rwhogg:
Fixed and pushed, thanks for the review Mike!

MikeMcQuaid:
Thanks again @rwhogg!

PR #2950, opened 2017-07-26
Added tests for utils/svn

mansimarkaur:

-----

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -1228,9 +1228,8 @@ class ResourceAuditor
         end
       elsif strategy <= SubversionDownloadStrategy
         next unless DevelopmentTools.subversion_handles_most_https_certificates?
-        if !Utils.svn_available?
-          problem "No valid version of svn found"
-        elsif Utils.svn_remote_exists url
+        next unless Utils.svn_available?
+        if Utils.svn_remote_exists url
           problem "The URL #{url} is not a valid svn URL"
         end
       end
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -1228,7 +1228,9 @@ class ResourceAuditor
         end
       elsif strategy <= SubversionDownloadStrategy
         next unless DevelopmentTools.subversion_handles_most_https_certificates?
-        unless Utils.svn_remote_exists url
+        if !Utils.svn_available?
+          problem "No valid version of svn found"
+        elsif Utils.svn_remote_exists url
           problem "The URL #{url} is not a valid svn URL"
         end
       end
--- a/Library/Homebrew/test/utils/svn_spec.rb
+++ b/Library/Homebrew/test/utils/svn_spec.rb
@@ -2,37 +2,40 @@ require "utils/svn"
 
 describe Utils do
   describe "#self.svn_available?" do
-    it "returns true if svn --version command succeeds" do
-      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
-      expect(described_class.svn_available?).to be_truthy
+    before(:each) do
+      if described_class.instance_variable_defined?(:@svn)
+        described_class.send(:remove_instance_variable, :@svn)
+      end
     end
 
-    it "returns false if svn --version command does not succeed" do
-      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
-      expect(described_class.svn_available?).to be_falsey
-    end
-
-    it "returns svn version if already set" do
-      described_class.instance_variable_set(:@svn, true)
+    it "returns svn version if svn available" do
       expect(described_class.svn_available?).to be_truthy
     end
   end
 
   describe "#self.svn_remote_exists" do
-    let(:url) { "https://dl.bintray.com/homebrew/mirror/" }
-
     it "returns true when svn is not available" do
-      described_class.instance_variable_set(:@svn, false)
-      expect(described_class.svn_remote_exists(url)).to be_truthy
+      allow(Utils).to receive(:svn_available?).and_return(false)
+      expect(described_class.svn_remote_exists("blah")).to be_truthy
     end
 
-    it "returns false when remote does not exist" do
-      expect(described_class.svn_remote_exists(url)).to be_falsey
-    end
+    context "when svn is available" do
+      before do
+        allow(Utils).to receive(:svn_available?).and_return(true)
+      end
 
-    it "returns true when remote exists" do
-      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
-      expect(described_class.svn_remote_exists(url)).to be_truthy
+      it "returns false when remote does not exist" do
+        expect(described_class.svn_remote_exists(HOMEBREW_CACHE/"install")).to be_falsey
+      end
+
+      it "returns true when remote exists", :needs_network do
+        remote = "http://github.com/Homebrew/install"
+        svn = HOMEBREW_SHIMS_PATH/"scm/svn"
+
+        HOMEBREW_CACHE.cd { system svn, "checkout", remote }
+
+        expect(described_class.svn_remote_exists(HOMEBREW_CACHE/"install")).to be_truthy
+      end
     end
   end
 end
--- a/Library/Homebrew/test/utils/svn_spec.rb
+++ b/Library/Homebrew/test/utils/svn_spec.rb
@@ -3,9 +3,7 @@ require "utils/svn"
 describe Utils do
   describe "#self.svn_available?" do
     before(:each) do
-      if described_class.instance_variable_defined?(:@svn)
-        described_class.send(:remove_instance_variable, :@svn)
-      end
+      described_class.clear_svn_version_cache
     end
 
     it "returns svn version if svn available" do
--- a/Library/Homebrew/test/utils/svn_spec.rb
+++ b/Library/Homebrew/test/utils/svn_spec.rb
@@ -2,11 +2,17 @@ require "utils/svn"
 
 describe Utils do
   describe "#self.svn_available?" do
-    it "processes value when @svn is not defined" do
+    it "returns true if svn --version command succeeds" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
       expect(described_class.svn_available?).to be_truthy
     end
 
-    it "returns value of @svn when @svn is defined" do
+    it "returns false if svn --version command does not succeed" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
+      expect(described_class.svn_available?).to be_falsey
+    end
+
+    it "returns svn version if already set" do
       described_class.instance_variable_set(:@svn, true)
       expect(described_class.svn_available?).to be_truthy
     end
@@ -15,17 +21,16 @@ describe Utils do
   describe "#self.svn_remote_exists" do
     let(:url) { "https://dl.bintray.com/homebrew/mirror/" }
 
-    it "gives true when @svn is false" do
-      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
+    it "returns true when svn is not available" do
       described_class.instance_variable_set(:@svn, false)
       expect(described_class.svn_remote_exists(url)).to be_truthy
     end
 
-    it "gives false when url is obscure" do
-      expect(described_class.svn_remote_exists(url)).to be_falsy
+    it "returns false when remote does not exist" do
+      expect(described_class.svn_remote_exists(url)).to be_falsey
     end
 
-    it "gives true when quiet_system succeeds with given url" do
+    it "returns true when remote exists" do
       allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
       expect(described_class.svn_remote_exists(url)).to be_truthy
     end
--- a/Library/Homebrew/test/utils/svn_spec.rb
+++ b/Library/Homebrew/test/utils/svn_spec.rb
@@ -1,0 +1,33 @@
+require "utils/svn"
+
+describe Utils do
+  describe "#self.svn_available?" do
+    it "processes value when @svn is not defined" do
+      expect(described_class.svn_available?).to be_truthy
+    end
+
+    it "returns value of @svn when @svn is defined" do
+      described_class.instance_variable_set(:@svn, true)
+      expect(described_class.svn_available?).to be_truthy
+    end
+  end
+
+  describe "#self.svn_remote_exists" do
+    let(:url) { "https://dl.bintray.com/homebrew/mirror/" }
+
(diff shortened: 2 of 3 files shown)


Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "utils/svn"
+
+describe Utils do
+  describe "#self.svn_available?" do
+    it "processes value when @svn is not defined" do
```
reitermarkus: You shouldn't use instance variables to describe a test, since this doesn't really explain what it is supposed to do.

MikeMcQuaid:
Some test failures:
```
Failures:
  1) Utils#self.svn_remote_exists returns true when svn is not available
     Failure/Error: expect(described_class.svn_remote_exists(url)).to be_truthy
       expected: truthy value
            got: false
     # ./test/utils/svn_spec.rb:26:in `block (3 levels) in <top (required)>'
  2) Utils#self.svn_available? returns false if svn --version command does not succeed
     Failure/Error: expect(described_class.svn_available?).to be_falsey
       expected: falsey value
            got: true
     # ./test/utils/svn_spec.rb:12:in `block (3 levels) in <top (required)>'
Finished in 7 minutes 2 seconds (files took 4.75 seconds to load)
985 examples, 2 failures
Failed examples:
rspec ./test/utils/svn_spec.rb:24 # Utils#self.svn_remote_exists returns true when svn is not available
rspec ./test/utils/svn_spec.rb:10 # Utils#self.svn_available? returns false if svn --version command does not succeed
```

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  describe "#self.svn_remote_exists" do
+    let(:url) { "https://dl.bintray.com/homebrew/mirror/" }
+
+    it "returns true when svn is not available" do
+      described_class.instance_variable_set(:@svn, false)
+      expect(described_class.svn_remote_exists(url)).to be_truthy
+    end
+
+    it "returns false when remote does not exist" do
+      expect(described_class.svn_remote_exists(url)).to be_falsey
```
mistydemeo: Is this test performing network access?

mansimarkaur: Yes, it is performing network access. Do you think I should add a local repo instead?

mistydemeo: As we discussed on Slack - preferably, it'd be good to use a local repo here. If something has to use the network and we don't have another option, it'll need the `:needs_network` tag so that it only gets run with `brew tests --online`.

mansimarkaur:
@apjanke  Why was this PR closed?

apjanke:
I have no idea! I think I accidentally hit the wrong button on GitHub. Please ignore.

Library/Homebrew/test/utils/svn_spec.rb
mansimarkaur: The tests are behaving weirdly on my system. In the first run, only 1 test case fails, in the second 3 tests cases fail. I can kind of see this pattern here. Could you please try running the tests on your system and let me know what happens, @mistydemeo ?

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  describe "#self.svn_remote_exists" do
+    it "returns true when svn is not available" do
+      described_class.instance_variable_set(:@svn, false)
+      expect(described_class.svn_remote_exists("blah")).to be_truthy
+    end
+
+    it "returns false when remote does not exist" do
+      (mktmpdir/"blah").mkpath
```
MikeMcQuaid: Shouldn't need `mktmpdir` here; this is already in a temporary directory.

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      expect(described_class.svn_remote_exists("blah")).to be_falsey
+      FileUtils.rm_rf("blah")
```
MikeMcQuaid: This won't be run if the expectation fails.

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      expect(described_class.svn_available?).to be_truthy
+    end
+
+    it "returns false if svn --version command does not succeed" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
+      expect(described_class.svn_available?).to be_falsey
+    end
+
+    it "returns svn version if already set" do
+      described_class.instance_variable_set(:@svn, true)
```
reitermarkus: This will not reset the variable after the test, can you find a way to test the functionality that actually sets this to `true`? Otherwise, the only thing this is testing is whether `svn_available?` uses this variable.

MikeMcQuaid: Or alternatively: call a method that calls this so it's covered but not by a unit test.

Library/Homebrew/test/utils/svn_spec.rb
mansimarkaur: I've added a `begin ... ensure` block to take care of this now.

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,60 @@
+require "utils/svn"
+
+describe Utils do
+  describe "#self.svn_available?" do
+    it "returns true if svn --version command succeeds" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
```
MikeMcQuaid: I think it may be worth allowing the `svn` shim to be called. What fails if it is?

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      expect(described_class.svn_available?).to be_truthy
+    end
+
+    it "returns false if svn --version command does not succeed" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
```
MikeMcQuaid: I think it may be worth allowing the `svn` shim to be called. What fails if it is?

mansimarkaur: On some systems it might be successful and on some it might not be. So, I can just check that quiet_system will return either true, nil or false.

Library/Homebrew/test/utils/svn_spec.rb
mansimarkaur: This variable will be set only if `svn_available?` has been called successfully atleast once. So, there is no method which can ensure `@svn` will already exist if svn is not available on the system.
I'm removing `@svn` before each test now. I don't think svn_version already set can work without setting `@svn` because on a system without svn, this test won't work the way it should.

Library/Homebrew/test/utils/svn_spec.rb
MikeMcQuaid: On any macOS system there will always be a working `svn` and `git` so let's assume that during the tests, too, thanks.

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      if described_class.instance_variable_defined?(:@svn)
+        described_class.send(:remove_instance_variable, :@svn)
+      end
+    end
+
+    it "returns svn version if svn available" do
+      expect(described_class.svn_available?).to be_truthy
+    end
+
+    it "returns svn version if already set" do
```
mistydemeo: Isn't this test just doing the same thing as the previous one?

mansimarkaur: True. I guess I should remove it. I've removed such similar tests in git tests as well.  Thanks :)

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "utils/svn"
+
+describe Utils do
+  describe "#self.svn_available?" do
+    before(:each) do
+      if described_class.instance_variable_defined?(:@svn)
```
MikeMcQuaid: Need to try and figure out a way to not set/get instance variables here. If need be: delete the relevant tests that need them set given we're almost at the end of GSoC. Ideally: just call other methods that end up calling these ones.

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      expect(described_class.svn_remote_exists("blah")).to be_truthy
+    end
+
+    context "when svn is available" do
+      before do
+        allow(Utils).to receive(:svn_available?).and_return(true)
+        url.mkpath
+      end
+
+      let(:url) { mktmpdir/"blah" }
```
MikeMcQuaid: Why does this need to use `mktmpdir` and not write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

MikeMcQuaid: It's also really confusing to have a directory named `url`. Instead just use `HOMEBREW_CACHE` and no `let`.

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      it "returns false when remote does not exist" do
+        expect(described_class.svn_remote_exists(url)).to be_falsey
+      end
+
+      it "returns true when remote exists", :needs_network do
+        remote = "http://amethyst.yui.ne.jp/svn/pc98/np2/trunk/"
```
MikeMcQuaid: Consider using https://github.com/Homebrew/install as it's really small and we control it (you can clone GitHub repositories as Subversion repositories).

Library/Homebrew/test/utils/svn_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        svn = HOMEBREW_SHIMS_PATH/"scm/svn"
+
+        Dir.chdir url do
```
MikeMcQuaid: `HOMEBREW_CACHE.cd { system ... }`

Library/Homebrew/test/utils/svn_spec.rb
mansimarkaur: Here, I'm trying to remove the instance variable if already set because once `svn_available?` is called, it will set the `@svn` instance variable. To make these tests independent of each other, I need to ensure that the instance variable is not set beforehand. What do you suggest in this case?

Library/Homebrew/test/utils/svn_spec.rb
mansimarkaur: Using the above url, I'm getting the following: 
```
Error validating server certificate for 'https://github.com:443':
 - The certificate is not issued by a trusted authority. Use the
   fingerprint to validate the certificate manually!
Certificate information:
 - Hostname: github.com
 - Valid: from Mar 10 00:00:00 2016 GMT until May 17 12:00:00 2018 GMT
 - Issuer: DigiCert SHA2 Extended Validation Server CA, www.digicert.com, DigiCert Inc, US
 - Fingerprint: D7:9F:07:61:10:B3:92:93:E3:49:AC:89:84:5B:03:80:C1:9E:2F:8B
(R)eject, accept (t)emporarily or accept (p)ermanently?
```
In the tests, how do I ensure that I don't need to manually answer the above posed question of accepting the host?

Library/Homebrew/test/utils/svn_spec.rb
mansimarkaur: Got it. It's better to use a location that is already deleted after tests. Thanks!

Library/Homebrew/test/utils/svn_spec.rb
MikeMcQuaid: Perhaps adding a new method `clear_svn_location_cache` or similar and call that in the `before`.

mansimarkaur: Thanks for the suggestion! :)

Library/Homebrew/utils/svn.rb
```diff
@@ -1,4 +1,8 @@
 module Utils
+  def self.clear_svn_version_cache
+    remove_instance_variable(:@svn) if instance_variable_defined?(:@svn)
```
reitermarkus: This is never needed, except in tests, so this shouldn‘t exist in the actual code.

MikeMcQuaid: @reitermarkus I'm not sure how this gets resolved without adding a method here. What would you propose?

mansimarkaur: How about adding a method in one of the support modules for tests to clear any instance variable? As in a method whose argument is the instance variable to be cleared. @MikeMcQuaid  @reitermarkus  @mistydemeo

mansimarkaur:
There is something wrong with codecov. The base coverage is missing. Could you please help?

Library/Homebrew/utils/svn.rb
MikeMcQuaid: @mansimarkaur If by support modules you mean in the `spec_helper` or something then: yeh, that could work. @reitermarkus?

MikeMcQuaid: Also, there's probably some times this method should be called anyway such as e.g. after a new installation of `svn` (and same thing for `git`)

mansimarkaur: Yes, I meant adding a method in the `spec_helper` but if we want to call this cache clear method after a new installation, then we don't need to add it in the `spec_helper`. Which is a better alternative? Also, `utils/git.rb` already has a `Utils.clear_git_available_cache` to remove `@git`, `@git_path` and `@git_version`.

MikeMcQuaid: @mansimarkaur I think this is fine as-is, then, thanks.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -1228,7 +1228,9 @@ def audit_urls
         end
       elsif strategy <= SubversionDownloadStrategy
         next unless DevelopmentTools.subversion_handles_most_https_certificates?
-        unless Utils.svn_remote_exists url
+        if !Utils.svn_available?
+          problem "No valid version of svn found"
```
MikeMcQuaid: Actually: this shouldn't be a `problem` because it's not an issue with the formula that's being `audit`ed. Instead I'd just do `next unless Utils.svn_available?`

mansimarkaur: Got it! Made the change. Thanks :)

mistydemeo:
Thanks!

MikeMcQuaid:
Thanks for this @mansimarkaur, nice work!

PR #2952, opened 2017-07-27
Added tests for utils/analytics

mansimarkaur:

-----
Please review, @mistydemeo 
Thanks!

### DIFF
--- a/Library/Homebrew/test/utils/analytics_spec.rb
+++ b/Library/Homebrew/test/utils/analytics_spec.rb
@@ -44,8 +44,8 @@ describe Utils::Analytics do
       end
 
       it "returns nil when HOMEBREW_ANALYTICS_DEBUG is true" do
-        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = nil
-        ENV["HOMEBREW_NO_ANALYTICS"] = nil
+        ENV.delete("HOMEBREW_NO_ANALYTICS_THIS_RUN")
+        ENV.delete("HOMEBREW_NO_ANALYTICS")
         ENV["HOMEBREW_ANALYTICS_DEBUG"] = "true"
         expect(described_class.report_event("install", action)).to be_nil
       end
@@ -53,8 +53,8 @@ describe Utils::Analytics do
 
     context "when ENV vars are nil" do
       before do
-        ENV["HOMEBREW_NO_ANALYTICS"] = nil
-        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = nil
+        ENV.delete("HOMEBREW_NO_ANALYTICS_THIS_RUN")
+        ENV.delete("HOMEBREW_NO_ANALYTICS")
       end
 
       it "returns waiting thread when HOMEBREW_ANALYTICS_DEBUG is not set" do
--- a/Library/Homebrew/test/utils/analytics_spec.rb
+++ b/Library/Homebrew/test/utils/analytics_spec.rb
@@ -1,19 +1,27 @@
 require "utils/analytics"
+require "formula_installer"
 
 describe Utils::Analytics do
   describe "::os_prefix_ci" do
     context "when anonymous_os_prefix_ci is not set" do
+      before(:each) do
+        if described_class.instance_variable_defined?(:@anonymous_os_prefix_ci)
+          described_class.send(:remove_instance_variable, :@anonymous_os_prefix_ci)
+        end
+      end
+
       it "returns OS_VERSION and prefix when HOMEBREW_PREFIX is not /usr/local" do
+        stub_const("HOMEBREW_PREFIX", "blah")
         expect(described_class.os_prefix_ci).to include("#{OS_VERSION}, non-/usr/local")
       end
 
       it "includes CI when ENV['CI'] is set" do
-        allow(ENV).to receive(:[]).with("CI").and_return("true")
+        ENV["CI"] = "true"
         expect(described_class.os_prefix_ci).to include("CI")
       end
 
       it "does not include prefix when HOMEBREW_PREFIX is usr/local" do
-        allow(HOMEBREW_PREFIX).to receive(:to_s).and_return("/usr/local")
+        stub_const("HOMEBREW_PREFIX", "/usr/local")
         expect(described_class.os_prefix_ci).not_to include("non-/usr/local")
       end
     end
@@ -28,7 +36,39 @@ describe Utils::Analytics do
     end
   end
 
-  describe "::" do
-    
+  describe "::report_event" do
+    let(:f) { formula { url "foo-1.0" } }
+    let(:options) { FormulaInstaller.new(f).display_options(f) }
+    let(:action)  { "#{f.full_name} #{options}".strip }
+
+    context "when ENV vars is set" do
+      it "returns nil when HOMEBREW_NO_ANALYTICS is true" do
+        ENV["HOMEBREW_NO_ANALYTICS"] = "true"
+        expect(described_class.report_event("install", action)).to be_nil
+      end
+
+      it "returns nil when HOMEBREW_NO_ANALYTICS_THIS_RUN is true" do
+        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = "true"
+        expect(described_class.report_event("install", action)).to be_nil
+      end
+
+      it "returns nil when HOMEBREW_ANALYTICS_DEBUG is true" do
+        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = nil
+        ENV["HOMEBREW_NO_ANALYTICS"] = nil
+        ENV["HOMEBREW_ANALYTICS_DEBUG"] = "true"
+        expect(described_class.report_event("install", action)).to be_nil
+      end
+    end
+
+    context "when ENV vars are nil" do
+      before do
+        ENV["HOMEBREW_NO_ANALYTICS"] = nil
+        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = nil
+      end
+
+      it "returns nil when HOMEBREW_ANALYTICS_DEBUG is not set" do
+        expect(described_class.report_event("install", action)).to be_an_instance_of(Thread)
+      end
+    end
   end
 end
--- a/Library/Homebrew/test/utils/analytics_spec.rb
+++ b/Library/Homebrew/test/utils/analytics_spec.rb
@@ -25,15 +25,6 @@ describe Utils::Analytics do
         expect(described_class.os_prefix_ci).not_to include("non-/usr/local")
       end
     end
-
-    context "when anonymous_os_prefix_ci is set" do
-      let(:anonymous_os_prefix_ci) { "macOS 10.11.6, non-/usr/local, CI" }
-
-      it "returns anonymous_os_prefix_ci" do
-        described_class.instance_variable_set(:@anonymous_os_prefix_ci, anonymous_os_prefix_ci)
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
A test failure: 

```
Failures:
  1) Utils::Analytics::os_prefix_ci when anonymous_os_prefix_ci is not set returns OS_VERSION and prefix when HOMEBREW_PREFIX is not /usr/local
     Failure/Error: expect(described_class.os_prefix_ci).to include("#{OS_VERSION}, non-/usr/local")
       expected "macOS 10.12.5, CI" to include "macOS 10.12.5, non-/usr/local"
     # ./test/utils/analytics_spec.rb:7:in `block (4 levels) in <top (required)>'
Finished in 7 minutes 20 seconds (files took 4.54 seconds to load)
971 examples, 1 failure
Failed examples:
rspec ./test/utils/analytics_spec.rb:6 # Utils::Analytics::os_prefix_ci when anonymous_os_prefix_ci is not set returns OS_VERSION and prefix when HOMEBREW_PREFIX is not /usr/local
```

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        expect(described_class.os_prefix_ci).to include("#{OS_VERSION}, non-/usr/local")
+      end
+
+      it "includes CI when ENV['CI'] is set" do
+        CI = ENV["CI"]
+        ENV["CI"] = "true"
+
+        expect(described_class.os_prefix_ci).to include("CI")
+
+        ENV["CI"] = CI
```
mistydemeo: Can you put this in an `ensure`? If an exception were raised in this test there's a chance that we would skip this, affecting the results of other tests.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,93 @@
+require "utils/analytics"
+
+describe Utils::Analytics do
+  describe "::os_prefix_ci" do
+    context "when anonymous_os_prefix_ci is not set" do
+      it "returns OS_VERSION and prefix when HOMEBREW_PREFIX is not /usr/local" do
+        expect(described_class.os_prefix_ci).to include("#{OS_VERSION}, non-/usr/local")
```
mistydemeo: I think you'll have to override `HOMEBREW_PREFIX` for this test - I'm seeing inconsistent results for this and the other test where you're testing this.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,123 @@
+require "utils/analytics"
+
+describe Utils::Analytics do
+  describe "::os_prefix_ci" do
+    context "when anonymous_os_prefix_ci is not set" do
+      it "returns OS_VERSION and prefix when HOMEBREW_PREFIX is not /usr/local" do
+        h_prefix = HOMEBREW_PREFIX
+        HOMEBREW_PREFIX = mktmpdir/"blah"
```
MikeMcQuaid: Consider refactoring this into a method you can stub rather than changing a system constant.

MikeMcQuaid: Also, again, `mktmpdir` should never be needed because all tests run in a temporary directory.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+      it "includes CI when ENV['CI'] is set" do
+        CI = ENV["CI"]
+        ENV["CI"] = "true"
+
+        begin
+          expect(described_class.os_prefix_ci).to include("CI")
+        rescue
+        ensure
+          ENV["CI"] = CI
```
MikeMcQuaid: I believe ENV is already reset automatically after every test.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+        end
+      end
+
+      it "includes CI when ENV['CI'] is set" do
+        CI = ENV["CI"]
+        ENV["CI"] = "true"
+
+        begin
+          expect(described_class.os_prefix_ci).to include("CI")
+        rescue
```
MikeMcQuaid: Don't need this line.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+          HOMEBREW_PREFIX = h_prefix
+        end
+      end
+    end
+
+    context "when anonymous_os_prefix_ci is set" do
+      let(:anonymous_os_prefix_ci) { "macOS 10.11.6, non-/usr/local, CI" }
+
+      it "returns anonymous_os_prefix_ci" do
+        described_class.instance_variable_set(:@anonymous_os_prefix_ci, anonymous_os_prefix_ci)
```
MikeMcQuaid: Is there a way we can avoid setting an instance variable here?

Library/Homebrew/test/utils/analytics_spec.rb
reitermarkus: Yep, it is.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+describe Utils::Analytics do
+  describe "::os_prefix_ci" do
+    context "when anonymous_os_prefix_ci is not set" do
+      it "returns OS_VERSION and prefix when HOMEBREW_PREFIX is not /usr/local" do
+        h_prefix = HOMEBREW_PREFIX
+        HOMEBREW_PREFIX = mktmpdir/"blah"
+
+        begin
+          expect(described_class.os_prefix_ci).to include("#{OS_VERSION}, non-/usr/local")
+        rescue
```
apjanke: I think this `rescue` is unnecessary, since everything's handled in the `ensure` block.

Library/Homebrew/test/utils/analytics_spec.rb
reitermarkus: It may be needed here, since it has to be in another directory than the default. Anyways, use [`stub_const`](https://relishapp.com/rspec/rspec-mocks/v/3-6/docs/mutating-constants/stub-defined-constant) instead of redefining it.

MikeMcQuaid: We should still be able to use a subdirectory of the current temporary directory rather than a new `mktmpdir` (which also won't get cleaned up/deleted afterwards)

Library/Homebrew/test/utils/analytics_spec.rb
MikeMcQuaid: Ok, so the `ENV` changes in `ensure` can be removed.

Library/Homebrew/test/utils/analytics_spec.rb
reitermarkus: Actually, it will get cleaned up, since it will create a new directory under `HOMEBREW_TEMP`, which is cleaned up after each spec.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        end
+      end
+
+      it "returns OS_VERSION and prefix when HOMEBREW_PREFIX is not /usr/local" do
+        stub_const("HOMEBREW_PREFIX", "blah")
+        expect(described_class.os_prefix_ci).to include("#{OS_VERSION}, non-/usr/local")
+      end
+
+      it "includes CI when ENV['CI'] is set" do
+        ENV["CI"] = "true"
```
mansimarkaur: ENV is already reset automatically after every test, so, it will get restored.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    context "when ENV vars are nil" do
+      before do
+        ENV["HOMEBREW_NO_ANALYTICS"] = nil
+        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = nil
+      end
+
+      it "returns nil when HOMEBREW_ANALYTICS_DEBUG is not set" do
+        expect(described_class.report_event("install", action)).to be_an_instance_of(Thread)
```
mansimarkaur: I've changed the test name to depict that we are testing that it returns a waiting thread.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        expect(described_class.report_event("install", action)).to be_nil
+      end
+
+      it "returns nil when HOMEBREW_NO_ANALYTICS_THIS_RUN is true" do
+        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = "true"
+        expect(described_class.report_event("install", action)).to be_nil
+      end
+
+      it "returns nil when HOMEBREW_ANALYTICS_DEBUG is true" do
+        ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = nil
```
reitermarkus: For all of these, use `ENV.delete(…)` instead of `ENV[…] = nil`.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,101 @@
+require "utils/analytics"
+require "formula_installer"
+
+describe Utils::Analytics do
+  describe "::os_prefix_ci" do
+    context "when anonymous_os_prefix_ci is not set" do
+      before(:each) do
+        if described_class.instance_variable_defined?(:@anonymous_os_prefix_ci)
+          described_class.send(:remove_instance_variable, :@anonymous_os_prefix_ci)
```
MikeMcQuaid: Need to try and figure out a way to not set/get instance variables here. If need be: delete the relevant tests that need them set given we're almost at the end of GSoC. Ideally: just call other methods that end up calling these ones.

mansimarkaur:
@MikeMcQuaid please review. Thanks!
Also, here is it alright to add a remove cache method for `@anonymous_os_prefix`?

MikeMcQuaid:
> Also, here is it alright to add a remove cache method for @anonymous_os_prefix?

Yes, thanks.

Library/Homebrew/test/utils/analytics_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+    end
+
+    context "when ENV vars are nil" do
+      before do
+        ENV.delete("HOMEBREW_NO_ANALYTICS_THIS_RUN")
+        ENV.delete("HOMEBREW_NO_ANALYTICS")
+      end
+
+      it "returns waiting thread when HOMEBREW_ANALYTICS_DEBUG is not set" do
```
mistydemeo: I don't think anything actually uses this - I believe this is just an implementation detail. We don't need to test that.

Library/Homebrew/utils/analytics.rb
```diff
@@ -3,6 +3,10 @@
 module Utils
   module Analytics
     class << self
+      def clear_anonymous_os_prefix_ci_cache
+        remove_instance_variable(:@anonymous_os_prefix_ci) if instance_variable_defined?(:@anonymous_os_prefix_ci)
```
MikeMcQuaid: Can you split this onto two lines or add an early return (`return unless instance_variable_defined`...)? Thanks!

mistydemeo:
Thanks!

MikeMcQuaid:
Well done again @mansimarkaur 😍

PR #2955, opened 2017-07-27
Added tests for utils/git

mansimarkaur:

-----

### DIFF
--- a/Library/Homebrew/test/utils/git_spec.rb
+++ b/Library/Homebrew/test/utils/git_spec.rb
@@ -45,3 +45,22 @@ describe Git do
     end
   end
 end
+
+describe Utils do
+  describe "::git_available?" do
+    it "returns true if git --version command succeeds" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
+      expect(described_class.git_available?).to be_truthy
+    end
+
+    it "returns false if git --version command does not succeed" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
+      expect(described_class.git_available?).to be_falsey
+    end
+
+    it "returns git version if already set" do
+      described_class.instance_variable_set(:@git, true)
+      expect(described_class.git_available?).to be_truthy
+    end
+  end
+end
--- a/Library/Homebrew/test/utils/git_spec.rb
+++ b/Library/Homebrew/test/utils/git_spec.rb
@@ -3,45 +3,44 @@ require "utils/git"
 describe Git do
   before(:all) do
     git = HOMEBREW_SHIMS_PATH/"scm/git"
-    file = "lib/blah.rb"
-    repo = Pathname.new("repo")
 
-    (repo/"lib").mkpath
+    @file = "lib/blah.rb"
+    @repo = Pathname.new("repo")
+
+    (@repo/"lib").mkpath
     system git, "init"
-    FileUtils.touch("repo/#{file}")
+    FileUtils.touch(@repo/@file)
 
-    File.open(repo/file, "w") { |f| f.write("blah") }
-    system git, "add", repo/file
+    File.open(@repo/@file, "w") { |f| f.write("blah") }
+    system git, "add", @repo/@file
     system git, "commit", "-m", "'File added'"
     @h1 = `git rev-parse HEAD`
 
-    File.open(repo/file, "w") { |f| f.write("brew") }
-    system git, "add", repo/file
+    File.open(@repo/@file, "w") { |f| f.write("brew") }
+    system git, "add", @repo/@file
     system git, "commit", "-m", "'written to File'"
     @h2 = `git rev-parse HEAD`
   end
 
-  let(:file) { "lib/blah.rb" }
-  let(:repo) { Pathname.new("repo") }
   let(:hash1) { @h1[0..6] }
   let(:hash2) { @h2[0..6] }
 
   after(:all) do
-    FileUtils.rm_rf("repo")
+    FileUtils.rm_rf(@repo)
   end
 
   describe "#last_revision_commit_of_file" do
     it "gives last revision commit when before_commit is nil" do
       expect(
-        described_class.last_revision_commit_of_file(repo, file),
+        described_class.last_revision_commit_of_file(@repo, @file),
       ).to eq(hash1)
     end
 
     it "gives revision commit based on before_commit when it is not nil" do
       expect(
-        described_class.last_revision_commit_of_file(repo,
-                                                    file,
-                                                    before_commit: "0..3"),
+        described_class.last_revision_commit_of_file(@repo,
+                                                    @file,
+                                                    before_commit: hash2),
       ).to eq(hash2)
     end
   end
@@ -49,14 +48,14 @@ describe Git do
   describe "#last_revision_of_file" do
     it "returns last revision of file" do
       expect(
-        described_class.last_revision_of_file(repo,
-                                              repo/file),
+        described_class.last_revision_of_file(@repo,
+                                              @repo/@file),
       ).to eq("blah")
     end
 
     it "returns last revision of file based on before_commit" do
       expect(
-        described_class.last_revision_of_file(repo, repo/file,
+        described_class.last_revision_of_file(@repo, @repo/@file,
                                               before_commit: "0..3"),
       ).to eq("brew")
     end
@@ -72,100 +71,71 @@ describe Utils do
 
   describe "::git_available?" do
     it "returns true if git --version command succeeds" do
-      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
       expect(described_class.git_available?).to be_truthy
     end
 
     it "returns false if git --version command does not succeed" do
-      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
       expect(described_class.git_available?).to be_falsey
-    end
-
-    it "returns git version if already set" do
-      described_class.instance_variable_set(:@git, true)
-      expect(described_class.git_available?).to be_truthy
-      described_class.instance_variable_set(:@git, nil)
     end
   end
 
   describe "::git_path" do
-    context "when git is not available" do
-      before do
-        described_class.instance_variable_set(:@git, false)
-      end
-
-      it "returns nil" do
-        expect(described_class.git_path).to eq(nil)
-      end
+    it "returns nil when git is not available" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
+      expect(described_class.git_path).to eq(nil)
     end
 
-    context "when git is available" do
-      before do
-        described_class.instance_variable_set(:@git, true)
-      end
-
-      it "returns path of git" do
-        expect(described_class.git_path).to end_with("git")
-      end
-
-      it "returns git_path if already set" do
-        described_class.instance_variable_set(:@git_path, "git")
-        expect(described_class.git_path).to eq("git")
-        described_class.instance_variable_set(:@git_path, nil)
-      end
+    it "returns path of git when git is available" do
+      expect(described_class.git_path).to end_with("git")
     end
   end
 
   describe "::git_version" do
-    context "when git is not available" do
-      before do
-        described_class.instance_variable_set(:@git, false)
-      end
-
-      it "returns nil" do
-        expect(described_class.git_path).to eq(nil)
-      end
+    it "returns nil when git is not available" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
+      expect(described_class.git_path).to eq(nil)
     end
 
-    context "when git is available" do
+    it "returns version of git when git is available" do
+      expect(described_class.git_version).not_to be_nil
+    end
+  end
+
+  describe "::ensure_git_installed!" do
+    it "returns nil if git already available" do
+      expect(described_class.ensure_git_installed!).to be_nil
+    end
+
+    context "when git is not already available" do
       before do
-        described_class.instance_variable_set(:@git, true)
+        stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
       end
 
-      it "returns version of git" do
-        expect(described_class.git_version).not_to be_nil
+      it "can't install brewed git if homebrew/core is unavailable" do
+        allow_any_instance_of(Pathname).to receive(:directory?).and_return(false)
+        expect { described_class.ensure_git_installed! }.to raise_error("Git is unavailable")
       end
 
-      it "returns git_version if already set" do
-        described_class.instance_variable_set(:@git_version, "v1")
-        expect(described_class.git_version).to eq("v1")
-        described_class.instance_variable_set(:@git_version, nil)
+      it "raises error if can't install git" do
+        stub_const("HOMEBREW_BREW_FILE", mktmpdir/"brew")
+        expect { described_class.ensure_git_installed! }.to raise_error("Git is unavailable")
+      end
+
+      it "installs git" do
+        allow(Homebrew).to receive(:_system).with(any_args).and_return(true)
+        described_class.ensure_git_installed!
       end
     end
   end
 
   describe "::git_remote_exists" do
-    context "when git is not available" do
-      before do
-        described_class.instance_variable_set(:@git, false)
-      end
-
-      it "returns true" do
-        expect(described_class.git_remote_exists("blah")).to be_truthy
-      end
+    it "returns true when git is not available" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
+      expect(described_class.git_remote_exists("blah")).to be_truthy
     end
 
     context "when git is available" do
-      before(:all) do
-        described_class.instance_variable_set(:@git, true)
-      end
-
-      after(:all) do
-        if described_class.instance_variable_defined?(:@git)
-          described_class.send(:remove_instance_variable, :@git)
-        end
-      end
-
       it "returns true when git remote exists", :needs_network do
         git = HOMEBREW_SHIMS_PATH/"scm/git"
         repo = Pathname.new("hey")
@@ -173,10 +143,10 @@ describe Utils do
 
         system "cd", repo
         system git, "init"
-        system git, "remote", "add", "origin", "git@github.com:Homebrew/brew"
+        system git, "remote", "add", "origin", "http://github.com/Homebrew/homebrew.github.io"
         system "cd .."
 
-        expect(described_class.git_remote_exists("git@github.com:Homebrew/brew")).to be_truthy
+        expect(described_class.git_remote_exists("http://github.com/Homebrew/homebrew.github.io")).to be_truthy
 
         FileUtils.rm_rf(repo)
       end
@@ -191,24 +161,15 @@ describe Utils do
     it "removes @git_path and @git_version if defined" do
       described_class.clear_git_available_cache
 
-      expect(@git_path).to be_nil
-      expect(@git_version).to be_nil
+      expect(described_class.instance_variable_get(:@git_path)).to be_nil
+      expect(described_class.instance_variable_get(:@git_version)).to be_nil
     end
 
     it "removes @git if defined" do
-      described_class.instance_variable_set(:@git, true)
+      described_class.git_available?
+      described_class.clear_git_available_cache
 
-      begin
-        described_class.clear_git_available_cache
-
-        expect(@git).to be_nil
-        expect(@git_path).to be_nil
-        expect(@git_version).to be_nil
-      ensure
-        if described_class.instance_variable_defined?(:@git)
-          described_class.send(:remove_instance_variable, :@git)
-        end
-      end
+      expect(described_class.instance_variable_get(:@git)).to be_nil
     end
   end
 end
--- a/Library/Homebrew/test/utils/git_spec.rb
+++ b/Library/Homebrew/test/utils/git_spec.rb
@@ -1,0 +1,47 @@
+require "utils/git"
+
+describe Git do
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+    file = "lib/blah.rb"
+    repo = Pathname.new("repo")
+    FileUtils.mkpath("repo/lib")
+    `#{git} init`
+    FileUtils.touch("repo/#{file}")
+    File.open(repo.to_s+"/"+file, "w") { |f| f.write("blah") }
+    `#{git} add repo/#{file}`
+    `#{git} commit -m"File added"`
+    @hash1 = `git rev-parse HEAD`
+    File.open(repo.to_s+"/"+file, "w") { |f| f.write("brew") }
+    `#{git} add repo/#{file}`
+    `#{git} commit -m"written to File"`
+    @hash2 = `git rev-parse HEAD`
+  end
+
+  let(:file) { "lib/blah.rb" }
+  let(:repo) { Pathname.new("repo") }
+
+  # after(:all) do
+  #   FileUtils.rm_rf("repo")
+  # end
+
+  describe "#last_revision_commit_of_file" do
+    it "sets args as --skip=1 when before_commit is nil" do
+      expect(described_class.last_revision_commit_of_file(repo, file)).to eq(@hash1[0..6])
+    end
+
+    it "sets args as --skip=1 when before_commit is nil" do
+      expect(described_class.last_revision_commit_of_file(repo, file, before_commit: "0..3")).to eq(@hash2[0..6])
+    end
+  end
+
+  describe "#last_revision_of_file" do
+    it "returns last revision of file" do
+      expect(described_class.last_revision_of_file(repo, repo.to_s+"/"+file)).to eq("blah")
+    end
+
+    it "returns last revision of file based on before_commit" do
+      expect(described_class.last_revision_of_file(repo, repo.to_s+"/"+file, before_commit: "0..3")).to eq("brew")
+    end
+  end
+end
--- a/Library/Homebrew/test/utils/git_spec.rb
+++ b/Library/Homebrew/test/utils/git_spec.rb
@@ -21,9 +21,9 @@ describe Git do
   let(:file) { "lib/blah.rb" }
   let(:repo) { Pathname.new("repo") }
 
-  # after(:all) do
-  #   FileUtils.rm_rf("repo")
-  # end
+  after(:all) do
+    FileUtils.rm_rf("repo")
+  end
 
   describe "#last_revision_commit_of_file" do
     it "sets args as --skip=1 when before_commit is nil" do
@@ -61,6 +61,116 @@ describe Utils do
     it "returns git version if already set" do
       described_class.instance_variable_set(:@git, true)
       expect(described_class.git_available?).to be_truthy
+      described_class.instance_variable_set(:@git, nil)
+    end
+  end
+
+  describe "::git_path" do
+    context "when git is not available" do
+      before do
+        described_class.instance_variable_set(:@git, false)
+      end
+
+      it "returns" do
+        expect(described_class.git_path).to eq(nil)
+      end
+    end
+
+    context "when git is available" do
+      before do
+        described_class.instance_variable_set(:@git, true)
+      end
+
+      it "returns path of git" do
+        expect(described_class.git_path).to eq("/Applications/Xcode.app/Contents/Developer/usr/bin/git")
+      end
+
+      it "returns git_path if already set" do
+        described_class.instance_variable_set(:@git_path, "git")
+        expect(described_class.git_path).to eq("git")
+        described_class.instance_variable_set(:@git_path, nil)
+      end
+    end
+  end
+
+  describe "::git_version" do
+    context "when git is not available" do
+      before do
+        described_class.instance_variable_set(:@git, false)
+      end
+
+      it "returns" do
+        expect(described_class.git_path).to eq(nil)
+      end
+    end
+
+    context "when git is available" do
+      before do
+        described_class.instance_variable_set(:@git, true)
+      end
+
+      it "returns version of git" do
+        expect(described_class.git_version).not_to be_nil
+      end
+
+      it "returns git_version if already set" do
+        described_class.instance_variable_set(:@git_version, "v1")
+        expect(described_class.git_version).to eq("v1")
+        described_class.instance_variable_set(:@git_version, nil)
+      end
+    end
+  end
+
+  describe "::git_remote_exists" do
+    context "when git is not available" do
+      before do
+        described_class.instance_variable_set(:@git, false)
+      end
+
+      it "returns true" do
+        expect(described_class.git_remote_exists("blah")).to be_truthy
+      end
+    end
+
+    context "when git is available" do
+      before(:all) do
+        described_class.instance_variable_set(:@git, true)
+        git = HOMEBREW_SHIMS_PATH/"scm/git"
+        @repo = Pathname.new("hey")
+        FileUtils.mkpath("hey")
+        `cd #{@repo}`
+        `#{git} init`
+        `#{git} remote add origin git@github.com:Homebrew/brew`
+        `cd ..`
+      end
+
+      after(:all) do
+        FileUtils.rm_rf(@repo)
+      end
+
+      it "returns true when git remote exists" do
+        expect(described_class.git_remote_exists("git@github.com:Homebrew/brew")).to be_truthy
+      end
+
+      it "returns false when git remote does not exist" do
+        expect(described_class.git_remote_exists("blah")).to be_falsey
+      end
+    end
+  end
+
+  describe "::clear_git_available_cache" do
+    it "removes @git_path and @git_version if defined" do
+      described_class.clear_git_available_cache
+      expect(@git_path).to be_nil
+      expect(@git_version).to be_nil
+    end
+
+    it "removes @git if defined" do
+      described_class.instance_variable_set(:@git, true)
+      described_class.clear_git_available_cache
+      expect(@git).to be_nil
+      expect(@git_path).to be_nil
+      expect(@git_version).to be_nil
     end
   end
 end
--- a/Library/Homebrew/test/utils/git_spec.rb
+++ b/Library/Homebrew/test/utils/git_spec.rb
@@ -4,37 +4,36 @@ describe Git do
   before(:each) do
     git = HOMEBREW_SHIMS_PATH/"scm/git"
 
-    @file = "blah.rb"
-
     HOMEBREW_CACHE.cd do
       system git, "init"
 
-      File.open(@file, "w") { |f| f.write("blah") }
-      system git, "add", HOMEBREW_CACHE/@file
+      File.open(file, "w") { |f| f.write("blah") }
+      system git, "add", HOMEBREW_CACHE/file
       system git, "commit", "-m", "'File added'"
       @h1 = `git rev-parse HEAD`
 
-      File.open(@file, "w") { |f| f.write("brew") }
-      system git, "add", HOMEBREW_CACHE/@file
+      File.open(file, "w") { |f| f.write("brew") }
+      system git, "add", HOMEBREW_CACHE/file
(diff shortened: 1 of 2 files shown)


Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
+require "utils/git"
+
+describe Git do
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+    file = "lib/blah.rb"
+    repo = Pathname.new("repo")
+    FileUtils.mkpath("repo/lib")
+    `#{git} init`
```
mistydemeo: Can you use `system git, "init"`? We only use backticks when we want the output as a string.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+  after(:all) do
+    FileUtils.rm_rf("repo")
+  end
+
+  describe "#last_revision_commit_of_file" do
+    it "sets args as --skip=1 when before_commit is nil" do
+      expect(described_class.last_revision_commit_of_file(repo, file)).to eq(@hash1[0..6])
+    end
+
+    it "sets args as --skip=1 when before_commit is nil" do
```
mistydemeo: Is this test title correct? `before_commit` isn't nil here.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      expect(described_class.last_revision_commit_of_file(repo, file, before_commit: "0..3")).to eq(@hash2[0..6])
+    end
+  end
+
+  describe "#last_revision_of_file" do
+    it "returns last revision of file" do
+      expect(described_class.last_revision_of_file(repo, repo.to_s+"/"+file)).to eq("blah")
```
mistydemeo: Nitpick, but how about `repo.join("file").to_s` instead?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    end
+  end
+
+  describe "::git_path" do
+    context "when git is not available" do
+      before do
+        described_class.instance_variable_set(:@git, false)
+      end
+
+      it "returns" do
```
mistydemeo: Returns what?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      end
+    end
+
+    context "when git is available" do
+      before do
+        described_class.instance_variable_set(:@git, true)
+      end
+
+      it "returns path of git" do
+        expect(described_class.git_path).to eq("/Applications/Xcode.app/Contents/Developer/usr/bin/git")
```
mistydemeo: We can't guarantee this path will always be this exact string - it could also be the Homebrew-installed git, the CLT-provided git, etc. For example, on my system:

```
       expected: "/Applications/Xcode.app/Contents/Developer/usr/bin/git"
            got: "/usr/local/bin/git"
```

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  describe "::git_remote_exists" do
+    context "when git is not available" do
+      before do
+        described_class.instance_variable_set(:@git, false)
+      end
+
+      it "returns true" do
+        expect(described_class.git_remote_exists("blah")).to be_truthy
```
mistydemeo: You're testing the right behaviour, but this one surprises me! I didn't realize this was the case.

mistydemeo:
Looks like we've got one failure:

```
  1) Utils::git_remote_exists when git is available returns true when git remote exists
     Failure/Error: expect(described_class.git_remote_exists(@repo)).to be_truthy
       expected: truthy value
            got: false
     # ./test/utils/git_spec.rb:151:in `block (4 levels) in <top (required)>'
```

mistydemeo:
I get this additional failure locally:

```
  2) Utils::git_available? returns false if git --version command does not succeed
     Failure/Error: expect(described_class.git_available?).to be_falsey

       expected: falsey value
            got: true
     # ./test/utils/git_spec.rb:58:in `block (3 levels) in <top (required)>'
```

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: So, should I just test if the returned value inlcudes only "git"?

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: Using system to execute these git commands are failing the tests. git add and commit aren't being executed properly.

Library/Homebrew/test/utils/git_spec.rb
mistydemeo: If we can stub it, let's do that - then it'll behave consistently even on systems that don't actually have git.

Library/Homebrew/test/utils/git_spec.rb
mistydemeo: What's failing about them? It'd be good to fix that, rather than ignore the failure using backticks.

mistydemeo:
A few failures on Linux:

```
  1) Utils::git_path when git is available returns path of git
     Failure/Error: expect(described_class.git_path).to eq("/Applications/Xcode.app/Contents/Developer/usr/bin/git")
       expected: "/Applications/Xcode.app/Contents/Developer/usr/bin/git"
            got: "/usr/bin/git"
```

As noted earlier, we can't count on the Xcode path to be there on every system.

```
  2) Utils::git_remote_exists when git is available returns true when git remote exists
     Failure/Error: expect(described_class.git_remote_exists("git@github.com:Homebrew/brew")).to be_truthy
       expected: truthy value
            got: false
```

Linuxbrew uses a different remote.

```
  3) Utils::git_available? returns true if git --version command succeeds
     Failure/Error: expect(described_class.git_available?).to be_truthy
       expected: truthy value
            got: false
```

Not sure what's going on here!

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: I've corrected it now.

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: Returns `nil`
Made the change

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+    file = "lib/blah.rb"
+    repo = Pathname.new("repo")
+    FileUtils.mkpath("repo/lib")
+    system "#{git} init"
+    FileUtils.touch("repo/#{file}")
+    File.open(repo.join("file").to_s, "w") { |f| f.write("blah") }
+    system "#{git} add repo/#{file}"
+    system "#{git} commit -m'File added'"
+    @hash1 = `git rev-parse HEAD`
```
MikeMcQuaid: Favour using `let` over `@` instance variables.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+
+describe Git do
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+    file = "lib/blah.rb"
+    repo = Pathname.new("repo")
+    FileUtils.mkpath("repo/lib")
+    system "#{git} init"
+    FileUtils.touch("repo/#{file}")
+    File.open(repo.join("file").to_s, "w") { |f| f.write("blah") }
```
MikeMcQuaid: `(repo/"file").write "blah"`

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,181 @@
+require "utils/git"
+
+describe Git do
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+    file = "lib/blah.rb"
+    repo = Pathname.new("repo")
+    FileUtils.mkpath("repo/lib")
```
MikeMcQuaid: `(repo/"lib").mkpath`

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    system "#{git} init"
+    FileUtils.touch("repo/#{file}")
+    File.open(repo.join("file").to_s, "w") { |f| f.write("blah") }
+    system "#{git} add repo/#{file}"
```
MikeMcQuaid: `system "git", "add", "repo/#{file}"` and same with all other calls

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+  let(:file) { "lib/blah.rb" }
+  let(:repo) { Pathname.new("repo") }
+
+  after(:all) do
+    FileUtils.rm_rf("repo")
+  end
+
+  describe "#last_revision_commit_of_file" do
+    it "sets args as --skip=1 when before_commit is nil" do
+      expect(described_class.last_revision_commit_of_file(repo, file)).to eq(@hash1[0..6])
```
MikeMcQuaid: Consider setting up the `hash1` variable to avoid needing to `[0..6]` in the test

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    end
+
+    it "sets args as --skip=1 when before_commit is nil" do
+      expect(described_class.last_revision_commit_of_file(repo, file, before_commit: "0..3")).to eq(@hash2[0..6])
+    end
+  end
+
+  describe "#last_revision_of_file" do
+    it "returns last revision of file" do
+      expect(described_class.last_revision_of_file(repo, repo.join("file").to_s)).to eq("blah")
```
MikeMcQuaid: Use `Pathname#/` rather than `.join` everywhere

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    end
+
+    it "returns last revision of file based on before_commit" do
+      expect(described_class.last_revision_of_file(repo, repo.join("file").to_s, before_commit: "0..3")).to eq("brew")
```
MikeMcQuaid: Split this and other long lines onto multiple lines.

MikeMcQuaid:
CC @reitermarkus for RSpec form check.

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: Any specific reasons for preferring `system "git", "add", "repo/#{file}"` over ` system "#{git} add repo/#{file}"` ?

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: What are your views about adding a brew style violation for a line length greater than a certain number of characters? @mistydemeo @MikeMcQuaid

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: I'll have to explicitly close the file as well if I write open and write separately. Why do you think `(repo/"file").write "blah"` is a better way?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+        described_class.last_revision_of_file(repo, repo/file,
+                                              before_commit: "0..3"),
+      ).to eq("brew")
+    end
+  end
+end
+
+describe Utils do
+  describe "::git_available?" do
+    it "returns true if git --version command succeeds" do
```
mansimarkaur: This test is behaving weird on my system. It passes sometimes and fails at other times.
Basically, `git_available?` is returning `nil` and `false` too at some times.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
+require "utils/git"
+
+describe Git do
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+    file = "lib/blah.rb"
+    repo = Pathname.new("repo")
+
+    (repo/"lib").mkpath
+    system "#{git} init"
```
MikeMcQuaid: `system git, "init"`

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+    FileUtils.touch("repo/#{file}")
+
+    File.open(repo/file, "w") { |f| f.write("blah") }
```
MikeMcQuaid: `(repo/file).write("blah")`. That doesn't require an explicit close because it's handled automatically by `Pathname#write`.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+    system git, "add", repo/file
+    system git, "commit", "-m", "'File added'"
+    @h1 = `git rev-parse HEAD`
+
+    File.open(repo/file, "w") { |f| f.write("brew") }
```
MikeMcQuaid: `(repo/file).write("blah")`. That doesn't require an explicit close because it's handled automatically by `Pathname#write`.

Library/Homebrew/test/utils/git_spec.rb
MikeMcQuaid: Don't see the change here?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+                                              before_commit: "0..3"),
+      ).to eq("brew")
+    end
+  end
+end
+
+describe Utils do
+  describe "::git_available?" do
+    it "returns true if git --version command succeeds" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
```
MikeMcQuaid: I think it may be worth allowing the `git` shim to be called. What fails if it is?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+      end
+    end
+
+    context "when git is available" do
+      before do
+        described_class.instance_variable_set(:@git, true)
+      end
+
+      it "returns path of git" do
+        allow(Utils).to receive(:popen_read).with(HOMEBREW_SHIMS_PATH/"scm/git",
```
MikeMcQuaid: I think it may be worth allowing the `git` shim to be called. What fails if it is?

Library/Homebrew/test/utils/git_spec.rb
reitermarkus: `its(:git_path) { is_expected.to be nil }`

Library/Homebrew/utils/git.rb
```diff
@@ -4,7 +4,7 @@ module Git
   module_function
 
   def last_revision_commit_of_file(repo, file, before_commit: nil)
-    args = [before_commit.nil? ? "--skip=1" : before_commit.split("..").first]
+    args = ["--skip=#{before_commit.nil? ? 1 : before_commit.split("..").first}"]
```
reitermarkus: This is not the same as before.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+      end
+    end
+
+    context "when git is available" do
+      before(:all) do
+        @initial_var = described_class.instance_variable_get(:@git)
+        described_class.instance_variable_set(:@git, true)
+
+        git = HOMEBREW_SHIMS_PATH/"scm/git"
+        @repo = Pathname.new("hey")
```
reitermarkus: Don't use instance variables. Use `let(:repo) { … }`.

Library/Homebrew/utils/git.rb
mansimarkaur: The previous behaviour was wrong. If `before_commit` is not nil, it did not work correctly.

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: How do I suggest I use `let` with `before(:all)` and `after(:all)` blocks?

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: It returns different path values for different systems. The mocking attempts to avoid that.

MikeMcQuaid: @mansimarkaur I think we could just check for `ends_with?("git")` or similar. That would make this partially an integration test for the shim too, which would be great.

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: I get the following error if I use `Pathname#write` : 
```
Failure/Error: (repo/file).write("blah")
RuntimeError:
Will not overwrite repo/lib/blah.rb
```

Library/Homebrew/test/utils/git_spec.rb
reitermarkus: Why does it have to be `before(:all)`?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,209 @@
...
+        @repo.mkpath
+
+        system "cd", @repo
+        system git, "init"
+        if OS.linux?
+          system git, "remote", "add", "origin", "git@github.com:Linuxbrew/brew"
```
reitermarkus: This isn't needed, since the test isn't OS-specific.

Library/Homebrew/utils/git.rb
reitermarkus: How exactly was it wrong?

mansimarkaur: If `before_commit` is `nil`, `args` become `"--skip=1"` but if `before_commit` is say, `"0..3"`, then args should be `"--skip=0" while previously `args` became just `0`.

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: Realised I can do away with the before(:all) block. Hence, removed it. Thanks :)

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+    end
+
+    it "can't install brewed git if homebrew/core is unavailable" do
+      described_class.instance_variable_set(:@git, false)
+      allow_any_instance_of(Pathname).to receive(:directory?).and_return(false)
+      expect { described_class.ensure_git_installed! }.to raise_error("Git is unavailable")
+    end
+
+    it "installs git if homebrew/core is available and git is not available already" do
+      # WIP
```
mansimarkaur: If I mock `git_available?` to return false and also mock the existence of `CoreTap.instance.formula`, running `safe_system` to install git raises an error since git is already installed on my system. What could be the workaround to mock a system that doesn't have brew installed git?
Thanks!
@mistydemeo @MikeMcQuaid @reitermarkus

mistydemeo: `expect(described_class).to receive(:safe_system).with(HOMEBREW_BREW_FILE, "install", "git").and_return(nil)` should work to mock the `safe_system` call.

mansimarkaur: If git was installed successfully, safe_system would return `true` and not `nil`. How should we check for that since git will be pre-installed and an error will come up?

mistydemeo: You can mock it to return `true`! It doesn't look like anything actually checks the return value of `safe_system` though - the error handling is in the `system` call it's wrapping, and since we're skipping that by mocking `safe_system` it should be okay.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+    file = "lib/blah.rb"
+    repo = Pathname.new("repo")
+
+    (repo/"lib").mkpath
+    system git, "init"
+    FileUtils.touch("repo/#{file}")
+
+    File.open(repo/file, "w") { |f| f.write("blah") }
```
MikeMcQuaid: `(repo/file).write("blah")` and skip the `touch` above.

Library/Homebrew/test/utils/git_spec.rb
MikeMcQuaid: I'd still say this form with a previous `unlink` is probably nicer but feel free to leave as-is and just change the case above.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(true)
+      expect(described_class.git_available?).to be_truthy
+    end
+
+    it "returns false if git --version command does not succeed" do
+      allow_any_instance_of(Process::Status).to receive(:success?).and_return(false)
+      expect(described_class.git_available?).to be_falsey
+    end
+
+    it "returns git version if already set" do
```
MikeMcQuaid: I think this gets a bit too far into the implementation details. As @reitermarkus said: we shouldn't need to set instance variables in a test.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      described_class.instance_variable_set(:@git, true)
+      expect(described_class.git_available?).to be_truthy
+      described_class.instance_variable_set(:@git, nil)
+    end
+  end
+
+  describe "::git_path" do
+    context "when git is not available" do
+      before do
+        described_class.instance_variable_set(:@git, false)
```
MikeMcQuaid: Is it possible instead to stub the lookup so it fails to find Git rather than overriding an instance variable?

mansimarkaur: Thanks for suggesting this, @MikeMcQuaid ! :)

Library/Homebrew/utils/git.rb
reitermarkus: That's the correct behaviour. When no commit range is given, it should just skip the latest commit, i.e. `--skip=1`, otherwise, it should pass the first commit of the commit range to `git log`, so it will only show commits before that commit `before_commit`.

mansimarkaur: So, then do you think the modifications made are correct? @reitermarkus

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: This gives the following error:
```
Failure/Error: (repo/file).write("brew")

RuntimeError:
       Will not overwrite repo/lib/blah.rb
```

Library/Homebrew/utils/git.rb
reitermarkus: No, it should either be `"--skip=1"`, or when there is a commit range, i. e. `before_commit = "aabbcc...ddeeff"`, it should be just `"aabbcc"`, not `"--skip=aabbcc"`.

Library/Homebrew/test/utils/git_spec.rb
reitermarkus: Have you removed the `touch` above?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+    it "returns false if git --version command does not succeed" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
+      expect(described_class.git_available?).to be_falsey
+    end
+  end
+
+  describe "::git_path" do
+    context "when git is not available" do
+      before do
+        stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
reitermarkus: Move this into the `it` block, since it is only used once.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+      it "returns path of git" do
+        expect(described_class.git_path).to end_with("git")
+      end
+    end
+  end
+
+  describe "::git_version" do
+    context "when git is not available" do
+      before do
+        stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
reitermarkus: Move this into the it block, since it is only used once.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+        allow(Homebrew).to receive(:_system).with(any_args).and_return(true)
+        described_class.ensure_git_installed!
+      end
+    end
+  end
+
+  describe "::git_remote_exists" do
+    context "when git is not available" do
+      before do
+        stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
reitermarkus: Move this into the `it` block, since it is only used once.

Library/Homebrew/test/utils/git_spec.rb
mansimarkaur: Yes. I have removed the `touch`. I still get the above mentioned error.

Library/Homebrew/utils/git.rb
mansimarkaur: I understand it now. I've reverted the change.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+describe Git do
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+
+    @file = "lib/blah.rb"
+    @repo = Pathname.new("repo")
+
+    (@repo/"lib").mkpath
+    system git, "init"
+    FileUtils.touch(@repo/@file)
```
MikeMcQuaid: You don't need to `touch` here given you're immediately writing to it.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
+require "utils/git"
+
+describe Git do
+  before(:all) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+
+    @file = "lib/blah.rb"
+    @repo = Pathname.new("repo")
```
MikeMcQuaid: Could put this in one of these dirs to avoid the manual cleanup (and be consistent): https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+                                              before_commit: "0..3"),
+      ).to eq("brew")
+    end
+  end
+end
+
+describe Utils do
+  before(:each) do
+    if described_class.instance_variable_defined?(:@git)
+      described_class.send(:remove_instance_variable, :@git)
```
MikeMcQuaid: Need to try and figure out a way to not set/get instance variables here. If need be: delete the relevant tests that need them set given we're almost at the end of GSoC. Ideally: just call other methods that end up calling these ones.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    it "returns false if git --version command does not succeed" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
+      expect(described_class.git_available?).to be_falsey
+    end
+  end
+
+  describe "::git_path" do
+    it "returns nil when git is not available" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
MikeMcQuaid: Instead of `mktmpdir` write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    end
+
+    it "returns path of git when git is available" do
+      expect(described_class.git_path).to end_with("git")
+    end
+  end
+
+  describe "::git_version" do
+    it "returns nil when git is not available" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
MikeMcQuaid: Instead of `mktmpdir` write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    end
+  end
+
+  describe "::git_available?" do
+    it "returns true if git --version command succeeds" do
+      expect(described_class.git_available?).to be_truthy
+    end
+
+    it "returns false if git --version command does not succeed" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
MikeMcQuaid: Instead of `mktmpdir` write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  describe "::ensure_git_installed!" do
+    it "returns nil if git already available" do
+      expect(described_class.ensure_git_installed!).to be_nil
+    end
+
+    context "when git is not already available" do
+      before do
+        stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
MikeMcQuaid: Instead of `mktmpdir` write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      end
+
+      it "can't install brewed git if homebrew/core is unavailable" do
+        allow_any_instance_of(Pathname).to receive(:directory?).and_return(false)
+        expect { described_class.ensure_git_installed! }.to raise_error("Git is unavailable")
+      end
+
+      it "raises error if can't install git" do
+        stub_const("HOMEBREW_BREW_FILE", mktmpdir/"brew")
```
MikeMcQuaid: Instead of `mktmpdir` write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      it "installs git" do
+        allow(Homebrew).to receive(:_system).with(any_args).and_return(true)
+        described_class.ensure_git_installed!
+      end
+    end
+  end
+
+  describe "::git_remote_exists" do
+    it "returns true when git is not available" do
+      stub_const("HOMEBREW_SHIMS_PATH", mktmpdir/"shim")
```
MikeMcQuaid: Instead of `mktmpdir` write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      expect(described_class.git_remote_exists("blah")).to be_truthy
+    end
+
+    context "when git is available" do
+      it "returns true when git remote exists", :needs_network do
+        git = HOMEBREW_SHIMS_PATH/"scm/git"
+        repo = Pathname.new("hey")
```
MikeMcQuaid: Write to an already existing location that's deleted after tests: for example: https://github.com/Homebrew/brew/blob/2c43e069c25e3c7cc255f5cd105df3597246b952/Library/Homebrew/test/spec_helper.rb#L110-L129

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+        repo.mkpath
+
+        system "cd", repo
```
MikeMcQuaid: Use `repo.cd do`

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+        expect(described_class.git_remote_exists("blah")).to be_falsey
+      end
+    end
+  end
+
+  describe "::clear_git_available_cache" do
+    it "removes @git_path and @git_version if defined" do
+      described_class.clear_git_available_cache
+
+      expect(described_class.instance_variable_get(:@git_path)).to be_nil
```
MikeMcQuaid: Need to try and figure out a way to not set/get instance variables here. If need be: delete the relevant tests that need them set given we're almost at the end of GSoC. Ideally: just call other methods that end up calling these ones.

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+      expect(described_class.instance_variable_get(:@git_version)).to be_nil
+    end
+
+    it "removes @git if defined" do
+      described_class.git_available?
+      described_class.clear_git_available_cache
+
+      expect(described_class.instance_variable_get(:@git)).to be_nil
```
MikeMcQuaid: Need to try and figure out a way to not set/get instance variables here. If need be: delete the relevant tests that need them set given we're almost at the end of GSoC. Ideally: just call other methods that end up calling these ones.

Library/Homebrew/utils/git.rb
```diff
@@ -16,11 +16,10 @@ def last_revision_commit_of_file(repo, file, before_commit: nil)
   def last_revision_of_file(repo, file, before_commit: nil)
     relative_file = Pathname(file).relative_path_from(repo)
 
-    commit_hash = last_revision_commit_of_file(repo, file, before_commit: before_commit)
-
+    commit_hash = last_revision_commit_of_file(repo, relative_file, before_commit: before_commit)
     out, = Open3.capture3(
       HOMEBREW_SHIMS_PATH/"scm/git", "-C", repo,
-      "show", "#{commit_hash}:#{relative_file}"
+      "show", "#{commit_hash}:#{file}"
```
MikeMcQuaid: @reitermarkus This look correct to you?

reitermarkus: No, it should still be the relative path. Might be that `git show` also supports absolute paths, but relative is definitely cleaner.

mansimarkaur:
Please review, @MikeMcQuaid . Thanks!

mistydemeo:
Looks good to me! Mind updating the commit history for merging?

Library/Homebrew/test/utils/git_spec.rb
```diff
@@ -0,0 +1,151 @@
+require "utils/git"
+
+describe Git do
+  before(:each) do
+    git = HOMEBREW_SHIMS_PATH/"scm/git"
+
+    @file = "blah.rb"
```
reitermarkus: Why are you using an instance variable here? Wasn't this only a workaround for `before(:all)`? `let` should work in `before(:each)`.

MikeMcQuaid:
Thanks again @mansimarkaur!

PR #2967, opened 2017-07-29
Silence all specs by default.

reitermarkus:

---

Remove `shutup`, and silence all specs by default.


Library/Homebrew/test/spec_helper.rb
```diff
@@ -88,11 +86,25 @@ def find_files
       @__argv = ARGV.dup
       @__env = ENV.to_hash # dup doesn't work on ENV
 
+      unless example.metadata.key?(:focus)
```
MikeMcQuaid: Let's also handle `VERBOSE_TESTS`

reitermarkus: Should we set this when `--verbose` is passed? It wasn't before.

PR #2976, opened 2017-07-30
audit: Port patches audit code to a rubocop and add tests

GauthamGoli:

-----
As of now `patch_problems` method is replicated at two different places(`audit.rb` and `legacy_patches_cop.rb`).  

In #2790 I told that `patch_problems` method would be moved to `formula_cop.rb` in future, since `audit.rb` code which calls `patch_problems` had not been completely ported yet then. 
 
In this PR, I ported rest of the code, hence we can remove `patch_problems` from `audit.rb`, and move its counterpart defined in `legacy_patches_cop.rb` to `formula_cop.rb` but I'm skeptical about it as I noticed that all the methods defined in `FormulaCop` class are very generalized methods.  
Is `FormulaCop` the right place for `patch_problems`? If not, I would refactor it accordingly.  Thanks!


MikeMcQuaid:
> Is FormulaCop the right place for patch_problems?

Yeh, I agree that it's probably not. Either you could have a shared class for the two classes using this or just put the external patch audits in the same class as the legacy ones.

GauthamGoli:
I will keep external patch audits in the same class as legacy patch audits.

JCount:
@GauthamGoli That approach seems to make the most sense. Grouping the logic of the patch audits together should hopefully help with maintenance, and perhaps coverage as well.

Library/Homebrew/rubocops/patches_cop.rb
```diff
@@ -4,16 +4,27 @@
 module RuboCop
   module Cop
     module FormulaAudit
-      # This cop checks for and audits legacy patches in Formulae
-      class LegacyPatches < FormulaCop
+      # This cop audits patches in Formulae
+      class Patches < FormulaCop
         def audit_formula(_node, _class_node, _parent_class_node, body)
+          # Audit External Patches
+          patches = find_all_blocks(body, :patch)
```
MikeMcQuaid: Could just name this `external_patches` and avoid the comments.

Library/Homebrew/test/rubocops/patches_cop_spec.rb
```diff
@@ -84,6 +85,15 @@ def patches
                                   line: 5,
                                   column: 5,
                                   source: source }]
+        elsif patch_url =~ %r{https?://patch-diff\.githubusercontent\.com/raw/(.+)/(.+)/pull/(.+)\.(?:diff|patch)}
+          expected_offenses = [{  message:  "use GitHub pull request URLs:\n"\
```
MikeMcQuaid: Favour using `EOS.indent` over embedding `\n` in multiline strings. If you've done this in other tests: make a new PR (or add to this one) that global style change. Alternatively, these could all be addressed in another PR after this is merged?

GauthamGoli: Yeah will change these in another PR.

MikeMcQuaid:
Thanks again @GauthamGoli!

JCount:
Good work, once again @GauthamGoli!

PR #2980, opened 2017-07-30
audit: fix bug where `brew audit foo` runs every style check.

GauthamGoli:

-----
This PR fixes bug, where every style check is ran when `brew audit <some_formula>` is executed which is not the expected behaviour.

Style checks are done only when `--strict` or `--new-formula` is passed to `brew audit`.
Without `--strict` or `--new-formula` , only RuboCop Custom Cops' style checks belonging to `FormulaAudit` department have to be checked. 

cc and Thanks @ilovezfs


MikeMcQuaid:
Thanks for the fix, @GauthamGoli!

PR #2982, opened 2017-07-30
audit: Port audit_class to rubocop, add tests and autocorrect

GauthamGoli:

-----
#569


Library/Homebrew/rubocops/class_cop.rb
```diff
@@ -0,0 +1,41 @@
...
+      class ClassName < FormulaCop
+        DEPRECATED_CLASSES = %w[
+          GithubGistFormula
+          ScriptFileFormula
+          AmazonWebServicesFormula
+        ].freeze
+
+        def audit_formula(_node, _class_node, parent_class_node, _body_node)
+          parent_class = class_name(parent_class_node)
+          return unless DEPRECATED_CLASSES.index(parent_class)
```
MikeMcQuaid: why `.index` and not `.include?` here?

Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -369,7 +369,14 @@ def problem(msg)
...
         _, class_node, = *node
-        class_node && string_content(class_node) == "Formula"
+        class_names = %w[
+          Formula
+          GithubGistFormula
+          ScriptFileFormula
+          AmazonWebServicesFormula
+        ]
+
+        class_node && class_names.index(string_content(class_node))
```
MikeMcQuaid: Not sure I understand this change, can you elaborate?

GauthamGoli: To Identify a formula, only classes which inherit from `Formula` were considered. This change adds the remaining 3 also, so that `class_cop.rb`'s checks can run.

Library/Homebrew/rubocops/class_cop.rb
GauthamGoli: Right! Sorry.

Library/Homebrew/rubocops/extend/formula_cop.rb
GauthamGoli: It should have been `class_node && class_names.include?(class_name(class_node))`

MikeMcQuaid: Gotcha. Those classes all inherit from Formula themselves, though, so perhaps it's worth looking through the entire inheritance hierarchy?

Library/Homebrew/rubocops/class_cop.rb
MikeMcQuaid: No worries, just checking!

Library/Homebrew/rubocops/extend/formula_cop.rb
GauthamGoli: I can't think of a way to look through inheritance hierarchy just using RuboCop without `require`-ing the Classes themselves. Do you think its possible?

MikeMcQuaid: If you can get the name or `Class` object of a class you could recurse through the inheritance hierarchy. That's maybe unnecessary, though but just an idea. I'm fine with this approach.

Library/Homebrew/rubocops/class_cop.rb
```diff
@@ -0,0 +1,41 @@
...
+        def autocorrect(node)
+          lambda do |corrector|
+            corrector.replace(node.source_range, "Formula")
+          end
+        end
+      end
+    end
+
+    module FormulaAuditStrict
+      # - `test do ..end` should be defined in a formula
```
JCount: `in a formula` => `in the formula` ?

JCount:
@GauthamGoli I apologize, I completely lost track of this. What is the current status of this PR, do you still want this merged as-is or do you want to make modifications?

GauthamGoli:
```
Gauthams-MacBook-Pro:Homebrew gautham$ brew audit --only-cops=FormulaAudit/ClassName
Warning: Calling GithubGistFormula.url is deprecated!
Use Formula.url instead.
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-science/nest.rb:4:in `<class:Nest>'
Please report this to the homebrew/science tap!

homebrew/science/nest:
  * C: 1: col 14: GithubGistFormula is deprecated, use Formula instead
Error: 1 problem in 1 formula
```
`odeprecated` is being called in `formula_specialities.rb` in the respective methods of deprecated classes which is causing the above duplication of warning/audit errors. I kind of don't know what to do here yet.

MikeMcQuaid:
> odeprecated is being called in formula_specialities.rb in the respective methods of deprecated classes which is causing the above duplication of warning/audit errors. I kind of don't know what to do here yet.

Could you `rescue MethodDeprecatedError` to handle this issue? I've not tried it so it may not work. If not we could consider removing this audit in favour of the message.

GauthamGoli:
Why are there warnings and audit rules at the same time?

MikeMcQuaid:
> Why are warnings and audit rules present simultaneously for audit_class?

@GauthamGoli The warnings are a runtime check to ensure that even people who don't use `brew audit` receive notification about their deprecation.

GauthamGoli:
Okay. These warnings are piped to $stderr, and can be temporarily redirected to elsewhere when the formulae are loaded in audit to prevent the warning from appearing. But doing so would redirect whole of $stderr and other warnings would get redirected too. I'm currently trying to find some workaround.

MikeMcQuaid:
@GauthamGoli In this case it may be reasonable to modify the deprecation code to handle this case.

GauthamGoli:
Modifying it to raise `MethodDeprecatedError` instead of piping to $stderr, and then `rescue`ing it is tricky too, as `rescue` would prevent instances of formulae from getting created, in `audit`

MikeMcQuaid:
Was `Library/Homebrew/plist` included by mistake?

MikeMcQuaid:
@GauthamGoli I reckon change https://github.com/Homebrew/brew/blob/2a75c6c591c6bb0562ed25c84c951f70c9d8d90d/Library/Homebrew/utils.rb#L101-L107 to have different behaviour when running `audit`. For example, you could have a `Homebrew.auditing?` or similar which instead of raising excepts adds things to a `Homebrew.audit_problems` global array and then `brew audit` outputs them at the end. This would allow you to remove the audits that duplicate the deprecation code. Thoughts?

GauthamGoli:
Yeah. Makes sense to have `audit` specific flag. 
Instead of adding to `Homebrew.audit_problems`, I think its better to suppress it, and let the Cop report it because it would have access to the node's location in the source. 

Yes. `Library/Hombrew/plist` has been added by mistake will remove it.

JCount:
That also additionally solves the problem of having to update, (or add to), the audit code when deprecations are modified within the reset of the codebase. A big 👍 to avoiding unnecessary duplication in the codebase.

MikeMcQuaid:
> Instead of adding to Homebrew.audit_problems, I think its better to suppress it, and let the Cop report it because it would have access to offending node's location in the source.

Sounds good to me 👍

GauthamGoli:
Made changes, can you please review?

Library/Homebrew/utils.rb
```diff
@@ -103,7 +103,7 @@ def odeprecated(method, replacement = nil, disable: false, disable_on: nil, call
      Homebrew.raise_deprecation_exceptions?
     raise MethodDeprecatedError, message
   else
```
MikeMcQuaid: `elsif !Homebrew.auditing?`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -121,6 +122,7 @@ def audit
     return if problem_count.zero?
 
     ofail "#{Formatter.pluralize(problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
+    Homebrew.auditing = false
```
MikeMcQuaid: Can probably skip setting this here as the end of a command is the end of termination, too.

Library/Homebrew/utils.rb
JCount: Doesn't this get rid of the default case? Currently, that should not be an issue, but it does add a single point of failure for the `odeprecated` code message printout.

MikeMcQuaid:
Thanks again @GauthamGoli!

JCount:
👍

PR #2990, opened 2017-08-01
formula_desc_cop: add 'x86' to VALID_LOWERCASE_WORDS

JCount:

-----
add `x86` to `VALID_LOWERCASE_WORDS`

-----
This is a perfectly valid "word" to use at the beginning of a formula's description that should not be capitalized. A current example of a formula that fails a strict audit because of this is [qemu](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qemu.rb).

(Should `x86-64` be added as well? My personal opinion is that it is probably unnecessary.)


MikeMcQuaid:
Nice spot!

PR #2991, opened 2017-08-01
audit: add additional php header ignore patterns

javian:

-----
Since https://github.com/Homebrew/homebrew-core/pull/16067 introduces a new php formula with the standard naming convention the header ignore rule needs to be updated for the formula to pass green in the CI. This new regexp should support the following formula names which ought to cover any future additions to the php formulas in core. The regexp should match
```
php (new Formula)
php@7.1 (new versioned formula, not yet exisiting)
php71 (convention used in the php tap)
```

### DIFF
--- a/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
@@ -6,7 +6,7 @@ module FormulaCellarChecks
       formula.name.start_with?(formula_name)
     end
 
-    return if formula.name =~ /^php\d+$/
+    return if formula.name =~ /^php(@?\d+\.?\d*?)?$/
 
     return if MacOS.version < :mavericks && formula.name.start_with?("postgresql")
     return if MacOS.version < :yosemite  && formula.name.start_with?("memcached")


Library/Homebrew/extend/os/mac/formula_cellar_checks.rb
```diff
@@ -6,7 +6,7 @@ def check_shadowed_headers
       formula.name.start_with?(formula_name)
     end
 
-    return if formula.name =~ /^php\d+$/
+    return if formula.name =~ /^php(?:\@\d\.\d|\d+)?$/
```
MikeMcQuaid: `/^php(@\d+(\.\d)?)?$/` is a little simpler and should work, I think?

javian: So the at sign doesn't need to be escaped the ?
That regexp does not seem to match php71 which is the current filenames used in the php tap. We could do /^php(\d+|@\d+\.\d+)?$/ which would also match 7.10 if the version would be written with double digits.  I think your regexp was also designed to match  php@7 as well as php@7.0 but other Formulas like ruby write the .0 so I'm not sure if we need to cover that case ?

MikeMcQuaid: > So the at sign doesn't need to be escaped ?

MikeMcQuaid: > So the at sign doesn't need to be escaped ?

Yup.

> We could do /^php(\d+|@\d+.\d+)?$/ which would also match 7.10 if the version would be written with double digits.

That works. Could also do `/^php@?\d+(.\d+)?$/` if you think that's more readable (I do but you may disagree. Either is fine, though 👍

javian: Regexp now update, looks similar but should cover everything now =)

MikeMcQuaid:
Thanks again @javian!

PR #2996, opened 2017-08-03
brew deps: add --include-requirements, plus some fixes

apjanke:

-----

Adds a `--include-requirements` option to `brew deps`, to display all requirements, including ones without a default formula. Ones with a default formula are displayed as a requirement, with the default formula noted in parentheses. Per a suggestion by @ilovezfs.

Mostly useful in conjunction with `--tree`. This helps hunt down where a deeply-nested requirement is coming from. For example, `octave` has an indirect dependency on `:x11`, but it has so many dependencies it's hard to tell where they're coming from. Or it lets you determine where the MinimumMacOS requirement for `mytop` is coming from.

Also fixes a bug in `--tree` where the dependencies of a Requirement's default formula were not recursed into.

Also fixes a bug where the `--tree` mode ignored the `--include-optional`, `--include-build`, and `--skip-recommended` filters, instead always displaying build and recommended dependencies, and never optionals.

Also adds an `--annotate` option to the `--tree` mode, marking up optional, recommended, and build dependencies as such in the display.

Also adds an undocumented `--for-each` switch to help debug the `--all`/`--installed` display mode, which couldn't previously be gotten to for a list of specified formulae.

Before:
```
$ brew deps --tree xpdf
xpdf (required dependencies)
├── openmotif
│   ├── automake
│   │   └── autoconf
│   ├── autoconf
...

$ brew deps --tree mytop
mytop (required dependencies)
├── :mysql
└── openssl
    └── makedepend
        └── pkg-config
```

After:

```
$ brew deps --tree --include-requirements xpdf
xpdf
├── :x11
├── openmotif
│   ├── :x11
│   ├── fontconfig
│   │   └── freetype
...

$ brew deps --tree mytop
mytop
├── mysql
│   └── openssl
└── openssl

$ brew deps --tree mytop --include-requirements
mytop
├── :mysql (mysql)
│   ├── :macOS >= 10.7
│   └── openssl
└── openssl

$ brew deps --tree mytop --include-requirements --include-build
mytop
├── :mysql (mysql)
│   ├── :macOS >= 10.7
│   ├── cmake
│   │   └── sphinx-doc
│   └── openssl
│       └── makedepend
│           └── pkg-config
└── openssl
    └── makedepend
        └── pkg-config
```

Removing the colon before the default formula when not doing `--include-requirements` makes the `--tree` output more consistent with the non-tree output.

Note that if you include all of `--tree --include-optional --include-build --include-requirements` on large formulae (like `brew deps octave --tree --include-optional --include-build --include-requirements`), you can get a stack overflow error. I suspect this is because there's a circular dependency in the formula definitions somewhere. This could be worked around by using a stack to detect circular dependencies inside the code here, but I haven't put it in yet.

###  Test failure

I'm getting this failure when I run `brew tests` locally, but it's also happening for me on `master`. Don't think it's introduced by this PR.

```
Failures:

  1) OS::Mac::language can be parsed by Locale::parse
     Failure/Error: expect { Locale.parse(subject.language) }.not_to raise_error

       expected no Exception, got #<Locale::ParserError: '' cannot be parsed to a Locale> with backtrace:
         # ./locale.rb:15:in `parse'
         # ./test/os/mac_spec.rb:19:in `block (4 levels) in <top (required)>'
         # ./test/os/mac_spec.rb:19:in `block (3 levels) in <top (required)>'
     # ./test/os/mac_spec.rb:19:in `block (3 levels) in <top (required)>'

Finished in 56.57 seconds (files took 1.49 seconds to load)
446 examples, 1 failure

Failed examples:

rspec ./test/os/mac_spec.rb:18 # OS::Mac::language can be parsed by Locale::parse
```

### DIFF
--- a/Library/Homebrew/cmd/deps.rb
+++ b/Library/Homebrew/cmd/deps.rb
@@ -1,4 +1,4 @@
-#:  * `deps` [`--1`] [`-n`] [`--union`] [`--full-name`] [`--installed`] [`--include-build`] [`--include-optional`] [`--skip-recommended`] <formulae>:
+#:  * `deps` [`--1`] [`-n`] [`--union`] [`--full-name`] [`--installed`] [`--include-build`] [`--include-optional`] [`--skip-recommended`] [`--include-requirements`] <formulae>:
 #:    Show dependencies for <formulae>. When given multiple formula arguments,
 #:    show the intersection of dependencies for <formulae>.
 #:
@@ -19,15 +19,22 @@
 #:    <formulae>. To include the `:build` type dependencies, pass `--include-build`.
 #:    Similarly, pass `--include-optional` to include `:optional` dependencies.
 #:    To skip `:recommended` type dependencies, pass `--skip-recommended`.
+#:    To include requirements in addition to dependencies, pass `--include-requirements`.
 #:
-#:  * `deps` `--tree` [<filters>] (<formulae>|`--installed`):
+#:  * `deps` `--tree` [`--1`] [<filters>] [`--annotate`] (<formulae>|`--installed`):
 #:    Show dependencies as a tree. When given multiple formula arguments, output
 #:    individual trees for every formula.
+#:
+#:    If `--1` is passed, only one level of children is displayed.
 #:
 #:    If `--installed` is passed, output a tree for every installed formula.
 #:
 #:    The <filters> placeholder is any combination of options `--include-build`,
-#:    `--include-optional`, and `--skip-recommended` as documented above.
+#:    `--include-optional`, `--skip-recommended`, and `--include-requirements` as
+#:    documented above.
+#:
+#:    If `--annotate` is passed, the build, optional, and recommended dependencies
+#:    are marked as such in the output.
 #:
 #:  * `deps` [<filters>] (`--installed`|`--all`):
 #:    Show dependencies for installed or all available formulae. Every line of
@@ -36,6 +43,10 @@
 #:
 #:    The <filters> placeholder is any combination of options `--include-build`,
 #:    `--include-optional`, and `--skip-recommended` as documented above.
+
+# The undocumented `--for-each` option will switch into the mode used by `deps --all`,
+# but only list dependencies for specified formula, one specified formula per line.
+# This is used for debugging the `--installed`/`--all` display mode.
 
 # encoding: UTF-8
 
@@ -52,20 +63,26 @@ module Homebrew
       all?: ARGV.include?("--all"),
       topo_order?: ARGV.include?("-n"),
       union?: ARGV.include?("--union"),
+      for_each?: ARGV.include?("--for-each"),
     )
 
-    if mode.installed? && mode.tree?
-      puts_deps_tree Formula.installed
+    if mode.tree?
+      if mode.installed?
+        puts_deps_tree Formula.installed, !ARGV.one?
+      else
+        raise FormulaUnspecifiedError if ARGV.named.empty?
+        puts_deps_tree ARGV.formulae, !ARGV.one?
+      end
     elsif mode.all?
       puts_deps Formula
-    elsif mode.tree?
-      raise FormulaUnspecifiedError if ARGV.named.empty?
-      puts_deps_tree ARGV.formulae
     elsif ARGV.named.empty?
       raise FormulaUnspecifiedError unless mode.installed?
       puts_deps Formula.installed
+    elsif mode.for_each?
+      puts_deps ARGV.formulae
     else
       all_deps = deps_for_formulae(ARGV.formulae, !ARGV.one?, &(mode.union? ? :| : :&))
+      all_deps = condense_requirements(all_deps)
       all_deps = all_deps.select(&:installed?) if mode.installed?
       all_deps = all_deps.map(&method(:dep_display_name)).uniq
       all_deps.sort! unless mode.topo_order?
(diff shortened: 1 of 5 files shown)


ilovezfs:
Thanks for this PR @apjanke! It seems to work as expected in my testing.

>you can get a stack overflow error. 

It would be good if this were fixed before this goes in. `brew deps --tree -include-optional ffmpeg` caused my terminal to lose its mind. :)

(Also, it would be neat if --tree respected --1.)

apjanke:
Added `--1` support to `--tree`. I don't see how that would be useful in practice, but it's nice in that it unifies the behavior of the different display modes; I like it.

I see how to implement cycle detection to avoid the stack overflow. I'll add it. The `brew deps --tree -include-optional ffmpeg` blow-up reproduces for me too, and that's not an obscure case.

apjanke:
Added circular-dependency detection. Now `brew deps --tree --include-optional ffmpeg` works:

```
$ brew deps --tree --include-optional ffmpeg
ffmpeg
├── lame
├── x264
│   └── l-smash
├── xvid
...
│       │   ├── jpeg
│       │   ├── libpng
│       │   └── libtiff
│       │       ├── jpeg
│       │       └── xz
│       └── ffmpeg (CIRCULAR DEPENDENCY)
```

and `brew deps octave --tree ...` is merely insane, not broken:

```
$ brew deps octave --tree --include-optional --include-build --include-requirements | wc -l
   25843
```

ilovezfs:
>and brew deps octave --tree ... is merely insane, not broken:
>```
>$ brew deps octave --tree --include-optional --include-build --include-requirements | wc -l
>   25843
>```

A generic `--depth` option would probably help to tame that particular crazy.

Library/Homebrew/cmd/deps.rb
```diff
@@ -52,45 +63,91 @@ def deps
       all?: ARGV.include?("--all"),
       topo_order?: ARGV.include?("-n"),
       union?: ARGV.include?("--union"),
+      for_each?: ARGV.include?("--for-each"),
     )
+    @with_reqs = ARGV.include?("--include-requirements")
+    @with_optional = ARGV.include?("--include-optional")
+    @with_build = ARGV.include?("--include-build")
+    @skip_recommended = ARGV.include?("--skip-recommended")
+    @annotate = ARGV.include?("--annotate")
```
MikeMcQuaid: Feels weird to use class variables for some of `ARGV` and not others. I'd lean on the side of never using them.

Library/Homebrew/cmd/deps.rb
```diff
@@ -52,45 +63,91 @@ def deps
...
       all_deps.sort! unless mode.topo_order?
       puts all_deps
     end
   end
 
+  def condense_requirements(deps)
+    if @with_reqs
+      deps
+    else
+      deps.map do |d|
```
MikeMcQuaid: `deps.map do |dep|`

Library/Homebrew/cmd/deps.rb
```diff
@@ -52,45 +63,91 @@ def deps
...
+        if d.is_a? Dependency
+          d
+        elsif d.default_formula?
+          d.to_dependency
+        end
+      end.compact
+    end
+  end
+
   def dep_display_name(d)
```
MikeMcQuaid: `def dep_display_name(dep)`

Library/Homebrew/cmd/deps.rb
```diff
@@ -129,41 +186,52 @@ def deps_for_formulae(formulae, recursive = false, &block)
...
+      next if !@with_reqs && dep.is_a?(Requirement) && !dep.default_formula?
       chr = if i == max
         "└──"
       else
         "├──"
       end
-      prefix_ext = (i == max) ? "    " : "│   "
-      puts prefix + "#{chr} #{dep_display_name(dep)}"
-      recursive_deps_tree(Formulary.factory(dep.name), prefix + prefix_ext)
+      display_s = "#{chr} #{dep_display_name(dep)}"
```
MikeMcQuaid: Can we figure out a better name than `chr` here? `line`, maybe?

Library/Homebrew/cmd/deps.rb
```diff
@@ -129,41 +186,52 @@ def deps_for_formulae(formulae, recursive = false, &block)
...
+      is_circular = @dep_stack.include?(dep.name)
+      display_s = "#{display_s} (CIRCULAR DEPENDENCY)" if is_circular
+      puts prefix + display_s
```
MikeMcQuaid: `puts "#{prefix}#{display_s}`

Library/Homebrew/cmd/deps.rb
```diff
@@ -129,41 +186,52 @@ def deps_for_formulae(formulae, recursive = false, &block)
...
+      if recursive && !is_circular
+        prefix_ext = (i == max) ? "    " : "│   "
```
MikeMcQuaid: `prefix_extension = if i == max`

Library/Homebrew/cmd/deps.rb
apjanke: So, I tried writing this to use function parameters for everything, and it got pretty hard to ready. Maybe it should fall back to using `ARGV.include?` everywhere? Or a single class instance variable for all the options? My impression is we're trying to move away from scattering global `ARGV` references everywhere.

apjanke:
Addressed most of the style pedantry, except for the use of class instance variables. Not sure what the right fix to do there is; using function parameters makes the code look pretty gross IMHO. I think class instance variables should be okay in the case of commands, since the modules are only loaded one per process, used once, and then the process ends. Deferring to the judgment of other maintainers here though.

apjanke:
And, thanks!

apjanke:
Updated the `zsh` completions to reflect the new `brew deps --tree` behavior in this PR.

Library/Homebrew/cmd/deps.rb
apjanke: Done.

Library/Homebrew/cmd/deps.rb
apjanke: Done.

Library/Homebrew/cmd/deps.rb
apjanke: how about `tree_lines`, to distinguish the graphical lines of the tree structure from a "line" of text?

Library/Homebrew/cmd/deps.rb
apjanke: Done.

Library/Homebrew/cmd/deps.rb
apjanke: Done.

Library/Homebrew/cmd/deps.rb
MikeMcQuaid: >  Maybe it should fall back to using ARGV.include? everywhere?

Yeh, I agree. 

> My impression is we're trying to move away from scattering global ARGV references everywhere.

I think broadly but at this point it's probably reliant on having a decent Command DSL first.

apjanke: I've updated it to use direct `ARGV` references everywhere, and removed the class instance variables.

apjanke:
Arg! `brew style`-ing now...

apjanke:
Merged. Thanks for the feedback!

chdiza:
> For example, octave has an indirect dependency on :x11, but it has so many dependencies it's hard to tell where they're coming from.

I just tried `brew deps octave --tree --include-requirements`, and no such `:x11` appears in the output.  Am I doing it wrong?

ilovezfs:
@chdiza no https://github.com/Homebrew/homebrew-core/pull/16193

PR #2662, opened 2017-05-21
audit: Port audit_text method to rubocop and add tests

GauthamGoli:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----
#569 
I ported most of the code in `audit_text` method except [few lines](https://github.com/GauthamGoli/brew/blob/ac5b43921fe84ed16f2b0540b28a5f0f2d5b7849/Library/Homebrew/dev-cmd/audit.rb#L855-L868) which use Homebrew's internal code.

PR #3012, opened 2017-08-06
Add node pattern methods to handle dependency audits in a better way

GauthamGoli:

-----
#569 and #2662 
RuboCop has pattern searching macros which simplify searching/matching nodes in AST. 
Found documentation for same in the source of [node_pattern.rb](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/node_pattern.rb)

This PR updates `depends_on` audits to use such pattern matching methods, so code becomes more readable and also fixes bug in previous code which wasn't detecting dependencies. 

(there is however one edge case where the value of a `depends_on` hash is an array, I know how exactly to implement that and will deal with that later)

**Edit:** How did I derive those patterns?  By running the below commands and looking at the AST.
`$ ruby-parse -e 'depends_on :x11 => optional'`
`$ ruby-parse -e 'depends_on "readline" => :recommended'`

cc @ilovezfs

### DIFF
--- a/Library/Homebrew/rubocops/extend/formula_cop.rb
+++ b/Library/Homebrew/rubocops/extend/formula_cop.rb
@@ -1,4 +1,5 @@
 require "parser/current"
+require_relative "../../extend/string"
 
 module RuboCop
   module Cop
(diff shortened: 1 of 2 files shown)


Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -161,6 +158,22 @@ def depends_on_name_type?(node, name = nil, type = :required)
         type_match && name_match
       end
 
+      def_node_search :required_dependency?, <<-PATTERN
```
MikeMcQuaid: Worth using `EOS` or `EOS.undent` here?

GauthamGoli: Yeah.

MikeMcQuaid:
Thanks again @GauthamGoli!

PR #3034, opened 2017-08-09
shims/super/cc: Fix c++ -xc++-header

sjackman:
Compiling a precompiled header should be mode `:cxx` rather than `:cxxld`.

When compiling a precompiled header with `c++ -xc++-header`, `shims/super/cc` believes the mode is `:cxxld` and adds `cxxflags + args + cppflags + ldflags`, when it should be `cxxflags + args + cppflags`. `ldflags` should not be included when compiling a precompiled header.

This issue caused `brew install homebrew/science/opencv` to fail on Linux. See https://github.com/Homebrew/homebrew-science/pull/6092

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -43,6 +43,8 @@ class Cmd
       else
         :cc
       end
+    elsif @args.include?("-xc++-header") || @args.each_cons(2).include?(["-x", "c++-header"])
+      :cxx
     elsif @args.include? "-E"
       :ccE
     else


MikeMcQuaid:
Any ideas why this works fine on macOS?

sjackman:
Yes. The `LDFLAGS` that are addd on macOS, namely `-L` options, don't interfere with the proper operation of `-xc++-header`. Linuxbrew adds to `LDFLAGS` `-Wl,-rpath=…` and `-Wl,--dynamic-linker=`, and the latter breaks `-xc++-header`. That's arguably a bug in GCC, since adding linker options to the command line of the GCC driver shouldn't break it when it's operating in a non-linking mode.

MikeMcQuaid:
@sjackman In that case: can this be guarded to be GCC only or will these flags literally never be used on Clang?

sjackman:
`clang` on Linux appears to support all three options mentioned above, `-xc++-header`, `-Wl,-rpath=…`, and `-Wl,-dynamic-linker=…`. The first two options work on macOS, and `-Wl,-dynamic-linker=…` is specific to Linux.

MikeMcQuaid:
Sounds like this fix should be scoped to Linux and/or GCC, then, no?

sjackman:
When compiling a pre-compiled header with `-xc++header`, which is also used on macOS with clang, there's no need to supply the linker flags `ldflags`, as they have no effect. Really your call though. I don't feel strongly. I haven't yet submitted my `--env=super` support for Linux to Homebrew. When I do it will include this fix for Linux.

MikeMcQuaid:
@ilovezfs any thoughts?

sjackman:
`brew install homebrew/science/opencv` uses `-x c++-header`, as one test.

MikeMcQuaid:
@sjackman Thanks. Have you been able to test this patch with it on macOS? If so: I trust you enough to not do it myself. No worries if not, though.

sjackman:
I've tested `brew install -s hello homebrew/science/opencv && brew test opencv` on macOS. That much works.

MikeMcQuaid:
Thanks @sjackman!

sjackman:
Thanks for merging, Mike!

Issue #3051, opened 2017-08-13
Bash completion incomplete for "uninstall"


rhwood:
# Please always follow these steps:
- [x] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed all issues and retried your prior step?
- [x] Ran `brew config` and `brew doctor` and included their output with your issue?

**Please note we will close your issue without comment if you delete or do not fill out the issue checklist and provide ALL the requested information.**

To help us debug your issue please explain:
- What you were trying to do (and why)
- What happened (include command output)
- What you expected to happen
- Step-by-step reproduction instructions (by running `brew` commands)

I was attempting to uninstall a formula, and discovered that bash completion for brew does not automatically complete "uninstall " because it wants me to decide if I want to run "uninstal" (one letter L) or "uninstall" (two letter Ls). I see that no such distinction is made in bash completion for "install" (in fact, the bash completion deliberately removes the alias "instal" to prevent that).

Steps to reproduce in bash shell (```<tab>``` represents pressing the tab key):
```
brew uni<tab>
```
expected results are:
```
brew uninstall <-- with space following word "uninstall"
```

actual results are:
```
brew uninstal<-- no space
<tab><tab>
uninstal   uninstall
```

rhwood:
Output of ```brew config``` and ```brew doctor```:

```
Calculus:JMRI rhwood$ brew config
HOMEBREW_VERSION: 1.3.1-40-ge1bab7e-dirty
ORIGIN: https://github.com/Homebrew/brew
HEAD: e1bab7ede9461a6d5c93fb7240ba5a1f3bf7a7fb
Last commit: 24 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 3a0dd5196ac84427ad394be4c3c81c5f49c569bc
Core tap last commit: 25 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: quad-core 64-bit broadwell
Homebrew Ruby: 2.0.0-p648
[...]

PR #3052, opened 2017-08-13
Remove "*instal" aliases from bash completion

rhwood:

-----
Remove all command aliases ending in "instal" (single letter L) from the bash completion. This has the practical effect of removing the aliases "instal" and "uninstal" from bash completion, allowing "install" and "uninstall" to be auto completed once the first three characters of each "ins" or "uni" are typed in.

This makes bash completion for "uninstall" complete with a following space on the first tab instead of offering two different versions of the same command.

Addresses #3051

### DIFF
--- a/completions/bash/brew
+++ b/completions/bash/brew
@@ -563,7 +563,7 @@ _brew() {
 
   if [[ "$i" -eq "$COMP_CWORD" ]]
   then
-    # Do not auto-complete "*instal" abbreviations for "*install" commands.
+    # Do not auto-complete "*instal" or "*uninstal" aliases for "*install" commands.
     # Prefix newline to prevent not checking the first command.
     local cmds=$'\n'"$(brew commands --quiet --include-aliases | grep -v instal$)"
     __brewcomp "${cmds}"
--- a/completions/bash/brew
+++ b/completions/bash/brew
@@ -563,10 +563,10 @@ _brew() {
 
   if [[ "$i" -eq "$COMP_CWORD" ]]
   then
-    # Do not auto-complete "instal" abbreviation for "install" command.
+    # Do not auto-complete "*instal" abbreviations for "*install" commands.
     # Prefix newline to prevent not checking the first command.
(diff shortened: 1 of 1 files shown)


completions/bash/brew
```diff
@@ -563,10 +563,10 @@ _brew() {
 
   if [[ "$i" -eq "$COMP_CWORD" ]]
   then
-    # Do not auto-complete "instal" abbreviation for "install" command.
+    # Do not auto-complete "*instal" abbreviations for "*install" commands.
```
MikeMcQuaid: `# Do not auto-complete "instal" or "uninstal" abbreviations for "*install" commands.` I think is a little easier to parse.

completions/bash/brew
```diff
@@ -563,10 +563,10 @@ _brew() {
...
     # Prefix newline to prevent not checking the first command.
-    local cmds=$'\n'"$(brew commands --quiet --include-aliases)"
-    __brewcomp "${cmds/$'\n'instal$'\n'/$'\n'}"
+    local cmds=$'\n'"$(brew commands --quiet --include-aliases | grep -v instal$)"
```
MikeMcQuaid: Any way we can do this without spawning a new `grep` process? No worries if not, just wanted to check.

MikeMcQuaid:
@rhwood Thanks for the PR! FYI for future if you're happy to open a PR for an issue you don't need to create an issue first.

completions/bash/brew
rhwood: I could get it to work without calling grep was:
```
local cmds=$'\n'"$(brew commands --quiet --include-aliases)"
cmds="${cmds/$'\n'uninstal$'\n'/$'\n'}"
__brewcomp "${cmds/$'\n'instal$'\n'/$'\n'}"
```
I could not figure out a regex pattern that would work to capture both instal and uninstal in a single line, and went for the one liner (which I also find more readable).

MikeMcQuaid:
Thanks for your contribution to Homebrew, @rhwood! Without people like you submitting PRs we couldn't run this project. You rock!

PR #3053, opened 2017-08-13
diagnostic: don't tap homebrew/core

alyssais:

-----

Attempting to look up a formula when the core tap has not been set up yet made Homebrew silently clone it, which meant that `brew doctor` took a very long time with no indication of what was happening.

Since both of the diagnostic checks that caused this behaviour will only fail if a homebrew/core formula is installed, it seems a little silly to clone it just for the purpose of checking that.

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -942,6 +942,8 @@ module Homebrew
       end
 
       def check_for_non_prefixed_coreutils
+        return unless CoreTap.instance.installed?
+
         coreutils = Formula["coreutils"]
         return unless coreutils.any_version_installed?
 
@@ -955,6 +957,8 @@ module Homebrew
       end
 
       def check_for_non_prefixed_findutils
+        return unless CoreTap.instance.installed?
+
         findutils = Formula["findutils"]
         return unless findutils.any_version_installed?
 


ilovezfs:
>when the core tap has not been set up yet

When would that ever be the case?

alyssais:
When Homebrew has been installed by cloning the Git repo

MikeMcQuaid:
Will `brew doctor` fail in this case after this change? A better fix may be to instead to verbosely tap Homebrew/core when running `brew doctor` (before calling these methods) if it is not already tapped as we don't support using Homebrew without Homebrew/core tapped.

alyssais:
`brew doctor` doesn't fail in this case currently, and it won't fail after.

I think making `brew doctor` automatically modify the installation would be pretty inconsistent with current behaviour and user expectations. I'd be open to adding a check that produced a warning saying that homebrew/core isn't tapped though? It just seems odd to me that it happens silently in these two checks that don't really seem related to whether homebrew/core is tapped.

MikeMcQuaid:
> I'd be open to adding a check that produced a warning saying that homebrew/core isn't tapped though?

Yeh, if that could be part of this PR that'd be good.

> It just seems odd to me that it happens silently in these two checks that don't really seem related to whether homebrew/core is tapped.

There's a lot of places in the codebase this will happen. In general any time a formula is referenced by name: if Homebrew/core isn't tapped it will be. This may feel weird but it's actually the nicest behaviour for users. `brew doctor` may be worth special-casing here though, I agree.

alyssais:
Further investigation reveals that the implicit tapping can be triggered by far more diagnostic checks than just these two, and it's extremely difficult to predict which ones. So, rather than trying to modify all of them, I wonder if it might just be better to make sure Homebrew *always* outputs "Tapping homebrew/core" or something while doing so, so it doesn't look like it's just hanging. This would be a much easier fix that addresses the same problem. Thoughts?

MikeMcQuaid:
> So, rather than trying to modify all of them, I wonder if it might just be better to make sure Homebrew always outputs "Tapping homebrew/core" or something while doing so, so it doesn't look like it's just hanging.

This seems fine, yeh.

MikeMcQuaid:
@alyssais Any updates here?

PR #3071, opened 2017-08-20
Update FAQ entry re. .apps and /usr/local/bin

### LINKED ISSUE #3064: FAQ entry for "My Mac .apps don’t find /usr/local/bin utilities!" is outdated
It says:

> If you’re on Mountain Lion, you can fix this by running launchctl setenv PATH "/usr/local/bin:$PATH".

however this changed again with 10.8.4: see https://apple.stackexchange.com/a/94972 .

Looks like, at least since 10.10, one should do `sudo launchctl config user path /usr/local/bin/:$PATH`, per https://apple.stackexchange.com/a/198282 .

akalin:
The existing instructions don't work for current versions of macOS.

This fixes #3064.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -71,9 +71,8 @@ Which is usually: `~/Library/Caches/Homebrew`
 GUI apps on macOS don’t have `/usr/local/bin` in their `PATH` by
 default.  If you’re on Mountain Lion or later, you can fix this by
 running `sudo launchctl config user path "/usr/local/bin:$PATH"` and
-then rebooting. Note that this sets the launchctl PATH for _all users_
-([more details here](https://apple.stackexchange.com/a/198282)). For
-earlier versions of macOS, see
+then rebooting, as documented in `man launchctl`. Note that this sets
+the launchctl PATH for _all users_. For earlier versions of macOS, see
 [this page](https://developer.apple.com/legacy/library/qa/qa1067/_index.html).
 
 ## How do I contribute to Homebrew?
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -69,7 +69,7 @@ Which is usually: `~/Library/Caches/Homebrew`
 
 ## My Mac `.app`s don’t find `/usr/local/bin` utilities!
 GUI apps on macOS don’t have `/usr/local/bin` in their `PATH` by
-default.  If you're on Mountain Lion or later, you can fix this by
+default.  If you’re on Mountain Lion or later, you can fix this by
 running `sudo launchctl config user path "/usr/local/bin:$PATH"` and
 then rebooting. Note that this sets the launchctl PATH for _all users_
 ([more details here](https://apple.stackexchange.com/a/198282)). For
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -73,9 +73,7 @@ default.  If you're on Mountain Lion or later, you can fix this by
 running `sudo launchctl config user path "/usr/local/bin:$PATH"` and
 then rebooting. Note that this sets the launchctl PATH for _all users_
 ([more details here](https://apple.stackexchange.com/a/198282)). For
-earlier versions,
-see [this page](https://stackoverflow.com/a/5444960/7869253),
-or
+earlier versions of macOS, see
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -68,13 +68,15 @@ Be careful as this is a destructive operation.
...
-including how to set this across reboots. If you’re pre-Mountain Lion,
-[here’s an official
-alternative](https://developer.apple.com/legacy/library/qa/qa1067/_index.html).
+GUI apps on macOS don’t have `/usr/local/bin` in their `PATH` by
+default.  If you're on Mountain Lion or later, you can fix this by
+running `sudo launchctl config user path "/usr/local/bin:$PATH"` and
+then rebooting. Note that this sets the launchctl PATH for _all users_
+([more details here](https://apple.stackexchange.com/a/198282)). For
+earlier versions,
+see [this page](https://stackoverflow.com/a/5444960/7869253),
```
alyssais: Do we need to add this link in addition to the official Apple document?

MikeMcQuaid: I don't think we do, let's just go for the official source, thanks!

akalin: Fixed!

akalin:
Please take another look.

docs/FAQ.md
```diff
@@ -68,13 +68,13 @@ Be careful as this is a destructive operation.
...
 ## My Mac `.app`s don’t find `/usr/local/bin` utilities!
-GUI apps on macOS don’t have `/usr/local/bin` in their `PATH` by default.
-If you’re on Mountain Lion, you can fix this by running
-`launchctl setenv PATH "/usr/local/bin:$PATH"`. [More details
-here](https://stackoverflow.com/questions/135688/setting-environment-variables-in-os-x/5444960#5444960),
-including how to set this across reboots. If you’re pre-Mountain Lion,
-[here’s an official
-alternative](https://developer.apple.com/legacy/library/qa/qa1067/_index.html).
+GUI apps on macOS don’t have `/usr/local/bin` in their `PATH` by
+default.  If you're on Mountain Lion or later, you can fix this by
```
apjanke: Pointing this out mostly for clarification from other maintainers: this PR changes a curly apostrophe to a straight one. Do we have a preference on which we use in docs? A naive `grep` through the source looks like a mix.

```
$ grep -r "'" * | wc -l
     235
[/usr/local/Homebrew/docs on ⇄ master]
$ grep -r "’" * | wc -l
     111
```

FWIW, I prefer the curlys because they look nicer, though they're certainly harder to type. (E.g. you may not I'm not using them in this comment.)

docs/FAQ.md
```diff
@@ -68,13 +68,13 @@ Be careful as this is a destructive operation.
...
+running `sudo launchctl config user path "/usr/local/bin:$PATH"` and
+then rebooting. Note that this sets the launchctl PATH for _all users_
+([more details here](https://apple.stackexchange.com/a/198282)). For
```
apjanke: Could this StackExchange link be replaced by a link to official Apple documentation, or something more authoritative? This Answer is basically "I did this, and it worked for me"; I think we might want something more official for the Homebrew docs.

apjanke: (And I realize you're basically inheriting a problem from the old version of the documentation here; just hoping we can find a better reference while we're going in and changing things.)

docs/FAQ.md
MikeMcQuaid: I think straight apostrophes are nicer because they avoid weird code monospace rendering issues so: 👍

akalin: I changed it back to a curly apostrophe to be consistent with how it was (and since I already had one in this PR anyway).

docs/FAQ.md
akalin: Unfortunately, it looks like the sole source of documentation for this is in the man page for `launchctl`, and it appears that Apple doesn't put those online anymore: https://forums.developer.apple.com/thread/46067 .

However, it's easy enough to just refer to the manpage. Did that.

akalin:
Please take another look!

MikeMcQuaid:
Thanks for your first contribution to Homebrew @akalin! Without people like you submitting PRs we couldn't run this project. You rock!

PR #3079, opened 2017-08-22
Added tests for os/mac/keg

mansimarkaur:

-----

### DIFF
--- a/Library/Homebrew/test/os/mac/keg_spec.rb
+++ b/Library/Homebrew/test/os/mac/keg_spec.rb
@@ -5,28 +5,54 @@ describe Keg do
 
   subject { described_class.new(keg_path) }
 
-  describe "#mach_o_files" do
-    let(:keg_path) { HOMEBREW_CELLAR/"a/1.0" }
+  let(:keg_path) { HOMEBREW_CELLAR/"a/1.0" }
+  let(:file) { keg_path/"lib/i386.dylib" }
 
-    before(:each) { (keg_path/"lib").mkpath }
+  before(:each) do
+    (keg_path/"lib").mkpath
+    cp dylib_path("i386"), file
+    subject.link
+  end
 
-    after(:each) { subject.unlink }
+  after(:each) { subject.unlink }
 
-    it "skips hardlinks" do
-      cp dylib_path("i386"), keg_path/"lib/i386.dylib"
-      ln keg_path/"lib/i386.dylib", keg_path/"lib/i386_hardlink.dylib"
+  describe "#change_dylib_id" do
+    it "does nothing if given id is same as file's dylib id" do
+      id = file.dylib_id
+      file.change_dylib_id(id)
+      expect(file.dylib_id).to eq(id)
+    end
+  end
 
-      subject.link
-      expect(subject.mach_o_files.count).to eq(1)
+  describe "#change_install_name" do
+    it "does nothing if given name is same as file's install name" do
+      file.ensure_writable do
+        subject.each_install_name_for(file) do |name|
+          file.change_install_name(name, name)
+          expect(name).to eq(name)
+        end
+      end
     end
 
-    it "isn't confused by symlinks" do
-      cp dylib_path("i386"), keg_path/"lib/i386.dylib"
-      ln keg_path/"lib/i386.dylib", keg_path/"lib/i386_hardlink.dylib"
-      ln_s keg_path/"lib/i386.dylib", keg_path/"lib/i386_symlink.dylib"
+    it "does nothing when install name start with '/'" do
+      file.ensure_writable do
+        subject.each_install_name_for(file) do |name|
+          new_name = subject.fixed_name(file, name)
+          file.change_install_name(name, new_name)
+          expect(name).not_to eq(new_name)
+        end
+      end
+    end
+  end
 
-      subject.link
-      expect(subject.mach_o_files.count).to eq(1)
+  describe "#require_relocation?" do
+    it "is set to false at initialization" do
+      expect(subject.require_relocation?).to be false
+    end
+
+    it "is set to true after linkage is fixed" do
+      subject.fix_dynamic_linkage
+      expect(subject.require_relocation?).to be true
     end
   end
 end


woodruffw:
> could you review this PR?

Sure, no problem.

Library/Homebrew/test/os/mac/keg_spec.rb
```diff
@@ -5,28 +5,65 @@
...
+    it "changes file's install name to given name" do
+      new_name = ""
+      file.ensure_writable do
+        subject.each_install_name_for(file) do |name|
+          new_name = subject.fixed_name(file, name)
+          subject.change_install_name(name, new_name, file)
+        end
+      end
+      subject.each_install_name_for(file) do |name|
+        expect(name).to eq(new_name)
```
woodruffw: Why are we expecting each install name in the file to be the same as the last fixed name (`new_name`)?

In the file we're testing against, the only install name is `/usr/lib/libSystem.B.dylib`. Since that begins with a `/`, it's one of the ones that we don't called `fixed_name` on in `fix_dynamic_linkage`:

```ruby
          # Don't fix absolute paths unless they are rooted in the build directory
          next if bad_name.start_with?("/") &&
                  !bad_name.start_with?(HOMEBREW_TEMP.to_s) &&
                  !bad_name.start_with?(HOMEBREW_TEMP.realpath.to_s)
```

Library/Homebrew/test/os/mac/keg_spec.rb
```diff
@@ -5,28 +5,65 @@
...
+    cp dylib_path("i386"), file
+  end
 
-    it "skips hardlinks" do
-      cp dylib_path("i386"), keg_path/"lib/i386.dylib"
-      ln keg_path/"lib/i386.dylib", keg_path/"lib/i386_hardlink.dylib"
+  after(:each) { subject.unlink }
 
+  describe "#change_dylib_id" do
+    before(:each) do
```
woodruffw: This can be moved into the toplevel `before`, can't it? Same with the one below.

Library/Homebrew/test/os/mac/keg_spec.rb
```diff
@@ -5,28 +5,65 @@
...
+  end
 
+  describe "#change_install_name" do
+    before(:each) do
       subject.link
-      expect(subject.mach_o_files.count).to eq(1)
+    end
+
+    it "does nothing if given name is same as file's install name" do
+      initial_name = ""
```
woodruffw: You shouldn't need to pre-declare `initial_name` here.

Library/Homebrew/test/os/mac/keg_spec.rb
```diff
@@ -5,28 +5,65 @@
...
+          subject.change_install_name(name, name, file)
+        end
+      end
+      subject.each_install_name_for(file) do |name|
+        expect(name).to eq(initial_name)
+      end
+    end
+
+    it "changes file's install name to given name" do
+      new_name = ""
```
woodruffw: Same here; you don't need to predeclare `new_name`.

Library/Homebrew/test/os/mac/keg_spec.rb
```diff
@@ -5,28 +5,65 @@
...
+      subject.change_dylib_id(id, file)
+      expect(file.dylib_id).to eq(id)
+    end
+
+    it "changes file's dylib id to given id" do
+      id = "blah.B.dylib"
+      file.ensure_writable do
+        subject.change_dylib_id(id, file)
+      end
+      expect(file.dylib_id).to eq(id)
```
woodruffw: I'm not sure why this is failing -- `change_dylib_id` is a pretty thin wrapper over `MachO::Tools.change_dylib_id`, and it works correctly for me in isolation:

```ruby
>> m.dylib_id #=> "libfoo.1.dylib"
>> MachO::Tools.change_dylib_id('i386.dylib', 'blah.B.dylib') #=> 4164
>> m = MachO.open('i386.dylib')
>> m.dylib_id #=> "blah.B.dylib"
```

Is it possible that `file` is being overwritten somewhere before this test?

mansimarkaur: The file object that is created when the file is opened (https://github.com/Homebrew/brew/blob/master/Library/Homebrew/vendor/macho/macho/tools.rb#L22) at has it's dylib_id changed while filename's dylib_id isn't changed.

Library/Homebrew/test/os/mac/keg_spec.rb
mansimarkaur: I was initially using `initial_name` in a different block from the one where it was declared. Hence, I was pre-declaring it. I'm using the expect assertion inside the same block now.

Library/Homebrew/test/os/mac/keg_spec.rb
mansimarkaur: We are expecting each install name in the file to be same as the last fixed name because there is only one file and only one install name for the file is /usr/lib/libSystem.B.dylib.
I'll try thinking of a case for which we actually modify the install name.

mansimarkaur:
Please review, @woodruffw 
Thanks!

Library/Homebrew/test/os/mac/keg_spec.rb
mistydemeo: It looks like the test's passing now - is it flaky? Did it get resolved since?

mansimarkaur:
The tests are still failing on my local. Could you try it on your system, @mistydemeo  and also re-run the travis build to ensure the tests are actually passing?

mistydemeo:
I can repro failures for `Keg#change_install_name changes file's install name to given name` and `Keg#change_dylib_id changes file's dylib id to given id` locally.

Library/Homebrew/test/os/mac/keg_spec.rb
woodruffw: Got it. I think that we can invert this test, then, since we actually expect install names that begin with a `/` to *never* be rewritten. `expect(name).to_not eq(new_name)` should do the trick, if I'm remembering my RSpec correctly.

We might also want to amend this test to only attempt to rewrite the first install name in the binary, since we might update the fixture in the future to contain additional `LC_LOAD_*` commands.

Library/Homebrew/test/os/mac/keg_spec.rb
woodruffw: @mansimarkaur You're right, but that shouldn't matter unless there's a race condition somewhere (i.e., the check is being performed before `write!` occurs inside of `change_install_name`). 

I'll do some digging when I have access to my OS X machine (tomorrow morning).

mansimarkaur: Even after the `write!`, file's dylib_id is changed while that is not true for filename.

Library/Homebrew/test/os/mac/keg_spec.rb
mansimarkaur: to rewrite the first install name in the binary, I should be running the loop only once. Is that what you mean, @woodruffw ?

MikeMcQuaid:
@woodruffw How's this looking to you now?

woodruffw:
@MikeMcQuaid I'm a bit stumped on the tests here - the only thing I can think of that would cause this behavior should be fixed in #3101, but @mansimarkaur rebased and confirmed that the tests are still failing.

I'm going to merge #3101 anyways since it'll bring a minor performance improvement with no breaking (public) changes, but otherwise those two failing tests might need to be stubbed out until I trace it down fully.

MikeMcQuaid:
@woodruffw Sounds good. Shout after that and we can rerun here and remove the tests if needed.

MikeMcQuaid:
@woodruffw What's up, here?

woodruffw:
#3139 covers this, I just haven't made time to add some unit tests for it yet.

I'll have that PR done tonight, and I think that should fix the behavior here. If not, then commenting those tests out for the time being is good by me (since they don't test the relocation behavior we currently use, only potential combinatons).

MikeMcQuaid:
> I'll have that PR done tonight, and I think that should fix the behavior here. If not, then commenting those tests out for the time being is good by me (since they don't test the relocation behavior we currently use, only potential combinatons).

:+1:. I think it's worth deleting rather than commenting out the tests. Would you mind doing that here and just merging when it's ✅ and you're happy with it? Thanks!

woodruffw:
No problem!

PR #3087, opened 2017-08-25
audit: Add previously unported GH patch audit rule in Patches cop

GauthamGoli:

-----
#2976 
Somehow I missed porting this one rule. Apologies for that :(

cc and Thanks @ilovezfs for finding this

### DIFF
--- a/Library/Homebrew/rubocops/patches_cop.rb
+++ b/Library/Homebrew/rubocops/patches_cop.rb
@@ -25,12 +25,22 @@ module RuboCop
 
         def patch_problems(patch)
           patch_url = string_content(patch)
+          gh_patch_param_pattern = %r{https?://github\.com/.+/.+/(?:commit|pull)/[a-fA-F0-9]*.(?:patch|diff)}
+          if regex_match_group(patch, gh_patch_param_pattern)
+            if patch_url !~ /\?full_index=\w+$/
+              problem <<-EOS.undent
+                GitHub patches should use the full_index parameter:
+                  #{patch_url}?full_index=1
+              EOS
+            end
+          end
+
           gh_patch_patterns = Regexp.union([%r{/raw\.github\.com/},
                                             %r{gist\.github\.com/raw},
                                             %r{gist\.github\.com/.+/raw},
                                             %r{gist\.githubusercontent\.com/.+/raw}])
           if regex_match_group(patch, gh_patch_patterns)
-            unless patch_url =~ /[a-fA-F0-9]{40}/
+            if patch_url !~ /[a-fA-F0-9]{40}/
               problem <<-EOS.undent.chomp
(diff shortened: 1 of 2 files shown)


Library/Homebrew/test/rubocops/patches_cop_spec.rb
```diff
@@ -108,6 +109,15 @@ def patches
                                   line: 5,
                                   column: 5,
                                   source: source }]
+        elsif patch_url =~ %r{https?://github\.com/.+/.+/(?:commit|pull)/[a-fA-F0-9]*.(?:patch|diff)}
+          expected_offenses = [{  message: <<-EOS.undent,
```
reitermarkus: Any reason why there are two spaces here after the bracket instead of one?

GauthamGoli: No. Will make it one.

reitermarkus: Also, it's weird that this isn't reported by RuboCop, since it usually suggests moving the variable outside of the `if`, when it is the same assignment in all branches.

GauthamGoli: https://github.com/Homebrew/brew/blob/master/Library/.rubocop.yml#L111

reitermarkus: I think this should be enabled for core code.

Library/Homebrew/rubocops/patches_cop.rb
```diff
@@ -25,6 +25,16 @@ def audit_formula(_node, _class_node, _parent_class_node, body)
 
         def patch_problems(patch)
           patch_url = string_content(patch)
+          gh_patch_param_pattern = %r{https?://github\.com/.+/.+/(?:commit|pull)/[a-fA-F0-9]*.(?:patch|diff)}
+          if regex_match_group(patch, gh_patch_param_pattern)
+            unless patch_url =~ /\?full_index=\w+$/
```
MikeMcQuaid: `if patch_url !~`

Library/Homebrew/test/rubocops/patches_cop_spec.rb
MikeMcQuaid: @reitermarkus @GauthamGoli I'm 🆒 with PRs to make changes like that. I'd like to see us get closer to the RuboCop defaults on as much as possible.

Library/Homebrew/rubocops/patches_cop.rb
GauthamGoli: Addressed.

MikeMcQuaid:
Thanks again for the great work @GauthamGoli!

Issue #3068, opened 2017-08-19
Add a check for branches of tapped repos


JasonQSY:
I think it is a good idea to add at least a check for the branches of tapped repos. Generally, we are always using the `master` branch. However, if it is incorrectly switched to another branch like `patch`, it is better that `brew update` or `brew doctor` could have a warning.

The situation is that if we switch to another branch `patch` in `usr/local/Homebrew/Library/Taps/homebrew/homebrew-core`, `brew update` works correctly because it just pull the remote `master` branch to the local one. However, the active branch is `patch` and there is no warning in both `brew update` and `brew doctor`. And if other formulas work correctly   requires a later version of formulas in `homebrew-core`, the installation fails.

It is not a big problem but tough to find the first time... That's why I think adding a check in `brew update` or at least `brew doctor` is a good idea. I could work on it if such a PR is welcomed.

alyssais:
Having a `doctor` warning sounds good to me. 👍

MikeMcQuaid:
@JasonQSY Could you try and open a pull request? [This document](http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html) should help and we're happy to walk you through anything else.

Thanks!

JasonQSY:
OK. I'll try to work on it.

MikeMcQuaid:
Fixed in https://github.com/Homebrew/brew/pull/3089.

PR #3089, opened 2017-08-26
Add check for HEAD ref in diagnostics

BenMusch:

-----

Addresses this issue: https://github.com/Homebrew/brew/issues/3068 by extending a diagnostic check so that it adds a warning to the console if `.git/HEAD` is not set to `/refs/heads/master`

Any thoughts on how to test this? Are there patterns established for testing things that interact with the filesystem?

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -787,19 +787,13 @@ module Homebrew
         end
 
         head = coretap_path.git_head
-        return unless head && head !~ %r{refs/heads/master}
+        return if head.nil? || head =~ %r{refs/heads/master}
 
         <<-EOS.undent
-          Suspicious #{CoreTap.instance} git head found.
+          Homebrew/homebrew-core is not on the master branch
 
-          With a non-standard head, your local version of Homebrew might not
-          have all of the changes intended for the most recent release. The
-          current git head is:
-            #{head}
-
-          Unless you have compelling reasons, consider setting the head to
-          point at the master branch by running:
-            git checkout master
+          Check out the master branch by running:
+            git -C "$(brew --repo homebrew/core)" checkout master
         EOS
       end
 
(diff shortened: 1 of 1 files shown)


Library/Homebrew/diagnostic.rb
```diff
@@ -785,6 +785,22 @@ def check_coretap_git_origin
               git -C "#{coretap_path}" remote set-url origin #{Formatter.url("https://github.com/Homebrew/homebrew-core.git")}
           EOS
         end
+
+        head = coretap_path.git_head
+        return unless head && head !~ %r{refs/heads/master}
```
MikeMcQuaid: Don't use `unless  ... &&`; it's hard to follow. Switch to an `if`.

Library/Homebrew/diagnostic.rb
```diff
@@ -785,6 +785,22 @@ def check_coretap_git_origin
...
+          Suspicious #{CoreTap.instance} git head found.
+
+          With a non-standard head, your local version of Homebrew might not
+          have all of the changes intended for the most recent release. The
+          current git head is:
+            #{head}
+
+          Unless you have compelling reasons, consider setting the head to
+          point at the master branch by running:
+            git checkout master
```
MikeMcQuaid: Replace all the above text with something like:

```
Homebrew/homebrew-core is not on the master branch

```

Want to avoid Git jargon where possible here.

MikeMcQuaid:
Fantastic work again, @BenMusch! Looking forward already to your next PR 😍

MikeMcQuaid:
@benmusch this had to be reverted unfortunately (see the linked stuff above) but I’d love if you could try again in a new PR and detail what commands you use to test this so we can try them too. Thanks again!

BenMusch:
Interesting, will investigate and re-open. Thanks for the heads up @MikeMcQuaid

ilovezfs:
Thanks, @BenMusch! For what it's worth, I think you need to use `--symbolic-full-name`.

PR #3094, opened 2017-08-26
doctor: Check for branch of core tap

BenMusch:

-----

Fixes the issues with https://github.com/Homebrew/brew/pull/3089 -- the `git_head` returns a commit sha.

I also went ahead and implemented the null object pattern in the git repo module -- was just a thought that came to me and I'm totally happy to drop the commit if anyone prefers it without that change

### DIFF
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -26,6 +26,7 @@ module Homebrew
         check_missing_deps
         check_for_linked_keg_only_brews
       ]
+      slow_checks = []
       methods = (checks.all.sort - slow_checks) + slow_checks
     else
       methods = ARGV.named
--- a/Library/Homebrew/cmd/doctor.rb
+++ b/Library/Homebrew/cmd/doctor.rb
@@ -26,7 +26,6 @@ module Homebrew
         check_missing_deps
         check_for_linked_keg_only_brews
       ]
-      slow_checks = []
       methods = (checks.all.sort - slow_checks) + slow_checks
     else
       methods = ARGV.named
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -1091,7 +1091,7 @@ module Homebrew
       end
 
       def all
-        methods.map(&:to_s).grep(/^check_coretap_git_config/)
+        methods.map(&:to_s).grep(/^check_/)
       end
     end # end class Checks
   end
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -786,8 +786,9 @@ module Homebrew
           EOS
         end
 
-        head = coretap_path.git_head
-        return if head.nil? || head =~ %r{refs/heads/master}
+        branch = coretap_path.git_branch
+        return if branch.nil? || branch =~ %r{master}
+
 
         <<-EOS.undent
           Homebrew/homebrew-core is not on the master branch
@@ -1090,7 +1091,7 @@ module Homebrew
       end
 
       def all
-        methods.map(&:to_s).grep(/^check_/)
+        methods.map(&:to_s).grep(/^check_coretap_git_config/)
       end
     end # end class Checks
   end
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -787,8 +787,7 @@ module Homebrew
         end
 
         branch = coretap_path.git_branch
-        return if branch.nil? || branch =~ %r{master}
-
+        return if branch.nil? || branch == "master"
 
         <<-EOS.undent
           Homebrew/homebrew-core is not on the master branch
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -786,11 +786,13 @@ module Homebrew
           EOS
         end
 
+        return if ENV["CI"] || ENV["JENKINS_HOME"]
+
         branch = coretap_path.git_branch
         return if branch.nil? || branch == "master"
 
         <<-EOS.undent
-          Homebrew/homebrew-core is not on the master branch
+        #{CoreTap.instance} is not on the master branch
 
           Check out the master branch by running:
             git -C "$(brew --repo homebrew/core)" checkout master
--- a/Library/Homebrew/extend/git_repository.rb
+++ b/Library/Homebrew/extend/git_repository.rb
@@ -36,6 +36,15 @@ module GitRepositoryExtension
     end
   end
 
+  def git_branch
+    return unless git? && Utils.git_available?
+    cd do
+      Utils.popen_read(
+        "git", "rev-parse", "--abbrev-ref", "HEAD"
+      ).chuzzle
+    end
+  end
+
   def git_last_commit_date
     return unless git? && Utils.git_available?
     cd do
--- a/Library/Homebrew/extend/git_repository.rb
+++ b/Library/Homebrew/extend/git_repository.rb
@@ -7,33 +7,39 @@ module GitRepositoryExtension
   end
 
   def git_origin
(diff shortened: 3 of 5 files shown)


Library/Homebrew/cmd/doctor.rb
```diff
@@ -26,6 +26,7 @@ def doctor
         check_missing_deps
         check_for_linked_keg_only_brews
       ]
+      slow_checks = []
```
BenMusch: Used this & the modification in `all` to debug. Removing now

MikeMcQuaid:
@BenMusch Can you elaborate what commands you're using to test this locally? Thanks!

Library/Homebrew/extend/git_repository.rb
```diff
@@ -7,41 +7,52 @@ def git?
...
-      Utils.popen_read(
-        "git", "show", "-s", "--format=%cd", "--date=short", "HEAD"
-      ).chuzzle
+    git_client.send_command("show", "-s", "--format=%cd", "--date=short", "HEAD")
+  end
+
+  private
+
+  def git_client
+    @client ||= (git? && Utils.git_available?) ? GitClient.new(self) : NullClient.new
```
MikeMcQuaid: Use `begin ... end` and `if ... `else` here rather than the ternary operator to make it a little easier to follow.

Library/Homebrew/extend/git_repository.rb
```diff
@@ -7,41 +7,52 @@ def git?
...
+
+    def send_command(*args)
+      path.cd do
+        Utils.popen_read("git", *args).chuzzle
+      end
+    end
+  end
+
+  class NullClient
+    def send_command(*_args)
```
MikeMcQuaid: Can just be `*` I believe?

Library/Homebrew/extend/git_repository.rb
```diff
@@ -7,41 +7,52 @@ def git?
...
+      nil
```
MikeMcQuaid: Can skip this `nil`.

Library/Homebrew/extend/git_repository.rb
```diff
@@ -7,41 +7,52 @@ def git?
...
+  end
+
+  class GitClient
+    attr_reader :path
+
+    def initialize(path)
+      @path = path
+    end
+
+    def send_command(*args)
```
MikeMcQuaid: Not a huge fan of this name but struggling to think of something better. Maybe `git` or `run` or something?

MikeMcQuaid:
> codecov/patch — 76.92% of diff hit (target 53.1%)

@BenMusch Would be good to try and raise this if possible, thanks!

BenMusch:
@MikeMcQuaid Here's how I tested locally: https://asciinema.org/a/Oh3pde0ptrUZTGab1rq6tfzz5

BenMusch:
I've added a test for the git_branch method, lemme know if there's anything else that comes across as test-worthy here

MikeMcQuaid:
@BenMusch Could you list the series of commands instead? Thanks!

Library/Homebrew/diagnostic.rb
```diff
@@ -785,6 +785,16 @@ def check_coretap_git_origin
               git -C "#{coretap_path}" remote set-url origin #{Formatter.url("https://github.com/Homebrew/homebrew-core.git")}
           EOS
         end
+
```
MikeMcQuaid: Add a `return if ENV["CI"] || ENV["JENKINS_HOME"]` so this isn't hit on CI.

Library/Homebrew/diagnostic.rb
```diff
@@ -785,6 +785,16 @@ def check_coretap_git_origin
...
+        branch = coretap_path.git_branch
+        return if branch.nil? || branch == "master"
```
MikeMcQuaid: When could `branch` be `nil`?

Library/Homebrew/extend/git_repository.rb
MikeMcQuaid: Let's go for `run`

Library/Homebrew/tap.rb
```diff
@@ -118,6 +119,12 @@ def git?
     path.git?
   end
 
+  # The current git branch of this {Tap}.
+  def git_branch
+    raise TapUnavailableError, name unless installed?
```
MikeMcQuaid: How about punting the `TapUnavailableError` somehow into the `GitClient` so you can avoid repeating this. All it's checking is `path.directory?` which you could do in the client.

BenMusch:
Sure:

```bash
$ cd /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core
$ git status
On branch master

Your branch is up-to-date with 'origin/master'.
nothing to commit, working tree clean
$ brew doctor
[some output, none about the branch]
$ git checkout -b test
Switched to branch 'test'
$ brew doctor

Warning: Homebrew/homebrew-core is not on the master branch

[more output]
$ git checkout master
$ brew doctor
[some output, none about the branch]
```

Library/Homebrew/diagnostic.rb
BenMusch: If git was unavailable it'd be nil

Library/Homebrew/diagnostic.rb
BenMusch: Happy to do that. Just curious, why does it matter if CI hits this? Is it just for test coverage?

Library/Homebrew/extend/git_repository.rb
```diff
@@ -43,16 +49,15 @@ def initialize(path)
       @path = path
     end
 
-    def send_command(*args)
+    def run(*args)
+      raise TapUnavailableError, path.name unless path.installed?
```
BenMusch: Maybe this should be raised on initialization? Or perhaps we can include this in the `git_client` method, and then use a NullClient when it's not available?

BenMusch: (I'll fix the test failure related to this once I have a clearer idea on where this should happen)

Library/Homebrew/diagnostic.rb
reitermarkus: It is because `brew doctor` should not fail on CI. PRs would fail, because they are not on `master`.

Library/Homebrew/extend/git_repository.rb
```diff
@@ -7,41 +7,57 @@ def git?
...
+
+  class GitClient
+    attr_reader :path
+
+    def initialize(path)
+      @path = path
+    end
+
+    def run(*args)
+      raise TapUnavailableError, path.name unless path.directory?
```
BenMusch: Posting again just because it's hidden using the "outdated diff" feature:

Maybe this should be raised on initialization? Or perhaps we can include this in the `git_client` method, and then use a NullClient when it's not available?

I'll address the related test failure once I've got a better idea of when I wanna raise this

reitermarkus: I think this is the right place to raise, but not the right exception, since the path doesn‘t have to be a Tap.

BenMusch: True. Although in the context that this is used in `tap.rb`, it seems like `TapUnavailableError` would make the most sense. What do you think about passing an `on_error` block to the constructor?

MikeMcQuaid: > What do you think about passing an on_error block to the constructor?

Give it a go and let's see what it looks like, thanks!

BenMusch:
Started classes last week so I'm pretty busy, should I pull out the actual feature (checking for master branch) into a separate PR from the refactoring? That way we can get that merged sooner

MikeMcQuaid:
> should I pull out the actual feature (checking for master branch) into a separate PR from the refactoring

👍

BenMusch:
Closing in favor of https://github.com/Homebrew/brew/pull/3229, will do the refactor some other time

PR #3095, opened 2017-08-27
Tests: override HOME

mistydemeo:

-----

This is an alternate fix for #2197.

This prevents tools like git and svn from reading their configuration in the user's home directory. This is useful, for example, if the user has git configured to GPG sign every commit, which would cause every place we `git commit` in the tests to fail.

I chose `HOMEBREW_CACHE` over `TEST_TMPDIR` in order to avoid introducing file leakage from newly-written dotfiles. We clear out `HOMEBREW_CACHE` before checking for leakage.

cc @alyssais, @woodruffw

### DIFF
--- a/Library/Homebrew/test/spec_helper.rb
+++ b/Library/Homebrew/test/spec_helper.rb
@@ -86,6 +86,9 @@ RSpec.configure do |config|
       @__argv = ARGV.dup
       @__env = ENV.to_hash # dup doesn't work on ENV
 
+      # Makes sure config files for git, svn, etc. aren't read
+      ENV["HOME"] = HOMEBREW_CACHE.to_s
+
       unless example.metadata.key?(:focus) || ENV.key?("VERBOSE_TESTS")
         @__stdout = $stdout.clone
         @__stderr = $stderr.clone


mistydemeo:
Hm, looks like changing the home directory interferes with the Linux tests.

woodruffw:
Weird. Only two failures in the `brew search` tests, and the last line they trace back to in our code is where we restore `ENV` (` ./test/spec_helper.rb:99`). Not sure what to make of that, but userland programs on Linux frequently find the user's "home" in all sorts of incorrect ways (`/home/$USER`, `/etc/passwd`, etc). Maybe this is an instance of that?

alyssais:
We might want to try setting this just once in `cmd/tests.rb`? It doesn't need to be set and restored for every test, and that's where all the other test environment setup happens. Might help with whatever weirdness @woodruffw is describing.

MikeMcQuaid:
> We might want to try setting this just once in cmd/tests.rb?

👍 if it works.

reitermarkus:
Another thought: I think `File.expand_path` doesn‘t use `HOME` to expand `~`, which might break stuff.

woodruffw:
`expand_path` respects `ENV["HOME"]` for me, on Linux:

```ruby
>> ENV["HOME"] = '/tmp' #=> "/tmp"
>> File.expand_path '~' #=> "/tmp"
```

That's 2.4.1, though, so maybe 2.3.x and below have different behavior?

reitermarkus:
I can‘t check right now, I just remember it from the Cask merger, where we had to use `HOME` instead of `expand_path` because it wouldn‘t respect Jenkins‘ `HOME` directory.

MikeMcQuaid:
> I can‘t check right now, I just remember it from the Cask merger, where we had to use HOME instead of expand_path because it wouldn‘t respect Jenkins‘ HOME directory.

Yes, correct 👍

PR #3096, opened 2017-08-27
audit: Add some heuristics to https upgrade checks

vszakats:

-----

Following up on https://github.com/Homebrew/homebrew-core/pull/17181#issuecomment-325120973, this patch tries to extend HTTPS-upgrade audit checks to detect additional cases that may allow an upgrade. The logic was borrowed from this [project](https://github.com/vszakats/https-check).

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -240,26 +240,25 @@ class FormulaAuditor
       return "The URL #{url} could use HTTPS rather than HTTP"
     end
 
-    if type == "homepage"
+    return if type != "homepage"
 
-      details[:file] = details[:file].gsub(/https?:\\?\/\\?\//, '/')
-      secure_details[:file] = secure_details[:file].gsub(/https?:\\?\/\\?\//, '/')
+    details[:file] = details[:file].gsub(%r{https?:\\?\/\\?\/}, "/")
+    secure_details[:file] = secure_details[:file].gsub(%r{https?:\\?\/\\?\/}, "/")
 
-      # Same content after normalization
-      if details[:file] == secure_details[:file]
-        return "The URL #{url} could use HTTPS rather than HTTP"
-      end
+    # Same content after normalization
+    if details[:file] == secure_details[:file]
+      return "The URL #{url} could use HTTPS rather than HTTP"
+    end
 
-      # Same size, different content after normalization
-      # (typical causes: Generated ID, Timestamp, Unix time)
-      if details[:file].length == secure_details[:file].length
-        return "The URL #{url} could use HTTPS rather than HTTP"
-      end
+    # Same size, different content after normalization
+    # (typical causes: Generated ID, Timestamp, Unix time)
+    if details[:file].length == secure_details[:file].length
+      return "The URL #{url} could use HTTPS rather than HTTP"
+    end
 
-      lenratio = (100 * secure_details[:file].length / details[:file].length).to_i
-      if lenratio >= 90 && lenratio <= 120
-        return "The URL #{url} may be able to use HTTPS rather than HTTP. Please verify it in a browser."
-      end
+    lenratio = (100 * secure_details[:file].length / details[:file].length).to_i
+    if lenratio >= 90 && lenratio <= 120
+      return "The URL #{url} may be able to use HTTPS rather than HTTP. Please verify it in a browser."
     end
   end
 
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -260,7 +260,7 @@ class FormulaAuditor
     end
 
     lenratio = (100 * secure_details[:file].length / details[:file].length).to_i
-    return if lenratio < 90 || lenratio > 110
+    return unless (90..110).cover?(lenratio)
     "The URL #{url} may be able to use HTTPS rather than HTTP. Please verify it in a browser."
   end
 
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -260,7 +260,7 @@ class FormulaAuditor
     end
 
     lenratio = (100 * secure_details[:file].length / details[:file].length).to_i
-    return if lenratio < 90 || lenratio > 120
+    return if lenratio < 90 || lenratio > 110
     "The URL #{url} may be able to use HTTPS rather than HTTP. Please verify it in a browser."
   end
 
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -201,7 +201,7 @@ class FormulaAuditor
     @specs = %w[stable devel head].map { |s| formula.send(s) }.compact
   end
 
-  def self.check_http_content(type, url, name, user_agents: [:default])
+  def self.check_http_content(url, name, user_agents: [:default], check_content: false)
     return unless url.start_with? "http"
 
     details = nil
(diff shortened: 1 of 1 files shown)


vszakats:
I'm in the dark how unit tests could be fabricated for this.

MikeMcQuaid:
@vszakats We can punt on unit tests for this for now, don't worry about it. Sorry for the delay, I will review this but just have a recent 👶 stealing my time.

vszakats:
@MikeMcQuaid Wow, congrats!! :) 🎉

(as for tests: maybe with `file:///` fixtures)

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -201,7 +201,7 @@ def initialize(formula, options = {})
     @specs = %w[stable devel head].map { |s| formula.send(s) }.compact
   end
 
-  def self.check_http_content(url, name, user_agents: [:default])
+  def self.check_http_content(type, url, name, user_agents: [:default])
```
MikeMcQuaid: Move `type` to be the last argument named `check_content: false` or similar.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
       details[:content_length] == secure_details[:content_length]
     file_match = details[:file_hash] == secure_details[:file_hash]
 
-    return if !etag_match && !content_length_match && !file_match
-    "The URL #{url} could use HTTPS rather than HTTP"
+    if etag_match || content_length_match || file_match
+      return "The URL #{url} could use HTTPS rather than HTTP"
```
MikeMcQuaid: While we're here let's go for `should use`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+    end
+
+    return if type != "homepage"
```
MikeMcQuaid: `return unless check_content`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+
+    details[:file] = details[:file].gsub(%r{https?:\\?\/\\?\/}, "/")
```
MikeMcQuaid: Use a new variable for this e.g. `no_protocol_file_contents =`. The `\/` should not be needed in a `%r`. What's the `\\?`; sorry, being stupid here.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+    secure_details[:file] = secure_details[:file].gsub(%r{https?:\\?\/\\?\/}, "/")
```
MikeMcQuaid: Same here: `no_protocol_https_file_contents` or similar

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+
+    # Same content after normalization
```
MikeMcQuaid: `# Check for the same content after removing all protocols`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+    if details[:file] == secure_details[:file]
+      return "The URL #{url} could use HTTPS rather than HTTP"
```
MikeMcQuaid: Let's go for `should` here.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+    end
+
+    # Same size, different content after normalization
+    # (typical causes: Generated ID, Timestamp, Unix time)
+    if details[:file].length == secure_details[:file].length
```
MikeMcQuaid: I think both this check and the one below should be only for `--strict` (maybe `--new-formula`?) because they (presumably) will produce some false positives. @ilovezfs, thoughts?

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,29 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+      return "The URL #{url} could use HTTPS rather than HTTP"
```
MikeMcQuaid: This message I think should be the same as below; it may be able to use it but should be verified.

Library/Homebrew/dev-cmd/audit.rb
vszakats: `\\?` is to catch protocol prefixes in embedded JavaScript code, escaped as: `https:\/\/`.

vszakats:
Thanks for your kind words @MikeMcQuaid - trying my best! Addressed your review notes, except for the one pending on @ilovezfs's input.

ilovezfs:
>I think both this check and the one below should be only for --strict (maybe --new-formula?) because they (presumably) will produce some false positives. @ilovezfs, thoughts?

`--strict` makes sense

Library/Homebrew/dev-cmd/audit.rb
vszakats: If there is precedent for an audit _warning_, it could be used here. Per past experience a length match at this point is an pretty-sure bet for upgradability.

vszakats:
Addressed all requested changes, including to limit some heuristics to strict mode.

MikeMcQuaid:
Great work here @vskatas! Looks good to go from my perspective. @ilovezfs, any thoughts?

@vszakats can you provide some URLs I can test this with?

vszakats:
@MikeMcQuaid Thanks, I'm glad you liked it! I've attached a `.zip` with two test formulae, for two upgrade cases: [tests.zip](https://github.com/Homebrew/brew/files/1268787/tests.zip)

Couldn't find a live example for the 'same size, different content after removing protocols' case though - it should also catch homepages with 'same size, different content and no Content-Length or ETag'.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,32 @@ def self.check_http_content(url, name, user_agents: [:default])
...
+    return unless strict
+
+    # Same size, different content after normalization
+    # (typical causes: Generated ID, Timestamp, Unix time)
+    if details[:file].length == secure_details[:file].length
+      return "The URL #{url} may be able to use HTTPS rather than HTTP. Please verify it in a browser."
+    end
+
+    lenratio = (100 * secure_details[:file].length / details[:file].length).to_i
+    return if lenratio < 90 || lenratio > 110
```
reitermarkus: `return unless (90..110).cover?(lenratio)`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,8 +236,32 @@ def self.check_http_content(url, name, user_agents: [:default])
       details[:content_length] == secure_details[:content_length]
     file_match = details[:file_hash] == secure_details[:file_hash]
 
-    return if !etag_match && !content_length_match && !file_match
-    "The URL #{url} could use HTTPS rather than HTTP"
+    if etag_match || content_length_match || file_match
+      return "The URL #{url} should use HTTPS rather than HTTP"
```
reitermarkus: Use `Formatter.url(url)` for the URLs.

vszakats: `#{url}` is used this way throughout this whole source file, so I think moving to `Formatter.url(url)` would better be done in a separate PR addressing it for all lines.
Unless I missing something and above is a special case, but since it is an existing line, I'm not sure.

Library/Homebrew/dev-cmd/audit.rb
vszakats: Committed, thanks!

Library/Homebrew/dev-cmd/audit.rb
reitermarkus: Yep, a separate PR is probable a better idea.

MikeMcQuaid:
Thanks for the great work @vszakats and for stepping up! I appreciate you getting stuck in and hope to see more PRs from you like this in future 👏

vszakats:
Thank you @MikeMcQuaid for the merge and all of you for the reviews! It's great to have some true code contributed to Homebrew.

PR #3098, opened 2017-08-28
gpg: tweaks

DomT4:
-----

Ref: https://github.com/Homebrew/homebrew-core/pull/17335

### DIFF
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -7,8 +7,7 @@ class Gpg
       next unless gpg_short_version
       gpg_version = Version.create(gpg_short_version.to_s)
       @version = gpg_version
-      gpg_version == Version.create("2.1") ||
-        gpg_version == Version.create("2.0")
+      gpg_version >= Version.create("2.0")
     end
   end
 
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -37,12 +37,21 @@ class Gpg
       Key-Length: 2048
       Subkey-Type: RSA
       Subkey-Length: 2048
-      Passphrase: ''
       Name-Real: Testing
       Name-Email: testing@foo.bar
       Expire-Date: 1d
+      %no-protection
       %commit
     EOS
     system GPG_EXECUTABLE, "--batch", "--gen-key", "batch.gpg"
+  end
+
+  def self.cleanup_test_processes!
+    odie "No GPG present to test against!" unless available?
+    gpgconf = Pathname.new(GPG_EXECUTABLE).parent/"gpgconf"
+
+    system gpgconf, "--kill", "gpg-agent"
+    system gpgconf, "--homedir", "keyrings/live", "--kill",
+                                 "gpg-agent"
   end
 end
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -54,4 +54,13 @@ class Gpg
     system gpgconf, "--homedir", "keyrings/live", "--kill",
                                  "gpg-agent"
   end
+
+  def self.test(path)
+    create_test_key(path)
+    begin
+      yield
+    ensure
+      cleanup_test_processes!
+    end
+  end
 end
--- a/Library/Homebrew/requirements/gpg2_requirement.rb
+++ b/Library/Homebrew/requirements/gpg2_requirement.rb
@@ -5,8 +5,8 @@ class GPG2Requirement < Requirement
   fatal true
   default_formula "gnupg"
 
-  # GPGTools installs GnuPG 2.0.x as a vanilla `gpg` symlink
-  # pointing to `gpg2`. Homebrew install 2.1.x as a non-symlink `gpg`.
-  # We support both the 2.0.x "stable" and 2.1.x "modern" series here.
+  # GPGTools installs GnuPG 2.0.x as a `gpg` symlink pointing
+  # to `gpg2`. Our `gnupg` installs only a non-symlink `gpg`.
+  # The aim is to retain support for any version above 2.0.
   satisfy(build_env: false) { Gpg.gpg || Gpg.gpg2 }
 end


DomT4:
Tests still fail locally, which is why I lopped off the last two checkboxes on the list, but nothing related:

```
Failures:

  1) Homebrew #search_taps
     Failure/Error:
       expect(described_class.search_taps("some-formula"))
         .to match(["homebrew/foo/some-formula"])

       expected [] to match ["homebrew/foo/some-formula"]
     # ./test/cmd/search_remote_tap_spec.rb:18:in `block (2 levels) in <top (required)>'

Finished in 1 minute 4.81 seconds (files took 3.55 seconds to load)
211 examples, 1 failure, 1 pending

Failed examples:

rspec ./test/cmd/search_remote_tap_spec.rb:4 # Homebrew #search_taps
```

Library/Homebrew/gpg.rb
```diff
@@ -7,7 +7,7 @@ def self.find_gpg(executable)
       next unless gpg_short_version
       gpg_version = Version.create(gpg_short_version.to_s)
       @version = gpg_version
-      gpg_version == Version.create("2.1") ||
+      gpg_version >= Version.create("2.1") ||
```
fxcoudert: Any reason why we don't want `gpg_version >= Version.create("2.0")`?

DomT4: That would work too, aye. I was partially thinking in the future that `2.0` is going to go away anyway because upstream are dropping support for it, but yeah, there's no actual reason to wait to compact it down to one line.

JCount:
We probably should have coverage for `cleanup_test_processes`.

DomT4:
I'm not sure we even have coverage for `create_test_key` at the moment because CI just skips over all the GPG stuff & it's a really solid way for tests to break if they only ever get run once in a blue moon, but I can look into it.

DomT4:
I did add initial testing last year. I can't find it around any more, so I'm presuming it was removed either for the above reason or something else. I'll try to find out why/where.

JCount:
I think the test are still there because I believe I tweaked it when I was updating this code as part of the switch to having `2.1` be the default.

JCount:
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/gpg2_requirement_spec.rb

DomT4:
Those are just the requirement tests though, rather than actual functionality testing. The last reference I can find to the latter is https://github.com/Homebrew/brew/commit/14a7ef591b68ebb4d901c2f2a9a1553be0e7af03.

JCount:
It appears that `gpg_spec.rb` was removed in https://github.com/Homebrew/brew/pull/2746 and I am not entirely certain if that was intentional or not, given the way it does not really seem to align with the stated purpose of the PR.

DomT4:
Not sure. If @MikeMcQuaid was removing tests which weren't actually getting executed on CI because they needed some special handling/capability it would be a candidate.

MikeMcQuaid:
> Not sure. If @MikeMcQuaid was removing tests which weren't actually getting executed on CI because they needed some special handling/capability it would be a candidate.

Yeh, that sounds like something I'd have done.

Library/Homebrew/gpg.rb
```diff
@@ -38,12 +37,21 @@ def self.create_test_key(path)
       Key-Length: 2048
       Subkey-Type: RSA
       Subkey-Length: 2048
-      Passphrase: ''
```
MikeMcQuaid: Think this was just added; does `%no-protection` do the same thing?

Library/Homebrew/gpg.rb
```diff
@@ -38,12 +37,21 @@ def self.create_test_key(path)
...
       Name-Real: Testing
       Name-Email: testing@foo.bar
       Expire-Date: 1d
+      %no-protection
       %commit
     EOS
     system GPG_EXECUTABLE, "--batch", "--gen-key", "batch.gpg"
   end
+
+  def self.cleanup_test_processes
```
MikeMcQuaid: `self.cleanup_test_processes!` probably because it's going to kill things.

MikeMcQuaid:
A few comments but otherwise 👍 from me.

Library/Homebrew/gpg.rb
DomT4: Aye.

Library/Homebrew/gpg.rb
DomT4: Sure, will change.

MikeMcQuaid:
> Which is just a bit nicer in terms of only having to change one block of code if anything needs modifying over time.

It might be nice to try and make that a single block like:

```ruby
Gpg.test(testpath) do
  system bin/"formula_executable", "do", "xyz"
end
```

which implicitly calls the other methods?

DomT4:
Will look into it 👍.

DomT4:
Tweaks pushed. Seems to work alright with a modified `duplicity` test locally.

ilovezfs:
I've been enjoying not having these tests depend on opaque backend magic.

DomT4:
I'm not sure opaque is... fair. The code is simply enough to understand what it's doing, and I can add code comments or broader documentation on usage if desired. A lot of Homebrew is nudged behind the scenes to reduce extensive code duplication.

There's also the downside that the current system in core _only_ supports Homebrew's `gnupg`, which means if you're updating those formulae locally & want to run the tests you can't unless you temporarily install a whole new GnuPG backend. Given the requirement explicitly supports external tools the current core setup is, arguably, a violation of that intention.

ilovezfs:
It's more that someone would have to go looking for this method and its docs to understand what in the world is going on in the tests. Note we've had exactly zero people complain that they only support Homebrew's gpg since I unwound them to no longer use the DSL.

DomT4:
> It's more that someone would have to go looking for this method and its docs to understand what in the world is going on in the tests

You can use this argument with so much of Homebrew though. Look at what we do with Python formulae that need `resource`s these days. Even I had to go check the documentation the other week to double check I wasn't misapplying the syntax tweaks I'd made to a local formula, and there's almost no commit in this project I miss being merged or discussed.

There's also the magic around our temporary hijacking of `HOME`, and a host of other variables. I'd have a lot of sympathy for your argument if I was proposing some significant departure from the status quo but backend magic is the status quo with Homebrew probably more often than is not the case.

Whether or not that's a good thing or not is something I'm sure could be debated, but by Homebrew's standards even if you disagree with backend magic this is a pretty minor piece of sorcery.

ilovezfs:
We've actually been talking about unwinding all of the language DSLs. Given I already did the work to make that happen with this one, this seems like a step backward.

>Even I had to go check the documentation the other week to double check I wasn't misapplying the syntax tweaks I'd made to a local formula, and there's almost no commit in this project I miss being merged or discussed.

That is a problem.

DomT4:
I've got to sleep. I'm happy for people to keep chucking discussion around here & I'll get back to it later today/sometime tomorrow depending on how tiring today proves to be 😸.

DomT4:
Any other opinions here? If we're going to be stuck debating it for a while I'll pull the non-controversial elements into another PR since currently the GPG Requirement is wrongly unfulfilled if you've updated your `gnupg` formula.

ilovezfs:
It's not actively making anything _worse_ so shipped :) Thanks Mr. T

DomT4:
Thanks ILZ. I'll hold off on filing anything against core in terms of test rewriting until there's some consensus on Homebrew's fondness for user-invisible magic.

ilovezfs:
@DomT4 it would make sense to open a PR against duplicity since that's actually broken if you want.

DomT4:
@ilovezfs I think that should be fixed now? Unless I'm missing something. But if you want me to look at it just let me know whether you want it written in the current style or using the new/proposed style & I'll do so.

ilovezfs:
@DomT4 It's using `Gpg.create_test_key(testpath)` without a block, so I assume it's not running the cleanup, right?

>whether you want it written in the current style or using the new/proposed style

I'd use the new style so we have at least one formula using the code :)

DomT4:
> so I assume it's not running the cleanup, right?

Aye, correct. It'll be leaving running processes.

> I'd use the new style so we have at least one formula using the code :)

Alright, I'll get that filed in the next 10 minutes or so. At least it's there to point to if y'all decide to press ahead with the changes or easy enough to revert if not 😸.

ilovezfs:
>At least it's there to point to if y'all decide to press ahead with the changes or easy enough to revert if not :smile_cat:.

Yup!

>Alright, I'll get that filed in the next 10 minutes or so. 

:bowing_man:

DomT4:
There's a new significant release available, hence the minor delay. Working on it.

ilovezfs:
no worries

DomT4:
Hitting this fun with `poet` again so I may end up doing the resources manually I guess:
```
pkg_resources.VersionConflict: (pbr 3.1.1 (/private/tmp/venv/lib/python2.7/site-packages), Requirement.parse('pbr<2.0,>=1.6'))
```

ilovezfs:
hmm

ilovezfs:
I think the offender is `lib/python2.7/site-packages/python_novaclient-2.27.0.dist-info/METADATA:Requires-Dist: pbr (<2.0,>=1.6)`

So that's the rabbit hole if you want to start digging :)

ilovezfs:
```
Collecting python-novaclient==2.27.0 (from pyrax)
```

ilovezfs:
Lovely: https://github.com/rackspace/pyrax/blame/master/setup.py#L87

ilovezfs:
Here's what I get if I unpin python-novaclient: https://gist.github.com/ilovezfs/c7e58b3cc6ffe3ef18fcf03b3b102703

ilovezfs:
which is competing with python-keystoneclient
```
Collecting pbr!=2.1.0,>=2.0.0 (from python-keystoneclient)
  Using cached pbr-3.1.1-py2.py3-none-any.whl
```

ilovezfs:
@DomT4 OK this should work:
```
pip2 install azure-storage boto pyrax dropbox fasteners kerberos mega.py paramiko pexpect pycrypto pycryptopp python-swiftclient 'python-keystoneclient<3.11.0' requests requests-oauthlib ndg-httpsclient pyOpenSSL 'pbr<2' 'stevedore<1.21.0' 'oslo.utils<3.23.0' 'oslo.serialization<2.17.0' 'oslo.i18n<3.13.0' 'keystoneauth1<2.19.0' 'debtcollector<1.12.0' .
Processing /Users/joe/Developer/duplicity0714/duplicity-0.7.14
```

https://gist.github.com/ilovezfs/31099485d8ad62edcc42f7e5d05b37da

azure_nspkg-2.0.0-py2.py3-none-any.whl I guess doesn't have an sdist

ilovezfs:
@DomT4 arguably we should just drop pyrax support from duplicity until/unless they unpin python_novaclient since it's having an undue impact on the rest of the tree.

DomT4:
Sorry, wandered off to eat. IIRC `pyrax` has been deprecated upstream, so dropping it makes sense to me given those dependency pins are going to be permanent & keep causing hassle every time `duplicity` sees a new upstream release.

DomT4:
(And thanks for spending the time looking into it! Wasn't intending to throw the work onto you, just needed to eat to feel a little more alive today heh).

ilovezfs:
Yup, deprecated:
https://github.com/rackspace/pyrax/commit/dc33a91e099643c19488cc94abbdcadaab67399d

Let's kill it.

PR #3099, opened 2017-08-28
popen: Do not suppress stderr

sjackman:

-----

Suppressing stderr makes troubleshooting failed commands difficult.

### DIFF
--- a/Library/Homebrew/utils/popen.rb
+++ b/Library/Homebrew/utils/popen.rb
@@ -13,7 +13,7 @@ module Utils
         return pipe.read unless block_given?
         yield pipe
       else
-        $stderr.reopen("/dev/null", "w")
+        $stderr.reopen("/dev/null", "w") if !ARGV.debug? && !ARGV.verbose?
         exec(*args)
       end
     end


sjackman:
One test is failing:

> expected block to not output to stderr, but output "Error: No such file or directory - /usr/bin/fgrep\nError: No such file or directory - /usr/bin/fgrep\nError: No such file or directory - /usr/bin/fgrep\n"

https://travis-ci.org/Homebrew/brew/jobs/269276988#L1745

JCount:
Would it make sense to guard this with a check to see whether `debug` is enabled, rather than unilaterally disabling it? Otherwise, this seems to have the potential to dump a lot of confusing output into the average user's stderr, potentially complicating matters.

(If something like this is already in place, feel free to disregard.)

sjackman:
I've addressed the above test failure by silencing the stderr of `fgrep`. The `-s` option to silence stderr is portable to both BSD grep and GNU grep.

sjackman:
@JCount I feel that ignoring error messages wholesale is a dangerous default. When a particular tool is expected to output error message (like `fgrep` which caused a test to fail), its stderr should be selectively silenced.

JCount:
I was advocating for selective silencing, not against it. Using the developer mode variable or the debug flag were merely a basic example to try to clarify what I meant.

MikeMcQuaid:
I'm not sure about making this change. This is essentially a public formula API at this stage and will make formulae start outputting things they didn't before. I think it's a possibility if, as @JCount suggests, it's guarded on `ARGV.debug?` and/or `ARGV.verbose?`. CC @ilovezfs for some thoughts here, too.

sjackman:
I've modified the logic to be
```ruby
$stderr.reopen("/dev/null", "w") unless ARGV.debug? || ARGV.verbose?
```
I'm still happy to receive comments and change it as desired.

ilovezfs:
So if we do `foo = Utils.popen_read("blah")` does foo get set to some mixture of stdout and stderr? If so, that is not good.

ilovezfs:
e.g. I'm pretty sure this will blow up https://github.com/Homebrew/brew/blob/master/Library/Homebrew/language/node.rb#L13

sjackman:
> So if we do `foo = Utils.popen_read("blah")` does foo get set to some mixture of stdout and stderr?

No, the `$stderr` of `blah` does not go to `foo`. `$stderr` of the child process goes to `$stderr` of the parent process, which is typically the console for an interactive `brew` process.

ilovezfs:
Why did the fgrep need to be changed if this doesn't affect stdout?

sjackman:
```
expected block to not output to stderr, but output "Error: No such file or directory - /usr/bin/fgrep\nError: No such file or directory - /usr/bin/fgrep\nError: No such file or directory - /usr/bin/fgrep\n"
```
https://github.com/Homebrew/brew/pull/3099#issuecomment-325436813

It caused a `brew tests` failure that checked that the output of `$stderr` was empty when installing a bottle with a broken symlink.

Library/Homebrew/utils/popen.rb
```diff
@@ -13,7 +13,7 @@ def self.popen(args, mode)
         return pipe.read unless block_given?
         yield pipe
       else
-        $stderr.reopen("/dev/null", "w")
+        $stderr.reopen("/dev/null", "w") unless ARGV.debug? || ARGV.verbose?
```
MikeMcQuaid: Can you make this an `if` instead; `unless` with multiple conditions are hard to parse. After that should be 🆗

sjackman: Done. What do you think of adding…
```ruby
$stderr.reopen("/dev/null", "w") if !ARGV.debug? && !ARGV.verbose? && !ENV["HOMEBREW_DEVELOPER"]?
```

ilovezfs: :-1: to that.

MikeMcQuaid: Debug or verbose seems sufficient, thanks.

sjackman: That's fine.

ilovezfs:
>codecov/patch — 0% of diff hit (target 50.49%)

@sjackman do you mind trying to raise that percent?

sjackman:
Both lines of this diff are hit by `test/dev-cmd/bottle_spec.rb:20`. I know because that test failed without adding the `-s` option to `grep`.
https://travis-ci.org/Homebrew/brew/jobs/269276988#L1745
https://github.com/Homebrew/brew/pull/3099#issuecomment-325436813

Doesn't that mean that 100% of the diff is covered? Or have I misunderstood?

ilovezfs:
>I know because that test failed without adding the -s option to grep.

was that before or after you scoped it to debug and verbose?

sjackman:
This line is definitely being run.
```ruby
Utils.popen_read("/usr/bin/fgrep", "-s", recursive_fgrep_args, string, to_s) do |io|
```
https://github.com/Homebrew/brew/pull/3099/files#diff-26acfe09641ad4e7cf4f22179625981eR99

sjackman:
```ruby
$stderr.reopen("/dev/null", "w") if !ARGV.debug? && !ARGV.verbose?
```
https://github.com/Homebrew/brew/pull/3099/files#diff-83e8c13ed518191365f4edaa7339293dR16
This line is run too. I believe `ARGV.debug?` and `ARGV.verbose?` are both false when Jenkins runs, so only the one code path is being tested, where `!ARGV.debug? && !ARGV.verbose?` is false and `$stderr` is closed.

MikeMcQuaid:
@sjackman It's unclear to me: are you saying the CI fails without that change being made?

sjackman:
CI passes. The `fgrep -s` change is required if `brew tests` is run with either `ARGV.debug?` or `ARGV.verbose?` enabled, but that doesn't seem to be the case.

MikeMcQuaid:
> The fgrep -s change is required if brew tests is run with either ARGV.debug? or ARGV.verbose? enabled, but that doesn't seem to be the case for CI brew test-bot.

:cool:, you can remove that change then, thanks.

sjackman:
Done!

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #3101, opened 2017-08-28
mach: Avoid reopening the file for relocation

woodruffw:
This commit allows the relocation code to perform install name
and dylib ID changes without reopening the file separately.

I still need to delete the old `MachO::Tools`-calling code and do some other
cleanup, so this is WIP.

Potential fix for the failing tests in #3079.


woodruffw:
cc @mansimarkaur

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -24,7 +24,7 @@ def relocate_dynamic_linkage(relocation)
       file.ensure_writable do
         if file.dylib?
           id = dylib_id_for(file).sub(relocation.old_prefix, relocation.new_prefix)
-          change_dylib_id(id, file)
+          file.change_dylib_id(id)
```
mansimarkaur: The `change_dylib_id` still accepts 2 arguments. Shouldn't this be  `file.change_dylib_id(id, file)`?

woodruffw: Whoops, that was a mistake. It should only take 1, I'll fix that up now.

ayan-das1234:
@mansimarkaur What in the hell did you do i am trying to clone into Homebrew and I am stuck with this error: HEAD is now at 2a75c6c Merge pull request #2706 from mansimarkaur/increase_test_cov

ilovezfs:
@ayan-das1234 please follow our [code of conduct](https://github.com/Homebrew/brew/blob/master/CODEOFCONDUCT.md) when interacting with other people here.

MikeMcQuaid:
@woodruffw 🚢 when you're happy.

PR #3103, opened 2017-08-29
linkage_checker: add a check for extraneous dependencies

maxim-belkin:

-----

Detect extraneous dependencies in formulae. Might be useful for new formulae.

- Have not written tests for my changes. Suggestions what tests to write are welcome
- `brew tests` fail for some other tests

### DIFF
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -5,7 +5,7 @@ require "formula"
 class LinkageChecker
   attr_reader :keg, :formula
   attr_reader :brewed_dylibs, :system_dylibs, :broken_dylibs, :variable_dylibs
-  attr_reader :undeclared_deps, :extraneous_deps, :reverse_links
+  attr_reader :undeclared_deps, :unnecessary_deps, :reverse_links
 
   def initialize(keg, formula = nil)
     @keg = keg
@@ -16,7 +16,7 @@ class LinkageChecker
     @variable_dylibs = Set.new
     @undeclared_deps = []
     @reverse_links = Hash.new { |h, k| h[k] = Set.new }
-    @extraneous_deps = []
+    @unnecessary_deps = []
     check_dylibs
   end
 
@@ -52,7 +52,7 @@ class LinkageChecker
       end
     end
 
-    @undeclared_deps, @extraneous_deps = check_undeclared_deps if formula
+    @undeclared_deps, @unnecessary_deps = check_undeclared_deps if formula
   end
 
   def check_undeclared_deps
@@ -78,11 +78,12 @@ class LinkageChecker
         a <=> b
       end
     end
-    extraneous_deps = declared_dep_names.reject do |full_name|
+    unnecessary_deps = declared_dep_names.reject do |full_name|
       name = full_name.split("/").last
-      Formula[name].bin.directory? || @brewed_dylibs.keys.map { |x| x.split("/").last }.include?(name)
+      next true if Formula[name].bin.directory?
+      @brewed_dylibs.keys.map { |x| x.split("/").last }.include?(name)
     end
-    [undeclared_deps, extraneous_deps]
+    [undeclared_deps, unnecessary_deps]
   end
 
   def display_normal_output
@@ -91,7 +92,7 @@ class LinkageChecker
     display_items "Variable-referenced libraries", @variable_dylibs
     display_items "Missing libraries", @broken_dylibs
     display_items "Possible undeclared dependencies", @undeclared_deps
-    display_items "Possible extraneous dependencies", @extraneous_deps
+    display_items "Possible unnecessary dependencies", @unnecessary_deps
   end
 
   def display_reverse_output
@@ -109,6 +110,7 @@ class LinkageChecker
 
   def display_test_output
     display_items "Missing libraries", @broken_dylibs
+    display_items "Possible unnecessary dependencies", @unnecessary_deps
     puts "No broken dylib links" if @broken_dylibs.empty?
   end
 
@@ -120,8 +122,8 @@ class LinkageChecker
     !@undeclared_deps.empty?
   end
 
-  def extraneous_deps?
-    !@extraneous_deps.empty?
+  def unnecessary_deps?
+    !@unnecessary_deps.empty?
   end
 
   private
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -80,7 +80,7 @@ class LinkageChecker
     end
     extraneous_deps = declared_dep_names.reject do |full_name|
       name = full_name.split("/").last
-      @brewed_dylibs.keys.map { |x| x.split("/").last }.include? name
+      Formula[name].bin.directory? || @brewed_dylibs.keys.map { |x| x.split("/").last }.include?(name)
     end
     [undeclared_deps, extraneous_deps]
   end
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -79,10 +79,10 @@ class LinkageChecker
       end
     end
     extraneous_deps = declared_dep_names.reject do |full_name|
-        name = full_name.split("/").last
-        @brewed_dylibs.keys.map{ |x| x.split("/").last }.include? name
+      name = full_name.split("/").last
+      @brewed_dylibs.keys.map { |x| x.split("/").last }.include? name
     end
-    return undeclared_deps, extraneous_deps
+    [undeclared_deps, extraneous_deps]
   end
 
   def display_normal_output
--- a/Library/Homebrew/os/mac/linkage_checker.rb
+++ b/Library/Homebrew/os/mac/linkage_checker.rb
@@ -5,7 +5,7 @@ require "formula"
 class LinkageChecker
   attr_reader :keg, :formula
   attr_reader :brewed_dylibs, :system_dylibs, :broken_dylibs, :variable_dylibs
-  attr_reader :undeclared_deps, :reverse_links
+  attr_reader :undeclared_deps, :extraneous_deps, :reverse_links
 
   def initialize(keg, formula = nil)
     @keg = keg
@@ -16,6 +16,7 @@ class LinkageChecker
     @variable_dylibs = Set.new
     @undeclared_deps = []
     @reverse_links = Hash.new { |h, k| h[k] = Set.new }
+    @extraneous_deps = []
     check_dylibs
   end
 
@@ -51,7 +52,7 @@ class LinkageChecker
       end
     end
 
-    @undeclared_deps = check_undeclared_deps if formula
+    @undeclared_deps, @extraneous_deps = check_undeclared_deps if formula
   end
 
   def check_undeclared_deps
@@ -77,6 +78,11 @@ class LinkageChecker
         a <=> b
       end
     end
+    extraneous_deps = declared_dep_names.reject do |full_name|
+        name = full_name.split("/").last
+        @brewed_dylibs.keys.map{ |x| x.split("/").last }.include? name
+    end
+    return undeclared_deps, extraneous_deps
   end
 
   def display_normal_output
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'm not sure I'm reading this right but is an extraneous dependency any where there's no linkage? If so, this will need to be an optional flag as there will be a _lot_ of false positives otherwise (i.e. anything that calls an executable from a dependency rather than linking against it, most scripting language dependencies, etc.)

sjackman:
A lot of false positives could be eliminated by ignoring any keg that has a `bin` directory. Many packages whose primary purpose is to provide a library have `include` and `lib` directories, but no `bin` directory.

maxim-belkin:
I agree: false positives is a possibility.
```
brew linkage ffmpeg
```
reports `xvid` as a possible extraneous dependency

maxim-belkin:
Those interested in testing this, try the following "one-liner" to check installed formulae:
```bash
for formula in $(brew list); do echo "* Formula: $formula"; brew linkage $formula | sed -n '/^Possible/,/^[^ ]/p;'; done
```

sjackman:
@ilovezfs identified that roughly half of `depends_on "boost"` dependencies in Homebrew/science use only the headers, and have no run-time linkage to `boost`, and so should in fact be `depends_on "boost" => :build`. This PR would help identify those cases. I'm having trouble finding the archived conversation—too many repos and slack channels.

scpeters:
@sjackman there could still be a need for `boost` to be installed for libraries that expose boost symbols in their own headers. Then anything that wants to build against the library would need the boost headers. It's kinda like the difference between `lib*` and `lib*-dev` packages on debian.

scpeters:
I just took a quick look at some of the packages in `homebrew/science` and found two that use the boost headers and install a library and header files ([cantera](https://github.com/Homebrew/homebrew-science/blob/master/cantera.rb#L23) and [xylib](https://github.com/Homebrew/homebrew-science/blob/master/xylib.rb#L15)). Despite having `#include <boost/` statements in the installed header files, both of these formulae declare `:build` dependencies on `boost`, possibly because they also install executables and want to reduce the footprint for those users. If you're building from source against a library, they might be knowledgeable to install boost on their own.

Maybe these types of header-only dependencies that the linkage test can't see could be marked with `:run` like the build dependencies in `audit`?

https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L499-L507

MikeMcQuaid:
> I agree: false positives are a possibility.

Given we're running this in CI for everything in Homebrew/homebrew-core: we really don't want false positives here.

> @ilovezfs identified that roughly half of depends_on "boost" dependencies in Homebrew/science use only the headers, and have no run-time linkage to boost, and so should in fact be depends_on "boost" => :build. This PR would help identify those cases.

I agree. Having this as an optional flag which is always run for new formulae in `brew test-bot`seems like a good idea.

maxim-belkin:
Please suggest test-cases or directions how to improve the code

MikeMcQuaid:
@maxim-belkin Use `ARGV.include?("--something")` and only enable the current functionality if that is passed.

sjackman:
@MikeMcQuaid The output of `Possible undeclared dependencies` is informative only. It doesn't affect the exit status of either `brew test` or `brew test --missing`. For example `brew linkage awk` reports `Possible undeclared dependencies: gmp`. Would that be an option for this feature as well? Enabled by default, but not affecting exit status. To make this feature useful, it's still important to eliminate as many false positives as possible.

MikeMcQuaid:
> It doesn't affect the exit status of either brew test or brew test --missing.

It doesn't affect the exit status or output of https://github.com/Homebrew/homebrew-test-bot/blob/80333911e7130b9542a6fe2b94c1311b52cef319/cmd/brew-test-bot.rb#L664 ? That's what we need to avoid.

sjackman:
Sorry, I wrote too quickly and didn't engage brain. I had meant to say…
`Possible undeclared dependencies` doesn't affect the output status of `brew linkage` or `brew linkage --check`.
If the same were true for this PR `Possible unnecessary dependencies` would it be okay without an option to enable it?

ilovezfs:
>I agree. Having this as an optional flag which is always run for new formulae in brew test-bot seems like a good idea.

Things that are known to give a lot of false positives aren't great even for --new-formula. We keep getting contributors asking, "Will my formula still qualify if those audits are failing?" and it's getting pretty old explaining that yes CI is bright red but no that doesn't matter because it will go away next time after it's merged.

sjackman:
This test will not affect the exit status of `brew linkage --test`, `brew audit --new-formula`, nor of CI.

MikeMcQuaid:
@sjackman in that case: how does anyone notice that output and make changes as a result?

scpeters:
I'm guessing it prints console messages but doesn't affect the exit status, which is what CI cares about.

sjackman:
Same as with the existing `Possible undeclared dependencies` message. It's only an informational message. If you ran `brew linkage` manually you'd see the message, or if you looked in the CI log, you'd see the message.

maxim-belkin:
Just an FYI: I've prepared a version with `--extraneous` so... just let me guys know what you decide. Personally, I think it is similar to undeclared dependencies, so I'd enable it by default.
As I mentioned in the OP, this could be used for new formulae: people could be asked to provide output of `brew linkage` when submitting a PR.

sjackman:
> @sjackman in that case: how does anyone notice that output and make changes as a result?

Ah, I had forgotten that `test-bot` prints only the output of `brew linkage --test` and not `brew linkage`. In that case this PR does not affect CI at all. I would suggest modifying test-bot to run `brew linkage`, which always succeeds, in addition to `brew linkage --test`, so that the output of `Possible undeclared dependencies` and the proposed `Possible unnecessary dependencies` are in the build log.

```diff
+      test "brew", "linkage", dependent.name
       test "brew", "linkage", "--test", dependent.name
```
https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L664

sjackman:
Or `brew linkage --test` could be modified to keep its existing exit status behaviour, but print the more verbose message of `brew linkage`.

maxim-belkin:
> Or brew linkage --test could be modified to keep its existing exit status behaviour, but print the more verbose message of brew linkage.

that's a possibility.

sjackman:
The current output of `brew linkage` and `brew linkage --test` is…
```
❯❯❯ brew linkage gawk
System libraries:
  /usr/lib/libSystem.B.dylib
Homebrew libraries:
  /Users/sjackman/.homebrew/opt/gmp/lib/libgmp.10.dylib (gmp)
  /Users/sjackman/.homebrew/opt/mpfr/lib/libmpfr.4.dylib (mpfr)
  /Users/sjackman/.homebrew/opt/readline/lib/libreadline.7.dylib (readline)
Possible undeclared dependencies:
  gmp
❯❯❯ brew linkage --test gawk
No broken dylib links
```

MikeMcQuaid:
> Or brew linkage --test could be modified to keep its existing exit status behaviour, but print the more verbose message of brew linkage.

This seems sensible. That said: how are people going to notice this for formulae in CI or is this intended to just be manually run by maintainers? I'm trying to understand the use case.

sjackman:
Of 47 formulae that depend on `boost` in Homebrew/science, @ilovezfs identified that 20 of those 47 formulae have no library linkage to `boost` (that is, they use only the headers), so should in fact be `depends_on "boost" => :build` rather than `depends_on "boost"`. This PR would allow a developer (maintainer or contributor) to run `brew linkage foo`, see the message `Possible unnecessary dependency: boost`, and change the dependency to a `:build` dependency. I would use this command when creating a new formula, or troubleshooting existing formulae.
https://github.com/Homebrew/homebrew-science/issues/6192#issuecomment-324509708

maxim-belkin:
1. `extraneous` or `unnecessary`?
2. what should I do?
    1. leave the PR "as is" 😊 
    2. move `extraneous/unnecessary` functionality under the `--test`
        1. move `undeclared` dependencies under `--test`

MikeMcQuaid:
> This PR would allow a developer (maintainer or contributor) to run brew linkage foo, see the message Possible unnecessary dependency: boost, and change the dependency to a :build dependency.

That seems like a good workflow but I guess I'm just wondering if this is reliant on individual developers to manually do this: is it going to be useful? I see no harm in this PR as-is being merged, to be clear, I'm just trying to figure out if there's some better workflow.

MikeMcQuaid:
> extraneous or unnecessary?

unnecessary

> what should I do?

As long as the exit code isn't modified I'm happy as-is. It does seem like it would be useful to have this under `--test` so `test-bot` runs it.

sjackman:
@maxim-belkin I'd suggest `Possible build or unnecessary dependencies`

sjackman:
> I'm just trying to figure out if there's some better workflow.

If it were possible to eliminate all the false positives, then it could affect the `brew linkage --test` exit status. It may be possible, but it would require significant work, and may be impossible. If it got to the point of being mostly correctly (but not entirely), perhaps it could be enabled as a check for new formula only.

We could add a tickbox to `.github/PULL_REQUEST_TEMPLATE.md` that asks the user to run `brew linkage foo` and consider whether any of the `Possible build or unnecessary dependencies` may in fact be a `:build` dependency or unnecessary.

Library/Homebrew/os/mac/linkage_checker.rb
```diff
@@ -77,6 +78,11 @@ def check_undeclared_deps
         a <=> b
       end
     end
+    extraneous_deps = declared_dep_names.reject do |full_name|
+      name = full_name.split("/").last
+      Formula[name].bin.directory? || @brewed_dylibs.keys.map { |x| x.split("/").last }.include?(name)
```
MikeMcQuaid: `next true if Formula[name].bin.directory?` and put the `map` on a new line so it is a little shorter.

MikeMcQuaid:
> unnecessary

Let's go for this and then we can merge as-is if it's useful to you like that.

sjackman:
I'd suggest also renaming the code `extraneous_deps` to `unnecessary_deps` in that case, to match.

maxim-belkin:
Sure, will do once I get to it. Should be today (evening).

Library/Homebrew/os/mac/linkage_checker.rb
maxim-belkin: perhaps `next true unless ...?`

maxim-belkin: nope, `next true if...` :)

maxim-belkin:
In my [second implementation](https://github.com/maxim-belkin/brew-macOS/blob/cb565e8e0155a3c023fdce1713f18faadfd4a2de/Library/Homebrew/os/mac/linkage_checker.rb#L60) of this PR (which I'm abandoning), I had to factor out `declared_dep_names` method from `check_undeclared_deps`

I can do that here too, if it seems reasonable/necessary.

maxim-belkin:
I wonder if we should report possible unnecessary dependencies (PUDs) with `bin` directories as "tier-2" PUDs

maxim-belkin:
I think the `linkage_checker` can be improved further. Code that detects `undeclared/unnecessary` dependencies is executed under `--test`, when, in fact, it should not.

maxim-belkin:
can someone re-iniate the build? I think there was an intermittent issue

ilovezfs:
Those errors are currently affecting most things :see_no_evil:

maxim-belkin:
💩

maxim-belkin:
Could someone have a look at the following lines of this PR more closely: line [8](https://github.com/maxim-belkin/brew-macOS/blob/756a3ad13873011400b5d6e86fce7271eb46734d/Library/Homebrew/os/mac/linkage_checker.rb#L8) and [125 through 127](https://github.com/maxim-belkin/brew-macOS/blob/756a3ad13873011400b5d6e86fce7271eb46734d/Library/Homebrew/os/mac/linkage_checker.rb#L125-L127).

ilovezfs:
@maxim-belkin you need to rebase the PR or it will keep failing.

maxim-belkin:
it did not fail - it did not run one of the tests. but I'll rebase the PR.

maxim-belkin:
any suggestion how to improve test fix "codecov/patch"?

MikeMcQuaid:
Looks good to me, nice work @maxim-belkin!

maxim-belkin:
Glad to contribute!

PR #3105, opened 2017-08-29
pour_bottle?: Pour local bottles without sha256

sjackman:

-----

Pouring a local bottle for a formula without a bottle sha256
in the formula for that OS would unexpectedly install from
source instead.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -91,7 +91,6 @@ class FormulaInstaller
     return false if ARGV.cc
     return false unless options.empty?
     return false if formula.bottle_disabled?
-    return true  if formula.local_bottle_path
     unless formula.pour_bottle?
       if install_bottle_options[:warn] && formula.pour_bottle_check_unsatisfied_reason
         opoo <<-EOS.undent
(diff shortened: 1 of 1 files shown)


Library/Homebrew/formula_installer.rb
```diff
@@ -85,7 +85,7 @@ def pour_bottle?(install_bottle_options = { warn: false })
     return false if @pour_failed
 
     bottle = formula.bottle
-    return false unless bottle
+    return false unless bottle || formula.local_bottle_path
```
MikeMcQuaid: `return true  if formula.local_bottle_path` is on line 94 already so it feels weird to duplicate it here. Can this logic be moved around instead so it's only checked once?

sjackman: I've removed `return true if formula.local_bottle_path`. Pouring a local bottle should still check that that bottle is compatible. `--force-bottle` may be used to pour a local bottle that is incompatible, as it is for remote bottles.

MikeMcQuaid: Ok, cool. Can you make this an `if` instead; `unless` with multiple conditions are hard to parse. After that should be 👍 to 🚢

sjackman: Done. Thanks for the feedback, Mike.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
My pleasure. Thanks for the feedback, and thanks for merging, Mike.

PR #3106, opened 2017-08-30
add -n flag for curl ( check .netrc for basic auth )

fsuhrau:
some cdm provider requires basic authentication or token authentication.
it is easy to provide them via ~/.netrc
with extending the curl download flags by -n, curl will try to automatically find the best fit for authentication.

### DIFF
--- a/Library/Homebrew/utils/curl.rb
+++ b/Library/Homebrew/utils/curl.rb
@@ -38,7 +38,7 @@ def curl(*args)
 end
 
 def curl_download(*args, to: nil, continue_at: "-", **options)
-  curl("--location", "--remote-time", "--continue-at", continue_at.to_s, "--output", to, *args, **options)
+  curl("-n", "--location", "--remote-time", "--continue-at", continue_at.to_s, "--output", to, *args, **options)
 rescue ErrorDuringExecution
   # `curl` error 33: HTTP server doesn't seem to support byte ranges. Cannot resume.
   if $CHILD_STATUS.exitstatus == 33 && continue_at == "-"


MikeMcQuaid:
> some cdm provider

What's a CDM provider?

fsuhrau:
sorry typo CDN provider in our case it is JFrog Artifactory

MikeMcQuaid:
Artifactory should be modified to support Homebrew and not the other way around, sorry.

fsuhrau:
sorry i didn't get your point.
homebrew has already custom strategies for amazon s3 authentication or github private repo auth.
netrc is just a way provided by unix/linux to achieve also authentication.
wget, for example, is looking in netrc for credentials by default also curl has the possibility but sadly curl requires an argument for it (which i would like to add)
in my opionion homebrew should also provide the os way for authentication.

what maybe makes sense to change is the flag to '--netrc-optional' wich is not that strict as '-n' so in case somebody uses basic auth in urls of formulas directly it will not block him.

Just a hint for other people who have the same problem it is also achieveable by creating a custom download strategy.

```
# uses netrc
class CurlNetrcDownloadStrategy < CurlDownloadStrategy
    def _fetch
        url = @url
        curl_download url, "--netrc-optional", to: temporary_path
    end
end

# uses artifactory token authentication
class ArtifactoryTokenDownloadStrategy < CurlDownloadStrategy
    def _fetch
        if ENV['ARTIFACTORY_TOKEN']
            ohai "Using ARTIFACTORY_TOKEN"
            curl_download url, "-H", "X-JFrog-Art-Api: #{ENV["ARTIFACTORY_TOKEN"]}", to: temporary_path
        end
    end
end

# example usage
class Foo < Formula
  desc ""
  homepage ""
  url "https://example.com/foo-0.1.tar.gz", :using => CurlNetrcDownloadStrategy
  sha256 "85cc828a96735bdafcf29eb6291ca91bac846579bcef7308536e0c875d6c81d7"

  # depends_on "cmake" => :build

  def install
    # ENV.deparallelize
    system "./configure", "--disable-debug",
                          "--disable-dependency-tracking",
                          "--disable-silent-rules",
                          "--prefix=#{prefix}"
    # system "cmake", ".", *std_cmake_args
    system "make", "install"
  end

  test do
    system "false"
  end
end
```

MikeMcQuaid:
> netrc is just a way provided by unix/linux to achieve also authentication.

The problem is if someone has a misconfigured `netrc` this will now break installation of all Homebrew formulae for them and we'll need to help them debug that.

> Just a hint for other people who have the same problem it is also achieveable by creating a custom download strategy.

Doing this in specific formulae seems like a better idea and if it's used in multiple, open-source formulae we could accept a new download strategy for this. CC @ilovezfs for thoughts here.

PR #3107, opened 2017-08-30
RFC: Parallel downloads.

reitermarkus:
I've been playing with this for a while now, but I'm still not sure about how to best handle the output, so I'd like to hear some feedback on the output “window”.

I have currently implemented a “window” which shows the URL and the download percentage, along with a progress bar (which is “behind” the URL using inverted colors).

bildschirmfoto 2017-08-30 um 20 39 16

So please, @Homebrew/maintainers, pull this and try it using `brew download`.

This is a very rough draft and far from being finished, so this should not be a code review but rather a discussion about the UI/UX.

### DIFF
--- a/Library/Homebrew/.rubocop.yml
+++ b/Library/Homebrew/.rubocop.yml
@@ -29,6 +29,9 @@ Lint/ParenthesesAsGroupedExpression:
 Metrics/ParameterLists:
   CountKeywordArgs: false
 
+Style/AsciiComments:
+  Enabled: false
+
 Style/BlockDelimiters:
   Exclude:
     - '**/*_spec.rb'
--- a/Library/Homebrew/cmd/download.rb
+++ b/Library/Homebrew/cmd/download.rb
@@ -23,14 +23,15 @@ class ParallelDownloader
   end
 
   def self.update_display(message, progress)
-    progress_str = "%0.1f %%" % progress
-    len = width - progress.to_s.length - 4
+    progress_str = " %0.1f %%" % progress
+
+    len = width - progress_str.to_s.length
 
     str = message.to_s[0, len].to_s.ljust(len)
 
     split_at = (len / 100.0 * progress).to_i
 
-    "\e[7m" << str[0, split_at].to_s << "\e[0m" << str[split_at..-1] << "  #{progress_str}\n"
+    "\e[7m" << str[0, split_at].to_s << "\e[0m" << str[split_at..-1] << progress_str << "\n"
   end
 
   def update
@@ -44,8 +45,7 @@ class ParallelDownloader
         lines = downloads.map { |dl| self.class.update_display(dl.uri, dl.progress) }.join
 
         if $stdout.tty?
-          print "\e[#{downloads.count}A"
-          print lines
+          print "\e[" << downloads.count.to_s << "A" << lines
         else
           print lines if downloads.all?(&:ended?)
         end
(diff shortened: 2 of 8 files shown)


MikeMcQuaid:
Does this have anything that will prevent parallel downloads from the same host? That'd be a blocker from my perspective.

FYI for an external command the best way to get people to test this and provide feedback would be an external command in a tap. For this to be included here we'd probably want it to be hooked into all the DownloadStrategy stuff (which may provide another opportunity to get Cask sharing more of our code, there).

reitermarkus:
> For this to be included here …

I have to clarify. This command shouldn't be included here, it only exists to demonstrate the behaviour of parallel downloads, which e.g. `brew fetch` and `brew install` would make use of when this is “hooked into all the DownloadStrategy stuff”.

> Does this have anything that will prevent parallel downloads from the same host?

I'm not sure what this means exactly.

alyssais:
Downloading multiple files at once from the same server doesn't buy us anything in terms of performance, it just means longer waits for any one thing to download.

reitermarkus:
> Downloading multiple files at once from the same server doesn't buy us anything in terms of performance

I tried with 8 parallel downloads from `bintray.com` vs. sequential downloads. Still 40 % faster. Then again, no one has commented yet on the behaviour. Do we want to download everything first? If so, downloading from the same host is not an issue. Or do we want to start installation as soon as one of the downloads finishes?

MikeMcQuaid:
> I tried with 8 parallel downloads from bintray.com vs. sequential downloads. Still 40-50 % faster.

Regardless of the speed it's kinda inconsiderate to the host to do so. In Bintray's case it's a CDN so I doubt they care but many hosts would.

> Then again, no one has commented yet on the behaviour.

I haven't tested it as I don't understand how to test it and don't want to `brew pull` this currently (but an external tap would be easier).

> Do we want to download everything first?

I'm not sure. If I pass e.g. `brew install foo bar foo2 bar2 foo3 bar3` then the advantage is once the downloads are done they won't fail if my internet drops once connections begin. On the flip side, it's longer until e.g. foo alone will be installed. I'm probably leaning a bit towards "yes we should download everything first" partly for the not needing to download during and partly for the UI.

alyssais:
> Or do we want to start installation as soon as one of the downloads finishes?

I think this would be technically preferable. No point letting the CPU idle while we're waiting on the network, and vice versa. Not sure how the UI would work, though.

reitermarkus:
> Not sure how the UI would work, though.

Yep, that's the open question.

ilovezfs:
>No point letting the CPU idle while we're waiting on the network, and vice versa.

There is a point, which is that the make-it-do-everything-in-parallel approach I think will be more confusing and increase support our burden. Also, waiting for everything to download before starting any installs seems really annoying in exchange for a minor benefit to those who are using brew with intermittent network access. So personally I prefer the current behavior to anything be proposed here.

MikeMcQuaid:
I think I'm 👍 (and based on Slack conversations with @ilovezfs he also seems to be) on implementing this now as a `brew fetch --parallel` option (possibly also enabled by default with an environment variable). If that ends up being a solution to a problem and not buggy we could look at promoting it to the `brew fetch` default. Only after we've evaluated all that would it be something we'd think about for `brew install`'s flow, I think.

As mentioned before: personally I don't find it useful to test a command like this without looking at the code in a way that could be integrated so I'd suggest this PR moves instead towards the `brew fetch --parallel` approach.

MikeMcQuaid:
Closing this as it can still be `brew pull`d, discussion has stopped here in favour of a future PR adding the functionality to `brew fetch`.

PR #3109, opened 2017-08-31
Java: deprecate ENV.java_cache in favour of setting it by default.

DomT4:
-----

https://github.com/Homebrew/homebrew-core/issues/17376 made me realise it's quite confusing we still permit this to be optional given the sandbox is now mandatory unless you pass an explicit flag.

`brew tests` still fails with the unrelated:
```
Failures:

  1) Homebrew #search_taps
     Failure/Error:
       expect(described_class.search_taps("some-formula"))
         .to match(["homebrew/foo/some-formula"])

       expected [] to match ["homebrew/foo/some-formula"]
     # ./test/cmd/search_remote_tap_spec.rb:18:in `block (2 levels) in <top (required)>'

Finished in 1 minute 12.21 seconds (files took 3.43 seconds to load)
311 examples, 1 failure
```

### DIFF
--- a/Library/Homebrew/compat/ENV/shared.rb
+++ b/Library/Homebrew/compat/ENV/shared.rb
@@ -3,4 +3,8 @@ module SharedEnvExtension
     odeprecated "ENV.j1", "ENV.deparallelize"
     deparallelize
   end
+
+  def java_cache
+    # odeprecated "ENV.java_cache"
+  end
 end
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -869,6 +869,10 @@ class FormulaAuditor
       problem "Use \"depends_on :x11\" instead of \"ENV.x11\""
     end
 
+    if line.include?("ENV.java_cache")
+      problem "In-formula ENV.java_cache usage has been deprecated & should be removed."
+    end
+
     # Avoid hard-coding compilers
     if line =~ %r{(system|ENV\[.+\]\s?=)\s?['"](/usr/bin/)?(gcc|llvm-gcc|clang)['" ]}
       problem "Use \"\#{ENV.cc}\" instead of hard-coding \"#{Regexp.last_match(3)}\""
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -260,10 +260,6 @@ module SharedEnvExtension
     set_cpu_flags(flags)
   end
 
-  def java_cache
-    append "_JAVA_OPTIONS", "-Duser.home=#{HOMEBREW_CACHE}/java_cache"
-  end
-
   # ld64 is a newer linker provided for Xcode 2.5
   # @private
   def ld64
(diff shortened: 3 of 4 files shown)


DomT4:
If this is merged, would likely need a new tag cutting afterwards, alongside a PR filed against core to remove all existing `ENV.java_cache` usage.

Library/Homebrew/compat/ENV/shared.rb
```diff
@@ -3,4 +3,8 @@ def j1
     odeprecated "ENV.j1", "ENV.deparallelize"
     deparallelize
   end
+
+  def java_cache
+    odeprecated "ENV.java_cache"
```
MikeMcQuaid: I'd hold off adding the `odeprecated` in favour of an `audit` for now. Comment it out and we can reenable it when people have had the chance to address the audit.

DomT4: 👍 Will tweak.

DomT4:
Tweaks pushed.

DomT4:
Also CC @apjanke if you're around. We've discussed this a bit in the past so you might have thoughts/feelings.

MikeMcQuaid:
Thanks again @DomT4!

DomT4:
Thanks Mike ❤️

ilovezfs:
In the tests should we be setting  "-Duser.home=#{testpath}" rather than "-Duser.home=#{HOMEBREW_CACHE}/java_cache"? There's some inconsistency in core over which is preferred.

DomT4:
`ENV.java_cache` was supposed to be used consistently IIRC. It wasn't, essentially, because very few people wanted to repeatedly run `test do` blocks that fetched a large amount of remote Java-related data & consequently the pressure for that to be something that was forced into consistency was low.

That discussion is still one that could be had but generally it seems easier to use the same location.

On the point discussed in the `core` PR things setting `ENV["_JAVA_OPTIONS"] =` unconditionally rather than prepending to that variable should now be more or less verboten. I can always throw together an audit rule on that if necessary/desired.

ilovezfs:
It sounds like we should just remove the instances of `-Duser.home=#{testpath}` unless for test purity reasons we don't want the tests using the cache.

DomT4:
PR incoming shortly. Just doing a few sanity checks locally. I did a fair bit of building stuff before this PR was filed but evidently I wasn't broad enough in remembering we hadn't used `ENV.java_cache` that consistently.

DomT4:
(Bad Dom 🙈)

ilovezfs:
Basically I either think we should automatically set `-Duser.home=#{testpath}` in the test blocks, or automatically set `-Duser.home=#{HOMEBREW_CACHE}/java_cache` in the test blocks. But either that will mean that the explicit `-Duser.home=#{testpath}` in the test blocks can be removed.

DomT4:
https://github.com/Homebrew/homebrew-core/pull/17591

apjanke:
Sorry for the slow response; it is seasonal crunch time at my work right now.

Just for the record, I'm in favor of this and agree with DomT4's reasoning.

DomT4:
No worries at all on the response time. Just wanted to tag you in since it was discussed previously & I didn't know whether it was still something that was on your radar. Thanks for the response! 🍻

PR #3111, opened 2017-08-31
Don't fail to migrate repo because of empty dir

alyssais:

-----

I just ran into an issue where a colleague's Homebrew couldn't update. It seems to have been because it failed once just after the new repo directory was created. Since there was nothing in this directory, there really isn't any reason for us to fail here, so to avoid this problem in the future, try to `rmdir` the directory before failing because it already exists.


MikeMcQuaid:
Seems reasonable, thanks @alyssais!

PR #3112, opened 2017-09-01
search_remote_tap spec: fix test offline

mistydemeo:

-----

As pointed out by @jcount, the method being tested always [returns `[]` immediately](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/search.rb#L106) if `HOMEBREW_NO_GITHUB_API` is set even though we stub the JSON response. As a result, we get an `expected [] to match ["homebrew/foo/some-formula"]` failure when `--online` isn't passed. Unsetting `HOMEBREW_NO_GITHUB_API` for this one test should be safe.


reitermarkus:
Thanks!

PR #3114, opened 2017-09-01
Shell escape printed commands when installing a formula

alyssais:

-----

This means commands can be copy/pasted into a terminal, which is very useful, for example, when debugging formulae installs with `brew install --interactive`.

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1,3 +1,4 @@
+require "shellwords"
 require "formula_support"
 require "lock_file"
 require "formula_pin"
@@ -1732,7 +1733,7 @@ class Formula
         pretty_args[i] = "import setuptools..."
       end
     end
-    ohai "#{cmd} #{pretty_args * " "}".strip
+    ohai [cmd, *pretty_args].shelljoin
 
     @exec_count ||= 0
     @exec_count += 1
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1867,8 +1867,8 @@ class Formula
       ENV.refurbish_args if setup_py_in_args || setuptools_shim_in_args
     end
 
-    $stdout.reopen(out)
-    $stderr.reopen(out)
+    $stdout = out
+    $stderr = out
     out.close
     args.collect!(&:to_s)
     begin
--- a/Library/Homebrew/test/formula_spec.rb
+++ b/Library/Homebrew/test/formula_spec.rb
@@ -763,6 +763,13 @@ describe Formula do
     expect(h["versions"]["bottle"]).to be_truthy
   end
 
+  describe "#system" do
+    it "escapes commands" do
+      f = formula { url "foo-1.0" }
+      expect { f.system("echo", "-n", "") }.to output("==> echo -n ''\n").to_stdout
+    end
+  end
+
   describe "#eligible_kegs_for_cleanup" do
     it "returns Kegs eligible for cleanup" do
       f1 = Class.new(Testball) do


Library/Homebrew/formula.rb
```diff
@@ -1867,8 +1868,8 @@ def exec_cmd(cmd, args, out, logfn)
       ENV.refurbish_args if setup_py_in_args || setuptools_shim_in_args
     end
 
-    $stdout.reopen(out)
-    $stderr.reopen(out)
+    $stdout = out
+    $stderr = out
     out.close
```
alyssais: Feels weird to be closing this since we're now still using that instance, but it does seem to work? Any `IO` experts able to explain this to me?

reitermarkus: According to my minimal example here, it really shouldn't work. If I uncomment `w.close`, the fork fails with an `IOError`.

```ruby
r, w = IO.pipe

pid = fork do
  r.close
  $stdout = w
  $stderr = w
  begin
    w.close
    $stdout.puts "STDOUT"
    $stderr.puts "STDERR"
  rescue => e
    STDOUT.puts e.class
    exit! 1
  end
  exit! 0
end

w.close

Process.wait(pid)

puts r.read

r.close
```

reitermarkus: Ok, it seems like `$stdout` and `$stderr` aren't even used, because the `exec` call below will just use `STDOUT` and `STDERR`, apparently. So instead of overriding these variables, use `exec(cmd, *args, err: out, out: out)` below.

MikeMcQuaid:
Can you provide some examples of what is/isn't escaped? I'm vaguely 👎 on this just because it'll make the default output look kinda ugly for most users for an edge-case that's useful to some maintainers sometimes.

reitermarkus:
> it'll make the default output look kinda ugly

I wouldn't consider an escaped space ugly when there is supposed to be one. Otherwise the output is exactly the same as before.

alyssais:
Yeah it just quotes strings with spaces and escapes apostrophes and stuff. Most of the time it makes no difference because nothing needs escaping. I think it would be better for users if the commands _were_ escaped, because otherwise it's impossible to tell what's actually happening — is `foo bar` one argument or two? I can't think why anyone would prefer seeing invalid, ambiguous shell commands over correct, valid ones.

reitermarkus:
> I can't think why anyone would prefer seeing invalid, ambiguous shell commands over correct, valid ones.

Exactly.

ilovezfs:
Before:
```
./configure --prefix=/usr/local/Cellar/wget/1.19.1_1 --sysconfdir=/usr/local/etc --with-ssl=openssl --with-libssl-prefix=/usr/local/opt/openssl@1.1 --disable-debug --disable-pcre
```

After:
```
./configure --prefix\=/usr/local/Cellar/wget/1.19.1_1 --sysconfdir\=/usr/local/etc --with-ssl\=openssl --with-libssl-prefix\=/usr/local/opt/openssl@1.1 --disable-debug --disable-pcre
```

If you're a user, I think you want to see the former not the latter.

MikeMcQuaid:
> If you're a user, I think you want to see the former not the latter.

I 100% agree.

> I can't think why anyone would prefer seeing invalid, ambiguous shell commands over correct, valid ones.

They aren't invalid shell commands, they are unescaped shell commands. These are provided for informational purposes rather than copy-paste purposes. As I said above:

> it'll make the default output look kinda ugly for most users for an edge-case that's useful to some maintainers sometimes.

reitermarkus:
Well, the escaped `=` seems like a bug in the standard library.

reitermarkus:
Ok, it needs to be escaped when `set -k` is specified.

MikeMcQuaid:
Passing on this, sorry @alyssais.

reitermarkus:
There is still a bug in here.

MikeMcQuaid:
@reitermarkus I'm not sure I see what the bug is here and I'm very dubious about fixing bugs that haven't been reported by users (as they may well break things elsewhere).

reitermarkus:
It seems the bug doesn't actually occur when using `IO#reopen`.

PR #3120, opened 2017-09-04
docs: Clarify availability of analytics

woodruffw:
This commit tweaks the language in Analytics.md to make it clear
that summaries of install/error data are publicly available,
while only Homebrew maintainers can access the full analytics
(as specified above in the document).


MikeMcQuaid:
Thanks @woodruffw!

JCount:
This is an important clarification to make. Good work @woodruffw! 👍

PR #3121, opened 2017-09-04
Support High Sierra

fxcoudert:

-----

macOS 10.13 is going to be released soon, and we are working on its support. As long as it's considered prerelease, we tell users not to file issues. At this point, I think it's counter-productive and we want to know about potential issues.

Also, the system Ruby on High Sierra is 2.3, so the diagnostics need to be adapted to expect that. However, the Ruby test fails but I'm not sure how to adapt it. Suggestions welcome!

### DIFF
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -34,12 +34,12 @@ module OS
 
     def prerelease?
       # TODO: bump version when new OS is released
-      version >= "10.13"
+      version >= "10.14"
     end
 
     def outdated_release?
       # TODO: bump version when new OS is released
-      version < "10.10"
+      version < "10.11"
     end
 
     def cat


ilovezfs:
We can wait for the GM ;)

MikeMcQuaid:
Yes, I want to wait for at least the GM if not the actual release. Apple still changes stuff under us right up until the GM and we don't have any CI boxes yet so we really can't claim to provide any level of support yet.

Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -195,7 +195,7 @@ def check_for_other_package_managers
       end
 
       def check_ruby_version
-        ruby_version = "2.0"
+        ruby_version = MacOS.version >= "10.13" ? "2.3" : "2.0"
```
MikeMcQuaid: I don't want to go back to supporting different Ruby versions in different places. CC @mistydemeo how's the Ruby `2.3.3p222` stuff looking? CC @alyssais @xu-cheng in case either of you have been working on the portable Ruby recently, either.

xu-cheng: I don't think there is any problem for portable ruby 2.3. except that you have to build it on your own (I don't have access to 10.5 VM).

fxcoudert:
I'm happy with waiting for the GM. Just for the sake of data: 10.13 currently represents 11% of our installs, compared to 75% for 10.12, 12% for 10.11, and 2.6% for 10.10.

DomT4:
> As long as it's considered prerelease, we tell users not to file issues.

Doesn't usually stop them much, to be honest 😸.

bai:
GM has been just released. Any blockers on getting this shipped?

DomT4:
Probably Mike's comment on this:
> I don't want to go back to supporting different Ruby versions in different places.

MikeMcQuaid:
> Yes, I want to wait for at least the GM if not the actual release.

Given we still don't have any CI running: I want to wait for the actual release, thanks.

DomT4:
Did `/usr/include` completely vanish for everyone else with the High Sierra GM lol? I guess it makes sense with the new `clang` behaviour, but, 😅.

ilovezfs:
Presumably you just need to reinstall the CLT.

DomT4:
Yeah, _almost_ the entire CLT installation got wiped by updating to the GM. A little unexpected. There was no indication that would happen from the upgrade or macOS in general, unless I missed a snippet in the release notes on the developer website.

DomT4:
Hmm. I think it has something to do with the apparently mandatory migration to APFS. Ho hum.

ilovezfs:
Sounds like a bug in previous installers if they weren't doing that. So is /usr/include back after CLT reinstall?

DomT4:
I deliberately stayed on HFS+ until now, and at least as far as I recall this is the first time `/usr/include` has vanished after updating/upgrading/your preferred term here.

I noticed during the CLT installation `diskarbitrationd` was asking for access to `/Volumes/Preboot` which hasn't happened before, obviously I guess since IIRC it's an APFS thing, so perhaps something hit a hiccup there during the update/upgrade/whatever.

>  So is /usr/include back after CLT reinstall?

Aye.

ilovezfs:
It's not actually mandatory. There's a `startosinstall` command in the installer .app with a `--converttoapfs NO` option.

DomT4:
> It's not actually mandatory

Extremely pushy rather than mandatory then, if you prefer 😉.

ilovezfs:
It's probably more about user confusion than pushiness. Think of how many useless "Should I check that box or not?" inquiries and FUD they've prevented.

DomT4:
Yeah, I get it from that point of view.

I have some nerves around using a new filesystem a whole bunch of third-parties haven't even considered the behaviour of yet, but I guess Apple are happy enough with the stability of it to be rolling it out by default without that _"Do you want APFS or na?"_ checkbox displayed as part of the graphical installation process.

Anyway, just wanted to see whether the `/usr/include` vanishing thing was a common issue or whether it was a snag in the HFS+ => APFS migration really.

ilovezfs:
>Anyway, just wanted to see whether the /usr/include vanishing thing was a common issue or whether it was a snag in the HFS+ => APFS migration really.

I'm curious whether `brew config` shows the CLT as being installed or as being not installed when you had the CLT installed before the upgrade to High Sierra, but afterward /usr/include has gone missing. If we are showing the CLT as being installed, we'll need to fix that so it shows as being not installed and/or add a doctor diagnostic about it.

DomT4:
After the update `brew config` printed `N/A` on CLT existence. The only thing left post-update in the `/Library/Developer/CommandLineTools/usr` folder was `share/man`.

ilovezfs:
>After the update brew config printed N/A on CLT existence. The only thing left post-update in the /Library/Developer/CommandLineTools/usr folder was share/man.

This just happened exactly as you described (only thing remaining is the `whatis` man page) on CI after updating to the GM. No APFS involved.

MikeMcQuaid:
@fxcoudert This can be squashed and merged 🎉

Issue #1191, opened 2016-09-30
-L#{sysroot}/usr/lib isn't always removed by superenv with Xcode 8


tdsmith:
Observations:
- superenv isn't stripping some paths that resolve to -L#{sysroot}/usr/lib: https://github.com/Homebrew/homebrew-core/issues/5347#issuecomment-250629471
- MacOSX10.12.sdk is a symlink to MacOSX.sdk so there exist non-canonical paths for sysroot
- MacOS.sdk_root is non-canonical on Xcode 8: `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk`
- We find sdk paths here: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/os/mac/sdk.rb#L48-L51 and don't immediately canonicalize them
- We blacklist paths here: https://github.com/Homebrew/brew/blame/master/Library/Homebrew/shims/super/cc#L281 and compare them to canonicalized paths

This seems likely to have been responsible for breaking the python `sqlite3` module on Xcode-only Sierra systems.

Inference:
- we should `realpath` all the things, or else
- anticipate Xcode's symlinking behavior in other ways, and/or
- ?

Is there a correct `realpath` discipline we should use? Point of introduction? Point of use? Everywhere?

MikeMcQuaid:
I guess I'd `realpath` everything and then use both the `real` and non-real version to decide what we should filter.

gregory-nisbet:
Is this issue still open?

I think the blacklist code has drifted since this issue was created. Where does it live now? 

You can compare paths for equivalence using the inode and device number and that can save you the trouble of needing to `realpath` and needing to keep track of whether your path has been canonicalized or not.

silly example that prints `true`.

```
#!/usr/bin/ruby

class Test
    def self.file_sameness(a, b)
        if File.exists?(a) and File.exists?(b)
            st_a = File::Stat.new(a)
            st_b = File::Stat.new(b)
            st_a.ino == st_b.ino and st_a.dev == st_b.dev
        else
            false
        end
    end
end

puts Test.file_sameness("/tmp", "/private/tmp")

```

MikeMcQuaid:
> Is this issue still open?

Yes.

> I think the blacklist code has drifted since this issue was created. Where does it live now?

[...]

PR #3124, opened 2017-09-04
[WIP] resolve path when checking include dirs

gregory-nisbet:

-----

adds a new method, `above_symbolically?` to heuristically check whether a path is a parent of another path when symbolic references are considered. Currently we perform the heuristic check by first seeing if the directory is textually a prefix of the path, and then comparing the `realpath`s. This doesn't catch everything... I don't think there's a way to do that without walking the file system and seeing which paths are accessible from `dir` ... and that seems like overkill.

I hunted around for a way to test the code in shims and couldn't find one ... so I wrote a quick test that copies the shim source over to a temporary file without the shebang preamble and then loads it.

The shim tester is kind of a hack (and it leaks temporary files). Is there a better approach for testing this code?

I'm also open to suggestions about how to test a scenario that more closely resembles a conflict with macports or fink or X11 or w/e.

Context: https://github.com/Homebrew/brew/issues/1191

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -224,7 +224,8 @@ class Cmd
       true
     else
       # ignore MacPorts, Boxen's Homebrew, X11, fink
-      !path.start_with?("/opt/local", "/opt/boxen/homebrew", "/opt/X11", "/sw", "/usr/X11")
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Thanks for the PR @gregory-nisbet but I'm not sure I understand what this is fixing. If this is related to https://github.com/Homebrew/brew/issues/1191 it should probably be touching code with the line `sysroot` in it somewhere.

PR #3125, opened 2017-09-05
mac/hardware/cpu: recognise Kaby Lake

DomT4:

-----

The 2017 MacBook Pro line shipped with Kaby Lake CPUs. Suspect this is what is causing the `dunno` in https://github.com/Homebrew/homebrew-core/issues/14418#issuecomment-327043548.

### DIFF
--- a/Library/Homebrew/extend/os/mac/hardware/cpu.rb
+++ b/Library/Homebrew/extend/os/mac/hardware/cpu.rb
@@ -50,6 +50,8 @@ module Hardware
             :broadwell
           when 0x37fc219f # Skylake
             :skylake
+          when 0x0f817246 # Kaby Lake
+            :kabylake
           else
             :dunno
           end


DomT4:
Not really any new testing that can be added here as far as I can see, so 🤷‍♂️.

Library/Homebrew/extend/os/mac/hardware/cpu.rb
```diff
@@ -50,6 +50,8 @@ def family
             :broadwell
           when 0x37fc219f # Skylake
             :skylake
+          when 0x0f817246 # Kaby Lake
+            :kabylake
```
ilovezfs: should it be `:kaby_lake`?

DomT4: IMO, yes. But the Linux side of this code & the CPU Hardware tests are already using this style so I just went with it.

ilovezfs: Yuck.

MikeMcQuaid:
Thanks again @DomT4!

PR #3134, opened 2017-09-06
pull: follow redirects when checking if bottle is published

iMichka:

-----

The error I was getting:
```
Verifying bottle: wpscan-2.9.3.x86_64_linux.bottle.tar.gz
Error: Failed to find published wpscan bottle at https://linuxbrew.bintray.com/bottles/wpscan-2.9.3.x86_64_linux.bottle.tar.gz (302 )!
```

Following the redirection fixes that.

### DIFF
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -552,28 +552,16 @@ module Homebrew
         # 401 error is normal while file is still in async publishing process
         url = URI(bottle_info.url)
         puts "Verifying bottle: #{File.basename(url.path)}"
-        http = Net::HTTP.new(url.host, url.port)
-        http.use_ssl = true
         retry_count = 0
-        http.start do
-          loop do
-            req = Net::HTTP::Head.new bottle_info.url
-            req.initialize_http_header "User-Agent" => HOMEBREW_USER_AGENT_RUBY
-            res = http.request req
-            break if res.is_a?(Net::HTTPSuccess)
-
-            unless res.is_a?(Net::HTTPClientError)
-              raise "Failed to find published #{f} bottle at #{url} (#{res.code} #{res.message})!"
-            end
-
-            if retry_count >= max_retries
-              raise "Failed to find published #{f} bottle at #{url}!"
-            end
-            print(wrote_dots ? "." : "Waiting on Bintray.")
-            wrote_dots = true
-            sleep poll_retry_delay_seconds
-            retry_count += 1
+        loop do
+          break if check_publication(f, url)
+          if retry_count >= max_retries
+            raise "Failed to find published #{f} bottle at #{url}!"
           end
+          print(wrote_dots ? "." : "Waiting on Bintray.")
+          wrote_dots = true
+          sleep poll_retry_delay_seconds
+          retry_count += 1
         end
 
         # Actual download and verification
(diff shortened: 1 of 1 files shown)


iMichka:
cc @sjackman @reitermarkus @jonchang @maxim-belkin

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -605,6 +593,37 @@ def verify_bintray_published(formulae_names)
     end
   end
 
+  def check_publication(f, url, max_retries, retry_count)
+    http = create_http(url)
+    http.start do
+      res = get_http_result(http, url)
+      return true if res.is_a?(Net::HTTPSuccess)
+      if res.is_a?(Net::HTTPRedirection)
+        retry_count += 1
```
maxim-belkin: I'd call it `redirect_count`.

iMichka: I used the existing `retry_count` variable, which is used on the main loop, so I kept the same name for consistency.

In fact I could have a separate `retry_count` and `redirect_count` variable, and have 2 different max retries/redirects, but I am not sure it is worth it. Max values of retries is 300, and I do not expect more than 5 redirects (though I am not a web developer, so I may be wrong).

maxim-belkin: Yes, I noticed the resuse... while I agree it's safe to expect just a few redirects, imagine every retry taking 10 redirects. This way you reduce the number of retries from 300 to 30.

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -605,6 +593,37 @@ def verify_bintray_published(formulae_names)
     end
   end
 
+  def check_publication(f, url, max_retries, retry_count)
```
maxim-belkin: I'd call it `check_<something>_status` where `<something>` could be `bottle`

maxim-belkin:
LGTM! :)

Library/Homebrew/dev-cmd/pull.rb
iMichka: What about `is_bottle_published` (it returns a boolean)?

maxim-belkin: I expect someone from upstream to chime in. I just expressed my opinion :)

MikeMcQuaid:
I'd like to verify this solves a macOS problem first.

sjackman:
To recap a conversation in Slack… unpublished or very recently published bottles are not redirected with an HTTP 302 redirect, but older bottles are redirected. It's unclear when that transition happens.

sjackman:
For example, this bottle published ten hours ago, is already being redirected:
```
❯❯❯ curl -sD- https://homebrew.bintray.com/bottles/libxml2-2.9.5.sierra.bottle.tar.gz | head -n1
HTTP/1.1 302
```

iMichka:
I'm with killing that pull request. It does not solve any practical problem on mac, and I will adapt my workflow for Linux.

Just tell me if you still would like to have these changes or not.

sjackman:
As @jcount discovered, though only HTTP `GET` are redirected. HTTP `HEAD` are not.
 ```
❯❯❯ curl -sI https://homebrew.bintray.com/bottles/libxml2-2.9.4_4.el_capitan.bottle.tar.gz | head -n1
HTTP/1.1 200 OK
```

MikeMcQuaid:
> I'm with killing that pull request. It does not solve any practical problem on mac, and I will adapt my workflow for Linux.
> 
> Just tell me if you still would like to have these changes or not.

:cool:, passing on this for now, thanks anyway @iMichka!

iMichka:
Thanks, no problem :) As it does not fix anything, no need to merge it :)

PR #3139, opened 2017-09-08
mach: Avoid reopening the file for relocation

woodruffw:
Second time's the charm...

cc @ilovezfs


woodruffw:
`openssl` now builds successfully with this patch and the tests are :green_apple:, so I think this is good to go. I'll merge it tonight unless anybody has new feedback on it.

ilovezfs:
@woodruffw I think it would be good to add a test here that would have caused a failure for https://github.com/Homebrew/brew/pull/3101.

Also, in https://github.com/Homebrew/homebrew-core/issues/17861#issuecomment-328269813 @philn raised a good point, I think, about whether those errors should in general be fatal, and why they aren't currently.

woodruffw:
> I think it would be good to add a test here that would have caused a failure for #3101.

Will do!

> about whether those errors should in general be fatal, and why they aren't currently.

Hmm. I think the original reason for them being non-fatal was because, in principle, executables with broken linkages can still be run (so long as the broken references are never loaded). That might only be true for lazy-loaded dylibs, however.

I'm :+1: on making install linkage errors fatal by default (with a manual override perhaps?).

MikeMcQuaid:
> I'm 👍 on making install linkage errors fatal by default (with a manual override perhaps?).

👍 on this but don't think we need an override (for now, anyway).

woodruffw:
> :+1: on this but don't think we need an override (for now, anyway).

Cool. I'll add it to this PR, along with the tests when I have time.

woodruffw:
@ilovezfs I think the failure in #3101 was caused by the change in class, so not something I can easily add to unit tests. Mind if I skip adding tests here, and shore up @mansimarkaur's work in #3079 instead?

MikeMcQuaid:
@woodruffw That works for me! 🚢 when you're confident.

ilovezfs:
Actually no, that's not OK with me, not that anyone waited for my response.

MikeMcQuaid:
Apologies to @ilovezfs and @woodruffw here: I should not have 👍d this being merged without a regression test that @ilovezfs had specifically requested.

PR #3154, opened 2017-09-13
vendor-install: Do not suppress stderr

sjackman:
Do not suppress error messages such as:
```
ruby: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

### DIFF
--- a/Library/Homebrew/cmd/vendor-install.sh
+++ b/Library/Homebrew/cmd/vendor-install.sh
@@ -139,7 +139,7 @@ install() {
   tar "$tar_args" "$CACHED_LOCATION"
   safe_cd "$VENDOR_DIR/portable-$VENDOR_NAME"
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'd like these messages to be suppressed for users.

sjackman:
The error message
```
ruby: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
Failed to install vendor ruby
```
is more Googlable and informative than simply
```
Failed to install vendor ruby
```
so my preference is the former. I'll modify the PR if you confirm that your preference is the latter.

ilovezfs:
It seems like it would be best to figure out how to make it succeed or at least tell the user how to fix it.

sjackman:
I plan to link `zlib` statically when I build the Linux bottle for `portable-ruby@2.3`. That would resolve this particular issue.

MikeMcQuaid:
Passing on this but thanks anyway @sjackman.

PR #3170, opened 2017-09-18
Check req.satisfied? before req formula install

### LINKED ISSUE #3159: Issue when a dependency depends on :mpi
- [X] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at the relevant tap e.g. for Homebrew/homebrew-core https://github.com/Homebrew/homebrew-core/issues/new
- [X] Ran `brew doctor`, fixed all issues and retried your prior step?
- [X] Ran `brew config` and `brew doctor` and included their output with your issue?

https://gist.github.com/dpo/dcb38eec73e0a4c9a7dc60b6941a96f6

Build `homebrew/science/mumps` and dependencies against `mpich` to investigate https://github.com/Homebrew/homebrew-science/issues/6318.

The dependencies `parmetis` and `scalapack` (which depend on `:mpi`) can be built from source against `mpich`, but `mumps` cannot:

```bash
$ brew install mpich # after uninstalling open-mpi and everything depending on it
$ brew install -s parmetis scalapack # works fine

$ brew install -s mumps
==> Installing mumps from homebrew/science
==> Installing dependencies for homebrew/science/mumps: open-mpi
==> Installing homebrew/science/mumps dependency: open-mpi
Error: Cannot install open-mpi because conflicting formulae are installed.
mpich: because both install mpi__ compiler wrappers

Please brew unlink mpich before continuing.
```

- What you expected to happen

See `mumps` compile against `mpich` since the `:mpi` dependency is satisfied.

# Symptom

It seems that if a formula has a dependency D that depends on `:mpi`, then Homebrew wants to install `open-mpi`, regardless of whether D is already installed and regardless of whether it's been built from source against `mpich`. Here's a minimal example with a trimmed down `mumps.rb`:
```ruby
class Mumps < Formula
  desc "Parallel Sparse Direct Solver"
  homepage "http://mumps-solver.org"
  url "http://mumps.enseeiht.fr/MUMPS_5.1.1.tar.gz"
  sha256 "a2a1f89c470f2b66e9982953cbd047d429a002fab9975400cef7190d01084a06"

  depends_on :mpi
  # depends_on "scalapack"  # scalapack depends_on :mpi

  def install
    puts "Hello, MUMPS!"
  end
end
```
Assume `open-mpi` is not installed but `mpich` is. Then `brew install -s` the above formula succeeds (i.e., does not complain about installing `open-mpi`).

Now, uncomment the commented out line (and even comment out the explicit `:mpi` dependency) and assume `scalapack` was built from source against `mpich`:
```
$ otool -L /usr/local/opt/scalapack/lib/libscalapack.dylib 
/usr/local/opt/scalapack/lib/libscalapack.dylib:
	...
	/usr/local/opt/mpich/lib/libmpifort.12.dylib (compatibility version 14.0.0, current version 14.0.0)
	/usr/local/opt/mpich/lib/libmpi.12.dylib (compatibility version 14.0.0, current version 14.0.0)
	/usr/local/opt/mpich/lib/libpmpi.12.dylib (compatibility version 14.0.0, current version 14.0.0)
	...
```
Then
```
$ brew install -s /tmp/mumps.rb
==> Installing dependencies for mumps: open-mpi
==> Installing mumps dependency: open-mpi
Error: Cannot install open-mpi because conflicting formulae are installed.
  mpich: because both install mpi__ compiler wrappers

Please `brew unlink mpich` before continuing.
```

I've inserted debugging statements in `mpi_requirement.rb` and `satisfy` returns `true`. Nevertheless, Homebrew wants `open-mpi`. It would be great to know if others can reproduce this.

scpeters:

-----

This changes the order of logic in `FormulaInstaller::expand_requirements` so that we check if a requirement is already satisfied before trying to install a formula to satisfy it.

This fixed #3159 for me.

cc @dpo

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -436,11 +436,11 @@ class FormulaInstaller
           Requirement.prune
         elsif req.build? && install_bottle_for_dependent
           Requirement.prune
+        elsif req.satisfied?
+          Requirement.prune
         elsif install_requirement_formula?(req_dependency, req, install_bottle_for_dependent)
           deps.unshift(req_dependency)
           formulae.unshift(req_dependency.to_formula)
-          Requirement.prune
-        elsif req.satisfied?
           Requirement.prune
         elsif !runtime_requirements.include?(req) && install_bottle_for_dependent
           Requirement.prune


ilovezfs:
The existing behavior is intentional so that the default formula is used with bottles.

dpo:
@ilovezfs In #3159, somehow the fact that `mpich` is installed still forces `open-mpi` to try to install, which makes it impossible for users to build a toolchain on top of `mpich` if they want to. That doesn't seem to be the intended behavior?!

scpeters:
@ilovezfs I think I understand what you're saying. I'll phrase things using the example of #3159 with `mpich` pre-installed and `scalapack` built from source against `mpich`, though `open-mpi` is the default formula for the `mpi` requirement.

## Without this patch:

* `brew install mumps` fails:

~~~
$ brew install mumps
==> Installing mumps from homebrew/science
==> Installing dependencies for homebrew/science/mumps: open-mpi
==> Installing homebrew/science/mumps dependency: open-mpi
Error: Cannot install open-mpi because conflicting formulae are installed.
  mpich: because both install mpi__ compiler wrappers

Please `brew unlink mpich` before continuing.
~~~

* `brew install -s mumps` also fails:

~~~
$ brew install -s mumps
==> Installing mumps from homebrew/science
==> Installing dependencies for homebrew/science/mumps: open-mpi
==> Installing homebrew/science/mumps dependency: open-mpi
Error: Cannot install open-mpi because conflicting formulae are installed.
  mpich: because both install mpi__ compiler wrappers

Please `brew unlink mpich` before continuing.
~~~

## With this patch

* `brew install mumps` installs a broken bottle because `open-mpi` is not installed

* `brew install -s mumps` succeeds

## Conclusion

So I think while this did fix building `mumps` from source against `mpich`, it fails to recognize that the bottle won't be installed correctly. I'll take another look at the logic.

PR #3173, opened 2017-09-19
Revert "popen: Do not suppress stderr"

ilovezfs:
Reverts Homebrew/brew#3099

### DIFF
--- a/Library/Homebrew/utils/popen.rb
+++ b/Library/Homebrew/utils/popen.rb
@@ -13,7 +13,7 @@ module Utils
         return pipe.read unless block_given?
         yield pipe
       else
-        $stderr.reopen("/dev/null", "w") if !ARGV.debug? && !ARGV.verbose?
+        $stderr.reopen("/dev/null", "w")
         exec(*args)
       end
     end


ilovezfs:
CC @sjackman @MikeMcQuaid #3099 is causing messages like
```
xcrun: error: unable to find utility "gcc-4.0", not a developer tool or in PATH
xcrun: error: unable to find utility "gcc-4.2", not a developer tool or in PATH
```
to get printed on every build failure.

sjackman:
`brew config -v` also replicates this issue as well. I've opened PR #3174 to address this issue.

ilovezfs:
@sjackman OK, thanks. I'll close this one out.

ilovezfs:
@sjackman @MikeMcQuaid going to revert because the `xcrun: error: unable to find utility "gcc-4.0"` isn't the only issue.

@chdiza reports in https://github.com/Homebrew/brew/issues/3181

>Also, brew config is showing the rather unhelpful and probably unintentional:
>```Unable to find any JVMs matching version "(null)".```

sjackman:
I'll address #3181 in #3174.

MikeMcQuaid:
@sjackman I don't really want to play whack-a-mole with this, I'm afraid.

MikeMcQuaid:
An undocumented environment variable for this may be the best call if you really need this functionality.

sjackman:
Note that these messages are only displayed when either `-d` or `-v` is enabled. I suggest changing it so that these messages are only displayed when both `HOMEBREW_DEVELOPER=1` and `-v` are enabled. Changing…
```ruby
$stderr.reopen("/dev/null", "w") if !ARGV.debug? && !ARGV.verbose?
```
to
```ruby
$stderr.reopen("/dev/null", "w") unless ARGV.homebrew_developer? && ARGV.verbose?
```

MikeMcQuaid:
@sjackman That's still going to hit Homebrew developers, non-tag users and CI.

sjackman:
How about…
```ruby
$stderr.reopen("/dev/null", "w") unless ARGV.homebrew_developer?&& ARGV.debug? && ARGV.verbose?
```
Does CI have `ARGV.debug?` enabled?

MikeMcQuaid:
An undocumented environment variable is the only thing that's not going to annoy existing maintainers/users.

sjackman:
I've updated PR https://github.com/Homebrew/brew/pull/3174 to use the environment variable `HOMEBREW_STDERR`.

PR #3174, opened 2017-09-19
popen: Do not suppress stderr when HOMEBREW_STDERR

sjackman:
Suppress the error message:
```
xcrun: error: unable to find utility "gcc-4.0", not a developer tool or in PATH
```

popen: Add an `options` argument. Useful for closing stderr, for example.
```ruby
popen_read("foo", err: :close)
```

### DIFF
--- a/Library/Homebrew/extend/os/mac/development_tools.rb
+++ b/Library/Homebrew/extend/os/mac/development_tools.rb
@@ -9,7 +9,7 @@ class DevelopmentTools
         @locate[key] = if (located_tool = original_locate(tool))
           located_tool
         elsif MacOS.version > :tiger
-          path = Utils.popen_read("/usr/bin/xcrun", "-no-cache", "-find", tool).chomp
+          path = Utils.popen_read("/usr/bin/xcrun", "-no-cache", "-find", tool, err: :close).chomp
           Pathname.new(path) if File.executable?(path)
         end
       end
--- a/Library/Homebrew/utils/popen.rb
+++ b/Library/Homebrew/utils/popen.rb
@@ -13,7 +13,7 @@ module Utils
         return pipe.read unless block_given?
         yield pipe
       else
-        options[:err] ||= :close
+        options[:err] ||= :close unless ENV["HOMEBREW_STDERR"]
         exec(*args, options)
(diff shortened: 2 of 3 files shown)


sjackman:
I had `err: :close` originally, but got the error:
```
== Library/Homebrew/extend/os/mac/development_tools.rb ==
C: 12: 81: Use hash rockets syntax.
```
Is my `rubocop` somehow misconfigured? My `brew` is up-to-date.

Library/Homebrew/utils/popen.rb
```diff
@@ -1,20 +1,22 @@
 module Utils
   def self.popen_read(*args, &block)
-    popen(args, "rb", &block)
+    options = args.last.is_a?(Hash) ? args.pop : {}
```
reitermarkus: Instead of this, you can use the double splat operator: 

```ruby
def self.popen_read(*args, **options, &block)
```

sjackman: Nice! Thanks, Markus. I've updated the PR.

Library/Homebrew/extend/os/mac/development_tools.rb
```diff
@@ -9,7 +9,7 @@ def locate(tool)
         @locate[key] = if (located_tool = original_locate(tool))
           located_tool
         elsif MacOS.version > :tiger
-          path = Utils.popen_read("/usr/bin/xcrun", "-no-cache", "-find", tool).chomp
+          path = Utils.popen_read("/usr/bin/xcrun", "-no-cache", "-find", tool, err: :close).chomp
```
MikeMcQuaid: You can probably remove these two now in favour of just the `HOMEBREW_STDERR` below, I think?

sjackman: I plan on enabling `HOMEBREW_STDERR=1` globally for myself. These two commits silence that wee bit of unnecessary noise from `brew config`. Happy to drop them though if you prefer.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging!

PR #3176, opened 2017-09-19
BottleLoader: Use the formula stored in the bottle

sjackman:

-----

When installing a bottle from the local file system or on http, use the formula stored in the bottle.


Library/Homebrew/formulary.rb
```diff
@@ -122,9 +122,22 @@ def initialize(bottle_name)
...
     end
 
+    def load_file
+      formula_path = HOMEBREW_CACHE_FORMULA/"#{name}.rb"
+      formula_path.dirname.mkpath
+      FileUtils.rm_f formula_path
+      bottle_version = Utils::Bottles.resolve_version @bottle_filename
+      system "tar", "-xf", @bottle_filename, "-C", HOMEBREW_CACHE_FORMULA,
+        "--strip-components", "3", "#{name}/#{bottle_version}/.brew/#{name}.rb",
+        err: :close
```
MikeMcQuaid: It would be good to avoid creating/deleting/writing files in the formulary loader. How about using something like `tar -O` to read it from `stdout`?

sjackman: Good suggestion, Mike. Done.

Library/Homebrew/formulary.rb
```diff
@@ -123,7 +123,16 @@ def initialize(bottle_name)
     end
 
     def get_formula(spec, alias_path: nil)
-      formula = super
+      bottle_version = Utils::Bottles.resolve_version @bottle_filename
+      path = "#{name}/#{bottle_version}/.brew/#{name}.rb"
+      contents = Utils.popen_read "tar", "-xO", "-f", @bottle_filename,
+        "--strip-components", "3", path, err: :close
```
MikeMcQuaid: I don't want `err: :close` to be going throughout the codebase when they are no-ops for most users and when, in this case, the error should never be hit. I'm tempted to say rather than falling back this should raise an exception instead. Thoughts?

Library/Homebrew/formulary.rb
```diff
@@ -123,7 +123,16 @@ def initialize(bottle_name)
...
+      formula = if $CHILD_STATUS.success?
+        Formulary.from_contents name, HOMEBREW_CELLAR/path, contents, spec
+      else
+        super
```
MikeMcQuaid: If this stays as-is: worth printing a `Warning` message in this case?

Library/Homebrew/formulary.rb
sjackman: An error is expected when used with a bottle that does not contain `.brew/#{name.rb}`, as for bottles older than about a year so. For these bottles without an internal formula, the behaviour of `brew install` remains unchanged.

Library/Homebrew/formulary.rb
sjackman: I've added a warning.

Library/Homebrew/formulary.rb
sjackman: Does it make more sense to change the warning to an error and get rid of the old code path?

sjackman: Agreed. I'll change it.

Library/Homebrew/formulary.rb
```diff
@@ -123,7 +123,17 @@ def initialize(bottle_name)
...
     def get_formula(spec, alias_path: nil)
-      formula = super
+      bottle_version = Utils::Bottles.resolve_version @bottle_filename
+      path = "#{name}/#{bottle_version}/.brew/#{name}.rb"
+      contents = Utils.popen_read "tar", "-xO", "-f", @bottle_filename,
+        "--strip-components", "3", path, err: :close
+      formula = if $CHILD_STATUS.success?
+        Formulary.from_contents name, HOMEBREW_CELLAR/path, contents, spec
+      else
+        opoo "This bottle does not contain a formula: #{path}"
```
MikeMcQuaid: I think this should probably just fail rather than have inconsistent behaviour. If you're loading bottles more than a year old you're going to have other problems.

sjackman: Agreed. Done.

Library/Homebrew/formulary.rb
```diff
@@ -122,14 +122,14 @@ def initialize(bottle_name)
       super name, Formulary.path(full_name)
     end
 
-    def get_formula(spec, alias_path: nil)
-      formula = super
+    def get_formula(spec, alias_path: nil) # rubocop:disable Lint/UnusedMethodArgument
```
sjackman: Is `# rubocop:disable Lint/UnusedMethodArgument` the best way to avoid this `brew style` error?
```
W:125: 27: Unused method argument - alias_path.
```
See https://github.com/bbatsov/rubocop/issues/3130

sjackman: Would you prefer…
```ruby
raise "Unexpected non-nil parameter alias_path" unless alias_path.nil?
```

sjackman:
I have a failed test to address. I'll get back to you with that later today.
```
  1) Formulary::factory returns a Formula when given a bottle
     Failure/Error: formula = subject.factory(bottle)
     
     BottleFormulaUnavailableError:
       This bottle does not contain the formula file:
         testball_bottle/0.1/.brew/testball_bottle.rb
```

Library/Homebrew/formulary.rb
MikeMcQuaid: I think `def get_formula(spec, **)` should work.

Library/Homebrew/exceptions.rb
```diff
@@ -555,12 +555,11 @@ def initialize(resource)
...
 
-class BottleVersionMismatchError < RuntimeError
-  def initialize(bottle_file, bottle_version, formula, formula_version)
+class BottleFormulaUnavailableError < RuntimeError
+  def initialize(formula_path)
     super <<-EOS.undent
-      Bottle version mismatch
-      Bottle: #{bottle_file} (#{bottle_version})
-      Formula: #{formula.full_name} (#{formula_version})
+      This bottle does not contain the formula file:
```
MikeMcQuaid: How about outputting the bottle file and expected formula path within it?

Library/Homebrew/formulary.rb
```diff
@@ -122,14 +122,14 @@ def initialize(bottle_name)
...
+      bottle_version = Utils::Bottles.resolve_version @bottle_filename
```
MikeMcQuaid: Can this fail?

Library/Homebrew/formulary.rb
```diff
@@ -122,14 +122,14 @@ def initialize(bottle_name)
...
+      formula_path = "#{name}/#{bottle_version}/.brew/#{name}.rb"
+      contents = Utils.popen_read "tar", "-xO", "-f", @bottle_filename, "--strip-components=3", formula_path
+      raise BottleFormulaUnavailableError, formula_path unless $CHILD_STATUS.success?
+      formula = Formulary.from_contents name, HOMEBREW_CELLAR/formula_path, contents, spec
+      formula.bottle_specification.sha256 @bottle_filename.sha256 => Utils::Bottles.tag
```
MikeMcQuaid: What happens without this, out of interest? I'm 👍 on it, just curious.

Library/Homebrew/formulary.rb
sjackman: Yes, that works, but it does alter the behavior of the method for style, in that `get_formula(spec, monkey: :banana)` no longer fails. https://github.com/bbatsov/rubocop/issues/3130#issuecomment-274852535
My minor style preference is `raise … unless nil?`. I'm fine with `**` though.

Library/Homebrew/formulary.rb
sjackman: If the bottle does not contain a file matching the regex `.+/.+/INSTALL_RECEIPT.json`, then yes, it could fail. It currently throws an exception:
```
Error: undefined method `split' for nil:NilClass
/Users/sjackman/.homebrew/Library/Homebrew/utils/bottles.rb:40:in `resolve_formula_names'
```

sjackman: I've added
```diff
--- a/Library/Homebrew/utils/bottles.rb
+++ b/Library/Homebrew/utils/bottles.rb
@@ -29,9 +29,11 @@ module Utils
       end
 
       def receipt_path(bottle_file)
-        Utils.popen_read("tar", "-tzf", bottle_file).lines.map(&:chomp).find do |line|
+        path = Utils.popen_read("tar", "-tzf", bottle_file).lines.map(&:chomp).find do |line|
           line =~ %r{.+/.+/INSTALL_RECEIPT.json}
         end
+        raise "This bottle does not contain the file INSTALL_RECEIPT.json: #{bottle_file}" if path.nil?
+        path
       end
 
       def resolve_formula_names(bottle_file)
```

Library/Homebrew/formulary.rb
sjackman: Because the formula stored in the bottle does not have a `bottle` block, the `bottle` method returns `nil`, and causes the error. In principle that error would have also been possible before if the formula in the repo did not contain a bottle block, which is uncommon.
```
==> Pouring the cached bottle
Error: undefined method `compatible_cellar?' for nil:NilClass
/Users/sjackman/.homebrew/Library/Homebrew/formula_installer.rb:104:in `pour_bottle?'
```
I'll add a commit to address this issue.

Library/Homebrew/exceptions.rb
sjackman: Done.

Library/Homebrew/formulary.rb
sjackman: Done.

Library/Homebrew/formulary.rb
sjackman: Done.

Library/Homebrew/formula_installer.rb
```diff
@@ -101,7 +101,7 @@ def pour_bottle?(install_bottle_options = { warn: false })
       return false
     end
 
-    unless bottle.compatible_cellar?
+    if bottle && !bottle.compatible_cellar?
```
MikeMcQuaid: `unless bottle&.compatible_cellar?`

Library/Homebrew/utils/bottles.rb
```diff
@@ -29,9 +29,11 @@ def native_regex
       end
 
       def receipt_path(bottle_file)
-        Utils.popen_read("tar", "-tzf", bottle_file).lines.map(&:chomp).find do |line|
+        path = Utils.popen_read("tar", "-tzf", bottle_file).lines.map(&:chomp).find do |line|
           line =~ %r{.+/.+/INSTALL_RECEIPT.json}
         end
+        raise "This bottle does not contain the file INSTALL_RECEIPT.json: #{bottle_file}" if path.nil?
```
MikeMcQuaid: `unless path`

Library/Homebrew/formula_installer.rb
sjackman: I didn't know about the `&.` operator! Fantastic!

MikeMcQuaid: Will be enforced by `brew style` from https://github.com/Homebrew/homebrew-portable/pull/50. Wasn't supported in Ruby 2.0.

sjackman: `formula.bottle` is `nil` when `bottle.compatible_cellar?` is false.
Using `formula.bottle_specification.compatible_cellar?` rather
than `formula.bottle.compatible_cellar?` avoids this problem

Library/Homebrew/utils/bottles.rb
sjackman: Done.

sjackman:
I found one odd use case: `brew install -s local_bottle.tar.gz` failed when copying the formula to `.brew/NAME.rb`, because the formula is inside the bottle and not stored anywhere on disk. I've fixed this issue with commit 731b1b1e9c36d59ce48c05799a3ddc172d916dd8.

sjackman:
Do you think it would be useful to add a `contents` method to the `Formula` class, that returns the Ruby code of the formula?

MikeMcQuaid:
@sjackman If there's 3 or more locations it would be used: yep, otherwise: nope.

ilovezfs:
Masking the formula that is in the git repository in favor of one buried in a binary and, in particular, allowing that to override the checksum, seems like a security risk to me.

MikeMcQuaid:
> Masking the formula that is in the git repository in favor of one buried in a binary and, in particular, allowing that to override the checksum, seems like a security risk to me.

I think the checksum should be able to be set if it's unset if that's needed for internal code but it shouldn't be able to be overridden, I agree. I wonder if it's worth reading the version from the both formulae and using the one in Git if the version/revision/`pkg_version` match (as otherwise we are unlikely to have old bottles remain working after e.g. major `post_install` changes). Alternatively, the best middle ground might be only using the `post_install` from the bottled formula as that's the only bit that affects bottles at all.

ilovezfs:
If we patch a critical CVE that would still let someone blithely `brew install ./badversion` since the versions wouldn't be the same.

sjackman:
Sure. I can do that.

sjackman:
My most recent commit displays a warning when downgrading a formula.

sjackman:
> I think the checksum should be able to be set if it's unset if that's needed for internal code

Adding the `sha256` to the bottle spec is no longer necessary thanks to fixing up the logic in `pour_bottle?` not to freak out when `formula.bottle` is nil. I've removed the line that sets the `sha256` of the bottle.

Library/Homebrew/formula_installer.rb
```diff
@@ -101,6 +100,7 @@ def pour_bottle?(install_bottle_options = { warn: false })
       return false
     end
 
+    bottle = formula.bottle_specification
```
MikeMcQuaid: Why is this needed below, out of interest, and `bottle_specification` rather than `bottle`?

Library/Homebrew/formula_installer.rb
```diff
@@ -235,6 +235,10 @@ def install
       end
       raise CannotInstallFormulaError, message
     end
+    if formula.opt_prefix.directory? &&
+       formula.pkg_version < (v = Keg.new(formula.opt_prefix.resolved_path).version)
+      opoo "Downgrading #{formula.name} from #{v} to #{formula.pkg_version}."
```
MikeMcQuaid: Sorry, rather than downgrading here would be good to check the tab in the bottle and see if the formula file exists and that one has a newer version.

Library/Homebrew/test/support/fixtures/testball_bottle.rb
```diff
@@ -6,7 +6,7 @@ def initialize(name = "testball_bottle", path = Pathname.new(__FILE__).expand_pa
       stable.bottle do
         cellar :any_skip_relocation
         root_url "file://#{TEST_FIXTURE_DIR}/bottles"
-        sha256 "9abc8ce779067e26556002c4ca6b9427b9874d25f0cafa7028e05b5c5c410cb4" => Utils::Bottles.tag
+        sha256 "d48bbbe583dcfbfa608579724fc6f0328b3cd316935c6ea22f134610aaf2952f" => Utils::Bottles.tag
```
MikeMcQuaid: Presumably: adding the formula?

Library/Homebrew/formula_installer.rb
sjackman: `formula.bottle` returns `nil` when `compatible_cellar?` is false.
See https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L330
and https://github.com/Homebrew/brew/blob/master/Library/Homebrew/software_spec.rb#L93

Rather than `formula.bottle_specification.compatible_cellar?` we could use `formula.bottled?`. Do you have a preference?

Library/Homebrew/test/support/fixtures/testball_bottle.rb
sjackman: Yep

Library/Homebrew/formula_installer.rb
sjackman: Ah, got it. Sorry I misunderstood. I've updated the warning.

Library/Homebrew/formula_installer.rb
MikeMcQuaid: `formula.bottled?` would be preferable, yeh.

Library/Homebrew/formula_installer.rb
```diff
@@ -236,6 +236,15 @@ def install
       raise CannotInstallFormulaError, message
     end
 
+    # Warn if a more recent version of this formula is available in the tap.
+    begin
+      if formula.pkg_version < (v = Formulary.factory(formula.full_name).pkg_version)
+        opoo "#{formula.name} #{v} is available and more recent than version #{formula.pkg_version}."
```
MikeMcQuaid: Use `formula.full_name` here? @ilovezfs, does this look OK?

sjackman: Done.

Library/Homebrew/formula_installer.rb
sjackman: Done.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Woo hoo! Thanks for merging, Mike!

PR #3177, opened 2017-09-19
brew info: report if no kegs are linked

maxim-belkin:
- [x] Have you added an explanation of what your changes do and why you'd like us to include them? **(see below)**

-----

**Purpose**: Instruct `brew info <formulae>` to print a message `No linked kegs found!` for every (installed) formula that has no linked kegs.

Currently, `brew` marks linked kegs with an asterisk (`*`). The absence of an asterisk signifies that there are no linked kegs. However, this is rather cryptic. This PR makes the absence of linked kegs obvious.

Related to `brew test <formula>` because formulae with no linked kegs fail tests.

### DIFF
--- a/Library/Homebrew/cmd/info.rb
+++ b/Library/Homebrew/cmd/info.rb
@@ -143,6 +143,7 @@ module Homebrew
         tab = Tab.for_keg(keg).to_s
         puts "  #{tab}" unless tab.empty?
       end
+      puts "No linked kegs found!" unless kegs.map(&:linked?).any?
     end
 
     puts "From: #{Formatter.url(github_info(f))}"


MikeMcQuaid:
I'm not sure what I think about this. The code seems sound but I'm not sure I see a widespread user need for this.

ilovezfs:
This is already covered by `doctor`, and the message looks like an error:
```
iMac-TMP:~ joe$ brew info zsh
zsh: stable 5.4.2 (bottled), HEAD
UNIX shell (command interpreter)
https://www.zsh.org/
/usr/local/Cellar/zsh/5.4.2 (1,392 files, 11.9MB)
  Poured from bottle on 2017-09-22 at 06:44:00
No linked kegs found!
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/zsh.rb
==> Dependencies
Required: gdbm ✔, pcre ✔
==> Options
--with-unicode9
	Build with Unicode 9 character width support
--without-etcdir
	Disable the reading of Zsh rc files in /etc
--HEAD
	Install HEAD version
```
so this seems more likely to cause confusion than anything else.

Also, we've now made the opt-linked status persistent (`brew unlink` will persist through upgrades and reinstalls), so the fact that there are no linked kegs for a given formula is not something to be exclamatory about.

>Related to brew test <formula> because formulae with no linked kegs fail tests.

You can use `brew test -f`

ilovezfs:
Thanks for the PR nonetheless @maxim-belkin!

PR #3178, opened 2017-09-20
Fix version parsing

naoty:

-----

I fixed version parsing. While I publish [naoty/git-misc](https://github.com/naoty/git-misc) from [naoty/homebrew-misc](https://github.com/naoty/homebrew-misc), I found that the upgrade of `git-misc` was failed because the name of Cache file is `git-misc-misc.tar.gz` and the version was missing.

When the download URL, https://github.com/naoty/git-misc/releases/download/0.2.0/git-misc.tar.gz, is parsed into a version, it is wrongly matched with [this line](https://github.com/Homebrew/brew/blob/b93fa96ec8b79bbe97253bc4a493c82ce1bb6d50/Library/Homebrew/version.rb#L439). So, I made this pattern more strict.

A change in `node_spec.rb` is needed to pass tests. It looks like a dummy text and so I fixed it to an actual download URL.

### DIFF
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -436,7 +436,7 @@ class Version
     return m.captures.first unless m.nil?
 
     # e.g. https://www.openssl.org/source/openssl-0.9.8s.tar.gz
-    m = /-v?(\d+(?:\.\d+)*[^-]*)/.match(stem)
+    m = /-v?(\d+[^-]+)/.match(stem)
     return m.captures.first unless m.nil?
 
     # e.g. astyle_1.23_macosx.tar.gz
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -436,7 +436,7 @@ class Version
     return m.captures.first unless m.nil?
 
     # e.g. https://www.openssl.org/source/openssl-0.9.8s.tar.gz
-    m = /-v?([^-]+)/.match(stem)
+    m = /-v?(\d+(?:\.\d+)*[^-]*)/.match(stem)
     return m.captures.first unless m.nil?
 
     # e.g. astyle_1.23_macosx.tar.gz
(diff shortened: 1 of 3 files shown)


Library/Homebrew/version.rb
```diff
@@ -436,7 +436,7 @@ def self._parse(spec)
     return m.captures.first unless m.nil?
 
     # e.g. https://www.openssl.org/source/openssl-0.9.8s.tar.gz
-    m = /-v?([^-]+)/.match(stem)
+    m = /-v?(\d+(?:\.\d+)*[^-]*)/.match(stem)
```
JCount: If you make this `m = /-v?(\d+[^-]+)/.match(stem)`, it should correctly fall through to the code designed to detect version numbers from this sort of url.

naoty:
@JCount Thank you for review! I just fixed it.

naoty:
CI was failed because https://stockfish.s3.amazonaws.com/stockfish-8-src.zip was failed to parse.

Library/Homebrew/version.rb
```diff
@@ -436,7 +436,7 @@ def self._parse(spec)
     return m.captures.first unless m.nil?
 
     # e.g. https://www.openssl.org/source/openssl-0.9.8s.tar.gz
-    m = /-v?([^-]+)/.match(stem)
+    m = /-v?(\d+[^-]+)/.match(stem)
     return m.captures.first unless m.nil?
```
JCount:
```
# e.g. https://stockfish.s3.amazonaws.com/stockfish-8-src.zip
# e.g. https://launchpad.net/widelands/build19/build19/+download/widelands-build19-src.tar.bz2
m = /-([a-zA-Z]*\d+)-/.match(stem)
```
I think this should work.

JCount: Sigh... I never get this review workflow right.

naoty:
I'm sorry for introducing new bugs. 😢 

I will try to add some tests about stockfish and widelands to `version_spec.rb`, and add your snippet.

ilovezfs:
>I'm sorry for introducing new bugs. :cry:

No worries :heart:

naoty:
I tried to add new tests and codes, but new bugs were introduced again. I give up to solve this problem by myself 😢

ilovezfs:
@naoty all right. Thanks for the PR, nonetheless!

JCount:
@naoty Don't worry about it. It was entirely my fault since I was trying to help you while I was half asleep. 😅

PR #3180, opened 2017-09-20
test: report summary of failed tests

maxim-belkin:

-----

Report formulae that fail their tests.
Example:
```
$ brew test adns gcc jmhobbs/parrot/terminal-parrot
Testing adns
==> /usr/local/Cellar/adns/1.5.1/bin/adnsheloex --version
Error: Testing requires the latest version of gcc
Testing jmhobbs/parrot/terminal-parrot
==> /usr/local/Cellar/terminal-parrot/1.1.0/bin/parrot -loops 1
Last 15 lines ...
[snip]
... 
Error: The following formulae failed to successfully complete the tests:
Error: jmhobbs/parrot/terminal-parrot
```
That is, only report formulae that are installed, linked and fail their tests.

### DIFF
--- a/Library/Homebrew/dev-cmd/test.rb
+++ b/Library/Homebrew/dev-cmd/test.rb
@@ -94,9 +94,10 @@ module Homebrew
         ENV.replace(env)
       end
     end
-    unless failed_formulae.empty?
-      ofail "The following formulae failed to successfully complete the tests:"
-      ofail failed_formulae.map(&:full_name).join(" ")
-    end
+    return false if failed_formulae.empty?
+    ofail <<-EOF.undent
+      The following formulae failed to complete the tests successfully:
+      #{failed_formulae.map(&:full_name).join(" ")}
+    EOF
   end
 end
--- a/Library/Homebrew/dev-cmd/test.rb
+++ b/Library/Homebrew/dev-cmd/test.rb
@@ -26,6 +26,7 @@ module Homebrew
   def test
     raise FormulaUnspecifiedError if ARGV.named.empty?
 
+    failed_formulae = []
     ARGV.resolved_formulae.each do |f|
       # Cannot test uninstalled formulae
       unless f.installed?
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/test.rb
```diff
@@ -86,7 +86,7 @@ def test
         ofail "#{f.full_name}: failed"
         puts e.message
         failed_formulae << f
-      rescue Exception => e
+      rescue StandardError => e
```
maxim-belkin: I added this change to fix the following stylistic error:

```
$ brew style ./Library/Homebrew/dev-cmd/test.rb
== Library/Homebrew/dev-cmd/test.rb ==
W: 87:  7: Avoid rescuing the Exception class. Perhaps you meant to rescue StandardError?

```

MikeMcQuaid:
Are there other command that do similarly? It occurs to me that this is inconsistent with how e.g. `install` might behave?

maxim-belkin:
If a user issues a call to test a large number of formulae, these tests might take a while. With this PR, a user will a summary of which formulae failed tests. If `brew test` is executed in a CI, it will be easier to report which formulae failed.

`install` is different because when we're dealing with installations there might be a waterfall effect.

MikeMcQuaid:
@maxim-belkin I'd like to see this implemented for e.g. `install`'s `caveats` before we do this for other commands.

Library/Homebrew/dev-cmd/test.rb
```diff
@@ -84,12 +85,19 @@ def test
       rescue ::Test::Unit::AssertionFailedError => e
         ofail "#{f.full_name}: failed"
         puts e.message
-      rescue Exception => e # rubocop:disable Lint/RescueException
```
MikeMcQuaid: You're changing behaviour here, please change it back, thanks.

Library/Homebrew/dev-cmd/test.rb
```diff
@@ -84,12 +85,19 @@ def test
...
         ofail "#{f.full_name}: failed"
         puts e, e.backtrace
+        failed_formulae << f
       ensure
         ENV.replace(env)
       end
     end
+    return false if failed_formulae.empty?
+    ofail <<-EOF.undent
+      The following formulae failed to complete the tests successfully:
```
MikeMcQuaid: This should not be output if a single formula was passed (which is the typical case which again makes me wonder if this is a useful change, I'm afraid).

PR #3182, opened 2017-09-21
superenv: help Autotools with 10.13 SDK on 10.12

ilovezfs:

-----

The GNU Autotools tests for whether futimens and utimensat are available reliably come to incorrect conclusions on 10.12 with the 10.13 SDK in Xcode 9. This overrides its decisions by forcing the right answer in superenv using ac_cv_func_* environment variables and setting them to "no" on 10.12.

### DIFF
--- a/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
@@ -96,9 +96,12 @@ module Superenv
       self["SDKROOT"] = MacOS.sdk_path
     end
 
-    # Filter out symbols known not to be defined on 10.11 since GNU Autotools
-    # can't reliably figure this out with Xcode 8 on its own yet.
-    if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+    # Filter out symbols known not to be defined since GNU Autotools can't
+    # reliably figure this out with Xcode 8 and above.
+    if MacOS.version == "10.12" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "9.0"
+      ENV["ac_cv_func_futimens"] = "no"
+      ENV["ac_cv_func_utimensat"] = "no"
+    elsif MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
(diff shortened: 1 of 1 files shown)


ilovezfs:
CC @hjmallon @MikeMcQuaid

andreygursky:
> The GNU Autotools tests for whether futimens and utimensat are available reliably come to incorrect conclusions on 10.12 with the 10.13 SDK in Xcode 9.

Why do you think the autotools' conclusion is incorrect if SDK reports utimensat() is indeed available? This report from SDK is incorrect if you selected to build for e.g. 10.12, thus your workaround with forcing the availability to "no". I'd prefer to change the wording from "autotools ... incorrect" to "SDK 10.13 ... incorrect".

Have you selected, that you're going to build for 10.12? If it is not possible at all, the wording could be something like "SDK 10.13 doesn't support building for pre-10.13."

ilovezfs:
@andreygursky because it's not actually in the libraries. See the linked issues. Note we do set deployment target so that's not the issue ;)

hjmallon:
It 'weak-links' the symbols if the deployment target is older than the oldest for the symbol to be available.

ilovezfs:
Anyway feel free to search "superenv" and review the same pull requests from last year :)

Library/Homebrew/extend/os/mac/extend/ENV/super.rb
```diff
@@ -96,9 +96,12 @@ def setup_build_environment(formula = nil)
...
     end
 
-    # Filter out symbols known not to be defined on 10.11 since GNU Autotools
-    # can't reliably figure this out with Xcode 8 on its own yet.
-    if MacOS.version == "10.11" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "8.0"
+    # Filter out symbols known not to be defined since GNU Autotools can't
+    # reliably figure this out with Xcode 8 and above.
+    if MacOS.version == "10.12" && MacOS::Xcode.installed? && MacOS::Xcode.version >= "9.0"
+      ENV["ac_cv_func_futimens"] = "no"
+      ENV["ac_cv_func_utimensat"] = "no"
```
MikeMcQuaid: If there's any more in a future PR let's make this a loop like below.

PR #3191, opened 2017-09-22
brew: keep HOMEBREW_PREFIX/bin in PATH with HOMEBREW_ENV_FILTERING

### LINKED ISSUE #2875: HOMEBREW_ENV_FILTERING ignores locally-satisfied requirements (?)

```
=> brew doctor
Your system is ready to brew.

=> brew config (without HOMEBREW_ENV_FILTERING set)
HOMEBREW_VERSION: 1.2.4-23-g8ec08479a
HEAD: 8ec08479ae71a849ea1e648ccda7a85771c14d01
Last commit: 53 minutes ago
Core tap HEAD: 695d9b28f83037ecd11bbb3be3b3a79db0e1b004
Core tap last commit: 5 hours ago
HOMEBREW_PREFIX: /usr/local
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.1 build 802
Git: 2.13.2 => /usr/local/opt/git/bin/git
Perl: /usr/local/bin/perl => /usr/local/Cellar/perl/5.26.0/bin/perl
Ruby: /usr/local/var/rbenv/shims/ruby => /usr/local/var/rbenv/versions/2.4.1/bin/ruby
Java: 1.8.0_131
macOS: 10.12.6-x86_64
Xcode: 8.3.3
CLT: 8.3.2.0.1.1492020469
X11: N/A
```

It appears that `HOMEBREW_ENV_FILTERING` stops Homebrew finding requirements that are satisfied locally and defaults to the system instead, which to be honest was somewhat unexpected behaviour even with `HOMEBREW_ENV_FILTERING` set. Not sure if this is intended on the Homebrew side or a legitimate bug.

Noticed this today, because I install `svn --with-python` which triggers a from-source build, and as a nice side-effect the `perl` bindings are built against the version of `perl` I actually use locally.

Reproduction:
```
=> export HOMEBREW_ENV_FILTERING="1"
=> brew install perl
=> perl --version
This is perl 5, version 26, subversion 0 (v5.26.0) built for darwin-thread-multi-2level
=> brew install subversion --with-python
=> ls -1 /usr/local/Cellar/subversion/1.9.6/lib/perl5/site_perl/
5.18.2
=> unset HOMEBREW_ENV_FILTERING
=> brew reinstall subversion
=> ls -1 /usr/local/Cellar/subversion/1.9.6/lib/perl5/site_perl/
5.26.0
```

maxim-belkin:
**Yes, <a href="#fix">here</a>**
**Nope.** Please suggest tests.
**Nope** 

-----
<a name="fix"></a>
Fixes #2875.

Basically, with `HOMEBREW_ENV_FILTERING` enabled, Homebrew filters out everything, including its own `bin` directory from PATH. This PR brings peace of mind to those who do `HOMEBREW_ENV_FILTERING` stuff occasionally.

```
$ export HOMEBREW_ENV_FILTERING=1
$ brew install subversion --with-python
$ ls -1 /usr/local/Cellar/subversion/1.9.7/lib/perl5/site_perl/
5.26.0
$ unset HOMEBREW_ENV_FILTERING
$ brew reinstall subversion
$ ls -1 /usr/local/Cellar/subversion/1.9.7/lib/perl5/site_perl/
5.26.0
```

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -60,7 +60,7 @@ done
 
 if [[ -n "$HOMEBREW_ENV_FILTERING" ]]
 then
-  PATH="/usr/bin:/bin:/usr/sbin:/sbin"
+  PATH="$HOMEBREW_PREFIX/bin:/usr/bin:/bin:/usr/sbin:/sbin"
 
   FILTERED_ENV=()
   for VAR in HOME SHELL PATH TERM LOGNAME USER "${!HOMEBREW_@}"


MikeMcQuaid:
Sorry but this ends up defeating much of the point of `HOMEBREW_ENV_FILTERING` by randomly adding things to the `PATH` based on user configuration (i.e. their chosen packages). 😭

maxim-belkin:
Are you suggesting #2875 is not an issue and should be, in fact, closed?

MikeMcQuaid:
I think it's an issue but the solution is using `HOMEBREW_PATH` inside relevant requirements.

maxim-belkin:
How can Homebrew even think about Homebrew-ed `perl` if user does not specify anything on the command line?

maxim-belkin:
Just a reminder, in #2875 the "problem" is that
```
brew install subversion --with-python
```
uses system `perl`, not Homebrew-ed one.

There is nothing on the command line that would suggest that using system's Perl is wrong

MikeMcQuaid:
I'm not sure I understand the question. Requirements can read the user's path with `HOMEBREW_PATH` and `which` through that to find what they want us to use but we don't want to expose that globally to e.g. the build process.

maxim-belkin:
Em... Could you please elaborate on: 
> Requirements can read the user's path with `HOMEBREW_PATH` and `which` through that to find what they want us to use


Can't we assume that those who use Homebrew have `HOMEBREW_PREFIX/bin` in their `PATH`?

maxim-belkin:
In `subversion` formula,
```
depends_on :perl => ["5.6", :recommended]
```
is satisfied by system's Perl even if Homebrew-ed Perl is installed and is better/newer.

So... because we do not tell Homebrew to use its own Perl and `PATH` is set in such a way that only system's Perl can be found, how can Homebrew even think about using Homebrew-ed Perl?

MikeMcQuaid:
> Can't we assume that those who use Homebrew have HOMEBREW_PREFIX/bin in their PATH?

Yes. Please check out the implementation of `HOMEBREW_ENV_FILTERING`. We reset the `PATH` and keep around `HOMEBREW_PATH` with the original user's value. The prior is used by the build process, the latter can be used by `which` in `Requirement`s to add the full `Cellar` path to the `PATH`.

maxim-belkin:
`def which` and `def which_all` in `requirement.rb` already consults `ORIGINAL_PATHS`.

MikeMcQuaid:
Yes, which is why an additional fix would be required here...

PR #3192, opened 2017-09-22
Add cmd/strip

sjackman:
Strip binary executables and libraries to reduce their size.

-----

```
❯❯❯ brew strip gcc
==> Stripping gcc...
  /usr/local/Cellar/gcc/7.2.0 (1,486 files, 289.8MB)
  /usr/local/Cellar/gcc/7.2.0 (1,486 files, 257.4MB)
```

### DIFF
--- a/Library/Homebrew/cmd/strip.rb
+++ b/Library/Homebrew/cmd/strip.rb
@@ -1,0 +1,60 @@
+#:  * `strip` [`--all`]
+#:    Strip binary executables and libraries to reduce their size.
+#:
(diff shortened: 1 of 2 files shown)


ilovezfs:
I don't think we should do this. It is useful to have the regular symbols even if we're not doing `-g` builds.

MikeMcQuaid:
I think this would be better suited as an external command that users could tap, sorry @sjackman.

sjackman:
No problem

PR #3202, opened 2017-09-24
audit: don't check for homebrew/science duplicates

ilovezfs:

-----

It's just noise to turn the migration PRs red over expected duplicates.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -419,6 +419,7 @@ class FormulaAuditor
 
     @@local_official_taps_name_map ||= Tap.select(&:official?).flat_map(&:formula_names)
                                           .each_with_object({}) do |tap_formula_full_name, name_map|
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -419,6 +419,7 @@ def audit_formula_name
 
     @@local_official_taps_name_map ||= Tap.select(&:official?).flat_map(&:formula_names)
                                           .each_with_object({}) do |tap_formula_full_name, name_map|
+      next if tap_formula_full_name.start_with?("homebrew/science/")
```
MikeMcQuaid: Arguably we could just kill these checks entirely given we're wanting to move more and more stuff from taps to core. Thoughts?

ilovezfs: Yeah. If anything, duplicates in non-Homebrew taps would be more pertinent information at this point.

PR #3209, opened 2017-09-25
install.rb: report caveats for installed formulae

maxim-belkin:
-----

**Pull Request template**:

-----

**Detailed explanation:**

When installing multiple formulae, caveats might get lost in the normal
output that accompanies normal installation. **This PR reminds the user of
all the caveats at the end of the installation.** Caveats are repeated only if
there are more than one formulae being installed. Following calls to
`brew install <formula-with-caveats>` does not result in caveat message
being printed again.

**Example:**
```
$ brew install go python
...
[snip]
...

==> Please take a note of the caveats for the following installed formulae:

==> Caveats for "go":
A valid GOPATH is required to use the `go get` command.
If $GOPATH is not specified, $HOME/go will be used by default:
  https://golang.org/doc/code.html#GOPATH

You may wish to add the GOROOT-based install location to your PATH:
  export PATH=$PATH:/usr/local/opt/go/libexec/bin

==> Caveats for "python":
This formula installs a python2 executable to /usr/local/bin.
If you wish to have this formula's python executable in your PATH then add
the following to ~/.bash_profile:
  export PATH="/usr/local/opt/python/libexec/bin:$PATH"

Pip and setuptools have been installed. To update them
  pip2 install --upgrade pip setuptools

You can install Python packages with
  pip2 install <package>

They will install into the site-package directory
  /usr/local/lib/python2.7/site-packages

See: https://docs.brew.sh/Homebrew-and-Python.html
```

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -277,7 +277,8 @@ module Homebrew
         puts "To install one of them, run (for example):\n  brew install #{taps_search_results.first}"
       end
     end
-    if formulae.size > 1 && formulae.map { |f| !f.caveats.to_s.empty? }.any?
+    return unless formulae.size > 1
+    if formulae.map { |f| !f.caveats.to_s.empty? }.any?
       puts ""
       ohai "Please review the caveats for the following installed formulae:"
       formulae.each do |f|
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -277,15 +277,20 @@ module Homebrew
         puts "To install one of them, run (for example):\n  brew install #{taps_search_results.first}"
       end
     end
-    return unless formulae.size > 1
-    if formulae.map { |f| !f.caveats.to_s.empty? }.any?
-      puts ""
-      ohai "Please review the caveats for the following installed formulae:"
-      formulae.each do |f|
-        puts ""
-        ohai "Caveats for \"#{f}\":", f.caveats.to_s unless f.caveats.to_s.empty?
-      end
-    end
+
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/install.rb
```diff
@@ -277,6 +277,15 @@ def install
         puts "To install one of them, run (for example):\n  brew install #{taps_search_results.first}"
       end
     end
+    return unless formulae.size > 1
```
MikeMcQuaid: Will this handle caveats from dependencies also? Maybe we don't care about outputting all of them. Thoughts, @ilovezfs?

As a code comment: please turn this `unless` into an `if`.

ilovezfs: If we want to improve the signal-to-noise-ratio by re-printing caveats at the end only for formulae the user explicitly installed, it would be good to at least still do so for formulae that had originally been installed-on-request, even if they're merely being upgraded because they're outdated dependencies in this go.

maxim-belkin: > Will this handle caveats from dependencies also?

currently, caveats for dependencies are not displayed. I'll look into this later

MikeMcQuaid: @maxim-belkin Cool! As @ilovezfs mentioned I think `install_on_request` dependencies are a reasonable subset.

maxim-belkin:
now caveats for `installed_on_request` deps are also reported. What is not taken into account is whether these dependencies are beings upgraded in the current run (not sure if commented out line would do exactly that or not).

maxim-belkin:
not ready yet, argh...

maxim-belkin:
Closing for now... will reopen when I have more time to work on it.

PR #3256, opened 2017-10-01
Unshadow `allow_untrusted` option for pkg artifact

claui:
The `pkg` stanza has an option called `allow_untrusted`, which is supposed to cause `/usr/sbin/installer` to be called with the `-allowUntrusted` switch.

PR #3141 seems to have renamed the `pkg_install_opts` field to `options`. At the same time, it introduces an `options` parameter for the `run_installer` method, which shadows the `options` getter method, causing the `allow_untrusted` option to be silently ignored.

The issue affects just the `pkg` stanza because `Hbc::Artifact::Pkg` is the only artifact class that has an `options` method.

This PR removes the shadowing ~~by renaming the method parameters~~ by renaming the field to `stanza_options`; in one case, it uses ~~`ignored_install_options`~~ `_options` for a parameter name (instead of the more canonical `_`) for the sake of clarity.

**Edit:** last paragraph as per @reitermarkus’s suggestions.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
@@ -9,17 +9,19 @@ module Hbc
     class Pkg < AbstractArtifact
       attr_reader :pkg_relative_path
 
-      def self.from_args(cask, path, **options)
-        options.extend(HashValidator).assert_valid_keys(:allow_untrusted, :choices)
-        new(cask, path, **options)
+      def self.from_args(cask, path, **stanza_options)
+        stanza_options.extend(HashValidator).assert_valid_keys(
+          :allow_untrusted, :choices
+        )
+        new(cask, path, **stanza_options)
       end
 
-      attr_reader :path, :options
+      attr_reader :path, :stanza_options
 
-      def initialize(cask, path, **options)
+      def initialize(cask, path, **stanza_options)
         super(cask)
         @path = cask.staged_path.join(path)
-        @options = options
+        @stanza_options = stanza_options
       end
 
       def summarize
@@ -32,7 +34,7 @@ module Hbc
 
       private
 
-      def run_installer(command: nil, verbose: false, **options)
+      def run_installer(command: nil, verbose: false, **_options)
         ohai "Running installer for #{cask}; your password may be necessary."
         ohai "Package installers may write to any location; options such as --appdir are ignored."
         unless path.exist?
@@ -43,7 +45,9 @@ module Hbc
           "-target", "/"
         ]
         args << "-verboseR" if verbose
-        args << "-allowUntrusted" if options.fetch(:allow_untrusted, false)
+        if stanza_options.fetch(:allow_untrusted, false)
+          args << "-allowUntrusted"
+        end
         with_choices_file do |choices_path|
           args << "-applyChoiceChangesXML" << choices_path if choices_path
           command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
@@ -51,7 +55,7 @@ module Hbc
       end
 
       def with_choices_file
-        choices = options.fetch(:choices, {})
+        choices = stanza_options.fetch(:choices, {})
         return yield nil if choices.empty?
 
         Tempfile.open(["choices", ".xml"]) do |file|


Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -26,13 +26,14 @@ def summarize
...
-        run_installer(**options)
+      def install_phase(**install_options)
+        run_installer(**install_options)
       end
 
       private
 
-      def run_installer(command: nil, verbose: false, **options)
+      def run_installer(command: nil, verbose: false,
+        **_ignored_install_options)
```
reitermarkus: Make this just `**_`.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -26,13 +26,14 @@ def summarize
         path.relative_path_from(cask.staged_path).to_s
       end
 
-      def install_phase(**options)
-        run_installer(**options)
+      def install_phase(**install_options)
+        run_installer(**install_options)
```
reitermarkus: I think it would be better to rename the `options` method to keep signatures consistently using `**options`.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
claui: Thanks a lot @reitermarkus for the review!

Yes, for ignored method parameters, I believe `_` is the best placeholder **as long as the thing it stands for is obvious, or at least unambiguous.**

An example to illustrate what I mean in this specific case:

Future dev wants to write a simple PR against `pkg.rb`, which e. g. is supposed to check if `--force` is enabled, and if so, to have the `/usr/sbin/installer` command line use `sudo -n` instead of `sudo`. To accomplish this, the dev replaces the single line:

```
command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
```

with:

```
command.run!("/usr/sbin/installer", sudo: true, sudo_n: options.fetch(:force, false), args: […])
```

(under the assumption that `:force` is in the `options` hash. This is obviously a mistake on the PR author’s end because the `:force` option actually isn’t in `options`, but in `_`. The thing however is, we have just caused the PR author to needlessly fall in a trap: **the most obvious fix is now exactly the wrong thing to do.** If the PR author now tries to fix this by renaming `_` to `options`, this will make the `:force` thing seem to work, but at the same time that’ll subtly break the `allow_untrusted` feature once again because of the shadowing.)

**It’s us who’d be responsible for this regression** (rather than the PR author) because we chose `_` as a placeholder for the _less_ obvious choice. Let’s be open about what we’re ignoring, at least in this particular case. That’s why I’m in favor of `_ignored_install_options` or `_install_options` as the least error-prone choice in this particular case. Not only does Rubocop allow this syntax for unused parameters, it even explicitly suggests it (try removing the leading `_` and then call `brew style`).

That said, in the light of your other suggestion, I’d like to rename it to `_options`. What do you think?

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
claui: Convinced, thanks!

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
reitermarkus: I wouldn't say that `options.fetch(:force, false)` is the *obvious* fix in that case when `**options` is not even in the signature.

I still think `**_` is better because `**_options` won't become `**options`, but rather an additional named parameter should be added.

claui: > I wouldn't say that `options.fetch(:force, false)` is the obvious fix

I feel like there’s a misunderstanding here. Maybe my wording has not been clear enough. Let me put it in other words.

This is what the code **says today:**

```
[…] stanza_options.fetch(:allow_untrusted, false) […]
```

Because **all we have** is `stanza_options `, and `:allow_untrusted` **works fine** because it’s in `stanza_options`, it’s a very **natural mistake** to assume that `:force` would be in `stanza_options` just as well:

```
[…] stanza_options.fetch(:allow_untrusted, false) […]
[…]
[…] stanza_options.fetch(:force, false) […]
```

Note: All I’m saying is that `stanza_options.fetch(:force, false)` is a _natural mistake_ to make while adding the feature. I’m not saying it’s an obvious fix. There’s not even an issue yet that would need fixing.

As `stanza_options.fetch(:force, false)` will of course **not work as expected**, a PR author might then be inclined to rename `_` to `stanza_options` because it seems to make perfect sense at this moment.

Renaming `_` to `stanza_options` would **seemingly fix** the issue; however, not only would such an _obvious fix_ constitute a second mistake, it would also subtly break `allow_untrusted` once again, just like it’s broken today.

<br/>

> I still think **_ is better

Why do you think it’s better? I’d really like to understand.
We’re dealing with two similar but different options hashes. It’s really easy to mix them up. I feel it’s important to give a hint for future developers here.

<br/>

> because **_options won't become **options, but rather an additional named parameter should be added.

How is this an argument in favor of `**_`?

Please keep in mind that this issue currently breaks a dozen of casks. That’s no big deal in itself but all I’m trying to do here is to fix it **and to prevent this particular mistake** from happening again.

claui:
Thanks for being rigorous with your reviews. I appreciate that a lot 👍

PR #3258, opened 2017-10-02
improve 'brew cask _stanza' by checking for known stanzas

Git-Jiro:

-----

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -45,7 +45,8 @@ module Hbc
         @format = :to_yaml if yaml?
 
         return if DSL::DSL_METHODS.include?(stanza)
-        raise ArgumentError, "Illegal stanza: '#{stanza}'"
+        raise ArgumentError,
+          "Unknown/unsupported stanza: '#{stanza}' Check cask reference for supported stanzas."
       end
 
       def run
@@ -72,7 +73,8 @@ module Hbc
 
           if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
              (value.respond_to?(:to_s) && value.to_s == "{}")
-            raise CaskError, "no such stanza '#{artifact_name ? artifact_name : stanza}' on Cask '#{cask}'"
+            stanza_name = artifact_name ? artifact_name : stanza
+            raise CaskError, "no such stanza '#{stanza_name}' on Cask '#{cask}'"
           end
 
           if format
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -48,7 +48,7 @@ module Hbc
         raise ArgumentError,
           <<~EOS
             Unknown/unsupported stanza: '#{stanza}'
-            Check cask reference for supported stanzas.
+            Check Cask reference for supported stanzas.
           EOS
       end
 
@@ -74,8 +74,7 @@ module Hbc
             value = value[artifact_name] if artifact_name
           end
 
-          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
-             (value.respond_to?(:to_s) && value.to_s == "{}")
+          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?)
             stanza_name = artifact_name ? artifact_name : stanza
             raise CaskError, "no such stanza '#{stanza_name}' on Cask '#{cask}'"
           end
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -71,16 +71,16 @@ module Hbc
             next
           end
 
-          if value.nil? || (value.is_a?(Array) && value.empty?)
-            opoo "no such stanza '#{stanza}' on Cask '#{cask}'" unless quiet?
-            puts ""
-            raise ArgumentError, "no such stanza '#{stanza}' on Cask '#{cask}'"
-          end
+          if (value.nil? || value.is_a?(Array) && value.empty?) ||
+             (artifact_name && !value.key?(artifact_name))
 
-          if artifact_name && !value.key?(artifact_name)
-            opoo "no such stanza '#{artifact_name}' on Cask '#{cask}'" unless quiet?
-            puts ""
-            raise ArgumentError, "no such stanza '#{artifact_name}' on Cask '#{cask}'"
+            if artifact_name
+              thing = artifact_name
+            else
+              thing = stanza
+            end
+
+            raise CaskError, "no such stanza '#{thing}' on Cask '#{cask}'"
           end
 
           if stanza == :artifacts
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -57,12 +57,6 @@ module Hbc
         casks(alternative: -> { Hbc.all }).each do |cask|
           print "#{cask}\t" if table?
 
-          unless cask.respond_to?(stanza)
-            opoo "no such stanza '#{stanza}' on Cask '#{cask}'" unless quiet?
-            puts ""
-            next
-          end
-
           begin
             value = cask.send(stanza)
           rescue StandardError
@@ -71,7 +65,9 @@ module Hbc
             next
           end
 
-          if (value.nil? || value.is_a?(Array) && value.empty?) ||
+          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
+             (value.respond_to?(:to_h) && value.to_h.empty?) ||
+             (value.respond_to?(:to_s) && value.to_s == "{}") ||
              (artifact_name && !value.key?(artifact_name))
 
             if artifact_name
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -67,7 +67,7 @@ module Hbc
 
           if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
              (value.respond_to?(:to_s) && value.to_s == "{}") ||
-             (artifact_name && !value.key?(artifact_name))
+             (artifact_name && !value.include?(artifact_name))
 
             if artifact_name
               thing = artifact_name
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -43,6 +43,10 @@ module Hbc
         @stanza = args.shift.to_sym
 
         @format = :to_yaml if yaml?
+
+        unless DSL::DSL_METHODS.include?(stanza)
+          raise ArgumentError, "Illegal stanza: '#{stanza}'"
+        end
       end
 
       def run
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -71,10 +71,16 @@ module Hbc
             next
           end
 
+          if value == nil || (value.kind_of?(Array) && value.empty?)
+            opoo "no such stanza '#{stanza}' on Cask '#{cask}'" unless quiet?
+            puts ""
+            raise ArgumentError, "no such stanza '#{stanza}' on Cask '#{cask}'"
+          end
+
           if artifact_name && !value.key?(artifact_name)
             opoo "no such stanza '#{artifact_name}' on Cask '#{cask}'" unless quiet?
             puts ""
-            next
+            raise ArgumentError, "no such stanza '#{artifact_name}' on Cask '#{cask}'"
           end
 
           if stanza == :artifacts
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -46,7 +46,8 @@ module Hbc
 
         return if DSL::DSL_METHODS.include?(stanza)
         raise ArgumentError,
-          "Unknown/unsupported stanza: '#{stanza}' Check cask reference for supported stanzas."
+          "Unknown/unsupported stanza: '#{stanza}' Check cask reference"\
+          " for supported stanzas."
       end
 
       def run
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -44,9 +44,8 @@ module Hbc
 
         @format = :to_yaml if yaml?
 
-        unless DSL::DSL_METHODS.include?(stanza)
-          raise ArgumentError, "Illegal stanza: '#{stanza}'"
-        end
+        return if DSL::DSL_METHODS.include?(stanza)
+        raise ArgumentError, "Illegal stanza: '#{stanza}'"
       end
 
       def run
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -71,7 +71,7 @@ module Hbc
             next
           end
 
-          if value == nil || (value.kind_of?(Array) && value.empty?)
+          if value.nil? || (value.is_a?(Array) && value.empty?)
             opoo "no such stanza '#{stanza}' on Cask '#{cask}'" unless quiet?
             puts ""
             raise ArgumentError, "no such stanza '#{stanza}' on Cask '#{cask}'"
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -24,9 +24,6 @@ module Hbc
       ARTIFACTS =
         DSL::ORDINARY_ARTIFACT_CLASSES.map(&:dsl_key) +
         DSL::ARTIFACT_BLOCK_CLASSES.map(&:dsl_key)
-      ARTIFACTS_CLASSES =
-        DSL::ORDINARY_ARTIFACT_CLASSES +
-        DSL::ARTIFACT_BLOCK_CLASSES
 
       option "--table",   :table,   false
       option "--quiet",   :quiet,   false
@@ -69,27 +66,12 @@ module Hbc
           end
 
           if stanza == :artifacts
-            result = {}
-            value.each do |e|
-              result[e.class.dsl_key] = e.summarize
-            end
-            value = result
-          end
-
-          if artifact_name
-            result = nil
-            value.each do |k, v|
-              if k == artifact_name
-                result = v
-                break
-              end
-            end
-            value = result
+            value = Hash[value.map { |v| [v.class.dsl_key, v] }]
+            value = value[artifact_name] if artifact_name
           end
 
           if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
-             (value.respond_to?(:to_s) && value.to_s == "{}") # ||
-
+             (value.respond_to?(:to_s) && value.to_s == "{}")
             raise CaskError, "no such stanza '#{artifact_name ? artifact_name : stanza}' on Cask '#{cask}'"
           end
 
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -66,7 +66,7 @@ module Hbc
           end
 
           if stanza == :artifacts
-            value = Hash[value.map { |v| [v.class.dsl_key, v] }]
+            value = Hash[value.map { |v| [v.class.dsl_key, v.to_s] }]
             value = value[artifact_name] if artifact_name
           end
 
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -46,8 +46,10 @@ module Hbc
 
         return if DSL::DSL_METHODS.include?(stanza)
         raise ArgumentError,
-          "Unknown/unsupported stanza: '#{stanza}' Check cask reference"\
-          " for supported stanzas."
+          <<~EOS
+            Unknown/unsupported stanza: '#{stanza}'
+            Check cask reference for supported stanzas.
+          EOS
       end
 
       def run
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -24,6 +24,9 @@ module Hbc
       ARTIFACTS =
         DSL::ORDINARY_ARTIFACT_CLASSES.map(&:dsl_key) +
         DSL::ARTIFACT_BLOCK_CLASSES.map(&:dsl_key)
+      ARTIFACTS_CLASSES =
+        DSL::ORDINARY_ARTIFACT_CLASSES +
+        DSL::ARTIFACT_BLOCK_CLASSES
 
       option "--table",   :table,   false
       option "--quiet",   :quiet,   false
@@ -65,33 +68,29 @@ module Hbc
             next
           end
 
-          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
-             (value.respond_to?(:to_s) && value.to_s == "{}") ||
-             (artifact_name && !value.include?(artifact_name))
-
-            if artifact_name
-              thing = artifact_name
-            else
-              thing = stanza
+          if stanza == :artifacts
+            result = {}
+            value.each do |e|
+              result[e.class.dsl_key] = e.summarize
             end
-
-            raise CaskError, "no such stanza '#{thing}' on Cask '#{cask}'"
+            value = result
           end
 
-          if stanza == :artifacts
-            value = Hash[
-              value.map do |k, v|
-                v = v.map do |a|
-                  next a.to_a if a.respond_to?(:to_a)
-                  next a.to_h if a.respond_to?(:to_h)
-                  a
-                end
-
-                [k, v]
+          if artifact_name
+            result = nil
+            value.each do |k, v|
+              if k == artifact_name
+                result = v
+                break
               end
-            ]
+            end
+            value = result
+          end
 
-            value = value.fetch(artifact_name) if artifact_name
+          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
+             (value.respond_to?(:to_s) && value.to_s == "{}") # ||
+
+            raise CaskError, "no such stanza '#{artifact_name ? artifact_name : stanza}' on Cask '#{cask}'"
           end
 
           if format
--- a/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
@@ -66,7 +66,6 @@ module Hbc
           end
 
           if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
-             (value.respond_to?(:to_h) && value.to_h.empty?) ||
              (value.respond_to?(:to_s) && value.to_s == "{}") ||
              (artifact_name && !value.key?(artifact_name))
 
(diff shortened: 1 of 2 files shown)


Git-Jiro:
@claui you are correct, 100 chars is way too long.
I also fixed the merge conflict by rebasing my changes on the proper commit from master.

claui:
Thanks a lot @Git-Jiro! 👍 

The unfortunate thing is, your improved version has now **caught the attention of Rubocop,** which our Travis CI calls everytime there’s a commit in a pull request that is to go into brew/master.

Actually, this was my fault; Rubocop [says it’s an offense](https://travis-ci.org/Homebrew/brew/jobs/282839457#L295) to have an `if`/`unless` block contain only a `raise`, a statement which yields control. Rubocop actually liked your old version better!

(We’re not the only ones to get caught by this; in case you’re curious, others [have also been bitten](https://github.com/bbatsov/rubocop/issues/1332) by this [GuardClause dilemma](https://stackoverflow.com/questions/40446219/rubocop-guard-clause-dilemma).)

Luckily, **there’s a way to satisfy both Rubocop and me** though. ☺️ Instead of `raise […] unless […]` you can write `return if […]`:

```
return if DSL::DSL_METHODS.include?(stanza)
```

You can then write an unconditional `raise` in the line after that:

```
raise ArgumentError, "Illegal stanza: '#{stanza}'"
```

This should make everyone happy 😉  It’s also a good practice to run `brew style` locally before pushing your commit, which essentially runs Rubocop on your local repo. (I forget it all the time.)

Git-Jiro:
I admit I am guilty as charged. ;-)
I indeed forgot to run brew style. Rubocop should be happy now.
Robocop - Thank you for your cooperation

reitermarkus:
What is the error message when the stanza exists, but not on the Cask? Just to make sure they are clearly distignuishable.

Also, can you add a small test for this? Thanks!

Git-Jiro:
@reitermarkus 

> What is the error message when the stanza exists, but not on the Cask?

`$ brew cask _stanza installer ableton-live`
`Warning: no such stanza 'installer' on Cask 'ableton-live'`
` `
`Error: nothing to print`
`$ echo $?`
`1`

> Also, can you add a small test for this?

Yes, I will add a test for this.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -71,10 +71,16 @@ def run
             next
           end
 
+          if value == nil || (value.kind_of?(Array) && value.empty?)
```
reitermarkus: `value.nil? || value&.to_a&.empty?`

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -71,10 +71,16 @@ def run
...
+            opoo "no such stanza '#{stanza}' on Cask '#{cask}'" unless quiet?
+            puts ""
```
reitermarkus: Remove this.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -71,10 +71,16 @@ def run
...
+            raise ArgumentError, "no such stanza '#{stanza}' on Cask '#{cask}'"
+          end
+
           if artifact_name && !value.key?(artifact_name)
             opoo "no such stanza '#{artifact_name}' on Cask '#{cask}'" unless quiet?
             puts ""
-            next
+            raise ArgumentError, "no such stanza '#{artifact_name}' on Cask '#{cask}'"
```
reitermarkus: I wouldn't classify this as an `ArgumentError`, maybe just raise without specifying a class.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,10 +71,16 @@ def run
             next
           end
 
+          if value.nil? || (value.is_a?(Array) && value.empty?)
```
reitermarkus: `value.nil? || value&.to_a&.empty?`

Library/Homebrew/test/cask/cli/internal_stanza_spec.rb
```diff
@@ -0,0 +1,27 @@
+describe Hbc::CLI::InternalStanza, :cask do
+  it "shows stanza of the Specified Cask" do
+    command = described_class.new("gpg", "with-gpg")
+    command.run
+    # TODO: check result
```
reitermarkus: Use this to check the output:

```ruby
expect {

}.to output("string").to_stdout
```

Also use curly braces for all `expect` calls so it reads nicer.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: Unfortunately this errors out:
`Martins-MacBook-Pro:~ martin$ brew cask _stanza version ableton-live`
`Error: undefined method 'to_a' for "9.7.4":Hbc::DSL::Version`
`    Did you mean?  to_yaml`
`             to_r`
`             to_f`
`              to_i`
`              to_s`
`              to_c`

Library/Homebrew/test/cask/cli/internal_stanza_spec.rb
Git-Jiro: Rubocop does not like curly braces:
`== /usr/local/Homebrew/Library/Homebrew/test/cask/cli/internal_stanza_spec.rb ==`
`C:  4: 12: Avoid using {...} for multi-line blocks.`

I also get the same offence if I 'brew style'  /usr/local/Homebrew/Library/Homebrew/test/cask/cli/fetch_spec.cli for example ...

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
reitermarkus: Ah, right. Ok, then make the second part `(value.respond_to?(:to_a) && value.to_a.empty?)`.

Library/Homebrew/test/cask/cli/internal_stanza_spec.rb
reitermarkus: Weird, it should be ignored for `*_spec.rb` files.

reitermarkus:
The `unless cask.respond_to?(stanza)` block can be removed, because this won't every be true, since we already checked if such a stanza exists by that point.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,10 +71,16 @@ def run
             next
           end
 
-          if artifact_name && !value.key?(artifact_name)
-            opoo "no such stanza '#{artifact_name}' on Cask '#{cask}'" unless quiet?
-            puts ""
-            next
+          if (value.nil? || value.is_a?(Array) && value.empty?) ||
+             (artifact_name && !value.key?(artifact_name))
```
reitermarkus: I think the `value.key?(artifact_name)` won't work anymore, since `cask.artifacts` now returns a `Set` of artifacts instead of a `Hash` of `name => [artifacts]`. So to check if there is an artifact with a given name, the logic has to be changed.

Library/Homebrew/test/cask/cli/internal_stanza_spec.rb
Git-Jiro: Yes, it is ignoring *_spec.rb files by default. I manually did a 'brew style /usr/local/Homebrew/Library/Homebrew/test/cask/cli/internal_stanza_spec.rb ' to specifically check the file.
OK, I will change the do ... end to { } so it reflects the way all the other spec files are written.

reitermarkus: Yeah, `brew style /usr/local/Homebrew/Library/Homebrew/test/cask/cli/internal_stanza_spec.rb` will assume that the Ruby file is a Formula, so it will run with a different configuration.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: I have a test case for this that seems to work fine ...

reitermarkus:
Can you look into the test failures?

Git-Jiro:
@reitermarkus  I already had a look at them. My Ruby-Foo is not strong enough to know what the problem is :-(. It seems to be related to the Ruby version. As far as I understand it, both CI tasks run the complete test-suite. The failing one uses a different Ruby version and has the "--coverage" flag set. I assume the "--coverage" flag is not the issue, because I can run the test-suite with "--coverage" on my machine without any issues.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,10 +65,18 @@ def run
             next
           end
 
-          if artifact_name && !value.key?(artifact_name)
-            opoo "no such stanza '#{artifact_name}' on Cask '#{cask}'" unless quiet?
-            puts ""
-            next
+          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
+             (value.respond_to?(:to_h) && value.to_h.empty?) ||
```
claui: The `Enumerable#to_h` method, which was introduced in Ruby 2.1, **will fail unless every element it contains is a two-element array.** This appears to be why `#to_h` currently fails for two of the test cases.

For example, it fails when `value` is a `Set` whose elements are `Artifact`s (and not pairs of something); see also @reitermarkus’s [comment](https://github.com/Homebrew/brew/pull/3258/commits/a605d74e6521181f3a0c2e46de95abadab01320f#r143356789).

Fortunately, you should be able to just lose that line altogether. It is redundant because whenever a hash is empty, its `#to_a` is also empty.

claui:
> It seems to be related to the Ruby version.
> As far as I understand it, both CI tasks run the complete test-suite.

@Git-Jiro The Linux CI job [excludes all macOS-specific tests](https://travis-ci.org/Homebrew/brew/jobs/285108126#L993). I think this is why it passes.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: @claui I removed the line as you suggested. The test is still failing on TravisCI and is still working fine on my machine. Any ideas what I can do to get the tests failing on my machine as well in order to speed up fixing this?

MikeMcQuaid: @Git-Jiro Run `brew tests --online` and `brew style`.

Git-Jiro: @MikeMcQuaid unfortunately the --online flag did not change the results: https://gist.github.com/Git-Jiro/2afc9133c700223fd67383030142f9ed

claui: @Git-Jiro Can you double-check that you have actually tapped into `caskroom/cask` on the system you use for testing? Maybe `brew untap caskroom/cask && brew tap caskroom/cask` for good measure? (This shouldn’t normally be necessary.)

Git-Jiro: Nope, tests are still fine on my system ...

claui: What happens if you drop a `raise` somewhere in the code under test?

Git-Jiro: If I drop a raise in my code under test, the raise works as expected:
https://gist.github.com/Git-Jiro/9254614151ac0a4809ed20a5e70b422d

claui: Now that you find the `raise` to work as expected, try moving the `raise` statement around in the code. Move it gradually near the `if stanza == artifacts` block (this is the region of the code where the test fails), and repeat this step until the `raise` ends up inside that block. If the `raise` stops working at some point, then find out why. Alternatively, use debug statements like `warn` to figure out what’s happening near the offending part of the code.

As you can see in the CI failures, the `if stanza == artifacts` block is one of the reasons why the tests are still failing. (This is most likely a side effect of the recent refactoring of the `#artifacts` method, which now returns a `Set` instead of a `Hash`.)

Also, try to stick with [your `brew tests` command line](https://gist.github.com/Git-Jiro/2afc9133c700223fd67383030142f9ed) instead of running `brew cask _stanza`.

Git-Jiro: Finally, I think I got all your hints about Sets vs. Hashes.
Hopefully the code works now ...

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -67,7 +67,7 @@ def run
 
           if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
              (value.respond_to?(:to_s) && value.to_s == "{}") ||
-             (artifact_name && !value.key?(artifact_name))
+             (artifact_name && !value.include?(artifact_name))
```
reitermarkus: Pretty sure that `value` will never include `artifact_name`. Did you check what is inside `value`?

Git-Jiro: Here is some example output of "value" when running "brew cask _stanza artifacts ableton-live" https://gist.github.com/Git-Jiro/6ee0421cf4ab1d59f008251760fac3a3 ...

reitermarkus: What if you run `brew cask _stanza app <cask>`?

Git-Jiro: This is the output of "brew cask _stanza app ableton-live" https://gist.github.com/Git-Jiro/66079ce487d44500207988eb54a6a4f2

Git-Jiro: I am still wondering why "brew tests" works on my machine and fails on Travis-CI.
As far as I can tell Travis-CI is using this bottled Ruby to run the tests: https://homebrew.bintray.com/bottles-portable/portable-ruby-2.3.3.leopard_64.bottle.1.tar.gz
As far as I can see my local Homebrew seems to use exactly the same bottled Ruby version: https://gist.github.com/Git-Jiro/30093a888c2c2ac929f195b2457cb6e6

claui: You mentioned that inserting a `raise` statement broke `brew cask _stanza`. Did it not break `brew tests`?

reitermarkus:
@Git-Jiro, can you rebase this? Maybe then you can reproduce the test error.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,10 +65,17 @@ def run
...
+            if artifact_name
+              thing = artifact_name
+            else
+              thing = stanza
+            end
+
+            raise CaskError, "no such stanza '#{thing}' on Cask '#{cask}'"
           end
 
           if stanza == :artifacts
```
reitermarkus: The test error is coming from this part here, because `value` is not a `Hash` anymore, the `v` variable will always be nil.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,10 +65,17 @@ def run
             next
           end
 
-          if artifact_name && !value.key?(artifact_name)
-            opoo "no such stanza '#{artifact_name}' on Cask '#{cask}'" unless quiet?
-            puts ""
-            next
+          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
+             (value.respond_to?(:to_s) && value.to_s == "{}") ||
```
reitermarkus: This does not apply to any case, because a hash responds to `to_a`, and will be handled by the check above.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,10 +65,17 @@ def run
...
+             (artifact_name && !value.include?(artifact_name))
+
+            if artifact_name
+              thing = artifact_name
```
reitermarkus: Can you find a better name than `thing`? Also, use a ternary for this.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: Yes, the raise breaks "brew tests" as well.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: OK, I understand now that Sets have to be handled differently than Hashes.
Yes, I will rebase and hopefully my local brew will behave the same way as "Travis" brew.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: As I said above, seems like I am lacking a bit of knowledge regarding Ruby Sets and Arrays, but I will get there ...

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: Ok!

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
Git-Jiro: @reitermarkus YES! Rebase did the trick! Now finally my local brew seems to behave like "Travis brew". This will help fixing the final issues tremendously! Thank you for pointing out this blunder of mine.

Git-Jiro:
@claui @reitermarkus Thanks to you guys the code seems to work now with SortedSet!

However the output changed due to my changes. I do not know if some other part of Homebrew relies on the format of the output. Therefore one of my tests still breaks because of that.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -24,6 +24,9 @@ class InternalStanza < AbstractInternalCommand
       ARTIFACTS =
         DSL::ORDINARY_ARTIFACT_CLASSES.map(&:dsl_key) +
         DSL::ARTIFACT_BLOCK_CLASSES.map(&:dsl_key)
+      ARTIFACTS_CLASSES =
```
reitermarkus: This constant doesn't seem to be used anywhere.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,26 +68,29 @@ def run
...
+                result = v
+                break
               end
-            ]
+            end
+            value = result
+          end
+
+          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
+             (value.respond_to?(:to_s) && value.to_s == "{}") # ||
```
reitermarkus: Remove this check.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,26 +68,29 @@ def run
...
-          if artifact_name && !value.key?(artifact_name)
-            opoo "no such stanza '#{artifact_name}' on Cask '#{cask}'" unless quiet?
-            puts ""
-            next
+          if stanza == :artifacts
+            result = {}
+            value.each do |e|
+              result[e.class.dsl_key] = e.summarize
+            end
+            value = result
```
reitermarkus: You can replace all of this with

```ruby
value = Hash[value.map { |v| [v.class.dsl_key, v] }]
```

It shouldn't use `#summarize` because otherwise it will not respond to `#to_a` or `#to_h` anymore below.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,26 +68,29 @@ def run
...
+            result = nil
+            value.each do |k, v|
+              if k == artifact_name
+                result = v
+                break
               end
-            ]
+            end
+            value = result
+          end
```
reitermarkus: Since `value` is a hash at this point, you can just do 

```ruby
if artifact_name
  value = value[artifact_name]
end
```

Also, I think this should be inside the `if stanza == :artifacts` block.

Git-Jiro:
@reitermarkus @claui What should I do not make the last failing test pass?

reitermarkus:
I think you need to map the values in the hash with `to_s`.

Git-Jiro:
@reitermarkus @claui The output is closer to the original now. However it still is not the same. Should I adapt the test to the new output? Or is there some script or other command which expects a certain output format of `brew cask _stanza` which will break if the format changes?

claui:
@Git-Jiro If I recall correctly, internal commands like `_stanza` are intended to be called by contributors for their convenience. Their output can’t be part of any contract. If anything relies on it, that would be on the caller’s own risk.

Pinging @vitorgalvao for a second opinion.

Library/Homebrew/test/cask/cli/internal_stanza_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    expect {
+      command.run
+    }.to raise_error(/no such stanza/)
+  end
+
+  it "shows all artifact stanzas when using 'artifacts' keyword" do
+    command = described_class.new("artifacts", "with-gpg")
+    expect {
+      command.run
+    }.to output("{:app=>[[\"Caffeine.app\"]]}\n").to_stdout
```
claui: As you already said, this example currently fails.
If you want to avoid escaping, consider a `%` string literal:

```
}.to output(%Q({:app=>"Caffeine.app (App)"}\n)).to_stdout
```

Or, if you’re bold:

```
}.to output(/Caffeine.app/).to_stdout
```

vitorgalvao:
> internal commands like `_stanza` are intended to be called by contributors (…) If anything relies on it, that would be on the caller’s own risk.

That is correct. The only notable exception would be [`_appcast_checkpoint`](https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/appcast.md), but even that is to be used when contributing, not in normal usage.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -43,6 +43,9 @@ def initialize(*)
         @stanza = args.shift.to_sym
 
         @format = :to_yaml if yaml?
+
+        return if DSL::DSL_METHODS.include?(stanza)
+        raise ArgumentError, "Illegal stanza: '#{stanza}'"
```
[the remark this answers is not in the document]
Git-Jiro: How do you like this one? `"Unknown/unsupported stanza: '#{stanza}' Check cask language reference for list of supported stanzas."`

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -43,6 +43,10 @@ def initialize(*)
         @stanza = args.shift.to_sym
 
         @format = :to_yaml if yaml?
+
+        return if DSL::DSL_METHODS.include?(stanza)
+        raise ArgumentError,
+          "Unknown/unsupported stanza: '#{stanza}' Check cask reference for supported stanzas."
```
claui: This is an example of an excellent error message. Not only does it clearly state what the problem is, it also tells the user exactly how to fix it. 👍 

For longer messages like this one, I suggest to use the `\` line continuation operator to keep your code within the 80-character boundary:

```
raise ArgumentError,
  "Unknown/unsupported stanza: '#{stanza}'. " \
  "Check cask reference for supported stanzas."
```

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -43,6 +43,11 @@ def initialize(*)
         @stanza = args.shift.to_sym
 
         @format = :to_yaml if yaml?
+
+        return if DSL::DSL_METHODS.include?(stanza)
+        raise ArgumentError,
+          "Unknown/unsupported stanza: '#{stanza}' Check cask reference"\
+          " for supported stanzas."
```
reitermarkus: I'd make this two lines.

```ruby
<<~EOS
  Unknown/unsupported stanza: '#{stanza}' 
  Check the Cask reference for supported stanzas.
EOS
```

Git-Jiro:
@vitorgalvao @claui @reitermarkus Anything else I could do to get this ready for merging?

claui:
> Anything else I could do to get this ready for merging?

@Git-Jiro Looks fine to me! Thanks a lot for your work and for bearing with change requests.  
I feel your PR is a step in the right direction to get #29289 done.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -43,6 +43,13 @@ def initialize(*)
         @stanza = args.shift.to_sym
 
         @format = :to_yaml if yaml?
+
+        return if DSL::DSL_METHODS.include?(stanza)
+        raise ArgumentError,
+          <<~EOS
+            Unknown/unsupported stanza: '#{stanza}'
+            Check cask reference for supported stanzas.
```
reitermarkus: Uppercase “Cask” here.

Library/Homebrew/cask/lib/hbc/cli/internal_stanza.rb
```diff
@@ -68,26 +69,15 @@ def run
...
-                [k, v]
-              end
-            ]
+            value = Hash[value.map { |v| [v.class.dsl_key, v.to_s] }]
+            value = value[artifact_name] if artifact_name
+          end
 
-            value = value.fetch(artifact_name) if artifact_name
+          if value.nil? || (value.respond_to?(:to_a) && value.to_a.empty?) ||
+             (value.respond_to?(:to_s) && value.to_s == "{}")
```
reitermarkus: Remove this check.

Git-Jiro: @reitermarkus The whole check? If I remove the whole check 3 tests will break. If I remove the "{}" part of the check, the check regarding "depends_on" will break.
Please clarify your thoughts about this.

reitermarkus: No, just this line with the string comparison.

Git-Jiro: OK, but how should I detect absence of the "depends_on" stanza instead in order to not break the corresponding test?

reitermarkus: When do you expect the string to be `"{}"`? A `Hash` responds to `to_a`, so it will be handled by the previous test.

Git-Jiro: I expect this when querying for the "depends_on" stanza. This returns an Hbc::DSL::DependsOn object. As far as I can see this object does not respond to "to_a". It only responds to "to_s" with a string of "{}" when there is no "depends_on" stanza in the cask.

reitermarkus: Once https://github.com/Homebrew/brew/pull/3432 is merged, `DependsOn` will respond to `to_a`.

Git-Jiro: Oh! Magical interventions ... ;-)

Git-Jiro:
@reitermarkus @claui I rebased my PR to make my tests work with #3432

claui:
Thanks a bunch @Git-Jiro! This is starting to look great.

In the meantime, I have noticed that this PR **introduces a regression** for two usage modes of `_stanza` which we may have overlooked in our discussion earlier. The problem is that there are two possible ways to call the `_stanza` command as of today:

1. `brew cask _stanza STANZA TOKEN […]`
2. `brew cask _stanza STANZA` without further arguments

With the first usage mode, I’ve noticed that the command now errors out early when we call it with more than one Cask token (note that the positional order of the arguments now influences the result):

```
$ brew cask _stanza depends_on docker dropbox
{:macos=>[[:>=, #<Gem::Version "10.10">]]}
Error: no such stanza 'depends_on' on Cask 'dropbox'
```

```
$ brew cask _stanza depends_on dropbox docker
Error: no such stanza 'depends_on' on Cask 'dropbox'
```

The second usage mode (i. e. no Cask given at all) seems to error out whenever we invoke it with any stanza relevant to this PR:

```
$ brew cask _stanza depends_on
Error: no such stanza 'depends_on' on Cask '010-editor'
```

I feel this is not exactly an issue with your implementation; I believe **we simply forgot to address these existing usage modes** when discussing the requirements in [issue #29289](https://github.com/caskroom/homebrew-cask/issues/29289). (Who knows, maybe no one except me has ever used the `_stanza` command like that?)

In my opinion, there are legitimate use cases for calling `_stanza` with no Cask tokens, or with multiple tokens; that said, we may want to re-consider whether to deprecate these usage modes, or keep supporting them.

**I suggest that we resume the discussion in [#29289](https://github.com/caskroom/homebrew-cask/issues/29289),** and label this PR do-not-merge until a decision is made.

claui:
@Git-Jiro The discussion has reached consensus today in that we will no longer be supporting any usage of `_stanza` except for the single-token form.

In my opinion, this means we can safely accept that this PR will break the other usage modes for now; guarding against those usage modes should be a topic for another PR.

I suggest that the do-not-merge label now be removed, and that this PR be merged. Pinging @reitermarkus to confirm.

reitermarkus:
Thanks @Git-Jiro, great work here!

PR #3259, opened 2017-10-03
Remove `UnderscoreSupportingURI`.

reitermarkus:

-----

Not needed anymore on Ruby 2.3.3.


claui:
I’m a bit on the fence here. While it’s true that the issue has been [fixed since Ruby 2.2](https://bugs.ruby-lang.org/issues/8241), and it makes perfect sense to ditch `UnderscoreSupportingURI`, I’m not sure if we must really lose the spec.

After all, we rely on the new behavior, and the spec should reflect this. Basically, it would act as a canary, which would give alarm in case HBC inadvertently fails to switch to the vendored Ruby 2.3.3, or falls back to Ruby 2.0 for some other reason.

On the other hand, testing an upstream API should be none of our business.

Would it hurt to keep the spec, and have it point to `URI`? I’m torn tbh.

MikeMcQuaid:
> Basically, it would act as a canary, which would give alarm in case HBC inadvertently fails to switch to the vendored Ruby 2.3.3, or falls back to Ruby 2.0 for some other reason.

A bunch of other things will fail before that point. I'm 👍 to deleting this if it's an upstream API. Keeping stuff like it around bloats the test suite, a bit.

claui:
> A bunch of other things will fail before that point.

That’s a good point, and comforting to know.
In that case, LGTM as is, too.

PR #3262, opened 2017-10-04
Add version detection support for php URL

kabel:
Pulls a semver from anywhere in the URL (not just stem).


MikeMcQuaid:
Thanks for your contribution to Homebrew/brew, @kabel! Without people like you submitting PRs we couldn't run this project. You rock!

PR #3264, opened 2017-10-04
Update my name in Kickstarter-Supporters, add .mailmap

mgol:
- ~~Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).~~

-----

### DIFF
--- a/docs/Kickstarter-Supporters.md
+++ b/docs/Kickstarter-Supporters.md
@@ -379,7 +379,7 @@ These wonderful people supported our Kickstarter by giving us £10 or more:
 * [Andrew Brown](http://pvalu.es)
 * [Bethany Sumner](http://www.bethanysumner.com/)
 * [Orta](http://orta.io)
-* [Michał Gołębiowski](https://github.com/mgol)
+* [Michał Gołębiowski-Owczarek](https://github.com/mgol)
 * [Adam C. Foltzer](http://www.acfoltzer.net/)
 * [Steve Hiemstra](https://www.speg.com)
 * [Anton Sipos](http://www.softwarefuturism.com)


.mailmap
```diff
@@ -0,0 +1 @@
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
```
MikeMcQuaid: Please remove this file.

.gitignore
```diff
@@ -53,6 +53,7 @@
 
 # Unignore our root-level metadata files.
 !/.gitignore
+!/.mailmap
```
MikeMcQuaid: Please remove this change.

.mailmap
mgol: Without this entry in this file Git will recognize me as two different people. For example:
```
$ git shortlog -nse | grep m.goleb
     1	Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
     1	Michał Gołębiowski <m.goleb@gmail.com>
```

MikeMcQuaid: Sorry, we're not adding a file in our repo just for that.

mgol: OK, I'll remove it. Note that such things may be more important to other people, e.g. trans folks after a name change who wouldn't appreciate seeing their old name when doing `git blame` etc.

MikeMcQuaid: Unfortunately Git uses that information as part of a SHA1 hash so there's no real way to work around that. We allow people to change their email and GitHub allows them to change their username.

mgol: It's true author information is kept in the commits and hence immutable (and that's what you'll see if you invoke `git show COMMIT_HASH`). For many operations, though, Git reads what's in the commit but then passes it through the `.mailmap` before returning to the user; one such example is `git blame`. When I invoke `git blame` on a file in the jQuery repo that I've last modified long ago, it marks the lines with my current name, not the one under which it was committed.

GitHub uses GitHub accounts to present some of this info so changing a GitHub username will change what you see in the UI. External apps, understandably, can't do it so they use what Git reports to them directly.

For example, if I use the WebStorm "annotate" feature which shows a sidebar left to the code with info about the author of each line, with a `.mailmap` present in the repo it always shows my current name, not the old one - simply because that's what `git blame` reports by itself and that's what WebStorm uses under the hood.

MikeMcQuaid: Yep, I'm familiar with that. For your own use I'd recommend maintaining your own `.mailmap` file.

mgol: PR updated.

MikeMcQuaid: @mgol also for what it’s worth I took my wife’s surname when I got married so there’s several years of Homebrew commits by me with my old name.

mgol: Thanks for the global on-machine `.mailmap` tip, I didn't know that's possible!

PR #3270, opened 2017-10-05
Add depends_on :linux and :macos

sjackman:

-----

Allow a formula to declare whether it depends on macOS or Linux.


Library/Homebrew/dependency_collector.rb
```diff
@@ -108,7 +108,8 @@ def parse_symbol_spec(spec, tags)
     case spec
     when :x11        then X11Requirement.new(spec.to_s, tags)
     when :xcode      then XcodeRequirement.new(tags)
-    when :macos      then MinimumMacOSRequirement.new(tags)
+    when :linux      then LinuxRequirement.new
+    when :macos      then tags.empty? ? MacOSRequirement.new : MinimumMacOSRequirement.new(tags)
```
MikeMcQuaid: make `MacOSRequirement.new` handle the `tags.empty?` case instead of having the meaning overloaded.

ilovezfs:
How do you specify _both_ that macOS is required _and_ that a minimum version is required? My understanding is `depends_on :macos => :el_capitan` is a no-op on Linux.

sjackman:
To specify that both that macOS is required and that a minimum version is required:
```
depends_on :macos
depends_on MinimumMacOSRequirement => :el_capitan
```

ilovezfs:
That seems like pretty yucky syntax to me.

Library/Homebrew/dependency_collector.rb
sjackman: Done.

Library/Homebrew/requirements/minimum_macos_requirement.rb
```diff
@@ -3,18 +3,32 @@
...
-  def initialize(tags)
-    @version = MacOS::Version.from_symbol(tags.first)
+  def initialize(tags = [])
+    @version = MacOS::Version.from_symbol tags.first unless tags.empty?
     super
   end
 
-  satisfy(build_env: false) { MacOS.version >= @version }
+  satisfy(build_env: false) do
+    if @version
```
MikeMcQuaid: `next false if !OS.mac? || !@version`

Library/Homebrew/requirements/minimum_macos_requirement.rb
```diff
@@ -3,18 +3,32 @@
 class MinimumMacOSRequirement < Requirement
```
MikeMcQuaid: Might be worth renaming this now to be consistent, what do you think?

Library/Homebrew/requirements/minimum_macos_requirement.rb
MikeMcQuaid: May be worth taking the same approach below, too.

Library/Homebrew/requirements/minimum_macos_requirement.rb
sjackman: The name of the class is a component of the DSL. The name `MinimumMacOSRequirement` is used by two formulae in core:
```
nordugrid-arc.rb:  depends_on MinimumMacOSRequirement => :yosemite
phantomjs.rb:  depends_on MinimumMacOSRequirement => :lion
```

MikeMcQuaid: We can add it to `compat` and update them to use `depends_on :macos`

sjackman: Got it. Would you like to suggest a name?

MikeMcQuaid: `MacOSRequirement`?

Library/Homebrew/requirements/minimum_macos_requirement.rb
sjackman: Done. Thanks for the suggestion, Mike. The logic needed tweaking. When `OS.mac?` is `true` and `@version` is `nil`, the result should be `true`.

Library/Homebrew/requirements/minimum_macos_requirement.rb
sjackman: I've renamed `MinimumMacOSRequirement` to `MacOSRequirement` and added `MinimumMacOSRequirement` to `compat`.

Library/Homebrew/compat/requirements/minimum_macos_requirement.rb
```diff
@@ -0,0 +1,3 @@
+require "requirements/macos_requirement"
```
MikeMcQuaid: Stick this file's contents in `homebrew/Library/Homebrew/compat/requirements.rb`

Library/Homebrew/requirements/macos_requirement.rb
```diff
@@ -0,0 +1,26 @@
+require "requirement"
+
+class MacOSRequirement < Requirement
+  fatal true
+
+  def initialize(tags = [])
+    @version = MacOS::Version.from_symbol tags.first unless tags.empty?
```
MikeMcQuaid: `MacOS::Version.from_symbol(tags.first)` just to aid visual parsing.

Library/Homebrew/requirements/macos_requirement.rb
```diff
@@ -0,0 +1,26 @@
...
+    super
+  end
+
+  satisfy(build_env: false) do
+    next false if !OS.mac? && !@version
+    next true if !OS.mac? || !@version
```
MikeMcQuaid: This is a little hard to parse above. Can you maybe make it `if`/`else`s? This may contradict what I said earlier: sorry!

Library/Homebrew/requirements/macos_requirement.rb
```diff
@@ -0,0 +1,26 @@
...
+    MacOS.version >= @version
+  end
+
+  def message
+    return "macOS is required." if !OS.mac? || !@version
```
MikeMcQuaid: `if !OS.mac? || !@version` could be a utility method, I think, it's repeated 3 times.

Library/Homebrew/requirements/macos_requirement.rb
sjackman: Factoring out the `minimum_version?` utility method also helped make this logic more clear.

Library/Homebrew/requirements/macos_requirement.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class MacOSRequirement < Requirement
+  fatal true
+
+  def initialize(tags = [])
+    @version = MacOS::Version.from_symbol(tags.first) unless tags.empty?
+    super
+  end
+
+  def minimum_version?
```
MikeMcQuaid: `minimum_version_provided` or `passed` or `specified` something would be nicer to make it clear that this doesn't mean the minimum version is met.

MikeMcQuaid:
Tests failing on Linux (shouldn't assume `MacOS` is specified there): https://travis-ci.org/Homebrew/brew/jobs/285144949

sjackman:
Fixed up.

MikeMcQuaid:
@sjackman Still got 🔴.

sjackman:
😬

sjackman:
Good to merge?

Library/Homebrew/requirements/macos_requirement.rb
```diff
@@ -0,0 +1,29 @@
...
+    @version = MacOS::Version.from_symbol(tags.first) unless tags.empty?
+    super
+  end
+
+  def minimum_version_specified?
+    OS.mac? && @version
+  end
+
+  satisfy(build_env: false) do
+    next OS.mac? || @version unless minimum_version_specified?
```
MikeMcQuaid: Sorry, last nite: can you split this into two `next` checks and return a boolean value?

sjackman: Done.

MikeMcQuaid:
Thanks @sjackman and sorry for the nit picking!

sjackman:

PR #3278, opened 2017-10-06
Clean up code style and remove `.rubocop_todo.yml`.

reitermarkus:


Library/Homebrew/.rubocop.yml
```diff
@@ -26,9 +25,26 @@ Lint/NestedMethodDefinition:
 Lint/ParenthesesAsGroupedExpression:
   Enabled: true
 
+Metrics/BlockNesting:
+  Max: 5
+
+Metrics/ModuleLength:
+  Max: 360
```
MikeMcQuaid: Good idea to just setting these to high values. Mind doing that for line length, too? Would be good to get it down eventually but at least stop it growing for now.

Library/Homebrew/cmd/tap.rb
```diff
@@ -54,8 +54,7 @@ def tap
                     quiet: ARGV.quieter?
       rescue TapRemoteMismatchError => e
         odie e
-      rescue TapAlreadyTappedError, TapAlreadyUnshallowError
-        # Do nothing.
+      rescue TapAlreadyTappedError, TapAlreadyUnshallowError # rubocop:disable Lint/HandleExceptions
```
MikeMcQuaid: Not worth keeping the comment and returning `nil` instead?

Library/Homebrew/extend/pathname.rb
```diff
@@ -186,7 +186,7 @@ def atomic_write(content)
       begin
         tf.chown(uid, gid)
         tf.chmod(old_stat.mode)
-      rescue Errno::EPERM
+      rescue Errno::EPERM # rubocop:disable Lint/HandleExceptions
```
MikeMcQuaid: Worth returning `nil` here and skipping the `disable (and everywhere below)?

ilovezfs:
@reitermarkus maybe should rebase now that #3284 is merged?

reitermarkus:
I'll leave the comments for now. I think it's more obvious to see the difference between explicitly returning `nil` and returning it just to silence RuboCop.

PR #3279, opened 2017-10-07
Support `brew pull --bottle` for taps in organizations other than Homebrew

sjackman:

-----

Infer the name of the Bintray organization and test-bot GitHub user if either is not provided.
If the GitHub organization is Homebrew, infer that the Bintray organization is homebrew, and the GitHub test-bot user is BrewTestBot.
Otherwise, if the GitHub organization is Foo, infer that the Bintray organization is foo, and the GitHub test-bot user is FooTestBot.
Add options `--bintray-org` and `--test-bot-user` to override these inferences.

### DIFF
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -58,6 +58,18 @@ require "tap"
 require "version"
 require "pkg_version"
 
+module GitHub
+  module_function
+
+  # Return the corresponding test-bot user name for the given GitHub organization.
+  def test_bot_user(user)
+    test_bot = ARGV.value "test-bot-user"
+    return test_bot if test_bot
+    return "BrewTestBot" if user.casecmp("homebrew").zero?
+    "#{user.capitalize}TestBot"
+  end
+end
+
 module Homebrew
   module_function
 
@@ -231,7 +243,7 @@ module Homebrew
           url
         else
           bottle_branch = "pull-bottle-#{issue}"
-          "https://github.com/#{test_bot_user user}/homebrew-#{tap.repo}/compare/#{user}:master...pr-#{issue}"
+          "https://github.com/#{GitHub.test_bot_user user}/homebrew-#{tap.repo}/compare/#{user}:master...pr-#{issue}"
         end
 
         curl "--silent", "--fail", "--output", "/dev/null", "--head", bottle_commit_url
@@ -257,13 +269,6 @@ module Homebrew
     # Verify bintray publishing after all patches have been applied
     bintray_published_formulae.uniq!
     verify_bintray_published(bintray_published_formulae)
-  end
-
-  def test_bot_user(user)
-    test_bot = ARGV.value "test-bot-user"
-    return test_bot if test_bot
-    return "BrewTestBot" if user.casecmp("homebrew").zero?
-    "#{user.capitalize}TestBot"
   end
 
   def force_utf8!(str)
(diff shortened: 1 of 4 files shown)


Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -253,11 +259,18 @@ def pull
     verify_bintray_published(bintray_published_formulae)
   end
 
+  def test_bot_user(user)
+    test_bot = ARGV.value "test-bot-user"
+    return test_bot if test_bot
+    return "BrewTestBot" if user.casecmp("homebrew").zero?
```
MikeMcQuaid: Not sure I understand `user.casecmp("homebrew").zero?`; what's it doing?

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -253,11 +259,18 @@ def pull
...
+    "#{user.capitalize}TestBot"
```
MikeMcQuaid: I wonder if this is a bit magic and we should just require this argument to get passed. Thoughts?

Library/Homebrew/dev-cmd/pull.rb
sjackman: If the GitHub organization that the PR is submitted to is `Homebrew` (case insensitive comparison), then the GitHub test-bot user is `BrewTestBot`.

Library/Homebrew/dev-cmd/pull.rb
sjackman: For a command run primarily by CI, I'd agree. Since `brew pull --bottle` is run interactively primarily, a reasonable default value is particularly helpful. How to name this test-bot user is arbitrary and may as well be consistent across taps. An alternative would be to store this value in the `.gitconfig` of each tap, or possibly in a dot file that's committed to the tap.

MikeMcQuaid: I think that might an argument for accepting environment variables for some of these options instead but I'm fine as-is, too.

Library/Homebrew/dev-cmd/pull.rb
MikeMcQuaid: Definitely need a case-insensitive comparison here? If so, `user.downcase == "homebrew"` is more typical style.

Library/Homebrew/dev-cmd/pull.rb
sjackman: They'd need to set/unset/change this environment variable depending on which tap organization they're currently working on, if they contribute to more than one tap organization. My preference is as-is.

MikeMcQuaid:
@sjackman Still 🔴

sjackman:
Ran `brew man`

MikeMcQuaid:
Still 🔴 (new Rubocop rules)

sjackman:
```
Library/Homebrew/dev-cmd/pull.rb:61:1: C: Module has too many lines. [364/360]
module Homebrew ...
```
https://travis-ci.org/Homebrew/brew/jobs/285897358#L843

MikeMcQuaid:
> What's the recommended way of dealing with Module has too many lines? I could move the test_bot_user utility function that I created out of the module.

Yeh, consider splitting some logic into another module in the same file.

sjackman:
Done.

sjackman:
Good to go? Or shall I write a test to address the failed `codecov/patch` coverage failure?

MikeMcQuaid:
Thanks again @sjackman!

PR #3287, opened 2017-10-08
audit: tweak assert/refute_predicate audit

DomT4:
-----

Addresses some of the issues I raised in https://github.com/Homebrew/homebrew-core/pull/19127#issue-263566817.

Would still like to work out a way to check `File.exist?`-style language inside the `test do` block exclusively but for now it's better to be too conservative on the audit than to be overly-zealous and flag too much.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -967,12 +967,12 @@ class FormulaAuditor
       problem "Use `assert_match` instead of `assert ...include?`"
     end
 
-    if line =~ /(assert File\.exist\?|File\.exist\?)/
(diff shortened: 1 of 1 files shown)


ilovezfs:
Thanks, Mr. T!

MikeMcQuaid:
@DomT4 It's worth perhaps chatting with @GauthamGoli some time about getting this into a RuboCop test which would be able to do things such as scope it to `test do` only and ignore weird use of whitespace.

DomT4:
@MikeMcQuaid Aye, probably. I tend to find Rubocop's syntax pretty grim to work with, but Rubocop seems to be the direction Homebrew is moving in so I guess I'll need to wrap my head around it more contently at some point.

PR #3290, opened 2017-10-09
Whitelist keg-only `ncurses` for `brew audit`

claui:
With regards to `brew audit`, this PR whitelists the `ncurses` formula, which is tagged `keg_only :provided_by_macos`.

This tag had caused `brew audit --new-formula sc-im` to fail with the following error message:

> Dependency 'ncurses' may be unnecessary as it is provided by
> macOS; try to build this formula without it.

However, the [new formula](https://github.com/Homebrew/homebrew-core/pull/14067) `sc-im` depends on the GNU ncurses library in a recent version, just like the one Homebrew provides as a separate formula.

Because the macOS-provided ncurses is no drop-in replacement, this PR whitelists ncurses in `audit.rb`.

See also: https://github.com/Homebrew/homebrew-core/pull/14067#issuecomment-335046151

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -450,7 +450,7 @@ class FormulaAuditor
         end
 
         if @new_formula && dep_f.keg_only_reason &&
-           !["openssl", "apr", "apr-util"].include?(dep.name) &&
+           !["apr", "apr-util", "ncurses", "openssl"].include?(dep.name) &&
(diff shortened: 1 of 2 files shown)


ilovezfs:
This is a --new-formula only audit and one we'd prefer to keep, but thanks nonetheless for the PR.

ilovezfs:
@claui if you'd like to put those new tests you wrote in another PR just for expanding the test coverage of this case, that would be cool :)

PR #3291, opened 2017-10-09
 lines_cop: add ENV.universal_binary audit exemption for wine

JCount:

-----
add ENV.universal_binary audit exception for wine and corresponding test coverage

### DIFF
--- a/Library/Homebrew/test/rubocops/lines_cop_spec.rb
+++ b/Library/Homebrew/test/rubocops/lines_cop_spec.rb
@@ -419,7 +419,22 @@ describe RuboCop::Cop::FormulaAudit::Miscellaneous do
       end
     end
 
-    it "with ENV.universal_binary" do
+    it "with ENV.universal_binary exempted formula" do
+      source = <<-EOS.undent
+        class Wine < Formula
+          desc "foo"
+          url 'http://example.com/foo-1.0.tgz'
+          if build?
+             ENV.universal_binary
+          end
+        end
+      EOS
+
+      inspect_source(source, "/homebrew-core/Formula/wine.rb")
+      expect(cop.offenses).to eq([])
+    end
+
+    it "with ENV.x11" do
       source = <<-EOS.undent
         class Foo < Formula
           desc "foo"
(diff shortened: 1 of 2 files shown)


JCount:
cc @ilovezfs

Library/Homebrew/test/rubocops/lines_cop_spec.rb
```diff
@@ -419,7 +419,22 @@ class Foo < Formula
       end
     end
 
-    it "with ENV.universal_binary" do
+    it "with ENV.universal_binary exempted formula" do
```
reitermarkus: Can you make this into a proper sentence? Thanks. 😉

JCount: I was only matching the one above it. So, inconsistency or proper sentences?

reitermarkus: Ideally proper sentences for all, or `specify` instead of `it`.

JCount: The full fragment is: `' When auditing formula with ENV.universal_binary exempted formula'` any suggestions?

JCount: (I am referring to the line https://github.com/Homebrew/brew/pull/3291/files#diff-9349b511bf2b9967d30735fbb26db7b7R383)

JCount:
@reitermarkus Is something like this what you had in mind?

JCount:
@reitermarkus gentle ping on this

reitermarkus:
[Something like this](https://github.com/Homebrew/brew/pull/3294#discussion_r144666636) is what I had in mind.

JCount:
Since this has the context `When auditing formulae` prepended to both tests, making the wording following the `it` into complete sentences, would, in my mind, only serve to make things less intelligible.

ilovezfs:
`it "is as quiet as a church mouse when there is an exemption to the deprecation of ENV.universal_binary" do`

reitermarkus:
> …  in my mind, only serve to make things less intelligible.

It should be possible to make it more intelligible than “when auditing formulae it with a build.universal? exemption”.

Also, the context is pretty redundant, so this could actually be removed, and we'd just have `it "does not report build.universal? offenses for Wine" do`, which would be pretty clear to me.

PR #3293, opened 2017-10-09
audit: more *_predicate nudging/tweaks

DomT4:
-----

CC @ilovezfs.

https://github.com/Homebrew/homebrew-core/pull/19213#issuecomment-335160034

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -974,8 +974,8 @@ class FormulaAuditor
       problem "Use `assert_predicate <path_to_file>, :exist?` instead of `#{Regexp.last_match(1)}`"
     end
 
-    if line =~ /assert !File\.exist\?/
-      problem "Use `refute_predicate <path_to_file>, :exist?` instead of `assert !File.exist?`"
+    if line =~ /(assert !File\.exist\?|assert !\(.*\)\.exist\?)/
+      problem "Use `refute_predicate <path_to_file>, :exist?` instead of `#{Regexp.last_match(1)}`"
     end
 
     return unless @strict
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -974,8 +974,16 @@ def line_problems(line, _lineno)
...
-      problem "Use `refute_predicate <path_to_file>, :exist?` instead of `assert !File.exist?`"
+    if line =~ /(assert !File\.exist\?|assert !\(.*\)\.exist\?)/
+      problem "Use `refute_predicate <path_to_file>, :exist?` instead of `#{Regexp.last_match(1)}`"
+    end
+
+    if line =~ /(assert File\.executable\?|assert \(.*\)\.executable\?)/
+      problem "Use `assert_predicate <path_to_file>, :executable?` instead of `#{Regexp.last_match(1)}`"
+    end
+
+    if line =~ /(assert !File\.executable\?|assert !\(.*\)\.executable\?)/
```
DomT4: We might not need this line really, because why would you logically check something _was not_ executable? Thoughts welcome.

MikeMcQuaid:
Given https://github.com/Homebrew/brew/pull/3313 I think this should probably be integrated in there instead.

ilovezfs:
Given it seems to be working fine I'm not sure why it needs to be made fancier and block on that happening.

MikeMcQuaid:
Ok then. Once https://github.com/Homebrew/brew/pull/3313 goes in, though, future changes will need to be to the Rubocop.

DomT4:
Thanks for merging! I was happy either way. At some point I'm going to have to sit down & wrap my head around the Rubocop format for everything, but I'll take being able to put that off for a little longer 🙈.

PR #3296, opened 2017-10-09
Refactor `MacOS::languages`.

reitermarkus:


Library/Homebrew/os/mac.rb
```diff
@@ -47,19 +47,15 @@ def cat
     end
 
     def languages
-      return @languages unless @languages.nil?
+      return @languages if defined?(@languages)
```
MikeMcQuaid: The `@languages ||= begin` pattern I think is generally nicest here but don't feel super strongly.

Library/Homebrew/os/mac.rb
```diff
@@ -47,19 +47,15 @@ def cat
...
-      if ENV["HOMEBREW_LANGUAGES"]
-        @languages = ENV["HOMEBREW_LANGUAGES"].split(",") + @languages
-      end
-
-      if ARGV.value("language")
-        @languages = ARGV.value("language").split(",") + @languages
-      end
-
-      @languages = @languages.uniq
+      @languages = [
```
MikeMcQuaid: `@languages.prepend` or something feels nicer?

reitermarkus: Doesn't really feel nicer to me.

MikeMcQuaid: It feels a little weird to be doing a splat on `@languages` inside it's own assignment.

reitermarkus: It's now all in one assignment.

PR #3298, opened 2017-10-10
pull: fix bintray verification failing due to redirection

vszakats:

-----

When doing a `brew pull --bottle <n>` for the second time, the command tries to verify if the uploaded package is there and correct. This failed in my case with an error:
```
Verifying bottle: caddy-0.10.10.high_sierra.bottle.tar.gz
Error: Failed to find published caddy bottle at https://homebrew.bintray.com/bottles/caddy-0.10.10.high_sierra.bottle.tar.gz (302 )!
```
What happens here is that `brew pull` is attempting a pre-check of the bottle URL (before starting the full download), by using the `Net::HTTP` Ruby class. In this particular case though — and this well may depend on geolocation — the URL returns a **302 redirect** (to a valid CloudFront URL). 302 is then considered and error and the command bails out without continuing to the curl download to do the actual verification.

`Net::HTTP` doesn't have an option to follow redirects automatically and 302 is _likely_ satisfying the goal of the pre-check (its stated goal is to wait in a loop till the upload finishes to propagate/publish upstream), so this patch simply accepts `302` as a success and let it proceed to the actual download.

### DIFF
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -560,7 +560,7 @@ module Homebrew
             req = Net::HTTP::Head.new bottle_info.url
             req.initialize_http_header "User-Agent" => HOMEBREW_USER_AGENT_RUBY
             res = http.request req
-            break if res.is_a?(Net::HTTPSuccess)
+            break if res.is_a?(Net::HTTPSuccess) || res.code == "302"
 
             unless res.is_a?(Net::HTTPClientError)
               raise "Failed to find published #{f} bottle at #{url} (#{res.code} #{res.message})!"


iMichka:
See #3134, where I tried to fix the same problem.

vszakats:
Thanks @iMichka, I haven't seen it.

Mine is the "lazy fix" though, yours is the complete package with following redirects.

FWIW in my case `302` happened _right after_ the initial upload and on macOS. I think the point is that such may happen anytime on any OS since it's up to Bintray to serve its content this way whenever it decides to.

IMO your original PR should be reopened and this closed in favour of it.

ilovezfs:
@vszakats if both solve the problem, why would you prefer a `+39 −20` change over a `+1 −1` change?

vszakats:
@ilovezfs Following the redirect is the canonical/future-proof way, though if the short fix is acceptable for everyone I for sure don't mind using it instead. It does fix the problem for now and may be extended later if it turns out a valid redirect would still not guarantee a propagated upload.

ilovezfs:
@vszakats In the 302 case, with your fix, it still ends up downloading the file, right?

vszakats:
@ilovezfs: It does, yes.

PR #3301, opened 2017-10-10
readall: by default, process homebrew/core tap only

maxim-belkin:

-----

There is a discrepancy between the description in `cmd/readall.rb` and what it actually does:
currently, `brew readall` reads all taps, not only the `homebrew/core` one. 

Steps to see the effect
1. Add `ohai tap` before `Homebrew.failed = true unless Readall.valid_tap?(tap, options)` in the `taps.each` loop in `Library/Homebrew/cmd/readall.rb`
2. Call `brew readall`

BEFORE (with the above debug information):
```
$ brew readall
==> jmhobbs/parrot
==> linuxbrew/developer
==> homebrew/science
==> homebrew/core
==> homebrew/dev-tools
==> homebrew/test-bot
==> caskroom/cask
```

AFTER:
```
$ brew readall
==> homebrew/core
```

### DIFF
--- a/Library/Homebrew/cmd/readall.rb
+++ b/Library/Homebrew/cmd/readall.rb
@@ -29,7 +29,7 @@ module Homebrew
 
     options = { aliases: ARGV.include?("--aliases") }
     taps = if ARGV.named.empty?
-      Tap
+      [Tap.fetch("homebrew/core")]
     else
       [Tap.fetch(ARGV.named.first)]
     end


scpeters:
I think you don't even need to add `ohai` to see the effect if you use `brew readall --debug`, though the latter is much more verbose

MikeMcQuaid:
> There is a discrepancy between the description in cmd/readall.rb and what it actually does:
> currently, brew readall reads all taps, not only the homebrew/core one.

I think we should adjust the documentation to match the behaviour not the behaviour to match the documentation.

maxim-belkin:
> I think we should adjust the documentation to match the behaviour not the behaviour to match the documentation.

ok :)

PR #3305, opened 2017-10-11
Use smart quotes and ellipses throughout the docs

koraktor:
… and add a note about this in the Prose Style Guidelines.


MikeMcQuaid:
Are you sure the rendered Markdown doesn't do this automatically? If so I think it would be worth keeping these as-is.

ilovezfs:
The rendered markdown does not do that automatically, but I still don't think we should do this.

koraktor:
@ilovezfs: Any specific reason why?

IMHO it’s a good idea to ask for a consistent typogaphic style. I know this doesn’t bring a huge benefit, but it’s an improvement nevertheless.

ilovezfs:
Consistency is good, but consistently not using smart quotes, etc. seems more convenient to maintain even if some people think it's less pretty.

MikeMcQuaid:
I agree, I'm afraid, sorry @koraktor. I would accept a PR (or this one can be modified) to ensure smart quotes are enabled in Markdown rendering on https://docs.brew.sh and https://brew.sh, though.

reitermarkus:
I wonder if there is a linter for prose.

 … of course there is: https://github.com/amperser/proselint 😄

MikeMcQuaid:
Passing on this, sorry @koraktor. I'd be happy to make a change to the GitHub Pages configuration that enables smart quotes.

PR #3306, opened 2017-10-12
Fixes querying Cask commands' help

amyspark:
-----

In caskroom/homebrew-cask#28977, two bugs were discovered (quote from my comment).

The first one:
> That last option is being “stolen” by Homebrew itself. Try something like brew cask example --help and you’ll get their help, not even ours.

It's a bug in Cask's `cli.rb`:
https://github.com/Homebrew/brew/blob/56458f03fcc68ef6d8ee3ee4a7c1d16021aa5800/Library/Homebrew/cask/lib/hbc/cli.rb#L227-L235
Line 234 specifies that it should print the first argument if available, which the example script does (instead of `example`, which doesn't exist as a command!)

The second one:
> And this looks like a bug, but I just tried brew services --help and my terminal is filled with the (correct) instructions. I then cannot leave/do anything until I <kbd>ctrl</kbd><kbd>C</kbd>.

This one only happens *if homebrew/services has not been previously tapped*, because in this case the help request is redirected to the `tap` command, which *always returns the help text without actually tapping the cask* 🤦‍♀️ .
The execution flow is roughly like this:

`brew services --help`
 - `help.rb` [does not find the command](https://github.com/Homebrew/brew/blob/021729e1646212d47d3f52c0b29034e7da97693b/Library/Homebrew/cmd/help.rb#L69) and so returns
 - now `brew.rb` tests [whether it belongs to a Tap](https://github.com/Homebrew/brew/blob/b488a89b5d2bd9c96e3128f598221fed1c3c1911/Library/Homebrew/brew.rb#L103-L107)
 - it does, so `brew tap homebrew/services` [is triggered](https://github.com/Homebrew/brew/blob/b488a89b5d2bd9c96e3128f598221fed1c3c1911/Library/Homebrew/brew.rb#L113-L115)
 -  ***because the `HOMEBREW_HELP` environment variable is still set, `help.rb` sets the `help_flag`, [and attempts to print `tap` help instead](https://github.com/Homebrew/brew/blob/021729e1646212d47d3f52c0b29034e7da97693b/Library/Homebrew/cmd/help.rb#L72-L73)***
 -  now, `puts(command_help(path))` does the job, but after doing so `exit 0` returns _from the "tap"_
 - [we execute `brew services --help` again](https://github.com/Homebrew/brew/blob/b488a89b5d2bd9c96e3128f598221fed1c3c1911/Library/Homebrew/brew.rb#L115), and since the tap didn't clone the repo but merely printed the help text, this triggers the loop.

The bug lies in the remarked step: if a tap is required, it should *not* obey the `HOMEBREW_HELP` environment variable.

I've fixed the first bug by specifying all `args` to be printed along with the offending command, and the second by forcing `brew.rb` to ignore the shell's `HOMEBREW_HELP` variable and verify that the help text was actually requested.

Fixes caskroom/homebrew-cask#28977

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -32,7 +32,7 @@ begin
 
   empty_argv = ARGV.empty?
   help_flag_list = %w[-h --help --usage -?]
-  help_flag = false
+  help_flag = !ENV["HOMEBREW_HELP"].nil?
   internal_cmd = true
   cmd = nil
 
@@ -119,8 +119,16 @@ begin
     if Process.uid.zero? && !brew_uid.zero?
       tap_commands += %W[/usr/bin/sudo -u ##{brew_uid}]
     end
+    if help_flag
+      # Unset HOMEBREW_HELP to avoid confusing the tap
+      ENV["HOMEBREW_HELP"] = nil
+    end
     tap_commands += %W[#{HOMEBREW_BREW_FILE} tap #{possible_tap}]
     safe_system(*tap_commands)
+    if help_flag
+      # Restore HOMEBREW_HELP after the tap
+      ENV["HOMEBREW_HELP"] = 1
+    end
     exec HOMEBREW_BREW_FILE, cmd, *ARGV
   end
 rescue UsageError => e
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -80,7 +80,13 @@ begin
   # - no arguments are passed
   if empty_argv || help_flag
     require "cmd/help"
-    Homebrew.help cmd, empty_argv: empty_argv
+    if cmd == "cask"
+      # Let Cask handle the help command
+      Homebrew.send cmd.to_s.tr("-", "_").downcase
+      exit 0
+    else
+      Homebrew.help cmd, empty_argv: empty_argv
+    end
     # `Homebrew.help` never returns, except for external/unknown commands.
   end
 
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -44,9 +44,6 @@ begin
       help_flag = true
     elsif !cmd && !help_flag_list.include?(arg)
       cmd = ARGV.delete_at(i)
-    elsif help_flag_list.include?(arg) && cmd
-      # cmd determined, and it needs help
-      help_flag = true
     end
   end
 
@@ -121,13 +118,13 @@ begin
     end
     if help_flag
       # Unset HOMEBREW_HELP to avoid confusing the tap
-      ENV["HOMEBREW_HELP"] = nil
+      ENV.delete("HOMEBREW_HELP")
     end
     tap_commands += %W[#{HOMEBREW_BREW_FILE} tap #{possible_tap}]
     safe_system(*tap_commands)
     if help_flag
       # Restore HOMEBREW_HELP after the tap
-      ENV["HOMEBREW_HELP"] = 1
+      ENV["HOMEBREW_HELP"] = "1"
     end
     exec HOMEBREW_BREW_FILE, cmd, *ARGV
   end
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -44,7 +44,7 @@ begin
       help_flag = true
     elsif !cmd && !help_flag_list.include?(arg)
       cmd = ARGV.delete_at(i)
-    elsif help_flag_list.include?(arg) & cmd
+    elsif help_flag_list.include?(arg) && cmd
       # cmd determined, and it needs help
       help_flag = true
     end
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -78,15 +78,10 @@ begin
   # - a help flag is passed AND a command is matched
   # - a help flag is passed AND there is no command specified
   # - no arguments are passed
-  if empty_argv || help_flag
+  # - if cmd is Cask, let Cask handle the help command instead
+  if (empty_argv || help_flag ) && cmd != "cask"
     require "cmd/help"
-    if cmd == "cask"
-      # Let Cask handle the help command
-      Homebrew.send cmd.to_s.tr("-", "_").downcase
-      exit 0
-    else
-      Homebrew.help cmd, empty_argv: empty_argv
-    end
+    Homebrew.help cmd, empty_argv: empty_argv
     # `Homebrew.help` never returns, except for external/unknown commands.
   end
 
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -32,7 +32,7 @@ begin
 
   empty_argv = ARGV.empty?
   help_flag_list = %w[-h --help --usage -?]
-  help_flag = !ENV["HOMEBREW_HELP"].nil?
+  help_flag = false
   internal_cmd = true
   cmd = nil
 
@@ -44,6 +44,9 @@ begin
       help_flag = true
     elsif !cmd && !help_flag_list.include?(arg)
       cmd = ARGV.delete_at(i)
+    elsif help_flag_list.include?(arg) & cmd
+      # cmd determined, and it needs help
+      help_flag = true
     end
   end
 
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -78,8 +78,8 @@ begin
   # - a help flag is passed AND a command is matched
   # - a help flag is passed AND there is no command specified
   # - no arguments are passed
-  # - if cmd is Cask, let Cask handle the help command instead
-  if (empty_argv || help_flag ) && cmd != "cask"
+  # - if cmd is Cask, let Cask handle the help command instead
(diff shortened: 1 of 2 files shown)


Library/Homebrew/brew.rb
```diff
@@ -44,6 +44,9 @@
       help_flag = true
     elsif !cmd && !help_flag_list.include?(arg)
       cmd = ARGV.delete_at(i)
+    elsif help_flag_list.include?(arg) & cmd
```
MikeMcQuaid: This is pretty opaque; can you elaborate on what this is doing?

Library/Homebrew/brew.rb
```diff
@@ -32,7 +32,7 @@
 
   empty_argv = ARGV.empty?
   help_flag_list = %w[-h --help --usage -?]
-  help_flag = !ENV["HOMEBREW_HELP"].nil?
```
MikeMcQuaid: This environment variable should not be ignored.

amyspark: The problem is, if the help request needs a `brew tap` before fulfilling it (as in the example above), it detects this variable and outputs the `tap` help instead of actually tapping the cask.

Library/Homebrew/brew.rb
amyspark: My bad, I meant `&&`. Fix incoming.
The two changes are meant to be read together: by ignoring the `HOMEBREW_HELP` variable I force Homebrew to detect if a help request was actually issued *to the current process*.
In the case above, when `brew tap homebrew/services` is issued, this case detects that *no* help flags or command are present and lets the tap proceed.

amyspark:
I've done the requested changes. Are they OK?

Library/Homebrew/brew.rb
MikeMcQuaid: Sure but in that case handle the `brew tap` case more specifically either be unsetting `HOMEBREW_HELP` in `brew.sh` when appropriate (https://github.com/Homebrew/brew/blob/cf349018d5cc508ddd2bd0797df47688b9f7346b/Library/Homebrew/brew.sh#L211-L218) or just for the `brew tap` here: https://github.com/Homebrew/brew/blob/cf349018d5cc508ddd2bd0797df47688b9f7346b/Library/Homebrew/brew.rb#L119-L120

amyspark: Done, see 0998e16.

Library/Homebrew/brew.rb
```diff
@@ -119,8 +119,16 @@
...
     end
+    if help_flag
+      # Unset HOMEBREW_HELP to avoid confusing the tap
+      ENV["HOMEBREW_HELP"] = nil
+    end
     tap_commands += %W[#{HOMEBREW_BREW_FILE} tap #{possible_tap}]
     safe_system(*tap_commands)
+    if help_flag
+      # Restore HOMEBREW_HELP after the tap
+      ENV["HOMEBREW_HELP"] = 1
```
reitermarkus: This has to be a string.

Library/Homebrew/brew.rb
```diff
@@ -116,8 +119,16 @@
     if Process.uid.zero? && !brew_uid.zero?
       tap_commands += %W[/usr/bin/sudo -u ##{brew_uid}]
     end
+    if help_flag
+      # Unset HOMEBREW_HELP to avoid confusing the tap
+      ENV["HOMEBREW_HELP"] = nil
```
reitermarkus: Use `ENV.delete`.

Library/Homebrew/brew.rb
```diff
@@ -116,8 +116,16 @@
     if Process.uid.zero? && !brew_uid.zero?
       tap_commands += %W[/usr/bin/sudo -u ##{brew_uid}]
     end
+    if help_flag
+      # Unset HOMEBREW_HELP to avoid confusing the tap
+      ENV.delete("HOMEBREW_HELP")
```
MikeMcQuaid: `ENV.delete("HOMEBREW_HELP") if help_flag`

Library/Homebrew/brew.rb
```diff
@@ -116,8 +116,16 @@
...
+    end
     tap_commands += %W[#{HOMEBREW_BREW_FILE} tap #{possible_tap}]
     safe_system(*tap_commands)
+    if help_flag
+      # Restore HOMEBREW_HELP after the tap
+      ENV["HOMEBREW_HELP"] = "1"
```
MikeMcQuaid: `ENV["HOMEBREW_HELP"] = "1" if help_flag` and can skip the comment as it's clear without it, thanks.

amyspark:
I'm testing the change I made to `cli.rb`, please do not merge yet.

amyspark:
@MikeMcQuaid, I've reverted the lines in cli.rb, I think a better option is to print the first offending command (`example`) instead of printing the whole Cask command (`help example -a`).
Please review, and if it is OK, do merge this.
Thanks for your comments!

Library/Homebrew/cask/lib/hbc/cli.rb
```diff
@@ -230,7 +230,7 @@ def run(*_args)
 
         return if @command == "help" && @args.empty?
 
-        unknown_command = @args.empty? ? @command : @args.first
+        unknown_command = @command == "help" ? @args.first : @command
```
MikeMcQuaid: CC @reitermarkus for a 👍 on this.

reitermarkus: This error is a bit useless actually, because `brew cask help install` will result in `Error: Unknown command: install`, because there is no per-command help implemented.

I think the error should be `"help does not take arguments."`.

Library/Homebrew/cask/lib/hbc/cli.rb
```diff
@@ -230,7 +230,7 @@ def run(*_args)
 
         return if @command == "help" && @args.empty?
 
-        unknown_command = @args.empty? ? @command : @args.first
+        unknown_command = (@command == "help") ? @args.first : @command
```
reitermarkus: https://github.com/Homebrew/brew/pull/3306#discussion_r146018543

reitermarkus:
I don't think this completely fixes https://github.com/caskroom/homebrew-cask/issues/28977. Does `brew cask -h` now show the same as `brew cask help`?

amyspark:
@reitermarkus, no, it currently doesn't. The problem is that with the current setup, Homebrew tries to print Cask's help by grepping `/usr/local/Homebrew/Library/Homebrew/cmd/cask.rb`, which is empty (set HOMEBREW_DEVELOPER to see the warning message). I can't yet find a way to send the request to Cask itself without breaking the `brew help` command.

Library/Homebrew/cask/lib/hbc/cli.rb
```diff
@@ -230,8 +230,7 @@ def run(*_args)
 
         return if @command == "help" && @args.empty?
 
-        unknown_command = @args.empty? ? @command : @args.first
-        raise ArgumentError, "Unknown command: #{unknown_command}"
+        raise ArgumentError, "help does not take arguments"
```
reitermarkus: End this with a `.`.

Library/Homebrew/cask/lib/hbc/cli.rb
```diff
@@ -154,7 +154,7 @@ def detect_command_and_arguments(*args)
     def run
       command_name, *args = detect_command_and_arguments(*@args)
       command = if help?
-        args.unshift(command_name)
+        args.unshift(command_name) if !command_name.nil?
```
reitermarkus: `unless` instead of `if !`

amyspark:
Um, I'm not sure why Travis failed with the last commit, it doesn't show anything in the build log.

MikeMcQuaid:
Tests failed again after rerun but I'm unable to figure out why.

MikeMcQuaid:
@reitermarkus What do you think about making Cask subcommand help use the same methods as Homebrew command help rather than just making this a blind passthrough?

amyspark:
@MikeMcQuaid I ran `brew tests` manually and the test that is failing is cask/cli:46:
https://github.com/Homebrew/brew/blob/ce0e96c3205e10fe975b8c107092c290b9373af9/Library/Homebrew/test/cask/cli_spec.rb#L46-L51

Library/Homebrew/brew.rb
```diff
@@ -80,7 +80,13 @@
   # - no arguments are passed
   if empty_argv || help_flag
```
MikeMcQuaid: Does this work if you add `cmd != "cask"` here and falls into the code below as expected? If so that seems like a more minimal and cleaner solution. If you change to that and it works: please add the `# Let Cask handle the help command` comment, too, thanks!

amyspark: Yes, it works! I've simplified it below. Thanks for the comment! 👍

MikeMcQuaid:
Thanks @amyspark! I'm good with this when @reitermarkus is; can you merge when you're happy (and CI is 💚), thanks!

reitermarkus:
Thank you for your first contribution, @amyspark!

PR #3315, opened 2017-10-13
caveats.rb: empty method on Linux

maxim-belkin:

-----

`plist_caveats` does not apply to Linux so we have to override it to an empty method.

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -161,49 +161,7 @@ class Caveats
     EOS
   end
 
-  def plist_caveats
-    s = []
-    if f.plist || (keg&.plist_installed?)
-      plist_domain = f.plist_path.basename(".plist")
-
-      # we readlink because this path probably doesn't exist since caveats
-      # occurs before the link step of installation
-      # Yosemite security measures mildly tighter rules:
-      # https://github.com/Homebrew/legacy-homebrew/issues/33815
-      if !plist_path.file? || !plist_path.symlink?
-        if f.plist_startup
-          s << "To have launchd start #{f.full_name} now and restart at startup:"
-          s << "  sudo brew services start #{f.full_name}"
-        else
-          s << "To have launchd start #{f.full_name} now and restart at login:"
-          s << "  brew services start #{f.full_name}"
-        end
-      # For startup plists, we cannot tell whether it's running on launchd,
-      # as it requires for `sudo launchctl list` to get real result.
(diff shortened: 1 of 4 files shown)


maxim-belkin:
Forgot to mention @sjackman to chime in/check

sjackman:
Thanks for this Maxim. There's a bit more work needed. The existing `plist_caveats` method needs to move to a file `extend/os/mac/caveats.rb`. Create a file `extend/os/caveats.rb` with contents
```ruby
if OS.mac?
  require "extend/os/mac/caveats"
elsif OS.linux?
  require "extend/os/linux/caveats"
end
```

Add to the end of `Library/Homebrew/caveats.rb`
```ruby
require "extend/os/system_config"
```

maxim-belkin:
OK, will do

If only there was a variable to add a single line, instead of 5
```
require "extend/os/#{os}/caveats"
```

sjackman:
There's `ENV["HOMEBREW_SYSTEM"]` which has the value `Macintosh`.
I'd suggest creating a global variable in `os.rb` named `OS::SYSTEM` that takes the values `mac`, `linux`, and `generic`.

The file `"extend/os/#{os}/caveats"` may not exist, so it'd have to be something like…
```ruby
require "extend/os/#{os}/caveats" if (HOMEBREW_REPOSITORY/"extend/os"/os/"caveats.rb").readable?
```

@MikeMcQuaid Your thought on the above, rather than the current…
```ruby
if OS.mac?
  require "extend/os/mac/system_config"
elsif OS.linux?
  require "extend/os/linux/system_config"
end
```

maxim-belkin:
> ```
> require "extend/os/#{os}/caveats" if (HOMEBREW_REPOSITORY/"extend/os"/os/"caveats.rb").readable?

Is it possible / OK to "extend" the `require` method?

sjackman:
I wouldn't extend it, but you could define a new method `require_if_available` (or some name).

maxim-belkin:
Made the changes without creating any new variable(s).

sjackman:
```
Library/Homebrew/extend/os/mac/caveats.rb:45:1: C: Extra empty line detected at class body end.
```
https://travis-ci.org/Homebrew/brew/jobs/288642476#L860

maxim-belkin:
Yes, I noticed that after I pushed it :)

maxim-belkin:
Fixed!

MikeMcQuaid:
Thanks again @maxim-belkin!

PR #3318, opened 2017-10-14
Add audit check for URL schema

Git-Jiro:

-----
This implements: https://github.com/caskroom/homebrew-cask/issues/25403

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/audit.rb
+++ b/Library/Homebrew/cask/lib/hbc/audit.rb
@@ -2,6 +2,7 @@ require "hbc/checkable"
 require "hbc/download"
 require "digest"
 require "utils/git"
+require "dev-cmd/audit"
 
 module Hbc
   class Audit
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cask/lib/hbc/audit.rb
```diff
@@ -243,6 +244,78 @@ def core_formula_url
...
+      #   0, nil           : URL unreachable
+      #  -1, effective URL : URL reachable, but schema changed
+
+      curl_executable, *args = curl_args(
+        "--compressed", "--location", "--fail",
+        "--write-out", "%{http_code} %{url_effective}",
+        "--output", "/dev/null",
+        url_to_access,
+        user_agent: :fake
+      )
```
reitermarkus: This will download the whole file to `/dev/null`, won't it? It should use `--head`.

MikeMcQuaid:
It would be good to share more code here with `audit`'s equivalent.

Git-Jiro:
@MikeMcQuaid Ok, I will have a look.

Git-Jiro:
@MikeMcQuaid is this the kind of code reuse you envisioned?

MikeMcQuaid:
@Git-Jiro Yes, something like that. You may want to extract the `FormulaAuditor.check_http_content` into `utils/curl` instead as importing a Homebrew `cmd/audit` inside Cask seems a little weird. Wait for @reitermarkus's thoughts first, though.

Git-Jiro:
@MikeMcQuaid Do you have an example for me where I can look at a similar refactoring? I would like to see how other people handled similar situations in the code base.

MikeMcQuaid:
@Git-Jiro I don't have a good example I can think of but I can answer more specific questions and provide review feedback.

Git-Jiro:
Ok

Git-Jiro:
@MikeMcQuaid What do you think about my first version of this refactoring?

MikeMcQuaid:
@Git-Jiro Looks great, thanks for that! I'm happy to 🚢 this when @reitermarkus is.

Library/Homebrew/cask/lib/hbc/audit.rb
```diff
@@ -275,6 +277,17 @@ def core_formula_url
       "#{core_tap.default_remote}/blob/master/Formula/#{cask.token}.rb"
     end
 
+    def check_https_availability
+      check_url_for_https_availability(cask.url) unless cask.url.to_s.empty?
```
reitermarkus: This should probably also use `cask.url.user_agent`.

Library/Homebrew/cask/lib/hbc/audit.rb
```diff
@@ -275,6 +277,18 @@ def core_formula_url
       "#{core_tap.default_remote}/blob/master/Formula/#{cask.token}.rb"
     end
 
+    def check_https_availability
+      user_agents = cask.url.user_agent.to_s.empty? ? [:default] : [cask.url.user_agent]
```
reitermarkus: `#user_agent` already defaults to `:default` by itself.

reitermarkus:
Thanks, @Git-Jiro, nice work here!

PR #3320, opened 2017-10-15
Install RuboCop with Bundler.

reitermarkus:

-----

Don't download RuboCop twice, once without, and once with Bundler.


Library/Homebrew/Gemfile
```diff
@@ -0,0 +1,20 @@
...
+  gem "rspec-wait", require: false
+end
+
+group :test, :coverage do
+  gem "codecov", require: false
+  gem "simplecov", require: false
+end
+
+group :rubocop do
+  gem "rubocop", HOMEBREW_RUBOCOP_VERSION
```
MikeMcQuaid: Is there any reason to have these constants in the `constants` file rather than just having them in the `Gemfile` after this change?

Library/Homebrew/utils.rb
```diff
@@ -187,7 +187,23 @@ def system(cmd, *args)
     _system(cmd, *args)
   end
 
+  def install_and_require_rubocop!
+    install_gem_setup_path! "bundler"
+    require "bundler"
+
+    quiet_system "bundle", "config", "--delete", "without"
+    quiet_system "bundle", "config", "--local", "without", "test"
```
MikeMcQuaid: Will this not uninstall the test gems meaning `brew tests && brew style` always cause a `bundle install` to be needed? I could be totally wrong here.

Library/Homebrew/Gemfile
```diff
@@ -0,0 +1,20 @@
+source "https://rubygems.org"
```
MikeMcQuaid: I think ideally this file would live in the repo root now.

Library/Homebrew/test/.bundle/config
```diff
@@ -1,4 +0,0 @@
----
```
MikeMcQuaid: Is this file still created by bundler itself? My issue is that by removing this we break a manually run `bundle install`

Library/Homebrew/utils.rb
reitermarkus: The `test` ones will still be installed.

Library/Homebrew/Gemfile
reitermarkus: No, just wanted to keep the changes here as minimal as possible for the time being.

Library/Homebrew/test/.bundle/config
reitermarkus: Needs to be checked out if we want to run `style` without installing everything else.

Library/Homebrew/utils.rb
MikeMcQuaid: Ok.

Library/Homebrew/Gemfile
MikeMcQuaid: Let's remove them then, thanks.

Library/Homebrew/test/.bundle/config
MikeMcQuaid: Ok, thanks.

ilovezfs:
@reitermarkus did you intend to let stale bot do that to you?

reitermarkus:
> @reitermarkus did you intend to let stale bot do that to you?

Yeah, there was too much trial and error here. 😄

PR #3329, opened 2017-10-17
Specifies which errant version of Ruby is used in raise.

ashfurrow:

-----

As [discussed here](https://github.com/Homebrew/brew/pull/3240/files#r145185621), it would be helpful to display which version of Ruby is actually being used in the error message when used in non-2.3 environments. This is especially useful when running commands remotely, such as on CI.

Regarding testing: I didn't see any existing tests (I don't think it'd be easy to test this, as the code is executed at the top level) but if you point me in the right direction I'm sure I could figure it out.

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -9,7 +9,7 @@ RUBY_VERSION_SPLIT = RUBY_VERSION.split "."
 RUBY_X = RUBY_VERSION_SPLIT[0].to_i
 RUBY_Y = RUBY_VERSION_SPLIT[1].to_i
 if RUBY_X < 2 || (RUBY_X == 2 && RUBY_Y < 3)
-  raise "Homebrew must be run under Ruby 2.3!"
+  raise "Homebrew must be run under Ruby 2.3! You're running #{RUBY_VERSION}."
 end
 
 require "pathname"


Library/Homebrew/brew.rb
```diff
@@ -9,7 +9,7 @@
 RUBY_X = RUBY_VERSION_SPLIT[0].to_i
 RUBY_Y = RUBY_VERSION_SPLIT[1].to_i
 if RUBY_X < 2 || (RUBY_X == 2 && RUBY_Y < 3)
-  raise "Homebrew must be run under Ruby 2.3!"
+  raise "Homebrew must be run under Ruby 2.3, but you're running #{RUBY_VERSION}!"
```
reitermarkus: I'd make this `"Homebrew must be run under Ruby 2.3! You're running #{RUBY_VERSION}."`

ashfurrow: Taken care of 👍

reitermarkus:
Thanks, @ashfurrow!

MikeMcQuaid:
> As discussed here, it would be helpful to display which version of Ruby is actually being used in the error message when used in non-2.3 environments. This is especially useful when running commands remotely, such as on CI.

@ashfurrow For what it's worth: any time you see this message unless you've manually set `HOMEBREW_RUBY_PATH` we consider it a bug. Can you elaborate on where you've seen this?

ashfurrow:
Interesting! The issue came up on our CI provider, fixed by [this PR](https://github.com/artsy/eigen/pull/2435). I was pointed in the direction of running `brew update` on [this thread](https://discuss.circleci.com/t/homebrew-must-be-run-under-ruby-2-3-runtimeerror/17232/8). Let me know if there's any more info I can provide.

MikeMcQuaid:
Ok, I think that's the same as a bug we've seen before. I'll look at addressing it, thanks.

MikeMcQuaid:
I'm afraid I've tried but cannot reproduce this on even updates from very old Homebrews now so I can only presume I've fixed it.

ashfurrow:
Thank you for following up 🌟

PR #3332, opened 2017-10-17
setup-ruby-path: Ensure that Ruby is installed

sjackman:

-----

Fix the error:
```
/usr/local/Homebrew/Library/Homebrew/brew.sh: line 372: exec: : not found
```

### DIFF
--- a/Library/Homebrew/utils/ruby.sh
+++ b/Library/Homebrew/utils/ruby.sh
@@ -40,7 +40,7 @@ setup-ruby-path() {
         ruby_old_version="$("$HOMEBREW_RUBY_PATH" -rrubygems -e "puts Gem::Version.new('$minimum_ruby_version') > Gem::Version.new(RUBY_VERSION)")"
       fi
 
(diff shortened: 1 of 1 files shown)


sjackman:
Previously when no Ruby is installed, the vendored Ruby would be installed.
Now when no Ruby is installed, the above error message is printed.
I believe the change of behaviour occurred here:
https://github.com/Homebrew/brew/commit/b2b413165f3553c48c8e060dd97ccde9c13cae80#diff-2012e7e4fc656781b9d32ead2909ee8bR43
This PR restores the previous behaviour. When no Ruby is installed, the vendored Ruby will be installed.

MikeMcQuaid:
> Previously when no Ruby is installed, the vendored Ruby would be installed.

Just trying to understand this: is that if there's no Ruby found even in e.g. `/usr/bin` i.e. this is a Linux only bug? Happy to merge regardless but just want to make sure I understand.

sjackman:
Yes, correct. This bug only affects Linux, since macOS always has `/usr/bin/ruby` available.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
My pleasure. Thanks again for merging, Mike.

Issue #3331, opened 2017-10-17
The JavaRequirement#satisfies_version breaks for JDK 9 version string scheme


OfTheDelmer:
# Java Requirement
### Brew Issue Check List

- [x] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at the relevant tap e.g. for Homebrew/homebrew-core https://github.com/Homebrew/homebrew-core/issues/new
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed all issues and retried your prior step?
- [x] Ran `brew config` and `brew doctor` and included their output with your issue?

------

`$ brew config`

```shell
HOMEBREW_VERSION: 1.3.5
ORIGIN: https://github.com/Homebrew/brew
HEAD: 91ab116ace7f4f97d3440190463c93be9ec6d675
Last commit: 9 days ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 1c4f70ce126c6aaea88ecd71a6318d35958137c2
Core tap last commit: 15 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: quad-core 64-bit broadwell
Homebrew Ruby: 2.3.3 => /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.3/bin/ruby
Clang: 8.1 build 802
Git: 2.14.1 => /usr/local/bin/git
Perl: /usr/local/bin/perl => /usr/local/Cellar/perl/5.24.0_1/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /Users/delmer.reed/.rvm/rubies/ruby-2.3.0/bin/ruby
Java: 9.0.1, 1.8.0_144
macOS: 10.12.5-x86_64
Xcode: 8.3.3
CLT: 8.3.2.0.1.1492020469
X11: 2.7.11 => /opt/X11
```

`$ brew doctor`

```shell
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
[...]

PR #3333, opened 2017-10-18
java_requirement: fix Java 9 version detection

DomT4:

-----

I went for pretty much the most obvious possible fix here, so feel free to debate/request changes/etc as desired. We don't really need to consider post-period digit with Java 9 onwards, but retaining the existing behaviour there shouldn't cause issues with existing Java dependencies in core _(as far as I can see)_ because 9 > 1.

Obviously the `/(1|9)\.\d/` won't support Java 10 but that problem is likely several years away.

Closes https://github.com/Homebrew/brew/issues/3331.

### DIFF
--- a/Library/Homebrew/requirements/java_requirement.rb
+++ b/Library/Homebrew/requirements/java_requirement.rb
@@ -11,7 +11,7 @@ class JavaRequirement < Requirement
   end
 
   def initialize(tags)
-    @version = tags.shift if /(\d\.)+\d/ =~ tags.first
+    @version = tags.shift if /(\d+\.)+\d/ =~ tags.first
     super
   end
 
@@ -103,7 +103,7 @@ class JavaRequirement < Requirement
   end
 
   def satisfies_version(java)
-    java_version_s = Utils.popen_read("#{java} -version 2>&1")[/1.\d/]
+    java_version_s = Utils.popen_read("#{java} -version 2>&1")[/\d+.\d/]
     return false unless java_version_s
     java_version = Version.create(java_version_s)
     needed_version = Version.create(version_without_plus)


Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -103,7 +103,7 @@ def oracle_java_os
   end
 
   def satisfies_version(java)
-    java_version_s = Utils.popen_read("#{java} -version 2>&1")[/1.\d/]
+    java_version_s = Utils.popen_read("#{java} -version 2>&1")[/(1|9)\.\d/]
```
ilovezfs: why not `/\d+.\d/`?

commitay:
Thanks for taking care of this @DomT4! I forgot about the `java_requirement` for Formula.
___
> Obviously the `/(1|9)\.\d/` won't support Java 10 but that problem is likely several years away.

Java 18.3 is scheduled for release March 2018

http://www.oracle.com/technetwork/java/javase/eol-135779.html

DomT4:
> Java 18.3 is scheduled for release March 2018

What the heck are ya doing with your version schemes Oracle 😅. Thanks for letting me know about that, I guess we shouldn't punt the problem then.

Library/Homebrew/requirements/java_requirement.rb
DomT4: I assumed it was hardcoded for some desirable reason, to be honest. If there's not a known one I have no objections to that.

Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -11,7 +11,7 @@ class JavaRequirement < Requirement
   end
 
   def initialize(tags)
-    @version = tags.shift if /(\d\.)+\d/ =~ tags.first
+    @version = tags.shift if /(\d\.?\d?)+\d/ =~ tags.first
```
ilovezfs: it'll need a \d+ up here too

MikeMcQuaid:
Thanks again @DomT4!

DomT4:
Thanks for the input all ❤️.

PR #3337, opened 2017-10-19
pull: Fix pull --bottle 1234

sjackman:

-----

Fix the error:
```
Error: undefined method `casecmp' for nil:NilClass
```

### DIFF
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -26,11 +26,11 @@ class Tap
       raise "Invalid tap name '#{args.join("/")}'"
     end
 
-    # we special case homebrew so users don't have to shift in a terminal
-    user = "Homebrew" if user == "homebrew"
+    # We special case homebrew and linuxbrew so that users don't have to shift in a terminal.
+    user = user.capitalize if ["homebrew", "linuxbrew"].include? user
     repo = repo.strip_prefix "homebrew-"
 
-    if user == "Homebrew" && (repo == "homebrew" || repo == "core")
+    if ["Homebrew", "Linuxbrew"].include?(user) && ["core", "homebrew"].include?(repo)
       return CoreTap.instance
     end
 
--- a/Library/Homebrew/test/dev-cmd/pull_spec.rb
+++ b/Library/Homebrew/test/dev-cmd/pull_spec.rb
@@ -22,7 +22,7 @@ describe "brew pull", :integration_test do
       .and output(/Current branch is new\-branch/).to_stderr
       .and be_a_failure
 
-    expect { brew "pull", "--bump", "8" }
+    expect { brew "pull", "--bump", "https://github.com/Homebrew/homebrew-core/pull/8" }
       .to output(/Fetching patch/).to_stdout
       .and output(/No changed formulae found to bump/).to_stderr
       .and be_a_failure
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -87,11 +87,8 @@ module Homebrew
     tap = nil
 
     ARGV.named.each do |arg|
-      if arg.to_i.positive?
-        issue = arg
-        url = "https://github.com/Homebrew/homebrew-core/pull/#{arg}"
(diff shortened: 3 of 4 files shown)


sjackman:
This PR fixes a bug that I introduced in PR #3279. Sorry for that.

MikeMcQuaid:
@sjackman This is failing on the Linux CI. Note the Linux CI should not end up cloning Linuxbrew in any situation.

scpeters:
> Note the Linux CI should not end up cloning Linuxbrew in any situation.

I see references to linuxbrew repositories in the following places:

https://github.com/Homebrew/brew/blob/ffb582b5a6d829c5825e6d6a40545cb3099001fb/Library/Homebrew/cmd/update.sh#L32-L33

https://github.com/Homebrew/brew/blob/3a606abaa13a3090f3c6336efed435006f091c60/Library/Homebrew/tap.rb#L544

MikeMcQuaid:
@scpeters Yep, those are fine but we've previously taken lengths to workaround that in CI itself.

JCount:
@sjackman 
The CI issue seems caused by the use of [CoreTap's default_remote method](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/tap.rb#L539-L546),  due to its platform-specific behavior. This is likely cause for the test suite to cloning `https://github.com/Linuxbrew/homebrew-core` when it runs `brew pull 1` as part of the [pull_spec](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/dev-cmd/pull_spec.rb#L20-L23) tests. I suspect that
```
arg = "#{CoreTap.instance.default_remote}/pull/#{arg}" if arg.to_i.positive?
```
 is evaluating to 
```
arg = "https://github.com/Linuxbrew/homebrew-core/pull/1"
```
on the Linux CI.

sjackman:
I've fixed the test failure. I've removed the synonym of the tap name `Homebrew/homebrew` for `Homebrew/core`. I can put it back in if you prefer.

Library/Homebrew/tap.rb
```diff
@@ -27,10 +27,10 @@ def self.fetch(*args)
     end
 
     # we special case homebrew so users don't have to shift in a terminal
-    user = "Homebrew" if user == "homebrew"
+    user = user.capitalize if ["homebrew", "linuxbrew"].include? user
     repo = repo.strip_prefix "homebrew-"
 
-    if user == "Homebrew" && (repo == "homebrew" || repo == "core")
```
MikeMcQuaid: Does the `repo == "homebrew"` break anything here? If it's not required for the fix can you put it back? Thanks!

Library/Homebrew/test/dev-cmd/pull_spec.rb
```diff
@@ -22,7 +22,7 @@
       .and output(/Current branch is new\-branch/).to_stderr
       .and be_a_failure
 
-    expect { brew "pull", "--bump", "8" }
+    expect { brew "pull", "--bump", "https://github.com/Homebrew/homebrew-core/pull/8" }
```
MikeMcQuaid: To be explicit: this is just to fix the test failure on Linux, right? Using `8` will still pull Homebrew/homebrew-core on macOS?

Library/Homebrew/tap.rb
```diff
@@ -27,10 +27,10 @@ def self.fetch(*args)
     end
 
     # we special case homebrew so users don't have to shift in a terminal
```
MikeMcQuaid: `and linuxbrew`

Library/Homebrew/tap.rb
sjackman: It doesn't break anything. Just seemed to be an anachronism. I can put it back.

Library/Homebrew/test/dev-cmd/pull_spec.rb
sjackman: Correct.

MikeMcQuaid:
Thanks again @sjackman and sorry for all the back and forth.

sjackman:

PR #3341, opened 2017-10-19
build-environment: OS-specific environment keys

maxim-belkin:

-----

CC @sjackman 

Provide a way to deal with OS-specific build environment variables. On Linux, those are: LD_LIBRARY_PATH, LD_RUN_PATH, LD_PRELOAD, LIBRARY_PATH.

### DIFF
--- a/Library/Homebrew/build_environment.rb
+++ b/Library/Homebrew/build_environment.rb
@@ -34,7 +34,7 @@ end
 module Homebrew
   module_function
 
-  def build_env_keys(env)
+  def common_build_env_keys
     %w[
       CC CXX LD OBJC OBJCXX
       HOMEBREW_CC HOMEBREW_CXX
@@ -46,7 +46,13 @@ module Homebrew
       HOMEBREW_SDKROOT HOMEBREW_BUILD_FROM_SOURCE
       MAKE GIT CPP
       ACLOCAL_PATH PATH CPATH
-    ].select { |key| env.key?(key) }
+    ]
+  end
+
+  def os_specific_build_env_keys; end
+
+  def build_env_keys(env)
+    (common_build_env_keys + os_specific_build_env_keys).select { |key| env.key?(key) }
   end
 
   def dump_build_env(env, f = $stdout)
@@ -64,3 +70,5 @@ module Homebrew
     end
   end
 end
+
+require "extend/os/build_environment"
--- a/Library/Homebrew/build_environment.rb
+++ b/Library/Homebrew/build_environment.rb
@@ -49,7 +49,9 @@ module Homebrew
     ]
   end
 
-  def os_specific_build_env_keys; end
+  def os_specific_build_env_keys
+    []
+  end
 
   def build_env_keys(env)
     (common_build_env_keys + os_specific_build_env_keys).select { |key| env.key?(key) }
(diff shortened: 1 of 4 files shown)


sjackman:
I'd be inclined to add `LD_LIBRARY_PATH LD_RUN_PATH LD_PRELOAD LIBRARY_PATH` to `build_env_keys on both OS. The variables that don't affect macOS, well, won't affect it, and so do no harm. Note that `LD_RUN_PATH` and `LIBRARY_PATH` affect `gcc` even on macOS. `LD_LIBRARY_PATH` and `LD_PRELOAD` are specific to Linux's dynamic linker. On macOS these variables are called `DYLD_LIBRARY_PATH` and `DYLD_INSERT_LIBRARIES`.

maxim-belkin:
ok, then it's up to @MikeMcQuaid: shall we add the aforementioned variables to `build_env_keys` or go through the approach implemented here?

maxim-belkin:
Just in case, I'm ready to submit another PR: https://github.com/maxim-belkin/brew-macOS/blob/build-env2/Library/Homebrew/build_environment.rb#L49

Library/Homebrew/build_environment.rb
```diff
@@ -46,7 +46,13 @@ def build_env_keys(env)
       HOMEBREW_SDKROOT HOMEBREW_BUILD_FROM_SOURCE
       MAKE GIT CPP
       ACLOCAL_PATH PATH CPATH
-    ].select { |key| env.key?(key) }
+    ]
+  end
+
+  def os_specific_build_env_keys; end
```
MikeMcQuaid: This should return an empty array or the code below won't work in the generic case.

MikeMcQuaid:
I'm good with this approach. Can you elaborate a bit more on why this is needed, though, and what it does? Is this environment variables Homebrew/brew or formulae or users set? If the latter: when do they need to set these?

Library/Homebrew/build_environment.rb
MikeMcQuaid: `brew tests --generic` reveals this error. It's probably worth always running locally for cross-platform changes.

sjackman:
As far as I can tell `build_env_keys` is only used by `dump_build_env` and `brew --env`. So it's for the purpose of displaying environment variables that the user has set in their environment. In a coming PR, these variables are added to `SANITIZED_VARS` of `extend/ENV/shared.rb` to ensure these environment variables don't affect `brew install`.

sjackman:
> I'm good with this approach.

Do you prefer this PR as is or https://github.com/maxim-belkin/brew-macOS/blob/build-env2/Library/Homebrew/build_environment.rb#L49
My preference is the latter. Whichever you prefer is fine.

MikeMcQuaid:
> As far as I can tell build_env_keys is only used by dump_build_env and brew --env. So it's for the purpose of displaying environment variables that the user has set in their environment. In a coming PR, these variables are added to SANITIZED_VARS of extend/ENV/shared.rb to ensure these environment variables don't affect brew install.

Thanks! Presumably those variables are only output if set? If so, I'm fine with just whitelisting them all without `os/extend` as you suggest. If not, to avoid extra empty output on macOS I either think we should keep this PR's approach or not output them if they aren't set.

> In a coming PR, these variables are added to SANITIZED_VARS of extend/ENV/shared.rb to ensure these environment variables don't affect brew install.

Cool. If those variables are being removed rather than added/modified I'm again 👍 on keeping that the same on macOS

maxim-belkin:
Closing in favor of #3358.

Issue #3100, opened 2017-08-28
brew leaves lists formulaes which are dependencies of installed packages


raszi:
**Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

# Please always follow these steps:
- [X] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at the relevant tap e.g. for Homebrew/homebrew-core https://github.com/Homebrew/homebrew-core/issues/new
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed all issues and retried your prior step?
- [X] Ran `brew config` and `brew doctor` and included their output with your issue?

# Outputs

## brew doctor

```
~ $ brew doctor
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: Homebrew's sbin was not found in your PATH but you have installed
formulae that put executables in /usr/local/sbin.
Consider setting the PATH for example like so
  echo 'export PATH="/usr/local/sbin:$PATH"' >> ~/.bash_profile
```

This is fine I am using fish.

## brew config

```
~ $ brew config
HOMEBREW_VERSION: 1.3.1
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 69799d97b1e7314912b2ee234dec2c179c5fb969
Last commit: 3 weeks ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 420482004c6f5f43e6f1c25bc00a53900ecd9ce7
Core tap last commit: 2 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: quad-core 64-bit haswell
Homebrew Ruby: 2.0.0-p648
Clang: 8.1 build 802
Git: 2.14.1 => /usr/local/bin/git
Perl: /usr/bin/perl
Python: /usr/bin/python
[...]

PR #3344, opened 2017-10-20
leaves: account for requirements

maxim-belkin:

-----

Fixes the problem with `leaves` described in #3100.

There is a potential to combine the new `each do` loop with the one right above it<sup>*</sup> but (for now) I decided just to post the solution. 

<sup>*</sup> For example, by resolving requirements to dependencies in the first place and then looping over formula dependencies together with the dependencies emerging from requirements.

### DIFF
--- a/Library/Homebrew/cmd/leaves.rb
+++ b/Library/Homebrew/cmd/leaves.rb
@@ -23,7 +23,19 @@ module Homebrew
         end
       end
 
-      deps_of_installed.merge(deps)
+      reqs = []
+
+      f.requirements.to_a.each do |req|
+        dep = req.to_dependency
+        next if dep.nil?
+        if req.optional? || req.recommended?
+          reqs << dep.to_formula.full_name if f.build.with?(req)
+        else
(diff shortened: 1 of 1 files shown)


maxim-belkin:
Before:
<pre>
$ brew leaves
atk
fdupes
<b>gcc</b>
gdk-pixbuf
gnu-sed
hicolor-icon-theme
jpegoptim
p7zip
pango
<b>perl</b>
pstree
pyenv
<b>python</b>
r
rbenv
<b>ruby</b>
tree
vim
</pre>
_Note gcc, perl, python, ruby_

After:
```
$ brew leaves
atk
fdupes
gdk-pixbuf
gnu-sed
hicolor-icon-theme
jpegoptim
p7zip
pango
pstree
pyenv
r
rbenv
tree
vim
```

MikeMcQuaid:
While this is an improvement I think ideally this would use `tab.runtime_dependencies` or `Keg.find_some_installed_dependents`.

PR #3349, opened 2017-10-21
Rubocop 0.51

reitermarkus:

-----

Update RuboCop to 0.51 and fix/refactor specs.


Library/Homebrew/rubocops/conflicts_cop.rb
```diff
@@ -6,10 +6,8 @@ module Cop
     module FormulaAudit
       # This cop audits versioned Formulae for `conflicts_with`
       class Conflicts < FormulaCop
-        MSG = <<~EOS.freeze
-          Versioned formulae should not use `conflicts_with`.
-          Use `keg_only :versioned_formula` instead.
-        EOS
+        MSG = "Versioned formulae should not use `conflicts_with`. " \
```
DomT4: Struggling to see how a bunch of backslashes with heavier indentation are any more readable than an `EOS` block, personally 🙈.

reitermarkus: It needs to be a single line for it to work with `expect_offense`, that's the reason. `EOS` was two lines.

DomT4: Does something like `.split("\n")` not work?

reitermarkus: No, it only supports a single line. Look at one of the specs where `expect_offense` is used. It parses the lines with `^^^^^^^^` and matches them.

DomT4: I think this is a fairly profound regression on readability, and honestly something that would deter me from wanting to spend much time with these files, but it is what it is. Mike's approved it & I stepped away, so I'll go quietly.

ilovezfs: @DomT4 yes this seems weird but note two things:

1. most of the code here is still using heredocs
2. the diff here is +375 −1,167

That said, I too am struggling to understand why rubocop is forcing us into this one line situation.

@reitermarkus is this something specific to 0.51 or have we had a bug all along? If it's the former, that seems like a pretty yucky regression that we should engage with upstream about. If the latter, oh well.

Library/Homebrew/test/rubocops/caveats_cop_spec.rb
```diff
@@ -1,41 +1,20 @@
...
-require_relative "../../extend/string"
 require_relative "../../rubocops/caveats_cop"
 
 describe RuboCop::Cop::FormulaAudit::Caveats do
   subject(:cop) { described_class.new }
 
   context "When auditing caveats" do
     it "When there is setuid mentioned in caveats" do
-      source = <<~EOS
+      expect_offense(<<~EOS)
```
ilovezfs: why is this one EOS not RUBY?

Library/Homebrew/rubocops/conflicts_cop.rb
MikeMcQuaid: @reitermarkus Is `expect_offense` not something we define ourselves in `Library/Homebrew/test/support/helper/rubocop.rb`?

ilovezfs:
@reitermarkus I notice that you've changed several structured expectations like
```
      expected_offenses = [{  message: "sha256 should be 64 characters",
                              severity: :convention,
                              line: 5,
                              column: 12,
                              source: source },
                           {  message: "sha256 should be 64 characters",
                              severity: :convention,
                              line: 9,
                              column: 14,
                              source: source }]
```

into exact string matches that involve things like
```
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Is this a mandatory change dictated by 0.51 or something you're electing to do, that we could choose not to do? I think we're going backwards in terms of maintainability and readability moving from the former to the latter, even if it is fewer lines.

Library/Homebrew/rubocops/conflicts_cop.rb
reitermarkus: We had `expect_offense` defined ourselves before, but now RuboCop's `rspec` support includes a `expect_offense` helper method, which they use for all of their specs.

reitermarkus:
> Is this a mandatory change dictated by 0.51

Not really mandatory, but none of the RuboCop specs is using the old way anymore, so this is basically depending on undocumented behaviour.

Library/Homebrew/test/rubocops/caveats_cop_spec.rb
reitermarkus: Must have gotten tired of copy-pasting. 😄

Library/Homebrew/rubocops/conflicts_cop.rb
ilovezfs: >Does something like .split("\n") not work?

@reitermarkus I think @DomT4 meant something like
```
        MSG = <<~EOS.chomp.tr("\n", " ").freeze
          Versioned formulae should not use `conflicts_with`.
          Use `keg_only :versioned_formula` instead.
        EOS
```

or

```
        MSG = <<~EOS.tr("\n", " ").strip.freeze
          Versioned formulae should not use `conflicts_with`.
          Use `keg_only :versioned_formula` instead.
        EOS
```
or equivalent.

Library/Homebrew/constants.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 # RuboCop version used for `brew style` and `brew cask style`
-HOMEBREW_RUBOCOP_VERSION = "0.50.0"
```
MikeMcQuaid: Can we put these in the `Gemfile` instead now?

MikeMcQuaid:
> We had expect_offense defined ourselves before, but now RuboCop's rspec support includes a expect_offense helper method, which they use for all of their specs.

I'm in favour of using RuboCop's version rather than ours here even if this means a slightly less structured approach.

Library/Homebrew/constants.rb
reitermarkus: Not in this PR.

Library/Homebrew/rubocops/conflicts_cop.rb
reitermarkus: That doesn't seem more readable to me.

ilovezfs: Personally I think some junk on the first line is better than quotes and backslashes at the end of every line but I will leave that between you and Dom.

DomT4: > I think @DomT4 meant something like

Yeah I didn't actually do testing, I was just tossing ideas off of the top of my head in terms of could we not do this literally any other way.

> That doesn't seem more readable to me.

I agree it's not brilliant but at least from my point of view there is nothing worse than backslashes for readability. For me there is an appreciable difference between:

```ruby
blah "Polly ate three crackers and two of those had butter on " \
     "except one had salted butter and the other unsalted. " \
     "Polly wasn't sure why she was documenting her lunch to " \
     "such a degree but wondered if she should join Instagram " \
     "where documenting lunch seems to be very in vogue."
```
And:
```ruby
blah <<~EOS.tr("\n", " ").strip.freeze
  Polly ate three crackers and two of those had butter on
  except one had salted butter and the other unsalted.
  Polly wasn't sure why she was documenting her lunch to
  such a degree but wondered if she should join Instagram
  where documenting lunch seems to be very in vogue.
EOS
```
The latter has a flow that isn't punched in the face by having to check for backslashes constantly. The longer the block of text the worse it becomes. And the latter also avoids that grim necessity for whitespace before the closing `"` on each line except the final one.

reitermarkus: It may be worth it for many lines, but not for only 2 as in this case.

DomT4: But then you're simply introducing inconsistency across a codebase 🙈.

PR #3352, opened 2017-10-22
java_requirement: support prompting users to install legacy Java casks

JCount:

-----
This adds support for prompting the user to install an older, "legacy" java cask when required.

Instead of always attempting to satisfy the requirement with the default java cask, currently Java 9, this prompts the user to install a specific java cask version as needed.

### DIFF
--- a/Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
+++ b/Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
@@ -1,12 +1,29 @@
 class JavaRequirement < Requirement
-  cask "java"
-
   env do
     env_java_common
     env_oracle_jdk || env_apple
   end
 
+  # A strict Java 8 requirement (1.8) should prompt the user to install
+  # the legacy java8 cask because the current version, Java 9, is not
+  # completely backwards compatible, and contains breaking changes such as
+  # strong encapsulation of JDK-internal APIs and a modified version scheme
+  # (9.0 not 1.9).
+  def cask
+    if @version.nil? || @version.to_s.end_with?("+") ||
+       @version.to_f >= JAVA_CASK_MAP.keys.max.to_f
+      JAVA_CASK_MAP.fetch(JAVA_CASK_MAP.keys.max)
+    else
+      JAVA_CASK_MAP.fetch("1.8")
+    end
+  end
+
   private
+
+  JAVA_CASK_MAP = {
+    "1.8" => "caskroom/versions/java8",
+    "9.0" => "java",
+  }.freeze
 
   def possible_javas
     javas = []


Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -34,4 +41,15 @@ def env_apple
   def oracle_java_os
     :darwin
   end
+
+  JAVA_CASK_MAP = {
+    "1.6" => "caskroom/versions/java6",
```
MikeMcQuaid: For the sake of not relying on these: can we limit this to Java 8?

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -34,4 +41,15 @@ def env_apple
...
+    "1.7" => "caskroom/versions/zulu7",
+    "1.8" => "caskroom/versions/java8",
+    "9.0" => "caskroom/cask/java",
+  }.freeze
+
+  def legacy_java_cask(cask_version = "1.8")
+    @cask = JAVA_CASK_MAP.fetch(cask_version.to_s)
```
MikeMcQuaid: Feels a bit weird to have `cask "java"` and setting this here, too? Also, `caskroom/cask/java` doesn't agree with the version above.

MikeMcQuaid:
Thanks for this, @JCount!

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
JCount: Looking at this now, the reasoning I had for having `"1.8"` as the default `cask_version` makes no sense.

Combined with your comment about limiting it to `Java 8` and not setting it in two places, I am considering refactoring this to completely replace setting the `cask` variable, even for the default case when a version tag isn't passed.

JCount:
Note that the changes I made are very rough around the edges, and will probably need a fair amount of polishing through code review. 🙂

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -1,13 +1,28 @@
...
+    super
+  end
 
   env do
     env_java_common
     env_oracle_jdk || env_apple
   end
 
+  def cask
+    which_cask_needed?
```
MikeMcQuaid: Any reason `which_cask_needed?` can't fill this method?

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -34,4 +49,13 @@ def env_apple
   def oracle_java_os
     :darwin
   end
+
+  def which_cask_needed?
+    if @version.nil? || @version.to_s.end_with?("+") ||
+       @version.to_f >= CURRENT_JAVA.to_f
+      JAVA_CASK_MAP.fetch(CURRENT_JAVA.to_s)
+    else
+      JAVA_CASK_MAP.fetch("1.8")
```
MikeMcQuaid: A comment explaining the 9.0/1.8 situation just now would be great 👍

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
JCount: Nope, good idea! 👍

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
JCount: 👍

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -1,13 +1,34 @@
...
+    @cask = cask
+    super
+  end
 
   env do
     env_java_common
     env_oracle_jdk || env_apple
   end
 
+  # A strict Java 8 requirement (1.8) should prompt the user to install java8
```
MikeMcQuaid: This describes what it does but not why it's doing what it's doing. The comment is relatively obvious from reading the code but it would be good to explain the whole "Java is usually backwards compatible but 9.0 is causing everything to explode due to breaking changes and changing version number format" stuff here instead.

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -1,13 +1,34 @@
 class JavaRequirement < Requirement
-  cask "java"
+  def initialize(tags)
```
ilovezfs: Is this custom initialize actually necessary?

The initialize for `requirements/java_requirement` is

```
  def initialize(tags)
    @version = tags.shift if /(\d+\.)+\d/ =~ tags.first
    super
  end
```
and for requirement is
```
  def initialize(tags = [])
    @default_formula = self.class.default_formula
    @cask ||= self.class.cask
    @download ||= self.class.download
    @formula = nil
    tags.each do |tag|
      next unless tag.is_a? Hash
      @cask ||= tag[:cask]
      @download ||= tag[:download]
    end
    @tags = tags
    @tags << :build if self.class.build
    @name ||= infer_name
  end
```
so I think this is redundant unless I've missed some detail.

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -1,13 +1,34 @@
...
+       @version.to_f >= CURRENT_JAVA.to_f
+      JAVA_CASK_MAP.fetch(CURRENT_JAVA.to_s)
+    else
+      JAVA_CASK_MAP.fetch("1.8")
+    end
+  end
+
   private
 
+  CURRENT_JAVA = 9.0
```
ilovezfs: @MikeMcQuaid should we retrieve this as `JAVA_CASK_MAP.keys.max` instead of having a separate variable?

MikeMcQuaid: Avoiding another variable would be ideal 👍

Library/Homebrew/extend/os/mac/requirements/java_requirement.rb
```diff
@@ -1,13 +1,30 @@
 class JavaRequirement < Requirement
-  cask "java"
```
MikeMcQuaid: The `@cask` doesn't seem to be set any more, though. Does the cask message still get output as expected now? If so: 🆒 and ignore me.

JCount: The base `Requirement` class has:
```
class Requirement
  include Dependable

  attr_reader :tags, :name, :cask, :download, :default_formula

  def initialize(tags = [])
    @default_formula = self.class.default_formula
    @cask ||= self.class.cask
    @download ||= self.class.download
    @formula = nil
    tags.each do |tag|
      next unless tag.is_a? Hash
      @cask ||= tag[:cask]
      @download ||= tag[:download]
    end
    @tags = tags
    @tags << :build if self.class.build
    @name ||= infer_name
  end
```

MikeMcQuaid: :cool:. If it works and you're happy with it: 🚢

PR #3366, opened 2017-10-25
lines_cop: Fix detection of negated expression

GauthamGoli:

-----
In chained conditional statements if any one condition is negated, false positives were being raised. 
Example: `build.with?("libcxx") || !MacOS::CLT.installed?` in `llvm` formula

### DIFF
--- a/Library/Homebrew/rubocops/extend/formula_cop.rb
+++ b/Library/Homebrew/rubocops/extend/formula_cop.rb
@@ -316,8 +316,10 @@ module RuboCop
       end
 
       # Check if negation is present in the given node
-      def negated?(node)
-        method_called?(node, :!)
+      def expression_negated?(node)
+        return false if node.parent.nil?
+        return false unless node.parent.method_name.equal?(:!)
+        offending_node(node.parent)
       end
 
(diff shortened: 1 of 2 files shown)


Library/Homebrew/rubocops/extend/formula_cop.rb
```diff
@@ -316,8 +316,12 @@ def component_precedes?(first_node, next_node)
       end
 
       # Check if negation is present in the given node
-      def negated?(node)
-        method_called?(node, :!)
+      def expression_negated?(node)
+        if !node.parent.nil? && node.parent.method_name.equal?(:!)
```
MikeMcQuaid:
```
return false if node.parent.nil?
return false unless node.parent.method_name.equal?(:!)
```
and then you can skip the `return true` and indented `if`. In general checking for "bad" input and exiting early is a nice practice.

GauthamGoli: @MikeMcQuaid Updated the PR.

MikeMcQuaid:
Thanks again @GauthamGoli!

PR #3377, opened 2017-10-26
pathname: new methods and improvements

maxim-belkin:

-----

CC @sjackman 

`extend/pathname.rb`

- `atomic_write`: close file before renaming to prevent 'Device or resource busy' error
- `ensure_writable`: preserve executable bit

`extend/os/pathname.rb`
`extend/os/linux/extend/pathname.rb`
    - new `elf?` and `dynamic?` methods


Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,12 @@
+class Pathname
+  # @private
+  def elf?
+    read(4) == "\x7fELF"
+  end
+
+  # @private
+  def dynamic?
+    @which_file ||= which("file")
+    !@which_file.nil? && `#{@which_file} -b #{self}` =~ /dynamic|shared/
```
sjackman: I suggest…
```ruby
@which_file ||= which "file"
return false unless @which_file
popen_read(@which_file, "-b", to_path) =~ /dynamic|shared/
```

Library/Homebrew/extend/pathname.rb
```diff
@@ -189,6 +189,7 @@ def atomic_write(content)
       rescue Errno::EPERM # rubocop:disable Lint/HandleExceptions
       end
 
+      tf.close # Close the file before renaming to prevent 'Device or resource busy' error
```
sjackman: I suggest…
```ruby
# Close the file before renaming to prevent the error: Device or resource busy
tf.close
```

sjackman:
A few style suggestions, otherwise looks good, Maxim!

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,12 @@
+class Pathname
+  # @private
+  def elf?
+    read(4) == "\x7fELF"
+  end
+
+  # @private
+  def dynamic?
```
MikeMcQuaid: Is this for a dynamic library? If so, would be good to be maybe `dynamic_library?` or similar?

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,12 @@
+class Pathname
+  # @private
+  def elf?
+    read(4) == "\x7fELF"
```
MikeMcQuaid: This magic warrants a comment or citation.

Library/Homebrew/extend/pathname.rb
MikeMcQuaid: When does this error occur? Is it not Linux-only?

Library/Homebrew/extend/pathname.rb
```diff
@@ -189,6 +189,7 @@ def atomic_write(content)
...
       File.rename(tf.path, self)
     ensure
       tf.close!
```
MikeMcQuaid: Is there no error if this is closed twice?

Library/Homebrew/extend/pathname.rb
```diff
@@ -376,7 +377,7 @@ def ensure_writable
     saved_perms = nil
     unless writable_real?
       saved_perms = stat.mode
-      chmod 0644
+      chmod "u+rw"
```
MikeMcQuaid: Motivation for this change?

Library/Homebrew/extend/pathname.rb
sjackman: It affects primarily NFS, which I believe can affect macOS as well.

Library/Homebrew/extend/os/linux/extend/pathname.rb
sjackman: Here's a citation: https://en.wikipedia.org/wiki/List_of_file_signatures

Library/Homebrew/extend/os/linux/extend/pathname.rb
sjackman: Both executables and libraries may be dynamic. I'd suggest `dynamic_binary` or `dynamic_elf`?

Library/Homebrew/extend/pathname.rb
sjackman: Closing the file twice does not produce an error.

Library/Homebrew/extend/pathname.rb
sjackman: As I recall, it related to modifying an executable script, and avoiding removing the executable bit.

Library/Homebrew/extend/os/linux/extend/pathname.rb
maxim-belkin: I don't think we should return `false` because that would be incorrect.
We should, perhaps, `raise StandardError.new("file command is not available")` error in that case.
I'm also considering adding a `if/else`:
`if which("file")... elsif which("readelf") ... elseif which("ldd") ... else raise...`

Library/Homebrew/extend/pathname.rb
maxim-belkin: should I add a comment like `# preserve executable bit for executable scripts`

maxim-belkin: ?

Library/Homebrew/extend/os/linux/extend/pathname.rb
sjackman: The current behaviour is to return `false` when `file` is not available. I understand your objection though. Trying first `readelf` and then `file` seems reasonable.

Library/Homebrew/extend/os/linux/extend/pathname.rb
MikeMcQuaid: Perfect, this would be great in a comment.

Library/Homebrew/extend/os/linux/extend/pathname.rb
MikeMcQuaid: Either of the latter two sound good to me.

Library/Homebrew/extend/os/linux/extend/pathname.rb
MikeMcQuaid: > We should, perhaps, raise StandardError.new("file command is not available") error in that case.

Something like this sounds good or the `if`/`else`.

Library/Homebrew/extend/pathname.rb
MikeMcQuaid: Ok. Would be good to add that to the comment, too.

Library/Homebrew/extend/pathname.rb
MikeMcQuaid: Executable scripts without shebangs are intentionally made non-executable unless `skip_clean` is used.

Library/Homebrew/extend/os/linux/extend/pathname.rb
sjackman: I suggest `dynamic_elf?`.

ilovezfs: `keebler_elf?`

Library/Homebrew/extend/pathname.rb
sjackman: I'm talking about executable scripts with a shebang line and that have the executable bit set. `chmod 0644` removes the executable bit, which is undesirable. `chmod "u+rw"` adds the read and write bits without modifying the executable bit. If it's set it remains set. If it's clear, it remains clear. This function is called `ensure_writable`.

Library/Homebrew/extend/os/linux/extend/pathname.rb
sjackman: 😂 You win! Make it so.

Library/Homebrew/extend/pathname.rb
MikeMcQuaid: My mistake; I guess I'm trying to figure out why this is something that worked fine as-is on macOS but is causing Linuxbrew issues?

sjackman: The saved permissions are restored later
https://github.com/Homebrew/brew/pull/3377/files#diff-06540c12bb8e84922c0d329ed05b19bfR384
```ruby
chmod saved_perms if saved_perms
```
It's only transient that the script loses the executable bit.

maxim-belkin: I'm still not comfortable running `chmod u+rw` in a method called _ensure\_**writable**_.  Either it should be `chmod u+w` or a different method name.

sjackman: The callers of this function read the file, modify it, and write it back to disk, so they need both read and write access. Perhaps it's a bit of a misnomer. I'm okay with that. Permission to write to a file usually implies permission to read a file, except in only the most unusual of exceptions.

maxim-belkin: I could argue... but arguing without real-life examples, which I don't have handy, is useless, so, I'll leave it "as is" :)

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,13 @@
+class Pathname
+  # @private
+  def elf?
+    # List of file signatures: https://en.wikipedia.org/wiki/List_of_file_signatures
```
maxim-belkin: should I add a pointer to ` /usr/share/file/magic` directory and/or `man magic`?

sjackman: One reference is good enough. The ELF article on Wikipedia is a better reference though:
https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header

Library/Homebrew/extend/os/linux/extend/pathname.rb
sjackman: If you're up for it, it wouldn't be too terribly difficult to parse the ELF header in Ruby code and determine whether the file is static or dynamic. The information isn't unfortunately at a fixed position, but it's not difficult to parse. Iterate through the vector of `Program header` entries. If you see a `PT_DYNAMIC` header, it's dynamic, otherwise it's static.

See https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#Program_header

If you're not into it, I don't have the time to implement a PR this week, but I could next week. I'd prefer to avoid having `brew` depend on external commands which may or may not be available.

Library/Homebrew/extend/os/linux/extend/pathname.rb
vitorgalvao: It might be useful to link to a specific revision: https://en.wikipedia.org/wiki/List_of_file_signatures?oldid=807859331

sjackman: The current ELF document is the better reference in my opinion.

maxim-belkin:
One of the error messages https://travis-ci.org/Homebrew/brew/jobs/294948205#L945:
```
1) Pathname#ensure_writable makes a file writable and restores permissions afterwards
     Failure/Error:
       file.ensure_writable do
         expect(file).to be_writable
       end
     TypeError:
       no implicit conversion of String into Integer
```

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -6,8 +6,19 @@ def elf?
   end
 
   # @private
-  def dynamic?
-    @which_file ||= which("file")
-    !@which_file.nil? && `#{@which_file} -b #{self}` =~ /dynamic|shared/
+  def dynamic_elf?
+    if !which("readelf").nil?
+      needed = `readelf -d #{self} | grep NEEDED`
+      return false if needed.empty?
```
sjackman: I'd suggest…
```ruby
popen_read("readelf", "-l", to_path)[/ DYNAMIC /]
```

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -6,8 +6,19 @@ def elf?
...
+    elsif !which("objdump").nil?
+      needed = `objdump -p #{self} | grep NEEDED`
+      return false if needed.empty?
```
sjackman: Since both `readelf` and `objdump` are provided by `binutils`, it's unlikely that the user would have one and not the other. I'd be inclined to get rid of `objdump`.

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -6,8 +6,19 @@ def elf?
...
+    elsif !which("file").nil?
+      match = `file -L -b #{self}` =~ /dynamic|shared/
+      return false if match.nil?
+    else
+      raise StandardError, "No tools available to detect dynamically linked ELF files."
```
sjackman: I'd suggest...
```
Neither `readelf` nor `file` is available to determine whether an ELF file is dynamically or statically linked.
```

sjackman:
```
 no implicit conversion of String into Integer
```
As far as I can tell, `FileUtils.chmod` accepts a symbolic string mode, but `Pathname.chmod` does not and `File.chmod` does not. You can instead use…
```ruby
chmod 0644 | saved_perms
```

maxim-belkin:
`chmod 0644 | saved_perms` is too cryptic. I'll try `FileUtils.chmod "u+rw", to_path`

Library/Homebrew/extend/os/linux/extend/pathname.rb
maxim-belkin: > it's unlikely that the user would have one and not the other. I'd be inclined to get rid of `objdump`.

macOS users aside - yes. I included it because it works with more formats and is as safe as `readelf`.

Library/Homebrew/extend/os/linux/extend/pathname.rb
maxim-belkin: ok

Library/Homebrew/extend/os/linux/extend/pathname.rb
maxim-belkin: ok

maxim-belkin:
[codecov/patch](https://codecov.io/gh/Homebrew/brew/compare/badbb00b0867846afa6d55275bca4edff959e908...fb37f12a0198a6e6db4f1a74a6e0cc5d6f790fa8) shows reduced coverage but otherwise success

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,21 @@
+class Pathname
+  # @private
+  def elf?
+    # List of file signatures: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
```
sjackman: Change `List of file signatures:` to `See` or `ELF file format:`

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,21 @@
...
+    read(4) == "\x7fELF"
+  end
+
+  # @private
+  def dynamic_elf?
+    if !which("readelf").nil?
+      match = popen_read("readelf", "-l", to_path)[/ DYNAMIC /]
+      return false if match.empty?
```
sjackman: I believe this suffers from `NoMethodError: undefined method `empty?' for nil:NilClass` when it doesn't match.
I suggest…
```ruby
return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ")
```
Sorry this contradicts my previous advice.

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,21 @@
...
+    elsif !which("file").nil?
+      match = `file -L -b #{self}` =~ /dynamic|shared/
+      return false if match.nil?
```
sjackman: I'd suggest…
```ruby
return !popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil?
```
@MikeMcQuaid What's your preferred style in this case?

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,21 @@
...
+    else
+      raise StandardError, "Neither `readelf` nor `file` is available to determine whether '#{self}' is dynamically or statically linked."
+    end
+    true
```
sjackman: Remove this line `true`.

Library/Homebrew/extend/os/pathname.rb
```diff
@@ -0,0 +1 @@
+require "extend/os/linux/extend/pathname" if OS.linux?
```
sjackman: The equivalent file on macOS is `os/mac/pathname.rb`. That file should probably be moved to `extend/os/mac/extend/pathname.rb`

Library/Homebrew/extend/os/linux/extend/pathname.rb
maxim-belkin: done

Library/Homebrew/extend/os/linux/extend/pathname.rb
maxim-belkin: done

Library/Homebrew/extend/os/linux/extend/pathname.rb
maxim-belkin: done

Library/Homebrew/extend/os/pathname.rb
maxim-belkin: done

maxim-belkin:
https://travis-ci.org/Homebrew/brew/jobs/295748996#L846
```
Offenses:
Library/Homebrew/extend/os/linux/extend/pathname.rb:10:5: C: Use a guard clause instead of wrapping the code inside a conditional expression.
    if !which("readelf").nil?
    ^^
Library/Homebrew/extend/os/linux/extend/pathname.rb:11:7: C: Redundant return detected.
      return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ")
      ^^^^^^
Library/Homebrew/extend/os/linux/extend/pathname.rb:13:7: C: Redundant return detected.
      return !popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil?
      ^^^^^^
663 files inspected, 3 offenses detected
```

So, I will change `dynamic_elf?` from:

```ruby
  def dynamic_elf?
    if !which("readelf").nil?
      return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ")
    elsif !which("file").nil?
      return !popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil?
    else
      raise StandardError, "Neither `readelf` nor `file` is available to determine whether '#{self}' is dynamically or statically linked."
    end
```
to
```ruby
  def dynamic_elf?
    return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ") unless which("readelf").nil?
    return !popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil? unless which("file").nil?
    raise StandardError, "Neither `readelf` nor `file` is available to determine whether '#{self}' is dynamically or statically linked."
  end 
```

Library/Homebrew/extend/os/mac/hardware/cpu.rb
```diff
@@ -1,4 +1,4 @@
-require "os/mac/pathname"
+require "extend/os/mac/extend/pathname"
```
sjackman: This line doesn't look like it should go here. I'm guessing that it preexisted `extend/os/mac`. Can you try removing this line, and adding it to `extend/os/pathname.rb`?

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -7,12 +7,8 @@ def elf?
...
   def dynamic_elf?
-    if !which("readelf").nil?
-      return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ")
-    elsif !which("file").nil?
-      return !popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil?
-    else
-      raise StandardError, "Neither `readelf` nor `file` is available to determine whether '#{self}' is dynamically or statically linked."
-    end
+    return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ") unless which("readelf").nil?
+    return !popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil? unless which("file").nil?
```
sjackman: Nice! I like it. Try…
```ruby
… if which "readelf"
… if which "file"
```

maxim-belkin: ok

Library/Homebrew/extend/os/mac/hardware/cpu.rb
maxim-belkin: I'm not sure I understand...
I can add `require "extend/os/mac/extend/pathname" if OS.mac?` to `extend/os/pathname.rb?` but do you want this line removed from `cpu.rb`?

maxim-belkin: got it

sjackman: > do you want this line removed from cpu.rb?

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,14 @@
...
+  def elf?
+    # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+    read(4) == "\x7fELF"
+  end
+
+  # @private
+  def dynamic_elf?
+    return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ") if which "readelf"
+    return !popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil? if which "file"
+    raise StandardError, "Neither `readelf` nor `file` is available to determine whether '#{self}' is dynamically or statically linked."
```
MikeMcQuaid: Can we get these lines wrapping at about 80 characters?

Library/Homebrew/extend/os/linux/extend/pathname.rb
```diff
@@ -0,0 +1,14 @@
+class Pathname
+  # @private
+  def elf?
+    # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+    read(4) == "\x7fELF"
+  end
+
+  # @private
+  def dynamic_elf?
+    return popen_read("readelf", "-l", to_path).include?(" DYNAMIC ") if which "readelf"
```
MikeMcQuaid: An `if`/`elsif` in block form would make this easier to follow, I think.

Library/Homebrew/extend/os/pathname.rb
MikeMcQuaid: `if/`elsif`

Library/Homebrew/extend/os/mac/hardware/cpu.rb
```diff
@@ -1,5 +1,3 @@
-require "os/mac/pathname"
```
MikeMcQuaid: ?

Library/Homebrew/extend/pathname.rb
```diff
@@ -189,6 +189,9 @@ def atomic_write(content)
       rescue Errno::EPERM # rubocop:disable Lint/HandleExceptions
       end
 
+      # Close the file before renaming to prevent the error: Device or resource busy
+      # Affects primarily NFS. Can affect macOS
```
MikeMcQuaid: Can remove `Can affect macOS`

Library/Homebrew/os/mac.rb
```diff
@@ -3,7 +3,7 @@
 require "os/mac/version"
 require "os/mac/xcode"
 require "os/mac/xquartz"
-require "os/mac/pathname"
+require "extend/os/mac/extend/pathname"
```
MikeMcQuaid: Feels weird to require deeply here.

Library/Homebrew/extend/os/linux/extend/pathname.rb
sjackman:
```
C: Use a guard clause instead of wrapping the code inside a conditional expression.
C: Redundant return detected.
```
@maxim-belkin If you go back to the `if/else` code that you had before, and remove the `return` statements, it should clear up both of these style errors.

Library/Homebrew/os/mac.rb
sjackman: My opinion…
`os/mac.rb` should not require `extend/os/mac/extend/pathname.rb`
`extend/pathname.rb` should require `extend/os/pathname.rb`

Library/Homebrew/extend/os/mac/hardware/cpu.rb
sjackman: I think this line was an anachronism, before `extend/os` existed. My opinion…
`extend/os/mac/hardware/cpu.rb` should not require `os/mac/pathname.rb`
`extend/pathname.rb` should require `extend/os/pathname.rb` which requires `extend/os/mac/extend/pathname.rb`

maxim-belkin: Question: is there a way to detect _anachronisms_ like this? That is, something is required but is not, in fact, necessary.

Library/Homebrew/extend/pathname.rb
maxim-belkin: done

maxim-belkin:
OK, here is the current dependency "diagram":

- `extend/pathname`    requires   `extend/os/pathname`
- `extend/os/pathname`    requires `extend/os/<OS>/extend/pathname`

maxim-belkin:
@sjackman @MikeMcQuaid 
is there anything else you guys want me to do/change in this PR?

MikeMcQuaid:
Nice work here, thanks again @maxim-belkin!

sjackman:
Thanks, Maxim! and thanks Mike for feedback and for merging. 👏

maxim-belkin:
yay!

Issue #3007, opened 2017-08-05
brew uses should be (much) faster


MikeMcQuaid:
The algorithm used by `brew uses` ends up being really slow. Through a new algorithm, rewrite or heavy caching it would be good to make it much faster. [`brew prof uses`](https://github.com/Homebrew/homebrew-dev-tools/blob/master/cmd/brew-prof.rb) is probably the best way to optimise this.

richiethomas:
Hi @MikeMcQuaid, I'm new to the Homebrew project and shell scripting as well, but I'm keen to take a crack at this if possible.

richiethomas:
UPDATE: I see the `uses` script is written in Ruby, so I'm guessing I'd learn less shell scripting than a pure BaSH script.  That's fine though, still very keen to pitch in on this issue.

MikeMcQuaid:
@richiethomas Yeh, we'd welcome any work on it. Feel free to open a PR!

richiethomas:
Thanks @MikeMcQuaid.  You mentioned caching as a potential solution, can you give me an example of where caching is currently used in the codebase so that I don't deviate too far from the best practice?

MikeMcQuaid:
@richiethomas I general the caching we do in cases like this looks a bit like:

```
@some_hash[formula_name] ||= begin
  # expensive operation here
end
```

In this case, doing effectively a `brew readall` and converting the entire formula tree into a hash seems like it may be a sensible approach and much quicker to query. For bonus points: make `brew readall` faster too will make this faster too.

richiethomas:
Hey @MikeMcQuaid, I noticed this doesn't appear to be current practice, but do you think it would be a wise move to store the entire formula tree locally, like in a JSON file in the Homebrew directory, in order to reduce the compile time of the formula tree?  The tradeoff would be adding that little extra work up from when adding or removing Brew formulas, but saving the compile time (which is substantial) on the `readall` and `uses` call, essentially making it into just a straight read operation on that JSON hash (which would theoretically be O(1) or close to it).  I'm sure you've thought of this before, just wondering if there were concerns about that approach that I as a newer dev aren't seeing.  Looking forward to your feedback!

MikeMcQuaid:
@richiethomas There's similar discussion in https://github.com/Homebrew/brew/issues/3072.

[...]

PR #3379, opened 2017-10-27
WIP: Implement on-disk formula cache

MichaelSquires:

-----
This PR addresses issue #3007. One of the reasons commands like `brew uses` and `brew readall` may be slow is because they load all formulae in all installed taps. The slowness is a result of ruby having to open and interpret thousands of formulae. In order to speed this process up, this PR adds functionality which pre-compiles all formulae and stores the ruby byte code on-disk in a marshaled hash. 

The formula store is saved at `HOMEBREW_CACHE/formula_store.bin`. The format of the formula store is an on-disk marshaled hash with the following format:
```
{
  <path to formula1> => {
    "byte_code" => <bytecode of formula stored in binary>,
    "extra_data" => <formula data after the __END__>,
    "mtime" => <integer of modified time>
  },
  ...
  <path to formulaN> => {
    "byte_code" => <bytecode of formula stored in binary>,
    "extra_data" => <formula data after the __END__>,
    "mtime" => <integer of modified time>
  },
}
```

Potential cache invalidation issues have been solved by storing the modified time of each formula in the formula store. When loading formulae from the store, the modified time is compared with the modified time of the file. If they are different, the store is updated so it will be up-to-date during the next run. Additionally, the store is updated when `brew tap`, `brew untap`, or `brew update` are run. 

Local testing has shown an approximate 40% performance gain when running commands which iterate over all loaded formula. 

An additional command, `brew update-store`, was added for generating/updating the store.

### DIFF
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -32,6 +32,7 @@
 #:    List all pinned taps.
 
 require "tap"
+require "formula_store"
 
 module Homebrew
   module_function
@@ -52,6 +53,8 @@ module Homebrew
         tap.install clone_target: ARGV.named[1],
                     full_clone: full_clone?,
                     quiet: ARGV.quieter?
+
+        FormulaStore.store_tap tap
       rescue TapRemoteMismatchError => e
         odie e
       rescue TapAlreadyTappedError, TapAlreadyUnshallowError # rubocop:disable Lint/HandleExceptions
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -54,7 +54,7 @@ module Homebrew
                     full_clone: full_clone?,
                     quiet: ARGV.quieter?
 
-        FormulaStore.store_tap tap
+        FormulaStore.store_tap tap if ENV["HOMEBREW_EXPERIMENTAL_FORMULA_STORE"]
       rescue TapRemoteMismatchError => e
         odie e
       rescue TapAlreadyTappedError, TapAlreadyUnshallowError # rubocop:disable Lint/HandleExceptions
--- a/Library/Homebrew/cmd/untap.rb
+++ b/Library/Homebrew/cmd/untap.rb
@@ -2,6 +2,7 @@
 #:    Remove a tapped repository.
 
 require "tap"
+require "formula_store"
 
 module Homebrew
   module_function
@@ -12,6 +13,7 @@ module Homebrew
     ARGV.named.each do |tapname|
       tap = Tap.fetch(tapname)
       raise "untapping #{tap} is not allowed" if tap.core_tap?
+      FormulaStore.unstore_tap tap
       tap.uninstall
     end
   end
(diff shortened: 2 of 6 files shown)


MichaelSquires:
I'm still working on writing some new tests for the added functionality. I thought this would be a good time to get a PR started.

MikeMcQuaid:
@MichaelSquires Let me know when you'd like some more detailed code review, here. For now could you have everything related to the cached guarded with an environment variable e.g. `ENV["HOMEBREW_EXPERIMENTAL_CACHE"]` or similar? This will allow us to ship this and test it further before enabling it for everyone.

A random other comment: it may be worth experimenting with something like a SQLite database for this and/or the output of `brew info --json` if that's sufficient to be able to calculate this dependency data. Storing structured data on disk should be similar or superior for speed and would be vastly superior for usage in other contexts. Just a thought!

MichaelSquires:
>Let me know when you'd like some more detailed code review, here. 

Thanks, I'm still experimenting but will let you know.

> For now could you have everything related to the cached guarded with an environment variable e.g. ENV["HOMEBREW_EXPERIMENTAL_CACHE"] or similar?

Done. 

> it may be worth experimenting with something like a SQLite database for this

I tried two different implementations with sqlite. The first one was a single table with columns identical to the hash keys of the marshaled data. The other had two tables, one table had the path, mtime, and id of a row in the second table. The second table was the byte code and extra data. My thought was that not retrieving the byte code and extra data each time would speed it up. That assumption proved to be incorrect so I'm back to using just one table.

Time comparisons between marshal and sqlite can be seen below. 

```
$ time HOMEBREW_EXPERIMENTAL_FORMULA_STORE=sqlite brew readall

real    0m3.684s
user    0m3.077s
sys     0m0.567s

$ time HOMEBREW_EXPERIMENTAL_FORMULA_STORE=marshal brew readall

real    0m3.177s
user    0m2.691s
sys     0m0.443s
```

> or the output of brew info --json if that's sufficient to be able to calculate this dependency data. Storing structured data on disk should be similar or superior for speed and would be vastly superior for usage in other contexts.

Agreed. I'm going to try that approach next. I think this will involve (re)building `Formula` instances from the structured data though. I haven't looked into it very much yet. I'd definitely welcome any quick pointers you might have. 

The last idea I had which might be a little radical is to bring along a portable version of redis and have a brew specific redis instance running as a fast persistent cache. It could easily be configured to listen on a unix socket so it doesn't use up a listening tcp port. If this did get implemented, I would suggest moving all brew formula cache data into it (brew desc cache comes to mind).

MichaelSquires:
Well, scratch that redis idea. I implemented a quick proof of concept and it turned out to be a lot slower than expected. I'll move on to investigating the structured data idea.

MikeMcQuaid:
> I think this will involve (re)building Formula instances from the structured data though.

I think there could be non-install use cases where querying the JSON alone would be sufficient. This would allow e.g. querying options, dependencies, reverse dependencies etc. which may be enough for now; at `install` time the slow bit is the installation itself anyway.

PR #3381, opened 2017-10-27
lock.sh: suppress error output when looking for flock

iMichka:

-----

Fixes (on centos6):
> which: no flock in (/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)

The checks for ruby and python at the same location have already their error message suppressed, so I think it makes sense to suppress that one too.

cc @sjackman


MikeMcQuaid:
Makes sense, thanks again @iMichka!

PR #3396, opened 2017-10-29
Implement `brew cask upgrade`

amyspark:

-----

This pull request fixes caskroom/homebrew-cask#29301 by implementing the `brew cask upgrade` command.
This is an initial version; I've added tests that cover a correct upgrade flow, but I would be especially interested if anyone could check that step 7 (if install fails, revert) works correctly.

Please review, and thanks for the comments!

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
@@ -8,11 +8,11 @@ module Hbc
       end
 
       def install_phase(**options)
-        move(**options)
+        move(source, target, **options)
       end
 
       def uninstall_phase(**options)
-        delete(**options)
+        move(target, source, **options)
       end
 
       def summarize_installed
@@ -25,12 +25,12 @@ module Hbc
 
       private
 
-      def move(force: false, command: nil, **options)
+      def move(source, target, force: false, command: nil, **options)
         if Utils.path_occupied?(target)
           message = "It seems there is already #{self.class.english_article} #{self.class.english_name} at '#{target}'"
           raise CaskError, "#{message}." unless force
           opoo "#{message}; overwriting."
-          delete(force: force, command: command, **options)
+          delete(target, force: force, command: command, **options)
         end
 
         unless source.exist?
@@ -49,7 +49,7 @@ module Hbc
         add_altname_metadata(target, source.basename, command: command)
       end
 
-      def delete(force: false, command: nil, **_)
+      def delete(target, force: false, command: nil, **_)
         ohai "Removing #{self.class.english_name} '#{target}'."
         raise CaskError, "Cannot remove undeletable #{self.class.english_name}." if MacOS.undeletable?(target)
 
--- a/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
@@ -25,7 +25,7 @@ module Hbc
 
       private
 
-      def move(source, target, force: false, command: nil, **options)
+      def move(source, target, skip: false, force: false, command: nil, **options)
         if Utils.path_occupied?(target)
           message = "It seems there is already #{self.class.english_article} #{self.class.english_name} at '#{target}'"
           raise CaskError, "#{message}." unless force
@@ -34,6 +34,8 @@ module Hbc
         end
 
         unless source.exist?
+          return if skip
+
           raise CaskError, "It seems the #{self.class.english_name} source '#{source}' is not there."
         end
 
--- a/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
@@ -8,11 +8,11 @@ module Hbc
       end
 
       def install_phase(**options)
-        move(source, target, **options)
+        move(**options)
       end
 
       def uninstall_phase(**options)
-        move(target, source, **options)
+        move_back(**options)
       end
 
       def summarize_installed
@@ -25,7 +25,7 @@ module Hbc
 
       private
 
-      def move(source, target, skip: false, force: false, command: nil, **options)
+      def move(force: false, command: nil, **options)
         if Utils.path_occupied?(target)
           message = "It seems there is already #{self.class.english_article} #{self.class.english_name} at '#{target}'"
           raise CaskError, "#{message}." unless force
@@ -34,8 +34,6 @@ module Hbc
         end
 
         unless source.exist?
-          return if skip
-
           raise CaskError, "It seems the #{self.class.english_name} source '#{source}' is not there."
         end
 
@@ -46,6 +44,31 @@ module Hbc
           FileUtils.move(source, target)
         else
           command.run("/bin/mv", args: [source, target], sudo: true)
+        end
+
+        add_altname_metadata(target, source.basename, command: command)
+      end
+
+      def move_back(skip: false, force: false, command: nil, **options)
+        if Utils.path_occupied?(source)
+          message = "It seems there is already #{self.class.english_article} #{self.class.english_name} at '#{source}'"
+          raise CaskError, "#{message}." unless force
+          opoo "#{message}; overwriting."
+          delete(source, force: force, command: command, **options)
+        end
+
+        unless target.exist?
+          return if skip
+          raise CaskError, "It seems the #{self.class.english_name} source '#{target}' is not there."
+        end
+
+        ohai "Moving #{self.class.english_name} '#{target.basename}' back to '#{source}'."
+        source.dirname.mkpath
+
+        if source.parent.writable?
+          FileUtils.move(target, source)
+        else
+          command.run("/bin/mv", args: [target, source], sudo: true)
         end
 
         add_altname_metadata(target, source.basename, command: command)
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -12,8 +12,11 @@ module Hbc
       end
 
       def run
-        outdated_casks = casks(alternative: -> { Hbc.installed.select { |cask|
-          cask.outdated?(greedy?) } }).select { |cask| cask.outdated?(true) }
+        outdated_casks = casks(alternative: lambda {
+          Hbc.installed.select do |cask|
+            cask.outdated?(greedy?)
+          end
+        }).select { |cask| cask.outdated?(true) }
 
         if outdated_casks.empty?
           oh1 "No Casks to upgrade"
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -13,14 +13,12 @@ module Hbc
       end
 
       def run
-        outdated_casks = casks(alternative: -> { Hbc.installed }).find_all { |cask| cask.outdated?(greedy?) }
+        outdated_casks = casks(alternative: -> { Hbc.installed }).select { |cask| cask.outdated?(greedy?) }
 
-        if outdated_casks.empty?
-          oh1 "No packages to upgrade"
-        else
-          oh1 "Upgrading #{Formatter.pluralize(outdated_casks.length, "outdated package")}, with result:"
-          puts outdated_casks.map { |f| "#{f.full_name} #{f.version}" } * ", "
-        end
+        return if outdated_casks.empty?
+
+        oh1 "Upgrading #{Formatter.pluralize(outdated_casks.length, "outdated package")}, with result:"
+        puts outdated_casks.map { |f| "#{f.full_name} #{f.version}" } * ", "
 
         outdated_casks.each do |old_cask|
           odebug "Uninstalling Cask #{old_cask}"
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -34,25 +34,29 @@ module Hbc
 
           old_cask_installer = Installer.new(old_cask, binaries: binaries?, verbose: verbose?, force: force?, upgrade: true)
 
-          old_cask_installer.uninstall
+          new_cask = CaskLoader.load(old_cask.to_s)
 
-          begin
-            odebug "Installing new version of Cask #{old_cask}"
-
-            new_cask = CaskLoader.load(old_cask.to_s)
-
+          new_cask_installer =
             Installer.new(new_cask, binaries:       binaries?,
                                     verbose:        verbose?,
                                     force:          force?,
                                     skip_cask_deps: skip_cask_deps?,
                                     require_sha:    require_sha?,
-                                    upgrade: true).install
+                                    upgrade: true)
 
+          begin
+            # purge artifacts BUT keep metadata aside
+            old_cask_installer.start_upgrade
+
+            # install BUT do not yet save metadata
+
+            new_cask_installer.install
+
+            # if successful, remove old metadata and install new
             old_cask_installer.finalize_upgrade
-          rescue CaskUnavailableError => e
+          rescue CaskError => e
             opoo e.message
-          rescue CaskAlreadyInstalledError => e
-            opoo e.message
+            old_cask_installer.revert_upgrade
           end
         end
       end
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -4,7 +4,6 @@ module Hbc
       option "--greedy", :greedy, false
       option "--quiet",  :quiet, false
       option "--force", :force, false
-      option "--force-update", :force_update, false
       option "--skip-cask-deps", :skip_cask_deps, false
 
       def initialize(*)
@@ -24,10 +23,9 @@ module Hbc
           odebug "Started upgrade process for Cask #{old_cask}"
           raise CaskNotInstalledError, old_cask unless old_cask.installed? || force?
 
-          unless old_cask.installed_caskfile.nil?
-            # use the same cask file that was used for installation, if possible
-            old_cask = CaskLoader.load(old_cask.installed_caskfile) if old_cask.installed_caskfile.exist?
-          end
+          raise CaskUnavailableError.new(old_cask, "The Caskfile is missing!") if old_cask.installed_caskfile.nil?
+
+          old_cask = CaskLoader.load(old_cask.installed_caskfile)
 
           old_cask_installer = Installer.new(old_cask, binaries: binaries?, verbose: verbose?, force: force?, upgrade: true)
 
@@ -66,10 +64,10 @@ module Hbc
             # If successful, wipe the old Cask from staging
             old_cask_installer.finalize_upgrade
           rescue CaskError => e
-            opoo e.message
             new_cask_installer.uninstall_artifacts if new_artifacts_installed
             new_cask_installer.purge_versioned_files
             old_cask_installer.revert_upgrade if started_upgrade
+            raise e
           end
         end
       end
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -1,0 +1,65 @@
+module Hbc
+  class CLI
+    class Upgrade < AbstractCommand
+      option "--greedy", :greedy, false
+      option "--quiet",  :quiet, false
+      option "--force", :force, false
+      option "--force-update", :force_update, false
+      option "--skip-cask-deps", :skip_cask_deps, false
+
+      def initialize(*)
+        super
+        self.verbose = ($stdout.tty? || verbose?) && !quiet?
+      end
+
+      def run
+        outdated_casks = casks(alternative: -> { Hbc.installed }).find_all { |cask| cask.outdated?(greedy?) }
+
+        if outdated_casks.empty?
+          oh1 "No packages to upgrade"
+        else
+          oh1 "Upgrading #{Formatter.pluralize(outdated_casks.length, "outdated package")}, with result:"
+          puts outdated_casks.map { |f| "#{f.full_name} #{f.version}" } * ", "
+        end
+
+        outdated_casks.each do |old_cask|
+          odebug "Uninstalling Cask #{old_cask}"
+
+          raise CaskNotInstalledError, old_cask unless old_cask.installed? || force?
+
+          unless old_cask.installed_caskfile.nil?
+            # use the same cask file that was used for installation, if possible
+            old_cask = CaskLoader.load(old_cask.installed_caskfile) if old_cask.installed_caskfile.exist?
+          end
+
+          old_cask_installer = Installer.new(old_cask, binaries: binaries?, verbose: verbose?, force: force?, upgrade: true)
+
+          old_cask_installer.uninstall
+
+          begin
+            odebug "Installing new version of Cask #{old_cask}"
+
+            new_cask = CaskLoader.load(old_cask.to_s)
+
+            Installer.new(new_cask, binaries:       binaries?,
+                                    verbose:        verbose?,
+                                    force:          force?,
+                                    skip_cask_deps: skip_cask_deps?,
+                                    require_sha:    require_sha?,
+                                    upgrade: true).install
+
+            old_cask_installer.finalize_upgrade
+          rescue CaskUnavailableError => e
+            opoo e.message
+          rescue CaskAlreadyInstalledError => e
+            opoo e.message
+          end
+        end
+      end
+
+      def self.help
+        "upgrades all outdated casks"
+      end
+    end
+  end
+end
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -12,7 +12,7 @@ module Hbc
       end
 
       def run
-        outdated_casks = casks(alternative: -> { Hbc.installed }).select { |cask| cask.outdated?(greedy?) }
+        outdated_casks = casks(alternative: -> { Hbc.installed }).select { |cask| cask.outdated?(greedy?) || (args.include?(cask.token) && cask.outdated?(true)) }
 
         if outdated_casks.empty?
           oh1 "No Casks to upgrade"
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -14,7 +14,10 @@ module Hbc
       def run
         outdated_casks = casks(alternative: -> { Hbc.installed }).select { |cask| cask.outdated?(greedy?) }
 
-        return if outdated_casks.empty?
+        if outdated_casks.empty?
+          oh1 "No Casks to upgrade"
+          return
+        end
 
         oh1 "Upgrading #{Formatter.pluralize(outdated_casks.length, "outdated package")}, with result:"
         puts outdated_casks.map { |f| "#{f.full_name} #{f.version}" } * ", "
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -21,8 +21,7 @@ module Hbc
         puts outdated_casks.map { |f| "#{f.full_name} #{f.version}" } * ", "
 
         outdated_casks.each do |old_cask|
-          odebug "Uninstalling Cask #{old_cask}"
-
+          odebug "Started upgrade process for Cask #{old_cask}"
           raise CaskNotInstalledError, old_cask unless old_cask.installed? || force?
 
           unless old_cask.installed_caskfile.nil?
@@ -43,17 +42,26 @@ module Hbc
                                     upgrade: true)
 
           begin
-            # purge artifacts BUT keep metadata aside
+            # Start new Cask's installation steps
+            new_cask_installer.check_conflicts
+
+            new_cask_installer.fetch
+
+            new_cask_installer.stage
+
+            # Move the old Cask's artifacts back to staging
             old_cask_installer.start_upgrade
 
-            # install BUT do not yet save metadata
+            # Install the new Cask
+            new_cask_installer.install_artifacts
 
-            new_cask_installer.install
+            new_cask_installer.enable_accessibility_access
 
-            # if successful, remove old metadata and install new
+            # If successful, wipe the old Cask from staging
             old_cask_installer.finalize_upgrade
           rescue CaskError => e
             opoo e.message
+            new_cask_installer.uninstall
             old_cask_installer.revert_upgrade
           end
         end
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -12,7 +12,8 @@ module Hbc
       end
 
       def run
-        outdated_casks = casks(alternative: -> { Hbc.installed }).select { |cask| cask.outdated?(greedy?) || (args.include?(cask.token) && cask.outdated?(true)) }
+        outdated_casks = casks(alternative: -> { Hbc.installed.select { |cask|
+          cask.outdated?(greedy?) } }).select { |cask| cask.outdated?(true) }
 
         if outdated_casks.empty?
           oh1 "No Casks to upgrade"
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -41,6 +41,9 @@ module Hbc
                                     require_sha:    require_sha?,
                                     upgrade: true)
 
+          started_upgrade = false
+          new_artifacts_installed = false
+
           begin
             # Start new Cask's installation steps
             new_cask_installer.check_conflicts
@@ -51,9 +54,12 @@ module Hbc
 
             # Move the old Cask's artifacts back to staging
             old_cask_installer.start_upgrade
+            # And flag it so in case of error
+            started_upgrade = true
 
             # Install the new Cask
             new_cask_installer.install_artifacts
+            new_artifacts_installed = true
 
             new_cask_installer.enable_accessibility_access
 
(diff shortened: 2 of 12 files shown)


reitermarkus:
> step 7 (if install fails, revert) 

What/where exactly is step 7?

Anyways, here is what I had in mind to handle this case:

Currently we have a `#install` and `#uninstall` method for every artifact class. In `#uninstall` the artifact is usually completely removed. To handle a failed upgrade, we should instead make `#uninstall` do the exact opposite of `#install`, i.e. if `#install` moves a file from A to B, `#uninstall` should move B back to A.

amyspark:
[Here](https://github.com/caskroom/homebrew-cask/issues/29301#issue-202421978):
> Behaviour for **`brew cask upgrade`**:
>
> Same steps 1–5.
>
> 6. For the ones that differ, run the uninstall and install steps. Uninstalling is necessary to avoid leaving old traces of it when uninstallation procedures change.
> 7. If install fails (after uninstall), revert to previous version.

`uninstall` purges the old installation's metadata, which I need to revert the upgrade.
I will try your approach and rewrite the `uninstall` step.

amyspark:
So far, I've added an extra `start_upgrade` step that uninstalls the app but preserves the outdated Cask's metadata. But I'd need a way to back it up so as to prevent Homebrew from detecting it when the new Cask is installed.
Any ideas?

amyspark:
@reitermarkus, I'm afraid I don't yet understand your intended `upgrade` flow. Could you be more specific?

So far, I came up with this flow:
1. Once the upgrade package is downloaded, `uninstall_artifacts` the old Cask (i.e. execute `uninstall` _without_ `purge_versioned_files` -- see later why)
2. `install` the new Cask.
3. If successful, `finalize_upgrade` the old Cask (i.e. execute the remaining `uninstall` steps). If not successful, `reinstall` the old Cask.

Since I reuse the  `reinstall` step in `revert_upgrade`, Cask can fetch again the old package if necessary and restore the old installation. Not deleting the old Cask's metadata is key here, because it must be copied again in `save_caskfile`.

reitermarkus:
> If not successful, `reinstall` the old Cask.

This is what I want to avoid, because often the old version is not available anymore.

We already had it installed before, so we can reinstall from the previous local version. So you are right that we don't want to call `purge_versioned_files` before the upgrade.

So how do we make it possible to reinstall from the previous local version? Make `uninstall_artifacts` not delete anything. A simple (hopefully more specific) example:

- A Cask installs `App.app` by moving it from the `staged_path`/`App.app` to `appdir`/`App.app`.
- When uninstalling this Cask, it will straight away delete `appdir`/`App.app`.

We don't want this. Instead, we want the `uninstall_phase` to move `appdir`/`App.app` back to `staged_path`. And if we want to actually uninstall the whole Cask, `purge_versioned_files` then removes `staged_path`.

A more general overview of the flow:

- **install** + **uninstall**
  `download` -> `stage` -> `install` -> `uninstall` -> `delete`

So `stage` and `uninstall` will result in the same state, because `uninstall` is the opposite of `install`.

- **upgrade**
  `download` -> `stage` new version & `uninstall` old version -> `install` new version
   -> success: `delete` old version
   -> error: `revert` partially installed new version -> `install` old version

amyspark:
I see now! Thanks for the explanation-- I think it's ready now.

amyspark:
@reitermarkus, with the last commit I've fixed the `upgrade` tests (I need the Casks' artifacts to be properly installed in order to test the `move` steps.)
The broken `install`, `uninstall`, and `reinstall` tests are still left. These are implemented against the `stdout` text; do you think I should update the tests or change the messages?
Also, please check that with the current steps a `==> Moving App 'xxx' to 'xxx'.`  is printed without a "Purging files" or something, which seems incomplete to me. Should I do something about this?

amyspark:
I've fixed all the remaining issues:
- Updated the `reinstall` and `uninstall` tests' text.
- I added a parameter to the `uninstall_artifacts` step, `clear`, so as to let it skip artifacts that have been manually uninstalled (eg. when testing the `with-uninstall-script` Cask). The `move` function was updated accordingly.

commitay:
Apologies in advance if I am misunderstanding this.

Currently this  uninstalls `pkg` Casks before downloading the upgrade, is this correct? It doesn't seem to follow the upgrade flow in https://github.com/Homebrew/brew/pull/3396#issuecomment-342315684

```
$ brew cask upgrade
vfuse 1.0.6
==> Starting upgrade for Cask vfuse
==> Running uninstall process for vfuse; your password may be necessary
==> Removing launchctl service com.chilcote.vfused
Password:
==> Uninstalling packages:
com.github.vfuse
==> Unlinking Binary '/usr/local/bin/vfuse'.
==> Caveats
Cask vfuse installs files under "/usr/local". The presence of such
files can cause warnings when running "brew doctor", which is considered
to be a bug in Homebrew-Cask.

==> Satisfying dependencies
==> Downloading https://github.com/chilcote/vfuse/releases/download/1.0.6/vfuse-
==> Verifying checksum for Cask vfuse
==> Installing Cask vfuse
==> Running installer for vfuse; your password may be necessary.
==> Package installers may write to any location; options such as --appdir are i
==> installer: Package name is vfuse-1.0.6
==> installer: Installing at base path /
==> installer:PHASE:Preparing for installation…
==> installer:PHASE:Preparing the disk…
==> installer:PHASE:Preparing vfuse-1.0.6…
==> installer:PHASE:Waiting for other installations to complete…
==> installer:PHASE:Configuring the installation…
==> installer:STATUS:
==> installer:%2.229250
==> installer:PHASE:Validating packages…
==> installer:%97.750000
==> installer:STATUS:
==> installer:PHASE:Finishing the Installation…
==> installer:STATUS:
==> installer:%100.000000
==> installer:PHASE:The software was successfully installed.
==> installer: The install was successful.
==> Linking Binary 'vfuse' to '/usr/local/bin/vfuse'.
🍺  vfuse was successfully upgraded!
==> Purging files for version 1.0.5 of Cask vfuse
```

amyspark:
@commitay, yes, you're right. I did it that way because I wanted to just `install` the new Cask (if it fails, it reverts everything on its own).

amyspark:
@commitay, now the flow looks like this:
```bash
amalia@Sakura:~$ brew cask upgrade
==> Upgrading 2 outdated packages, with result:
libreoffice 5.4.3, libreoffice-language-pack 5.4.3
==> Satisfying dependencies
==> Downloading https://download.documentfoundation.org/libreoffice/stable/5.4.3
######################################################################## 100,0%
==> Verifying checksum for Cask libreoffice
==> Starting upgrade for Cask libreoffice
==> Moving App 'LibreOffice.app' to '/usr/local/Caskroom/libreoffice/5.4.2/Libre
==> Unlinking Binary '/usr/local/bin/regmerge'.
==> Unlinking Binary '/usr/local/bin/regview'.
==> Unlinking Binary '/usr/local/bin/senddoc'.
==> Unlinking Binary '/usr/local/bin/ui-previewer'.
==> Unlinking Binary '/usr/local/bin/uno'.
==> Unlinking Binary '/usr/local/bin/urelibs'.
==> Unlinking Binary '/usr/local/bin/uri-encode'.
==> Unlinking Binary '/usr/local/bin/xpdfimport'.
==> Unlinking Binary '/usr/local/bin/soffice'.
==> Unlinking Binary '/usr/local/bin/unopkg'.
==> Unlinking Binary '/usr/local/bin/gengal'.
==> Moving App 'LibreOffice.app' to '/Applications/LibreOffice.app'.
==> Linking Binary 'regmerge' to '/usr/local/bin/regmerge'.
==> Linking Binary 'regview' to '/usr/local/bin/regview'.
==> Linking Binary 'senddoc' to '/usr/local/bin/senddoc'.
==> Linking Binary 'ui-previewer' to '/usr/local/bin/ui-previewer'.
==> Linking Binary 'uno' to '/usr/local/bin/uno'.
==> Linking Binary 'urelibs' to '/usr/local/bin/urelibs'.
==> Linking Binary 'uri-encode' to '/usr/local/bin/uri-encode'.
==> Linking Binary 'xpdfimport' to '/usr/local/bin/xpdfimport'.
==> Linking Binary 'soffice.wrapper.sh' to '/usr/local/bin/soffice'.
==> Linking Binary 'unopkg' to '/usr/local/bin/unopkg'.
==> Linking Binary 'gengal' to '/usr/local/bin/gengal'.
==> Purging files for version 5.4.2 of Cask libreoffice
🍺  libreoffice was successfully upgraded!
==> Satisfying dependencies
All Cask dependencies satisfied.
==> Downloading http://download.documentfoundation.org/libreoffice/stable/5.4.3/
######################################################################## 100,0%
==> Verifying checksum for Cask libreoffice-language-pack
==> Starting upgrade for Cask libreoffice-language-pack
==> Purging files for version 5.4.2 of Cask libreoffice-language-pack
🍺  libreoffice-language-pack was successfully upgraded!
```

amyspark:
Is there anything I can do wrt the `java` and `virtualbox` Casks, @commitay?

commitay:
I've open PRs with fixes for their dependent Casks, thanks anyway for the offer!

amyspark:
I added two more tests to check that failed upgrades are properly reverted (if necessary).

Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
```diff
@@ -0,0 +1,82 @@
...
+        return if outdated_casks.empty?
+
+        oh1 "Upgrading #{Formatter.pluralize(outdated_casks.length, "outdated package")}, with result:"
+        puts outdated_casks.map { |f| "#{f.full_name} #{f.version}" } * ", "
+
+        outdated_casks.each do |old_cask|
+          odebug "Started upgrade process for Cask #{old_cask}"
+          raise CaskNotInstalledError, old_cask unless old_cask.installed? || force?
+
+          unless old_cask.installed_caskfile.nil?
```
reitermarkus: We should error here if there is no old cask file.

Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
```diff
@@ -0,0 +1,82 @@
+module Hbc
+  class CLI
+    class Upgrade < AbstractCommand
+      option "--greedy", :greedy, false
+      option "--quiet",  :quiet, false
+      option "--force", :force, false
+      option "--force-update", :force_update, false
```
reitermarkus: This flag isn't used anywhere.

Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
```diff
@@ -0,0 +1,82 @@
...
+            # Install the new Cask
+            new_cask_installer.install_artifacts
+            new_artifacts_installed = true
+
+            new_cask_installer.enable_accessibility_access
+
+            # If successful, wipe the old Cask from staging
+            old_cask_installer.finalize_upgrade
+          rescue CaskError => e
+            opoo e.message
```
reitermarkus: Instead outputting the error message here, re-raise the exception at the end. Otherwise it will exit with `0` when it fails.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -83,7 +84,7 @@ def install
       odebug "Hbc::Installer#install"
 
       if @cask.installed? && !force? && !@reinstall
-        raise CaskAlreadyInstalledError, @cask
+        raise CaskAlreadyInstalledError, @cask unless upgrade?
```
reitermarkus: Add `!upgrade?` to the `if` above instead, or move `force? || @reinstall` down here.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -367,12 +368,34 @@ def save_caskfile
...
       disable_accessibility_access
-      uninstall_artifacts
+      uninstall_artifacts(clear: true)
       purge_versioned_files
       purge_caskroom_path if force?
     end
 
-    def uninstall_artifacts
+    def start_upgrade
+      return unless upgrade?
```
reitermarkus: Remove this, this shouldn't be called anyway if this isn't an upgrade. Same goes for the methods below.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,10 +443,10 @@ def purge_versioned_files
         end
       end
       @cask.metadata_versioned_path.rmdir_if_possible
-      @cask.metadata_master_container_path.rmdir_if_possible
+      @cask.metadata_master_container_path.rmdir_if_possible unless upgrade?
```
reitermarkus: Is this check here necessary, since this is not `rm_rf` but only `rmdir`?

amyspark: Yes. Line 445 clears the Cask's versioned metadata (which is what we need for `finalize_upgrade`), but 446 below deletes the whole metadata folder (that is, `$(brew --prefix)/Caskroom/<cask_name>/`, which we only need for `uninstall`).

Library/Homebrew/cask/lib/hbc/artifact/moved.rb
```diff
@@ -25,15 +25,17 @@ def summarize_installed
...
         if Utils.path_occupied?(target)
           message = "It seems there is already #{self.class.english_article} #{self.class.english_name} at '#{target}'"
           raise CaskError, "#{message}." unless force
           opoo "#{message}; overwriting."
-          delete(force: force, command: command, **options)
+          delete(target, force: force, command: command, **options)
         end
 
         unless source.exist?
+          return if skip
```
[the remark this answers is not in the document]
amyspark: You're right here. However, we will still need this var because `uninstall` tests expect `delete`'s behaviour (skip if artifact not found).

amyspark:
@vitorgalvao @reitermarkus, can you check if everything was sorted out? Thanks! 👍

reitermarkus:
Ok, I tested this now, everything seems to be working fine. One small thing: There should be some kind of message if there is no upgrade available.

amyspark:
@reitermarkus done! Thanks for the checkup.

reitermarkus:
One other thing. I shouldn't have to specify `--greedy` if I specifically ask for a Cask to be upgraded by name.

I.e. if I run `brew cask outdated --greedy` which returns an outdated Cask with `auto_updates`, I would expect `brew cask upgrade {{cask}}` to still upgrade it.

amyspark:
@reitermarkus Done again, thanks for the comment.

Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
```diff
@@ -12,7 +12,7 @@ def initialize(*)
       end
 
       def run
-        outdated_casks = casks(alternative: -> { Hbc.installed }).select { |cask| cask.outdated?(greedy?) }
+        outdated_casks = casks(alternative: -> { Hbc.installed }).select { |cask| cask.outdated?(greedy?) || (args.include?(cask.token) && cask.outdated?(true)) }
```
reitermarkus: I think this should work here:

```ruby
casks(alternative: -> { Hbc.installed.select { |cask| cask.outdated?(greedy?)  } })
  .select { |cask| cask.outdated?(true) }
```

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -76,7 +76,7 @@
       expect(Hbc::CaskLoader.load("local-transmission").versions).to include("2.60")
     end
 
-    it 'and ignores "auto_updates" and "latest" Casks even when their tokens are provided in the command line' do
+    it 'but updates "auto_updates" and "latest" Casks when their tokens are provided in the command line' do
```
reitermarkus: Remove the “but” here.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+  end
+
+  describe "handles borked upgrades" do
+    include_context "Casks that will fail upon upgrade"
+
+    output_reverted = Regexp.new <<~EOS
+      Warning: Reverting upgrade for Cask .*
+    EOS
+
+    it "by restoring the old Cask if the upgrade's install failed" do
```
reitermarkus: “restores” instead of “by restoring” (Basically all of these should read like full sentences.)

reitermarkus: Maybe also just “upgrade“ instead of “upgrade's install”.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+
+      installed.each { |cask| Hbc::CLI::Install.run(cask) }
+
+      allow_any_instance_of(described_class).to receive(:verbose?).and_return(true)
+    end
+  end
+
+  shared_context "Casks that will fail upon upgrade" do
+    before(:example) do
+      installed =
```
reitermarkus: Use the `let` syntax to set these. 

```ruby
let(:installed) { 
  [
    "outdated/bad-checksum",
    "outdated/will-fail-if-upgraded",
  ]
}
```

reitermarkus:
In the tests, don't reload the Casks everytime, save them in a variable. Of course they have to be reloaded once after the upgrade, but don't use `Hbc::CaskLoader.load` in consecutive checks.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+    it "by restoring the old Cask if the upgrade's install failed" do
+      expect(Hbc::CaskLoader.load("will-fail-if-upgraded")).to be_installed
+      expect(Hbc.appdir.join("container")).to be_a_file
+      expect(Hbc::CaskLoader.load("will-fail-if-upgraded").versions).to include("1.2.2")
+
+      expect {
+        expect {
+          described_class.run("will-fail-if-upgraded")
+        }.to raise_error(Hbc::CaskError)
+      }.to output(output_reverted).to_stderr
```
reitermarkus: Does `.and` work here instead of nesting these?

amyspark: Yes, but I'll need to set up a negated matcher for another test.

amyspark:
I've refactored the tests to load the Casks only once. Thanks for the comment.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      described_class.run("auto-updates", "--greedy")
+
+      expect(auto_updates).to be_installed
+      expect(auto_updates_path).to be_a_directory
+      expect(auto_updates.versions).to include("2.61")
+    end
+  end
+
+  describe "handles borked upgrades" do
+    include_context "Casks that will fail upon upgrade"
```
reitermarkus: This context is only used once, so you can just include it directly here.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      expect(local_caffeine.versions).to include("1.2.3")
+
+      expect(auto_updates).to be_installed
+      expect(auto_updates_path).to be_a_directory
+      expect(auto_updates.versions).to include("2.61")
+    end
+  end
+
+  describe "with --greedy it checks additional Casks" do
+    include_context "Proper Casks"
```
reitermarkus: Instead of including this twice, create a super-context, i.e. `context "successful upgrade" do`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      expect(local_caffeine).to be_installed
+      expect(local_caffeine_route).to be_a_directory
+      expect(local_caffeine.versions).to include("1.2.3")
+
+      expect(local_transmission).to be_installed
+      expect(local_transmission_route).to be_a_directory
+      expect(local_transmission.versions).to include("2.61")
+    end
+
+    it "and updates only the Casks specified in the command line" do
```
reitermarkus: Remove the “and” here.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      installed.each { |cask| Hbc::CLI::Install.run(cask) }
+
+      allow_any_instance_of(described_class).to receive(:verbose?).and_return(true)
+    end
+  end
+
+  describe 'without --greedy it ignores the Casks with "version latest" or "auto_updates true"' do
+    include_context "Proper Casks"
+
+    it "and updates all the installed Casks when no token is provided" do
```
reitermarkus: Remove the “and” here.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+
+      described_class.run("auto-updates", "--greedy")
+
+      expect(auto_updates).to be_installed
+      expect(auto_updates_path).to be_a_directory
+      expect(auto_updates.versions).to include("2.61")
+    end
+  end
+
+  describe "handles borked upgrades" do
```
reitermarkus: Make this a `context`, i.e. `context "failed upgrade" do`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      expect(auto_updates_path).to be_a_directory
+      expect(auto_updates.versions).to include("2.61")
+
+      expect(local_transmission).to be_installed
+      expect(local_transmission_route).to be_a_directory
+      expect(local_transmission.versions).to include("2.61")
+    end
+
+    it 'does not include the Casks with "auto_updates true" when the version did not change' do
+      auto_updates = Hbc::CaskLoader.load("auto-updates")
```
reitermarkus: Call this `cask` instead of `auto_updates `.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+  end
+
+  describe 'without --greedy it ignores the Casks with "version latest" or "auto_updates true"' do
+    include_context "Proper Casks"
+
+    it "and updates all the installed Casks when no token is provided" do
+      local_caffeine = Hbc::CaskLoader.load("local-caffeine")
+      local_caffeine_route = Hbc.appdir.join("Caffeine.app")
+      local_transmission = Hbc::CaskLoader.load("local-transmission")
+      local_transmission_route = Hbc.appdir.join("Transmission.app")
```
reitermarkus: This should also be `_path`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      expect(local_transmission).to be_installed
+      expect(local_transmission_route).to be_a_directory
+      expect(local_transmission.versions).to include("2.61")
+    end
+
+    it "and updates only the Casks specified in the command line" do
+      local_caffeine = Hbc::CaskLoader.load("local-caffeine")
+      local_caffeine_route = Hbc.appdir.join("Caffeine.app")
+      local_transmission = Hbc::CaskLoader.load("local-transmission")
+      local_transmission_route = Hbc.appdir.join("Transmission.app")
```
reitermarkus: This should also be `_path`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+  describe "with --greedy it checks additional Casks" do
+    include_context "Proper Casks"
+
+    it 'includes the Casks with "auto_updates true" or "version latest" with --greedy' do
+      local_caffeine = Hbc::CaskLoader.load("local-caffeine")
+      local_caffeine_route = Hbc.appdir.join("Caffeine.app")
+      auto_updates = Hbc::CaskLoader.load("auto-updates")
+      auto_updates_path = Hbc.appdir.join("MyFancyApp.app")
+      local_transmission = Hbc::CaskLoader.load("local-transmission")
+      local_transmission_route = Hbc.appdir.join("Transmission.app")
```
reitermarkus: This should also be _path, right?

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      expect(local_caffeine.versions).to include("1.2.3")
+
+      expect(local_transmission).to be_installed
+      expect(local_transmission_route).to be_a_directory
+      expect(local_transmission.versions).to include("2.60")
+    end
+
+    it 'updates "auto_updates" and "latest" Casks when their tokens are provided in the command line' do
+      local_caffeine = Hbc::CaskLoader.load("local-caffeine")
+      local_caffeine_route = Hbc.appdir.join("Caffeine.app")
```
reitermarkus: This should also be `_path`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      allow_any_instance_of(described_class).to receive(:verbose?).and_return(true)
+    end
+  end
+
+  describe 'without --greedy it ignores the Casks with "version latest" or "auto_updates true"' do
+    include_context "Proper Casks"
+
+    it "and updates all the installed Casks when no token is provided" do
+      local_caffeine = Hbc::CaskLoader.load("local-caffeine")
+      local_caffeine_route = Hbc.appdir.join("Caffeine.app")
```
reitermarkus: This should also be `_path`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      expect(local_caffeine.versions).to include("1.2.3")
+
+      expect(local_transmission).to be_installed
+      expect(local_transmission_route).to be_a_directory
+      expect(local_transmission.versions).to include("2.61")
+    end
+
+    it "and updates only the Casks specified in the command line" do
+      local_caffeine = Hbc::CaskLoader.load("local-caffeine")
+      local_caffeine_route = Hbc.appdir.join("Caffeine.app")
```
reitermarkus: This should also be `_path`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+      expect(auto_updates.versions).to include("2.61")
+    end
+  end
+
+  describe "with --greedy it checks additional Casks" do
+    include_context "Proper Casks"
+
+    it 'includes the Casks with "auto_updates true" or "version latest" with --greedy' do
+      local_caffeine = Hbc::CaskLoader.load("local-caffeine")
+      local_caffeine_route = Hbc.appdir.join("Caffeine.app")
```
reitermarkus: This should also be `_path`.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+        described_class.run("will-fail-if-upgraded")
+      }.to raise_error(Hbc::CaskError).and output(output_reverted).to_stderr
+
+      expect(will_fail_if_upgraded).to be_installed
+      expect(will_fail_if_upgraded_path).to be_a_file
+      expect(will_fail_if_upgraded.versions).to include("1.2.2")
+      expect(will_fail_if_upgraded.staged_path).to_not exist
+    end
+
+    it "by not restoring the old Cask if the upgrade failed pre-install" do
```
reitermarkus: This doesn't read like a sentence.

amyspark:
Thanks again for the comments, @reitermarkus. Fixes incoming.

Library/Homebrew/test/cask/cli/upgrade_spec.rb
```diff
@@ -0,0 +1,211 @@
...
+        expect(local_caffeine.versions).to include("1.2.3")
+
+        expect(auto_updates).to be_installed
+        expect(auto_updates_path).to be_a_directory
+        expect(auto_updates.versions).to include("2.61")
+      end
+    end
+
+    describe "with --greedy it checks additional Casks" do
+      it 'includes the Casks with "auto_updates true" or "version latest" with --greedy' do
```
reitermarkus: The “with --greedy” here is redundant.

amyspark:
And there we go 😉

reitermarkus:
Thank you for your great work here, @amyspark! 👍 

🎉

monouser7dig:
great to have this natively supported now however what about upgrading auto-updating casks? --greedy calls all "latest" casks which is pretty verbose.

I like a package manager because it provides version information and can act based on that but now the version information of the auto-updatable casks is rendered useless.

(I currently use `brew cu -a` all the time)

thanks for your efforts!

toonetown:
> @monouser7dig
> great to have this natively supported now however what about upgrading auto-updating casks? --greedy calls all "latest" casks which is pretty verbose.

I agree - I'd like to have the ability to upgrade greedily...but without "latest".

vitorgalvao:
> I like a package manager because it provides version information

We can’t have it every way. Apps will auto-update independently of what we do, and we’re not going to cripple that (nor could we even if we wanted). *So versions are going to be auto of sync anyway for casks that auto-update, no way around that*.

> but now the version information of the auto-updatable casks is rendered useless.

Maybe from an `info` perspective, but not from a cask file perspective.

> I agree - I'd like to have the ability to upgrade greedily...but without "latest".

And what about the casks that are `:latest` but do not auto-update? You don’t want those to remain outdated.

koutsenko:
Hello, how can i test that feature (brew cask upgrade)?
```
bash-3.2$ brew cask --version
Homebrew-Cask 1.3.8
caskroom/homebrew-cask (git revision ce2cb; last commit 2017-11-28)
bash-3.2$ brew cask upgrade
brew-cask provides a friendly homebrew-style CLI workflow for the
administration of macOS applications distributed as binaries.

Commands:
...
(and just help printed instead of upgrading)  
...
```

toonetown:
@vitorgalvao 

Yes - those are all points that have been used in the past to argue *against* having an `brew cask upgrade` in the first place.  I'm fine with whatever is decided - you can do some command line magic to achieve what I'm looking for anyway:

```/bin/bash
brew cask upgrade $(brew cask outdated --greedy --verbose | grep -v '(latest)' | cut -d' ' -f1)
```

monouser7dig:
1. I usually run upgrade before the apps auto update themselves so no problem on that side (Or on the rare occasion just reinstall using brew cu -a, on the other hand always upgrading „latest“ is a much! bigger overhead)
2. I manyually uodate the apps that have „latest“ using brew reinstall or using their build in upgrade mechanism (chrome canary) and this number of apps decreases 

Just use brew upgrade for everything that has Version Info so as soon as I run my brew command everything is ready to go. (Or at least as ready as it can be)

MikeMcQuaid:
> And what about the casks that are :latest but do not auto-update? You don’t want those to remain outdated.

@vitorgalvao Out of interest, is there any metadata that indicates if autoupdate is something a Cask does. If so, perhaps that could be used here?

vitorgalvao:
@MikeMcQuaid Yes. [It’s a simple stanza we add to casks](https://github.com/caskroom/homebrew-cask/blob/ce2cbe3b4d917cd360c51c5544909076d81b9448/Casks/alfred.rb#L9).

I guess that could work: if `:latest` and `auto_updates true`, never include in `upgrade`.

I’m now running a script to determine how many casks we have in that situation. I’ll wager they’re not many. For reference, we have (across all official repos):

+ 5272 total casks.
+ 1655 casks with `latest`.

reitermarkus:
I would actually be in favour of also updating Casks with `auto_updates`, even without `--greedy`. Reason being that an app only actually auto-updates if I open it.

MikeMcQuaid:
@reitermarkus I think it's a reasonable assumption that people are opening the Casks they've installed, though (at least for a default).

monouser7dig:
@Mike But you do not open them on a regular basis, yet if you open them you still want them up to date, the main reason to use a package manager

vitorgalvao:
> I’m now running a script to determine how many casks we have in that situation.

The verdict is in. From 5272 casks, only 48 are `version :latest` and `auto_updates true`:

<details>
<summary>latest auto_updates casks</summary>
<ul>
<li>avira-antivirus</li>
<li>basecamp</li>
<li>bluesense</li>
<li>bookmacster</li>
<li>bordertool</li>
<li>bordertool2</li>
<li>busycal</li>
<li>chatwork</li>
<li>coccoc</li>
<li>domino-cli</li>
<li>doxie</li>
<li>dropbox</li>
<li>expressscribe</li>
<li>feed-the-beast</li>
<li>google-chrome-canary</li>
<li>gpg-suite-nightly</li>
<li>huamim</li>
<li>ipass</li>
<li>iterm2-nightly</li>
<li>jandi</li>
<li>jdownloader</li>
<li>joinme</li>
<li>mailspring</li>
<li>megasync</li>
<li>miniconda</li>
<li>miniconda2</li>
<li>moxtra</li>
<li>myworkspace</li>
<li>orfo</li>
<li>orfo-plus</li>
<li>plantronics-hub</li>
<li>pokerstars</li>
<li>remote-play</li>
<li>riverdesign-sparkle</li>
<li>routeconverter</li>
<li>sketchpacks</li>
<li>soda-player</li>
<li>spatial</li>
<li>spotify</li>
<li>starleaf-breeze</li>
<li>steam</li>
<li>steamcmd</li>
<li>stride</li>
<li>vk-messenger</li>
<li>xnviewmp</li>
<li>xoctave</li>
<li>zeplin</li>
<li>zoomus-outlook-plugin</li>
</ul>
</details>

I don’t think it’s worth losing sleep over this, then. Especially since ideally the number of `latest` casks will continue to decrease, and if they have `auto_updates` there’s a better chance we can get find `appcast`s for them (meaning they’ll stop being `latest`).

MikeMcQuaid:
@vitorgalvao Seems reasonable to me, thanks for checking ❤️

tgdnt:
@koutsenko this feature will presumably be in the next stable release of homebrew, that's why we don't have it yet by default. I tested it by creating a new branch within `cd $(brew --repository)` and merging origin/master into it.

ev0rtex:
I just pulled master to try this out and ran into a small issue. On my system I have the `virtualbox` and `virtualbox-extension-pack` casks installed. When I tried `brew cask upgrade` it successfully upgraded Virtualbox but it did not install the `virtualbox-extension-pack` as expected...and in fact seemed to have removed it. For whatever reason I seem to have broken my copy/paste so here's a screenshot of what happened:

brew_cask_issue

Could this be a side-effect of the upgrade process for `virtualbox`? It's a minor annoyance but still unexpected behavior.

reitermarkus:
@ev0rtex, this seems to be specific to `virtualbox`, which runs `brew cask uninstall virtualbox-extension-pack` in `uninstall_preflight`. Could you open an issue at https://github.com/caskroom/homebrew-cask?

PR #3401, opened 2017-10-31
go: various tweaks to recommendations

DomT4:

-----

CC @ilovezfs re https://github.com/Homebrew/homebrew-core/pull/19791#issuecomment-338984764.

### DIFF
--- a/Library/Homebrew/rubocops/text_cop.rb
+++ b/Library/Homebrew/rubocops/text_cop.rb
@@ -46,7 +46,7 @@ module RuboCop
           end
 
           find_method_with_args(body_node, :system, "go", "get") do
-            problem "Formulae should not use `go get`. If non-vendored resources are required use `go_resource`s."
+            problem "Do not use `go get`. Please ask upstream to implement Go vendoring"
           end
         end
       end
--- a/Library/Homebrew/rubocops/text_cop.rb
+++ b/Library/Homebrew/rubocops/text_cop.rb
@@ -51,5 +51,14 @@ module RuboCop
         end
       end
     end
+    module FormulaAuditStrict
+      class Text < FormulaCop
+        def audit_formula(_node, _class_node, _parent_class_node, body_node)
+          find_method_with_args(body_node, :go_resource) do
+            problem "`go_resource`s are deprecated. Please ask upstream to implement Go vendoring"
+          end
+        end
+      end
+    end
   end
 end
--- a/Library/Homebrew/test/rubocops/text_cop_spec.rb
+++ b/Library/Homebrew/test/rubocops/text_cop_spec.rb
@@ -66,7 +66,7 @@ describe RuboCop::Cop::FormulaAudit::Text do
 
           def install
             system "go", "get", "bar"
-            ^^^^^^^^^^^^^^^^^^^^^^^^^ Formulae should not use `go get`. If non-vendored resources are required use `go_resource`s.
+            ^^^^^^^^^^^^^^^^^^^^^^^^^ Do not use `go get`. Please ask upstream to implement Go vendoring
           end
         end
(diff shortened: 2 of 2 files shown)


ilovezfs:
The only issue I can foresee here would be that the mere fact that someone tried to use `go get` or a `go_resource` block in a formula doesn't itself imply that upstream hasn't already implemented Go vendoring. But I suppose if they go ask upstream for vendoring that is actually already there, upstream can tell them that is the case, which might be annoying for some upstreams, but is probably acceptable since it's a one-time annoyance.

DomT4:
Aye. I mean we could tell people to open the source tarball & look for a `vendor` directory before asking upstream but I couldn't think of a way to communicate that succinctly in a way that would be accessible to everyone however new to the project or not they are.

One thing that is potentially a hiccup: Things like `gox` could be considered semi-legitimate `go_resource`s. We could either whitelist `gox` and its deps or add a `gox` formula breaking the normal rule that formulae pegged to `revision`s should also be pegged to a `tag`, since the last `gox` tag was years ago.

ilovezfs:
>One thing that is potentially a hiccup: Things like gox could be considered semi-legitimate go_resources. We could either whitelist gox and its deps or add a gox formula breaking the normal rule that formulae pegged to revisions should also be pegged to a tag, since the last gox tag was years ago.

Right. I'm perfectly happy to ignore that for now since it's --strict only. My sense is that everything that "has to" be a `go_resource` could be made a formula one way or another but a little whitelist seems like a good interim solution.

DomT4:
> I'm perfectly happy to ignore that for now since it's --strict only.

Aye. I wanted to enable it always but it's just way too common at the moment & would be CI chaos 😢. I'll chuck a whitelist for the `gox` stuff in here with a small note that the whitelist should ideally not live on forever.

ilovezfs:
probably also `go-bindata`

ilovezfs:
and `github.com/rogpeppe/godeps` (and its dep `github.com/kisielk/gotool`), but maybe we can convince `charm` to adopt something more mainstream.

ilovezfs:
(only 12 installs of charm in the last 30 days …)

DomT4:
Will look at this one tonight after I get back from the gym.

DomT4:
Sigh. _Will_ look at this one soon. Having a bit of a not ideal end to the week 🙄. Apologies for the delay.

DomT4:
Oh heck, thanks for the reminder bot. I did indeed forget.

ilovezfs:
That stupid :robot: needs to learn to mind its own business.

DomT4:
1% of the time it's a useful reminder. It makes a nice change from me simply telling it to go away, I guess.

DomT4:
Go away.

MikeMcQuaid:
Would like to see this merged but let's not let it sit open fighting with stale bot forever.

MikeMcQuaid:
@DomT4 can you open a new PR when you've addressed the comments? Thanks (and do shout at me if you feel this process is offensively annoying).

DomT4:
Aye, sure, no worries. I'm home from uni in a few days so I should have a more predictable blob of time to look into Rubocop's syntax a little more.

ilovezfs:
I'm actually happy with this as-is without the whitelist. Gox finally created a tag for us so the biggest offender will be dispensed with shortly.

MikeMcQuaid:
Thanks again @DomT4!

ilovezfs:
homebrew/core tracking issue for `go_resource` block removals: https://github.com/Homebrew/homebrew-core/issues/21801

ilovezfs:
New formula for Gox as promised: https://github.com/Homebrew/homebrew-core/pull/22005

PR #3406, opened 2017-10-31
style: --staged option

maxim-belkin:

-----

`--staged` flag for `style` command to check only staged changes in files that have been staged for a commit.

### DIFF
--- a/Library/Homebrew/cmd/style.rb
+++ b/Library/Homebrew/cmd/style.rb
@@ -15,6 +15,8 @@
 #:
 #:    If `--except-cops` is passed, the given Rubocop cop(s)' checks would be skipped.
 #:
+#:    If `--staged` is passed, perform check of files staged for commit.
+#:
 #:    Exits with a non-zero status if any style violations are found.
 
 require "utils"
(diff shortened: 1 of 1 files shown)


ilovezfs:
```
git diff --name-only --cached | xargs -n1 brew style
```

maxim-belkin:
> ```
> git diff --name-only --cached | xargs -n1 brew style
> ```

That run `style` for all changes in the staged files.

maxim-belkin:
OK, I clarified what the changes do: check only staged changes in staged files.

ilovezfs:
I don't think we should be building git operations into `brew style`. I'd suggest putting this in a function in your bash profile or similar.

Library/Homebrew/cmd/style.rb
```diff
@@ -35,6 +37,15 @@ def style
       ARGV.formulae.map(&:path)
     end
 
+    # `--staged`: detect staged files
+    if ARGV.include? "--staged"
+      staged = `git -C "#{HOMEBREW_REPOSITORY}" diff --cached --name-only`.split
```
DomT4: `Utils.popen_read` looks less grim. Just saying 🙈

maxim-belkin: Are you suggesting something like this:
```
staged = Utils.popen_read("git", "-C", HOMEBREW_REPOSITORY.to_s, "diff", "--cached", "--name-only").split
```

ilovezfs:
Sorry @maxim-belkin going to pass on this PR.

Library/Homebrew/cmd/style.rb
DomT4: Yeah, basically. I don't know if Homebrew has an official policy on this & literally any maintainer can overrule me but I never miss an opportunity to get anal about backticks.

MikeMcQuaid:
Agreed here; I think things that can be achieved relatively simply on the shell are worth doing there unless there's a clear, wide demand for it.

PR #3411, opened 2017-11-01
formula_desc_cop: Whitelist "etc." and fix the full stop autocorrect

issyl0:
- [] Have you successfully run `brew tests` with your changes locally?

-----

As per https://github.com/Homebrew/brew/pull/2795#issuecomment-340287453, here's a PR to whitelist "etc." only in the full stop format_desc_cop, and fix the autocorrect so it works with quotes at the end of the line too. I ran a `brew audit --strict` and it stopped complaining about formulae with `etc.` at the end.

### DIFF
--- a/Library/Homebrew/rubocops/formula_desc_cop.rb
+++ b/Library/Homebrew/rubocops/formula_desc_cop.rb
@@ -103,7 +103,7 @@ module RuboCop
             correction.gsub!(/(^|[^a-z])#{@formula_name}([^a-z]|$)/i, "\\1\\2")
             correction.gsub!(/^(['"]?)\s+/, "\\1")
             correction.gsub!(/\s+(['"]?)$/, "\\1")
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Looks good, thanks @issyl0! Would you mind writing some tests to check these work as expected? Thanks again!

MikeMcQuaid:
@issyl0 Any news on that? If you'd rather not: no worries.

issyl0:
Sorry @MikeMcQuaid, this week has been stupidly busy and I'm unlikely to get time to do this soon. Sorry for leaving this hanging too. Also, from what I can see, the existing tests always test that violations are reported, not that they're ignored, if that makes any sense, and negating that for this one case feels odd.

MikeMcQuaid:
Will merge without tests this time but please write some in future. Thanks @issyl0!

PR #3457, opened 2017-11-18
Use patch found in the PATH

sjackman:

### DIFF
--- a/Library/Homebrew/patch.rb
+++ b/Library/Homebrew/patch.rb
@@ -64,10 +64,9 @@ class EmbeddedPatch
 
   def apply
     data = contents.gsub("HOMEBREW_PREFIX", HOMEBREW_PREFIX)
-    cmd = "/usr/bin/patch"
     args = %W[-g 0 -f -#{strip}]
-    IO.popen("#{cmd} #{args.join(" ")}", "w") { |p| p.write(data) }
-    raise ErrorDuringExecution.new(cmd, args) unless $CHILD_STATUS.success?
+    Utils.popen_write("patch", *args) { |p| p.write(data) }
+    raise ErrorDuringExecution.new("patch", args) unless $CHILD_STATUS.success?
   end
 
   def inspect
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'd rather we avoid any more of these until environment filtering is on by default. I think then we can just remove the hardcoded `/usr/bin` pretty much everywhere. Thoughts?

sjackman:
> I think then we can just remove the hardcoded `/usr/bin` pretty much everywhere.

That would be awesome.

sjackman:
Before I read your comment, I came up with a better solution using `with_system_path`. It may still be worth applying, even in advance of environment filtering. After environment filtering, we can probably get rid of `with_system_path` and its callers.

MikeMcQuaid:
> Before I read your comment, I came up with a better solution using with_system_path

I think this kinda changes the meaning of `with_system_path`, though.

Library/Homebrew/patch.rb
```diff
@@ -64,9 +64,9 @@ def contents; end
 
   def apply
     data = contents.gsub("HOMEBREW_PREFIX", HOMEBREW_PREFIX)
-    cmd = "/usr/bin/patch"
+    cmd = "patch"
```
MikeMcQuaid: Let's try just changing this to `patch` below and not use a variable (i.e. remove the `/usr/bin`). If it goes 💥 
I'll revert it.

sjackman: I got rid of the `cmd` variable. Did you mean also get rid of `with_system_path`?

sjackman:
>> Before I read your comment, I came up with a better solution using with_system_path

> I think this kinda changes the meaning of with_system_path, though.

You man not find the desired utility in `/usr/bin` or `/bin` on a Linux system, where it makes sense to use the brewed version of the utility, if it's installed. It could be renamed to `with_sanitized_path`.

Library/Homebrew/patch.rb
MikeMcQuaid: @sjackman Yep; make this just remove `/usr/bin/` and the variable usage.

sjackman: Done. I also replaced `IO.popen` with `Utils.popen_write`.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:

MikeMcQuaid:
@sjackman Now that environment filtering is on by default @sjackman you can feel free to do this anywhere else that makes sense (and remove a bunch of `with_system_path` usage, too.

sjackman:
Fantastic!

PR #3459, opened 2017-11-19
Custom commands can show help

wendorf:
#3385 introduced a change to path loading, and loads custom tap paths after trying to render help text. When trying to run something like `brew bundle --help`, the path is totally missed and bundle ends up getting called directly.

I tried to write an integration test for this in `custom-external-command_spec.rb`, but figuring out the interactions between the test, brew path loading, and the test fixture was too tough. I need to not specify a custom `PATH` in order to properly test the path recomposition in `brew.rb`, but I was unable to do that and still have my custom command detected. If you have guidance for this, I'll gladly update my PR.

### DIFF
--- a/Library/Homebrew/brew.rb
+++ b/Library/Homebrew/brew.rb
@@ -69,6 +69,14 @@ begin
     end
   end
 
+  unless internal_cmd
+    # Add contributed commands to PATH before checking.
+    homebrew_path.append(Tap.cmd_directories)
+
+    # External commands expect a normal PATH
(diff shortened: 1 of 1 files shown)


maelvalais:
Also broken:

    brew test-bot --help

MikeMcQuaid:
Nice work again @wendorf, thanks for the fix!

maelvalais:
Excellent! `brew test-bot --help` and `brew bundle --help` (tap version) work again!
Thank you @wendorf

PR #3465, opened 2017-11-21
Add CVSRequirement and SubversionRequirement

sjackman:

-----

Use these two requirements in DependencyCollector.

My primary intention for this PR is to install CVS and SVN when necessary on a Linux system.

### DIFF
--- a/Library/Homebrew/requirements.rb
+++ b/Library/Homebrew/requirements.rb
@@ -114,10 +114,15 @@ class ArchRequirement < Requirement
   end
 end
 
+class CVSRequirement < Requirement
+  fatal true
+  default_formula "cvs"
+  satisfy { which "cvs" }
+end
+
 class MercurialRequirement < Requirement
   fatal true
   default_formula "mercurial"
-
   satisfy { which("hg") }
 end
 
@@ -125,4 +130,10 @@ class GitRequirement < Requirement
   fatal true
   default_formula "git"
   satisfy { Utils.git_available? }
+end
+
+class SubversionRequirement < Requirement
+  fatal true
+  default_formula "subversion"
+  satisfy { Utils.svn_available? }
 end
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -159,7 +159,9 @@ class DependencyCollector
     elsif strategy <= BazaarDownloadStrategy
       Dependency.new("bazaar", tags)
     elsif strategy <= CVSDownloadStrategy
-      Dependency.new("cvs", tags) if MacOS.version >= :mavericks || !MacOS::Xcode.provides_cvs?
+      CVSRequirement.new(tags)
+    elsif strategy <= SubversionDownloadStrategy
+      SubversionRequirement.new(tags)
     elsif strategy < AbstractDownloadStrategy
       # allow unknown strategies to pass through
     else
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -174,10 +174,6 @@ module OS
         version < "4.3"
       end
 
-      def provides_cvs?
-        version < "5.0"
-      end
-
       def default_prefix?
         if version < "4.3"
           prefix.to_s.start_with? "/Developer"
--- a/Library/Homebrew/test/dependency_collector_spec.rb
+++ b/Library/Homebrew/test/dependency_collector_spec.rb
@@ -84,6 +84,18 @@ describe DependencyCollector do
       expect(subject.add(resource)).to be_an_instance_of(GitRequirement)
     end
 
+    it "creates a resource dependency from a CVS URL" do
+      resource = Resource.new
+      resource.url(":pserver:anonymous:@example.com:/cvsroot/foo/bar", using: :cvs)
+      expect(subject.add(resource)).to be_an_instance_of(CVSRequirement)
+    end
+
+    it "creates a resource dependency from a Subversion URL" do
+      resource = Resource.new
+      resource.url("svn://example.com/foo/bar")
+      expect(subject.add(resource)).to be_an_instance_of(SubversionRequirement)
+    end
+
     it "creates a resource dependency from a '.7z' URL" do
       resource = Resource.new
       resource.url("http://example.com/foo.7z")


Library/Homebrew/requirements.rb
```diff
@@ -114,10 +114,15 @@ def message
   end
 end
 
+class CVSRequirement < Requirement
+  fatal true
+  default_formula "cvs"
+  satisfy { which "cvs" }
```
reitermarkus: Do we still want the `MacOS.version >= :mavericks || !MacOS::Xcode.provides_cvs?` check on Mac? If not, is `MacOS::Xcode.provides_cvs?` still used anywhere else?

sjackman: It looks like `Xcode.provides_cvs?` is not used anywhere else, and we could get rid of it.
If we wanted to keep `MacOS.version >= :mavericks || !MacOS::Xcode.provides_cvs?`, then that check could be moved to `extend/os/mac/requirements.rb`.

sjackman: I've modified the PR to remove `MacOS::Xcode.provides_cvs?`.

Library/Homebrew/requirements.rb
```diff
@@ -114,10 +114,15 @@ def message
   end
 end
 
+class CVSRequirement < Requirement
+  fatal true
+  default_formula "cvs"
```
MikeMcQuaid: This isn't going to work as it's going to mean that bottled formulae always use the CVS formula rather than the system version.

Library/Homebrew/requirements.rb
```diff
@@ -126,3 +131,9 @@ class GitRequirement < Requirement
   default_formula "git"
   satisfy { Utils.git_available? }
 end
+
+class SubversionRequirement < Requirement
+  fatal true
+  default_formula "subversion"
```
MikeMcQuaid: This isn't going to work as it's going to mean that bottled formulae on macOS always use the Subversion formula rather than the system version.

Library/Homebrew/requirements.rb
MikeMcQuaid: Ignore my previous comment; it's right for a CVS formula to be used here 👍

Library/Homebrew/requirements.rb
MikeMcQuaid: Also, this probably warrants version check to make sure that the Subversion client is e.g. 1.8 or higher.

ilovezfs:
I think we should at least consider using `depends_on "cvs" => :build` and not adding a new requirement.

There is extremely limited use in formulae right now
```
curlftpfs.rb:7:  head ":pserver:anonymous:@curlftpfs.cvs.sourceforge.net:/cvsroot/curlftpfs", :using => :cvs
id3lib.rb:5:  head ":pserver:anonymous:@id3lib.cvs.sourceforge.net:/cvsroot/id3lib",
libowfat.rb:6:  head ":pserver:cvs:@cvs.fefe.de:/cvs", :using => :cvs
mandoc.rb:7:  head "anoncvs@mandoc.bsd.lv:/cvs", :using => :cvs
myman.rb:6:  head ":pserver:anonymous:@myman.cvs.sourceforge.net:/cvsroot/myman", :using => :cvs
ncp.rb:6:  head ":pserver:cvs:@cvs.fefe.de:/cvs", :using => :cvs
```
and I'd be :+1: on auditing it out of existence given that Apple removed cvs.
```
bash-3.2$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.1
BuildVersion:	17B48
bash-3.2$ which cvs
bash-3.2$ 
```

Library/Homebrew/requirements.rb
ilovezfs: How so? This will force the formula even `MacOS::Xcode.provides_cvs?` is true.

MikeMcQuaid: I think that's fine given how old those versions of macOS are at this point. I don't mind either way, though; if you'd rather this was handled by the Requirement: 👍 from me.

ilovezfs: I suppose someday our CVS formula might be something other than identical to Apple's.

Library/Homebrew/requirements.rb
sjackman: > This isn't going to work as it's going to mean that bottled formulae on macOS always use the Subversion formula rather than the system version.

Since it's a `:build` dependency, bottled formula will not require the subversion formula.

> Also, this probably warrants version check to make sure that the Subversion client is e.g. 1.8 or higher.

I'll add the version check.

ilovezfs: We don't want to be pouring the subversion bottle while it's building either.

sjackman: I've added the version check.

sjackman: > We don't want to be pouring the subversion bottle while it's building either.

`brew install -s foo` will not pour the subversion bottle. `brew install --build-bottle foo` will pour the subversion bottle. As far as I can tell, there are only two formulae that use `svn://` for their stable URL: `smpeg` and `smpeg2`. So I don't think it's a big concern in practice. If it's something that you'd like to fix all-the-same, I can create a `SubversionRequirement` for macOS in `extend/os/mac/requirements.rb` that avoids pulling in a `subversion` dependency on macOS for `svn://`.

```sh
grep 'url.*svn:.*:revision' -r $(brew --repo homebrew/core)
./smpeg2.rb:  url "svn://svn.icculus.org/smpeg/tags/release_2_0_0/", :revision => "408"
./smpeg.rb:  url "svn://svn.icculus.org/smpeg/tags/release_0_4_5/", :revision => "399"
```

ilovezfs: This thing should be a no-op on macOS as far as I'm concerned.

sjackman: Another option if it's palatable to you is…
```ruby
default_formula "subversion" unless OS.mac?
```

> This thing should be a no-op on macOS as far as I'm concerned.

Cool. In that case, on macOS I'll make it…
```ruby
class SubversionRequirement < Requirement
  satisfy { true }
end
```

sjackman: Do you prefe

1. in the same file
```ruby
satisfy {
  next true if OS.mac?
  …
}
```

2. in extend/os/mac/requirements.rb
```ruby
satisfy { true }
```

sjackman: Now that I'm looking at both, I prefer option `extend/os/mac/requirements.rb` myself.

MikeMcQuaid: > Now that I'm looking at both, I prefer option extend/os/mac/requirements.rb myself.

agreed

sjackman: Done.

sjackman:
CI is 🍏 again after fixing a style issue.

Library/Homebrew/extend/os/mac/requirements.rb
```diff
@@ -0,0 +1,5 @@
+class SubversionRequirementMac < Requirement
+  satisfy { true }
```
MikeMcQuaid: Maybe `which "subversion"` which is equivalent?

MikeMcQuaid: Also: thoughts on doing the opposite and making just a Linux requirement with the same name `SubversionRequirement` and just defines `default_formula`? Feel free to disagree; I realise you may have tried that.

MikeMcQuaid: Scratch the last comment: I no longer like it. Interested why this uses a different name, though.

sjackman: If it uses the same name `class SubversionRequirement`, the behaviour is to re-open and modify the existing generic class. That'd need something like `deafault_formula nil` to override the `default_formula` specified by the generic class. `satisfy { true }` also fails to replace existing `satisfied?` method due to this caching:
```ruby
@satisfied ||= Requirement::Satisfier.new(options, &block)
```
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirement.rb#L174

It seemed cleaner to just replace the existing generic class, rather than extend it.

MikeMcQuaid: Ok, seems fine. Let's just use `which "subversion"` here and I think it's good to go.

ilovezfs: Won't that change the current behavior and use whatever subversion is in the user's PATH?

ilovezfs: Also: isn't this supposed to be the purview of the svn superenv shim?

MikeMcQuaid: Actually, not quite: it seems the version check below should still work on macOS if you need to require a version that's higher than the system version. @ilovezfs, thoughts?

MikeMcQuaid: > Actually, not quite:

MikeMcQuaid: > Actually, not quite: 

Posted before @ilovezfs's comment showed up. Yes, good points about the `svn` superenv shim. @sjackman does that not work right on Linux? What's the difference between Subversion and Git, here?

ilovezfs: If we see such a situation in the wild where macOS system version is insufficient, we could consider revisiting this, but I don't think that is likely. And without a default formula, using `which "svn"` (I assume you meant `which "svn"` not `which "subversion"`) is a security vulnerability because it won't ensure the installed version of the formula is up to date if that's the svn first in the PATH.

MikeMcQuaid: > is a security vulnerability because it won't ensure the installed version of the formula is up to date if that's the svn first in the PATH.

It will, actually; if a requirement's executable resolves to a formula then it'll upgrade it.

ilovezfs: OK. I still think this is a bad idea because it allows the dependencies of subversion to be involved in the macOS builds unnecessarily.

ilovezfs: Also Homebrew svn doesn't recover from a canceled build.
```
==> Checking out 3094
svn up /Users/joe/Library/Caches/Homebrew/netpbm--svn -r 3094
svn: E155004: Run 'svn cleanup' to remove locks (type 'svn help cleanup' for details)
svn: E155004: Working copy '/Users/joe/Library/Caches/Homebrew/netpbm--svn' locked.
svn: E155004: '/Users/joe/Library/Caches/Homebrew/netpbm--svn' is already locked.
Error: Failed to download resource "netpbm"
Failure while executing: svn up /Users/joe/Library/Caches/Homebrew/netpbm--svn -r 3094
```

ilovezfs: It also fails to respect `export HOMEBREW_SVN=/usr/bin/svn` which simply hangs.

ilovezfs: `which "subversion"` will also make `brew pin subversion` impossible.

sjackman: I'm happy with either `true` or `which "svn"`. Let me know your preference or whether you need any additional input from me.

MikeMcQuaid: @sjackman I guess I'm interested on why a `SubversionRequirement` is needed but `GitRequirement` is not? FWIW if this PR (or a new PR) is scoped to just the `CVSRequirement` I'm happy to merge that as-is.

MikeMcQuaid: > I guess I'm interested on why a SubversionRequirement is needed but GitRequirement is not?

To expand on that further: both provide shims in `Library/Homebrew/shims/scm/` so it would be good to use a similar method to `Utils.git_path` in any `Requirement` if we need one (which the lack of a `GitRequirement` draws into question)

sjackman: `GitRequirement` does exist. Does that answer your question, or did I misunderstand?
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements.rb#L124
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dependency_collector.rb#L153-L154

ilovezfs: Why wouldn't an svn requirement do the same shim-aware
```
  def self.git_available?
    @git ||= quiet_system HOMEBREW_SHIMS_PATH/"scm/git", "--version"
  end
```

ilovezfs: and the same automatic install thing https://github.com/Homebrew/brew/blob/82aa148cb262cad831daa800da4f02d1b6e95dbf/Library/Homebrew/utils/git.rb#L47-L61

MikeMcQuaid: I think it's fine to have the automatic install be through the `default_formula` like Git is but I agree a `satisfy { Utils.svn_available? }` would be a good idea. Sorry for all the back and forth here, @sjackman.

sjackman: I think you meant for macOS, so I've changed `satisfy { true }` to `satisfy { Utils.svn_available? }`.

MikeMcQuaid: @sjackman Sorry for all the run around here but I think we can have a single, generic `SubversionRequirement` with a `default_formula` after all that checks `Utils.svn_available?` (i.e. exactly mirrors the `GitRequirement`). A version would be a nice addition if the tag is specified but I think you could punt on that for now if you wanted, too.

Apologies again for this; my knowledge of this part of the codebase is a bit rusty.

ilovezfs: :+1:

sjackman: Done!

MikeMcQuaid:
@sjackman https://github.com/Homebrew/brew/pull/3479 may also be of interest as that's going to make the `:build` requirements for e.g. Subversion less of a big deal for bottles.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #3469, opened 2017-11-22
Cleaner: ELF files are executable on Linux

sjackman:


MikeMcQuaid:
Breaking some Linux tests: https://travis-ci.org/Homebrew/brew/jobs/305958848

sjackman:
The binary files in `test/support/fixtures/tarballs/testball-0.1.tbz` are Mach-O executables.
We can add `testball-0.1-linux.tbz` with ELF executables.
Should I rename `testball-0.1.tbz` to `testball-0.1-mac.tbz`?

MikeMcQuaid:
`brew tests` still failing here but the approach in general looks sound.

sjackman:
Fixed the test failures.

Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
```diff
@@ -126,8 +126,8 @@ def setup_test_formula(name, content = nil)
       content = <<~EOS
         desc "Some test"
         homepage "https://example.com/#{name}"
-        url "file://#{TEST_FIXTURE_DIR}/tarballs/testball-0.1.tbz"
-        sha256 "#{TESTBALL_SHA256}"
+        url "file://#{TEST_FIXTURE_DIR}/tarballs/#{OS.linux? ? "testball-0.1-linux.tbz" : "testball-0.1.tbz"}"
+        sha256 "#{OS.linux? ? TESTBALL_SHA256_LINUX : TESTBALL_SHA256}"
```
MikeMcQuaid: Can this use the same approach as above?

Library/Homebrew/test/cleaner_spec.rb
```diff
@@ -20,7 +20,8 @@
 
       subject.clean
 
-      expect((f.bin/"a.out").stat.mode).to eq(0100555)
+      mach_executable_perm = OS.linux? ? 0100444 : 0100555
```
MikeMcQuaid: Could we set these to the same in the tarball instead?

sjackman: The file permissions in the tarball are the same. This test is related to a file outside of the tarball,`test/support/fixtures/mach/a.out` is a Mach-O file. On Linux, `Cleaner::executable_path?` changes ELF files and shell scripts to executable, but not Mach-O files. Options: I can create a separate `test/support/fixtures/elf` directory for Linux. Alternatively, I can simply disable this particular test on Linux using `:needs_macos`.

Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
sjackman: No. This code defines the location of the tarball and its checksum, which is platform dependent.
```ruby
Formulary.factory("testball1").stable.checksum.hexdigest
```

sjackman: It could calculate the checksum from the contents of the file, rather than store it in Ruby code.

MikeMcQuaid: > It could calculate the checksum from the contents of the file, rather than store it in Ruby code.

Yes, let's do this.

Library/Homebrew/test/cleaner_spec.rb
MikeMcQuaid: > I can create a separate test/support/fixtures/elf directory for Linux.

Sounds good 👍

sjackman: Done!

Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
sjackman: Done!

Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
```diff
@@ -123,11 +123,12 @@ def brew(*args)
   def setup_test_formula(name, content = nil)
     case name
     when /^testball/
+      tarball = TEST_FIXTURE_DIR/"tarballs/#{OS.linux? ? "testball-0.1-linux.tbz" : "testball-0.1.tbz"}"
```
MikeMcQuaid: Can you use an `if`/`else` here?

sjackman: Sure thing.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:

PR #3470, opened 2017-11-22
setup-analytics: Handle missing /proc/.../uuid

sjackman:


MikeMcQuaid:
When is this missing?

sjackman:
On [Windows Subsystem for Linux](https://msdn.microsoft.com/en-us/commandline/wsl/about), for example.

MikeMcQuaid:
And does it have `/usr/bin/uuidgen` present? If so, can we reverse the ordering so `/proc` is only checked if `uuidgen` isn't present?

sjackman:
Yes, it does have `/usr/bin/uuidgen`, at least with the default Ubuntu file system image.

sjackman:
> If so, can we reverse the ordering so /proc is only checked if uuidgen isn't present?

:+1: Done.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:

PR #3474, opened 2017-11-23
Fixed typo

melnikaite:
Replaced `formulae` with `formulas`


ilovezfs:
We use `formulae` for the plural through the code and documentation but thanks for the PR nonetheless @melnikaite!

PR #3478, opened 2017-11-24
lines_cop: Add condition to prevent false positives

GauthamGoli:

-----
Fixes this bug 
https://github.com/Homebrew/homebrew-core/pull/18736
https://github.com/Homebrew/homebrew-core/pull/18736#issuecomment-346739752

This was the original audit rule.
```ruby
if line =~ /(Dir\[("[^\*{},]+")\])/
      problem "#{Regexp.last_match(1)} is unnecessary; just use #{Regexp.last_match(2)}"
end
```
Issue:  I was just running the check on first argument without checking number of arguments but original rule I think doesn't match multiple arguments


MikeMcQuaid:
Thanks again @GauthamGoli!

PR #3483, opened 2017-11-26
Implement X11Requirement for Linux

sjackman:

-----

Move the XQuartz implementation to extend/os/mac/requirements.

### DIFF
--- a/Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
+++ b/Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
@@ -1,0 +1,42 @@
+require "requirement"
+
+class XQuartzRequirement < Requirement
+  include Comparable
+
+  fatal true
+  cask "xquartz"
+  download "https://xquartz.macosforge.org"
+
+  env { ENV.x11 }
+
+  def initialize(name = "x11", tags = [])
+    @name = name
+    # no-op on version specified as a tag argument
+    tags.shift if /(\d\.)+\d/ =~ tags.first
+    super(tags)
+  end
+
+  def min_version
+    MacOS::XQuartz.minimum_version
+  end
+
+  satisfy build_env: false do
+    next false unless MacOS::XQuartz.installed?
+    min_version <= MacOS::XQuartz.version
+  end
+
+  def message
+    "XQuartz #{min_version} (or newer) is required to install this formula. #{super}"
+  end
+
+  def <=>(other)
+    return unless other.is_a? X11Requirement
+    0
+  end
+
+  def inspect
+    "#<#{self.class.name}: #{name.inspect} #{tags.inspect}>"
+  end
+end
+
+X11Requirement = XQuartzRequirement
--- a/Library/Homebrew/requirements/x11_requirement.rb
+++ b/Library/Homebrew/requirements/x11_requirement.rb
@@ -4,8 +4,6 @@ class X11Requirement < Requirement
   include Comparable
 
   fatal true
-  cask "xquartz"
-  download "https://xquartz.macosforge.org"
 
   env { ENV.x11 }
 
@@ -17,24 +15,33 @@ class X11Requirement < Requirement
   end
 
   def min_version
-    # TODO: remove in https://github.com/Homebrew/brew/pull/3483
-    return Version::NULL unless OS.mac?
+    "1.12.2"
+  end
 
-    MacOS::XQuartz.minimum_version
+  def min_xdpyinfo_version
+    "1.3.0"
   end
 
   satisfy build_env: false do
-    # TODO: remove in https://github.com/Homebrew/brew/pull/3483
-    next false unless OS.mac?
-
-    next false unless MacOS::XQuartz.installed?
-    min_version <= MacOS::XQuartz.version
+    if which_xorg = which("Xorg")
+      version = Utils.popen_read which_xorg, "-version", err: :out
+      next false unless $CHILD_STATUS.success?
+      version = version[/X Server (\d+\.\d+\.\d+)/, 1]
+      next false unless version
+      Version.new(version) >= min_version
+    elsif which_xdpyinfo = which("xdpyinfo")
+      version = Utils.popen_read which_xdpyinfo, "-version"
+      next false unless $CHILD_STATUS.success?
+      version = version[/^xdpyinfo (\d+\.\d+\.\d+)/, 1]
+      next false unless version
+      Version.new(version) >= min_xdpyinfo_version
+    else
+      false
+    end
   end
 
   def message
-    s = "XQuartz #{min_version} (or newer) is required to install this formula."
-    s += super
-    s
+    "X11 is required to install this formula, either Xorg #{min_version} or xdpyinfo #{min_xdpyinfo_version}, or newer. #{super}"
   end
 
   def <=>(other)
@@ -46,3 +53,5 @@ class X11Requirement < Requirement
     "#<#{self.class.name}: #{name.inspect} #{tags.inspect}>"
   end
 end
+
+require "extend/os/requirements/x11_requirement"
--- a/Library/Homebrew/extend/os/requirements/x11_requirement.rb
+++ b/Library/Homebrew/extend/os/requirements/x11_requirement.rb
@@ -1,0 +1,1 @@
+require "extend/os/mac/requirements/x11_requirement" if OS.mac?


Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -5,8 +5,7 @@ class X11Requirement < Requirement
   attr_reader :min_version
 
   fatal true
-  cask "xquartz"
-  download "https://xquartz.macosforge.org"
+  default_formula "linuxbrew/xorg/xorg"
```
ilovezfs: having a default formula in a non-core tap seems like asking for trouble

MikeMcQuaid: Yeh, I'd rather avoid `linuxbrew` tap usage for now until we're closer to merging the projects.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
```diff
@@ -0,0 +1,49 @@
+require "requirement"
+
+Object.send(:remove_const, :X11Requirement)
```
MikeMcQuaid: This is a bit weird.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
```diff
@@ -0,0 +1,49 @@
...
+  fatal true
+  cask "xquartz"
+  download "https://xquartz.macosforge.org"
+
+  env { ENV.x11 }
+
+  def initialize(name = "x11", tags = [])
+    @name = name
+    if /(\d\.)+\d/ =~ tags.first
+      @min_version = Version.create(tags.shift)
```
MikeMcQuaid: Do XQuartz versions map to X11/Xorg versions at all? Would be good to share the version logic if possible.

scpeters: The `About` window in my xquartz says: "XQuartz 2.7.11 (xorg-server 1.18.4)"

Library/Homebrew/requirements/x11_requirement.rb
sjackman: Okay. How about `default_formula "linuxbrew/xorg/xorg"` in `extend/os/linux/requirements/x11_requirement.rb`?

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
sjackman: This line removes the class `X11Requirement` so that `extend/os/mac/requirements/x11_requirement.rb` can redefine it. Creating a new class with a new name like `XQuartzRequirement` and then assigning it with `X11Requirement = XQuartzRequirement` causes a test to fail because `be_an_instance_of(X11Requirement)` is false. We could make `X11Requirement` be a base class, and have `XQuartzRequirement` inherit from it. Alternatively, we could remove this test. Your thoughts?

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
sjackman: We haven't yet found a need in Linuxbrew for a version check in the X11Requirement. This code is here primarily so that this Requirement exposes the same version API as the XQuartz Requirement, so that the tests pass. I can instead remove this version logic and make that particular test (just one) `:needs_macos`.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
sjackman: Never mind. I replaced it with `X11Requirement = XQuartzRequirement`.

Library/Homebrew/requirements/x11_requirement.rb
ilovezfs: How does that change anything? It's still a non-core (i.e. not Linuxbrew/homebrew-core) tap.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
ilovezfs: The point is when it's satisfied by something other than the default formula (which will be forced to be up to date regardless), which I'm assuming remains untested.

sjackman: Before the `xorg` formula existed, I used Linuxbrew with the host's X11 implementation. When installing a bottle that depends on `:x11`, the `default_formula` is used. The host's Xorg may be used when installing something from source that depends on `:x11`.

ilovezfs: That's the case where there should be a version check.

sjackman: I haven't run into a failure in the wild where the host X11 was too old. I'd rather not fix problems that haven't been encountered by users.

ilovezfs: The lack of proper version handling is why the Java requirement is currently broken on Linux. I don't think we should repeat that again.

sjackman:
```
❯❯❯ grep -r 'depends_on :x11 =>.*[0-9]' .
./gv.rb:  depends_on :x11 => "2.7.2"
```
Is `gv` the only formula that has a minimum version requirement for X11?

ilovezfs: The fact that they all don't have it is basically a bug. The bottles usually break when they were built with a different version from what the user has. We even discussed disabling all bottles on formula with an x11 requirement at one point because of that.

sjackman: We haven't encountered that issue with Xorg on Linux. When the user installs a bottle that depends on `:x11`, it uses the `xorg` formula, which upgrades that formula.

ilovezfs: Same behavior for ruby and perl and yet we have version checks.

ilovezfs: What is the downside to having a feature complete implementation of the requirement?

sjackman: It's unnecessary code if it solves a hypothetical problem.

ilovezfs: And yet here we have the broken Java requirement, now with a stale label, https://github.com/Linuxbrew/brew/issues/509.

sjackman: Java has 168 formulae with a versioned dependency. X11 has one. These are not comparable.
```
❯❯❯ grep -r 'depends_on :java => "[0-9]' $(brew --repo homebrew/core) | wc -l
     168
```

ilovezfs: That's because we've set a lax >= on those as a matter of course.

ilovezfs: More generally, I don't like the continuing precedent this is setting. macOS has a requirement with a version check, and Linuxbrew has a default formula without a version check until such time as something breaks, incurring a technical debt and burden shifting the completion of the work to the future.

ilovezfs: Also, given over 40% of the x11 requirements are optional, meaning the dependent's bottle won't be used, it seems particularly relevant to have some sort of version check on the system x11 before brew decides the default formula shouldn't be used to satisfy the requirement after all.

ilovezfs: Note, too, that `check_xquartz_up_to_date` (https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/os/mac/diagnostic.rb#L290-L299) warns about outdated XQuartz on macOS, which likely papers over some of the issues with older versions.

Library/Homebrew/extend/os/linux/requirements/x11_requirement.rb
```diff
@@ -0,0 +1,3 @@
+class X11Requirement
+  default_formula "linuxbrew/xorg/xorg"
```
MikeMcQuaid: I'd rather avoid any `linuxbrew` tap references until we are further along with the migration to merge Linuxbrew/brew and Homebrew/brew.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
MikeMcQuaid: I agree it would be good to have a version check here.

Library/Homebrew/extend/os/linux/requirements/x11_requirement.rb
sjackman: I've removed `extend/os/linux/requirements/x11_requirement.rb`

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
sjackman: I've added a version check that `xdpyinfo -version` is at least version 1.3.1. The most recent version is 1.3.2. I tested that `gv` works on CentOS 6 with the host's Xorg (with `xdpyinfo -version` of 1.3.1) using `brew install -s gv`.

MikeMcQuaid: How does that correspond to e.g. the X.org 7.7 release: https://www.x.org/wiki/Releases/ ?

MikeMcQuaid: Looks like checking `/usr/bin/X --version` may be the most authoritative with that in mind.

sjackman: Headless Linux servers don't usually have the X server installed (`/usr/bin/X`), just the X client. The X server is run on your local machine (which is not necessarily running Linuxbrew), whereas the X client is run on the headless server (which is running Linuxbrew). For this reason, I'm checking the version of a X client executable (`/usr/bin/xdpyinfo`) rather than the X server.

The most recent complete release of Xorg X11R7.7 shipped with `xdpyinfo 1.3.0`.
The previous release of Xorg X11R7.6 shipped with `xdpyinfo 1.2.0`.
See https://www.x.org/releases/X11R7.7/changelog.html

MikeMcQuaid: That's the `xdpyinfo` version rather than the X11 version, though; it's not particularly useful in this case. Can you see if there's anything else that provides the actual X11 version even with the X server not installed? If not, I think this should just assume the server is not headless and the Linux requirement can always override that.

ilovezfs: The katamari 7.7 had server version 1.12.2 and xdpyinfo 1.3.0. So I think starting with `Xorg -version` >= 1.12.2 makes sense, and then if the `Xorg` command isn't available, fall back on `xdpyinfo -version` > = 1.3.0.

Linux:
```
openzfs ~ # Xorg -version 2>&1|grep Server
X.Org X Server 1.19.3
openzfs ~ # xdpyinfo -version
xdpyinfo 1.3.2
```

macOS with XQuartz `2.7.11`:
```
iMac-TMP:~ joe$ Xorg -version 2>&1|grep Server
X.Org X Server 1.18.4
iMac-TMP:~ joe$ xdpyinfo -version
xdpyinfo 1.3.2
```

ilovezfs: @sjackman are you planning on actually using the `satisfy` block from this upstreamed code or just continuing to use
```
  satisfy build_env: false do
    Formula["linuxbrew/xorg/xorg"].installed?
  end
```
in Linuxbrew itself?

MikeMcQuaid:
@sjackman My apologies but I misunderstood my conversation previously with @ilovezfs. Instead of needing a version tag that can be passed through let's just make the requirement enforce a minimum `X -version` and, if not available, `xdpyinfo -version` within the requirement to a value you're happy with. I'd suggest after that (doesn't need to be part of this PR) we remove the existing version that's passed through as part of the requirement to make it a no-op and then remove it from formulae. Sorry about all the back and forth 😭

ilovezfs:
minor thing: I don't think `X` command prints a version for Xquartz but `Xorg` command does.
```
iMac-TMP:~ joe$ X -version
iMac-TMP:~ joe$ Xorg -version

X.Org X Server 1.18.4
Release Date: 2016-07-19
X Protocol Version 11, Revision 0
Build Operating System: Darwin 14.0 x86_64 
Current Operating System: Darwin iMac-TMP.local 15.6.0 Darwin Kernel Version 15.6.0: Sun Jun  4 21:43:07 PDT 2017; root:xnu-3248.70.3~1/RELEASE_X86_64 x86_64
Build Date: 25 October 2016  10:12:52PM
 
Current version of pixman: 0.34.0
	Before reporting problems, check http://wiki.x.org
	to make sure that you have the latest version.
iMac-TMP:~ joe$ 
```

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
sjackman: Yes, once it's merged into Homebrew/brew, and then after that, merged into Linuxbrew/brew.

Library/Homebrew/test/x11_requirement_spec.rb
```diff
@@ -20,7 +20,7 @@
       expect(subject).not_to eql(other)
     end
 
-    it "returns false if the minimum version differs" do
+    it "returns false if the minimum version differs", :needs_macos do
```
MikeMcQuaid: Can you instead modify this test so it does something on Linux?

sjackman: I've added a test for Linux.

Library/Homebrew/test/spec_helper.rb
```diff
@@ -61,6 +61,10 @@
     skip "Needs official command Taps." unless ENV["HOMEBREW_TEST_OFFICIAL_CMD_TAPS"]
   end
 
+  config.before(:each, :needs_linux) do
```
MikeMcQuaid: I'd rather we didn't have these as they show up as `pending` tests on macOS (I realise the inverse is the case on Linux). Here's a solution for pending tests (that I think we should do for Linux and, I guess, macOS): https://github.com/rspec/rspec-core/issues/2377#issuecomment-275131981 but in this particular case I think an `if`/`else` within the test itself is better as they are testing essentially the same functionality. CC @reitermarkus for RSpec thoughts.

MikeMcQuaid: I've tweaked the macOS underlying code so this may no longer be needed.

sjackman: > I've tweaked the macOS underlying code so this may no longer be needed.

Meaning the PR is good as is, or you'd prefer the `if`/`else` instead?

sjackman: Ah, I think you're referring to 8b9ac2b2e030a70bd6ee99ac8b0c9b9f92fd621b `x11_requirement: remove custom minimum version.` I'll rebase onto `master`.

sjackman:
I've rebased this PR onto `master`.

sjackman:
Thanks for the reorganization of the macOS X11Requirement, Mike.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  fatal true
+  cask "xquartz"
+  download "https://xquartz.macosforge.org"
+
+  env { ENV.x11 }
+
+  def initialize(name = "x11", tags = [])
+    @name = name
+    # no-op on version specified as a tag argument
```
MikeMcQuaid: Let's do the same thing on the generic X11 requirement; they should have the same interface. I think the only thing they should probably differ on is the `satisfy` and `min_version` stuff (and would be good to have a `min_version` method there too).

Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -17,32 +13,26 @@ def initialize(name = "x11", tags = [])
   end
 
   def min_version
```
MikeMcQuaid: Maybe a `min_xdpyinfo_version` method too would be good.

Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -17,32 +13,26 @@ def initialize(name = "x11", tags = [])
...
-    # TODO: remove in https://github.com/Homebrew/brew/pull/3483
-    return Version::NULL unless OS.mac?
-
-    MacOS::XQuartz.minimum_version
+    Version.new "1.12.2"
```
MikeMcQuaid: Don't think this actually needs a `Version.new` given you're always comparing them to `Version`s below.

Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -17,32 +13,26 @@ def initialize(name = "x11", tags = [])
...
-    min_version <= MacOS::XQuartz.version
-  end
-
-  def message
-    s = "XQuartz #{min_version} (or newer) is required to install this formula."
-    s += super
-    s
-  end
-
-  def <=>(other)
```
MikeMcQuaid: Want to keep this method in some form for API compatability

Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -17,32 +13,26 @@ def initialize(name = "x11", tags = [])
...
 
   satisfy build_env: false do
-    # TODO: remove in https://github.com/Homebrew/brew/pull/3483
-    next false unless OS.mac?
-
-    next false unless MacOS::XQuartz.installed?
-    min_version <= MacOS::XQuartz.version
-  end
-
-  def message
```
MikeMcQuaid: Want to keep this method in some form.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
sjackman: These two `initialize` functions are identical. Can you please elaborate?

Generic: https://github.com/sjackman/homebrew/blob/9f7278899e24fa3567db7e15050a63e8f3242679/Library/Homebrew/requirements/x11_requirement.rb#L8

macOS: https://github.com/sjackman/homebrew/blob/9f7278899e24fa3567db7e15050a63e8f3242679/Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb#L12

sjackman:
Fixed up.

Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb
MikeMcQuaid: Apologies, I misread this.

Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -17,32 +15,39 @@ def initialize(name = "x11", tags = [])
...
+    else
+      false
+    end
   end
 
   def message
-    s = "XQuartz #{min_version} (or newer) is required to install this formula."
-    s += super
-    s
+    "X11 is required to install this formula.\n"
```
MikeMcQuaid: Insert `#{min_version}` in here somewhere and may as well `+= super`, too.

Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -17,32 +15,39 @@ def initialize(name = "x11", tags = [])
...
   end
 
   def <=>(other)
     return unless other.is_a? X11Requirement
     0
   end
-
-  def inspect
-    "#<#{self.class.name}: #{name.inspect} #{tags.inspect}>"
-  end
```
MikeMcQuaid: Let's keep this method as-is.

Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -17,32 +15,39 @@ def initialize(name = "x11", tags = [])
...
   end
 
   satisfy build_env: false do
-    # TODO: remove in https://github.com/Homebrew/brew/pull/3483
-    next false unless OS.mac?
-
-    next false unless MacOS::XQuartz.installed?
-    min_version <= MacOS::XQuartz.version
+    if which "Xorg"
+      version = Utils.popen_read "Xorg", "-version", err: :out
```
ilovezfs: @sjackman I think this needs to pass the return value of `which "Xorg"` to `Utils.popen_read`, otherwise when env filtering is enabled, it will end up with `command not found: Xorg`. Ditto `xdpyinfo` below.

ilovezfs:
@sjackman once this is merged, non-formulae (i.e. x11 other than linuxbrew/xorg/xorg)  will be allowed to satisfy the x11 requirement. But as far as I can tell, Linuxbrew doesn't have an `ENV.x11` implementation. So won't non-formulae satisfiers cause a problem until `ENV.x11` is actually implemented?

It's not relevant to whether this PR gets merged in Homebrew/brew now as-is, but it does seem like in its current form it can't be merged into Linuxbrew/brew.

Library/Homebrew/requirements/x11_requirement.rb
sjackman: I believe that both `which` and `Utils.popen_read` will experience the same `PATH`. Either they'll both succeed, or they'll both fail. Do you believe that one of the two experiences env filtering, while the other does not? When `Xorg` and `xdpyinfo` are in the standard system path, both `which` and `Utils.popen_read` should work with env filtering enabled.

sjackman:
With stdenv, the default no-op `dev x11; end` is sufficient. Linuxbrew/brew has an implementation of `ENV.x11` for superenv, but superenv hasn't yet been contributed back to Homebrew/brew. It's coming.

Library/Homebrew/requirements/x11_requirement.rb
ilovezfs: >I believe that both which and Utils.popen_read will experience the same PATH. 

They will not. My comment was not speculative but based on testing. `which` is Homebrew's custom `which`, which brings back in the ORIGINAL_PATH.

Other requirements use Homebrew's `which` for this reason. There's no guarantee things used to satisfy requirements are in the env filtered PATH.

sjackman: Please point me to the relevant code.
`which` uses `PATH` 
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils.rb#L309
`popen_read` uses `PATH`
https://github.com/Homebrew/brew/blob/master/Library/Homebrew/utils/popen.rb#L18
What you're saying may well be true, but it's not evident from the code.

ilovezfs: [Requirement#which](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirement.rb#L158-L160) and [Requirment#which_all](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirement.rb#L162-L164)

sjackman: Fixed. Thanks for the feedback, Joe.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks again for merging, Mike! Thanks for the feedback, @ilovezfs!

MikeMcQuaid:
```
$ brew tests --only=x11_requirement 
Using recorded test runtime
1 processes for 1 specs, ~ 1 specs per process
/usr/local/Homebrew/Library/Homebrew/extend/os/mac/requirements/x11_requirement.rb:42: warning: already initialized constant X11Requirement
/usr/local/Homebrew/Library/Homebrew/requirements/x11_requirement.rb:3: warning: previous definition of X11Requirement was here
```

@sjackman can you address this (today if possible of I may need to revert)? Check out how `JavaRequirement` avoids this.

sjackman:
Yes, I can address it now.

sjackman:
Fixed.

PR #3492, opened 2017-11-28
Properly upgrade Casks with version :latest

amyspark:

-----

When I proposed #3396, I forgot to test if the upgrade process worked for `version :latest` Casks.
Obviously, this came back to bite me 😢 

In this initial state, I've submitted a test to summarize the issue. The missing part would be to tamper with `staged_path` if `version :latest` and use a randomized? path, but I'd like to request your comments on the best way to do it.

This pull request, once finished, fixes caskroom/homebrew-cask#41373.

Thanks for your comments!

/cc @MikeMcQuaid @vitorgalvao @reitermarkus

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
@@ -70,8 +70,6 @@ module Hbc
         else
           command.run("/bin/mv", args: [target, source], sudo: true)
         end
-
-        add_altname_metadata(source, target.basename, command: command)
       end
 
       def delete(target, force: false, command: nil, **_)
--- a/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
@@ -71,7 +71,7 @@ module Hbc
           command.run("/bin/mv", args: [target, source], sudo: true)
         end
 
-        add_altname_metadata(target, source.basename, command: command)
+        add_altname_metadata(source, target.basename, command: command)
       end
 
       def delete(target, force: false, command: nil, **_)
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -29,9 +29,10 @@ module Hbc
       @require_sha = require_sha
       @reinstall = false
       @upgrade = upgrade
+      @backed_up = false
     end
 
-    attr_predicate :binaries?, :force?, :skip_cask_deps?, :require_sha?, :upgrade?, :verbose?
+    attr_predicate :binaries?, :force?, :skip_cask_deps?, :require_sha?, :upgrade?, :verbose?, :backed_up?
 
     def self.print_caveats(cask)
       odebug "Printing caveats"
@@ -378,26 +379,26 @@ module Hbc
 
       disable_accessibility_access
       uninstall_artifacts
-      backup if version_is_latest?
+      backup
     end
 
     def backup
-      @cask.staged_path.rename backup_path(@cask.staged_path)
+      @backed_up = true
+      @cask.staged_path.rename backup_path
     end
 
     def restore_backup
-      path = backup_path(@cask.staged_path)
-
-      return unless path.directory?
+      return unless backup_path.directory?
 
       Pathname.new(@cask.staged_path).rmtree if @cask.staged_path.exist?
 
-      path.rename @cask.staged_path
+      backup_path.rename @cask.staged_path
+      @backed_up = false
     end
 
     def revert_upgrade
       opoo "Reverting upgrade for Cask #{@cask}"
-      restore_backup if version_is_latest?
+      restore_backup
       install_artifacts
       enable_accessibility_access
     end
@@ -436,8 +437,13 @@ module Hbc
       purge_caskroom_path
     end
 
-    def backup_path(path)
-      Pathname.new "#{path}.upgrading" unless path.nil?
+    def backup_path
+      return nil if @cask.staged_path.nil?
+      if backed_up?
+        Pathname.new "#{@cask.staged_path}.upgrading"
+      else
+        @cask.staged_path
+      end
     end
 
     def version_is_latest?
@@ -452,8 +458,8 @@ module Hbc
       ohai "Purging files for version #{@cask.version} of Cask #{@cask}"
 
       # versioned staged distribution
-      if upgrade? && version_is_latest?
-        staged_path = backup_path(@cask.staged_path)
+      if upgrade?
+        staged_path = backup_path
       else
         staged_path = @cask.staged_path
       end
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -378,10 +378,26 @@ module Hbc
 
       disable_accessibility_access
       uninstall_artifacts
+      backup if version_is_latest?
+    end
+
+    def backup
+      @cask.staged_path.rename backup_path(@cask.staged_path)
+    end
+
+    def restore_backup
+      path = backup_path(@cask.staged_path)
+
+      return unless path.directory?
+
+      Pathname.new(@cask.staged_path).rmtree if @cask.staged_path.exist?
+
+      path.rename @cask.staged_path
     end
 
     def revert_upgrade
       opoo "Reverting upgrade for Cask #{@cask}"
+      restore_backup if version_is_latest?
       install_artifacts
       enable_accessibility_access
     end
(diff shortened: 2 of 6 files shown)


reitermarkus:
> Obviously, this came back to bite me 😢

No worries, there had to be *something*. 😜

> and use a randomized? path

`brew reinstall` appends `.reinstall` to the Keg path: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/reinstall.rb#L74-L76

I'd suggest to just move `staged_path` to `"#{staged_path}.backup"` (or `.upgrading`), and either remove it after the upgrade, or move it back if the upgrade fails.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,6 +436,14 @@ def zap
       purge_caskroom_path
     end
 
+    def backup_path(path)
+      Pathname.new "#{path}.upgrading" unless path.nil?
```
reitermarkus: When would `path` be `nil`?

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -378,10 +378,26 @@ def start_upgrade
 
       disable_accessibility_access
       uninstall_artifacts
+      backup if version_is_latest?
```
reitermarkus: Let's always create a backup, not just when `version_is_latest?`.

Library/Homebrew/cask/lib/hbc/installer.rb
reitermarkus: Also, since this is only used for `staged_path`, you could make this take no arguments and just return `Pathname("#{@cask.staged_path}.upgrading")`.

amyspark: `path` should never be  `nil`, but I wanted to preserve the semantics of `gain_permissions_remove(staged_path) if !staged_path.nil? && staged_path.exist?`.

Library/Homebrew/cask/lib/hbc/installer.rb
amyspark: If we always create a backup, we'll need a way to know if this is the new or old cask when `purge_versioned_files`. Otherwise, it will leave a dangling, empty folder if the upgrade fails (see the test `"restores the old Cask if the upgrade failed"` in `upgrade_spec.rb`)

reitermarkus: Isn't `start_upgrade` only called on the old cask?

amyspark: Yes, I took advantage of that; see next commit.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,6 +437,19 @@ def zap
       purge_caskroom_path
     end
 
+    def backup_path
+      return nil if @cask.staged_path.nil?
+      if backed_up?
+        Pathname.new "#{@cask.staged_path}.upgrading"
+      else
+        @cask.staged_path
```
reitermarkus: I think this is really hard to follow. The only reason for this is `purge_versioned_files` using it. I'd suggest making two specialised versions of `purge_versioned_files`, one normal one, and one for the `upgrade?` case.

amyspark: I'll do it that way, and then factor out the whole metadata handling (to avoid code duplication).

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,19 +434,43 @@ def zap
...
       ohai "Purging files for version #{@cask.version} of Cask #{@cask}"
 
       # versioned staged distribution
       gain_permissions_remove(@cask.staged_path) if !@cask.staged_path.nil? && @cask.staged_path.exist?
 
       # Homebrew-Cask metadata
+      purge_metadata
+    end
+
+    def purge_metadata
```
reitermarkus: The lines in here which have `unless upgrade?` could be moved to `purge_versioned_files` now.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,19 +434,43 @@ def zap
...
       if @cask.metadata_versioned_path.respond_to?(:children) &&
-         @cask.metadata_versioned_path.exist?
+         @cask.metadata_versioned_path.exist? &&
+         !(upgrade? && version_is_latest?)
```
reitermarkus: Do we still need `version_is_latest?` now?

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,19 +434,43 @@ def zap
...
+
+    def version_is_latest?
+      @cask.versions.include?("latest")
+    end
+
     def gain_permissions_remove(path)
       Utils.gain_permissions_remove(path, command: @command)
     end
 
+    def purge_backed_versioned_files
```
reitermarkus: `purge_backed_up_versioned_files`

Library/Homebrew/cask/lib/hbc/installer.rb
amyspark: Yes, the backup mechanism doesn't involve the Cask's metadata; if the version is `latest`, deleting this line will make `purge_metadata` "uninstall" the Cask while leaving all artifacts in place.

amyspark: On second thoughts, I'll just go ahead with two fully separate methods and get rid of that line.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,10 +437,46 @@ def zap
...
+      return nil if @cask.staged_path.nil?
+      Pathname.new "#{@cask.staged_path}.upgrading"
+    end
+
+    def backup_metadata_path
+      return nil if @cask.metadata_versioned_path.nil?
+      Pathname.new "#{@cask.metadata_versioned_path}.upgrading"
+    end
+
+    def version_is_latest?
```
reitermarkus: This can now be removed.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,10 +437,46 @@ def zap
...
+      if backup_metadata_path.respond_to?(:children) &&
+         backup_metadata_path.exist? &&
+         backup_metadata_path.children.each do |subdir|
+           unless PERSISTENT_METADATA_SUBDIRS.include?(subdir.basename)
+             gain_permissions_remove(subdir)
+           end
+         end
+      end
+      backup_metadata_path.rmdir_if_possible
+      @cask.metadata_master_container_path.rmdir_if_possible unless upgrade?
```
reitermarkus: The `unless upgrade?` here doesn't make sense, because this method is only called when upgrading.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -31,7 +31,7 @@ def initialize(cask, command: SystemCommand, force: false, skip_cask_deps: false
       @upgrade = upgrade
     end
 
-    attr_predicate :binaries?, :force?, :skip_cask_deps?, :require_sha?, :upgrade?, :verbose?
+    attr_predicate :binaries?, :force?, :skip_cask_deps?, :require_sha?, :upgrade?, :verbose?, :backed_up?
```
reitermarkus: This is not needed anymore.

Library/Homebrew/cask/lib/hbc/artifact/moved.rb
```diff
@@ -71,7 +71,7 @@ def move_back(skip: false, force: false, command: nil, **options)
           command.run("/bin/mv", args: [target, source], sudo: true)
         end
 
-        add_altname_metadata(target, source.basename, command: command)
+        add_altname_metadata(source, target.basename, command: command)
```
reitermarkus: Remove this completely. We don‘t need to add this when moving it back, actually.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -420,10 +437,38 @@ def zap
...
 
+    def purge_backed_up_versioned_files
+      ohai "Purging files for version #{@cask.version} of Cask #{@cask}"
+
+      # versioned staged distribution
+      gain_permissions_remove(backup_path) if !backup_path.nil? && backup_path.exist?
+
+      # Homebrew-Cask metadata
+      if backup_metadata_path.respond_to?(:children) &&
+         backup_metadata_path.exist? &&
```
reitermarkus: The `&&` shouldn‘t be here.

reitermarkus: The whole condition here could probably just be `backup_metadata_path.directory?`.

reitermarkus:
You probably missed this comment by a minute: https://github.com/Homebrew/brew/pull/3492#discussion_r153669179

amyspark:
Yep, I wasn't using my inbox to keep up with comments. Thanks for the heads up @reitermarkus.

reitermarkus:
Thanks again, @amyspark! 👍

PR #3501, opened 2017-11-29
Adding proxy variables to FILTERED_ENV

### LINKED ISSUE #3500: https_proxy not in FILTERED_ENV

- [x] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at the relevant tap e.g. for Homebrew/homebrew-core https://github.com/Homebrew/homebrew-core/issues/new

- run `brew update` behind proxy
- got `fatal: unable to access 'https://github.com/Homebrew/brew.git/': Failed to connect to github.com port 443: Operation timed out`
  - git command failed (`/Applications/Xcode.app/Contents/Developer/usr/bin/git fetch --tags --force -q origin refs/heads/master:refs/remotes/origin/master`)

# Features

- https://github.com/Homebrew/brew/blob/master/bin/brew#L91
  - there is no http**s**_proxy in FILTERED_ENV (but http_proxy)
    - see: #3494 / 63408a10ff677909f4397c5c94cfa1f2007c0826
  - please add `https_proxy`, `ftp_proxy`, `no_proxy`

We will close this issue or ask you to create a pull-request if it's something we're not actively planning to work on.

barreyra:
Fixes #3500

-----

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -88,8 +88,8 @@ then
   PATH="/usr/bin:/bin:/usr/sbin:/sbin"
 
   FILTERED_ENV=()
-  for VAR in HOME SHELL PATH TERM LOGNAME USER CI TRAVIS SUDO_ASKPASS
-             http_proxy HTTPS_PROXY FTP_PROXY \
+  for VAR in HOME SHELL PATH TERM LOGNAME USER CI TRAVIS SUDO_ASKPASS \
+             http_proxy https_proxy ftp_proxy HTTPS_PROXY FTP_PROXY \
              "${!HOMEBREW_@}" "${!TRAVIS_@}" "${!JENKINS_@}"
   do
     # Skip if variable value is empty.
--- a/bin/brew
+++ b/bin/brew
@@ -88,7 +88,8 @@ then
   PATH="/usr/bin:/bin:/usr/sbin:/sbin"
 
   FILTERED_ENV=()
-  for VAR in HOME SHELL PATH TERM LOGNAME USER CI TRAVIS http_proxy \
+  for VAR in HOME SHELL PATH TERM LOGNAME USER CI TRAVIS \
+             http_proxy https_proxy ftp_proxy no_proxy \
              "${!HOMEBREW_@}" "${!TRAVIS_@}" "${!JENKINS_@}"
   do
     # Skip if variable value is empty.


bin/brew
```diff
@@ -89,6 +89,7 @@ then
 
   FILTERED_ENV=()
   for VAR in HOME SHELL PATH TERM LOGNAME USER CI TRAVIS http_proxy \
+             https_proxy ftp_proxy no_proxy \
```
MikeMcQuaid: What is `no_proxy` used for? Can you move `http_proxy` from the previous line, thanks!

barreyra: `no_proxy` lets you list internal domains that should not use the proxy server. Patch updated.

MikeMcQuaid: Why would you use this with Homebrew?

barreyra: What if somebody had some private casks in their intranet, for instance?

MikeMcQuaid: @barreyra Which of these variables do you personally need for Homebrew to function? I don't want to whitelist everything that could be useful but just things that individuals need.

barreyra: I don't personally need no_proxy, but I think all others are likely needed. It just depends on what protocol is required to retrieve a given dependency (http/https/ftp).

MikeMcQuaid: @barreyra I'd suggest just going for `HTTPS_PROXY` (https://github.com/Homebrew/brew/issues/3500) and, if you personally use it, `FTP_PROXY` for now (both uppercase). If you also use `ALL_PROXY` or `NO_PROXY` they can be added too (but only if you personally need them).

barreyra: So *not* setting the http proxy?

What's the difference between the lower and uppercase versions of the variables? I've always used the lowercase versions.

barreyra: Answering my own question, from the curl manpage:

>The  environment  variables  can  be  specified  in lower case or upper case. The lower case version has precedence. http_proxy is an exception as it is only available in lower case.

Why do you prefer the uppercase version?

MikeMcQuaid: > Why do you prefer the uppercase version?

That's the version that's documented in the man page if you're copy-pasting.

barreyra: I'm fine going with this for now. I had added no_proxy because it was one of the asks in #3500

MikeMcQuaid:
Thanks for your first contribution to Homebrew, @barreyra! Without people like you submitting PRs we couldn't run this project. You rock!

barreyra:
Thanks, but it's not my first commit, I added libbluray support to ffmpeg back in the day :)

https://github.com/Homebrew/homebrew-core/commit/e451b7f62df820e39d997aaeea3da7ba59af46ba
https://github.com/Homebrew/homebrew-core/commit/d12c637c38a0bfd8c26ce2aaa38a054a3712224a

MikeMcQuaid:
First commit to Homebrew/brew then 😁

PR #3507, opened 2017-11-29
Hardware::CPU: Implement OS-agnostic methods

sjackman:

### DIFF
--- a/Library/Homebrew/extend/os/linux/hardware/cpu.rb
+++ b/Library/Homebrew/extend/os/linux/hardware/cpu.rb
@@ -1,26 +1,13 @@
 module Hardware
   class CPU
     class << self
-      def universal_archs
-        [].extend ArchitectureListExtension
-      end
-
       def cpuinfo
         @cpuinfo ||= File.read("/proc/cpuinfo")
       end
 
-      def type
-        @type ||= if cpuinfo =~ /Intel|AMD/
-          :intel
-        elsif cpuinfo =~ /ARM|Marvell/
-          :arm
-        else
-          :dunno
-        end
-      end
-
       def family
         return :arm if arm?
+        return :ppc if ppc?
         return :dunno unless intel?
         # See https://software.intel.com/en-us/articles/intel-architecture-and-processor-identification-with-cpuid-model-and-family-numbers
         cpu_family = cpuinfo[/^cpu family\s*: ([0-9]+)/, 1].to_i
@@ -70,12 +57,9 @@ module Hardware
         end
       end
 
-      def cores
-        cpuinfo.scan(/^processor/).size
-      end
-
       def flags
-        @flags ||= cpuinfo[/^(flags|Features).*/, 0].split
+        @flags ||= cpuinfo[/^(flags|Features).*/, 0]&.split
+        @flags ||= []
       end
 
       # Compatibility with Mac method, which returns lowercase symbols
@@ -94,12 +78,6 @@ module Hardware
 
       def sse4?
         flags.include? "sse4_1"
-      end
-
-      alias is_64_bit? lm?
-
-      def bits
-        is_64_bit? ? 64 : 32
       end
     end
   end
--- a/Library/Homebrew/hardware.rb
+++ b/Library/Homebrew/hardware.rb
@@ -19,16 +19,48 @@ module Hardware
       end
 
       def arch_32_bit
-        :i386
+        if arm?
+          :arm
+        elsif intel?
+          :i386
+        elsif ppc?
+          :ppc32
+        else
+          :dunno
+        end
       end
 
       def arch_64_bit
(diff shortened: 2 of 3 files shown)


Library/Homebrew/extend/os/linux/hardware/cpu.rb
```diff
@@ -70,12 +57,8 @@ def family
         end
       end
 
-      def cores
-        cpuinfo.scan(/^processor/).size
-      end
-
       def flags
-        @flags ||= cpuinfo[/^(flags|Features).*/, 0].split
+        @flags ||= cpuinfo[/^(flags|Features).*/, 0]&.split || []
```
MikeMcQuaid: Stick the `||= []` on a new line.

Library/Homebrew/hardware.rb
```diff
@@ -39,14 +71,17 @@ def family
       end
 
       def cores
-        1
+        return @cores if @cores
+        @cores = Utils.popen_read("getconf", "_NPROCESSORS_ONLN").chomp.to_i
```
MikeMcQuaid: This is new code, no?

Library/Homebrew/hardware.rb
```diff
@@ -39,14 +71,17 @@ def family
...
+        @cores
       end
 
       def bits
-        case RUBY_PLATFORM
-        when /x86_64/, /ppc64/ then 64
-        when /i\d86/, /ppc/ then 32
-        end
+        return @bits if @bits
+        @bits = Utils.popen_read("getconf", "LONG_BIT").chomp.to_i
```
MikeMcQuaid: This is new code, no?

Library/Homebrew/hardware.rb
sjackman: Yes

Library/Homebrew/hardware.rb
sjackman: New to Homebrew, yes.

Library/Homebrew/extend/os/linux/hardware/cpu.rb
sjackman: Fixed.

sjackman:
I've reworded the commit log message to the more accurate:
```
Hardware::CPU: Implement OS-agnostic methods
```

Library/Homebrew/hardware.rb
MikeMcQuaid: Can the previous approach of using `RUBY_PLATFORM` be used here instead and e.g. fixed to work on ARM instead?

Library/Homebrew/hardware.rb
```diff
@@ -39,14 +71,17 @@ def family
...
+        raise ErrorDuringExecution.new("getconf", ["LONG_BIT"]) unless $CHILD_STATUS.success?
```
MikeMcQuaid: Feels weird that this behaves differently to `cores` above; why should one raise an exception but not the other?

Library/Homebrew/hardware.rb
sjackman: `getconf LONG_BIT` works on both macOS and Linux, and is a POSIX standard. Using it would be my preference. If not, yes, I can adapt the previous approach for ARM.

Library/Homebrew/hardware.rb
sjackman: With `cores` we have a reasonable default value that we can use, namely `1`. With `bits`, I'm not sure what a reasonable default value should be. I don't expect `getconf LONG_BIT` to fail though. Every system I've encountered has it. I was just being paranoid.

sjackman:
I've modified `bits` to use `RUBY_PLATFORM` first and fall back to `getconf LONG_BIT` otherwise.

Library/Homebrew/hardware.rb
```diff
@@ -39,13 +71,17 @@ def family
...
       end
 
       def bits
-        case RUBY_PLATFORM
-        when /x86_64/, /ppc64/ then 64
-        when /i\d86/, /ppc/ then 32
+        @bits ||= case RUBY_PLATFORM
+        when /x86_64/, /ppc64/, /aarch64/, /arm64/ then 64
+        when /i\d86/, /ppc/, /arm/ then 32
+        else Utils.popen_read("getconf", "LONG_BIT").chomp.to_i
```
MikeMcQuaid: Let's avoid the `else` here unless you have a platform that you know it will run on. Alternatively, `else 64` because we default to 64-bit everywhere.

sjackman: Done.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike! We're getting there. 🏁

PR #3515, opened 2017-12-01
Utils.popen_read: Nonexistent program should fail

sjackman:

-----

Ensure that `$CHILD_STATUS.success?` is false when the command does not exist.

### DIFF
--- a/Library/Homebrew/utils/popen.rb
+++ b/Library/Homebrew/utils/popen.rb
@@ -14,7 +14,15 @@ module Utils
         yield pipe
       else
         options[:err] ||= :close unless ENV["HOMEBREW_STDERR"]
-        exec(*args, options)
+        begin
+          exec(*args, options)
+        rescue Errno::ENOENT
+          $stderr.puts "brew: command not found: #{args[0]}" unless options[:err] == :close
+          exit! 127
+        rescue SystemCallError
+          $stderr.puts "brew: exec failed: #{args[0]}" unless options[:err] == :close
+          exit! 1
+        end
       end
     end
   end
--- a/Library/Homebrew/test/utils/popen_spec.rb
+++ b/Library/Homebrew/test/utils/popen_spec.rb
@@ -15,6 +15,12 @@ describe Utils do
       ).to eq("success")
       expect($CHILD_STATUS).to be_a_success
     end
+
+    it "fails when the command does not exist" do
+      expect(subject.popen_read("./nonexistent", err: :out))
+        .to eq("brew: command not found: ./nonexistent\n")
+      expect($CHILD_STATUS).to be_a_failure
+    end
   end
 
   describe "::popen_write" do


sjackman:
Currently `$CHILD_STATUS.success?` is `true` after executing a non-existant program.
```
❯❯❯ brew irb <<<'Utils.popen_read("nonexistant"); $CHILD_STATUS.success?'
Switch to inspect mode.
Utils.popen_read("nonexistant"); $CHILD_STATUS.success?
true
```

With this patch, it's false, as expected.
```
❯❯❯ brew irb <<<'Utils.popen_read("nonexistant"); $CHILD_STATUS.success?'
Switch to inspect mode.
Utils.popen_read("nonexistant"); $CHILD_STATUS.success?
false
```

sjackman:
I tried to write a test, but ran into some trouble. It looks as though the test harness is catching the non-existant file before `Utils.popen_read` does.
```ruby
❯❯❯ brew tests --only=utils/popen                         popen
Failures:

  1) Utils::popen_read fails when the command does not exist
     Failure/Error:
       Dir.foreach(@path) {|e|
         next if e == '.' || e == '..'
         if with_directory
           result << self.class.new(File.join(@path, e))
         else
           result << self.class.new(e)
         end
       }
     
     Errno::ENOENT:
       No such file or directory @ dir_initialize - /tmp/homebrew-tests-20171201-56677-bq281e/prefix/Library/Taps/homebrew/homebrew-core/Formula
     # ./test/spec_helper.rb:115:in `map'
     # ./test/spec_helper.rb:115:in `block (2 levels) in <top (required)>'
```

MikeMcQuaid:
> It looks as though the test harness is catching the non-existant file before Utils.popen_read does.

That doesn't seem to be the exception here because that's not the directory it's trying to create. You should be able to catch `Errno::ENOENT` and inspect it if that's the problem.

Library/Homebrew/utils/popen.rb
```diff
@@ -14,7 +14,12 @@ def self.popen(args, mode, options = {})
         yield pipe
       else
         options[:err] ||= :close unless ENV["HOMEBREW_STDERR"]
-        exec(*args, options)
+        begin
+          exec(*args, options)
+        rescue SystemCallError
+          $stderr.puts "brew: command not found: #{args[0]}" unless options[:err] == :close
```
MikeMcQuaid: This message doesn't seem to match the exception; presumably there are other `SystemCallError`s?

ilovezfs:
s/existant/existent/

Library/Homebrew/utils/popen.rb
sjackman: > You should be able to catch Errno::ENOENT and inspect it if that's the problem.

Thanks for the tip. Fixed.

sjackman:
`exit` should have been `exit!`. `exit` incorrectly invokes the `at_exit` handlers. The child process should not invoke the `at_exit` handlers. That should be left to the parent process.

MikeMcQuaid:
`brew style` is 🔴

MikeMcQuaid:
Thanks again @sjackman!

sjackman:

PR #3516, opened 2017-12-01
config: Print the linked_version [Linux]

sjackman:

-----

Otherwise when the keg was installed but out-of-date, it would display "N/A".


MikeMcQuaid:
Does this work for stuff that's keg-only?

sjackman:
Good question. I had to look it up. No, it doesn't. I do want to print the linked version. This function is used only in this file, and only with non-keg-only formulae. I've renamed this function to `formula_linked_version` to make it clear.

MikeMcQuaid:
Thanks gain @sjackman!

sjackman:

PR #3517, opened 2017-12-02
Implement linkage for Linux

sjackman:
extend/pathname: Add os/linux/elf.rb

### DIFF
--- a/Library/Homebrew/os/linux/elf.rb
+++ b/Library/Homebrew/os/linux/elf.rb
@@ -1,0 +1,160 @@
+module ELFShim
+  # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+  MAGIC_NUMBER_OFFSET = 0
+  MAGIC_NUMBER_ASCII = "\x7fELF".freeze
+
+  OS_ABI_OFFSET = 0x07
+  OS_ABI_SYSTEM_V = 0
+  OS_ABI_LINUX = 3
+
+  TYPE_OFFSET = 0x10
+  TYPE_EXECUTABLE = 2
+  TYPE_SHARED = 3
+
+  ARCHITECTURE_OFFSET = 0x12
+  ARCHITECTURE_I386 = 0x3
+  ARCHITECTURE_POWERPC = 0x14
+  ARCHITECTURE_ARM = 0x28
+  ARCHITECTURE_X86_64 = 0x62
+  ARCHITECTURE_AARCH64 = 0xB7
+
+  def read_uint8(offset)
+    read(1, offset).unpack("C").first
+  end
+
+  def read_uint16(offset)
+    read(2, offset).unpack("v").first
+  end
+
+  def elf?
+    return @elf if defined? @elf
+    return @elf = false unless read(MAGIC_NUMBER_ASCII.size, MAGIC_NUMBER_OFFSET) == MAGIC_NUMBER_ASCII
+
+    # Check that this ELF file is for Linux or System V.
+    # OS_ABI is often set to 0 (System V), regardless of the target platform.
+    @elf = [OS_ABI_LINUX, OS_ABI_SYSTEM_V].include? read_uint8(OS_ABI_OFFSET)
+  end
+
+  def arch
+    return :dunno unless elf?
+
+    @arch ||= case read_uint16(ARCHITECTURE_OFFSET)
+    when ARCHITECTURE_I386 then :i386
+    when ARCHITECTURE_X86_64 then :x86_64
+    when ARCHITECTURE_POWERPC then :powerpc
+    when ARCHITECTURE_ARM then :arm
+    when ARCHITECTURE_AARCH64 then :arm64
+    else :dunno
+    end
+  end
+
+  def elf_type
+    return :dunno unless elf?
+
+    @elf_type ||= case read_uint16(TYPE_OFFSET)
+    when TYPE_EXECUTABLE then :executable
+    when TYPE_SHARED then :dylib
+    else :dunno
+    end
+  end
+
+  def dylib?
+    elf_type == :dylib
+  end
+
+  def binary_executable?
+    elf_type == :executable
+  end
+
+  def dynamic_elf?
+    return @dynamic_elf if defined? @dynamic_elf
+
+    if which "readelf"
+      Utils.popen_read("readelf", "-l", to_path).include?(" DYNAMIC ")
+    elsif which "file"
+      !Utils.popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil?
+    else
+      raise "Please install either readelf (from binutils) or file."
+    end
+  end
+
+  class Metadata
+    attr_reader :path, :dylib_id, :dylibs
+
+    def initialize(path)
+      @path = path
+      @dylibs = []
+      @dylib_id, needed = needed_libraries path
+      return if needed.empty?
+
+      ldd = DevelopmentTools.locate "ldd"
+      ldd_output = Utils.popen_read(ldd, path.expand_path.to_s).split("\n")
+      return unless $CHILD_STATUS.success?
+
+      ldd_paths = ldd_output.map do |line|
+        match = line.match(/\t.+ => (.+) \(.+\)|\t(.+) => not found/)
+        next unless match
+        match.captures.compact.first
+      end.compact
+      @dylibs = ldd_paths.select do |ldd_path|
+        next true unless ldd_path.start_with? "/"
+        needed.include? File.basename(ldd_path)
+      end
+    end
+
+    private
+
+    def needed_libraries(path)
+      if DevelopmentTools.locate "readelf"
+        needed_libraries_using_readelf path
+      elsif DevelopmentTools.locate "patchelf"
+        needed_libraries_using_patchelf path
+      else
+        raise "patchelf must be installed: brew install patchelf"
+      end
+    end
+
+    def needed_libraries_using_patchelf(path)
+      patchelf = DevelopmentTools.locate "patchelf"
+      if path.dylib?
+        command = [patchelf, "--print-soname", path.expand_path.to_s]
+        soname = Utils.popen_read(*command).chomp
+        raise ErrorDuringExecution, command unless $CHILD_STATUS.success?
+      end
+      command = [patchelf, "--print-needed", path.expand_path.to_s]
+      needed = Utils.popen_read(*command).split("\n")
+      raise ErrorDuringExecution, command unless $CHILD_STATUS.success?
+      [soname, needed]
+    end
+
+    def needed_libraries_using_readelf(path)
+      soname = nil
+      needed = []
+      command = ["readelf", "-d", path.expand_path.to_s]
+      lines = Utils.popen_read(*command).split("\n")
+      raise ErrorDuringExecution, command unless $CHILD_STATUS.success?
+      lines.each do |s|
+        filename = s[/\[(.*)\]/, 1]
+        next if filename.nil?
+        if s.include? "(SONAME)"
+          soname = filename
+        elsif s.include? "(NEEDED)"
+          needed << filename
+        end
+      end
+      [soname, needed]
+    end
+  end
+
+  def metadata
+    @metadata ||= Metadata.new(self)
+  end
+
+  def dylib_id
+    metadata.dylib_id
+  end
+
+  def dynamically_linked_libraries(*)
+    metadata.dylibs
+  end
+end
--- a/Library/Homebrew/extend/os/linux/extend/pathname.rb
+++ b/Library/Homebrew/extend/os/linux/extend/pathname.rb
@@ -1,19 +1,5 @@
-class Pathname
-  # @private
-  def elf?
-    # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
-    read(4) == "\x7fELF"
-  end
+require "os/linux/elf"
 
-  # @private
-  def dynamic_elf?
-    if which "readelf"
-      popen_read("readelf", "-l", to_path).include?(" DYNAMIC ")
-    elsif which "file"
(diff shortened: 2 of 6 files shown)


Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
+module ELFShim
+  def elf?
+    return @elf if defined? @elf
+    # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+    return @elf = false unless read(4) == "\x7fELF"
+    # Check that the ELF file is for Linux.
+    @elf = case read(1, 7).unpack("C")[0]
+    when 0, 3 then true
```
MikeMcQuaid: The above two lines aren't particularly obvious in what they are doing. Can you add a comment and then we can hopefully figure out how to make them more readable?

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
+module ELFShim
+  def elf?
+    return @elf if defined? @elf
+    # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+    return @elf = false unless read(4) == "\x7fELF"
```
MikeMcQuaid: I'd say pretty much every number in this file should be a constant explaining what it means otherwise they are a bit "magic".

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
...
+  end
+
+  def dynamic_elf?
+    return @dynamic_elf if defined? @dynamic_elf
+    if which "readelf"
+      Utils.popen_read("readelf", "-l", to_path).include?(" DYNAMIC ")
+    elsif which "file"
+      !Utils.popen_read("file", "-L", "-b", to_path)[/dynamic|shared/].nil?
+    else
+      raise StandardError, "Neither `readelf` nor `file` is available "\
```
MikeMcQuaid: Is `StandardError` not the default for `raise? I can't remember. Also, make this message no more than 80 characters or use `<<~EOS`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
...
+        "to determine whether '#{self}' is dynamically or statically linked."
+    end
+  end
+
+  class Metadata
+    LDD_RX = /\t.* => (.*) \(.*\)|\t(.*) => not found/
```
MikeMcQuaid: Don't know what `LDD_RX` means

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    attr_reader :path, :dylib_id, :dylibs
+
+    def initialize(path)
+      @path = path
+      @dylibs = []
+      @dylib_id, needed = needed_libraries path
+      return if needed.empty?
+      # ldd requires that the file be executable, and all ELF files should be executable.
+      FileUtils.chmod "+x", path unless path.executable?
```
MikeMcQuaid: A `Metadata.new` shouldn't be modifying files.

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
...
+    def needed_libraries_using_readelf(path)
+      soname = nil
+      needed = []
+      command = ["readelf", "-d", path.expand_path.to_s]
+      lines = Utils.popen_read(*command).split("\n")
+      raise ErrorDuringExecution, command unless $CHILD_STATUS.success?
+      lines.each do |s|
+        case s
+        when /\(SONAME\)/
+          soname = s[/\[(.*)\]/, 1]
```
MikeMcQuaid: This regex is repeated so consider making it a constant

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    def needed_libraries_using_readelf(path)
+      soname = nil
+      needed = []
+      command = ["readelf", "-d", path.expand_path.to_s]
+      lines = Utils.popen_read(*command).split("\n")
+      raise ErrorDuringExecution, command unless $CHILD_STATUS.success?
+      lines.each do |s|
+        case s
+        when /\(SONAME\)/
```
MikeMcQuaid: This doesn't need to be a regex.

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
...
+          soname = s[/\[(.*)\]/, 1]
+        when /\(NEEDED\)/
```
MikeMcQuaid: This doesn't need to be a regex.

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  def metadata
+    @metadata ||= Metadata.new(self)
+  end
+
+  def dylib_id
+    metadata.dylib_id
+  end
+
+  def dynamically_linked_libraries(**_options)
```
MikeMcQuaid: Can probably just make this `(*)`

sjackman:
Thanks for the feedback, Mike. I've addressed your comments.

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
+module ELFShim
+  # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+  MAGIC_OFFSET = 0
```
MikeMcQuaid: `MAGIC_NUMBER_OFFSET`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  MAGIC_SIZE = 4
```
MikeMcQuaid: Use `MAGIC_NUMBER.size`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  MAGIC_NUMBER = "\x7fELF".freeze
+
+  OSABI_OFFSET = 0x07
```
MikeMcQuaid: `OS_ABI_OFFSET`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  OSABI_SIZE = 1
```
MikeMcQuaid: Use `OSABI_LINUX.to_s.size`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  OSABI_SYSTEM_V = 0
+  OSABI_LINUX = 3
+
+  TYPE_OFFSET = 0x10
+  TYPE_SIZE = 2
```
MikeMcQuaid: Use `TYPE_EXECUTABLE.to_s.size`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  TYPE_EXECUTABLE = 2
+  TYPE_SHARED = 3
+
+  MACHINE_OFFSET = 0x12
+  MACHINE_SIZE = 2
```
MikeMcQuaid: `MACHINE_X86_64.to_s.size`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  OSABI_SIZE = 1
+  OSABI_SYSTEM_V = 0
+  OSABI_LINUX = 3
+
+  TYPE_OFFSET = 0x10
+  TYPE_SIZE = 2
+  TYPE_EXECUTABLE = 2
+  TYPE_SHARED = 3
+
+  MACHINE_OFFSET = 0x12
```
MikeMcQuaid: `ARCHITECTURE_OFFSET`?

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
+module ELFShim
+  # See: https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#File_header
+  MAGIC_OFFSET = 0
+  MAGIC_SIZE = 4
+  MAGIC_NUMBER = "\x7fELF".freeze
```
MikeMcQuaid: This isn't a number. `MAGIC_NUMBER_ASCII`?

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  MACHINE_ARM = 0x28
+  MACHINE_X86_64 = 0x62
+  MACHINE_AARCH64 = 0xB7
+
+  def elf?
+    return @elf if defined? @elf
+    return @elf = false unless read(MAGIC_SIZE, MAGIC_OFFSET) == MAGIC_NUMBER
+    # Check that this ELF file is for Linux or System V.
+    # OSABI is often set to 0 (System V), regardless of the target platform.
+    osabi = read(OSABI_SIZE, OSABI_OFFSET).unpack("C")[0]
```
MikeMcQuaid: `.first` rather than `[0]`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+    @elf = osabi == OSABI_LINUX || osabi == OSABI_SYSTEM_V
+  end
+
+  def arch
+    return :dunno unless elf?
+    @arch ||= case read(MACHINE_SIZE, MACHINE_OFFSET).unpack("v")[0]
```
MikeMcQuaid: `.first` rather than `[0]`

Library/Homebrew/os/linux/elf.rb
MikeMcQuaid: `os_abi`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  MACHINE_X86_64 = 0x62
+  MACHINE_AARCH64 = 0xB7
+
+  def elf?
+    return @elf if defined? @elf
+    return @elf = false unless read(MAGIC_SIZE, MAGIC_OFFSET) == MAGIC_NUMBER
+    # Check that this ELF file is for Linux or System V.
+    # OSABI is often set to 0 (System V), regardless of the target platform.
+    osabi = read(OSABI_SIZE, OSABI_OFFSET).unpack("C")[0]
+    @elf = osabi == OSABI_LINUX || osabi == OSABI_SYSTEM_V
```
MikeMcQuaid: `[OSABI_SYSTEM_V, OSABI_LINUX].include?(osabi)`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+    when MACHINE_POWERPC then :powerpc
+    when MACHINE_ARM then :arm
+    when MACHINE_AARCH64 then :arm64
+    else :dunno
+    end
+  end
+
+  def elf_type
+    return :dunno unless elf?
+    @elf_type ||= case read(TYPE_SIZE, TYPE_OFFSET).unpack("v")[0]
```
MikeMcQuaid: `.first` rather than `[0]`

Library/Homebrew/os/linux/elf.rb
```diff
@@ -0,0 +1,151 @@
...
+  end
+
+  class Metadata
+    attr_reader :path, :dylib_id, :dylibs
+
+    def initialize(path)
+      @path = path
+      @dylibs = []
+      @dylib_id, needed = needed_libraries path
+      return if needed.empty?
```
MikeMcQuaid: Stick a newline after each `return`; makes this easier to follow/scan (and consider doing this elsewhere in the file).

Library/Homebrew/os/linux/elf.rb
sjackman: These are not equivalent.
`TYPE_EXECUTABLE.to_s.size` ≠ `TYPE_SIZE`
`TYPE_EXECUTABLE.to_s.size` = 1
`TYPE_SIZE` = 2

sjackman: I've added a `read_uint16` helper function.

Library/Homebrew/os/linux/elf.rb
sjackman: I've added a `read_uint8` helper function.

Library/Homebrew/os/linux/elf.rb
sjackman: Done.

sjackman:
Fixed. Thanks again for the feedback, Mike.

MikeMcQuaid:
Thanks again @sjackman!

MikeMcQuaid:
@sjackman Merging this because there's been a lot of back and forth but ideally we'd get some tests to cover a bit more of the diff, here. Can do that in follow-up PR.

sjackman:
It looks as though Codecov doesn't count lines-of-code as covered that are covered by tests that run only on Linux. Is that the case?
https://codecov.io/gh/Homebrew/brew/compare/0ad42ebbcb735422612a7c6d4edc5ad88044dd4b...d79c5ade1abaad0fb1ea88f2b949def8dcfddb81/diff#diff-TGlicmFyeS9Ib21lYnJldy9vcy9saW51eC9lbGYucmI=

sjackman:

MikeMcQuaid:
> It looks as though Codecov doesn't count lines-of-code as covered that are covered by tests that run only on Linux. Is that the case?

@sjackman Correct. If we can figure out a way to upload and combine the coverage that would be good. @reitermarkus looked at this before I think; any thoughts Markus?

PR #3526, opened 2017-12-04
Enable quarantining of Cask apps

amyspark:
-----

This pull request fixes caskroom/homebrew-cask#22388 by quarantining apps that explicitly request it.
To implement this, I picked up reitermarkus's work and integrated it into Cask's Utils.

I think this still need a lot of polish; thanks for your comments!

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/download_strategy.rb
+++ b/Library/Homebrew/cask/lib/hbc/download_strategy.rb
@@ -12,8 +12,8 @@ module Hbc
     attr_reader :cask, :name, :url, :uri_object, :version
 
     APP_NAME = "Homebrew-Cask".b.freeze
-    TYPE = format("%04d", (LSQuarantine::Database::TYPE_NUMBERS[:web_download]).freeze)
-    STATUS = format("%04d",  (81).freeze)
+    TYPE = format("%04d", (LSQuarantine::Database::TYPE_NUMBERS[:web_download])).freeze
+    STATUS = format("%04d", LSQuarantine::ExtendedAttribute::QUARANTINED_FILE).freeze
 
     def initialize(cask, command: SystemCommand)
       @cask       = cask
@@ -108,7 +108,9 @@ module Hbc
         begin
           LockFile.new(temporary_path.basename).with_lock do
             _fetch
-            LSQuarantine.add(temporary_path, type: TYPE, app_name: APP_NAME, title: @name, url: @url, status: STATUS) unless @cask.quarantine
+            if temporary_path.exist? && @cask.quarantine
+              LSQuarantine.add(temporary_path, type: TYPE, app_name: APP_NAME, title: @name, url: cask.homepage, status: STATUS) unless cask.quarantine
+            end
           end
         rescue ErrorDuringExecution
           # 33 == range not supported
--- a/Library/Homebrew/cask/lib/hbc/download_strategy.rb
+++ b/Library/Homebrew/cask/lib/hbc/download_strategy.rb
@@ -12,7 +12,8 @@ module Hbc
     attr_reader :cask, :name, :url, :uri_object, :version
 
     APP_NAME = "Homebrew-Cask".b.freeze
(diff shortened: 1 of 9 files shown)


Library/Homebrew/cask/lib/hbc/download_strategy.rb
```diff
@@ -10,6 +11,10 @@ module Hbc
   class AbstractDownloadStrategy
     attr_reader :cask, :name, :url, :uri_object, :version
 
+    APP_NAME = "Homebrew-Cask".b.freeze
+    TYPE = format("%04d", (LSQuarantine::Database::TYPE_NUMBERS[:web_download]).freeze)
+    STATUS = format("%04d",  (81).freeze)
```
amyspark: I got this constant empirically (downloading lots of apps via Firefox and Safari and seeing which value corresponded to an unopened package).

Library/Homebrew/cask/lib/hbc/utils/ls_quarantine/extended_attributes.rb
```diff
@@ -0,0 +1,67 @@
...
+    def clear
+      command("-c")[1].success?
+    end
+
+    private
+
+    XATTR_EXECUTABLE = Open3.capture2("which", "xattr")[0].chomp
+
+    def command(*args)
+      raise StandardError, "xattr not installed." if XATTR_EXECUTABLE.empty?
```
commitay: Not sure if this or `which xattr` is needed as we use `/usr/bin/xattr` for [add_altname_metadata](https://github.com/Homebrew/brew/blob/f3ec40d696b421e89f462d32f6f30a81f60d3e73/Library/Homebrew/cask/lib/hbc/artifact/relocated.rb#L56-L73)

reitermarkus:
> I think this still need a lot of polish

I think the most polishing would be needed in the code I wrote, so it might be a good idea to clean this up at https://github.com/reitermarkus/quarantine before incorporating it in Homebrew.

Especially magic numbers such as `81` above should be documented in the code, or included as a constant. If i remember correctly, those numbers were different on different versions of macOS, that's where I gave up on this.

> by quarantining apps that explicitly request it.

I don't think this should apply only to specific Casks.

reitermarkus:
As I already said, the numbers seem to be different on different versions of macOS.

Regarding `81`, can you check if it corresponds to https://github.com/Homebrew/brew/pull/3526/files#diff-105df28dac19bb98ad26194a4cfd6458R10?

Does it change if you cancel the prompt, and does it change if you open it?

amyspark:
@reitermarkus yes, it does change if one cancels the prompt or chooses to open it.
From my notes:

DMG as downloaded from Safari: `0x0081`

App bundle (e.g. `Firefox.app`)
--
Extracted from DMG, unopened: `0x0181`
Extracted from DMG, opened: `0x01E1`
Extracted from DMG, cancelled: `0x01A1`
Extracted from ZIP, unopened: `0x0081`
Extracted from ZIP, opened: `0x01E1` => `100`, `E0`
Extracted from ZIP, cancelled: `0x00A1`=> `A0`

`181` =` 1 1000 0001` => `0x00` is the unopened flag
`1a1` = `1 1010 0001`  => `0xA0` cancelled
`1e1` = `1 1110 0001` => `0xE0` opened

App executable (`Firefox.app/Contents/MacOS/...`)
--

manually set, unopened: `0001`
extracted from app bundle, unopened: `0x0181` = `1 1000 0001`
cancelled: no changes
opened: `0x1c1` = `1 1100 0001` => `0x40` (as per the VirusTotal docs!)

It seems that the minimum requested bit for Gatekeeper to work is `0x1`. When I set it to `0x0`, opening the app corrupted (in my tests) the security policy database.

vitorgalvao:
How did I miss this the first time?!

> > by quarantining apps that explicitly request it.
>
> I don't think this should apply only to specific Casks.

Agreed. Quarantining should apply to *all* casks.

> I think the most polishing would be needed in the code I wrote, so it might be a good idea to clean this up at https://github.com/reitermarkus/quarantine before incorporating it in Homebrew.

Again, agreed. If your code could be improved into a CLI tool that anyone could use, that’d be a boon to any project.

amyspark:
@vitorgalvao @reitermarkus Thanks for your comments. Do you need me to perform any further changes?

PR #3530, opened 2017-12-05
config: Print host glibc version [Linux]

sjackman:

-----

### DIFF
--- a/Library/Homebrew/extend/os/linux/system_config.rb
+++ b/Library/Homebrew/extend/os/linux/system_config.rb
@@ -1,4 +1,5 @@
 require "formula"
+require "os/linux/glibc"
 
 class SystemConfig
   class << self
@@ -12,6 +13,12 @@ class SystemConfig
       else
         "N/A"
       end
+    end
+
+    def host_glibc_version
+      version = OS::Linux::Glibc.system_version
+      return "N/A" if version.null?
+      version
     end
 
     def host_gcc_version
@@ -31,6 +38,7 @@ class SystemConfig
       dump_generic_verbose_config(out)
       out.puts "Kernel: #{`uname -mors`.chomp}"
       out.puts "OS: #{host_os_version}"
+      out.puts "Host glibc: #{host_glibc_version}"
       out.puts "/usr/bin/gcc: #{host_gcc_version}"
       ["glibc", "gcc", "xorg"].each do |f|
         out.puts "#{f}: #{formula_linked_version f}"
--- a/Library/Homebrew/os/linux/glibc.rb
+++ b/Library/Homebrew/os/linux/glibc.rb
@@ -1,0 +1,14 @@
+module OS
+  module Linux
+    module Glibc
+      module_function
+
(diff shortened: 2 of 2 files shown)


Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -14,6 +14,18 @@ def host_os_version
       end
     end
 
+    def maybe_na(version)
+      if version.nil? || version.null?
```
MikeMcQuaid: When is it nil and when is it null? I'd say it'd be better to return a consistent value.

Library/Homebrew/os.rb
```diff
@@ -21,6 +21,7 @@ def self.linux?
     end
     PATH_OPEN = "/usr/bin/open".freeze
   elsif OS.linux?
+    require "os/linux"
```
MikeMcQuaid: Does this need to be required globally? I wish we didn't do that for `os/mac`

Library/Homebrew/os/linux/glibc.rb
```diff
@@ -0,0 +1,14 @@
+module OS
+  module Linux
+    module Glibc
```
MikeMcQuaid: Could this just be required from `system_config` or even just be in that file itself if it's only used there?

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: It's always `null?`. I was being unnecessarily paranoid.

Library/Homebrew/os.rb
sjackman: No, it doesn't. I was just following the same pattern. Fixed.

Library/Homebrew/os/linux/glibc.rb
sjackman: Yes, it can be required by `system_config`. Done.
This method will be used later by the bottle-pouring logic to ensure that a new enough `glibc` is installed to use bottles.

Library/Homebrew/extend/os/linux/system_config.rb
```diff
@@ -14,6 +15,18 @@ def host_os_version
       end
     end
 
+    def maybe_na(version)
```
MikeMcQuaid: Is this method going to be used in other PRs? Otherwise can you just roll it into `host_glibc_version` with a `return "N/A" if version.null?`

Library/Homebrew/os/linux/glibc.rb
MikeMcQuaid: Cool, that's fine.

Library/Homebrew/extend/os/linux/system_config.rb
sjackman: Done.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #3586, opened 2017-12-19
formula_installer: prevent endless loop in some cases

iMichka:
This fixes an endless loop, where the xorg requirement
was added to the formulae array, and looped
over again and again. If the requirement is satisfied,
we can bail out earlier.

This fixes the endless loop you got when calling:
brew install python3 --with-tcl-tk

See https://github.com/Linuxbrew/homebrew-core/issues/5218
for the initial bug report. Adding @sjackman as cc.

-----

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -449,11 +449,11 @@ class FormulaInstaller
           Requirement.prune
         elsif req.build? && use_default_formula && req_dependency&.installed?
           Requirement.prune
+        elsif req.satisfied?
+          Requirement.prune
         elsif install_requirement_formula?(req_dependency, req, dependent, install_bottle_for_dependent)
           deps.unshift(req_dependency)
           formulae.unshift(req_dependency.to_formula)
-          Requirement.prune
-        elsif req.satisfied?
           Requirement.prune
         elsif !runtime_requirements.include?(req) && install_bottle_for_dependent
           Requirement.prune


Library/Homebrew/formula_installer.rb
```diff
@@ -449,12 +449,12 @@ def expand_requirements
           Requirement.prune
         elsif req.build? && use_default_formula && req_dependency&.installed?
           Requirement.prune
+        elsif req.satisfied?
```
ilovezfs: I think moving this up will mess up the use of default formula for e.g. `:python` using the python formula

MikeMcQuaid: Agreed.

MikeMcQuaid:
> brew install python3 --with-tcl-tk

Just in Linuxbrew or in Homebrew? Can you find any way of reproducing this bug in Homebrew?

Library/Homebrew/formula_installer.rb
MikeMcQuaid: To be more explicit: yes, it definitely will break this.

iMichka:
The bug can currently only be reproduced in Linuxbrew. But I do believe it could happen in Homebrew too, it is just not visible because you have a slightly different dependency tree.

Just by looking at the code, you see that you can end up in an endless loop, because the while loops over the same array than the one on which unshift is called.

@sjackman proposed that this patch could fix it (https://github.com/Linuxbrew/brew/pull/542), but it didn't. The thing is also that we are still using brew 1.3.9 for the moment, so the overall behaviour may be different.

Let me close this. I'll need to dig more through that part of the code and understand what is really going on.

PR #3588, opened 2017-12-21
brew.sh: Use shims/scm/git for HOMEBREW_VERSION

sjackman:

-----

Move the use of git to determine HOMEBREW_VERSION below the definition of the function git.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -19,14 +19,6 @@ case "$*" in
   --repository|--repo) echo "$HOMEBREW_REPOSITORY"; exit 0 ;;
 esac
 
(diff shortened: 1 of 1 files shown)


alyssais:
Can you describe what the effects of this change are? I think I could guess but don't want to make a wrong assumption.

MikeMcQuaid:
LGTM, thanks again @sjackman!

sjackman:
Before this patch,  the command
```sh
HOMEBREW_VERSION="$(git -C "$HOMEBREW_REPOSITORY" describe --tags --dirty --abbrev=7 2>/dev/null)"
```
is running using the `git` executable found in `PATH`.
After this patch, the command is run using the `git` shell function, which runs `$HOMEBREW_LIBRARY/Homebrew/shims/scm/git`.

PR #3590, opened 2017-12-21
caveats: correct completion/function paths for keg_only

DomT4:

-----

Can do a bunch of different things here but this seems like the simplest place to continue the discussion from https://github.com/Homebrew/homebrew-core/pull/21941.

### DIFF
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -86,11 +86,7 @@ class Caveats
     installed << "completions" if completion_installed
     installed << "functions" if functions_installed
 
-    if f.keg_only?
-      root_dir = f.opt_prefix
-    else
-      root_dir = HOMEBREW_PREFIX
-    end
+    root_dir = f.keg_only? ? f.opt_prefix : HOMEBREW_PREFIX
 
     case shell
     when :bash
--- a/Library/Homebrew/caveats.rb
+++ b/Library/Homebrew/caveats.rb
@@ -86,21 +86,27 @@ class Caveats
     installed << "completions" if completion_installed
     installed << "functions" if functions_installed
 
+    if f.keg_only?
+      root_dir = f.opt_prefix
+    else
+      root_dir = HOMEBREW_PREFIX
+    end
+
     case shell
     when :bash
       <<~EOS
         Bash completion has been installed to:
-          #{HOMEBREW_PREFIX}/etc/bash_completion.d
+          #{root_dir}/etc/bash_completion.d
       EOS
     when :zsh
       <<~EOS
         zsh #{installed.join(" and ")} have been installed to:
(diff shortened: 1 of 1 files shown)


Library/Homebrew/caveats.rb
```diff
@@ -76,6 +76,7 @@ def keg_only_text
 
   def function_completion_caveats(shell)
     return unless keg
+    return if f.keg_only? # https://github.com/Homebrew/homebrew-core/pull/21941
```
MikeMcQuaid: Would be good to make this print the correct paths instead.

alyssais: Would completion work for keg-only formulae though? The commands wouldn't (normally) be available.

DomT4: Above comment is correct. See the linked discussion in the core thread with ILZ. On the phone so will reply properly later from the laptop, but yeah, Alyssa is correct.

MikeMcQuaid: Keg-only doesn't imply unlinked. We should explain that bash completion has been installed and where and, if needed, elaborate further on what's required to make it work.

DomT4: I'm happy to do it, but it will significantly bulk out this section of code given `zsh` and `fish` need covering too, all with slightly different sets of instructions. If you're happy with code bulk though, I'm happy to put the time in.

MikeMcQuaid: I think it would be sufficient to just edit the paths noted below e.g. `#{HOMEBREW_PREFIX}/etc/bash_completion.d` in the keg-only case.

DomT4: Except that introduces a new failure model if people source it without handling the PATH issue at the same time, as outlined in https://github.com/Homebrew/homebrew-core/pull/21941#issuecomment-353217445.

MikeMcQuaid: We are telling people where they are installed, though; I think that's preferable and if we have users who actually get confused by that and end up on our issue tracker: we can deal with that.

DomT4: Okay. Let's start fairly minimal and see if users report confusion/etc, that seems a reasonable approach, I agree.

DomT4:
Rebased & tweaked.

DomT4:
> caveats: tweak root_dir definition style.

Sorry, completely forgot you were cool with ternary operators. I think it was you who objected to them being confusing at some point in the past & after that I stopped trying to sneak them into things 😄.

MikeMcQuaid:
@DomT4 No worries, I didn't care enough and would have merged already but Travis CI was being slow!

MikeMcQuaid:
Thanks again @DomT4!

DomT4:
Thanks Mike.

PR #3591, opened 2017-12-21
gpg: relocate significant logic to requirement

DomT4:

-----

Handles the recent ENV filtering changes. Theoretically. I mean it seems to be working locally, but personally still consider this a bit of a WIP.

CC @ilovezfs re https://github.com/Homebrew/homebrew-core/issues/21903.

Replaces https://github.com/Homebrew/brew/pull/3587 as desired.


DomT4:
Still need to do something about `gpg2_requirement_spec.rb`, I'm aware, as CI is highlighting.

Library/Homebrew/gpg.rb
```diff
@@ -1,26 +1,10 @@
...
-    end
-  end
-
-  def self.executable
-    find_gpg("gpg") || find_gpg("gpg2")
-  end
-
   def self.available?
-    File.executable?(executable.to_s)
+    File.executable?(GPG2Requirement.new.executable.to_s)
```
MikeMcQuaid: I'd rather not have requirements getting called by utility code this way.

Library/Homebrew/requirements/gpg2_requirement.rb
```diff
@@ -1,12 +1,31 @@
...
+      next unless gpg_short_version
+      gpg_version = Version.create(gpg_short_version.to_s)
+      @version = gpg_version
+      gpg_version >= Version.create("2.0")
+    end
+  end
+
+  env do
+    # If Homebrew's GPG is installed, prioritise it.
+    ENV.prepend_path "PATH", Formula["gnupg"].opt_bin
```
MikeMcQuaid: Why is that? Will this not contradict the PATH you're using for `satisfy`?

Library/Homebrew/gpg.rb
```diff
@@ -1,26 +1,10 @@
...
   end
 
-  def self.version
-    @version if available?
+  def self.executable
+    GPG2Requirement.new.executable
```
MikeMcQuaid: This will not work as expected. Fundamentally `Requirement` GPG lookup and install-time lookup need to find GPG differently: the prior needs to search `HOMEBREW_PATH`, the latter `PATH` (based on the `gpg` that was inserted by the requirement). I think this method should just be `which "gpg" || which "gpg2"`.

Library/Homebrew/gpg.rb
DomT4: Alright, I can rework it to not call the requirement, that's not a major issue.

Library/Homebrew/gpg.rb
DomT4: Thanks. To be honest whilst I've kept an eye on Homebrew I'm not as intimately familiar with it as I used to be simply because I don't have the same level of time available as I did once 😅.

This should be okay using `which` because the only usage in core of `Gpg.executable` comes inside `post_install`, but I'll check. The `gpg2` probing is something I dropped from the Requirement side of these changes because I'm no longer all that sure it's necessary.

Library/Homebrew/requirements/gpg2_requirement.rb
DomT4: This is something I might drop. My primary concern is things hardcoding the Cellar path to `gpg`, which would become a pain around updates/revision bumps/etc.

ILZ said `/usr/local/bin` is rejected even if it's where `satisfy` points to? I think? Unless I'm misunderstanding his words there, which is why I inserted the `append` line below.

Library/Homebrew/gpg.rb
MikeMcQuaid: Cool 👍

Library/Homebrew/requirements/gpg2_requirement.rb
MikeMcQuaid: > My primary concern is things hardcoding the Cellar path to gpg

How would they get that?



> ILZ said /usr/local/bin is rejected even if it's where satisfy points to? I think? Unless I'm misunderstanding his words there, which is why I inserted the append line below.

Let's pass on it unless you've got a reproducible case in which it's needed.

Library/Homebrew/gpg.rb
MikeMcQuaid: Cool. It would be good to have `Gpg.executable` so it works consistently in `install`, `post_install` and `test` which should be the case with using `which` as-is.

MikeMcQuaid: > The gpg2 probing is something I dropped from the Requirement side of these changes because I'm no longer all that sure it's necessary.

This suggests to me that we may want to kill this utility method entirely.

Library/Homebrew/requirements/gpg2_requirement.rb
ilovezfs: I was referring to 5de7ec394bf9f24423e0b380b39c6c2801eca81e.

MikeMcQuaid: At that point: why don't we just `depends_on "gpg"`?

DomT4: The point was to not trample all over MacGPG2, but that's _somewhat_ less of an issue now you can permanently `brew unlink` things. It would still be an issue the first time it happened, but. Ultimately it comes down to the question of how broad a use case are you willing to accommodate from a Homebrew point of view.

DomT4: Er, GPGTools even 😅.

DomT4: > How would they get that?

This was something I stumbled upon in testing but can't reproduce now, so 🤷‍♂️.

Library/Homebrew/gpg.rb
DomT4: It seems logical/sensible to me to check we are actually using a 2.x version of GPG still, which can either be enforced via Requirement or simply be always using Homebrew's `gnupg` formula if you don't mind throwing GPGTools support out of the window.

PR #3592, opened 2017-12-21
DevelopmentTools.installed?: Use locate

sjackman:

-----

Use locate rather than which to search for clang or gcc.
locate searches both $HOMEBREW_PREFIX/bin and /usr/bin.

### DIFF
--- a/Library/Homebrew/development_tools.rb
+++ b/Library/Homebrew/development_tools.rb
@@ -16,7 +16,7 @@ class DevelopmentTools
     end
 
     def installed?
-      which("clang") || which("gcc")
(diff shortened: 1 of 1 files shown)


alyssais:
My understanding of this is that it's searching `HOMEBREW_PREFIX/"bin"` again now, which it stopped doing after env filtering was introduced. See https://github.com/Linuxbrew/brew/issues/538.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for the clarification, Alyssa, and thanks for merging, Mike!

PR #3594, opened 2017-12-21
Revert "shims/scm/git: Fix the search for brewed git"

ilovezfs:
Reverts Homebrew/brew#3589

PR #3589, opened 2017-12-21
shims/scm/git: Fix the search for brewed git

sjackman:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew tests` with your changes locally?

-----

Search for brewed git in both locations:
```
$HOMEBREW_PREFIX/Homebrew/Library/Homebrew/shims/scm/../../../../../bin/git
$HOMEBREW_PREFIX/Library/Homebrew/shims/scm/../../../../bin/git
```

It previously searched only the latter location.

PR #3602, opened 2017-12-23
Revert "Revert "shims/scm/git: Fix the search for brewed git""

ilovezfs:
Reverts Homebrew/brew#3594

#3589 should be OK now that https://github.com/Homebrew/homebrew-core/pull/22071 is merged.


ilovezfs:
CC @sjackman

sjackman:
Thanks!

PR #3607, opened 2017-12-29
gpg: use with_homebrew_path

DomT4:

-----

Fixes calling things like `Gpg.available?` in `brew irb`, etc. Or anywhere that uses the protected `PATH`, really.

### DIFF
--- a/Library/Homebrew/gpg.rb
+++ b/Library/Homebrew/gpg.rb
@@ -4,7 +4,7 @@ module Gpg
   module_function
 
   def executable
-    which "gpg"
+    with_homebrew_path { which("gpg") }
   end
 
   def available?


MikeMcQuaid:
This isn't correct behaviour in an `install` method where it should be doing PATH lookup based on a Requirement or Dependency addition.

DomT4:
Sure. I'm open to suggestions that are better than having an audit to bar `Gpg.` calls inside `def install`. It seems like `brew irb` is a perfectly valid place where you'd want to be able to test/use `Gpg.` stuff though and currently you can't, because `brew irb` is PATH-shielded under the new ENV filtering.

ilovezfs:
Can we remove the Gpg module altogether?
```
Josephs-MacBook-Pro:Formula joe$ ag 'Gpg\.'
duplicity.rb
327:    Gpg.test(testpath) do

rpm.rb
52:    gnupg = Gpg.executable || HOMEBREW_PREFIX/"bin/gpg"
```
It would take about 5 minutes to remove its use from formulae.

DomT4:
It's final broad use wasn't exclusively formula-based, although I appreciate you'd have to have a long memory to remember that by now.

ilovezfs:
Can you elaborate? I'm not understanding what function it's serving in `brew` at this point.

DomT4:
https://github.com/Homebrew/brew/pull/786, https://github.com/Homebrew/brew/pull/676, https://github.com/Homebrew/brew/pull/675, https://github.com/Homebrew/brew/pull/674, https://github.com/Homebrew/brew/pull/386.

ilovezfs:
OK, thanks. All of those look self-referential as opposed to anything in use currently.

The only actual, current use of gpg appears to be in casks right now with stuff like
```
gpg "#{url}.asc", key_id: '9032cae4cbfa933a5a2145d5ff97c53f183c045d'
```

I don't know if that depends on the Gpg module or not. CC @reitermarkus

DomT4:
> I don't know if that depends on the Gpg module or not.

It doesn't.

> All of those look self-referential as opposed to anything in use currently.

Yes, this is correct. A lot of the back & forth over the GPG stuff and the resulting stress over effectively turning Homebrew into a job contributed significantly to my decision to walk away as a maintainer, and indeed more or less stop contributing for a while. But the intention wasn't to abandon it, which is why I've kept trimming & tweaking it since.

MikeMcQuaid:
@DomT4 I think it's just worth deprecating at this point I'm afraid as it's not been widely used in formulae. Doing that as part of https://github.com/Homebrew/brew/pull/3612.

DomT4:
Alright.

Issue #1860, opened 2017-01-16
Emit warnings for mistyped options


MikeMcQuaid:
e.g. `brew cleanup --hlep` should refuse to run rather than silently ignoring the `--hlep` argument. This will likely require implementing a new DSL for Homebrew commands that lists all the possible arguments (where currently just `ARGV.include?` is used).

GauthamGoli:
Why can't we use [this](https://docs.ruby-lang.org/en/2.1.0/OptionParser.html), instead of implementing a new DSL for parsing options? 
https://docs.ruby-lang.org/en/2.1.0/OptionParser.html

MikeMcQuaid:
@GauthamGoli We could use that but it wouldn't solve the problem by itself. The issue is that `ARGV.include?` is used throughout the codebase as global state. This is solveable with or without a new option parser.

Qchmqs:
the way it is now, those are hardcoded, can we extract them into a yaml, and try to apply fuzzy search first ? i feel like this will prove slow

MikeMcQuaid:
@Qchmqs Rather than a YAML we'd use a command DSL but yes, this might enable fuzzy search.

lwe:
Another simple solution could be to create an `ARGV.whitelist!(%w{--foobar --help})` in the commands that checks all args and raises on unknown arguments found. This approach would of course not work for install, but for other commands it would work very well. Biggest downside being that there is a large disconnect between actual place where the arg is used (could be anywhere in any file) vs. where the _known arguments_ are enforced. It still relies on globals and starts adding more dependency to `ARGV`.

A command/arg DSL would of course solve this problem, but this would be a rather large refactoring and there are about 500 call sites that use that global variable `ARGV`. First I think it would require changes that ARGV is passed as argument, rather than used as a global variable, before a command DSL can be implemented.

MikeMcQuaid:
@lwe Agreed that the command/arg DSL would be the longer-term solution; I think that's probably the best thing to push towards at this point.

apjanke:
There's no work-in-progress on this at #2420.

RandomDSdevel:
@apjanke:  Um, don't you mean something more along the lines of, "There's _some_ work in progress on this at #2420."…?

apjanke:
Uh, yes. I meant "now". Sorry.

RandomDSdevel:
[...]

PR #3610, opened 2017-12-30
cli : Add basic arg parser and use it for parsing `brew audit` args

GauthamGoli:

-----
#1860 **Emit warnings for mistyped options**

This a very basic WIP PR,  to get feedback on its usability & to demonstrate DSL of the parser; 

**Requirements:** 
 - DSL should be in declarative way
 - Reduce number of places `ARGV` is used, abstract it in DSL

**Example Code:**

```ruby
args = Homebrew::CLI::Parser.new do
  switch       "--strict",      "description of strict"
  switch       "--online",      "description of online"
  switch       "--new-formula", "description of new-formula"
  comma_array  "--only-cops",   "description of only-cops"
end.parse
```
3 types are supported as of now.
1. `switch` 
  A switch is an option that does not take arguments. It is used to turn options on/off . It can be both long and short
  Ex: `--new-formula` , `-D`  

2. `flag` 
 Flags take arguments. Long form flags take arguments with equal sign and short form flags don't require equal sign
  Ex: `--log-level=DEBUG` 

3. `comma_array` 
 Custom "flag" option type for Homebrew, which can take comma separated arguments
  Ex: `--only-cops=FormulaAudit,FormulaAuditStrict` 

**`Usage:`** 
**Example 1:** (All valid args)
`$ brew audit --new-formula wget` would result `args` to be as follows

`> p args` 
`> #<OpenStruct new_formula?=true>`

**Example 2:** (wrong args)

`$ brew audit --unknown-arg wget` would result in

`Error invalid option: --unknown-arg`

To start with, I have replaced usages of `ARGV` in `audit.rb` with this implementation.

### DIFF
--- a/Library/Homebrew/cli_parser.rb
+++ b/Library/Homebrew/cli_parser.rb
@@ -1,0 +1,55 @@
+require "optparse"
+require "ostruct"
+
+module Homebrew
+  module CLI
+    class Parser
+      def initialize(&block)
+        @parser = OptionParser.new
+        @parsed_args = OpenStruct.new
+        instance_eval(&block)
+      end
+
+      def switch(*names, description: nil)
+        description = option_to_description(*names) if description.nil?
+        @parser.on(*names, description) do
+          names.each do |name|
+            @parsed_args["#{option_to_name(name)}?"] = true
+          end
+        end
+      end
+
+      def comma_array(name, description: nil)
+        description = option_to_description(name) if description.nil?
+        @parser.on(name, OptionParser::REQUIRED_ARGUMENT, Array, description) do |list|
+          @parsed_args[option_to_name(name)] = list
+        end
+      end
+
+      def flag(name, description: nil, required: false)
+        if required
+          option_required = OptionParser::REQUIRED_ARGUMENT
+        else
+          option_required = OptionParser::OPTIONAL_ARGUMENT
+        end
+        description = option_to_description(name) if description.nil?
+        @parser.on(name, description, option_required) do |option_value|
+          @parsed_args[option_to_name(name)] = option_value
+        end
+      end
+
+      def option_to_name(name)
+        name.sub(/\A--?/, "").tr("-", "_")
+      end
+
+      def option_to_description(*names)
+        names.map { |name| name.sub(/\A--?/, "").tr("-", " ") }.sort.last
+      end
+
+      def parse(cmdline_args = ARGV)
+        @parser.parse!(cmdline_args)
+        @parsed_args
+      end
+    end
+  end
+end
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -49,20 +49,34 @@ require "cmd/style"
 require "date"
 require "missing_formula"
 require "digest"
+require "cli_parser"
 
 module Homebrew
   module_function
 
   def audit
-    inject_dump_stats!(FormulaAuditor, /^audit_/) if ARGV.switch? "D"
+    args = Homebrew::CLI::Parser.new do
+      switch      "--strict"
+      switch      "--online"
+      switch      "--new-formula"
+      switch      "--fix"
+      switch      "--display-cop-names"
+      switch      "--display-filename"
+      switch      "-D", "--audit-debug", description: "Activates debugging and profiling"
+      comma_array "--only"
+      comma_array "--except"
+      comma_array "--only-cops"
+      comma_array "--except-cops"
+    end.parse
+
     Homebrew.auditing = true
+    inject_dump_stats!(FormulaAuditor, /^audit_/) if args.audit_debug?
 
     formula_count = 0
     problem_count = 0
-
-    new_formula = ARGV.include? "--new-formula"
-    strict = new_formula || ARGV.include?("--strict")
-    online = new_formula || ARGV.include?("--online")
+    new_formula = args.new_formula?
+    strict = new_formula || args.strict?
+    online = new_formula || args.online?
 
     ENV.activate_extensions!
     ENV.setup_build_environment
@@ -75,35 +89,36 @@ module Homebrew
       files = ARGV.resolved_formulae.map(&:path)
     end
 
-    only_cops = ARGV.value("only-cops").to_s.split(",")
-    except_cops = ARGV.value("except-cops").to_s.split(",")
+    only_cops = args.only_cops
+    except_cops = args.except_cops
 
-    if !only_cops.empty? && !except_cops.empty?
+    if only_cops && except_cops
       odie "--only-cops and --except-cops cannot be used simultaneously!"
-    elsif (!only_cops.empty? || !except_cops.empty?) && (strict || ARGV.value("only"))
+    elsif (only_cops || except_cops) && (strict || args.only)
       odie "--only-cops/--except-cops and --strict/--only cannot be used simultaneously"
     end
 
-    options = { fix: ARGV.flag?("--fix"), realpath: true }
+    options = { fix: args.fix?, realpath: true }
 
-    if !only_cops.empty?
+    if only_cops
       options[:only_cops] = only_cops
-      ARGV.push("--only=style")
-    elsif new_formula
+      args.only = ["style"]
+    elsif args.new_formula?
       nil
     elsif strict
       options[:except_cops] = [:NewFormulaAudit]
-    elsif !except_cops.empty?
+    elsif except_cops
       options[:except_cops] = except_cops
     elsif !strict
       options[:only_cops] = [:FormulaAudit]
     end
 
+    options[:display_cop_names] = args.display_cop_names?
     # Check style in a single batch run up front for performance
     style_results = check_style_json(files, options)
 
     ff.sort.each do |f|
-      options = { new_formula: new_formula, strict: strict, online: online }
+      options = { new_formula: new_formula, strict: strict, online: online, only: args.only, except: args.except }
       options[:style_offenses] = style_results.file_offenses(f.path)
       fa = FormulaAuditor.new(f, options)
       fa.audit
@@ -113,7 +128,7 @@ module Homebrew
       formula_count += 1
       problem_count += fa.problems.size
       problem_lines = fa.problems.map { |p| "* #{p.chomp.gsub("\n", "\n    ")}" }
-      if ARGV.include? "--display-filename"
+      if args.display_filename?
         puts problem_lines.map { |s| "#{f.path}: #{s}" }
       else
         puts "#{f.full_name}:", problem_lines.map { |s| "  #{s}" }
@@ -196,6 +211,9 @@ class FormulaAuditor
     @new_formula = options[:new_formula]
     @strict = options[:strict]
     @online = options[:online]
+    @display_cop_names = options[:display_cop_names]
+    @only = options[:only]
+    @except = options[:except]
     # Accept precomputed style offense results, for efficiency
     @style_offenses = options[:style_offenses]
     @problems = []
@@ -205,9 +223,8 @@ class FormulaAuditor
 
   def audit_style
     return unless @style_offenses
-    display_cop_names = ARGV.include?("--display-cop-names")
     @style_offenses.each do |offense|
-      problem offense.to_s(display_cop_name: display_cop_names)
+      problem offense.to_s(display_cop_name: @display_cop_names)
     end
   end
 
@@ -777,17 +794,17 @@ class FormulaAuditor
   end
 
   def audit
-    only_audits = ARGV.value("only").to_s.split(",")
-    except_audits = ARGV.value("except").to_s.split(",")
-    if !only_audits.empty? && !except_audits.empty?
+    only_audits = @only
+    except_audits = @except
+    if only_audits && except_audits
       odie "--only and --except cannot be used simultaneously!"
     end
 
     methods.map(&:to_s).grep(/^audit_/).each do |audit_method_name|
       name = audit_method_name.gsub(/^audit_/, "")
-      if !only_audits.empty?
+      if only_audits
         next unless only_audits.include?(name)
-      elsif !except_audits.empty?
+      elsif except_audits
         next if except_audits.include?(name)
       end
       send(audit_method_name)
--- a/Library/Homebrew/test/cli_parser_spec.rb
+++ b/Library/Homebrew/test/cli_parser_spec.rb
@@ -1,0 +1,60 @@
+require_relative "../cli_parser"
+
+describe Homebrew::CLI::Parser do
+  describe "test switch options" do
+    subject(:parser) {
+      described_class.new do
+        switch "-v", "--verbose", description: "Flag for verbosity"
+        switch "--more-verbose",  description: "Flag for higher verbosity"
+      end
+    }
+
+    it "parses short option" do
+      args = parser.parse(["-v"])
+      expect(args).to be_verbose
+    end
+
+    it "parses a single valid option" do
+      args = parser.parse(["--verbose"])
+      expect(args).to be_verbose
+    end
+
+    it "parses a valid option along with few unnamed args" do
+      args = %w[--verbose unnamed args]
+      parser.parse(args)
+      expect(args).to eq %w[unnamed args]
+    end
+
+    it "parses a single option and checks other options to be nil" do
+      args = parser.parse(["--verbose"])
+      expect(args).to be_verbose
+      expect(args.more_verbose?).to be nil
+    end
+
+    it "raises an exception when an invalid option is passed" do
+      expect { parser.parse(["--random"]) }.to raise_error(OptionParser::InvalidOption, /--random/)
+    end
+  end
+
+  describe "test long flag options" do
+    subject(:parser) {
+      described_class.new do
+        flag        "--filename", description: "Name of the file", required: true
+        comma_array "--files",    description: "Comma separated filenames"
+      end
+    }
+    it "parses a long flag option with its argument" do
+      args = parser.parse(["--filename=random.txt"])
+      expect(args.filename).to eq "random.txt"
+    end
+
+    it "raises an exception when a flag's required arg is not passed" do
+      expect { parser.parse(["--filename"]) }.to raise_error(OptionParser::MissingArgument, /--filename/)
+    end
+
+    it "parses a comma array flag option" do
+      args = parser.parse(["--files=random1.txt,random2.txt"])
+      expect(args.files).to eq %w[random1.txt random2.txt]
+    end
+  end
+end


GauthamGoli:
Hi @alyssais 
>I suggest that, as a first pass, HomebrewCli.options be refactored to an instance method on a class that takes ARGV in the constructor. Something like:

> I also think that something like this would need comprehensive specs to demonstrate what functionality it provides, and to give confidence to other developers that it works as intended.

This existing code is already written in a way you described to enable writing tests. I'm adding a new commit where tests are added to show the existing functionality of the arg parser.

> I suggest that features like depends and conflicts be removed for now

`conflicts` is required for `audit` command, I'm adding a commit to demonstrate their use. Although I'm totally okay with adding them later.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -50,19 +50,33 @@
 require "missing_formula"
 require "digest"
 
+require_relative "../cli_parser"
```
MikeMcQuaid: Would be preferable to use a regular `require` here.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -50,19 +50,33 @@
...
+
 module Homebrew
   module_function
 
   def audit
     inject_dump_stats!(FormulaAuditor, /^audit_/) if ARGV.switch? "D"
     Homebrew.auditing = true
+    args, _unnamed_args = HomebrewCli.options do
```
MikeMcQuaid: `args, _ =`

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -50,19 +50,33 @@
...
+      option :strict, "strict", type: :boolean, default: false
```
MikeMcQuaid: The second argument should also be named and just default to a modified version of the symbol name,

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -50,19 +50,33 @@
...
+      option :online, "online", type: :boolean, default: false
+      option :new_formula, "new-formula", type: :boolean, default: false
+      option :fix, "fix", type: :boolean, default: false
+      option :display_cop_names, "display cop names", type: :boolean, default: false
+      option :display_filename, "display file names", type: :boolean, default: false
+      option :only, "only", type: :string
+      option :except, "except", type: :string
+      option :only_cops, "only cops", type: :string
```
MikeMcQuaid: `type: comma_array` or something might make this nicer to access.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -75,21 +89,15 @@ def audit
...
 
-    only_cops = ARGV.value("only-cops").to_s.split(",")
-    except_cops = ARGV.value("except-cops").to_s.split(",")
-
-    if !only_cops.empty? && !except_cops.empty?
-      odie "--only-cops and --except-cops cannot be used simultaneously!"
-    elsif (!only_cops.empty? || !except_cops.empty?) && (strict || ARGV.value("only"))
-      odie "--only-cops/--except-cops and --strict/--only cannot be used simultaneously"
-    end
+    only_cops = args.only_cops.to_s.split(",")
```
MikeMcQuaid: Maybe push the `to_s.split(",")` into a `comma_array` type?

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -50,19 +50,33 @@
...
+      conflicts :only_cops, :except_cops, :strict, :only
+    end
 
     formula_count = 0
     problem_count = 0
-
-    new_formula = ARGV.include? "--new-formula"
-    strict = new_formula || ARGV.include?("--strict")
-    online = new_formula || ARGV.include?("--online")
+    new_formula = args.new_formula
```
MikeMcQuaid: `args.new_formula?` for all of these reads a bit nicer.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+  class Parser
+    attr_accessor :ignore_invalid_options
+    attr_reader :unnamed
+
+    def initialize(&block)
+      @specs = {}
+      @long = {}
+      @short = {}
+      @order = []
+      @constraints = []
```
MikeMcQuaid: Not obvious to me from names what any of these are unfortunately. I'd make their names much longer or: could this perhaps use an ordered array of hashes?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+
+      instance_eval(&block)
+    end
+
+    def option(name, desc = "", opts = {})
```
MikeMcQuaid: Generate `desc` from the name, by default at least. Consider using named parameters over an options hash.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+      raise StandardError, "option #{o.name} already present" if @specs.member?(o.name) || @long[o.long]
+
+      @long[o.long] = o.name
+      @specs[o.name] = o
+
+      @order << [:opt, o.name]
+    end
+
+    def depends(*options)
+      options.each { |option| raise StandardError, "unknow #{option}" unless @specs[option] }
```
MikeMcQuaid: `unknow` typo here.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+      @constraints << [:depends, options]
+    end
+
+    def conflicts(*options)
+      options.each { |option| raise StandardError, "unknow #{option}" unless @specs[option] }
```
MikeMcQuaid: `unknow` typo here.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+      @constraints << [:conflicts, options]
+    end
+
+    def parse(cmdline = ARGV)
```
MikeMcQuaid: I'd consider not having a default argument here to decouple things a bit. This function is also pretty long and might be a good candidate to split into a few smaller functions.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+      vals = {}
+      required = {}
```
MikeMcQuaid: Not sure what these variables mean. I'd make their names much longer.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+        end
+        next nil if ignore_invalid_options && !sym
+        raise StandardError, "unknown argument #{arg}" unless sym
+
+        given_args[sym] ||= {}
+        given_args[sym][:arg] = arg
+        given_args[sym][:params] ||= []
+
+        num_params_taken = 0
+        unless params.empty?
```
MikeMcQuaid: `next num_params_taken if params.empty?`

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+          if @specs[sym].single_arg?
+            given_args[sym][:params] << params[0, 1]
```
MikeMcQuaid: Very confusing as to what this line is doing.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+        num_params_taken
+      end
+
+      @constraints.each do |type, syms|
+        constraint_sym = syms.find { |sym| given_args[sym] }
+        next unless constraint_sym
+
+        case type
+        when :depends
+          syms.each { |sym| raise StandardError, "--#{@specs[constraint_sym].long} requires --#{@specs[sym].long}" unless given_args.include? sym }
```
MikeMcQuaid: There's a lot of `raise StandardError` with long strings going on; I think there's probably a better pattern for these.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+        case opts.type
+        when :string, :strings
+          vals[sym] = params.map { |pg| pg.map(&:to_s) }
+        when :flag
+          vals[sym] = true
+        end
+
+        if opts.single_arg?
+          if opts.multi?
+            vals[sym] = vals[sym].map { |p| p[0] }
```
MikeMcQuaid: This line is pretty confusing.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+          else
+            vals[sym] = vals[sym][0][0]
```
MikeMcQuaid: This line is also pretty confusing.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+
+    SINGLE_ARG_TYPES = [:string].freeze
+
+    MULTI_ARG_TYPES = [:strings].freeze
+
+    def initialize(name, _desc = "", opts = {})
+      opts[:type] = case opts[:type]
+      when :boolean, :bool then :flag
+      when :string then :string
+      when Class
```
MikeMcQuaid: What's this used for currently?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+        case opts[:type].name
+        when "String" then :string
+        else
+          raise StandardError, "unsupported arg type"
+        end
+      when nil then nil
+      else
+        raise StandardError
```
MikeMcQuaid: Should have a message.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+      end
+
+      opts[:type] || :flag
+
+      opts[:long] = opts[:long] ? opts[:long].to_s : name.to_s.tr("_", "-")
```
MikeMcQuaid: I think `if` would make this easier to follow.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+      opts[:long] = case opts[:long]
```
MikeMcQuaid: consider a different variable to make the transformation more obvious

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,280 @@
...
+      when /^--([^-].*)$/ then Regexp.last_match(1)
+      when /^[^-]/ then opts[:long]
+      else raise StandardError, "invalid long option name #{opts[:long].inspect}"
+      end
+
+      opts[:default] = false if opts[:type] == :flag && opts[:default].nil?
+      opts[:default] = [opts[:default]] if opts[:default] && opts[:multi] && !opts[:default].is_a?(Array)
+      opts[:default] = "" if opts[:type] == :string && opts[:default].nil?
+
+      opts[:default] ||= false
```
MikeMcQuaid: The above logic for `opts[:default]` is all pretty confusing.

alyssais:
Thanks for that @GauthamGoli! Those tests do a good job of demonstrating that the code is easily testable. I think it would be good to have more examples at some point, but those ones are fine for now.

I had another look through the PR and was about to leave another review, but then I realised that it'll probably change enough as a result of you addressing Mike's that any comments I made now would probably be irrelevant immediately. So I'll wait until you've dealt with Mike's comments and then try to have another look. :)

Library/Homebrew/cli_parser.rb
GauthamGoli: This can be used for may be datetime class , I’ve seen some date arguments in Homebrew . Anyways I think I’m trying to do too much in this PR supporting named args , switches and named args with multiple params. I’m stripping all the other features amd rewriting with implementation of named and boolean arguments

alyssais: I think that sounds very sensible @GauthamGoli.

Library/Homebrew/cli_parser.rb
GauthamGoli: Now this is split this into smaller functions.
Can you explain how not having a default argument would decouple things? The default argument is to prevent passing ARGV explicitly everywhere.

Library/Homebrew/cli_parser.rb
GauthamGoli: Addressed

Library/Homebrew/cli_parser.rb
GauthamGoli: Tried making it simpler now

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: Addressed

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: Addressed

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: Figuring out a way to do this.

Library/Homebrew/cli_parser.rb
GauthamGoli: Addressed.

GauthamGoli:
Hi, I tried addressing almost all the review comments. Please let me know your thoughts on the new code, if it's still difficult to follow.

Library/Homebrew/cli_parser.rb
GauthamGoli: Wrote a base `CommandLineError` exception class and inherited it in other cli exception classes

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+          if num_params_taken.nil?
+            remaining_args << args[i]
+          else
+            i += num_params_taken
+          end
+        else
+          remaining_args << args[i]
+        end
+        i += 1
+      end
```
alyssais: I think it would be more readable if we duplicated the string and then just shifted arguments off it, rather than using an `i` variable that has to be manually incremented.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+
+      until i >= args.length
+        case args[i]
+
+        # match: --argument=<params>
+        when /^--(\S+?)=(.*)$/
+          num_params_taken = yield "--#{Regexp.last_match(1)}", [Regexp.last_match(2)]
+          remaining_args << args[i] if num_params_taken.nil?
+
+        # match: --argument param1 param2
```
alyssais: I'm not sure this is the correct behaviour. As a UNIX user, if I do `brew install --env std hello`, I expect `hello` to be interpreted as an unnamed argument, rather than as a parameter to `--env`. In my experience, if an argument has to take multiple values, it's more common to either comma-seperate them, or allow the argument to be supplied multiple times.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+            super
+          end
+        end
+
+        # https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding
+        def respond_to_missing?(m, include_private = false)
+          super
+        end
+      end
+      parsed_args
```
alyssais: This could be much simpler:

```ruby
OpenStruct.new(parsed_args)
```

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+    # Returns array of unnamed args
+    def each_arg(args)
+      remaining_args = []
+      i = 0
+
+      until i >= args.length
+        case args[i]
+
+        # match: --argument=<params>
+        when /^--(\S+?)=(.*)$/
```
alyssais: Shouldn't we reuse `PARAM_RE` here?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+      while args[pos] && args[pos] !~ PARAM_RE
+        params << args[pos]
+        pos += 1
+      end
+      params
+    end
+  end
+
+  # Regex to match command line options (but not their parameters)
+  PARAM_RE = /^-(-|\.$|[^\d\.])/
```
alyssais: I think this could have a better name. Even if it were `PARAM_REGEXP`.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+    attr_accessor :name, :option_attributes
+
+    FLAG_TYPES = [:flag, :bool, :boolean].freeze
+
+    SINGLE_ARG_TYPES = [:string, :comma_array].freeze
+
+    def initialize(name, description, type, default, required)
+      option_attributes = {}
+      option_attributes[:type] = case type
+      when :boolean, :bool then :flag
```
alyssais: I'm not sure we need to support both of these *in addition to `flag`*.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+  # Declared options are encapsulated by `Option` class
+  class Option
+    attr_accessor :name, :option_attributes
+
+    FLAG_TYPES = [:flag, :bool, :boolean].freeze
+
+    SINGLE_ARG_TYPES = [:string, :comma_array].freeze
+
+    def initialize(name, description, type, default, required)
+      option_attributes = {}
```
alyssais: Why do you use a hash here instead of just instance variables?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+      when nil then :flag # By default an option is considered a flag
+      else
+        raise UnknownOptionTypeError, type
+      end
+
+      option_attributes[:name] = convert_name_to_opt(name)
+
+      # Only store the option name
+      option_attributes[:name] = case option_attributes[:name]
+      when /^--([^-].*)$/ then Regexp.last_match(1)
```
alyssais: Can we reuse `PARAM_RE` here?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+      when :comma_array then :comma_array
+      when nil then :flag # By default an option is considered a flag
+      else
+        raise UnknownOptionTypeError, type
+      end
+
+      option_attributes[:name] = convert_name_to_opt(name)
+
+      # Only store the option name
+      option_attributes[:name] = case option_attributes[:name]
```
alyssais: Why do you set this, then immediately change its value on the next line?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
...
+      option_attributes[:description]
+    end
+
+    def required?
+      option_attributes[:required]
+    end
+
+    def self.create(name, description, type, default, required)
+      new(name, description, type, default, required)
+    end
```
alyssais: What is the purpose of this method?

Library/Homebrew/test/cli_parser_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    it "with multiple flags and few args passed" do
+      parser = HomebrewCli::Parser.new do
+        option :verbose, description: "Flag for verbosity", type: :bool
+        option :more_verbose, description: "Flag for higher verbosity", type: :bool
+      end
+      args = parser.parse(["--verbose"])
+      expect(args.verbose).to be true
+      expect(args.more_verbose).to be nil
```
alyssais: Shouldn't this be `false`? `flag`s are defined to default to `false` above.

Library/Homebrew/test/cli_parser_spec.rb
```diff
@@ -0,0 +1,62 @@
+require_relative "../cli_parser"
+
+describe HomebrewCli::Parser do
+  describe "test boolean flags" do
+    it "with a valid flag and valid arg" do
+      parser = HomebrewCli::Parser.new do
+        option :verbose, description: "Flag for verbosity", type: :bool
```
alyssais: Could we define an RSpec `subject` here with a single parser created with a bunch of different options? Would make a big difference to the length of each test, which is a big win IMO.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,271 @@
+module HomebrewCli
```
alyssais: I think `HomebrewCLI` or, even better `Homebrew::CLI` reads nicer and is more Ruby-ish.

MikeMcQuaid:
I agree with pretty much all @alyssais's comments so I'd be happy to see you both iterate on this and do more review myself later. Have a chat with each other on Slack about whether @GauthamGoli would be happy with @alyssais pushing commits directly to this PR (assuming @alyssais would want to do that). Thanks both of you ❤️!

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,50 @@
...
+      def handle_comma_array(name, description, _)
+        args = [name, "=MANDATORY", Array, description]
+        @parser.on(*args) do |list|
+          @parsed_args.send("#{option_to_name(name)}=", list)
+        end
+      end
+
+      def handle_flag(name, description, required)
+        args = [name, description, required ? "=MANDATORY" : "=[OPTIONAL]"]
+        @parser.on(*args) do |option_value|
```
alyssais: Why an `args` array here, instead of just passing the arguments normally?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      def handle_comma_array(name, description, _)
+        args = [name, "=MANDATORY", Array, description]
+        @parser.on(*args) do |list|
+          @parsed_args.send("#{option_to_name(name)}=", list)
+        end
+      end
+
+      def handle_flag(name, description, required)
+        args = [name, description, required ? "=MANDATORY" : "=[OPTIONAL]"]
```
alyssais: We can use `OptionParser::REQUIRED_ARGUMENT` and `OptionParser::OPTIONAL_ARGUMENT` here instead of these strings, which would be a bit nicer.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      def option(name, description: nil, type: nil, required: false)
+        handler_name = "handle_#{type}".to_sym
+        send(handler_name, name, description, required)
+      end
+
+      def handle_switch(name, description, _)
+        @parser.on(name, description) do
+          @parsed_args.send("#{option_to_name(name)}?=", true)
```
alyssais: Using `OpenStruct#[]=` would read nicer here:

```ruby
@parsed_args["#{option_to_name(name)}?"] = true
```

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,50 @@
...
+    class Parser
+      attr_accessor :parser
+
+      def initialize(&block)
+        @parser = OptionParser.new
+        @parsed_args = OpenStruct.new
+        instance_eval(&block)
+      end
+
+      def option(name, description: nil, type: nil, required: false)
```
alyssais: I think it would be a nicer API if `description` was just the second positional parameter, not a keyword argument. And, rather than a type parameter, it might be nicer to have different methods for different types of options? That would probably make the implementation simpler too.

So, the `audit` option definition would look something like this:

```ruby
switch       "--strict",             "strict"
switch       "--online",             "online"
switch       "--new-formula",        "new-formula"
switch       "--fix",                "fix"
switch       "--display-cop-names",  "display cop names"
switch       "--display-filename",   "display file names"

comma_array  "--only",               "only"
comma_array  "--except",             "except"
comma_array  "--only-cops",          "only cops"
comma_array  "--except-cops",        "except cops"
```

GauthamGoli: Yeah, I was also thinking of having different methods! 👍

Library/Homebrew/cli_parser.rb
GauthamGoli: There's a conditional argument, hence thought of writing it this way, but passing normally also is fine I think.

Library/Homebrew/test/cli_parser_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    it "with valid option and unnamed args passed" do
+      args = %w[--verbose unnamed args]
+      parser.parse(args)
+      expect(args).to eq %w[unnamed args]
+    end
+
+    it "with a single option passed" do
+      args = parser.parse(["--verbose"])
+      expect(args.verbose?).to be true
```
alyssais: I prefer `expect(args).to be_verbose`, but if you like it this way I think it's fine.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      def flag(name, description, required: false)
+        option_required = required ? OptionParser::REQUIRED_ARGUMENT : OptionParser::OPTIONAL_ARGUMENT
+        @parser.on(name, description, option_required) do |option_value|
+          @parsed_args[option_to_name(name)] = option_value
+        end
+      end
+
+      def option_to_name(name)
+        name.gsub(/^--?/, "").tr("-", "_")
```
alyssais: `sub` would work here instead of `gsub`

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      def comma_array(name, description)
+        @parser.on(name, OptionParser::REQUIRED_ARGUMENT, Array, description) do |list|
+          @parsed_args[option_to_name(name)] = list
+        end
+      end
+
+      def flag(name, description, required: false)
+        option_required = required ? OptionParser::REQUIRED_ARGUMENT : OptionParser::OPTIONAL_ARGUMENT
```
alyssais: Could this be a multi-line `if`/`else`? The line is quite long.

Library/Homebrew/test/cli_parser_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    it "with a long flag option and its parameter" do
+      args = parser.parse(["--filename=random.txt"])
+      expect(args.filename).to eq "random.txt"
+    end
+
+    it "with flag with its required parameter not passed" do
+      expect { parser.parse(["--filename"]) }.to raise_error(OptionParser::MissingArgument, /--filename/)
+    end
+
+    it "with comma array flag" do
```
alyssais: These test names could be improved. They should read like English sentences, including the `it`, eg `it "parses a comma array"`.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,44 @@
+require "optparse"
+require "ostruct"
+
+module Homebrew
+  module CLI
+    class Parser
+      attr_accessor :parser
```
alyssais: This doesn't need to be public, so this line could be removed.

GauthamGoli:
@alyssais Thanks! Pushed changes addressing review comments.

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,46 @@
...
+        else
+          option_required = OptionParser::OPTIONAL_ARGUMENT
+        end
+        @parser.on(name, description, option_required) do |option_value|
+          @parsed_args[option_to_name(name)] = option_value
+        end
+      end
+
+      def option_to_name(name)
+        name.sub(/^--?/, "").tr("-", "_")
```
alyssais: Use `\A` (start of string) rather than `^` (start of line).

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -49,20 +49,34 @@
...
+      switch      "--display-filename",   "display file names"
+      switch      "-D",                   "Activates debugging and profiling"
+      comma_array "--only",               "only"
+      comma_array "--except",             "except"
+      comma_array "--only-cops",          "only cops"
+      comma_array "--except-cops",        "except cops"
+    end.parse
+
     Homebrew.auditing = true
+    inject_dump_stats!(FormulaAuditor, /^audit_/) if args.D?
```
alyssais: `args.D?` feels a little odd. Shall we give it a long name for which `-D` on the command line can be a shorthand? `--audit-debug` or something?

GauthamGoli: Yeah. I've updated the code to support both short and long switches now, and added corresponding test.

Library/Homebrew/test/cli_parser_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  describe "test switch options" do
+    subject(:parser) {
+      described_class.new do
+        switch "-v", "--verbose", "Flag for verbosity"
+        switch "--more-verbose",  "Flag for higher verbosity"
+      end
+    }
+    it "parses a single valid option" do
+      args = parser.parse(["--verbose"])
+      expect(args.verbose?).to be true
```
alyssais: `expect(args).to be_verbose` reads nicer IMO

GauthamGoli: Agree. 👍 Updated it accordingly.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -49,20 +49,34 @@
...
 require "digest"
+require "cli_parser"
 
 module Homebrew
   module_function
 
   def audit
-    inject_dump_stats!(FormulaAuditor, /^audit_/) if ARGV.switch? "D"
+    args = Homebrew::CLI::Parser.new do
+      switch      "--strict",             "strict"
```
MikeMcQuaid: I think it would be good if the second argument was inferred for all of these except `"Activates debugging and profiling"`.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -49,20 +49,34 @@
...
+    inject_dump_stats!(FormulaAuditor, /^audit_/) if args.audit_debug?
 
     formula_count = 0
     problem_count = 0
-
-    new_formula = ARGV.include? "--new-formula"
-    strict = new_formula || ARGV.include?("--strict")
-    online = new_formula || ARGV.include?("--online")
+    new_formula = args.new_formula?
+    strict = new_formula || args.strict?
```
MikeMcQuaid: I like this API 👍

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -75,35 +89,36 @@ def audit
       files = ARGV.resolved_formulae.map(&:path)
     end
 
-    only_cops = ARGV.value("only-cops").to_s.split(",")
-    except_cops = ARGV.value("except-cops").to_s.split(",")
+    only_cops = args.only_cops
```
MikeMcQuaid: Could maybe avoid setting these variables at all and just use `args.only_cops`?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+module Homebrew
+  module CLI
+    class Parser
+      def initialize(&block)
+        @parser = OptionParser.new
+        @parsed_args = OpenStruct.new
+        instance_eval(&block)
+      end
+
+      def switch(*names, description)
```
MikeMcQuaid: Maybe make description a named parameter i.e. `description:`?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+        @parser.on(*names, description) do
+          names.each do |name|
+            @parsed_args["#{option_to_name(name)}?"] = true
+          end
+        end
+      end
+
+      def comma_array(name, description)
```
MikeMcQuaid: Maybe make description a named parameter i.e. `description:`?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+        @parser.on(name, OptionParser::REQUIRED_ARGUMENT, Array, description) do |list|
+          @parsed_args[option_to_name(name)] = list
+        end
+      end
+
+      def flag(name, description, required: false)
```
MikeMcQuaid: Maybe make description a named parameter i.e. `description:`?

MikeMcQuaid: What differentiates a flag and a switch?

Library/Homebrew/cli_parser.rb
```diff
@@ -0,0 +1,48 @@
...
+          option_required = OptionParser::REQUIRED_ARGUMENT
+        else
+          option_required = OptionParser::OPTIONAL_ARGUMENT
+        end
+        @parser.on(name, description, option_required) do |option_value|
+          @parsed_args[option_to_name(name)] = option_value
+        end
+      end
+
+      def option_to_name(name)
```
MikeMcQuaid: Could have a similar `option_to_description` method?

Library/Homebrew/cli_parser.rb
GauthamGoli: `switch` is an option that does not take arguments. It is used to turn options on/off . It can be both long and short
Ex: --new-formula , -D

`flag`s take arguments. Long form flags take arguments with equal sign and short form flags don't require equal sign
Ex: --log-level=DEBUG

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: `only_cops` and `expect_cops` are passed as arguments to `check_style_json` call below. We can't avoid setting them.

GauthamGoli:
@MikeMcQuaid Addressed comments and pushed my changes.

MikeMcQuaid:
Looks good here and locally, thanks again @GauthamGoli!

MikeMcQuaid:
Note this works because it caught the use of `brew audit --devel` in `brew test-bot` which did nothing 🎉

PR #3627, opened 2018-01-03
pathname: Use CI env variable rather than TRAVIS

sjackman:

-----

Add `CIRCLECI`, `CIRCLE_*`, and `CI_*` to the white list.
See https://circleci.com/docs/2.0/env-vars/
and https://circleci.com/docs/1.0/environment-variables/

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -499,7 +499,7 @@ module ObserverPathnameExtension
     MAXIMUM_VERBOSE_OUTPUT = 100
 
     def verbose?
-      return ARGV.verbose? unless ENV["TRAVIS"]
+      return ARGV.verbose? unless ENV["CI"]
       return false unless ARGV.verbose?
 
       if total < MAXIMUM_VERBOSE_OUTPUT


MikeMcQuaid:
Can you elaborate on why this is needed?

MikeMcQuaid:
(To elaborate: Travis only is whitelisted because we use it for our Homebrew/brew CI).

sjackman:
I feel that it's worth supporting external taps that wish to use CircleCI for CI. CircleCI may be used to build bottles for both macOS and Linux. `Homebrew/homebrew-test-bot` supports CircleCI.

MikeMcQuaid:
@sjackman Circle CI is not unsupported as-is and none of these variables would be used in Homebrew internals. `brew test-bot` does not filter the environment. Can you be more specific about what the actual problem is you're trying to solve here?

sjackman:
> `brew test-bot` does not filter the environment.

Ah, I didn't realize that. Thanks for the info. I suppose it's just for this one line then. Sorry that I hadn't included it in the original PR.

https://github.com/Homebrew/brew/pull/3627/files#diff-06540c12bb8e84922c0d329ed05b19bfR502

We run Linuxbrew/brew CI on both CircleCI and TravisCI. If it's not needed for Homebrew/brew though, then I'll close this PR.

sjackman:
I suppose this PR could be useful for external taps running on CircleCI with
```
HOMEBREW_VERBOSE=1
HOMEBREW_VERBOSE_USING_DOTS=1
```
to avoid timeouts.

MikeMcQuaid:
> Ah, I didn't realize that. Thanks for the info. I suppose it's just for this one line then. Sorry that I hadn't included it in the original PR.

Could that be `ENV["CI"]` or similar instead?

> We run Linuxbrew/brew CI on both CircleCI and TravisCI.

Why?

sjackman:
> Could that be ENV["CI"] or similar instead?

Good idea. Works for me.

sjackman:
>> We run Linuxbrew/brew CI on both CircleCI and TravisCI.

> Why?

We build bottles on CircleCI, so it's important that `brew tests` passes on both CircleCI and TravisCI. It's reasonably rare, but occasionally a test will pass on TravisCI but fail on CircleCI, due to differences in their configuration. For example, the version of `git` at `/usr/bin/git`. For example, `git clone` with `git` 1.8 outputs `Cloning into` to standard output, whereas recent versions of `git` output `Cloning into` to standard error. That difference causes a  `brew test` failure. If something like that slips through, every tap PR fails due to `brew tests` failing.

sjackman:
TravisCI isn't actually using `git` 1.8. It's just an example of how a configuration difference of the host can cause `brew tests` to fail.

MikeMcQuaid:
> That difference causes a brew test failure. If something like that slips through Linuxbrew/brew, every tap PR fails due to brew tests failing.

Tap PRs do not run `brew tests` so I'm still not sure I understand here.

sjackman:
> Tap PRs do not run brew tests so I'm still not sure I understand here.

I got that wrong, sorry. Nonetheless, we run Linuxbrew/brew `test-bot` on CircleCI to ensure that `brew tests` passes on Ubuntu 16. TravisCI runs Ubuntu 14.

MikeMcQuaid:
Thanks @sjackman!

sjackman:
Thanks for merging, Mike!

PR #3632, opened 2018-01-06
Preserve HOMEBREW_CELLAR if already specified

khenriks:

-----
Currently, users can freely choose HOMEBREW_PREFIX (based on where the symlink is located) and HOMEBREW_REPOSITORY (based on the location of the repository/where the symlink points) but cannot specify HOMEBREW_CELLAR. This provides an easy way to make that specifiable by users.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -5,8 +5,12 @@ then
 fi
 
 # Where we store built products; a Cellar in HOMEBREW_PREFIX (often /usr/local
-# for bottles) unless there's already a Cellar in HOMEBREW_REPOSITORY.
-if [[ -d "$HOMEBREW_REPOSITORY/Cellar" ]]
+# for bottles) unless there's already a Cellar in HOMEBREW_REPOSITORY or the
+# cellar is already defined via HOMEBREW_CELLAR.
+if [[ -n "$HOMEBREW_CELLAR" ]]
+then
+  :
+elif [[ -d "$HOMEBREW_REPOSITORY/Cellar" ]]
 then
   HOMEBREW_CELLAR="$HOMEBREW_REPOSITORY/Cellar"
 else


MikeMcQuaid:
Sorry, this is not desired functionality as there's assumptions made internally and in external scripts about the relationship in location between cellar and prefix.

khenriks:
Can I at least make a case for it?

Given that right now, the cellar can be defined based on the repository instead, that means that there **isn't** necessarily going to be any relationship between prefix and cellar, since the prefix and repository can be in completely unrelated locations.

On top of that, there are good reasons to want something like this functionality. Specifically, this allows a user to have an arbitrary prefix while still using /usr/local/Cellar as the cellar location, which permits most bottles to work as-is without any problems. This is useful for example if the user's administrator wants to lock down paths like /usr/local/bin and /usr/local/lib but has no concerns with leaving paths like /usr/local/Cellar alone. (This is not hypothetical, and represents a possible corporate computing environment.) Being able to still use bottles in that situation is very useful.

Could you provide additional information on the internal requirements on cellar and prefix location? I'm happy to attempt to patch that as well. I've been using a patch very much like this one for quite a while now, and haven't run into problems, so I'm a bit perplexed as to what the issues could be.

As for external scripts, this doesn't change the default configuration of Homebrew, so doesn't automatically break any scripts. If users customize the cellar path, the script would still continue to work if it relies on `brew --cellar` to find the cellar. Only if it makes assumptions about Homebrew internals would it break, in which case the user can fix it, the script author can fix it, the user can stop using it, or the user can stop using this new feature. That seems like a fine outcome.

MikeMcQuaid:
> This is useful for example if the user's administrator wants to lock down paths like /usr/local/bin and /usr/local/lib but has no concerns with leaving paths like /usr/local/Cellar alone. (This is not hypothetical, and represents a possible corporate computing environment.) Being able to still use bottles in that situation is very useful.

This is exactly what we do not want to permit as it's extremely unlikely that those bottles will continue to work consistently and not fail with hard-to-debug errors.


> Only if it makes assumptions about Homebrew internals would it break, in which case the user can fix it, the script author can fix it, the user can stop using it, or the user can stop using this new feature.

Unless it breaks in a way that isn't obvious that it is the problem. 

---

Sorry, adding additional configurability for extremely niche use-cases is not something we can feasibly support without compromising the quality of Homebrew which we don't want to do. If your local patch works for you: great, keep using that.

PR #3636, opened 2018-01-07
rubocop 0.52.1

MikeMcQuaid:
Let's see how this goes.

### DIFF
--- a/Library/.rubocop.yml
+++ b/Library/.rubocop.yml
@@ -123,6 +123,14 @@ Style/Documentation:
 Style/Encoding:
   Enabled: true
 
+# disabled until we get the Metrics/LineLength down to 80.
+Style/IfUnlessModifier:
+  Enabled: false
+
+# messes with existing plist/caveats style
+Style/TrailingBodyOnMethodDefinition:
+  Enabled: false
+
 # use spaces for indentation; detect tabs
 Layout/Tab:
   Enabled: true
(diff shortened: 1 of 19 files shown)


Library/.rubocop.yml
```diff
@@ -123,6 +123,10 @@ Style/Documentation:
 Style/Encoding:
   Enabled: true
 
+# disabled until we get the Metrics/LineLength down to 80.
```
alyssais: Is that something we're actively working on/making progress in?

MikeMcQuaid: Not currently.

MikeMcQuaid:
@GauthamGoli any ideas on about the failures here?

GauthamGoli:
@MikeMcQuaid Yeah https://github.com/bbatsov/rubocop/issues/3558
We have to make `autocorrect` public in all the cops. Just remove `private` keyword in following cops
 - `bottle_block_cop.rb`
 - `checksum_cop.rb`
 - `class_cop.rb`
 - `components_order_cop.rb`
 - `formula_desc_cop.rb`
 - `urls_cop.rb`

MikeMcQuaid:
Thanks @GauthamGoli! Let's give it another shot.

MikeMcQuaid:
CC @reitermarkus Can you add `rubocop-cask` support? Thanks!

ilovezfs:
This PR caused `--display-cop-names` to become the default. So there is now a redundant command line option in both `brew audit` and `brew style`, and there is no way to suppress the display of the fairly-useless-to-users cop names.

ilovezfs:
https://github.com/Homebrew/brew/pull/3710

PR #3639, opened 2018-01-07
Update formulae.brew.sh URL

alyssais:
The Homebrew/formulae.brew.sh repo implies that this is now the preferred URL.


MikeMcQuaid:
Cheers @alyssais.

PR #1301, opened 2016-10-16
dependency_collector: need xz dep on <= macOS 10.8

ilovezfs:
Mountain Lion: http://opensource.apple.com/source/libarchive/libarchive-25.1/config.h
  `/* #undef HAVE_LIBLZMA */`

Mavericks: http://opensource.apple.com/source/libarchive/libarchive-29/config.h
  `#define HAVE_LIBLZMA 1`

See https://github.com/Homebrew/brew/issues/1299.

PR #3641, opened 2018-01-08
Feature: support bottles compressed with XZ

kabel:

-----
This will allow for formulae to be bottled with XZ compression, which results in greatly reducing the bandwidth required to download without too much sacrifice in terms of decompression speed. This flexibility allows formulae to transition to XZ bottles gradually.

The motivation is to follow the trend of package managers using this compression format, and to reduce the amount of bandwidth (and possibly time) required for installing bottles. I see from the source comments that bzip2 was likely considered earlier, however it did perform as well enough to justify it's compression ratios. XZ has even better compression ratios and is not too much slower that gzip for decompression. Compression _would_ take noticeably longer, but that's only a user-experience degradation for bottle builders (CI and the like).

I may need some feedback on how best to provide the xz binary for the bottle command, since it isn't included by macOS by default (only the tar binary has support).


scpeters:
Prior discussion on this topic:

* https://github.com/Homebrew/brew-evolution/pull/17

It's a long thread, but the last comment by @MikeMcQuaid included "we'd accept a PR for this."

scpeters:
Other background info: #1279, #1301

Library/Homebrew/shims/super/xz
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+XZ="$(/usr/bin/which "$HOMEBREW_PREFIX/opt/xz/bin/xz" ||
+      echo /usr/bin/xz)"
```
MikeMcQuaid: I'd just always use the `HOMEBREW_PREFIX/opt/xz/bin/xz` and not create a shim given this path doesn't exist on macOS.

Library/Homebrew/test/formula_installer_bottle_spec.rb
```diff
@@ -59,6 +60,25 @@ def temporarily_install_bottle(formula)
       bin = HOMEBREW_PREFIX/"bin"
       expect(bin).to be_a_directory
     end
+
+    temporarily_install_bottle(TestballXzBottle.new) do |f|
```
MikeMcQuaid: Make this a separate test.

Library/Homebrew/test/support/fixtures/bottles/testball_bottle-0.1.mountain_lion.bottle.tar.xz
```diff
@@ -0,0 +1 @@
+testball_bottle-0.1.yosemite.bottle.tar.xz
```
MikeMcQuaid: Do we need a fixture for every version? Would be good to remove at least the need for mountain lion/yosemite fixtures.

MikeMcQuaid:
Oh, also: would be good to have `xz` be autoinstalled when needed for bottle time. It's not needed at extraction time, right?

scpeters:
@MikeMcQuaid that sounds right to have `xz` autoinstalled at bottle time to speed up with parallel compression, but it shouldn't be needed for extraction (based on my reading of https://github.com/Homebrew/brew-evolution/pull/17)

DomT4:
> (based on my reading of Homebrew/brew-evolution#17)

Only for Mavericks and newer versions of macOS, right? Older versions will still need the `xz` dep for extraction?

Not that it's a major use-case these days given Homebrew only bottles for three versions of macOS versions, but it might be worth considering for custom taps/etc? Unsure 🤷‍♂️.

scpeters:
@DomT4 thanks, you are right; older OS versions will still need `xz` for extraction

iMichka:
And linuxbrew will also need it, because we can’t rely on xz being installed. Don’t worry too much about this, we can make a second pass later to implement this. But if there is some easy way to prepare the code for this, so we do not need to rewrite half of it later, that would be great. Maybe @sjackman may want to have a look at this patch set.

But of course do not worry too much about Linux, we can handle that later if needed :)

sjackman:
We can make a portable bottle for `xz` for Linux. That bottle would have to be `gz` compressed though rather than `xz`. Other than that exception, it should be straight forward.

sjackman:
By a portable bottle, I mean a bottle in `homebrew-core` compiled against an older version of `glibc` (not a bottle in `homebrew-portable`).

kabel:
Thanks for the feedback. I'm working on the changes, but am running into issues testing the auto install of xz. I'm not very familiar with rspec or the Homebrew integration test framework, and I don't know if the tests can use the formulae from the core tap.

alyssais:
They can’t, no. You’ll have to use stubbing/mocking.

kabel:
> You’ll have to use stubbing/mocking.

That's problematic for an _integration_ test. Needing to test that the bottle command is able to install and use a working xz binary doesn't seem like something that should/can be mocked.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -190,7 +190,10 @@ def bottle_formula(f)
       rebuild = rebuilds.empty? ? 0 : rebuilds.max.to_i + 1
     end
 
-    compression_type = f.bottle_specification.compression_type
+    compression_type = :xz
```
DomT4: You're going to have to carve out an exemption here for the `xz` formula itself, at the very least on Linux, presumably. Otherwise you're going to end up with the `xz` bottle compressed with `xz`, which is problematic.

DomT4: 😆 You beat me to it. I'll shut up and let ya do your thing for a bit!

Library/Homebrew/utils/bottles.rb
```diff
@@ -29,16 +29,16 @@ def native_regex
       end
 
       def receipt_path(bottle_file)
-        path = Utils.popen_read("tar", "-tf", bottle_file).lines.map(&:chomp).find do |line|
-          line =~ %r{.+/.+/INSTALL_RECEIPT.json}
-        end
+        command = tar_command "tf", bottle_file, "*/*/INSTALL_RECEIPT.json"
```
sjackman: `"*/*/INSTALL_RECEIPT.json"` requires the option `--wildcards` with GNU `tar` on Linux.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -187,11 +190,15 @@ def bottle_formula(f)
       rebuild = rebuilds.empty? ? 0 : rebuilds.max.to_i + 1
     end
 
-    filename = Bottle::Filename.create(f, Utils::Bottles.tag, rebuild)
+    compression_type = :xz
```
MikeMcQuaid: Would be good for this to be an `else`

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -239,10 +246,15 @@ def bottle_formula(f)
           tar_path.utime(tab.source_modified_time, tab.source_modified_time)
           relocatable_tar_path = "#{f}-bottle.tar"
           mv tar_path, relocatable_tar_path
-          # Use gzip, faster to compress than bzip2, faster to uncompress than bzip2
-          # or an uncompressed tarball (and more bandwidth friendly).
-          safe_system "gzip", "-f", relocatable_tar_path
-          mv "#{relocatable_tar_path}.gz", bottle_path
+          case compression_type
+          when :xz
+            safe_system "#{HOMEBREW_PREFIX}/opt/xz/bin/xz", "-f", relocatable_tar_path
```
MikeMcQuaid: I'd just use a blanket `xz` here and ensure it was added to the `PATH` previously if needed.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -239,10 +246,15 @@ def bottle_formula(f)
...
+            tar_suffix = compression_type.to_s
```
MikeMcQuaid: I'd consider just making this `xz` for readability.

Library/Homebrew/software_spec.rb
```diff
@@ -244,17 +244,23 @@ def verify_download_integrity(_fn)
...
 
-    def initialize(name, version, tag, rebuild)
+    def initialize(name, version, tag, rebuild, compression_type = :gzip)
       @name = name
       @version = version
       @tag = tag.to_s.gsub(/_or_later$/, "")
       @rebuild = rebuild
+      @compression_suffix = case compression_type
+      when :xz
+        compression_type.to_s
```
MikeMcQuaid: Again just using `xz` here might be more readable.

Library/Homebrew/utils/bottles.rb
```diff
@@ -25,20 +25,20 @@ def file_outdated?(f, file)
       end
 
       def native_regex
-        /(\.#{Regexp.escape(tag.to_s)}\.bottle\.(\d+\.)?tar\.gz)$/o
+        /(\.#{Regexp.escape(tag.to_s)}\.bottle\.(\d+\.)?tar\.(?:gz|xz))$/o
       end
 
       def receipt_path(bottle_file)
-        path = Utils.popen_read("tar", "-tzf", bottle_file).lines.map(&:chomp).find do |line|
```
MikeMcQuaid: Please keep using `Utils.popen_read` instead of backticks in this file.

Library/Homebrew/utils/bottles.rb
```diff
@@ -59,10 +59,38 @@ def formula_contents(bottle_file,
...
         raise BottleFormulaUnavailableError.new(bottle_file, formula_path) unless $CHILD_STATUS.success?
         contents
       end
+
+      def install_dependencies(compression_type, ignore_tar: false)
+        return unless compression_type == :xz
+        return unless ignore_tar || DependencyCollector.tar_needs_xz_dependency?
+        f = Formula["xz"]
+        return if f.optlinked?
+        fi = FormulaInstaller.new(f)
```
MikeMcQuaid: I don't think `FormulaInstaller` should be invoked here. Where this happens as part of a `brew install` it should be installed as a dependency (and could use `FormulaInstaller` if necessary). Otherwise it should error out and request the user run `brew install xz` manually themselves. We should be optimising primarily for the macOS support versions case.

Library/Homebrew/utils/bottles.rb
```diff
@@ -59,10 +59,38 @@ def formula_contents(bottle_file,
...
+        fi.prelude
+        oh1 "Installing bottle dependency: #{Formatter.identifier(f.name)}"
+        fi.install
+        fi.finish
+      end
+
+      private
+
+      def xzpath
+        "#{HOMEBREW_PREFIX}/opt/xz/bin/xz"
```
MikeMcQuaid: I'd just use a blanket xz here and ensure it was added to the PATH previously if needed.

MikeMcQuaid:
> Not that it's a major use-case these days given Homebrew only bottles for three versions of macOS, but it might be worth considering for custom taps/etc? Unsure 🤷‍♂️.

Yeh, I think it's worth punting pretty hard on this and requesting the manual `brew install` when needed rather than autoinstalling.

> Don’t worry too much about this, we can make a second pass later to implement this.

Agreed.

> That's problematic for an integration test. Needing to test that the bottle command is able to install and use a working xz binary doesn't seem like something that should/can be mocked.

That integration testing can/will happen in `brew test-bot` on every build instead. It won't get added to code coverage but that would be a good thing to add at some point in another PR.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -187,11 +190,15 @@ def bottle_formula(f)
       rebuild = rebuilds.empty? ? 0 : rebuilds.max.to_i + 1
     end
 
-    filename = Bottle::Filename.create(f, Utils::Bottles.tag, rebuild)
+    compression_type = :xz
+    compression_type = :gzip if ARGV.include?("--use-gzip") || f.name == "xz"
```
DomT4: I don't know if it's worth being explicit about this? Such as throwing an `ohai "Using gzip rather than xz to bottle xz"` or something. This isn't really public-facing code in the sense that users don't routinely use it locally, but it feels a bit "eh" to make folks read the code to understand what's going on.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -306,6 +318,7 @@ def bottle_formula(f)
       bottle.prefix prefix
     end
     bottle.rebuild rebuild
+    bottle.compression_type compression_type
```
DomT4: I have a minor question here. It's reasonably common when new major versions of macOS arrive to mass bottle `homebrew/core` but _not_ replace the old bottles at the same time, rather simply chucking a new one onto the pile. What happens with this there? Because you'll inevitably for quite some time end up with a mixture of `gz` and `xz` bottles. You wouldn't want to have the `compression_type` set to `xz` for all the bottles in that situation, presumably?

Library/Homebrew/dev-cmd/bottle.rb
MikeMcQuaid: @DomT4 The output spits out the filename already:
`Bottling wget-1.19.2_1.high_sierra.bottle.tar.gz...`

DomT4: @MikeMcQuaid Aye, I know. Just wondering if that's explicit enough or whether the concept that `xz` shouldn't be bottled using `xz` needs some kind of brief further elaboration.

Library/Homebrew/dev-cmd/bottle.rb
MikeMcQuaid: https://github.com/kabel/brew/blob/142204dc4e7477f6b5500971982b49aa0c40ce7d/Library/Homebrew/dev-cmd/bottle.rb#L327 should be tweaked to check for mismatches here

Library/Homebrew/utils/bottles.rb
```diff
@@ -25,20 +25,20 @@ def file_outdated?(f, file)
...
       def native_regex
-        /(\.#{Regexp.escape(tag.to_s)}\.bottle\.(\d+\.)?tar\.gz)$/o
+        /(\.#{Regexp.escape(tag.to_s)}\.bottle\.(\d+\.)?tar\.(?:gz|xz))$/o
       end
 
       def receipt_path(bottle_file)
-        path = Utils.popen_read("tar", "-tzf", bottle_file).lines.map(&:chomp).find do |line|
-          line =~ %r{.+/.+/INSTALL_RECEIPT.json}
-        end
+        command = tar_command "tf", bottle_file, "*/*/INSTALL_RECEIPT.json"
```
sjackman: `*/*/INSTALL_RECEIPT.json` requires the option `--wildcards` with GNU tar on Linux. I'd prefer to keep the previous code for compatibility with both BSD tar and GNU tar.
```
tar: Pattern matching characters used in file names
tar: Use --wildcards to enable pattern matching, or --no-wildcards to suppress this warning
tar: */*/INSTALL_RECEIPT.json: Not found in archive
```

kabel: Which version of tar and on which host? My Fedora 27 has 1.29 and returns as expected.

sjackman: GNU tar 1.30 on Linuxbrew. For example…
```sh
❯❯❯ brew install gnu-tar
❯❯❯ brew fetch hello
❯❯❯ gtar -tf ~/Library/Caches/Homebrew/hello-2.10.high_sierra.bottle.tar.gz '*/*/INSTALL_RECEIPT.json'
gtar: Pattern matching characters used in file names
gtar: Use --wildcards to enable pattern matching, or --no-wildcards to suppress this warning
gtar: */*/INSTALL_RECEIPT.json: Not found in archive
gtar: Exiting with failure status due to previous errors
```

MikeMcQuaid:
> XZ has even better compression ratios and is not too much slower that gzip for decompression.

@kabel Would you mind attempting a benchmark in Travis CI's Mac workers of the time to download and pour a boost bottle with gz vs. xz? I'm interested whether the bandwidth savings outweigh the decompression increases.

kabel:
@MikeMcQuaid Any specifics on how you might like to see that? A new test in the test-bot repo? I'll need to find a place to host the temporary bottle(s) to accurately test the download time difference.

MikeMcQuaid:
@kabel I can upload a bottle you specify to Bintray temporarily. You can do the test as a one-off in a CI job here.

kabel:
Got a benchmark command added now. Interesting results from the [log](https://travis-ci.org/Homebrew/brew/jobs/336837987):
```
                 user     system      total        real
xz:fetch     0.000000   0.000000   0.350000 (  0.847190)
xz:stage     0.640000   1.020000   9.760000 ( 10.753786)
gzip:fetch   0.000000   0.010000   0.690000 (  1.435669)
gzip:stage   0.790000   1.020000   6.550000 (  7.556624)
>xz          0.640000   1.020000  10.110000 ( 11.600976)
>gzip        0.790000   1.030000   7.240000 (  8.992293)
               xz    gzip    diff  diff%
Filesize   41.7MB  84.6MB  42.9MB   50.7
Real                       -2.609  -29.0
```

The fetch results seems a bit odd, compared to my local results.
```
                 user     system      total        real
xz:fetch     0.000000   0.000000   2.180000 ( 13.263047)
xz:stage     0.520000   1.610000   8.580000 (  8.781193)
gzip:fetch   0.000000   0.010000   4.510000 ( 18.964659)
gzip:stage   0.690000   1.650000   6.000000 (  6.118618)
>xz          0.520000   1.610000  10.760000 ( 22.044240)
>gzip        0.690000   1.660000  10.510000 ( 25.083276)
               xz    gzip    diff  diff%
Filesize   41.7MB  84.6MB  42.9MB   50.7
Real                        3.039   12.1
```

Maybe the CI infrastructure is just fast/close. In any case, the time difference of ~3 seconds total feels like a good trade-off considering we're saving ~43MB in network traffic (would take ~5 sec additional on an average US broadband connection at 64 Mbps). In the best case (as seen locally), xz's time savings from network make it smaller _and_ faster than gzip.

ilovezfs:
On the other hand, if you already have all the bottles you're working with cached, it's all downside from that point forward in terms of performance.

kabel:
> it's all downside from that point forward in terms of performance.

That seems like a pessimistic view of it. To get the bottles cached, they would have had to be downloaded at some point. Then, the savings would just be disk space; not a huge sell, but not a downside.

ilovezfs:
Yup. I look at those numbers, see 50% longer stage times going forward, and begin to weep.

kabel:
50% is a bit of an exaggeration and more of a worst case scenario (more analysis could be done). I don't weep over bandwidth/disk savings when we're talking about about a few seconds slower to decompress. We live in a world where bandwidth will continue to be a premium and access to quality broadband is difficult. YMMV.

kabel:
Ha. I guess that's more of a foot-in-mouth moment for me. The deeper analysis from https://github.com/Homebrew/brew-evolution/pull/17#issuecomment-252432739 shows increases too. We know it's not a win-win. So it comes down to what we think will be of more value to the community.

MikeMcQuaid:
I think we're going to pass on this, sorry. As a summary:
- compression is much slower and requires additional dependencies
- decompression is slower and requires additional dependencies on some platforms
- download is faster

Given the status-quo is working fine the above reasons wouldn't necessarily be a slam-dunk if we were deciding the initial compression format to use for bottles but considering the effort required to rebottle things and the returns being less than expected (and in some cases negative) it's not worth the investment at this time. I'm really sorry for this @kabel and, genuinely, thanks for all your hard work here.

kabel:
Thanks @MikeMcQuaid for the opportunity to present this idea and learn more about the system. In the interest of clarity in summary, could I get feedback on the following questions:

1. Is there harm in providing support for this bottle compression format without it being immediately used by the core tap? 

Both common Linux binary package formats (deb and rpm) support compression in a variety of formats, giving repository (tap) maintainers flexibility in how they approach CPU, network, and storage optimization.

The question is in other words, would _separating_ the decision to bottle (core) with xz by default and the decision to support bottles with xz have a different outcome?

2. Are there quantitative/objective measures used in terms of determining the trade-off in performance (time/space)?

Without hard numbers it becomes very subjective for someone to analyze cost/benefit. If the overall goal of bottles is to reduce CPU utilization on the end-user side, can a secondary goal like reducing end-user network utilization be introduced that cuts into the former? A 50.7% savings in space and the time it would take to transfer that space over the internet (30-40.9% depending on ISP speed/location), was not worth the additional CI CPU load and 49% increase in end-user time to stage. The percentages mask some of the marketable net-time improvement many end-users would have likely seen. Clearly there are some strong opinions here.

It seems like a mathematical formula could have been used to determine if the savings for a particular formula was worth the trade-off and the winning compression format would win for each specific case.

With respect to
> requires additional dependencies
3. Dependencies are only _necessary_ for older OS versions. The bottling process could be further refactored to eliminate the need for using gzip/xz outside of BSD tar. Do end-of-life OS versions hold back future enhancement?

> considering the effort required to rebottle things [...] not worth the investment at this time

4. As a community member with nothing but ideas and time to contribute, is there anything that I can do (research, automation, etc.) to alleviate some of that burden?

ilovezfs:
I don't think we should add support for additional compression formats if it's not going to be used in core as that increases code complexity unnecessarily.

DomT4:
> We live in a world where bandwidth will continue to be a premium and access to quality broadband is difficult.

Is there anything appreciable to be gained turning the `gzip` compression up from its default 6 to `--best`, as something of a middle ground?

ilovezfs:
Reduced performance.

ilovezfs:
Not worth it.

sjackman:
The changes in file size are surprisingly small between `gzip -6` and `--best`.

ilovezfs:
Not surprising, and if anything, those stats point toward using a lower gzip number if we were to make any change there.

MikeMcQuaid:
> Is there harm in providing support for this bottle compression format without it being immediately used by the core tap?

Yes. It's more code that needs to be maintained, tested and updated that I want to avoid if we don't have a guaranteed major user on day 1.

> The percentages mask some of the marketable net-time improvement many end-users would have likely seen.

The point is: this change would be a compromise that maybe weights towards an overall improvement but it's not a massive win for all users. As a result, I don't think it justifies the complexity of the change.

> Do end-of-life OS versions hold back future enhancement?

No. Impact on them is considered as part of enhancements, though. Even if every macOS version back to 10.1 included `xz` I'd still not include this.

> As a community member with nothing but ideas and time to contribute, is there anything that I can do (research, automation, etc.) to alleviate some of that burden?

To be clear: I'm really grateful you created this PR and performed all the tests we requested. That has saved us a bunch of time and our hats are tipped to you for it. In general: these are the bigger things we're likely to accept PRs for: https://github.com/homebrew/brew/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

Similarly, your help on https://github.com/Homebrew/homebrew-php/issues/4721 would be fantastic.

Issue #3625, opened 2018-01-03
Bash completion for "brew cask" does not work


rhwood:
Useful arguments for brew commands except for `brew cask` can be automatically completed in bash.

# Please always follow these steps:
- [x] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at the relevant tap e.g. for Homebrew/homebrew-core https://github.com/Homebrew/homebrew-core/issues/new
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed all issues and retried your prior step?
- [x] Ran `brew config` and `brew doctor` and included their output with your issue?

To help us debug your issue please explain:
- What you were trying to do (and why)
  Use tab completion with `brew cask`
- What happened (include command output)
  `brew cask<tab><tab>` listed the contents of the current directory as useful completions
- What you expected to happen
   `brew cask<tab><tab>` to list usable commands
- Step-by-step reproduction instructions (by running `brew` commands)
   Type `brew cask<tab><tab>`

**Note:** installing the formula `brew-cask-completion` seems to make no difference, but I don't think I should need to install that formula anyways--this should just be part of brew.

# Features
Please replace this section with:
- a detailed description of your proposed feature
- the motivation for the feature
- how the feature would be relevant to at least 90% of Homebrew users
- what alternatives to the feature you have considered

We will close this issue or ask you to create a pull-request if it's something we're not actively planning to work on.

```
Randalls-MacBook-Pro:~ rwood$ brew update
Updated 1 tap (homebrew/core).
==> Updated Formulae
macvim
Randalls-MacBook-Pro:~ rwood$ brew config
HOMEBREW_VERSION: 1.4.2
ORIGIN: https://github.com/Homebrew/brew
HEAD: 4e08da95c34c27ba6d5ca5f80718fe2788337cdf
Last commit: 2 days ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: a7bb7a89db59ab518dcf93f849473ec9fba3f6d7
Core tap last commit: 27 minutes ago
HOMEBREW_PREFIX: /usr/local
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.3.3 => /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/bin/ruby
Clang: 9.0 build 900
[...]

PR #3647, opened 2018-01-09
bash completion: add null completion where appropriate

apjanke:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). *Nope - we don't test shell completion.*

-----

In some cases, in bash completion, there are no valid arguments discoverable from the command line. For example, with `brew search <tab><tab>`, it could take any string. (And, if #3625 is merged, `brew cask doctor <tab><tab>` has no valid completions.) In this case, I think we should supply no completions.

The current behavior is to fall back to completing on file names relative to the current directory. I think this is wrong, because those file names are not valid arguments in this context. Or at least, they're not expected to be relevant.

This PR changes the behavior to have it complete to nothing. (Unfortunately, adding a blank space on the command line, but I don't see a way around that.) I think that's more appropriate.

### DIFF
--- a/completions/bash/brew
+++ b/completions/bash/brew
@@ -1,5 +1,10 @@
 # Bash completion script for brew(1)
 
+# Indicates there are no completions
+__brewcomp_null() {
+  COMPREPLY=""
+}
+
 __brewcomp_words_include() {
   local i=1
   while [[ "$i" -lt "$COMP_CWORD" ]]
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @apjanke. Does the zsh completion need a similar check?

apjanke:
No prob!

> Does the zsh completion need a similar check?

No. In the same cases, our existing zsh completion reaches a no-completions state, and doesn't even add an extra space.

PR #3654, opened 2018-01-09
check_install_sanity: Use formula.runtime_dependencies

sjackman:

-----

Fix…
```
$ brew install wget
Error: wget contains conflicting version recursive dependencies:
openssl, openssl@1.1
View these with `brew deps --tree wget`.
$ brew deps --tree wget
wget
├── libidn2
│ └── libunistring
├── openssl@1.1
│ └── perl
│ ├── gdbm
│ ├── berkeley-db
│ └── expat
│ └── libbsd
└── util-linux
└── ncurses
```

On Linux, `libidn2` has no linkage to `gettext` and so is a `:build` dependency rather than a run-time dependency. `libintl` is a header-only and provided by `glibc` rather than by `gettext`, as it is on macOS.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -149,7 +149,7 @@ class FormulaInstaller
 
     recursive_deps = formula.recursive_dependencies
     recursive_formulae = recursive_deps.map(&:to_formula)
-    recursive_runtime_deps = formula.recursive_dependencies.reject(&:build?)
+    recursive_runtime_deps = formula.runtime_dependencies
     recursive_runtime_formulae = recursive_runtime_deps.map(&:to_formula)
 
     recursive_dependencies = []


MikeMcQuaid:
I don't agree with this. The goal of this check is to avoid multiple versions of the same library being pulled in during an installation which is almost always a sign of bad packaging.

sjackman:
Multiple versions are not being pulled in. In the dependency tree above, there's only `openssl@1.1` and no `openssl`.

MikeMcQuaid:
Can you explain why the existing code is incorrect in this case?

MikeMcQuaid:
i.e. why are multiple non-build dependencies for openssl appearing in recursive_dependencies?

sjackman:
The current code
```ruby
formula.recursive_dependencies.reject(&:build?)
```
doesn't prune the recursive dependencies of a build dependency.
```ruby
formula.runtime_dependencies
```
does prune the recursive dependencies of a build dependency.
See https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L1490

MikeMcQuaid:
Cool. Objection retracted, 🤐 attached to mouth. Thanks again @sjackman.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

ilovezfs:
In some circumstances, this has broken the check. If I make the `node` `--with-openssl@1.1` option the default, it is still possible to install emscripten despite the recursive dependency on both openssl and openssl@1.1. `git bisect` led me to 95bf23404ea39a8847b56612c3fef44497daf893.

ilovezfs:
`runtime_dependencies` seems to be first level (`deps -1`) only
```
irb(main):004:0> :emscripten.f.recursive_dependencies
=> [#<Dependency: "pkg-config" [:build]>, #<Dependency: "sphinx-doc" [:build]>, #<Dependency: "gdbm" []>, #<Dependency: "makedepend" [:build]>, #<Dependency: "openssl" []>, #<Dependency: "readline" []>, #<Dependency: "sqlite" []>, #<Dependency: "python@2" []>, #<Dependency: "cmake" [:build]>, #<Dependency: "icu4c" [:recommended]>, #<Dependency: "openssl@1.1" [:recommended]>, #<Dependency: "node" []>, #<Dependency: "yuicompressor" []>]
irb(main):005:0> :emscripten.f.runtime_dependencies
=> [#<Dependency: "gdbm" []>, #<Dependency: "openssl" []>, #<Dependency: "readline" []>, #<Dependency: "sqlite" []>, #<Dependency: "python@2" []>, #<Dependency: "node" []>, #<Dependency: "yuicompressor" []>]
irb(main):006:0> 
```

ilovezfs:
hmm no there's more to it than that. `openssl` is still indirect.

ilovezfs:
OK the issue is that `runtime_dependencies` ignores all optional and recommended dependencies, and only looks at required dependencies. 🐛

ilovezfs:
CC @sjackman

ilovezfs:
I've opened https://github.com/Homebrew/brew/issues/4124

PR #3656, opened 2018-01-10
enhance write_jar_script to launch specified java version

ryenus:

-----

Certain java apps require a specific java version, meanwhile the current java version from `$PATH` could be another incompatible version,  e.g. one might have installed both java 8 and java 9,
 meanwhile [**Jenkins**][1] only works with java 8.

[1]: https://issues.jenkins-ci.org/browse/JENKINS-40689

This change improves the `bin.write_jar_script`, so that it takes one more optional argument `java_version`, once specified, it use `/usr/libexec/java_home` to launch the specified java version.

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -436,11 +436,12 @@ class Pathname
   end
 
   # Writes an exec script that invokes a java jar
-  def write_jar_script(target_jar, script_name, java_opts = "")
+  def write_jar_script(target_jar, script_name, java_opts = "", java_version = "")
     mkpath
+    launcher = java_version.empty? ? "java" : "/usr/libexec/java_home -v #{java_version} --exec java"
     join(script_name).write <<~EOS
       #!/bin/bash
-      exec java #{java_opts} -jar #{target_jar} "$@"
+      exec #{launcher} #{java_opts} -jar #{target_jar} "$@"
     EOS
   end
 


DomT4:
I think you're essentially describing:
```ruby
 bin.env_script_all_files(libexec/"bin", Language::Java.java_home_env("1.8"))
```
no?

ryenus:
@DomT4 thanks for helping review! 

My answer would be yes and no. but I might be wrong, and there is a case I'd discuss with you.
I checked `env_script_all_files`, so it relies on `write_env_script`, my understanding is that it would generate a script file like: 

	JAVA_HOME=/path/to/java java ...

This surely would work, but only when `/path/to/java` used upon installation still exists when the generated script is used later.  When a new java version is installed, the `/path/to/java` could change, then the generated script would fail.

However, using `/usr/libexec/java_home -v #{java_version} --exec java` it can always look for the current java installation (of the specified version) and use it.

If possibly the generated script is something like:

	JAVA_HOME=$(/usr/libexec/java_home #{ava_version}) java ...

Then, just as you proposed, we can use: 

	bin.env_script_all_files(libexec/"bin", Language::Java.java_home_env("1.8"))

But I'm not sure about what would be translated from `Language::Java.java_home_env("1.8")`, first case? no no, 2nd case? fine, but it looks no better than using `/usr/libexec/java --exec` IMHO.

MikeMcQuaid:
Can you provide an example of the formulae in which you were planning on using this? Thanks!

DomT4:
> However, using /usr/libexec/java_home -v #{java_version} --exec java it can always look for the current java installation (of the specified version) and use it.

This is what it does, yeah. Example:

```bash
#!/bin/bash
JAVA_HOME="$(/usr/libexec/java_home --version 1.8)" exec "/usr/local/Cellar/ringojs/1.2.1/libexec/bin/ringo" "$@"
```

Which if you don't have that version locally errors out with `Unable to find any JVMs matching version "1.8".`

ilovezfs:
If I understand this PR, it's yet-another-way (I'm not really sure we actually need yet-another-way) to do this: https://github.com/Homebrew/homebrew-core/commit/39e0b03cdc6d8001aeed4881310faed7c7b75916

alyssais:
Having a utility method for this rather than having to copy/paste between formulae seems useful to me?

ilovezfs:
I don't particularly like any of our language DSLs so am not eager to add more "features" to them, but :man_shrugging:

DomT4:
I like language/etc shorthand more than ILZ does but I'd agree with him here. I'm struggling to see that it's worth the effort here specifically.

ryenus:
@DomT4 and all who expressed your thoughts, thanks!

I agree that `JAVA_HOME="$(/usr/libexec/java_home #{ava_version})" java` would work in most cases, usually, the following would report a version of java `1.8`:

	JAVA_HOME="$(/usr/libexec/java_home -v 1.8)" java -version

However, when `$PATH` provides another `java` which is different from `/System/Library/Frameworks/JavaVM.framework/Versions/A/Commands/java`. it might work differently:

	export PATH="$(/usr/libexec/java_home -v 1.8)/bin:$PATH"

	JAVA_HOME="$(/usr/libexec/java_home -v 9)" java -version

In this case it would report a version of java `1.8`, rather than java ~~`9`~~:

```
$ PATH="$(/usr/libexec/java_home -v 1.8)/bin:$PATH" JAVA_HOME="$(/usr/libexec/java_home -v 9)" java -version
java version "1.8.0_152"
Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.152-b16, mixed mode)
```

Meanwhile `/usr/libexec/java_home -v 9 --exec java -version` works as expected:

```
$ PATH="$(/usr/libexec/java_home -v 1.8)/bin:$PATH" /usr/libexec/java_home -v 9 --exec java -version
java version "9.0.1"
Java(TM) SE Runtime Environment (build 9.0.1+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.1+11, mixed mode)
```

This is the case where the two are not always the same and the proposed path is actually safer.

DomT4:
I'm not saying there's absolutely no benefit here or with a mechanism similar to this, and certainly don't want to give off that impression.

I guess my point is more that very few people have ever really complained our current two methods for handling this are insufficient or problematic, and I don't know if a problem users aren't vocally having is really a problem... as such.

I don't know. I'm gonna hush my mouth & leave this one down to ILZ/Mike/Alyssa.

ryenus:
@DomT4 no worries, I sincerely thank you for helping me understand the current offerings, hence to figure out the gap, now we have a more clear justification for the change, which is not possible without your help, thanks!

@MikeMcQuaid, thanks! one example is my related Jenkins PR: https://github.com/Homebrew/homebrew-core/pull/22671

And @ilovezfs could you please review again? appreciated!

Library/Homebrew/extend/pathname.rb
```diff
@@ -436,11 +436,12 @@ def env_script_all_files(dst, env)
   end
 
   # Writes an exec script that invokes a java jar
-  def write_jar_script(target_jar, script_name, java_opts = "")
+  def write_jar_script(target_jar, script_name, java_opts = "", java_version = "")
```
MikeMcQuaid: `java_version: nil` and then you can specify it like this: `bin.write_jar_script libexec/"jenkins-cli.jar", "jenkins-lts-cli", java_version: "1.8"` rather than needing an empty `java_opts` argument.

Library/Homebrew/extend/pathname.rb
```diff
@@ -436,11 +436,12 @@ def env_script_all_files(dst, env)
...
     mkpath
+    launcher = java_version.empty? ? "java" : "/usr/libexec/java_home -v #{java_version} --exec java"
```
MikeMcQuaid:
```ruby
java = if java_version
  "/usr/libexec/java_home -v #{java_version} --exec java"
else
  "java"
end
```

Library/Homebrew/extend/pathname.rb
```diff
@@ -436,11 +436,12 @@ def env_script_all_files(dst, env)
...
     join(script_name).write <<~EOS
       #!/bin/bash
-      exec java #{java_opts} -jar #{target_jar} "$@"
+      exec #{launcher} #{java_opts} -jar #{target_jar} "$@"
```
MikeMcQuaid: `exec "#{java}"` as it may have spaces.

Library/Homebrew/extend/pathname.rb
MikeMcQuaid: After speaking to @ilovezfs instead let's do something like:

```ruby
java_home = if java_version
  <<~EOS
    export JAVA_HOME="$(#{Language::Java.java_home_cmd(java_version)})"

  EOS
end
```

and below:

```bash
       #!/bin/bash
       #{java_home}exec java #{java_opts} -jar #{target_jar} "$@"
```

This will be consistent with what `bin.env_script_all_files(libexec/"bin", Language::Java.java_home_env("1.8"))` ends up doing.

Library/Homebrew/extend/pathname.rb
ryenus: right, this is indeed better, thanks!

Library/Homebrew/extend/pathname.rb
ryenus: If my understanding is correct, this only specifies `JAVA_HOME`, hence still relies on an unpolluted `$PATH`, if `$PATH` provides another `/usr/local/java-1.7/bin/java`, setting `JAVA_HOME` only no longer has the magic, and will launch java 1.7 instead, as I presented in my comment above: https://github.com/Homebrew/brew/pull/3656#issuecomment-359259313

MikeMcQuaid: We'd rather be consistent between approaches. Let's go for this one in both places for now and then if desired you can submit an update for both later.

PR #3657, opened 2018-01-10
Fix cached download file extension for certain URLs

kabel:

-----
PHP URLs have the downloadable file in the middle of the pathname. If no extension is detected, continue up the pathname. Ex: `https://php.net/get/php-7.2.1.tar.xz/from/this/mirror` should cache to `php-7.2.1.tar.xz`.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -306,7 +306,7 @@ class AbstractFileDownloadStrategy < AbstractDownloadStrategy
     url_pathname = Pathname.new(@url)
     until ext = url_pathname.extname[/[^?]+/]
       url_pathname = url_pathname.dirname
-      return if url_pathname.to_s == "."
+      return if url_pathname.to_s == "." || url_pathname.to_s == "/"
     end
     ext
   end
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -306,6 +306,7 @@ class AbstractFileDownloadStrategy < AbstractDownloadStrategy
     url_pathname = Pathname.new(@url)
     until ext = url_pathname.extname[/[^?]+/]
       url_pathname = url_pathname.dirname
+      return if url_pathname.to_s == "."
     end
     ext
   end
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -303,12 +303,11 @@ class AbstractFileDownloadStrategy < AbstractDownloadStrategy
     # We can't use basename_without_params, because given a URL like
     #   https://example.com/download.php?file=foo-1.0.tar.gz
     # the extension we want is ".tar.gz", not ".php".
-    url_pathname = Pathname.new(@url)
-    until ext = url_pathname.extname[/[^?]+/]
-      url_pathname = url_pathname.dirname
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
> PHP URLs have the downloadable file in the middle of the pathname. If no extension is detected, continue up the pathname. Ex: https://php.net/get/php-7.2.1.tar.xz/from/this/mirror should cache to php-7.2.1.tar.xz.

My concern is this is a widely generic change that will affect the behaviour of a bunch of other formulae in a way that may not be correct. The downloaded filename is based off the formula and version and not the upstream name.

kabel:
Right, I should have used a better example. I can certainly change it to make it fix the very specific case of php's file suffix, but I don't think it will change any other formulae that don't also exhibit the same problem. That is, tarballs in the cache directory that don't have any extension. I could also crawl through hombrew-core to get a better representation of downloaded resources and put in more tests.

Library/Homebrew/download_strategy.rb
```diff
@@ -303,7 +303,11 @@ def ext
     # We can't use basename_without_params, because given a URL like
     #   https://example.com/download.php?file=foo-1.0.tar.gz
     # the extension we want is ".tar.gz", not ".php".
-    Pathname.new(@url).extname[/[^?]+/]
+    url_pathname = Pathname.new(@url)
+    until ext = url_pathname.extname[/[^?]+/]
```
MikeMcQuaid: This should probably handle the case where the extname never matches and/or `dirname` fails.

MikeMcQuaid:
> I can certainly change it to make it fix the very specific case of php's file suffix, but I don't think it will change any other formulae that don't also exhibit the same problem. That is, tarballs in the cache directory that don't have any extension.

Ok. I think it's fine then if better error handling is added.

Library/Homebrew/download_strategy.rb
MikeMcQuaid: @kabel Ping on this

kabel: I addressed this in https://github.com/Homebrew/brew/pull/3657/commits/0254605cd52b1bb92ffe33114c671cc111c5a687 by checking the return value of dirname.

MikeMcQuaid: @kabel That's not sufficient. For example, if `url_pathname` ends up as `/` then this will be an infinite loop. I think it's worth changing the approach so this loop can never be infinite.

kabel: I did test it out through irb. I could add the root `/` as another escape condition, but these are going to be URLs that we're dealing with. So the top-level domain would (in most cases where it going that far) be returned as the extension. If the URL didn't have a domain (`file:///`), dirname would still default to `.`. I guess it can't hurt to be even more robust.

MikeMcQuaid: @kabel I'm sorry if I'm not being explicit enough but I'm just not happy with a change that adds an infinite loop if the return values are ever something you do not expect. Instead of making this loop indefinitely until there's an escape condition instead consider splitting on `/` and iterating over that or another approach that makes this a bounded loop.

kabel:
@MikeMcQuaid I understand the displeasure with the potential unbounded loop. Should be better now with `.ascend`.

MikeMcQuaid:
Nice find on `ascend` @kabel (TIL) and thanks again for the PR.

PR #3666, opened 2018-01-11
# ADD zh_TW.UTF-8 support.

ralic:
Before Change :
```
 bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
```
After Change : 
```
No more warning
```
- [v ] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?

- [v ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?

- [v ] Have you added an explanation of what your changes do and why you'd like us to include them?
## To add zh_TW.UTF-8 support and remove annoying warning from bash

- [x ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).

- [v ] Have you successfully run `brew tests` with your changes locally?

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -3,7 +3,7 @@ case "$(locale | grep LANG= | sed 's/LANG=//')" in
   *UTF-8*)
     LC_ALL=$(locale | grep LANG= | sed 's/LANG=//')
     ;;
-  default)
+  *)
 	LC_ALL="en_US.UTF-8"
 	;;
 esac
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -1,8 +1,13 @@
 # Force UTF-8 to avoid encoding issues for users with broken locale settings.
-if [[ "$(locale charmap 2>/dev/null)" != "UTF-8" ]]
-then
-  export LC_ALL="en_US.UTF-8"
-fi
+case "$(locale | grep LANG= | sed 's/LANG=//')" in 
+  *UTF-8*)
+    LC_ALL=$(locale | grep LANG= | sed 's/LANG=//')
+    ;;
+  default)
+	LC_ALL="en_US.UTF-8"
+	;;
+esac
+
 
 # Where we store built products; a Cellar in HOMEBREW_PREFIX (often /usr/local
 # for bottles) unless there's already a Cellar in HOMEBREW_REPOSITORY.


Library/Homebrew/brew.sh
```diff
@@ -1,8 +1,13 @@
 # Force UTF-8 to avoid encoding issues for users with broken locale settings.
-if [[ "$(locale charmap 2>/dev/null)" != "UTF-8" ]]
```
MikeMcQuaid: What does `locale charmap` output for `zh_TW.UTF-8`?

ralic: It shows UTF-8. However the warning will still be there, because LC_ALL was set to en_US.UTF-8. Here is the locale details.


```
=>locale
LANG=zh_TW.UTF-8
LANGUAGE=zh_TW:zh
LC_CTYPE="zh_TW.UTF-8"
LC_NUMERIC="zh_TW.UTF-8"
LC_TIME="zh_TW.UTF-8"
LC_COLLATE="zh_TW.UTF-8"
LC_MONETARY="zh_TW.UTF-8"
LC_MESSAGES="zh_TW.UTF-8"
LC_PAPER="zh_TW.UTF-8"
LC_NAME="zh_TW.UTF-8"
LC_ADDRESS="zh_TW.UTF-8"
LC_TELEPHONE="zh_TW.UTF-8"
LC_MEASUREMENT="zh_TW.UTF-8"
LC_IDENTIFICATION="zh_TW.UTF-8"
LC_ALL=zh_TW.UTF-8
```

apjanke:
Are you getting this warning on macOS or Linux?

I suspect this doesn't happen on macOS because the `en_US` locale is installed (at least by default) on all macOS systems, including non-US ones. And also because it doesn't seem to issue a warning when setting the locale to a noninstalled locale via bash. Instead, it silently falls back to the C locale. (Which is also not good.)

On my Sierra machine:

```
[~] $ locale
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL=
[~] $ export LC_ALL=xx_XX.UTF-8
[~] $ locale
LANG="en_US.UTF-8"
LC_COLLATE="C"
LC_CTYPE="C"
LC_MESSAGES="C"
LC_MONETARY="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_ALL="C"
[~] $
```

Does your machine have the `en_US.UTF-8` locale installed on it?

Library/Homebrew/brew.sh
apjanke: Was your locale set to `zh_TW.UTF-8` before running the `brew` bash script? If it was `zh_TW.UTF-8` originally, then the `export LC_ALL="en_US.UTF-8"` shouldn't have happened in the first place.

MikeMcQuaid: As @apjanke said: if `locale charmap` outputs `UTF-8` it'll never set `LC_ALL`.

MikeMcQuaid:
> I am using Debian

Closing as this isn't a macOS bug. Discuss this in https://github.com/Linuxbrew/brew/issues/568.

PR #3679, opened 2018-01-13
Don't auto-tap caskroom/cask in ZSH completion

### LINKED ISSUE #3677: The shell completion does not work well when cask repo is not cloned.

- [x] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at the relevant tap e.g. for Homebrew/homebrew-core https://github.com/Homebrew/homebrew-core/issues/new

- What you expected to happen

# Features

Input `brew cask install ...` and press TAB key, the output gets messed up if cask is not cloned yet.

screen shot 2018-01-13 at 10 39 09 am

It seems that it takes too long for a completion even if the output log if omitted, so I recommend to disable shell completion in this situation.

alyssais:
The Bash completion just uses the directory tree rather than invoking `brew cask`, so doesn't exhibit the same behaviour.

Fixes #3677.

### DIFF
--- a/completions/zsh/_brew_cask
+++ b/completions/zsh/_brew_cask
@@ -12,17 +12,21 @@
 zstyle -T ':completion:*:*:*:brew-cask:*' tag-order && \
 	zstyle ':completion:*:*:*:brew-cask:*' tag-order 'commands'
 
+__brew_cask() {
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Nice work @alyssais. 🚢 when 💚

PR #3684, opened 2018-01-15
Revert "Use Homebrew's gitconfig for various commands"

ilovezfs:
Reverts Homebrew/brew#3670


ilovezfs:
see https://github.com/Homebrew/brew/pull/3670#issuecomment-357602388

PR #3696, opened 2018-01-18
`prepend` instead of alias hacks for deprecations

alyssais:

-----

I proposed this [here](https://github.com/Homebrew/brew/pull/3694#discussion_r162320914).

While I was doing this, I added some tests for the deprecations so I could have confidence that I wasn't breaking anything. Writing the tests also made me realise that compat.rb didn't actually load the formula_support deprecations, so I fixed that too. We probably should test these going forward - I've found it to be pretty simple to do in these two cases.

I also noticed that the deprecation message for e.g. `depends_on ... => :lua` was inaccurate, so I fixed that too.


alyssais:
I recommend `?w=1` to review the diff on this ;)

MikeMcQuaid:
Thanks again @alyssais!

MikeMcQuaid:
@alyssais Could maybe also take the `prepend` approach for `Library/Homebrew/extend` too? What do you think?

Issue #3708, opened 2018-01-20
FEATURE (regression?): `brew doctor` doesn't check permission on /usr/local/man


sesam:
**Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

# Please always follow these steps:
- [X] Confirmed this is a problem with running a `brew` command and not `brew install`ing or the post-install behaviour of one or more formulae? If it's a formulae-specific problem please file this issue at the relevant tap e.g. for Homebrew/homebrew-core https://github.com/Homebrew/homebrew-core/issues/new
- [X] Ran `brew update` and retried your prior step?
- [X] Ran `brew doctor`, fixed all issues and retried your prior step?
- [X] Ran `brew config` and `brew doctor` and included their output with your issue?

To help us debug your issue please explain:
- What you were trying to do (and why)

Tried a simple `brew upgrade`

- What happened (include command output)

Various permission errors
````
Linking /usr/local/Cellar/cmake/3.10.2... Error: Permission denied @ dir_s_mkdir - /usr/local/share/man/man1
````

- What you expected to happen
Expected `brew upgrade` to function normally.

- Step-by-step reproduction instructions (by running `brew` commands)
1. have a Sierra mac, upgrade to High Sierra
2. run `brew upgrade` now and then normally
3. at some unhappy timepoint, it stops working normally, with complaints about failing to add links to /usr/local/share/man/man1 (and other manX and possibly also other places under /usr/local)
4. `brew update` says all is updated.
5. `brew doctor` says nothing useful and does not help
6. Try all ideas on Stackoverflow etc (also finding #3322 which seems prematurely abandoned)
7. cry a bit about maybe having to make a full brew reinstall
8. cross fingers and run the shell command from https://brew.sh/  `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
`

VOILA! That install script says:
````
==> The following existing directories will be made group writable:
/usr/local/share/man

Press RETURN to continue or any other key to abort
````

So why didn't `brew doctor` suggest or fix this?

# Features

Please, please, make `brew doctor` check all things that the brew install script checks, including permissions on `/usr/local/share/man`

sesam:
$ brew config # (config after things are ok)
````
HOMEBREW_VERSION: 1.5.0-6-gd14fd49
[...]

Issue #3228, opened 2017-09-27
Can't chown /usr/local in High Sierra


dtamayo:
I am running OS X High Sierra and need to change permissions for /usr/local, but it won't let me. I first ran brew update, and after fixing issues, brew doctor just gives:

Warning: You have unlinked kegs in your Cellar
Leaving kegs unlinked can lead to build-trouble and cause brews that depend on
those kegs to fail to run properly once built. Run `brew link` on these:
  gdbm
  python
  libyaml
  ruby

brew link ruby gives:

Linking /usr/local/Cellar/ruby/2.4.2... 
Error: Could not symlink include/ruby-2.4.0
/usr/local/include is not writable.

sudo chown -R $(whoami) $(brew --prefix) gives
chown: /usr/local: Operation not permitted

ls -lO /usr gives
total 0
drwxr-xr-x  975 root  wheel  restricted 31200 26 Sep 21:55 bin
drwxr-xr-x  292 root  wheel  restricted  9344 26 Sep 22:06 lib
drwxr-xr-x  233 root  wheel  restricted  7456 26 Sep 21:55 libexec
drwxr-xr-x   13 root  wheel  sunlnk       416 26 Sep 22:05 local
drwxr-xr-x  246 root  wheel  restricted  7872 26 Sep 21:55 sbin
drwxr-xr-x   46 root  wheel  restricted  1472 26 Sep 21:55 share
drwxr-xr-x    5 root  wheel  restricted   160 21 Sep 00:32 standalone

brew config gives 

HOMEBREW_VERSION: 1.3.4
ORIGIN: https://github.com/Homebrew/brew
HEAD: ac8b58537fe3dd71098a43afcc6a1785dcde1d39
Last commit: 2 days ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 168b65e2b525fa6a881e2480145ab1aa1fce2895
Core tap last commit: 2 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: quad-core 64-bit kabylake
Homebrew Ruby: 2.3.3 => /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/bin/ruby
Clang: 9.0 build 900
[...]

PR #3713, opened 2018-01-21
add check that ALL_TOP_LEVEL_DIRECTORIES exist

sesam:
First PR-half requested in #3708

To prevent messy errors from `brew upgrade` - add check to make sure that ALL_TOP_LEVEL_DIRECTORIES exist.

As noticed in #3228 this is necessary for at least some users for users of High Sierra. Worst case, you get an empty folder you don't need.

Everybody typically installs python, and in case some of these directories aren't necessary to sudo mkdir we can fix that in a later PR.

-----

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -345,6 +345,23 @@ module Homebrew
         EOS
       end
 
+      def check_missing_prefix_directories
+        missing_dirs = []
+        Keg::ALL_TOP_LEVEL_DIRECTORIES.each do |dir|
+          path = HOMEBREW_PREFIX/dir
+          missing_dirs << path unless path.exist?
+        end
+
+        return if missing_dirs.empty?
+
+        <<~EOS
+          The following directories are missing:
+          #{missing_dirs.join("\n")}
+          You should create them:
+            sudo mkdir -p #{missing_dirs.join(" ")}
+        EOS
+      end
+
       def check_access_site_packages
         return unless Language::Python.homebrew_site_packages.exist?
         return if Language::Python.homebrew_site_packages.writable_real?
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -346,6 +346,9 @@ module Homebrew
       end
 
       def check_missing_prefix_directories
+        path = HOMEBREW_PREFIX/('.brew-detect' + Time.now.to_i.to_s)
+        Dir.mkdir(path) and Dir.rmdir(path) # and return nil for all-ok
+      rescue SystemCallError # what mkdir/rmdir throws
         missing_dirs = []
         Keg::ALL_TOP_LEVEL_DIRECTORIES.each do |dir|
           path = HOMEBREW_PREFIX/dir
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -357,7 +357,7 @@ module Homebrew
       def check_existing_prefix_directories_are_searchable
         return if HOMEBREW_PREFIX.to_s != "/usr/local"
 
-        existing_dirs = Keg::ALL_TOP_LEVEL_PATHS.select(&:exist?)
+        existing_dirs = (Keg::ALL_TOP_LEVEL_PATHS + Keg::ALL_SHARE_PATHS).select(&:exist?)
         not_searchable_dirs = existing_dirs.reject(&:executable_real?)
         return if not_searchable_dirs.empty?
 
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -320,15 +320,8 @@ module Homebrew
       end
 
       def check_access_prefix_directories
-        not_writable_dirs = []
-
-        Keg::ALL_TOP_LEVEL_DIRECTORIES.each do |dir|
-          path = HOMEBREW_PREFIX/dir
-          next unless path.exist?
-          next if path.writable_real?
-          not_writable_dirs << path
-        end
-
+        existing_dirs = Keg::ALL_TOP_LEVEL_PATHS.select(&:exist?)
+        not_writable_dirs = existing_dirs.reject(&:writable_real?)
         return if not_writable_dirs.empty?
 
         <<~EOS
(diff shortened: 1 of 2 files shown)


Library/Homebrew/diagnostic.rb
```diff
@@ -345,6 +345,23 @@ def check_access_prefix_directories
         EOS
       end
 
+      def check_missing_prefix_directories
```
MikeMcQuaid: `return if HOMEBREW_PREFIX.to_s != "/usr/local"` as as-is this breaks CI and isn't really needed for the non-`/usr/local` case (as `sudo` wouldn't be needed there).

sesam: Sorry, I don't understand this comment. Maybe pasted on the wrong PR?

The code I added checks that the following paths exist:

````
#<Pathname:/usr/local/bin>
#<Pathname:/usr/local/etc>
#<Pathname:/usr/local/include>
#<Pathname:/usr/local/lib>
#<Pathname:/usr/local/sbin>
#<Pathname:/usr/local/share>
#<Pathname:/usr/local/var>
#<Pathname:/usr/local/Frameworks>
#<Pathname:/usr/local/lib/pkgconfig>
#<Pathname:/usr/local/share/locale>
#<Pathname:/usr/local/share/man>
#<Pathname:/usr/local/opt>
````

File: Library/Homebrew/diagnostic.rb [hunk 2]
```diff
@@ -345,6 +345,26 @@ def check_access_prefix_directories
...
+          missing_dirs << path unless path.exist?
+        end
+
+        return if missing_dirs.empty?
+
+        <<~EOS
+          The following directories are missing:
+          #{missing_dirs.join("\n")}
+          You should create them:
+            sudo mkdir -p #{missing_dirs.join(" ")}
```
MikeMcQuaid: This will also need a `chown` immediately afterwards.

Library/Homebrew/diagnostic.rb
```diff
@@ -345,6 +345,26 @@ def check_access_prefix_directories
         EOS
       end
 
+      def check_missing_prefix_directories
+        path = HOMEBREW_PREFIX/('.brew-detect' + Time.now.to_i.to_s)
+        Dir.mkdir(path) and Dir.rmdir(path) # and return nil for all-ok
+      rescue SystemCallError # what mkdir/rmdir throws
```
MikeMcQuaid: Skip this and make it just the message, please (as you did before)

Library/Homebrew/diagnostic.rb
MikeMcQuaid: > The code I added checks that the following paths exist:

It doesn't do that because `HOMEBREW_PREFIX` isn't always `/usr/local`. Please add `return if HOMEBREW_PREFIX.to_s != "/usr/local"` as the first line of this function.

Library/Homebrew/keg.rb
```diff
@@ -88,6 +88,10 @@ def to_s
     mime-info pixmaps sounds postgresql
   ].freeze
 
+  ALL_TOP_LEVEL_PATHS = ALL_TOP_LEVEL_DIRECTORIES.map { |dir| HOMEBREW_PREFIX/dir }
+  ALL_SHARE_PATHS = SHARE_PATHS.map { |dir| HOMEBREW_PREFIX/"share"/dir }
+  ALL_TOP_AND_SHARE_PATHS = ALL_TOP_LEVEL_PATHS + ALL_SHARE_PATHS
```
MikeMcQuaid: I think these should all live in `diagnostic.rb` as they aren't used in `keg.rb`

Library/Homebrew/diagnostic.rb
```diff
@@ -320,15 +320,8 @@ def check_access_homebrew_repository
...
-        not_writable_dirs = []
-
-        Keg::ALL_TOP_LEVEL_DIRECTORIES.each do |dir|
-          path = HOMEBREW_PREFIX/dir
-          next unless path.exist?
-          next if path.writable_real?
-          not_writable_dirs << path
-        end
-
+        existing_dirs = Keg::ALL_TOP_AND_SHARE_PATHS.select(&:exist?)
```
MikeMcQuaid: Every time this is used you check if they exist so maybe it's worth just having the constant having those that exist?

Library/Homebrew/diagnostic.rb
```diff
@@ -345,6 +338,40 @@ def check_access_prefix_directories
...
 
+      def check_missing_prefix_directories
+        return if HOMEBREW_PREFIX.to_s != "/usr/local"
+        missing_dirs = Keg::ALL_TOP_LEVEL_PATHS.reject(&:exist?)
+        return if missing_dirs.empty?
+
+        <<~EOS
+          The following directories are missing:
+          #{missing_dirs.join("\n")}
+          You should create them:
```
MikeMcQuaid: `create and change ownership of them:` and then can kill `And change the permissions of these directories:` below

Library/Homebrew/diagnostic.rb
```diff
@@ -345,6 +338,40 @@ def check_access_prefix_directories
...
+            sudo mkdir -p #{missing_dirs.join(" ")}
+          And change the permissions of these directories:
+          #{permission_commands_for(missing_dirs).join("\n")}
+        EOS
+      end
+
+      def check_existing_prefix_directories_are_searchable
```
MikeMcQuaid: I think think it's just worth checking if the desired directories are readable, writeable and searchable within a single function as they offer the same permission commands to fix them.

Library/Homebrew/diagnostic.rb
```diff
@@ -1088,6 +1115,12 @@ def check_for_tap_ruby_files_locations
       def all
         methods.map(&:to_s).grep(/^check_/)
       end
+
+      def permission_commands_for(dirs)
```
MikeMcQuaid: Just inline this method above with `chmod 775` to help with readability.

MikeMcQuaid:
@sesam Any updates?

sesam:
I liked the idea of checking folders for existence only once, avoiding unnecessary slow downs.

I did one attempt 2w ago to decrese the line count, but it got complex and with code repetition, so I reverted. A good solution would mean taking apart this PR + some existing checks, build a list of paths and needed operations. Maybe show a nice readable explanation before the tangle of paths and commands to make it happen.

Though if practically nobody has problems with permissions and paths, then maybe it's not worth bogging down doctor with such checks. I've also been thinking maybe some part of this issue is better solved in the brew install code paths, maybe through capturing the errors from missing/inaccesible folders and suggest a solution right there.

MikeMcQuaid:
Ok, passing on this then.

PR #3714, opened 2018-01-21
resource: allow patches to be applied

ilovezfs:

-----

Allow resource blocks to contain patch blocks so we can stop downloading
the resource patches as separate resources and applying them manually.

### DIFF
--- a/Library/Homebrew/resource.rb
+++ b/Library/Homebrew/resource.rb
@@ -8,13 +8,13 @@ require "version"
 class Resource
   include FileUtils
 
-  attr_reader :mirrors, :specs, :using, :source_modified_time
+  attr_reader :mirrors, :specs, :using, :source_modified_time, :patches, :owner
   attr_writer :version
   attr_accessor :download_strategy, :checksum
 
   # Formula name must be set after the DSL, as we have no access to the
   # formula name before initialization of the formula
-  attr_accessor :name, :owner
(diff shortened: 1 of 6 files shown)


Library/Homebrew/resource.rb
```diff
@@ -82,9 +88,24 @@ def stage(target = nil, &block)
...
+    patches.grep(DATAPatch) { |p| p.path = owner.owner.path }
+
+    patches.each do |patch|
+      patch.verify_download_integrity(patch.fetch) if patch.external?
+    end
+  end
+
+  def apply_patches
+    return if patches.empty?
+    ohai "Patching resource"
```
MikeMcQuaid: Could this also include the `name`?

Library/Homebrew/test/patching_spec.rb
```diff
@@ -73,6 +84,22 @@ def patches
...
+          url TESTBALL_URL
+          sha256 TESTBALL_SHA256
+
+          patch do
+            url PATCH_URL_A
+            sha256 PATCH_A_SHA256
+          end
+        end
+      end,
+    ).to have_its_resource_patched
```
MikeMcQuaid: Nice test 👍

ilovezfs: Thanks :)

PR #3722, opened 2018-01-23
Add doctor and style to checklist.

sesam:
Ensures that contributors know about `brew doctor` and `brew style` before submitting a PR, since CI fails on `brew style` breaches.

-----

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,6 +2,8 @@
 - [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
 - [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
 - [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
+- [ ] Have you successfully run `brew style` with your changes locally?
+- [ ] Have you successfully run `brew doctor` with your changes locally?
 - [ ] Have you successfully run `brew tests` with your changes locally?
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,6 +2,8 @@
 - [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
 - [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
 - [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
+- [ ] Have you successfully run `brew style` with your changes locally?
+- [ ] Have you successfully run `brew doctor` with your changes locally?
```
ilovezfs: there are too many benign things flagged by doctor so let's not add that bullet

sesam: ok, removing that row again. While CI seems to fail if `brew doctor` fails, it's maybe unusual for doctor to fail due to new changes.

MikeMcQuaid:
I don't think this is necessary as CI will provide this information.

MikeMcQuaid:
To elaborate more: these were necessary for your PR because you were modifying `doctor`'s code. I don't think people not following `brew style` is occurring often enough to justify the addition to this list (but others may disagree).

MikeMcQuaid:
Actually, sure, let's add `brew style` as we tell people to run `brew tests`. 
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @sesam!

PR #3724, opened 2018-01-23
build: call satisified? before modifying env.

ilovezfs:

-----

Requirement#modify_build_environment may depend on the side effects of
Requirement#satisfied?, so make sure the latter is called at least once
for each requirement during formula installation. Prior to b70b5429d09,
Requirement#satisfied? would usually be called during Build#expand_reqs,
but that is longer the case.


ilovezfs:
Right now, `JAVA_HOME` is no longer getting set during `def install`, which leads to failures like https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/16982/version=high_sierra/console

PR #3763, opened 2018-02-05
audit: Warn on new formulae containing binary URLs

woodruffw:
Checking for the presence of "darwin" and either "x86_64" or "amd64" is a good heuristic for binary installs, which aren't allowed in official taps.

This doesn't check resource URLs, which are allowed to be binaries in some cases (e.g., `cargo` for bootstrapping).

cc @ilovezfs

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -198,6 +198,8 @@ class FormulaAuditor
     @online = options[:online]
     # Accept precomputed style offense results, for efficiency
     @style_offenses = options[:style_offenses]
+    # Allow the actual official-ness of a formula to be overridden, for testing purposes
+    @official_tap = formula.tap&.official? || options[:official_tap]
     @problems = []
     @text = FormulaText.new(formula.path)
     @specs = %w[stable devel head].map { |s| formula.send(s) }.compact
@@ -304,7 +306,7 @@ class FormulaAuditor
   def audit_formula_name
     return unless @strict
     # skip for non-official taps
-    return unless formula.tap&.official?
+    return unless @official_tap
 
     name = formula.name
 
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
@woodruffw Any chance this could be a URL check in `Library/Homebrew/rubocops/urls_cop.rb` instead?

woodruffw:
Yeah, that shouldn't be a problem. I might need some help retrieving just the main `url` instead of all urls, though -- any advice on that @GauthamGoli?

ilovezfs:
I think it's worth noting that the spec-specific url check for beta urls being used in stable, to which this new check is most similar, is still in dev-cmd/audit.rb https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L566-L592 not in the cops.

MikeMcQuaid:
> I think it's worth noting that the spec-specific url check for beta urls being used in stable, to which this new check is most similar, is still in dev-cmd/audit.rb https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/audit.rb#L540-L592 not in the cops.

Ignore me then!

ilovezfs:
👍 to making this a hard, regular audit failure (not restricted to --strict or --new-formula). It looks like the only affected homebrew/core formulae are apple-gcc42 and clozure-cl, which are true positives, and app-engine-go-64 and app-engine-go-32, which are being removed in https://github.com/Homebrew/homebrew-core/pull/23680. I guess we might need to white list apple-gcc42 or make a cask for it.

woodruffw:
> 👍 to making this a hard, regular audit failure (not restricted to --strict or --new-formula)

That's good by me. I'll make the changes this afternoon (along with figuring out how to stub out the official tap check for the specs).

woodruffw:
Alright, the build is passing now.

Aside from the new `audit` behavior introduced, the only big change in this PR is the introduction of an instance variable, `@official_tap`, in `FormulaAuditor` where `formula.tap&.official?` was used before. Introducing this allowed me to toggle official tap-dependent behavior, including this new check.

ilovezfs:
maybe also add a test that it doesn't trigger on a resource?

woodruffw:
> maybe also add a test that it doesn't trigger on a resource?

Added!

PR #3775, opened 2018-02-07
Do not show git username prompt during 'brew update'

alldne:

-----

Hi there,

1.
Whenever I do `brew update` on my laptop, I got the error saying:
```
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

2.
I found that this is the exactly same issue with the one in the following link:
https://github.com/Homebrew/homebrew-core/issues/10067#issuecomment-280637028
https://github.com/Homebrew/homebrew-core/issues/10067#issuecomment-349995050

In my case, the reason was that the certain repo referred by my homebrew was deleted. (So`git fetch` failed)
https://github.com/Flinesoft/BartyCrouch
https://github.com/flinesoft/homebrew-bartycrouch (now deleted)
```
Bartycrouch now is part of Homebrew Core! No tap needed any more.
If you had installed a previous version (<= 3.8.0) via the tap, you should run the following once:
$ brew untap flinesoft/bartycrouch
```

3.
git client tries to get credential from the user by showing prompt if it couldn't get access to the remote repository like this:

```
Username for 'https://github.com':
```
And I think we don't have to show this prompt because:

First, the error message saying that `terminal prompts disabled`. (git terminal prompt is disabled by design maybe? just guessing)
```
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

Second, if this prompt is shown, there are 2 possible cases:
a) There is no such repo (404)
=> no need to ask credential and we can just fail.

b) Private repo
=> Can a private repo be used in homebrew? I have no idea. **Help needed**

Also, if it just fails instead of showing prompt => more clear error message like `remote: Repository not found`.

4.
So, I set `GIT_ASKPASS` to `true`, which is the shell command that does nothing and always succeeds.

For the `GIT_ASKPASS`:
https://git-scm.com/docs/gitcredentials
```
If the GIT_ASKPASS environment variable is set, the program specified by the variable is invoked. A suitable prompt is provided to the program on the command line, and the user’s input is read from its standard output.
```

With this environment variable set, git client doesn't ask credential and there is no prompt.
I'm worrying about that this solution seems quite tricky and pretty misleading (`ASKPASS = true` but never ask? 😂 )

---
I think this is not the feature or bug reporting so I didn't make the issue and just sent the PR. If I was wrong, please let me know! Thank you.

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -512,7 +512,7 @@ EOS
         git fetch --tags --force "${QUIET_ARGS[@]}" origin \
           "refs/heads/$UPSTREAM_BRANCH_DIR:refs/remotes/origin/$UPSTREAM_BRANCH_DIR" 2>/dev/null
       else
-        if ! git fetch --tags --force "${QUIET_ARGS[@]}" origin \
+        if ! GIT_ASKPASS=true git fetch --tags --force "${QUIET_ARGS[@]}" origin \
           "refs/heads/$UPSTREAM_BRANCH_DIR:refs/remotes/origin/$UPSTREAM_BRANCH_DIR"
         then
           echo "Fetching $DIR failed!" >>"$update_failed_file"


ilovezfs:
Without this PR

iMac-TMP:Homebrew joe$ git rev-parse @
b4443e1cc0b41d2b25ca15b87c3d4cb2ab65837f
iMac-TMP:Homebrew joe$ brew update
fatal: could not read Username for 'https://github.com': terminal prompts disabled
Error: Fetching /usr/local/Homebrew/Library/Taps/ilovezfs/homebrew-evanescent failed!
```

With this PR
```
iMac-TMP:Homebrew joe$ brew pull https://github.com/Homebrew/brew/pull/3775
==> Fetching patch 
Patch: https://github.com/Homebrew/brew/pull/3775.patch
==> Applying patch
Applying: Do not show git username prompt during 'brew update'
==> Patch closes issue #3775
==> Patch changed:
 Library/Homebrew/cmd/update.sh | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
iMac-TMP:Homebrew joe$ brew update
remote: Repository not found.
fatal: Authentication failed for 'https://github.com/ilovezfs/homebrew-evanescent/'
Error: Fetching /usr/local/Homebrew/Library/Taps/ilovezfs/homebrew-evanescent failed!
iMac-TMP:Homebrew joe$ 
```

So it's not clear it's getting better. It looks like it's just adding an additional error.

**Comment by MikeMcQuaid** [comment]
@alldne any thoughts on @ilovezfs's comment?

**Comment by alldne** [comment]
Thank you for the response.

IMHO, this one 
```
remote: Repository not found.
fatal: Authentication failed for 'https://github.com/ilovezfs/homebrew-evanescent/'
Error: Fetching /usr/local/Homebrew/Library/Taps/ilovezfs/homebrew-evanescent failed!
```
is slightly better than this one
```
fatal: could not read Username for 'https://github.com': terminal prompts disabled
Error: Fetching /usr/local/Homebrew/Library/Taps/ilovezfs/homebrew-evanescent failed!
```
because, at least it says `Repository not found`, which could be a hint to resolve the issue.

When I faced the latter, I was confused like _'Why asking username?'_ , _'How can I type the username even though prompts is disabled?'_. 
It says 2 error messages and the first one says 'could not read user name' and the second one says 'fetching failed'. I thought that the reason of failure is that username wasn't given. I think most normal user including me might tend to think like that.

BTW I agree with you that this PR is not the best solution because it still produces an additional, unnecessary and even confusing error message: `fatal: Authentication failed for ...`.

I just want for brew to show some clear error message that helps.
To do so, 
1) Some confusing messages such as `fatal: could not read Username for 'https://github.com': terminal prompts disabled` or `fatal: Authentication failed for 'https://github.com/...` should be deleted
2) It would be great that a message helping user to resolve the issue is added. `remote: Repository not found.` might be the one of them or there might be better messages.

How about making `git fetch` command silent? At least it does 1.

Furthermore, how about making the existing message `Error: Fetching /usr/local/Homebrew/Library/Taps/... failed!` better?
For example, by showing remote repository URL with that message, it's gonna be easier for user to check the repository website, which is mostly GitHub and see '404 repository not found' page. I believe this helps user to recognize something happened on this repo.

MikeMcQuaid:
> How about making git fetch command silent? At least it does 1.

My concern is this hides more helpful error messages in some situations. Thoughts?

> Furthermore, how about making the existing message Error: Fetching /usr/local/Homebrew/Library/Taps/... failed! better?
> For example, by showing remote repository URL with that message, it's gonna be easier for user to check the repository website, which is mostly GitHub and see '404 repository not found' page. I believe this helps user to recognize something happened on this repo.

This seems like a good idea. Could you change this PR to take this approach? Thanks!

MikeMcQuaid:
Sorry, passing on this @alldne but thanks anyway!

PR #3782, opened 2018-02-10
write_jar_script: add java version option

apjanke:
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change? *This supersedes #3656*

-----

This expands `write_jar_script` to allow specifying the Java version. This will be needed to fix some of the Java 9 breakage in https://github.com/Homebrew/homebrew-core/issues/19696, for those formulae which use `write_jar_script` instead of `write_env_script`.

Closes #3656. This PR supersedes it; I'm putting in a new version since there's been no activity there in two weeks.

This is indeed another way to do https://github.com/Homebrew/homebrew-core/commit/39e0b03cdc6d8001aeed4881310faed7c7b75916, and I know ILZ won't like it. But consider this my vote that it's nicer to have `write_jar_script` do it. Let me know what you decide, and if it's a nay, I'll stick with the manual script generation like the `metabase` formula does does. (Feel free to just downvote this PR to vote "No"; I won't take it personal.)

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -436,11 +436,14 @@ class Pathname
   end
 
   # Writes an exec script that invokes a java jar
-  def write_jar_script(target_jar, script_name, java_opts = "")
+  def write_jar_script(target_jar, script_name, java_opts = "", java_version: nil)
     mkpath
+    java_home = if java_version
+      "JAVA_HOME=\"$(#{Language::Java.java_home_cmd(java_version)})\" "
+    end
     join(script_name).write <<~EOS
       #!/bin/bash
-      exec java #{java_opts} -jar #{target_jar} "$@"
+      #{java_home}exec java #{java_opts} -jar #{target_jar} "$@"
     EOS
   end
 


ilovezfs:
As mentioned [here](https://github.com/Homebrew/brew/pull/3656#discussion_r163800942) as long as it's not using `java_home --exec`, while `env_script_all_files` and `write_env_script` use `export JAVA_HOME=` I see this as unobjectionable.

apjanke:
Gotcha; I thought your objection was more about adding more DSL features. Thanks.

ilovezfs:
Yeah, I'd just as soon get rid of all of them ;)

Library/Homebrew/extend/pathname.rb
```diff
@@ -436,11 +436,17 @@ def env_script_all_files(dst, env)
   end
 
   # Writes an exec script that invokes a java jar
-  def write_jar_script(target_jar, script_name, java_opts = "")
+  def write_jar_script(target_jar, script_name, java_opts = "", java_version: nil)
     mkpath
+    java_home = if java_version
+      <<~EOS
+        export JAVA_HOME="$(#{Language::Java.java_home_cmd(java_version)})"
```
MikeMcQuaid: Is it worth skipping the `export` and just putting the var on the same line?

Library/Homebrew/extend/pathname.rb
```diff
@@ -436,11 +436,17 @@ def env_script_all_files(dst, env)
...
+
+      EOS
+    end
     join(script_name).write <<~EOS
       #!/bin/bash
-      exec java #{java_opts} -jar #{target_jar} "$@"
+      #{java_home}exec java #{java_opts} -jar #{target_jar} "$@"
```
MikeMcQuaid: Utter pedantry I realise (sorry) but does this not end up with yucky indentation in the output file?

Library/Homebrew/extend/pathname.rb
apjanke: Sure. Amended.

Library/Homebrew/extend/pathname.rb
apjanke: I don't think so? Unless you mean the two spaces between "java" and "-jar" when "java_opts" is empty?

MikeMcQuaid: @apjanke np; when it was using `EOS` I think the beginning of this line would have had weird indentation.

apjanke: I actually think it wouldn't've, because of `~EOS`'s de-indentation. (That's the sort of thing that would bother me too, no worries on the pedantry.) But... moot point now. ;)

MikeMcQuaid:
Thanks again @apjanke!

apjanke:
:+1:

Issue #2173, opened 2017-02-25
Undeclared, non-recursive dependencies are not recorded in INSTALL_RECEIPT.json


ilovezfs:
- [x] Ran `brew update` and retried your prior step?
- [x] Ran `brew doctor`, fixed as many issues as possible and retried your prior step?
- [x] Confirmed this is problem with Homebrew/brew and not specific formulae? If it's a formulae-specific problem please file this issue at https://github.com/Homebrew/homebrew-core/issues/new

### Bug reports:

```
iMac-TMP:~ joe$ brew doctor
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: The filesystem on / appears to be case-sensitive.
The default macOS filesystem is case-insensitive. Please report any apparent problems.
```
```
iMac-TMP:~ joe$ brew config
HOMEBREW_VERSION: 1.1.10-295-g93e2cb3
ORIGIN: https://github.com/Homebrew/brew
HEAD: 93e2cb31afa1671c36023d119d9a539373b4be43
Last commit: 61 minutes ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: d88f2d6d27258b31c8f61d91df9aa6b3cb175c2c
Core tap last commit: 65 minutes ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/Homebrew
HOMEBREW_CELLAR: /usr/local/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
CPU: octa-core 64-bit skylake
Homebrew Ruby: 2.0.0-p648 => /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.0.0-p648/bin/ruby
Clang: 8.0 build 800
Git: 2.10.1 => /Applications/Xcode.app/Contents/Developer/usr/bin/git
Perl: /usr/bin/perl
Python: /usr/local/bin/python => /usr/local/Cellar/python/2.7.13/Frameworks/Python.framework/Versions/2.7/bin/python2.7
Ruby: /usr/local/bin/ruby => /usr/local/Cellar/ruby/2.4.0/bin/ruby
Java: 1.8.0_92, 1.8.0_77, 1.7.0_80, 1.6.0_65-b14-468
macOS: 10.11.6-x86_64
[...]

PR #3785, opened 2018-02-10
linkage_checker: Distinguish indirect deps from undeclared deps

apjanke:
- [X] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [X] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). *No - there are no `linkage_checker` tests yet, and I don't know where to start for that.*
- [ ] Have you successfully run `brew style` with your changes locally? *No, but I don't think it's a regression.*
- [X] Have you successfully run `brew tests` with your changes locally?

-----

Follows up on #2297 and https://github.com/Homebrew/homebrew-core/pull/23933#issuecomment-364633487.

This modifies `brew linkage` to separate out the "undeclared dependencies" in to two lists: an "indirect dependencies" list for linkage dependencies that are picked up transitively but not directly declared in the formula, and "undeclared dependencies" for those that are not in the recursive dependencies at all.

 Before:

```
[/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula on ⇄ ad8a26413e]
$ brew linkage mldonkey
System libraries:
  /usr/lib/libSystem.B.dylib
  /usr/lib/libbz2.1.0.dylib
  /usr/lib/libc++.1.dylib
  /usr/lib/libcharset.1.dylib
  /usr/lib/libiconv.2.dylib
  /usr/lib/libz.1.dylib
Homebrew libraries:
  /usr/local/opt/fontconfig/lib/libfontconfig.1.dylib (fontconfig)
  /usr/local/opt/freetype/lib/libfreetype.6.dylib (freetype)
  /usr/local/opt/gd/lib/libgd.3.dylib (gd)
  /usr/local/opt/jpeg/lib/libjpeg.9.dylib (jpeg)
  /usr/local/opt/libmagic/lib/libmagic.1.dylib (libmagic)
  /usr/local/opt/libpng/lib/libpng16.16.dylib (libpng)
  /usr/local/opt/libtiff/lib/libtiff.5.dylib (libtiff)
  /usr/local/opt/webp/lib/libwebp.7.dylib (webp)
Undeclared dependencies with linkage:
  fontconfig
  freetype
  jpeg
  libmagic
[...]

PR #3789, opened 2018-02-11
Undeclared dependencies in runtime_dependencies

alyssais:

-----

Fixes https://github.com/Homebrew/brew/issues/2173.

I've just noticed #3785. That probably interplays with this change in some way, but I haven't had a chance to have a proper look at it yet.


Library/Homebrew/formula.rb
```diff
@@ -1835,6 +1832,23 @@ def eligible_kegs_for_cleanup
...
+      Dependency.prune if !dependency.required? && build.without?(dependency)
+    end
+  end
+
+  def undeclared_runtime_dependencies
+    return [] unless optlinked?
+
+    keg = Keg.new(opt_prefix)
+    linkage_checker = LinkageChecker.new(keg, self)
+    dylib_formula_names = linkage_checker.brewed_dylibs.keys
```
ilovezfs: This includes the declared ones.

ilovezfs:
This PR does not save the undeclared ones to the `INSTALL_RECEIPT.json` and so unsafe removal of opportunistically linked deps is still allowed.

alyssais:
> This PR does not save the undeclared ones to the `INSTALL_RECEIPT.json` and so unsafe removal of opportunistically linked deps is still allowed.

Ah, I'd assumed `Tab.create` would pick them up, but it doesn't because the formula doesn't get linked into opt until after the tab is created. We can work around that by falling back to the formula's prefix if it isn't optlinked.

Library/Homebrew/formula.rb
alyssais: Ah, so it does. This didn't matter when I had this all in one method because we took the union, but it does now that it's in a method called `undeclared_runtime_dependencies`.

Library/Homebrew/formula.rb
```diff
@@ -13,6 +13,7 @@
 require "tap"
 require "keg"
 require "migrator"
+require "os/mac/linkage_checker"
```
alyssais: Will fix this to properly use the OS layer on the next pass.

ilovezfs:
how is this not also a problem for Linux?

alyssais:
It is a problem on Linux, but we don't have a LinkageChecker there

alyssais:
`brew linkage` also doesn't work on Linux

ilovezfs:
Oh joy

alyssais:
I think there was a WIP implementation for Linuxbrew at some point, but not sure what happened to it.

sjackman:
`brew linkage` works on Linuxbrew.

MikeMcQuaid:
@sjackman any chance of getting it upstreamed in the near future for this PR?

sjackman:
Done! I'm that good. (It works already too.) =p
```
$ brew linkage hello
System libraries:
  /lib/x86_64-linux-gnu/libc.so.6
```

MikeMcQuaid:
@alyssais What's the latest here?

Library/Homebrew/extend/os/mac/formula.rb
```diff
@@ -0,0 +1,16 @@
...
+    elsif prefix.directory?
+      keg = Keg.new(prefix)
+    else
+      return []
+    end
+
+    linkage_checker = LinkageChecker.new(keg, self)
+    linkage_checker.undeclared_deps.map { |n| Dependency.new(n) }
+  end
+end
```
sjackman: Why is `undeclared_runtime_dependencies` undefined for Linux?

alyssais: Because I didn't realise we had working `linkage` on Linux :P

alyssais:
Oh, I see. os/mac/linkage_checker is used on Linux even though it's in os/mac? I'll update the PR accordingly.

ilovezfs:
@alyssais Also, have you determined what impact this has on `brew deps`? Note that https://github.com/Homebrew/brew/pull/3785 was merged.

sjackman:
> Oh, I see. os/mac/linkage_checker is used on Linux even though it's in os/mac? I'll update the PR accordingly.

Hmm, how sketchy. I'll submit a PR to move `linkage_checker` to generic.

sjackman:
PR https://github.com/Homebrew/brew/pull/3852 moves `linkage_checker.rb` to generic.

alyssais:
> Also, have you determined what impact this has on brew deps?

AFAICT `brew deps` doesn't call `runtime_dependencies`, so it should have no impact.

> Note that #3785 was merged.

#3785 will stop indirect dependencies suddenly showing up as runtime dependencies, which would have happened otherwise, I think.

alyssais:
> `SystemStackError`

yikes

Library/Homebrew/formula.rb
```diff
@@ -1835,6 +1833,26 @@ def eligible_kegs_for_cleanup
...
+    if optlinked?
+      keg = Keg.new(opt_prefix)
+    elsif prefix.directory?
+      keg = Keg.new(prefix)
+    else
+      return []
+    end
+
+    linkage_checker = LinkageChecker.new(keg, self)
+    linkage_checker.undeclared_deps.map { |n| Dependency.new(n) }
```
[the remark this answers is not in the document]
alyssais: Cheers Mike. Will get that taken care of.

Library/Homebrew/formula.rb
```diff
@@ -1487,12 +1488,29 @@ def recursive_requirements(&block)
   # Returns a list of Dependency objects that are required at runtime.
   # @private
   def runtime_dependencies
```
MikeMcQuaid: I was thinking: given this is going to change the output of `runtime_dependencies` again quite drastically: I think it would be worth having the `runtime_dependencies` in the Tab be versioned. This means we don't need to do something like `parsed_homebrew_version < "1.1.6"` (in `tab.rb`) again in order to make sure these tabs have what we expect/need to rely on them.

MikeMcQuaid:
This PR will be affected by https://github.com/Homebrew/brew/pull/3979 (which I think should probably go in first).

Library/Homebrew/formula.rb
MikeMcQuaid: I've changed my mind on this given https://github.com/Homebrew/brew/pull/3979 so: ignore it.

MikeMcQuaid:
@alyssais What's the latest here? Thanks!

Library/Homebrew/formula.rb
```diff
@@ -1494,10 +1495,20 @@ def runtime_dependencies(read_from_tab: true)
...
 
-    recursive_dependencies do |_, dependency|
-      Dependency.prune if dependency.build?
-      Dependency.prune if !dependency.required? && build.without?(dependency)
+    declared_runtime_dependencies | undeclared_runtime_dependencies
+  end
+
+  def undeclared_runtime_dependencies
+    if optlinked?
+      keg = Keg.new(opt_prefix)
```
MikeMcQuaid: What's the advantage of using the `opt_prefix` when it's `optlinked?`?

Library/Homebrew/linkage_checker.rb
```diff
@@ -63,7 +63,7 @@ def check_undeclared_deps
       formula.build.without?(dep)
     end
     declared_deps = formula.deps.reject { |dep| filter_out.call(dep) }.map(&:name)
-    runtime_deps = keg.to_formula.runtime_dependencies(read_from_tab: false)
+    runtime_deps = keg.to_formula.declared_runtime_dependencies
```
MikeMcQuaid: This method doesn't seem to be defined.

MikeMcQuaid: https://travis-ci.org/Homebrew/brew/jobs/360237284

Library/Homebrew/linkage_checker.rb
```diff
@@ -63,7 +63,7 @@ def check_undeclared_deps
       formula.build.without?(dep)
     end
     declared_deps = formula.deps.reject { |dep| filter_out.call(dep) }.map(&:name)
-    runtime_deps = keg.to_formula.runtime_dependencies(read_from_tab: false)
```
MikeMcQuaid: Maybe the `read_from_tab` argument can go away entirely? The only other caller would be in `tab.rb`

Library/Homebrew/linkage_checker.rb
MikeMcQuaid: Fixed now.

Library/Homebrew/formula.rb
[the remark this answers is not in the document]
MikeMcQuaid: Fixed now.

Library/Homebrew/formula.rb
alyssais: Uses the currently active version of a formula rather than the latest. Can happen with `brew switch` or even just if `brew upgrade` hasn't been run yet.

Library/Homebrew/formula.rb
```diff
@@ -1486,20 +1487,40 @@ def recursive_requirements(&block)
...
+    if installed_prefix.directory? &&
+       (tab_deps = Keg.new(installed_prefix)&.runtime_dependencies)
+
       return tab_deps.map { |d| Dependency.new d["full_name"] }.compact
     end
 
+    compute_runtime_dependencies
+  end
+
+  def compute_runtime_dependencies
```
MikeMcQuaid: Sorry to be a pain but I think this is probably worse than the `read_from_tab` argument was 😭. It's not clear when you need to call `runtime_dependencies` vs. `compute_runtime_dependencies`. Could maybe fiddle with naming options but I'm tempted to just say put it back to how it was. Sorry.

Library/Homebrew/linkage_checker.rb
MikeMcQuaid: Changed my mind on this: ignore!

MikeMcQuaid:
@ilovezfs Mind kicking this around a bit and seeing if it does what you expect/want before I merge? Thanks ❤️

ilovezfs:
It seems to work.
```
iMac-TMP:freeciv joe$ diff before after
1c1
< {"homebrew_version":"1.6.2-14-g29f7d3d"
---
> {"homebrew_version":"1.6.2-29-g8035afc"
12c12
< "time":1524657205
---
> "time":1524657606
14c14
< "HEAD":"29f7d3d7ef5b2b939021ac247ef042228b8cc8a2"
---
> "HEAD":"8035afcc36ac45ebd127f670b13ab32414c310fa"
73c73,75
< "version":"2.24.32"}]
---
> "version":"2.24.32"}
> {"full_name":"xz"
> "version":"5.2.3"}]
```
for `freeciv` opportunistically linked to xz.

Do you want to bump any sort of INSTALL_RECEIPT version here since it changes the meaning/reliability of that data?

MikeMcQuaid:
> It seems to work.

🎉 

> Do you want to bump any sort of INSTALL_RECEIPT version here since it changes the meaning/reliability of that data?

I reckon we can leave it as-is because this is "value add" rather than "broken without".

MikeMcQuaid:
Thanks @alyssais!

ilovezfs:
Thanks @alyssais and @MikeMcQuaid!

ilovezfs:
This PR broke upgrades when there are formulae with the same name in two different taps even if the formula dependency is fully qualified. https://github.com/Homebrew/brew/commit/8386e6a8c18f5dee30f17386cab652b653b1b860 is the breaking commit.

```
iMac-TMP:Homebrew joe$ git rev-parse @
27ec9dae5fa4c2c5b19f8bfafaefbdbb3e9a9c8c
iMac-TMP:Homebrew joe$ brew upgrade
==> Upgrading 2 outdated packages, with result:
pothosware/pothos/pothoscomms 0.2.2 -> 0.3.1, pothosware/pothos/pothosplotters 0.3.1 -> 0.4.1
Error: Formulae found in multiple taps: 
       * dholm/sdr/spuce
       * pothosware/pothos/spuce

Please use the fully-qualified name e.g. dholm/sdr/spuce to refer the formula.
iMac-TMP:Homebrew joe$ 
```
with debug
```
==> Upgrading 2 outdated packages, with result:
pothosware/pothos/pothoscomms 0.2.2 -> 0.3.1, pothosware/pothos/pothosplotters 0.3.1 -> 0.4.1
Error: Formulae found in multiple taps: 
       * dholm/sdr/spuce
       * pothosware/pothos/spuce

Please use the fully-qualified name e.g. dholm/sdr/spuce to refer the formula.
/usr/local/Homebrew/Library/Homebrew/formulary.rb:377:in `loader_for'
/usr/local/Homebrew/Library/Homebrew/formulary.rb:280:in `factory'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1444:in `[]'
/usr/local/Homebrew/Library/Homebrew/linkage_checker.rb:141:in `block in check_undeclared_deps'
/usr/local/Homebrew/Library/Homebrew/linkage_checker.rb:139:in `reject'
/usr/local/Homebrew/Library/Homebrew/linkage_checker.rb:139:in `check_undeclared_deps'
/usr/local/Homebrew/Library/Homebrew/linkage_checker.rb:104:in `check_dylibs'
/usr/local/Homebrew/Library/Homebrew/linkage_checker.rb:19:in `initialize'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1530:in `new'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1530:in `undeclared_runtime_dependencies'
/usr/local/Homebrew/Library/Homebrew/formula.rb:1511:in `runtime_dependencies'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:424:in `runtime_requirements'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:437:in `expand_requirements'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:381:in `compute_dependencies'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:136:in `verify_deps_exist'
/usr/local/Homebrew/Library/Homebrew/formula_installer.rb:129:in `prelude'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:139:in `upgrade_formula'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:95:in `block in upgrade'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:92:in `each'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:92:in `upgrade'
/usr/local/Homebrew/Library/Homebrew/brew.rb:101:in `<main>'
```

PR #3791, opened 2018-02-12
diagnostic: guard cache check on Jenkins too

DomT4:

-----

Apparently Jenkins doesn't set `CI` in the `env` automatically? This should get CI working again. Sorry!

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -837,7 +837,8 @@ module Homebrew
 
       def check_for_large_cache
         return unless HOMEBREW_CACHE.exist?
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@DomT4 Thanks for the quick fix. Merging for now but I want to tweak this so we don't have to keep remembering to set `JENKINS_HOME` here.

DomT4:
Maybe `Utils.is_a_robot?` or something that combines these variables into a blob, since we're using it more than once now.

DomT4:
Thanks for merging. Apologies for this one; was sure Jenkins set `CI` by default sigh. Maybe I was misremembering and thinking of Travis, which is possible.

PR #3806, opened 2018-02-15
upgrade: changed pinned message from Error to Warning

sferrini:
This commit https://github.com/Homebrew/brew/commit/3a2e6b82fd008974c47b87a98c6c7bbf1dcdcea7#diff-aa6192b25b5254b9e0d4b95005b69a44R65 changed the message type from `oh1` to `onoe`. I think that having pinned packages that don't upgrade it's not an error (`onoe`), instead I would give the user just a warning (`opoo`).

### DIFF
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -62,7 +62,7 @@ module Homebrew
     end
 
     unless pinned.empty?
-      onoe "Not upgrading #{Formatter.pluralize(pinned.length, "pinned package")}:"
+      opoo "Not upgrading #{Formatter.pluralize(pinned.length, "pinned package")}:"
       puts pinned.map { |f| "#{f.full_specified_name} #{f.pkg_version}" } * ", "
     end
 


MikeMcQuaid:
I disagree, I do consider this an error. @ilovezfs, thoughts?

sferrini:
@MikeMcQuaid why do you consider it as an error?

If I'm pinning one package, it's because I don't want it to be upgraded.
It's not an error if it doesn't get upgraded, it's just the expected behavior.

I would put just a warning to remind users that they have pinned packages, nothing more.

that said, I'm open to discussions and to change my idea if you give me valid a explanation.

ilovezfs:
>why do you consider it as an error?

Let's say in Homebrew/homebrew-core we upgrade ruby in such a way that vim's linkage to ruby is broken unless you also rebuild vim against the new ruby. We will have upgraded ruby and revision bumped vim (which brew then treats as an upgrade) in the same pull request, and if you run `brew upgrade` without any pins in place, then nothing will break. Ruby will be upgraded to the new version, vim will be upgraded to the rebuilt bottle that links against the new ruby, and all linkage is fine.

If, however, you've pinned vim, then you will now have a broken vim installation as it's still linked against the old ruby, and the fact that you pinned vim will in no way have stopped `brew upgrade` from upgrading ruby, nor will any magic kick in to rebuild the old pinned version of vim against the new ruby. Your vim is now toast and you will likely end up filing an issue on our issue tracker unless you happen to understand what happened and how to fix it.

So there is a good reason to consider `brew upgrade` in the face of pins to be a failed command regardless of whether it has behaved "as expected."

That said, I'm not sure it's the most useful behavior because it becomes impossible to use, for example, `brew upgrade && do something` if you have any pins at all, and `brew upgrade; do something` is not the same thing, and may not be what you want to happen if any of the unpinned formulae failed to upgrade.

MikeMcQuaid:
I agree with @ilovezfs. Sorry, passing on this.

ilovezfs:
As a workaround,
```
brew upgrade $(brew outdated | grep -v -f <(brew list --pinned))
```
probably does what you want.

ilovezfs:
Actually, despite printing the error, currently it looks like `brew upgrade` _does_ exit 0 contrary to my assumption:
```
Josephs-MacBook-Pro:~ joe$ brew upgrade
==> No packages to upgrade
Error: Not upgrading 3 pinned packages:
bash 4.4.19, fd 6.3.0, ripgrep 0.8.0
Josephs-MacBook-Pro:~ joe$ echo $?
0
```

sferrini:
I hate to see in the output the `Error` word, when instead, in this (my) case is not an error at all.

If "broken dependencies" is your concern, print the error only if/when this happens.
I also spoke with other people and they all agree with me.

I'll create an alias/function and `sed` out the word if I match the `Not upgrading $n pinned packages` string, or maybe, it's time to keep using my own (brew) fork.

MikeMcQuaid:
> it's time to keep using my own (brew) fork.

Feel free, our license permits it.

PR #3812, opened 2018-02-15
update.sh: use version-sort option of sort

maxim-belkin:
Also, use more general regular expression in grep

-----

I think `--version-sort` is a more robust approach than specifying individual fields for sorting.
Not sure why `grep` is used instead of `head -n 1` but I kept it and generalized its pattern a bit.

Also, not sure why this is not something like:
```
git describe --abbrev=0 --tags origin/master
```


MikeMcQuaid:
I'd rather avoid refactorings like this that may cause end user bugs when they aren't fixing end user bugs.

maxim-belkin:
image

:smile:  https://youtu.be/nNv0IoE5XHw

PR #3820, opened 2018-02-20
Define OS::Mac on Linux

sjackman:
Define `OS::Mac` on Linux for formula API compatibility.

-----

Define `MacOS.version`, `MacOS.full_version`, and `MacOS::Xcode.version` to `Version::NULL` on Linux so that `brew readall` succeeds and `Homebrew/brew` can tap `Homebrew/core` on Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,22 +36,13 @@ before_install:
       mv "$HOMEBREW_REPOSITORY/Library/Taps" "$PWD/Library";
       sudo rm -rf "$HOMEBREW_REPOSITORY";
       sudo ln -s "$PWD" "$HOMEBREW_REPOSITORY";
-    else
-      HOMEBREW_CORE_TAP_DIR="$(brew --repo "homebrew/core")";
-      mkdir -p "$HOMEBREW_CORE_TAP_DIR";
     fi
   - if [ "$MACOS" ]; then
       travis_retry git -C Library/Taps/homebrew/homebrew-core fetch --depth=1 origin;
+    else
+      travis_retry git clone --depth=1 https://github.com/Homebrew/homebrew-core Library/Taps/homebrew/homebrew-core;
     fi
   - travis_retry git clone --depth=1 https://github.com/Homebrew/homebrew-test-bot Library/Taps/homebrew/homebrew-test-bot
-  - if [ "$LINUX" ]; then
-      HOMEBREW_TEST_BOT_TAP_DIR="$(brew --repo "homebrew/test-bot")";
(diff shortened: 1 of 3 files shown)


maxim-belkin:
Hmmm... wouldn't it be better to define two new methods instead, _i.e._ `olderthan?` and `newerthan?` so that, as an example
```
MacOS.version <= :snow_leopard
```
would become
```
MacOS.version.older_than? :snow_leopard
```
And make both of the methods return `false` on Linux... Otherwise, the above comparison will be returning `true` on Linux

maxim-belkin:
```
$ uname -s && brew irb <<< "Version::NULL < :snow_leopard"
Linux
Switch to inspect mode.
Version::NULL < :snow_leopard
true
```

sjackman:
The primary intention of this patch is to make it possible to tap `Homebrew/core` using `Homebrew/brew` running on Linux. We can discuss in a separate PR possible changes to the Homebrew formula API.

maxim-belkin:
Four methods would be required, actually :thinking: :
```
<    :  older_than?
<=   :  not_newer_than?
>=   :  not_older_than?
>    :  newer_than?
```

maxim-belkin:
Sure, but this is a hack. `MacOS.version` is meaningless on Linux and should not be returning NULL. If anything, it should give something like "not defined". We can not assume that all Linux distributions correspond to old macOS.

sjackman:
`Version:NULL` does mean not defined.

maxim-belkin:
`Version::NULL < something` evaluates to `true` for any number and any string, whereas
`Version::NULL > something` evaluates to `false` even for negative numbers.
So, whenever there is `MacOS.version < something` comparison in a formula, it will be considered as `true` on Linux. So, unless it's augmented/prefixed by `OS.mac? &&` this might lead to incorrect/undesired behavior or unnecessary patches being applied

```bash
$ uname -s
Linux

$ cd $(brew --repo homebrew/core)/Formula

$ grep -w "MacOS.version <" * | wc -l
312

$ grep -w "MacOS.version <" * | grep 'OS.mac?' | wc -l
14
```

MikeMcQuaid:
> Version::NULL does mean not defined. You can use Version.null? to test whether a version is null.

I agree, let's keep with this approach.

@sjackman Does this patch enable an otherwise unpatched Homebrew/brew to `brew readall` on Homebrew/homebrew-core?

ilovezfs:
>So, whenever there is MacOS.version < something comparison in a formula, it will be considered as true on Linux.

This is caused by spaghetti formulae that attempt to interweave macOS and Linux support and/or reuse macOS formulae from Homebrew/homebrew-core.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
> Does this patch enable an otherwise unpatched Homebrew/brew to brew readall on Homebrew/homebrew-core?

Yes!

Thanks for merging, Mike.

PR #3831, opened 2018-02-23
update some doc references [skip ci]

arthurnn:
This is a small change, just as I was reading the docs I found some 404 links. This fixes the ones I found.

cc @MikeMcQuaid


MikeMcQuaid:
Thanks for the PR @arthurnn. This is something I'm working on a bigger change for but it's not time critical because docs.brew.sh is considered the authoritative source.

PR #3833, opened 2018-02-24
rubocop: Add cop to check `depends_on` order and tests

GauthamGoli:

-----
This new cop orders `depends_on` in the following sequence

=> :build first
=> :run second
=> regular third
=> :recommended fourth
=> :optional fifth

and _within_ each category alphabetized.

Also, conditional dependency ordering is taken care of 
```ruby
depends_on "apple" if build.with? "foo" || build.with? "bar"
depends_on "foo" => :optional
depends_on "bar"
```
After auto correct
```ruby
depends_on "bar"
depends_on "foo" => :optional
depends_on "apple" if build.with? "foo" || build.with? "bar"
```

cc @ilovezfs / @MikeMcQuaid


MikeMcQuaid:
I tested this on `wget` and it fixed the order as expected 👍. `brew style --fix wget` did not, however.

GauthamGoli:
@MikeMcQuaid I think thats because this is a `FormulaAuditStrict` cop, which doesn't get executed for `brew style`

ilovezfs:
Let's merge this as a new formula only audit, and start using it. Then if all goes well, and we've worked out any rough spots, we can consider upgrading it to --strict or even non-strict and mass re-ordering everything.

GauthamGoli:
Updated. This cop is now a `--new-formula` audit.

MikeMcQuaid:
Thanks again @GauthamGoli!

PR #3835, opened 2018-02-25
audit:  `--new-formula`, Don't fail CI instead comment on PR about audit violations

GauthamGoli:

-----
WIP 

For `--new-formula` , dont fail CI when there are audit violations. Instead, comment on formula's PR. 

@BrewTestBot user could do the commenting part? 

cc @MikeMcQuaid


ilovezfs:
This somehow would need to guard against making the same comments over and over.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -121,8 +121,16 @@ def audit
     end
 
     return if problem_count.zero?
+    errors_summary = "#{Formatter.pluralize(problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
+    create_pr_comment("name-of-repo", "Pull Request Number", errors_summary) if new_formula
+    ofail errors_summary unless new_formula
+  end
 
-    ofail "#{Formatter.pluralize(problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
+  def create_pr_comment(repo, number, body)
```
MikeMcQuaid: I reckon this should live in `github.rb`. Also it should ideally avoid creating duplicate comments; perhaps it's worth accessing the existing comments on a PR and only creating a comment if the string-identical comment doesn't already exist?

MikeMcQuaid:
Nice work on this so far!

GauthamGoli:
@MikeMcQuaid Added a check to see if same comment already exists.

MikeMcQuaid:
This is my bad, I'll look at this more.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -121,8 +121,12 @@ def audit
     end
 
     return if problem_count.zero?
-
-    ofail "#{Formatter.pluralize(problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
+    errors_summary = "#{Formatter.pluralize(problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
+    if strict
```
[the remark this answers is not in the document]
MikeMcQuaid: @GauthamGoli Any thoughts on this?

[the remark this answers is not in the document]
GauthamGoli: @MikeMcQuaid  Yeah. separate method call is easier. What errors do we want to start with ?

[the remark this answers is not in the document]
MikeMcQuaid: @GauthamGoli All errors that are currently just `new_formula`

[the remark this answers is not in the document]
GauthamGoli: Custom Rubocop audit errors for new formula can be easily differentiated as they are `NewFormulaAudit` department, but a regular audit `problem` cannot be differentiated from `new_formula` problem. One way to do it is have some sort of flag to filter, would you suggest this / better way ?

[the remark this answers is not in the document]
MikeMcQuaid: @GauthamGoli You can look for the use of `if @new_formula`, `return unless @new_formula`, `next unless @new_formula` to see all these.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -517,7 +532,8 @@ def audit_specs
 
       next if spec.patches.empty?
       next unless @new_formula
```
MikeMcQuaid: Might be good to be able to remove the use of both `@new_formula` and `new_formula: true` being needed

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -605,7 +621,8 @@ def audit_revision_and_version_scheme
       next unless spec = formula.send(spec_sym)
       next unless previous_version_and_checksum[spec_sym][:version] == spec.version
       next if previous_version_and_checksum[spec_sym][:checksum] == spec.checksum
-      problem "#{spec_sym}: sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed."
+      problem "#{spec_sym}: sha256 changed without the version also changing; please create an issue upstream to rule out malicious circumstances and to find out why the file changed.",
```
MikeMcQuaid: Everything in `audit_revision_and_version_scheme` is _never run_ for new formulae.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -121,8 +123,18 @@ def audit
     end
 
     return if problem_count.zero?
+    errors_summary = "#{Formatter.pluralize(problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
+    if new_formula && new_formula_problem_lines.size.zero?
+      GitHub.create_issue_comment("repository-name", "pull-request-number",
```
MikeMcQuaid: I think we should pick a `HOMEBREW_` environment variable for these values which can be passed in by `brew test-bot`. I think if they are missing or the API call fails it'd be good to just output them as usual. Thoughts?

MikeMcQuaid: @GauthamGoli I think if the environment variables are set: yes.

GauthamGoli: @MikeMcQuaid  Updated the PR accordingly. Let me know what you think.

Library/Homebrew/utils/github.rb
```diff
@@ -291,4 +291,30 @@ def search(entity, *queries, **qualifiers)
     uri.query = query_string(*queries, **qualifiers)
     open_api(uri) { |json| json.fetch("items", []) }
   end
+
+  def create_issue_no_op?
+    !ENV.key?("HOMEBREW_GH_REPO") || !ENV.key?("HOMEBREW_GH_PR") || ENV["HOMEBREW_NO_GH_API"]
```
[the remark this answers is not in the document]
GauthamGoli: Updated the PR!.

MikeMcQuaid:
@GauthamGoli @ilovezfs I've tested this and done some minor cleanup and it looks good to me. If you're 👍 I'll merge this when I can be around to babysit it for a bit.

GauthamGoli:
Looks good to me.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -508,8 +535,7 @@ def audit_specs
         end
 
         next if spec.patches.empty?
-        next unless @new_formula
```
commitay: Is this line still needed?

ilovezfs:
I'd like to see the references to "new formulae" removed from the messaging, again to avoid planting the immediate-follow-up-PR-to-cheat-the-audit seed in contributors' minds.

Library/Homebrew/dev-cmd/audit.rb
MikeMcQuaid: Yep.

MikeMcQuaid:
Thanks again @GauthamGoli!

commitay:
`brew audit --new-formula` doesn't error or have any output when run locally.

ilovezfs:
Hmm... we do need some way to run this locally and see the new formula audit errors.

GauthamGoli:
Its possible. `@new_formula_problems` can be inspected

ilovezfs:
I mean from the CLI.

GauthamGoli:
Oh. Yeah even that is possible , and if `HOMEBREW_NEW_FORMULA_PULL_REQUEST_URL` environment variable is not set, new formula audit errors are outputted normally. 

@commitay  `brew audit --new-formula` works fine for me. Which audit specifically were you looking for ?

ilovezfs:
I've suspended all builds on Jenkins as @commitay indicates even regular audits seem not to be running.

commitay:
e.g `brew audit --new-formula imagemagick`

No output currently.

Previously: 
```
imagemagick:
  * C: 29: col 3: New Formula should not use `deprecated_option`
  * C: 40: col 3: dependency "freetype" (line 40) should be put before dependency "libtiff" (line 39)
  * C: 43: col 3: dependency "fontconfig" (line 43) should be put before dependency "x11" (line 42)
  * C: 44: col 3: dependency "little-cms" (line 44) should be put before dependency "x11" (line 42)
  * C: 45: col 3: dependency "little-cms2" (line 45) should be put before dependency "x11" (line 42)
  * C: 46: col 3: dependency "libde265" (line 46) should be put before dependency "x11" (line 42)
  * C: 47: col 3: dependency "libwmf" (line 47) should be put before dependency "x11" (line 42)
  * C: 48: col 3: dependency "librsvg" (line 48) should be put before dependency "x11" (line 42)
  * C: 49: col 3: dependency "liblqr" (line 49) should be put before dependency "x11" (line 42)
  * C: 50: col 3: dependency "openexr" (line 50) should be put before dependency "x11" (line 42)
  * C: 51: col 3: dependency "ghostscript" (line 51) should be put before dependency "x11" (line 42)
  * C: 52: col 3: dependency "webp" (line 52) should be put before dependency "x11" (line 42)
  * C: 53: col 3: dependency "openjpeg" (line 53) should be put before dependency "x11" (line 42)
  * C: 54: col 3: dependency "fftw" (line 54) should be put before dependency "x11" (line 42)
  * C: 55: col 3: dependency "pango" (line 55) should be put before dependency "x11" (line 42)
  * C: 56: col 3: dependency "perl" (line 56) should be put before dependency "x11" (line 42)
  * New formulae should not have a HEAD spec
Error: 17 problems in 1 formula
```

GauthamGoli:
`brew audit --new-formula wget` worked fine , while `brew audit --new-formula imagemagick` strangely doesnt return. Am Looking into this right now.

GauthamGoli:
Found the issue. 

`next if fa.problems.empty?` should be replaced with 
`next if fa.problems.empty? && fa.new_formula_problems.empty?`

Opening a PR soon.

commitay:
`brew audit --new-formula python` is another with no output

GauthamGoli:
@commitay This issue shouldn't be affecting regular audits. Can you tell me which non new-formula audit weren't getting reported ? So that I can triage

commitay:
This build failed with a `brew test` error only.

https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/24451/version=high_sierra/console

It should have regular and new formula audit failures

GauthamGoli:
The exit code wasn't being set to a non-zero value. Conditions have been fixed in the PR referenced above.

commitay:
https://github.com/Homebrew/homebrew-core/pull/28141

This PR failed GitHub notability but the bot didn't comment on it?

commitay:
https://github.com/Homebrew/brew/pull/4201

GauthamGoli:
@MikeMcQuaid what was the issue..?

MikeMcQuaid:
@GauthamGoli `brew-test-bot` wasn't setting the correct environment variable (https://github.com/Homebrew/homebrew-test-bot/commit/b03e66770b07465d851bdab42f9650c0e24e72d3). Also, the GitHub API failure exception message was never shown (
https://github.com/Homebrew/brew/pull/4538) and the API token needed updated to have the correct scopes. After that: 🎉

commitay:
It's a hypothetical concern at the moment but having green CI when a formula has problems that should prevent it being merged might mean the comment doesn't get noticed and fixed in longer discussions?

MikeMcQuaid:
@commitay Personally I still prefer the approach before this PR but deferred to @ilovezfs who’s has been requesting this for a while.

ilovezfs:
The idea is that we should be able to expect all PRs to pass CI, including new formula PRs. If, for example, a new formula must depend on a system-duplicate formula to prevent build failure, the approach before this PR meant that we would have to merge the new formula PR despite its failing CI.

The new approach may ultimately prove unworkable, but it's worth trying given the confusion the prior approach has caused for some contributors who think red means something must not be merged, as is the case in many projects (e.g. those with protected branches).

PR #3840, opened 2018-02-27
Add macOS libs ahead of brewed llvm libs in lib search path

aw1621107:

-----
This ensures that libraries that are built with brewed LLVM but not
included in the Command Line Tools/Xcode (e.g. libomp) can be found
during a build, while still using system libraries for the essential
stuff (e.g. libc++)

Sorry it took me so long. Given how short this change is, I'm not sure why I haven't done this ages ago...

This should finally resolve [homebrew-core/pull/10456](https://github.com/Homebrew/homebrew-core/pull/10456) and #2234, and hopefully we can start using clang for OpenMP stuff too. Just got Fortran left...

### DIFF
--- a/Library/Homebrew/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/ENV/super.rb
@@ -187,7 +187,7 @@ module Superenv
       paths << Formula["llvm"].opt_lib.to_s
     end
 
-    paths << homebrew_extra_library_paths
+    paths += homebrew_extra_library_paths
     PATH.new(paths).existing
   end
 
--- a/Library/Homebrew/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/ENV/super.rb
@@ -176,6 +176,8 @@ module Superenv
     PATH.new(
       keg_only_deps.map(&:opt_lib),
       HOMEBREW_PREFIX/"lib",
+      "#{MacOS.sdk_path}/usr/lib",
+      (compiler == :llvm_clang ? Formula["llvm"].opt_lib.to_s : ""),
       homebrew_extra_library_paths,
     ).existing
   end
--- a/Library/Homebrew/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/ENV/super.rb
@@ -173,31 +173,22 @@ module Superenv
   end
 
   def determine_library_paths
+    paths = [
+      keg_only_deps.map(&:opt_lib),
+      HOMEBREW_PREFIX/"lib",
+    ]
+
     if compiler == :llvm_clang
       if MacOS::CLT.installed?
-        PATH.new(
-          keg_only_deps.map(&:opt_lib),
-          HOMEBREW_PREFIX/"lib",
-          "/usr/lib",
-          Formula["llvm"].opt_lib.to_s,
-          homebrew_extra_library_paths,
-        ).existing
+        paths << "/usr/lib"
       else
-        PATH.new(
-          keg_only_deps.map(&:opt_lib),
-          HOMEBREW_PREFIX/"lib",
-          "#{MacOS.sdk_path}/usr/lib",
-          Formula["llvm"].opt_lib.to_s,
-          homebrew_extra_library_paths,
-        ).existing
+        paths << "#{MacOS.sdk_path}/usr/lib"
       end
-    else
-      PATH.new(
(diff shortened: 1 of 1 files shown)


Library/Homebrew/extend/ENV/super.rb
```diff
@@ -176,6 +176,8 @@ def determine_library_paths
     PATH.new(
       keg_only_deps.map(&:opt_lib),
       HOMEBREW_PREFIX/"lib",
+      "#{MacOS.sdk_path}/usr/lib",
```
MikeMcQuaid: Are you 100% sure adding this isn't going to affect any other builds in any other way?

Library/Homebrew/extend/ENV/super.rb
```diff
@@ -176,6 +176,8 @@ def determine_library_paths
...
+      (compiler == :llvm_clang ? Formula["llvm"].opt_lib.to_s : ""),
```
MikeMcQuaid: Don't use a ternary here; split out this logic by building the `PATH` differently.

Library/Homebrew/extend/ENV/super.rb
MikeMcQuaid: Actually regardless: only do this on `llvm_clang` being the compiler; it's not worth the risk.

ilovezfs: I think if CLT is installed we want `"/usr/lib"` not `""#{MacOS.sdk_path}/usr/lib"`

ilovezfs: Also, I'm not sure what happens if we do that and have also set
```
ENV["SDROOT"] = MacOS.sdk_path
```
in a formula, whether they're going to be competing with each other.

ilovezfs:
@aw1621107 if the goal is primarily to be able to use llvm's libomp.dylib, it may be sufficient to
```
depends_on "libomp"
```
if we merge https://github.com/Homebrew/homebrew-core/pull/20589.

Or is there some additional benefit/reason to actually build with llvm_clang at the same time as using libomp.dylib?

Library/Homebrew/extend/ENV/super.rb
aw1621107: @ilovezfs If both CLT and Xcode are installed, which should be preferred?

Library/Homebrew/extend/ENV/super.rb
aw1621107: I'm guessing that a ternary also shouldn't be used for separate CLT/Xcode SDK include paths?

Library/Homebrew/extend/ENV/super.rb
ilovezfs: CLT

Library/Homebrew/extend/ENV/super.rb
MikeMcQuaid: Correct, thanks.

Library/Homebrew/extend/ENV/super.rb
MikeMcQuaid: Isn't `MacOS.sdk_path` empty or `/` on CLT anyway?

aw1621107:
@ilovezfs I'm not aware of a reason to require llvm_clang, but I'm not really familiar with OpenMP and if it requires the same version of the compiler/dylib. I figured that this might be more straightforward for users, since OpenMP support would be pretty much "invisible" in terms of special requirements, beyond using brew's llvm instead of Apple's.

MikeMcQuaid:
@aw1621107 Out of interest: do you have any use for `llvm_clang` except OpenMP?

fxcoudert:
Regarding libomp specifically: I'm wary of using libomp from one version of upstream's llvm, with Apple's different version of clang. It might be tricky ABI incompatibility, and I have not found any definite statement on ABI compatibility between versions. I also don't know of any project actually using this and having done serious testing, so we would need to triple-check on an extensive testsuite before we use it for OpenMP support in our formulas.

Library/Homebrew/extend/ENV/super.rb
aw1621107: I just ended up splitting it into another if/else. Don't know if the nesting will be an issue.

MikeMcQuaid:
@aw1621107 Made some tweaks. Thanks again!

PR #3848, opened 2018-02-27
language/python: handle python@* in each_python

ilovezfs:

-----

Needed for the python3 -> python, python -> python@2 rename.

This should handle both the old formula names and the new formula names since python@2 is an alias before the rename and python@3 is an alias after the rename.

See https://github.com/Homebrew/homebrew-core/pull/24604.


ilovezfs:
The
```
next if build.without? python_formula
```
seems to work. Not sure if it should actually be
```
next if build.without? python_formula.to_s
```
to be safer.

Issue #3849, opened 2018-02-27
Allow user to set HOMEBREW_CELLAR


cpcallen:
- [ ] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new. If it's a `brew cask` problem please file this issue at https://github.com/caskroom/homebrew-cask/issues/new. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew update` and can still reproduce the problem?
- [ ] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [ ] ran `brew config` and `brew doctor` and included their output with your issue?

# Detailed description of feature

Allow the user to specify the value of `HOMEBREW_CELLAR` by setting an environment variable of the same name.

# Motivation for the feature

In my organisation, mandatory access controls prevent Homebrew from functioning normally if installed with `HOMEBREW_PREFIX=/usr/local`; instead, we normally install it in `~/Homebrew` or `/usr/local/brew` instead.

Unfortunately, an unrelated system integrity protection system prevents executing any locally-compiled binaries or pre-built binaries installed anywhere other than the default path of `/usr/local/Cellar`.

# Alternative solutions

Two solutions are in use but neither is satisfactory:

   * Manually patch `brew.sh` with a hardcoded value for `HOMEBREW_CELLAR`, then set `HOMEBREW_NO_UPDATE_CLEANUP=1` to try to preserve this hack when `brew` updates itself, or
   * Symlink `$HOMEBREW_PREFIX/Cellar` to point at `/usr/local/Cellar`

The first is an ugly and fragile hack; the second causes avoidable `brew doctor` warnings and actually prevents installation (without `--force`) of certain bottles such as `openssl` that complain the Cellar is not in its default location (despite it *actually* being there).

# Relevancy

This is probably only of interest to a small minority of users, but the change is trivial (see PR #3850) and potentially useful to anyone who needs to maintain a non-standard configuration.

MikeMcQuaid:
[...]

PR #3850, opened 2018-02-27
Allow user to set HOMEBREW_CELLAR

cpcallen:
Use environment variable in preference to looking in usual locations.

-----

This implements #3849, allowing user to specify value of HOMEBREW_CELLAR via environment variable.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -5,12 +5,16 @@ then
 fi
 
 # Where we store built products; a Cellar in HOMEBREW_PREFIX (often /usr/local
-# for bottles) unless there's already a Cellar in HOMEBREW_REPOSITORY.
-if [[ -d "$HOMEBREW_REPOSITORY/Cellar" ]]
+# for bottles) unless there's already a Cellar in HOMEBREW_REPOSITORY or the
+# user has specified HOMEBREW_CELLAR explicitly.
+if [[ -z "$HOMEBREW_CELLAR" ]]
 then
-  HOMEBREW_CELLAR="$HOMEBREW_REPOSITORY/Cellar"
-else
-  HOMEBREW_CELLAR="$HOMEBREW_PREFIX/Cellar"
+    if [[ -d "$HOMEBREW_REPOSITORY/Cellar" ]]
+    then
+	HOMEBREW_CELLAR="$HOMEBREW_REPOSITORY/Cellar"
+    else
+	HOMEBREW_CELLAR="$HOMEBREW_PREFIX/Cellar"
+    fi
 fi
 
 case "$*" in


MikeMcQuaid:
Sorry, we do not wish to support the customisation of this variable. Please note for future reference if you're going to submit a PR anyway we don't need an issue opened first.

cpcallen:
> Sorry, we do not wish to support the customisation of this variable.

Why not?

I appreciate the benefits of encouraging everyone to use the standard configuration, but what purpose is served by making it unnecessarily difficult to use homebrew when the standard configuration is not viable?  Or to put it a different way: would you prefer that I fork Homebrew, or not use it at all?

MikeMcQuaid:
For anyone else following this PR but not the issue (https://github.com/Homebrew/brew/issues/3849):

> In my organisation, mandatory access controls prevent Homebrew from functioning normally if installed with HOMEBREW_PREFIX=/usr/local; instead, we normally install it in ~/Homebrew or /usr/local/brew instead.
> 
> Unfortunately, an unrelated system integrity protection system prevents executing any locally-compiled binaries or pre-built binaries installed anywhere other than the default path of /usr/local/Cellar

This issue is in essence that your organisation wishes to configure macOS in a way that is incompatible with Homebrew but still wishes to use Homebrew's binary packages with this hack. For Homebrew to merge this would be adding another configuration that is not tested by our CI and it's unclear what proportion of binary packages would work or not with this approach.

The options for your organisation are to either figure out how to make your existing setup work with Homebrew's binary packages or to build your own binary packages for Homebrew.

MikeMcQuaid:
> Or to put it a different way: would you prefer that I fork Homebrew, or not use it at all?

Finally on this: this is not a helpful or fair way of presenting your argument. Even if those are your options: that's up to you and your employer to decide.

alyssais:
As a maintainer, I volunteer my free time to try to keep Homebrew working well for as many of its (many, many) users as I can. In general, I do this work expecting nothing in return from any individual user.

However, when somebody wants to increase the maintenance cost of the software (and thus reduce the ability of myself and other maintainers to keep it working well for those many users) for a reason they refuse to even attempt to justify, it does not exactly encourage me to want to dedicate any of my time to helping them specifically at the expense of the vast majority of our users.

If you wish to either fork Homebrew or not use it all rather than fixing your broken macOS configuration (that, again, you have refused to even explain), it makes absolutely no difference to Homebrew’s volunteer maintainers, while taking a significant amount of effort from yourself and/or your employer for no good reason.

ilovezfs:
>Or to put it a different way: would you prefer that I fork Homebrew, or not use it at all?

Neither. I think you're best off continuing to use Homebrew and not forking it.

Building your own bottles that are for use in a prefix other than /usr/local and setting the `root_url` for your bottle hosting domain is relatively straightforward and painless using https://github.com/Homebrew/homebrew-test-bot and should be a scalable solution for your organization (better than having everyone build the non-relocatable bottles from source, obviously).

FYI, Facebook actually creates their own bottles for both macOS and Linux as part of their `osquery` build system. See https://github.com/facebook/osquery/tree/master/tools/provision/formula. I don't think you'll find this, or similar, difficult or costly to do.

PR #3858, opened 2018-03-02
utils: stop suggesting fixing hard deprecations

DomT4:

-----

```
Error: Calling 'brew linkapps' is deprecated!
There is no replacement.
/usr/local/Homebrew/Library/Homebrew/cmd/linkapps.rb:18:in `linkapps'
Or, even better, submit a PR to fix it!
```

I'm just a mere mortal, but I have yet to find a way to fix permanent death, so it seems reasonable to not expect folks to try and find a way.

I accept `linkapps` isn't an ideal example here & in some cases it _is appropriate_ to tell developers to submit fixes. This could/should _(open for debate)_ be scoped to formulae or even just exempt things in `cmd` and `dev-cmd`, but although I know the intention behind it I can imagine it would be confusing to someone newer to be told to fix something that's, well, never ever coming back.

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -109,7 +109,11 @@ def odeprecated(method, replacement = nil, disable: false, disable_on: nil, call
 
   if ARGV.homebrew_developer? || disable ||
      Homebrew.raise_deprecation_exceptions?
-    developer_message = message + "Or, even better, submit a PR to fix it!"
+    if replacement_message != "There is no replacement."
+      developer_message = message + "Or, even better, submit a PR to fix it!"
+    else
+      developer_message = message
+    end
     raise MethodDeprecatedError, developer_message
   elsif !Homebrew.auditing?
     opoo "#{message}\n"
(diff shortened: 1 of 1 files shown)


DomT4:
I've pushed a second commit doing this:

> even just exempt things in cmd and dev-cmd

Let me know if that's a better avenue and I'll squash things as necessary.

Library/Homebrew/utils.rb
```diff
@@ -109,7 +109,13 @@ def odeprecated(method, replacement = nil, disable: false, disable_on: nil, call
 
   if ARGV.homebrew_developer? || disable ||
      Homebrew.raise_deprecation_exceptions?
-    developer_message = message + "Or, even better, submit a PR to fix it!"
```
MikeMcQuaid: How about only setting this if `tap` and `replacement` are both set?

DomT4: 👍Happy to look at that.

DomT4:
@MikeMcQuaid Sorry! Hadn't even realised it'd been 5 days here 😓.

MikeMcQuaid:
@DomT4 No worries!

MikeMcQuaid:
Thanks @DomT4!

PR #3861, opened 2018-03-02
language/python: version 3.6 in site_packages

scpeters:

-----

The `python` formula now refers to python3, but there is a diagnostic complaint that recommends setting up a site-packages file that refers to python 2.7. So I've updated the `homebrew_site_packages` method. I have tested the changes locally.

There's potentially a similar change to be made at the following line, but I wasn't sure about it:

https://github.com/Homebrew/brew/blob/master/Library/Homebrew/language/python.rb#L140

~~~
++ brew doctor
working fine: please don't worry or file an issue; just ignore this. Thanks!

Warning: Your default Python does not recognize the Homebrew site-packages
directory as a special site-packages directory, which means that .pth
files will not be followed. This means you will not be able to import
some modules after installing them with Homebrew, like wxpython. To fix
this for the current user, you can run:
  mkdir -p /Users/jenkins/Library/Python/2.7/lib/python/site-packages
  echo 'import site; site.addsitedir("/usr/local/lib/python2.7/site-packages")' >> /Users/jenkins/Library/Python/2.7/lib/python/site-packages/homebrew.pth
~~~

### DIFF
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -9,7 +9,7 @@ module Language
       Version.create(version.to_s)
     end
 
-    def self.homebrew_site_packages(version = "2.7")
+    def self.homebrew_site_packages(version = "3.6")
       HOMEBREW_PREFIX/"lib/python#{version}/site-packages"
     end
 


MikeMcQuaid:
Thanks again @scpeters!

PR #3892, opened 2018-03-08
keg: `delete_pyc_files!` should also remove `__pycache__`

lembacon:

-----

As Python 3 becomes the default Python in the core tap, bottles built against Python 3 become a thing. Both `.pyc`and `.pyo` have already been removed from bottles, we should also remove `__pycache__`. Otherwise people who have `PYTHONDONTWRITEBYTECODE` set will find this really annoying.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -519,7 +519,7 @@ class Keg
 
   def delete_pyc_files!
     find do |path|
-      if %w[.pyc .pyo].include?(pn.extname) || pn.basename.to_s == "__pycache__"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks again @lembacon!

MikeMcQuaid:
Oh, this was your first Homebrew/brew PR! In that case: 
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @lembacon 😁

MikeMcQuaid:
FYI I reverted my two tweak commits because `__pycache__` is a directory containing those cache files so can only be removed after they are.

PR #3893, opened 2018-03-08
keg: don't delete main opt links of other formulae.

ilovezfs:

-----

If an old alias, according to the tab, of a formula being upgraded (e.g.
python) has the same name as the main opt link of some other formula
(e.g. python@2), don't remove the main opt link of that other formula.

Specifically,

  brew install python@2
  brew upgrade python

should not accidentally delete /usr/local/opt/python@2 even though
python@2 used to be an alias of python.

Fixes https://github.com/Homebrew/brew/pull/3893.

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -256,7 +256,11 @@ class Keg
 
     aliases.each do |a|
       alias_symlink = opt/a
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Looks good, thanks for the quick fix @ilovezfs!

PR #3908, opened 2018-03-11
Version: Reduce complexity of so many regex

kabel:

-----

Cleans up many of the regular expressions used to do version detection. Performance benefit is likely minimal, from local tests. This cleanup results in 3 formulae in the core tap having a [different detected version](https://gist.github.com/kabel/7204d68977a1380814c75086861fe8b8#file-official-versions-diff-csv), although I don't think it's a significant difference. The cleanup also fixes version detection for 133 core formulae, which could remove the version definition now.

I tried to add tests to cover more of the more unique URL/version formats, but adding all 133 from the newly detected core formulae seemed like test overkill. To be more certain I wasn't regressing detection, I used a [custom brew command](https://gist.github.com/kabel/7204d68977a1380814c75086861fe8b8#file-dump-versions-rb) to dump the versions for all formulae. My [results](https://gist.github.com/kabel/7204d68977a1380814c75086861fe8b8#file-official-versions-audit-test-csv) are also available on the gist.

### DIFF
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -218,6 +218,22 @@ class Version
     version.nil? ? NULL : new(version)
   end
 
+  COMMON_VERSION_PATTERN = /(?:(?:2\.1[12]|2\.20\.11)-|%\d{2})?((?:\d+\.)+\d+(?:[-_.~+]?(?:(?:alpha|beta|preview|pre|rc|cr|final|dev|snapshot|git|gfm)(?:\.?\d+)*|(?:ga|rel|build|[abmprv])?\.?(?:\d+\.)*\d+)|[a-f0-9]{6,}|[a-z0-9]{1,2})?)/i
+  #                         ^----------------A----------------^^^-----C-----^^--^--E---^^--^---------------------------------G---------------------------------^ ^---------------------H--------------------^ ^----I-----^ ^-----J-----^^^^
+  #                                                            |             |          +-----------------------------------------------------------------------F-----------------------------------------------------------------------+||
+  #                                                            |             +-----------------------------------------------------------------------------D-----------------------------------------------------------------------------+|
+  #                                                            +------------------------------------------------------------------------------------B-------------------------------------------------------------------------------------+
+  # A: Known version-like prefixes that are _not_ part of version
+  # B: Captured version
+  # C: Dotted digit series
+  # D: Version suffix & separator
+  # E: Version suffix separator
+  # F: Suffix alternatives
+  # G: String suffixes with optional dotted digits
+  # H: String suffixes with manditory digits
+  # I: 6+ hexadecimal character suffix
+  # J: 1-2 alphanumeric character suffix
+
   def self._parse(spec)
     spec = Pathname.new(spec) unless spec.is_a? Pathname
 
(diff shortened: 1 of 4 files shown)


Library/Homebrew/version.rb
```diff
@@ -218,6 +218,22 @@ def self.parse(spec)
     version.nil? ? NULL : new(version)
   end
 
+  COMMON_VERSION_PATTERN = /(?:(?:2\.1[12]|2\.20\.11)-|%\d{2})?((?:\d+\.)+\d+(?:[-_.~+]?(?:(?:alpha|beta|preview|pre|rc|cr|final|dev|snapshot|git|gfm)(?:\.?\d+)*|(?:ga|rel|build|[abmprv])?\.?(?:\d+\.)*\d+)|[a-f0-9]{6,}|[a-z0-9]{1,2})?)/i
```
MikeMcQuaid: Sorry, I don't think this is an improvement, it's almost impossible to follow.

kabel: That's fair, and why I put in the description comments below. Is there anything specific that makes it impossible to follow, or is it just that it's large with a bunch of alternatives?

alyssais: Could we use a `//x` regular expression, across multiple lines with inline comments, to improve readability?

ilovezfs: This still seems like a regression in comprehensibility even if it's split over multiple lines.

MikeMcQuaid: It's just too long and will be extremely difficult to debug or modify (particularly compared to the existing code).

MikeMcQuaid:
> The cleanup also fixes version detection for 133 core formulae, which could remove the version definition now.

It would be great to get these hit another way 👍

kabel:
I understand the expression, at first glance, is a bit intimidating. However, I think we can strike a balance of maintaining the ability to debug and modify and reducing the amount of times it's necessary to call `#match`. Even if the performance gain is minimal, it's still a factor and the numerous amount of patterns make it difficult for someone to know which pattern was used to detect a given URL. So the trade-off is to get folks to stop adding new expressions and to start refining the ones that exist.

There are tools to help developers test and maintain regular expressions (I use [regex101](https://regex101.com/r/qbIMDh/1) all the time), so it seems a little confusing why we should avoid using them to their full potential when they're a bit cumbersome (long) to read.

I would consider @alyssais's recommendation or perhaps building the `Regexp` from a string algorithm. Are either of the following any easier to read/maintain? 

```rb
version_pattern = /(?: # [GROUP] optional version-like prefix to not capture
  (?: # product-specific prefixes followed by dash
    2\.1[12]
    |
    2\.20\.11
  )-
  |
  %\d{2} # URL encoded character
  )?
  ( # [GROUP] captured version
    (?:\d+\.)+\d+ # at least major.minor digits
    (?: # [GROUP] optional version suffix
      [-_.~+]? # optional separator character
      (?: # [GROUP] suffixes that can have a separator
        # common pre-releases with optional dotted digits
        (?:alpha|beta|preview|pre|rc|cr|final|dev|snapshot|git|gfm) 
        (?:\.?\d+)*
        |
        # other releases with required dotted digits
        (?:ga|rel|build|[abmprv])?
        (?:\.?\d+)+
        |
        # at least 6 hex characters (commit-like)
        [a-f0-9]{6,}
      )
      |
      # 1-2 alphanumeric character (openssl-like: 1.0.2n)
      [a-z0-9]{1,2}
    )?
  )
/ix
```

It's certainly a bummer not being able to copy the expression (from source below) into a testing tool.

```rb
# Prefixes that look like versions, but should not be captured
# e.g. 2.11- 2.12- 2.20.11-
# e.g. %20
ignore_version_like_alts = %w[
  (?:2\.1[12]|2\.20\.11)-
  %\d{2}
]. join "|"

# Major.minor required: 1.0
dotted_digits = '(?:\d+\.)+\d+'

# Digits with an optional dot prefix
# e.g. 34 .3
opt_dot_digits = '(?:\.?\d+)'

# Suffix separator characters
suffix_sep = '[-_.~+]'

# Suffixes that can have separator
# e.g. beta beta1 beta.1 beta1.1 beta.1.1
# e.g. p1 p.1 p1.1 p.1.1
# e.g. 123ABC CBA321
suffix_with_sep_alts = %W[
  (?:alpha|beta|preview|pre|rc|cr|final|dev|snapshot|git|gfm)#{opt_dot_digits}*
  (?:ga|rel|build|[abmprv])?#{opt_dot_digits}+
  [a-f0-9]{6,}
].join "|"

# Suffix that cannot have separator
suffix_no_sep = '[a-z0-9]{1,2}'

suffixes = %W[
  #{suffix_sep}?(?:#{suffix_with_sep_alts})
  #{suffix_no_sep}
].join "|"

version_pattern_str = %W[
  (?:#{ignore_version_like_alts})?
  (#{dotted_digits}(?:#{suffixes})?)
].join
version_pattern = Regexp.new(version_pattern_str, true)
```

MikeMcQuaid:
> I understand the expression, at first glance, is a bit intimidating. However, I think we can strike a balance of maintaining the ability to debug and modify and reducing the amount of times it's necessary to call #match. Even if the performance gain is minimal, it's still a factor and the numerous amount of patterns make it difficult for someone to know which pattern was used to detect a given URL. So the trade-off is to get folks to stop adding new expressions and to start refining the ones that exist.

It's very, very rarely worth trading readability for performance. Users (and maintainers) would rather something that worked correctly but took a little longer. In this case it's definitely not worth the trade-off because it's not in a part of the code that causes measurable delays for users. Given the complexity of this regex I would not be surprised (although I'm not claiming it's the case) if it actually ends up taking the same amount or more time than the existing code.

In short, if you're going to work on performance improvements start with `brew prof` to figure out what's slow and demonstrate a measurable improvement in total command execution. In this case that would be measurably (i.e. >=5% improvement) in a `brew` command.

MikeMcQuaid:
Sorry @kabel, passing on this.

kabel:
That's fine. I'll come back when I have more time to get the new detected versions using a different approach.

PR #3911, opened 2018-03-12
build: do not skip runtime deps of buildtime deps

### LINKED ISSUE #3883: Run-time deps of build-time deps aren't available to builds

- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new. If it's a `brew cask` problem please file this issue at https://github.com/caskroom/homebrew-cask/issues/new. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [ ] ran `brew config` and `brew doctor` and included their output with your issue?

use `depends_on "opam" => :build` and not also have to use `depends_on "aspcud" => :build`

`aspcud` is not available to the build unless `aspcud` is itself added as a build-time dependency, even though it's a run-time dependency of `opam`.

- What you expected to happen

`/usr/local/opt/aspcud/bin` to be in the `PATH`

```
iMac-TMP:homebrew-core joe$ git diff -- Formula/corectl.rb 
diff --git a/Formula/corectl.rb b/Formula/corectl.rb
index d21c3c9..5f168d7 100644
--- a/Formula/corectl.rb
+++ b/Formula/corectl.rb
@@ -22,6 +22,7 @@ class Corectl < Formula
   depends_on :macos => :yosemite
 
   def install
+raise "PATH: #{ENV["PATH"]}"
     ENV["GOPATH"] = buildpath
 
     path = buildpath/"src/github.com/TheNewNormal/#{name}"
iMac-TMP:homebrew-core joe$ brew install -s corectl
==> Downloading https://github.com/TheNewNormal/corectl/archive/v0.7.18.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/corectl-0.7.18.tar.gz
Error: PATH: /usr/local/Homebrew/Library/Homebrew/shims/super:/usr/local/opt/go/bin:/usr/local/opt/ocaml/bin:/usr/local/opt/aspcud/bin:/usr/local/opt/opam/bin:/usr/bin:/bin:/usr/sbin:/sbin
iMac-TMP:homebrew-core joe$ brew edit corectl
Editing /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/corectl.rb
Warning: Using vim because no editor was set in the environment.
This may change in the future, so we recommend setting EDITOR,
or HOMEBREW_EDITOR to your preferred text editor.
iMac-TMP:homebrew-core joe$ git diff -- Formula/corectl.rb 
diff --git a/Formula/corectl.rb b/Formula/corectl.rb
index d21c3c9..17fba92 100644
--- a/Formula/corectl.rb
+++ b/Formula/corectl.rb
@@ -17,11 +17,11 @@ class Corectl < Formula
   depends_on "go" => :build
   depends_on "libev"
   depends_on "ocaml" => :build
-  depends_on "aspcud" => :build
   depends_on "opam" => :build
   depends_on :macos => :yosemite
 
   def install
+raise "PATH: #{ENV["PATH"]}"
     ENV["GOPATH"] = buildpath
 
     path = buildpath/"src/github.com/TheNewNormal/#{name}"
iMac-TMP:homebrew-core joe$ brew install -s corectl
==> Downloading https://github.com/TheNewNormal/corectl/archive/v0.7.18.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/corectl-0.7.18.tar.gz
Error: PATH: /usr/local/Homebrew/Library/Homebrew/shims/super:/usr/local/opt/go/bin:/usr/local/opt/ocaml/bin:/usr/local/opt/opam/bin:/usr/bin:/bin:/usr/sbin:/sbin
iMac-TMP:homebrew-core joe$ 
```

stek29:

-----

Fix #3883
Revert 2fb5ead38a7632ff6f941909474767883a9e0252

I'm not sure on how to write tests for it without introducing a new fixture Formula with deps layout needed for test.

### DIFF
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -59,8 +59,6 @@ class Build
         Dependency.prune
       elsif dep.build? && dependent != formula
         Dependency.prune
-      elsif dep.build?
-        Dependency.keep_but_prune_recursive_deps
       end
     end
   end


Library/Homebrew/test/dependency_expansion_spec.rb
```diff
@@ -101,18 +101,6 @@ def build_dep(name, tags = [], deps = [])
     expect(deps).to eq([bar, baz])
   end
 
-  it "keeps dependency but prunes recursive dependencies with ::keep_but_prune_recursive_deps" do
```
MikeMcQuaid: Could you write/modify a test for the reverse instead i.e. verify that these aren't pruned?

stek29: This test was removed not because of fix itself but because the method it was using was removed.

I tried to (by making a test which should fail with old build.rb but pass with new one), but it didn't work.
I guess it should be in some other place, since the change to be tested is actually `Build#expand_deps` and not something in `dependency_expansion`

MikeMcQuaid: > I guess it should be in some other place, since the change to be tested is actually Build#expand_deps and not something in dependency_expansion

👍 That's be great.

ilovezfs:
Thanks for working on this @stek29! You're awesome

Library/Homebrew/test/dependency_expansion_spec.rb
stek29: But I'm not sure on how to test that -- I guess a new spec needs to be created for "Build"?

stek29:
Speaking of testing `build.rb` -- it starts with `trap`, and relies on Formula, BuildOptions and ARGV -- are there any good examples on tests for such classes/files?

Library/Homebrew/test/dependency_expansion_spec.rb
MikeMcQuaid: @stek29 Actually you can leave this as-is; I will create another PR that plays around a bit with this first. Nice work again!

stek29: @MikeMcQuaid Sure.

Library/Homebrew/dependency.rb
```diff
@@ -123,11 +121,6 @@ def skip
       throw(:action, :skip)
     end
 
-    # Keep a dependency, but prune its dependencies
-    def keep_but_prune_recursive_deps
```
MikeMcQuaid: @stek29 Leave all definitions of this method/test as-is: I'll use them on https://github.com/Homebrew/brew/pull/3914.

stek29: @MikeMcQuaid Sure

stek29:
Oh, actually `keep_but_prune_recursive_deps` is used in `cmd/uses.rb`, but I didn't catch that (because there's no testing on cmds? :x )

MikeMcQuaid:
Thanks again @stek29!

> Oh, actually keep_but_prune_recursive_deps is used in cmd/uses.rb, but I didn't catch that (because there's no testing on cmds? :x )

Yeh, our testing isn't comprehensive enough 😭. We'd love your help with that!

ilovezfs:
@stek29 unfortunately, I needed to revert this in https://github.com/Homebrew/brew/pull/3952. Please see my comment here:
https://github.com/Homebrew/brew/pull/3952#issuecomment-374922680

Issue #3891, opened 2018-03-07
Allow downloading bottles via ssh


ryangreenberg:
## Motivation
I would like to make it easy to install bottles that are stored in a secure location to further enable internal/corporate use of homebrew. I propose doing this by introducing a file download strategy that uses `scp`.

## Detailed Description

@eedwards-sk's work in #3841 will make it possible to specify download strategies
for bottles. Once that change is merged, you will be able to use `S3DownloadStrategy`,
`GitHubPrivateRepositoryDownloadStrategy`, or
`GitHubPrivateRepositoryReleaseDownloadStrategy` to download bottles from protected locations. A disadvantage to these, however, is that they require
individual developers to have the appropriate environment variables set
(e.g. `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` or `HOMEBREW_GITHUB_API_TOKEN`). This in turns means those credentials have to be provisioned and distributed to individual developer's machines.

Most companies already distribute and maintain ssh keys and configuration for developers, which makes it a natural fit for downloading internal files securely. I would like to introduce an `ScpFileDownloadStrategy` which would support fetching bottles over ssh via `scp`.

The schema for this strategy would be `scp://` and the format would be:

```
scp://(optional user)@host(:optional port)/path
```

The `fetch` method of `ScpFileDownloadStrategy` would translate this to `scp [-P optional port] host:/path`

In a formula this might look like:

```
bottle do
  root_url "scp://bottles.company.com/home/homebrew-bottles", :using => ScpFileDownloadStrategy
  ...
  sha256 "3a75f421c2b52c0fd0ee6e16a3adb6ecc4ac9ed8b4b73182515ff28b7a12a656" => :sierra
  sha256 "cce3572ec0448e5dbbd3719870c15cc534a0b3dd9cfdd048610e8234302f6e1a" => :high_sierra
end
```

## Relevance
The feature template asks "how the feature would be relevant to at least 90% of Homebrew users" and I admit that this is a niche addition (on top of internal/corporate use of Homebrew which is itself probably a niche). My apologies if this seems like a waste of time.

## Alternatives
Write this strategy, distribute it with our internal formulae, and `require` it at the top.

---

I would be happy to create a pull request for this change if the proposal seems acceptable.

MikeMcQuaid:
> I would be happy to create a pull request for this change if the proposal seems acceptable.

Yep, this sounds good. If there's a PR with sufficient test coverage we'd consider accepting this.

PR #3938, opened 2018-03-16
brew: Add ScpDownloadStrategy

ryangreenberg:
## Checklist

## Description

This PR implements the change proposed in #3891.

> I would like to make it easy to install bottles that are stored in a secure location to further enable internal/corporate use of homebrew. I propose doing this by introducing a file download strategy that uses scp.

The new `ScpDownloadStrategy` allows users to download source code (and with #3841, bottles) using scp. Here's what it looks like in practice:

```
$ head ./etl.rb 
class Etl < Formula
  # Note URL and :using => ScpDownloadStrategy
  url "scp://example.slack.com/homebrew/etl-0.04.19.tar.gz", :using => ScpDownloadStrategy
  sha256 "ba944c1a07fd321488f9d034467931b8ba9e48454abef502a633ff4835380c1c"

$ ./bin/brew fetch ./etl.rb
==> Downloading scp://example.slack.com/homebrew/etl-0.04.19.tar.gz
etl-0.04.19.tar.gz                    100%  363KB 701.9KB/s   00:00
Downloaded to: /Users/greenberg/Library/Caches/Homebrew/etl-0.04.19.tar.gz
SHA256: ba944c1a07fd321488f9d034467931b8ba9e48454abef502a633ff4835380c1c
```

As a convenient accident, since scp is run via `system`, it prompts the user for two-factor authentication or other credentials if required.

A couple implementation notes:

- I reused `CurlDownloadStrategyError` when the URL pattern doesn't match. That's not really accurate. I can introduce a new name, simply use `ArgumentError`, or leave as-is.
- I did not add `scp` to `DownloadStrategyDetector`.


Library/Homebrew/download_strategy.rb
```diff
@@ -608,6 +608,59 @@ def fetch_release_metadata
...
+
+  def initialize(name, resource)
+    super
+    @tarball_path = HOMEBREW_CACHE/"#{name}-#{version}#{ext}"
+    @temporary_path = Pathname.new("#{cached_location}.incomplete")
+    parse_url_pattern
+  end
+
+  def parse_url_pattern
+    url_pattern = %r{scp://([^@]+@)?([^@:/]+)(:[0-9]+)?/(\S+)}
```
MikeMcQuaid: This regex is a bit gnarly. Can it e simplified at all? e.g. at least `[0-9]` can be `\d`

Library/Homebrew/download_strategy.rb
```diff
@@ -608,6 +608,59 @@ def fetch_release_metadata
...
+    unless @url =~ url_pattern
```
MikeMcQuaid: Use `if` and `!~`

Library/Homebrew/download_strategy.rb
```diff
@@ -608,6 +608,59 @@ def fetch_release_metadata
...
+
+  def clear_cache
+    super
+    rm_rf(temporary_path)
+  end
+
+  private
+
+  def scp_source
+    path_prefix = @path.start_with?("~") ? "" : "/"
```
MikeMcQuaid: `path_prefix = "/" if @path.start_with?("~")`

Library/Homebrew/download_strategy.rb
```diff
@@ -608,6 +608,59 @@ def fetch_release_metadata
...
+    port_arg = @port ? "-P #{@port[1..-1]} " : ""
```
MikeMcQuaid: `port_arg = "-P #{@port[1..-1]} " if @port`

Library/Homebrew/test/download_strategies_spec.rb
```diff
@@ -224,6 +224,72 @@ def setup_git_repo
...
+      http://example.com/foo.tar.gz
+      scp://@example.com/foo.tar.gz
+      scp://example.com:/foo.tar.gz
+      scp://example.com
+    ]
+
+    invalid_urls.each do |invalid_url|
+      context "with invalid URL #{invalid_url}" do
+        it "raises CurlDownloadStrategyError" do
+          expect { described_class.new(name, resource_for(invalid_url)) }.to raise_error(CurlDownloadStrategyError)
```
MikeMcQuaid: Split this (and the other really long lines) only multiple lines.

Library/Homebrew/download_strategy.rb
ryangreenberg: I'll change `[0-9]` to `\d`.

As far as simplification, I'm not sure how much I can do to change it and still extract the desired parts. Two options:
1. Use whitespace mode and document the regex with comments
2. Put the regex in a constant and test it specifically so people changing it can at least see what behavior they need to maintain (this is already done indirectly via the tests for `fetch` but I'd be happy to move them to make it easier to understand/change)

Library/Homebrew/download_strategy.rb
ryangreenberg: Nifty—never thought of doing it that way.

ryangreenberg: Not sure if this changes your guidance, but I realized this would be `path_prefix = "/" unless @path.start_with("~")`

Library/Homebrew/download_strategy.rb
```diff
@@ -608,6 +610,60 @@ def fetch_release_metadata
...
+
+    if cached_location.exist?
+      puts "Already downloaded: #{cached_location}"
+      return
+    end
+    
+    begin
+      safe_system "scp", scp_source, temporary_path.to_s
+    rescue ErrorDuringExecution
+      raise CurlDownloadStrategyError, @url
```
MikeMcQuaid: As you've mentioned: this should be a different exception.

Library/Homebrew/download_strategy.rb
```diff
@@ -608,6 +610,60 @@ def fetch_release_metadata
...
+
+  def initialize(name, resource)
+    super
+    @tarball_path = HOMEBREW_CACHE/"#{name}-#{version}#{ext}"
+    @temporary_path = Pathname.new("#{cached_location}.incomplete")
+    parse_url_pattern
+  end
+
+  def parse_url_pattern
+    url_pattern = %r{scp://([^@]+@)?([^@:/]+)(:\d+)?/(\S+)}
```
MikeMcQuaid: Can you supply an example test formula as a comment? I'll obviously have to edit it to match something I have `scp` access to but it would be helpful.

ryangreenberg:
OK, I'm back post-merge-conflicts-resolution! Thanks for your help on this PR, @MikeMcQuaid. Let me know if there's anything else you'd like to see or think is missing.

On a side note, if you think now (or later) that having this extra download strategy is a burden, I think you could probably remove it and ask people who want to use it to require it in their own formula.

Library/Homebrew/download_strategy.rb
ryangreenberg: Done!

PR #3939, opened 2018-03-16
Don't use Linuxbrew/homebrew-core (yet, at least).

MikeMcQuaid:
This can't be tapped on vanilla Homebrew/brew because things like e.g. `GlibcRequirement` are missing. We will put this back when the Linuxbrew to Homebrew migration is complete.

CC @sjackman @maxim-belkin FYI


sjackman:
👍 Thanks for the heads up, Mike.

PR #3941, opened 2018-03-17
fix: crash when curl is not installed

bukzor:
Migrated from https://github.com/Linuxbrew/brew/issues/640

### DIFF
--- a/Library/Homebrew/utils/curl.rb
+++ b/Library/Homebrew/utils/curl.rb
@@ -6,8 +6,8 @@ def curl_executable
     ENV["HOMEBREW_CURL"],
     which("curl"),
     "/usr/bin/curl",
-  ].map { |c| Pathname(c) }.find(&:executable?)
-  raise "curl is not executable" unless @curl
+  ].compact.map { |c| Pathname(c) }.find(&:executable?)
+  raise "no executable curl was found" unless @curl
   @curl
 end
 


Library/Homebrew/utils/curl.rb
```diff
@@ -6,8 +6,8 @@ def curl_executable
     ENV["HOMEBREW_CURL"],
     which("curl"),
     "/usr/bin/curl",
-  ].map { |c| Pathname(c) }.find(&:executable?)
-  raise "curl is not executable" unless @curl
+  ].reject(&:nil?).map { |c| Pathname(c) }.find(&:executable?)
```
MikeMcQuaid: `.reject(&:nil?)` should be `.compact`

Library/Homebrew/utils/curl.rb
```diff
@@ -6,8 +6,8 @@ def curl_executable
...
+  raise "brew requires curl, which is not available" unless @curl
```
MikeMcQuaid: `no executable curl was found`

Library/Homebrew/utils/curl.rb
bukzor: My ruby-fu is quite weak. I've changed it but haven't tested.

Library/Homebrew/utils/curl.rb
```diff
@@ -6,8 +6,8 @@ def curl_executable
     ENV["HOMEBREW_CURL"],
     which("curl"),
     "/usr/bin/curl",
-  ].map { |c| Pathname(c) }.find(&:executable?)
-  raise "curl is not executable" unless @curl
+  ].compact.map { |c| Pathname(c) }.find(&:executable?)
+  raise "brew requires curl, but no executable curl was found" unless @curl
```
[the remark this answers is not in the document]
bukzor: Done.

MikeMcQuaid:
Thanks again @bukzor!

PR #3943, opened 2018-03-18
fix clang build version on Ubuntu Artful

bukzor:
Ported from https://github.com/Linuxbrew/brew/pull/621

Before:

```
$ brew config | grep Clang:
Clang: 4.0 build 
```

After:

```
$ brew config | grep Clang:     
Clang: 4.0 build 401
```

### DIFF
--- a/Library/Homebrew/development_tools.rb
+++ b/Library/Homebrew/development_tools.rb
@@ -74,7 +74,7 @@ class DevelopmentTools
     def clang_build_version
       @clang_build_version ||= begin
         if (path = locate("clang")) &&
-           build_version = `#{path} --version`[/clang-(\d{2,})/, 1]
+           build_version = `#{path} --version`[%r{clang(-| version [^ ]+ \(tags/RELEASE_)(\d{2,})}, 2]
           Version.new build_version
         else
           Version::NULL
--- a/Library/Homebrew/system_config.rb
+++ b/Library/Homebrew/system_config.rb
@@ -212,16 +212,15 @@ class SystemConfig
       f.puts "GCC-4.2: build #{gcc_4_2}" unless gcc_4_2.null?
       f.print "Clang: "
       if clang.null?
-        f.print "N/A"
+        f.puts "N/A"
       else
         f.print "#{clang} build "
         if clang_build.null?
-          f.print "(parse error)"
+          f.puts "(parse error)"
         else
-          f.print clang_build
+          f.puts clang_build
         end
       end
-      f.print "\n"
       f.puts "Git: #{describe_git}"
       f.puts "Curl: #{describe_curl}"
       f.puts "Perl: #{describe_perl}"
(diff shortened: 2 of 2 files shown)


Library/Homebrew/development_tools.rb
```diff
@@ -74,7 +74,7 @@ def clang_version
     def clang_build_version
       @clang_build_version ||= begin
         if (path = locate("clang")) &&
-           build_version = `#{path} --version`[/clang-(\d{2,})/, 1]
+          build_version = `#{path} --version`[/clang(-|.*RELEASE_)(\d{2,})/, 2]
```
MikeMcQuaid: This indentation change is incorrect; it's part of the `if` statement above. Also, what is present in the `.*` before `RELEASE`?

bukzor: The string is `clang version 5.0.1 (tags/RELEASE_501/final)`.
According to the pastebins I was able to find, the "501" is what's found under osx.

MikeMcQuaid: Cool, thanks. What about making it `tags/RELEASE_` instead of `.*`?

bukzor: I was less certain that would continue to match, and I don't think the current pattern will over-match.

MikeMcQuaid: @bukzor I'd rather we go for something more conservative for now and extended when needed. Thanks!

bukzor: Can dooo!

Library/Homebrew/system_config.rb
```diff
@@ -210,7 +210,7 @@ def dump_verbose_config(f = $stdout)
       f.puts "Homebrew Ruby: #{describe_homebrew_ruby}"
       f.puts "GCC-4.0: build #{gcc_4_0}" unless gcc_4_0.null?
       f.puts "GCC-4.2: build #{gcc_4_2}" unless gcc_4_2.null?
-      f.puts "Clang: #{clang.null? ? "N/A" : "#{clang} build #{clang_build}"}"
+      f.puts "Clang: #{clang.null? ? "N/A" : "#{clang} build #{clang_build.null? ? "???" : clang_build}"}"
```
bukzor: The intent is to make it more noticeable when the above regex doesn't match. I think it does a good job of that.
The output in that case looks like so:

```
Clang: 5.0 build ???
```

sjackman: I'd suggest `Clang: 5.0 build N/A`.

bukzor: "Not applicable" is the correct answer when a tool is not installed, which is its usage in the rest of this file, but the build number is indeed applicable when clang is installed, as it is here.

sjackman: Fair enough. I don't feel strongly.

MikeMcQuaid:
@bukzor You can reproduce those errors locally with `brew style`.

Library/Homebrew/system_config.rb
bukzor: <3

bukzor: I changed the "???" string to "(parse error)", which should be more clear.

bukzor:
I'm not 100% happy with the result, but `brew style` is happy now.

MikeMcQuaid:
Thanks again @bukzor!

bukzor:
Thank you too @MikeMcQuaid. It's a pleasure to work with you all. You and the rest of the homebrew team are a role model for open-source upstreams, and collaborative software in general.

MikeMcQuaid:
@bukzor Thanks for the kind words and taking the time to say them. We end up dealing with a lot of bile so comments like this are touching ❤️

PR #3944, opened 2018-03-18
/usr/bin/which -> type -P

bukzor:
This allows odd POSIXen (i.e. Android) to (at least attempt to) function.

Ported from Linuxbrew/brew#621

### DIFF
--- a/Library/Homebrew/shims/scm/git
+++ b/Library/Homebrew/shims/scm/git
@@ -85,10 +85,10 @@ fi
 
 case "$(lowercase "$SCM_FILE")" in
   git)
-    [[ -n "$HOMEBREW_GIT" ]] && safe_exec "$(which "$HOMEBREW_GIT")" "$@"
+    [[ -n "$HOMEBREW_GIT" ]] && safe_exec "$(type -P "$HOMEBREW_GIT")" "$@"
     ;;
   svn)
-    [[ -n "$HOMEBREW_SVN" ]] && safe_exec "$(which "$HOMEBREW_SVN")" "$@"
+    [[ -n "$HOMEBREW_SVN" ]] && safe_exec "$(type -P "$HOMEBREW_SVN")" "$@"
     ;;
 esac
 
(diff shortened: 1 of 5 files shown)


Library/Homebrew/shims/scm/git
```diff
@@ -99,7 +99,7 @@ brew_repo_version="$(quiet_safe_cd "$SCM_DIR/../../../../bin" && pwd -P)/$SCM_FI
 safe_exec "$brew_repo_version" "$@"
 
 IFS=$'\n'
-for path in $(/usr/bin/which -a "$SCM_FILE" 2>/dev/null)
+for path in $(command -v "$SCM_FILE" 2>/dev/null)
```
MikeMcQuaid: This isn't the same behaviour as `which -a`; it produces a single result rather than multiple.

sjackman:
I suggest using the bash builtin `type -ap` rather than `which -a`. This script is `#!/bin/bash`.
```
❯❯❯ type type which
type is a shell builtin
which is hashed (/usr/bin/which)
❯❯❯ which -a git
/usr/local/bin/git
/usr/bin/git
❯❯❯ type -ap git
/usr/local/bin/git
/usr/bin/git
```

MikeMcQuaid:
> I suggest using the bash builtin type -ap rather than the external command which -a. This script is #!/bin/bash.

Works for me. There may be other locations where that would be useful/needed too.

MikeMcQuaid:
> Do you like bash-isms in this project?

Yup, we use a bunch and we require `/bin/bash` 👍

maxim-belkin:
To be absolutely sure that you're using a builtin, a not `/usr/bin/type` alter-ego, prefix the command with `builtin`: `builtin type -ap` :)

MikeMcQuaid:
Thanks again @bukzor!

bukzor:
@maxim-belkin Is there any case where $PATH will override a built-in? I was pretty sure that `builtin` is only useful for avoiding a function or alias of the same name.

Also, I can't find any mention of `/usr/bin/type`. What is it?

maxim-belkin:
builtins can be disabled with `enable -n` but otherwise - you're correct.
On my mac, there is a bunch of `/usr/bin/<filename>` with the following contents:
```bash
#!/bin/sh
# $FreeBSD: src/usr.bin/alias/generic.sh,v 1.2 2005/10/24 22:32:19 cperciva Exp $
# This file is in the public domain.
builtin `echo ${0##*/} | tr \[:upper:] \[:lower:]` ${1+"$@"}
```
and `type` is one of them...
See https://stackoverflow.com/a/38809607

MikeMcQuaid:
I'm happy with this as-is. We can investigate further if we hit any bugs.

PR #3949, opened 2018-03-20
Upgraded use of AWS Ruby SDK for S3

amancevice:
When `S3DownloadStrategy` is detected in `DownloadStrategyDetector`, an attempt is made to install & require the `aws-sdk-s3` gem.

The AWS S3 SDK is used to sign the URL and handed off to `curl_download`.

This PR is meant as a companion to https://github.com/Homebrew/brew/pull/3841.

ATTN @MikeMcQuaid, I couldn't figure out how to do this in `formula_installer.rb` because I don't think the `formula` object has access to the `@url` of the formula at that time. However, I think putting it in the `DownloadStrategyDetector` effectively front-runs the sandbox.

I'm a newbie when it comes to writing rspec tests so I've done a pretty barebones job at this. If I get some guidance I can expand the tests.

-----

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -473,16 +473,6 @@ end
 # distribution.  (It will work for public buckets as well.)
 class S3DownloadStrategy < CurlDownloadStrategy
   def _fetch
-    # Put the aws gem requirement here (vs top of file) so it's only
-    # a dependency of S3 users, not all Homebrew users
-    require "rubygems"
-    begin
-      require "aws-sdk-v1"
-    rescue LoadError
-      onoe "Install the aws-sdk gem into the gem repo used by brew."
-      raise
-    end
-
     if @url !~ %r{^https?://([^.].*)\.s3\.amazonaws\.com/(.+)$}
       raise "Bad S3 URL: " + @url
     end
@@ -492,12 +482,12 @@ class S3DownloadStrategy < CurlDownloadStrategy
     ENV["AWS_ACCESS_KEY_ID"] = ENV["HOMEBREW_AWS_ACCESS_KEY_ID"]
     ENV["AWS_SECRET_ACCESS_KEY"] = ENV["HOMEBREW_AWS_SECRET_ACCESS_KEY"]
 
-    obj = AWS::S3.new.buckets[bucket].objects[key]
     begin
-      s3url = obj.url_for(:get)
-    rescue AWS::Errors::MissingCredentialsError
+      signer = Aws::S3::Presigner.new
+      s3url = signer.presigned_url :get_object, bucket: bucket, key: key
+    rescue Aws::Sigv4::Errors::MissingCredentialsError
       ohai "AWS credentials missing, trying public URL instead."
-      s3url = obj.public_url
+      s3url = @url
     end
 
     curl_download s3url, to: temporary_path
@@ -1116,6 +1106,9 @@ class DownloadStrategyDetector
   def self.detect(url, strategy = nil)
     if strategy.nil?
       detect_from_url(url)
+    elsif strategy == S3DownloadStrategy
+      require_aws_sdk
+      strategy
     elsif strategy.is_a?(Class) && strategy < AbstractDownloadStrategy
       strategy
     elsif strategy.is_a?(Symbol)
@@ -1168,5 +1161,10 @@ class DownloadStrategyDetector
     else
       raise "Unknown download strategy #{symbol} was requested."
     end
+  end
+
+  def self.require_aws_sdk
+    Homebrew.install_gem! "aws-sdk-s3", "~> 1.8"
+    require "aws-sdk-s3"
   end
 end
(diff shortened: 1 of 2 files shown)


Library/Homebrew/download_strategy.rb
```diff
@@ -1148,6 +1142,9 @@ def self.detect_from_url(url)
       SubversionDownloadStrategy
     when %r{^https?://(.+?\.)?sourceforge\.net/hgweb/}
       MercurialDownloadStrategy
+    when %r{^s3://}
+      require_aws_sdk
```
MikeMcQuaid: Why is this needed in two places?

Library/Homebrew/download_strategy.rb
```diff
@@ -1169,4 +1166,11 @@ def self.detect_from_symbol(symbol)
       raise "Unknown download strategy #{symbol} was requested."
     end
   end
+
+  def self.require_aws_sdk
+    require "aws-sdk-s3"
+  rescue LoadError
```
MikeMcQuaid: Make this unconditionally run `install_gem_setup_path` instead of the `rescue`

Library/Homebrew/download_strategy.rb
```diff
@@ -1169,4 +1166,11 @@ def self.detect_from_symbol(symbol)
...
+    Homebrew.install_gem_setup_path! "aws-sdk-s3", "~> 1.8", "ruby"
```
MikeMcQuaid: Don't pass `ruby` here.

Library/Homebrew/download_strategy.rb
```diff
@@ -473,17 +473,8 @@ def stage
...
-    begin
-      require "aws-sdk-v1"
-    rescue LoadError
-      onoe "Install the aws-sdk gem into the gem repo used by brew."
-      raise
-    end
-
-    if @url !~ %r{^https?://([^.].*)\.s3\.amazonaws\.com/(.+)$}
+    unless @url =~ %r{^https?://([^.].*)\.s3\.amazonaws\.com/(.+)$} ||
+           @url =~ %r{^s3://([^.].*?)/(.+)$}
```
MikeMcQuaid: Could you keep the changes like this into a separate PR? It makes it easier to review and test this one. Thanks!

Library/Homebrew/download_strategy.rb
amancevice: This is necessary to support both cases when the formula explicitly states `:using => S3DownloadStrategy` and again when it’s implied by using the `s3://` scheme in the download URL.

Ex:

```
# Traditional URL with :using
url “https://but.s3.anazonaws.com/key” :using => S3DownloadStrategy

# S3 URL
url “s3://bkt/key”
```

Library/Homebrew/download_strategy.rb
amancevice: Sure

Library/Homebrew/download_strategy.rb
amancevice: As noted above, I can separate these into different PRs, but I think it’s needed in both places

amancevice: As noted above, I can separate these into different PRs, but I think it’s needed in both places

Library/Homebrew/download_strategy.rb
amancevice: Completely leave off the argument? I tried passsing `nil` and it threw an error at me.

Library/Homebrew/download_strategy.rb
MikeMcQuaid: It would be good to have:

1. the gem changes to the existing code
2. the `s3://` extensions

be separate PRs, yeh.

Library/Homebrew/download_strategy.rb
MikeMcQuaid: Try to omit the argument entirely and if that fails post both that and the `nil` errors here and we'll talk about an API fix. Thanks!

amancevice: After uninstalling the gem from the Homebrew bundle...

With `Homebrew.install_gem_setup_path! "aws-sdk-s3", "~> 1.8"` I get:

```
==> Installing or updating 'aws-sdk-s3' gem
Fetching: aws-sdk-s3-1.8.2.gem (100%)
Successfully installed aws-sdk-s3-1.8.2
1 gem installed
Error: The 'aws-sdk-s3' gem is installed but couldn't find 'aws-sdk-s3' in the PATH:
```

With `Homebrew.install_gem_setup_path! "aws-sdk-s3", "~> 1.8", nil` I get:

```
==> Installing or updating 'aws-sdk-s3' gem
Fetching: aws-sdk-s3-1.8.2.gem (100%)
Successfully installed aws-sdk-s3-1.8.2
1 gem installed
Error: no implicit conversion of nil into String
```

amancevice: Any advice, @MikeMcQuaid?

MikeMcQuaid:
Looks good, nice work again @amancevice!

PR #3952, opened 2018-03-21
Revert "build: do not skip runtime deps of buildtime deps"

ilovezfs:
Reverts Homebrew/brew#3911

### DIFF
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -59,6 +59,8 @@ class Build
         Dependency.prune
       elsif dep.build? && dependent != formula
         Dependency.prune
+      elsif dep.build?
+        Dependency.keep_but_prune_recursive_deps
       end
     end
   end


ilovezfs:
Unfortunately, if you have a formula with
```
depends_on "cabal-install" => :build
depends_on "ghc" => :build
```

and the cabal-install formula has
```
depends_on "ghc@8.2"
```

Then GHC 8.2 gets used in your build instead of GHC 8.4.
```
bash-3.2$ echo $PATH
/private/tmp/elm-format-20180321-79762-97tae1/.cabal-sandbox/bin:/usr/local/Homebrew/Library/Homebrew/shims/super:/usr/local/opt/ghc@8.2/bin:/usr/local/opt/cabal-install/bin:/usr/local/opt/ghc/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

This could potentially be fixed by prioritizing in PATH the bin directory of deps that are more immediate over the bin directory of deps that are more indirect.

The same would need to be the case for PKG_CONFIG_PATH, etc. too.

PR #3957, opened 2018-03-22
Added Homebrew.install_gem!

amancevice:
Useful when gem does not install with executable.

The current `Homebrew.install_gem_setup_path!` method raises an error when installing a gem that doesn't come with its own executable (e.g. `aws-sdk-s3`).

I moved the code that installs the gem into `Homebrew.install_gem!`, And call this method from `Homebrew.install_gem_setup_path!` to maintain original functionality.

I had to add a make a slight change to add a guard clause so `brew style` would approve of the changes, but the method is functionally equivalent.

**NOTE** I didn't add/change tests because these changes should not affect any existing behavior.

-----

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -199,7 +199,7 @@ module Homebrew
     _system(cmd, *args)
   end
 
-  def install_gem_setup_path!(name, version = nil, executable = name)
+  def install_gem!(name, version = nil)
     # Match where our bundler gems are.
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks so much for your first contribution (and for beating me to it)! Without people like you submitting PRs we couldn't run this project. You rock, @amancevice!

eedwards-sk:
@amancevice Nice!!

Funny because I ran into this too, but I originally just modified the method itself to support a `nil` exe name, which skipped the 'exe check'. I like your approach, though, to re-use the core behavior but adding an extra check with the 'exe version'.

amancevice:
Thanks, @eedwards-sk -- I opened a sister PR to install the gem before the sandboxing of `brew install`: https://github.com/Homebrew/brew/pull/3949

Your feedback is welcomed!

PR #3971, opened 2018-03-24
irb: Use Parser to parse args

GauthamGoli:

-----
#1860

### DIFF
--- a/Library/Homebrew/dev-cmd/irb.rb
+++ b/Library/Homebrew/dev-cmd/irb.rb
@@ -5,6 +5,9 @@
 #:    If `--pry` is passed or HOMEBREW_PRY is set, pry will be
 #:    used instead of irb.
 
+require "cli_parser"
+require "utils/env"
+
 class Symbol
   def f(*args)
     Formulary.factory(to_s, *args)
@@ -21,7 +24,12 @@ module Homebrew
   module_function
 
   def irb
-    if ARGV.include? "--examples"
+    args = Homebrew::CLI::Parser.new do
+      switch "--examples"
+      switch "--pry"
+    end.parse
+
+    if args.examples?
       puts "'v8'.f # => instance of the v8 formula"
       puts ":hub.f.installed?"
       puts ":lua.f.methods - 1.methods"
@@ -29,7 +37,7 @@ module Homebrew
       return
     end
 
-    if ARGV.pry?
+    if args.pry? || Utils::EnvVars.pry?
       Homebrew.install_gem_setup_path! "pry"
       require "pry"
       Pry.config.prompt_name = "brew"
@@ -45,7 +53,7 @@ module Homebrew
 
     ohai "Interactive Homebrew Shell"
     puts "Example commands available with: brew irb --examples"
-    if ARGV.pry?
+    if args.pry? || Utils::EnvVars.pry?
       Pry.start
     else
       IRB.start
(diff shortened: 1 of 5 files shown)


Library/Homebrew/dev-cmd/irb.rb
```diff
@@ -21,15 +24,20 @@ module Homebrew
...
+    if args.examples?
       puts "'v8'.f # => instance of the v8 formula"
       puts ":hub.f.installed?"
       puts ":lua.f.methods - 1.methods"
       puts ":mpd.f.recursive_dependencies.reject(&:installed?)"
       return
     end
 
-    if ARGV.pry?
+    if args.pry? || Utils::EnvVars.pry?
```
MikeMcQuaid: Is it possible for `args.pry?` to also handle the environment variable somehow? Also, you can probably delete https://github.com/Homebrew/brew/blob/121780b41701d1d4857b65c4a3f0ecd511807f75/Library/Homebrew/extend/ARGV.rb#L271-L273 now (and in similar other PRs).

Library/Homebrew/dev-cmd/irb.rb
```diff
@@ -21,15 +24,20 @@ module Homebrew
   module_function
 
   def irb
-    if ARGV.include? "--examples"
+    args = Homebrew::CLI::Parser.new do
```
MikeMcQuaid: I was just thinking while reading this: what do you think about a refactoring that looks like this:
```ruby
    args = Homebrew::CLI::Parser.parse do
      switch "--examples"
      switch "--pry"
    end
```
or similar as you never seem to make use of the `Parser` object directly.

Library/Homebrew/dev-cmd/irb.rb
GauthamGoli: For `args.OPTION?` to handle environment variables, would probably have to assume that env var naming has the naming convention `HOMEBREW_<OPTION>`

MikeMcQuaid: Yeh, that would work. Additionally, we could allow them to be specified with the DSL e.g. `env: :pry` or similar to mean `HOMEBREW_PRY` can trigger it.

GauthamGoli: Explicitly specifying in DSL seems better to me.

Library/Homebrew/dev-cmd/irb.rb
GauthamGoli: Yeah, seems better & more readable this way. Will open a seperate PR for this.

MikeMcQuaid: @GauthamGoli Cool. Feel free to combine both ideas above into same PR (although different commits ideally).

GauthamGoli: @MikeMcQuaid I've added option for specifying environment variables and while adding a test for this, I remembered that I do make use of `Parser` Object directly in tests. It was written in existing way to make testing easier.

MikeMcQuaid:
Thanks again @GauthamGoli, nice work!

Library/Homebrew/dev-cmd/irb.rb
MikeMcQuaid: @GauthamGoli Should still be able to have a `.parse` class method which calls `new` and then runs `parse` on it so it's a short-hand of the version here without needing to change the existing `new` or `#parse` methods. As a general principle I'd say a public API should never be less readable/harder to use just to make testing easier.

GauthamGoli: Agree. Will do it. 👍

PR #3976, opened 2018-03-25
tests: Use Parser to parse args

GauthamGoli:

-----
#1860 

This is a backwards incompatible change for `--seed`

### DIFF
--- a/Library/Homebrew/dev-cmd/tests.rb
+++ b/Library/Homebrew/dev-cmd/tests.rb
@@ -1,4 +1,4 @@
-#:  * `tests` [`--verbose`] [`--coverage`] [`--generic`] [`--no-compat`] [`--only=`<test_script>[`:`<line_number>]] [`--seed` <seed>] [`--online`] [`--official-cmd-taps`]:
+#:  * `tests` [`--verbose`] [`--coverage`] [`--generic`] [`--no-compat`] [`--only=`<test_script>[`:`<line_number>]] [`--seed=`<seed>] [`--online`] [`--official-cmd-taps`]:
 #:    Run Homebrew's unit and integration tests. If provided,
 #:    `--only=`<test_script> runs only <test_script>_spec.rb, and `--seed`
 #:    randomizes tests with the provided value instead of a random seed.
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
> This is a backwards incompatible change for --seed

@GauthamGoli We'll need to support the `--key=value` format for some other commands. Why is it not posssible for `--seed` but is for `--only`?

GauthamGoli:
@MikeMcQuaid `--seed` was of `--key value` format, now it is of `--key=value` format

MikeMcQuaid:
@GauthamGoli Ah, gotcha, yes, misread! Thanks again!

ilovezfs:
It would be nice if, in general, `--key val` and `--key=val` both worked for Homebrew commands. I think it's a common error for people to
```
brew bump-formula-pr --url https://github.com/macvim-dev/macvim/archive/snapshot-146.tar.gz --version=8.0-146 macvim
```
when they mean
```
brew bump-formula-pr --url=https://github.com/macvim-dev/macvim/archive/snapshot-146.tar.gz --version=8.0-146 macvim
```
and the error message is about as strange as you could imagine
```
Josephs-MacBook-Pro:Formula joe$ brew bump-formula-pr --url https://github.com/macvim-dev/macvim/archive/snapshot-146.tar.gz --version=8.0-146 macvim
##################################################################################################################################################################################################### 100.0%
Error: snapshot-146.tar.gz: /Users/joe/Library/Caches/Homebrew/Formula/snapshot-146.tar.gz:1: Invalid char `\x1F' in expression
/Users/joe/Library/Caches/Homebrew/Formula/snapshot-146.tar.gz:1: invalid multibyte char (UTF-8)
Josephs-MacBook-Pro:Formula joe$ 
```

GauthamGoli:
@ilovezfs Actually, `OptionParser` supports both, so both ways would work for Homebrew and I was wrong in saying this is a backwards incompatible change

GauthamGoli:
And no more weird errors once I add Parser support in `bump-formula-pr` 👍

PR #3980, opened 2018-03-25
cli_parser: Add class method to make options declaration more readable

GauthamGoli:

-----
#1860 
https://github.com/Homebrew/brew/pull/3971#discussion_r176933717

### DIFF
--- a/Library/Homebrew/cli_parser.rb
+++ b/Library/Homebrew/cli_parser.rb
@@ -4,6 +4,10 @@ require "ostruct"
 module Homebrew
   module CLI
     class Parser
+      def self.parse(&block)
+        new(&block).parse
+      end
+
       def initialize(&block)
         @parser = OptionParser.new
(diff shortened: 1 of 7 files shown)


Library/Homebrew/cli_parser.rb
```diff
@@ -4,6 +4,12 @@
 module Homebrew
   module CLI
     class Parser
+      class << self
+        def parse(&block)
```
MikeMcQuaid: `def self.parse(&block)` would work, right?

GauthamGoli: yeah, but I read that this way is more popular in RoR codebase :grimacing:

GauthamGoli: Nevermind. Changing it to `def self.parse(&block)`

MikeMcQuaid:
Thanks again @GauthamGoli!

PR #3987, opened 2018-03-26
bump-formula-pr: Use Parser to parse args

GauthamGoli:

-----
#1860

### DIFF
--- a/Library/Homebrew/cli_parser.rb
+++ b/Library/Homebrew/cli_parser.rb
@@ -13,6 +13,8 @@ module Homebrew
         @parsed_args = OpenStruct.new
         # undefine tap to allow --tap argument
         @parsed_args.instance_eval { undef tap }
+        @depends = []
+        @conflicts = []
         instance_eval(&block)
       end
 
@@ -47,6 +49,14 @@ module Homebrew
         end
       end
 
+      def depends(primary, secondary, mandatory: false)
+        @depends << [primary, secondary, mandatory]
+      end
+
+      def conflicts(primary, secondary)
+        @conflicts << [primary, secondary]
+      end
+
       def option_to_name(name)
         name.sub(/\A--?/, "").tr("-", "_")
       end
@@ -57,6 +67,7 @@ module Homebrew
 
       def parse(cmdline_args = ARGV)
         @parser.parse(cmdline_args)
+        check_constraint_violations
         @parsed_args
       end
 
@@ -81,6 +92,57 @@ module Homebrew
         when :force   then [["-f", "--force"], :force]
         else name
         end
+      end
+
+      def option_passed?(name)
+        @parsed_args.respond_to?(name) || @parsed_args.respond_to?("#{name}?")
+      end
+
+      def check_depends
+        @depends.each do |primary, secondary, required|
+          primary_passed = option_passed?(primary)
+          secondary_passed = option_passed?(secondary)
+          raise OptionDependencyError.new(primary, secondary) if required && primary_passed &&
+                                                                 !secondary_passed
+          raise OptionDependencyError.new(primary, secondary, missing: true) if secondary_passed &&
+                                                                                !primary_passed
+        end
+      end
+
+      def check_conflicts
+        @conflicts.each do |primary, secondary|
+          primary_passed = option_passed?(primary)
+          secondary_passed = option_passed?(secondary)
+          raise OptionConflictError.new(primary, secondary) if primary_passed && secondary_passed
+        end
+      end
+
+      def check_constraint_violations
+        check_conflicts
+        check_depends
+      end
+    end
+
+    class OptionDependencyError < RuntimeError
+      def initialize(arg1, arg2, missing: false)
+        if !missing
+          message = <<~EOS
+            `#{arg1}` and `#{arg2}` should be passed together
+          EOS
+        else
+          message = <<~EOS
+            `#{arg2}` cannot be passed without `#{arg1}`
+          EOS
+        end
+        super message
+      end
+    end
+
+    class OptionConflictError < RuntimeError
+      def initialize(arg1, arg2)
+        super <<~EOS
+          `#{arg1}` and `#{arg2}` should not be passed together
+        EOS
       end
     end
   end
--- a/Library/Homebrew/cli_parser.rb
+++ b/Library/Homebrew/cli_parser.rb
@@ -1,5 +1,6 @@
 require "optparse"
 require "ostruct"
+require "set"
 
 module Homebrew
   module CLI
@@ -13,7 +14,7 @@ module Homebrew
         @parsed_args = OpenStruct.new
         # undefine tap to allow --tap argument
         @parsed_args.instance_eval { undef tap }
-        @depends = []
+        @constraints = []
         @conflicts = []
         instance_eval(&block)
       end
@@ -36,7 +37,7 @@ module Homebrew
         end
       end
 
-      def flag(name, description: nil)
+      def flag(name, description: nil, required_for: nil, depends_on: nil)
         if name.end_with? "="
           required = OptionParser::REQUIRED_ARGUMENT
           name.chomp! "="
@@ -47,18 +48,16 @@ module Homebrew
         @parser.on(name, description, required) do |option_value|
           @parsed_args[option_to_name(name)] = option_value
         end
+
+        set_constraints(name, required_for: required_for, depends_on: depends_on)
       end
 
-      def depends(primary, secondary, mandatory: false)
-        @depends << [primary, secondary, mandatory]
-      end
-
-      def conflicts(primary, secondary)
-        @conflicts << [primary, secondary]
+      def conflicts(*options)
+        @conflicts << options.map { |option| option_to_name(option) }
       end
 
       def option_to_name(name)
-        name.sub(/\A--?/, "").tr("-", "_")
+        name.sub(/\A--?/, "").tr("-", "_").delete("=")
       end
 
       def option_to_description(*names)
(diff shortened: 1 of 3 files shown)


Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
...
+    @args = Homebrew::CLI::Parser.parse do
+      switch "--devel"
+      switch "-n", "--dry-run"
+      switch "--write"
+      switch "--audit"
+      switch "--strict"
+      switch "--no-browse"
+      switch "--quiet"
+      switch "--force"
+      flag   "--url",      required: true
```
ilovezfs: This is not required

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
...
+      flag   "--sha256",   required: true
```
ilovezfs: This is not required

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
...
+      flag   "--mirror",   required: true
```
ilovezfs: This is not required

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
...
+      flag   "--tag",      required: true
```
ilovezfs: This is not required

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
...
+      flag   "--revision", required: true
```
ilovezfs: This is not required

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
...
+      flag   "--version",  required: true
```
ilovezfs: This is not required

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
...
+      flag   "--message",  required: true
```
ilovezfs: This is not required

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -116,6 +117,24 @@ def check_for_duplicate_pull_requests(formula)
   end
 
   def bump_formula_pr
```
MikeMcQuaid: Would be good to move this method to the top again.

Library/Homebrew/dev-cmd/bump-formula-pr.rb
MikeMcQuaid: We'll need a way to do stuff like this: 
```
(`--url=`<URL> `--sha256=`<sha-256>|`--tag=`<tag> `--revision=`<revision>)
```

i.e. have arguments that are grouped such that if `--url` is specified then `--sha256` is also needed and that only `--url` or `--tag` is allowed/required.

ilovezfs: Note that is true in principle, but in this command --sha256 is optional even when using --url.

ilovezfs: So it would need the ability to say `--sha256=` exists, and is optional, if `--url` is passed, but otherwise the option does not exist and is invalid.

GauthamGoli: Thanks @ilovezfs, @MikeMcQuaid : Addressed above comments and added tests for the same.

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -42,80 +42,35 @@
...
+      switch :verbose
+      switch :debug
+      flag   "--url",      required: true
+      flag   "--sha256",   required: true
+      flag   "--mirror",   required: true
+      flag   "--tag",      required: true
+      flag   "--revision", required: true
+      flag   "--version",  required: true
+      flag   "--message",  required: true
+      depends :url, :sha256
```
MikeMcQuaid: Not sure about the naming here but can't quite think of something better. Could maybe have some named parameters to make clear the differences between `url` and a `sha256` too.

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -42,80 +42,35 @@
...
+      depends :tag, :revision, mandatory: true
```
MikeMcQuaid: Not clear what `mandatory` means here; how does it differ from the `url` stuff above.

GauthamGoli: `--sha256` is valid only when`--url` is passed, but `--sha256` need not be passed when `--url` is passed. 

`--revision` is valid only when `--tag` is passed, and `mandatory: true` indicates, `--revision` has to be passed when `--tag` is passed.

Library/Homebrew/dev-cmd/bump-formula-pr.rb
reitermarkus: What does this mean exactly? That the `--sha256` is only valid if the `--url` flag is also specified?

If so, why not something like this?

```ruby
flag "--sha256", required: true, depends_on: "--url"
```

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -42,80 +42,38 @@
...
+      switch "--write"
+      switch "--audit"
+      switch "--strict"
+      switch "--no-browse"
+      switch :quiet
+      switch :force
+      switch :verbose
+      switch :debug
+
+      flag   "--url",      required: true,
```
ilovezfs: I find it confusing to say it's `required: true` when it's not required if you've passed --tag.

ilovezfs: I keep forgetting `required` refers to whether it needs a value not whether the flag is ... required. Maybe it could be `value_required: true`, or even
```
flag "--url="
```
and let it speak for itself.

GauthamGoli: `flag "--url="` seems nice.

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -42,80 +42,36 @@
...
+      switch :force
+      switch :verbose
+      switch :debug
+
+      flag   "--url=",     dependent_group: { "--sha256" => :optional },
+                           conflicting_group: ["--tag"]
+
+      flag   "--tag=",     dependent_group: { "--revision" => :mandatory }
+
+      flag   "--sha256="
```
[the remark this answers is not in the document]
reitermarkus: `conflicts_with` doesn't scale, but that's just in its nature, same as in Formulae. Then at least it's clear how to use it, or how do you argue on which of the conflicting flags you have to specify it?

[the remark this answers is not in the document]
reitermarkus: Or we could do this:

```ruby
raise UXTooComplicatedError if conflicts.count > 5
```

:trollface:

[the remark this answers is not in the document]
reitermarkus: It should either be specified on every `flag`, or separately, otherwise how to you choose which `flag` to specify it on?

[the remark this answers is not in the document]
GauthamGoli: Can be specified on any one `flag`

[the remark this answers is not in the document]
reitermarkus: My point is that it isn't consistent if it can be on any `flag`.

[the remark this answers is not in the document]
GauthamGoli: Agree. Let's decide/finalize the naming convention.
1. For dependant options.
```ruby
flag "--revision=", required_for: { "--tag" => :mandatory }
```

2. For mutually exclusive options, we'll have a seperate declaration
```ruby
conflicts "--flag1", "--flag2", "--flag3" .. "--flag n"
```



We can't use
```ruby
flag "--revision=", required_for: "--tag="
```
 for dependant options as `=` specifies that value is mandatory but not the option itself.

Suggest any appropriate improvements over this.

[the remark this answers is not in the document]
reitermarkus: > We can't use
> ```ruby
> flag "--revision=", required_for: "--tag="
> ```
> for dependant options as `=` specifies that value is mandatory but not the option itself.

That's why I suggested both `depends_on` and `required_for`:

```ruby
flag "--sha256=", depends_on: "--url="
flag "--revision=", required_for: "--tag="
```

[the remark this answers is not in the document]
GauthamGoli: Cool. Seems more readable 👍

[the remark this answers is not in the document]
GauthamGoli: Updated the PR.

Library/Homebrew/cli_parser.rb
```diff
@@ -34,7 +37,7 @@ def comma_array(name, description: nil)
         end
       end
 
-      def flag(name, description: nil)
+      def flag(name, description: nil, required_for: nil, depends_on: nil)
```
MikeMcQuaid: Can just omit the `nil`s I think and they will be defaulted.

Library/Homebrew/cli_parser.rb
```diff
@@ -45,10 +48,16 @@ def flag(name, description: nil)
...
+        set_constraints(name, required_for: required_for, depends_on: depends_on)
+      end
+
+      def conflicts(*options)
+        @conflicts << options.map { |option| option_to_name(option) }
       end
 
       def option_to_name(name)
-        name.sub(/\A--?/, "").tr("-", "_")
+        name.sub(/\A--?/, "").tr("-", "_").tr("=", "")
```
MikeMcQuaid: `.tr("=", "")` should be `.delete("=")`

Library/Homebrew/cli_parser.rb
```diff
@@ -82,6 +92,90 @@ def common_switch(name)
...
+      end
+
+      def check_constraints
+        @constraints.each do |primary, secondary, constraint_type|
+          primary_passed = option_passed?(primary)
+          secondary_passed = option_passed?(secondary)
+          if :mandatory.equal?(constraint_type) && primary_passed && !secondary_passed
+            raise OptionConstraintError.new(primary, secondary)
+          end
+          raise OptionConstraintError.new(primary, secondary, missing: true) if secondary_passed &&
```
MikeMcQuaid: Make this an `elsif`?

Library/Homebrew/cli_parser.rb
GauthamGoli: Seems like `nil` can't be omitted, then they become mandatory named arguments.

Library/Homebrew/cli_parser.rb
GauthamGoli: This leads to RuboCop style error:
> Use a guard clause instead of wrapping the code inside a conditional expression.

Library/Homebrew/cli_parser.rb
MikeMcQuaid: Ah, gotcha, ignore me then!

Library/Homebrew/cli_parser.rb
MikeMcQuaid: Instead could you put the `if secondary_passed ...` on it's own line (rather than a suffix)

GauthamGoli: This works, Updated the code.

MikeMcQuaid:
Thanks again @GauthamGoli!

ilovezfs:
`--force` no longer works.

```
Josephs-MacBook-Pro:bfpr joe$ brew bump-formula-pr --no-browse --tag=v1.10.2 --revision=81753b10df112992bf51bbc2c2f85208aad78335 kubernetes-cli --force
Error: These open pull requests may be duplicates:
allow kops to work with kubernetes-cli https://github.com/Homebrew/homebrew-core/pull/26175
Duplicate PRs should not be opened. Use --force to override this error.
```

GauthamGoli:
Issue is, `@args` in `bump-formula-pr.rb` is overwriting class instance var `@args` defined in `global.rb`, at `Homebrew` module level, 
we'd have to rename either of these to resolve this. Opening a PR soon.

PR #3993, opened 2018-03-28
Posted a solution to make document use unversioned python3.

in9lude:


docs/Homebrew-and-Python.md
```diff
@@ -14,7 +14,7 @@ Homebrew provides one formula for Python 3.x (`python`) and another for Python 2
 The executables are organized as follows so that Python 2 and Python 3 can both be installed without conflict:
 * `python3` points to Homebrew's Python 3.x (if installed)
 * `python2` points to Homebrew's Python 2.7.x (if installed)
-* `python` points to Homebrew's Python 2.7.x (if installed) otherwise the macOS system Python. Check out `brew info python` if you wish to add Homebrew's 3.x `python` to your `PATH`.
+* `python` points to Homebrew's Python 2.7.x (if installed) otherwise the macOS system Python. Add `/usr/local/opt/python/libexec/bin` into your PATH if you wish to add Homebrew's 3.x `python` to your `PATH`.
```
MikeMcQuaid: This isn't correct; it varies on your Homebrew installation location. What's wrong with the `brew info python` pointer?

in9lude: Oh...I misunderstood that the variable is fixed.In that case `brew info python` is good.

PR #4003, opened 2018-03-29
Fix CLI argument OpenStruct naming conditions

alyssais:

-----

Using OpenStruct prevents us from naming an option after any method defined on OpenStruct, for example `tap`. This showed up in this Linuxbrew PR: https://github.com/Linuxbrew/brew/pull/667#issuecomment-377372812, but would also bite us were we to convert `brew create` to use the new parser, since it also has a `tap` option.

Ideally we'd probably want to only define methods for the defined options themselves, but I think it's fine leaving that as an enhancement for now.


GauthamGoli:
@alyssais: The problem was, I missed declaring `--tap` here Homebrew/brew#3983 as it was not listed in man page on brew pull. `flag "--tap", required: true` would have prevented  Linuxbrew/brew#667 (comment)
I don't understand how its going to be an issue in `brew create` if we were to declare it as a `flag`

MikeMcQuaid:
> I don't understand how its going to be an issue in brew create if we were to declare it as a flag
> Please correct me if I'm wrong

@GauthamGoli Check out this test that fails locally without these changes.

```ruby
    it "can handle options that are named the same as Object instance methods" do
      parser.switch "--tap"
      args = parser.parse(["--tap"])
      expect(args.tap).to be true
    end
```

I'd like to give a bit of time here for @GauthamGoli to see if there's an `OpenStruct` solution that's possible.

GauthamGoli:
@MikeMcQuaid It passes locally for me without these changes.
For this test to pass, expect statement has to be `expect(args.tap?).to be true` because `--tap` is a `switch`

alyssais:
Oh, yes, I should have written the test as a flag, sorry. The names confused be a bit.

alyssais:
@GauthamGoli see that new test. I meant `flag` rather than `switch` there, apologies.

GauthamGoli:
@MikeMcQuaid @alyssais This change makes more sense. I was not able to get the above test to pass. Do you know the reason why this is happening for only `--tap` ?

reitermarkus:
> Do you know the reason why this is happening for only `--tap` ?

It's because every object in Ruby inherits the `#tap` method from `Object`, and `OpenStruct` seems to prohibit redefining it.

GauthamGoli:
Ah! everything makes sense now 👍 Thanks

sjackman:
What do you know. `tap` is like `tee` for pipelines in Ruby. https://ruby-doc.org/core-2.5.0/Object.html#method-i-tap

MikeMcQuaid:
@alyssais Any idea if there's a way to do this by just defining methods on an `OpenStruct` subclass instead? Would be nice to save the code from building this ourselves from scratch.

PR #4011, opened 2018-04-01
brew create: fix meson template

tschoonj:


MikeMcQuaid:
Thanks again @tschoonj!

PR #4012, opened 2018-04-01
cli_parser: Common switch options to be accessible across brew

GauthamGoli:

-----
#1860 

~As indicated above this a WIP PR, open to feedback.~

Context: Arguments like `--verbose` , `--debug`, `--quiet`, `--force` are to be accessible across commands, they cannot be stored as `Hombrew::CLI::Parser`'s instance variables.
Example: Command `A` might use a method defined elsewhere, and this method could check for `--debug` 

~To achieve this, I'm storing common `switch`es listed above in a class variable of `Homebrew::CLI::Parser` , and I've declared `verbose` and `debug` in every `dev-cmd` that uses `Homebrew::CLI::Parser`~

### DIFF
--- a/Library/Homebrew/cli_parser.rb
+++ b/Library/Homebrew/cli_parser.rb
@@ -18,11 +18,13 @@ module Homebrew
 
       def switch(*names, description: nil, env: nil)
         description = option_to_description(*names) if description.nil?
-        names, env = common_switch(*names) if names.first.is_a?(Symbol)
+        global_switch = names.first.is_a?(Symbol)
+        names, env = common_switch(*names) if global_switch
         @parser.on(*names, description) do
-          enable_switch(*names)
+          enable_switch(*names, global_switch)
         end
-        enable_switch(*names) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
+        enable_switch(*names, global_switch) if !env.nil? &&
+                                                !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
       end
 
       def comma_array(name, description: nil)
@@ -53,23 +55,29 @@ module Homebrew
       end
 
       def parse(cmdline_args = ARGV)
-        @parser.parse!(cmdline_args)
+        @parser.parse(cmdline_args)
         @parsed_args
       end
 
       private
 
-      def enable_switch(*names)
+      def enable_switch(*names, global_switch)
         names.each do |name|
+          if global_switch
+            Homebrew.args["#{option_to_name(name)}?"] = true
+            next
+          end
           @parsed_args["#{option_to_name(name)}?"] = true
         end
       end
 
+      # These are common/global switches accessible throughout Homebrew
       def common_switch(name)
         case name
         when :quiet   then [["-q", "--quiet"], :quiet]
         when :verbose then [["-v", "--verbose"], :verbose]
         when :debug   then [["-d", "--debug"], :debug]
+        when :force   then [["-f", "--force"], :force]
         else name
         end
       end
--- a/Library/Homebrew/dev-cmd/edit.rb
+++ b/Library/Homebrew/dev-cmd/edit.rb
@@ -11,8 +11,10 @@ module Homebrew
   module_function
 
   def edit
-    args = Homebrew::CLI::Parser.parse do
-      switch "--force"
+    Homebrew::CLI::Parser.parse do
+      switch :force
+      switch :verbose
+      switch :debug
     end
 
     unless (HOMEBREW_REPOSITORY/".git").directory?
@@ -41,7 +43,7 @@ module Homebrew
       paths = ARGV.named.map do |name|
         path = Formulary.path(name)
 
-        raise FormulaUnavailableError, name unless path.file? || args.force?
+        raise FormulaUnavailableError, name unless path.file? || Homebrew.args.force?
 
         path
       end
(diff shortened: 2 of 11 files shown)


Library/Homebrew/cli_parser.rb
```diff
@@ -4,10 +4,17 @@
 module Homebrew
   module CLI
     class Parser
+      # rubocop:disable Style/ClassVars
+      @@args = OpenStruct.new
```
GauthamGoli: @MikeMcQuaid Storing this in `Homebrew::CLI::Parser` class as class variable currently, as Parser is not going to be inherited but used as is. 
If we'd consider class variables evil, I'll move this to `Homebrew` class instance variable as discussed on slack.

Library/Homebrew/extend/ARGV.rb
```diff
@@ -156,15 +158,16 @@ def values(name)
...
 
   def force?
-    flag? "--force"
+    flag?("--force") || Homebrew::CLI::Parser.args.force?
   end
 
   def verbose?
-    flag?("--verbose") || !ENV["VERBOSE"].nil? || !ENV["HOMEBREW_VERBOSE"].nil?
+    flag?("--verbose") || !ENV["VERBOSE"].nil? || !ENV["HOMEBREW_VERBOSE"].nil? ||
+      Homebrew::CLI::Parser.args.verbose?
```
GauthamGoli: Instead of adding an `or` condition (`|| Homebrew::CLI::Parser.args.verbose?`) wherever `ARGV.verbose?` has been used throughout codebase, I'm adding it here and ideally would replace occurrences `ARGV.verbose?` with `Homebrew::CLI::Parser.args.verbose?` once we use `Parser` everywhere else.

Library/Homebrew/cli_parser.rb
```diff
@@ -4,10 +4,17 @@
...
+      # rubocop:enable Style/ClassVars
       def self.parse(&block)
         new(&block).parse
       end
 
+      def self.args
+        @@args
```
reitermarkus: You can use `@args ||= OpenStruct.new` here instead of the class variable.

Library/Homebrew/cli_parser.rb
MikeMcQuaid: > If we'd consider class variables evil

Yup 👍 

> I'll move this to Homebrew class instance variable as discussed on slack.

To clarify to other readers here: I suggested having `Homebrew.args` as a global instance kinda like `Homebrew.failed` currently is.

Library/Homebrew/cli_parser.rb
MikeMcQuaid: This is another option but I do think a global `Homebrew.args` is nice and snappy.

Library/Homebrew/cli_parser.rb
```diff
@@ -16,11 +23,13 @@ def initialize(&block)
...
         description = option_to_description(*names) if description.nil?
-        names, env = common_switch(*names) if names.first.is_a?(Symbol)
+        global_switch = names.first.is_a?(Symbol)
+        names, env = common_switch(*names) if global_switch
         @parser.on(*names, description) do
-          enable_switch(*names)
+          enable_switch(*names, global_switch)
         end
-        enable_switch(*names) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
+        enable_switch(*names, global_switch) if !env.nil? &&
```
MikeMcQuaid: I'd use an `if ... end` here.

Library/Homebrew/cli_parser.rb
```diff
@@ -57,17 +66,20 @@ def parse(cmdline_args = ARGV)
 
       private
 
-      def enable_switch(*names)
+      def enable_switch(*names, global_switch)
         names.each do |name|
+          next @@args["#{option_to_name(name)}?"] = true if global_switch
```
MikeMcQuaid: I'd use an `if .. end` here and split out the `next` from assignment.

Library/Homebrew/dev-cmd/edit.rb
```diff
@@ -41,7 +43,7 @@ def edit
       paths = ARGV.named.map do |name|
         path = Formulary.path(name)
 
-        raise FormulaUnavailableError, name unless path.file? || args.force?
+        raise FormulaUnavailableError, name unless path.file? || ARGV.force?
```
MikeMcQuaid: Why this change?

Library/Homebrew/extend/ARGV.rb
MikeMcQuaid: I think it would be good to not add any references to `Homebrew::CLI::Parser.args` or `Homebrew.args` in `ARGV` but instead work towards replacing calls to `ARGV` with calls to `Homebrew.args` instead.

Library/Homebrew/cli_parser.rb
GauthamGoli: This leads to Rubocop error  : `Use a guard clause instead of wrapping the code inside a conditional expression.`

GauthamGoli:
@MikeMcQuaid Addressed comments and updated the PR

GauthamGoli:
@MikeMcQuaid I've updated `bottle.rb` to use `Homebrew.args` for global options

MikeMcQuaid:
Thanks again @GauthamGoli, great work as usual!

PR #4022, opened 2018-04-06
Tell curl to not use .curlrc

stefansundin:
Searching the issues and PRs, I wasn't able to find prior discussion of this. Apologies if this is something that has already been decided that it can't be used for some reason.

I have known that Homebrew does not like `~/.curlrc` because `brew doctor` warns about it. But it hasn't been an issue for me until I tried creating a cask (https://github.com/caskroom/homebrew-cask/pull/45612).

In deciding between `-q` and `--disable`, I chose `-q` because it appears that `--disable` was only added in [7.49.0 (May 18 2016)](https://curl.haxx.se/changes.html). I wasn't able to determine when `-q` was initially added.

One reason I can think about against not merging this, is if a user puts special proxy settings in their curlrc. But even in that case I'm not sure, because the user might not expect Homebrew to use that. Is there a way to tell curl to use the macos proxy settings?

Let me know what you think! Thanks!

---

### DIFF
--- a/Library/Homebrew/test/utils/curl_spec.rb
+++ b/Library/Homebrew/test/utils/curl_spec.rb
@@ -1,0 +1,10 @@
+require "utils/curl"
+
+describe "curl" do
+  describe "curl_args" do
+    it "returns -q as the first argument" do
+      # -q must be the first argument according to "man curl"
+      expect(curl_args("foo")[1]).to eq("-q")
+    end
+  end
+end
--- a/Library/Homebrew/test/utils/curl_spec.rb
+++ b/Library/Homebrew/test/utils/curl_spec.rb
@@ -2,9 +2,14 @@ require "utils/curl"
 
 describe "curl" do
   describe "curl_args" do
-    it "returns -q as the first argument" do
+    it "returns -q as the first argument when HOMEBREW_CURLRC is set" do
+      ENV["HOMEBREW_CURLRC"] = "1"
       # -q must be the first argument according to "man curl"
       expect(curl_args("foo")[1]).to eq("-q")
+    end
+
+    it "doesn't return -q as the first argument when HOMEBREW_CURLRC is not set" do
+      expect(curl_args("foo")[1]).to_not eq("-q")
     end
   end
 end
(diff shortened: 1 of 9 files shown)


Library/Homebrew/test/utils/curl_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "utils/curl"
+
+describe "curl" do
+  describe "curl_args" do
+    it "returns -q as the first argument" do
+      expect(curl_args("foo")[1]).to eq("-q")
```
reitermarkus: `expect(curl_args("foo")).to include("-q")`

reitermarkus: Ah, ok, fine as is then.

MikeMcQuaid:
Thanks for the PR!

> One reason I can think about against not merging this, is if a user puts special proxy settings in their curlrc.

Perhaps we could have a `HOMEBREW_CURLRC` variable or similar that tells Homebrew (not by default) to use the system `.curlrc`? We could document that under both the Environment Variables and Proxy settings of the manpage.

CC @ilovezfs for thoughts on this PR and the above given we've discussed this before.

Library/Homebrew/test/utils/curl_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "utils/curl"
+
+describe "curl" do
+  describe "curl_args" do
+    it "returns -q as the first argument" do
+      # -q must be the first argument, or else it does not work
```
reitermarkus: Replace “or else it does not work” with ”according to `man curl`”.

stefansundin:
Let me know if you want me to remove `check_user_curlrc` from `diagnostics.rb`.

Library/Homebrew/utils/curl.rb
```diff
@@ -12,11 +12,14 @@ def curl_executable
...
 def curl_args(*extra_args, show_output: false, user_agent: :default)
-  args = [
-    curl_executable.to_s,
-    "-q", # do not load .curlrc (must be the first argument)
-    "--show-error",
-  ]
+  args = [curl_executable.to_s]
+
+  # do not load .curlrc unless requested (must be the first argument)
+  if ENV["HOMEBREW_CURLRC"]
```
stefansundin: Isn't this backwards? Now we will still load curlrc by default.

MikeMcQuaid: @stefansundin It is indeed, I'm a moron! Fixed in https://github.com/Homebrew/brew/pull/4045, thanks for noticing!

PR #4025, opened 2018-04-06
Add support for filtering custom variables

jeroen:

-----

The filtering of environment variables is great. But sometimes you want to pass a custom variable to the build, such as `PKG_CONFIG_PATH` or `XDG_DATA_DIRS`. It would be great if this can be done without fully opting out of filtering.

Here is one simple proposal. This would allow a user to do e.g.:

```sh
export HOMEBREW_BUILD_ENV="PKG_CONFIG_PATH=/opt/X11/lib/pkgconfig"
brew install whatever --build-from-source
```

Which is much nicer than using `HOMEBREW_NO_ENV_FILTERING=1`. Thank you for your time!

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -64,7 +64,7 @@ if [[ -z "$HOMEBREW_NO_ENV_FILTERING" && "$1" != "test-bot" ]]
 then
   PATH="/usr/bin:/bin:/usr/sbin:/sbin"
 
-  FILTERED_ENV=()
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'm afraid this defeats the purpose of this feature, sorry. It exists specifically to stop users random variables (which may be set intentionally) leaking into and breaking the build.

PR #4027, opened 2018-04-07
Simplify `search_taps` method and handle errors

stvhwrd:

-----

This commit simplifies the code for the `search_taps` method called by a user with `brew search`.

The logic currently searches more broadly, and then processes the JSON response locally to filter some results out.  This proposed change makes the search more specific to avoid having to filter out the response.

Simpler code is easier to maintain (and more newbie friendly), and potentially easier to troubleshoot bugs like #3834 (which is what led me to this).  It seems reasonable to expect that with some [considerable?] effort, the search functionality in each of `cask` and `brew` could be unified.

Additionally, I've borrowed the begin/rescue error handling logic from the cask-specific `search_remote` method.  This allows `brew search` to continue its check for "blacklisted, migrated and deleted formulae" despite a failure in the original `search_code` method.  Despite the fact that this encourages a second query to an API that's just errored out, it seems like a better practice than just failing on the error.

This is my first PR on `brew`...    <sup><sup>please be gentle</sup></sup>

Thanks for all the work you guys do in maintaining Homebrew!  Cheers

### DIFF
--- a/Library/Homebrew/cmd/search.rb
+++ b/Library/Homebrew/cmd/search.rb
@@ -110,10 +110,10 @@ module Homebrew
       $stderr.puts Formatter.headline("Searching taps on GitHub...", color: :blue)
     end
 
-    valid_dirnames = ["Formula", "HomebrewFormula", "Casks", "."].freeze
     matches = begin
       GitHub.search_code(
         user: ["Homebrew", "caskroom"],
+        path: ["Formula", "HomebrewFormula", "Casks", "."],
         filename: query,
         extension: "rb",
       )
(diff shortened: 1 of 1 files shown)


stvhwrd:
Thanks @MikeMcQuaid - omitting the test info was definitely an oversight.

I wrote a [test script](https://gist.github.com/stvhwrd/c8f90171d61fc4b523c6b435354960b6) that just ensures that the output of `brew search` is identical after the changes in this PR.  The script searches for 10 valid packages and 5 invalid packages, then compares the output of my fork with `Homebrew/brew:master`.

Here's the output:

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @stvhwrd!

PR #4031, opened 2018-04-07
enhance write_jar_script to launch specific java version

ryenus:

-----
On macOS, the default `java` launcher available from `$PATH` is: 

	/System/Library/Frameworks/JavaVM.framework/Versions/A/Commands/java

which is a specical launcher that allows one to launch certain version
of `java` simplify by specifying the `JAVA_HOME` environment variable.
For example:

	JAVA_HOME="$(/usr/libexec/java_home -v 9)" java -version

It would report java version 9, as expected

However, when `PATH` is modified and pointing to another java, like one
from the standard JDK/JRE, that version of `java` would always be used
regardless the value of `$JAVA_HOME`:

	export PATH="$(/usr/libexec/java_home -v 1.8)/bin:$PATH"
	JAVA_HOME="$(/usr/libexec/java_home -v 9)" java -version

This time the result is java version 1.8, instead of java 9.

To avoid this issue, we can directly use `/usr/libexec/java_home`, which
can be used to launch the specified java version correctly:

	/usr/libexec/java_home -v 9 --exec java -version

and this works regardless whether `$PATH` is messed up or not.

Note: this is the next step for https://github.com/Homebrew/brew/pull/3782, which partly solved the original problem.

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -438,12 +438,10 @@ class Pathname
   # Writes an exec script that invokes a java jar
   def write_jar_script(target_jar, script_name, java_opts = "", java_version: nil)
     mkpath
-    java_home = if java_version
-      "JAVA_HOME=\"$(#{Language::Java.java_home_cmd(java_version)})\" "
-    end
+    launcher = java_version.nil? ? "java" : "/usr/libexec/java_home -v #{java_version} --exec java"
     join(script_name).write <<~EOS
       #!/bin/bash
-      #{java_home}exec java #{java_opts} -jar #{target_jar} "$@"
+      exec #{launcher} #{java_opts} -jar #{target_jar} "$@"
     EOS
   end
 


ryenus:
So, to be consistent or to be correct, is that a matter now? 

I think this is more about that `bin.env_script_all_files` applies to cases where one needs to pass necessary environment variables for the process to be launched, meanwhile to launch a java program, simply passing the `JAVA_HOME` environment variable doesn't always work as expected, so why do we still insist to use the former?

MikeMcQuaid:
> Therefore in this case the inconsistence is more of a good match for a specific case, rather than being an anomaly.

I also would rather we were consistent here even if `java_home` is a better fit.

ryenus:
@MikeMcQuaid I understand that when the two are equal, consistency is always a good thing. 

But the problem here is that the two options are not equal, in that `bin.env_script_all_files(libexec/"bin", Language::Java.java_home_env("1.8"))` sometimes doesn't work due to the issue explained above.

MikeMcQuaid:
> sometimes doesn't work due to the issue explained above
 
We assume/rely on it working. As a result, the limitations there are also acceptable here and we value the consistency.

PR #4036, opened 2018-04-08
Diagnostiks

maxim-belkin:

-----

CC @sjackman 

Rearrange code related to `diagnostic.rb`.

Methods moved:
1. `check_for_gettext`, `check_for_iconv`, `check_dyld_vars`, `check_for_multiple_volumes`, `check_for_non_prefixed_findutils`
 from: `Library/Homebrew/diagnostic.rb`
    to: `Library/Homebrew/extend/os/mac/diagnostic.rb`
 
2. `#check_dyld_vars`
from: `Library/Homebrew/test/diagnostic_spec.rb`
to : `Library/Homebrew/test/os/mac/diagnostic_spec.rb`

3. New file `Library/Homebrew/extend/os/linux/diagnostic.rb`:
that defines Linux-specific methods: `check_tmpdir_sticky_bit`, `check_ld_vars`, `check_xdg_data_dirs`, `check_tmpdir_executable`.


Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -0,0 +1,65 @@
...
+      alias generic_check_tmpdir_sticky_bit check_tmpdir_sticky_bit
+
+      def check_tmpdir_sticky_bit
+        message = generic_check_tmpdir_sticky_bit
+        return if message.nil?
+        message + <<~EOS
+          If you don't have administrative privileges on this machine,
+          create a directory and set the HOMEBREW_TEMP environment variable,
+          for example:
+            export HOMEBREW_TEMP=~/tmp
```
MikeMcQuaid: No need to `export` here.

Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -0,0 +1,65 @@
...
+            mkdir -p "$HOMEBREW_TEMP"
+            chmod 1755 "$HOMEBREW_TEMP"
+            echo "export HOMEBREW_TEMP=\"$HOMEBREW_TEMP\"" >> #{shell_profile}
```
MikeMcQuaid: Are no `Utils::Shell` methods useful here?

Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -0,0 +1,65 @@
...
+        EOS
+      end
+
+      def check_ld_vars
```
MikeMcQuaid: I reckon this could exist in `diagnostic.rb` and check both `DYLD` and `LD` variables there. Thoughts?

Library/Homebrew/extend/os/linux/diagnostic.rb
sjackman: I'm afraid that I don't follow, Mike. If the `HOMEBREW_TEMP` shell variable were not exported to the environment, it would not be available to `brew`.

sjackman: Ah, I understand what you're saying. `export` is not needed to modify the shell profile. It is however needed for `brew` to work in this shell, which is the purpose of the `export`.

Library/Homebrew/extend/os/linux/diagnostic.rb
sjackman: Yes, both `export_value` looks useful.
A new function `export_value_in_profile` similar to `prepend_path_in_profile` would be quite useful.

Library/Homebrew/extend/os/linux/diagnostic.rb
sjackman: I suggested this structure. I'm just fine with a single `check_ld_vars` in generic `diagnostic.rb` that checks both if you're happy with that.

Library/Homebrew/extend/os/linux/diagnostic.rb
maxim-belkin: Yes, that's the reason for having `export` in front of the first assignment (idea is that people can simply copy-paste these instructions). So, are we good here or should I remove this `export`?

Library/Homebrew/extend/os/linux/diagnostic.rb
maxim-belkin: Will do. Note, however, that `prepend_path_in_profile` should, in fact, be `append_path_in_profile`.

Library/Homebrew/extend/os/linux/diagnostic.rb
maxim-belkin: Is the suggestion to check these variables unconditionally on both platforms?

Library/Homebrew/extend/os/linux/diagnostic.rb
maxim-belkin: I take it back. The path it prepended :) No one cares that this is done at the end of a `profile` file.

Library/Homebrew/extend/os/linux/diagnostic.rb
MikeMcQuaid: @maxim-belkin Yep 👍

Library/Homebrew/extend/os/linux/diagnostic.rb
MikeMcQuaid: @maxim-belkin We don't normally tell people to `export` as well as the shell changes; the assumption are people will open a new shell after the change.

MikeMcQuaid: I'm tempted to say remove the use of the `HOMEBREW_TEMP` bash variable here (while still setting it in the profile) altogether.

sjackman: > We don't normally tell people to export as well as the shell changes; the assumption are people will open a new shell after the change.

That instruction would need to be given explicitly, if so.

MikeMcQuaid: > That instruction to close the current shell and open a new one ought to be given explicitly, in that case. I'd prefer to export.

It's not elsewhere and I think consistency is more important.

sjackman: I defer.

MikeMcQuaid: @sjackman 🙇

maxim-belkin:
implemented suggested changes and rebased on top of master to sync it up.

sjackman:
`install` seems to be present on each of the systems (Docker images) that I quickly tested.

Library/Homebrew/extend/os/linux/diagnostic.rb
sjackman: `install` seems to be present on each of the systems (Docker images) that I quickly tested.

maxim-belkin:
If my changes to `utils/shell.rb` seem reasonable, I'd like to add more tests for the new methods.

Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -275,6 +290,92 @@ def check_homebrew_prefix
...
       end
+
+      def check_for_gettext
+        find_relative_paths("lib/libgettextlib.dylib",
+                            "lib/libintl.dylib",
+                            "include/libintl.h")
+        return if @found.empty?
+
+        # Our gettext formula will be caught by check_linked_keg_only_brews
+        gettext = begin
```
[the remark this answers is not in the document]
maxim-belkin: oops, will fix this. Forgot to switch on `paste` mode in `vim`

Library/Homebrew/utils/shell.rb
```diff
@@ -40,14 +40,36 @@ def profile
...
       when :csh, :tcsh
-        "echo 'setenv PATH #{csh_quote(path)}:$PATH' >> #{profile}"
+        "echo 'setenv #{variable} #{csh_quote(value)}' >> #{profile}"
+      when :fish
+        "echo 'set -g #{variable} \"#{sh_quote(value)}\"' >> #{profile}"
+      end
+    end
+
+    def prepend_variable_in_profile(variable, value)
+      return set_variable_in_profile(variable, value) if ENV[variable].nil?
```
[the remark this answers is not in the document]
maxim-belkin: This part handles the case when a variable is not initially set. When this happens
```
export VAR=/some/value:$VAR
```
results in
```
export VAR=/some/value:
```
This extra colon at the end can lead to undesired behaviours.

Library/Homebrew/extend/os/mac/diagnostic.rb
```diff
@@ -275,6 +290,92 @@ def check_homebrew_prefix
...
+          are compiled with improper architectures. Consider removing these files:
+        EOS
+      end
+
+      def check_for_iconv
+        find_relative_paths("lib/libiconv.dylib", "include/iconv.h")
+        return if @found.empty?
+
+        libiconv = begin
+                    Formulary.factory("libiconv")
```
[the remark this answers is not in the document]
maxim-belkin: done

Library/Homebrew/utils/shell.rb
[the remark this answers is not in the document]
sjackman: Here's an option using `${foo+bar}`
```sh
export PATH=/new/path${PATH+:$PATH}
```
For example
```sh
❯❯❯ env FOO=/old/path sh -c 'echo FOO=/new/path${FOO+:$FOO}'
FOO=/new/path:/old/path
❯❯❯ env -u FOO sh -c 'echo FOO=/new/path${FOO+:$FOO}'
FOO=/new/path
```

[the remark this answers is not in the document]
maxim-belkin: @sjackman This is a (useful) Bash-ism and I don't know how to do the same in Csh or Fish

Library/Homebrew/extend/os/mac/diagnostic.rb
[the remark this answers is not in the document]
maxim-belkin: this has been fixed...

Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -0,0 +1,49 @@
...
+          Consider adding Homebrew's share directory to XDG_DATA_DIRS like so
+            #{prepend_variable_in_profile("XDG_DATA_DIRS", HOMEBREW_PREFIX/"share")}
+        EOS
+      end
+
+      def check_tmpdir_executable
+        f = Tempfile.new(%w[homebrew_check_tmpdir_executable .sh], HOMEBREW_TEMP)
+        f.write "#!/bin/sh\n"
+        f.chmod 0700
+        f.close
```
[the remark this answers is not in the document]
maxim-belkin: Removed though I had to add `flush` to make sure file is updated accordingly.

Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -0,0 +1,49 @@
...
+          Homebrew's share was not found in your XDG_DATA_DIRS but you have
+          this variable set to include other locations.
+          Some programs like `vapigen` may not work correctly.
+          Consider adding Homebrew's share directory to XDG_DATA_DIRS like so
+            #{prepend_variable_in_profile("XDG_DATA_DIRS", HOMEBREW_PREFIX/"share")}
+        EOS
+      end
+
+      def check_tmpdir_executable
+        f = Tempfile.new(%w[homebrew_check_tmpdir_executable .sh], HOMEBREW_TEMP)
```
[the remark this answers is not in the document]
maxim-belkin: I think using `Tempfile` is beneficial here to make sure we're not overwriting anything.

Library/Homebrew/diagnostic.rb
```diff
@@ -571,17 +517,17 @@ def check_for_config_scripts
...
-        return if dyld_vars.empty?
+      def check_ld_vars
+        ld_vars = ENV.keys.grep(/^(LD|DYLD)_/)
+        return if ld_vars.empty?
 
-        values = dyld_vars.map { |var| "#{var}: #{ENV.fetch(var)}" }
+        values = ld_vars.map { |var| "#{var}: #{ENV.fetch(var)}" }
         message = inject_file_list values, <<~EOS
-          Setting DYLD_* vars can break dynamic linking.
+          Setting LD_* and DYLD_* variables can break dynamic linking.
```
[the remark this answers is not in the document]
maxim-belkin: I wonder if this method is, actually, necessary. I added `ENV.keys.sort.each { |x| ohai x }` at the beginning of the method for debugging purposes and all it got me was just a few standard env. variables...

maxim-belkin:
most comments have been addressed. Ready for another round of reviews.

maxim-belkin:
```
Warning: The directory /tmp does not permit executing
programs. It is likely mounted as "noexec". Please set HOMEBREW_TEMP
in your ~/.bash_profile to a different directory, for example:
  echo 'export HOMEBREW_TEMP="\~/tmp"' >> ~/.bash_profile
```

note `HOMEBREW_TEMP="\~/tmp"`

Library/Homebrew/utils/shell.rb
```diff
@@ -40,14 +40,36 @@ def profile
...
+    def set_variable_in_profile(variable, value)
       case preferred
       when :bash, :ksh, :sh, :zsh, nil
-        "echo 'export PATH=\"#{sh_quote(path)}:$PATH\"' >> #{profile}"
+        "echo 'export #{variable}=\"#{sh_quote(value)}\"' >> #{profile}"
       when :csh, :tcsh
-        "echo 'setenv PATH #{csh_quote(path)}:$PATH' >> #{profile}"
+        "echo 'setenv #{variable} #{csh_quote(value)}' >> #{profile}"
+      when :fish
+        "echo 'set -g #{variable} \"#{sh_quote(value)}\"' >> #{profile}"
```
reitermarkus: `-gx`

Library/Homebrew/utils/shell.rb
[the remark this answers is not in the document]
reitermarkus: In `fish` you do 

```fish
set VAR value1 value2 ... $VAR
```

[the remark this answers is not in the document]
maxim-belkin: I think this will work: https://stackoverflow.com/a/21564464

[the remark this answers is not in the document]
maxim-belkin: in csh we can't write that in a single line. we have to write:

```
if ( $?variable == 0 )
  setenv variable "$value"
else
  setenv variable "$value":"$variable"
endif
```
I guess we can use `<<~EOS` for that...

Library/Homebrew/test/utils/shell_spec.rb
```diff
@@ -69,25 +69,49 @@
...
+      expect(subject.prepend_path_in_profile(path))
+        .to eq("setenv PATH ~/some/path/../with\\ spaces:\"$PATH\" >> ~/.cshrc")
     end
 
     it "supports Fish" do
       ENV["SHELL"] = "/usr/local/bin/fish"
+      ENV["fish_user_paths"] = "/some/path"
       expect(subject.prepend_path_in_profile(path))
-        .to start_with("echo 'set -g fish_user_paths \"#{path}\" $fish_user_paths' >>")
+        .to eq("echo 'set -gx PATH ~/some/path/../with\\ spaces \"$PATH\"' >> ~/.config/fish/config.fish")
```
reitermarkus: This should still be `fish_user_paths`, right?

Library/Homebrew/diagnostic.rb
```diff
@@ -571,17 +517,18 @@ def check_for_config_scripts
...
+      def check_ld_vars
+        ld_vars = ENV.keys.grep(/^(|DY)LD_/)
+        return if ld_vars.empty?
 
-        values = dyld_vars.map { |var| "#{var}: #{ENV.fetch(var)}" }
+        values = ld_vars.map { |var| "#{var}: #{ENV.fetch(var)}" }
+        vars = ld_vars.map { |x| x.sub(/_.*/, "") }.uniq.sort.map { |x| x + "_*" }.join(" and ")
         message = inject_file_list values, <<~EOS
-          Setting DYLD_* vars can break dynamic linking.
+          Setting #{vars} variables can break dynamic linking.
```
MikeMcQuaid: Let's just do `DYLD_* or LD_*` variables and kill the `vars` line above.

Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def check_xdg_data_dirs
+        return if ENV["XDG_DATA_DIRS"].to_s.empty?
+        return if ENV["XDG_DATA_DIRS"].split("/").include?(HOMEBREW_PREFIX/"share")
+        <<~EOS
+          Homebrew's share was not found in your XDG_DATA_DIRS but you have
+          this variable set to include other locations.
+          Some programs like `vapigen` may not work correctly.
+          Consider adding Homebrew's share directory to XDG_DATA_DIRS like so
```
MikeMcQuaid: Add a trailing `:`

Library/Homebrew/utils/shell.rb
```diff
@@ -40,14 +40,31 @@ def profile
...
+        EOS
+      when :fish
+        "echo 'set -gx #{variable} #{sh_quote(value)} \"$#{variable}\"' >> #{profile}"
+      end
+    end
+
     def prepend_path_in_profile(path)
       case preferred
       when :bash, :ksh, :sh, :zsh, nil
-        "echo 'export PATH=\"#{sh_quote(path)}:$PATH\"' >> #{profile}"
```
MikeMcQuaid: Can you explain all the changes in this function and why they are needed?

Library/Homebrew/utils/shell.rb
```diff
@@ -40,14 +40,31 @@ def profile
       SHELL_PROFILE_MAP.fetch(preferred, "~/.bash_profile")
     end
 
+    def prepend_set_variable_in_profile(variable, value)
```
MikeMcQuaid: It seems like `prepend_path_in_profile(path)` should just be `prepend_set_variable_in_profile(PATH, path)`?

Library/Homebrew/diagnostic.rb
maxim-belkin: ok

Library/Homebrew/extend/os/linux/diagnostic.rb
maxim-belkin: ok

Library/Homebrew/utils/shell.rb
maxim-belkin: it can be (and this PR did that before). The difference between `prepend_set_variable` and `prepend_path` is that when `PATH` is unset users most likely won't be able to run `brew` at all. But this is an assumption, of course, and it is more reliable to use `prepend_set`

MikeMcQuaid: > But this is an assumption, of course, and it is more reliable to use prepend_set

I think we can assume that all Homebrew's users have a `PATH`.

Library/Homebrew/utils/shell.rb
MikeMcQuaid: I'm a bit dubious that this is all completely broken but yet we've not handle a single related used quote. Can you extract all the changes to this function (and the equivalents for `prepend_set_variable_in_profile`) into a separate PR where it can be discussed there. For example:

> export PATH="~/test test:$PATH"

This seems vastly preferable to:

> export PATH=~/test\ test:"$PATH"

maxim-belkin: OK, I will move it to a different PR.

MikeMcQuaid: Thanks!

maxim-belkin: It may be preferable, but not what Homebrew does. Homebrew says:

> export PATH="~/test\ test:$PATH"

maxim-belkin: > `~` must be outside of the quotes.

exactly. When it's outside of double quotes, it is correctly translated to user's home directory

maxim-belkin:
@MikeMcQuaid apologies for the delay... I removed changes to functions in `Library/Homebrew/utils/shell.rb` as you requested to another branch and will file a separate PR after this one is merged.

maxim-belkin:
Not sure how to handle this at the moment (will have time to think about it after Wednesday).
In the meantime, ping @sjackman 😄

MikeMcQuaid:
@maxim-belkin No rush. Just FYI: we can't have `brew doctor` failing on Travis CI by default (and we don't want to special case Travis CI either; it shows there's something up with the warning).

MikeMcQuaid:
You could consider removing the Linux `check_tmpdir_executable` for now and we could merge as-is.

maxim-belkin:
> You could consider removing the Linux check_tmpdir_executable for now

Good idea. Will actually do that.

PR #4044, opened 2018-04-09
urls_cop: Search for func calls to match `url` and `mirror` calls

GauthamGoli:

-----
Fixes bug found in https://github.com/Homebrew/homebrew-core/pull/26325

`curl.rb` has `system "#{bin}/curl", "-L", stable.url, "-o", filename` and `find_every_method_call_by_name` in `urls_cop.rb` was matching `stable.url` with empty params leading to NPE.

We also need a way to print RuboCop tracebacks when running `brew audit --debug` (Currently it doesn't) which i'll open a PR soon

cc @MikeMcQuaid

### DIFF
--- a/Library/Homebrew/rubocops/extend/formula_cop.rb
+++ b/Library/Homebrew/rubocops/extend/formula_cop.rb
@@ -99,6 +99,17 @@ module RuboCop
         end
       end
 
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Great and quick work here @GauthamGoli, thanks so much ❤️

PR #4046, opened 2018-04-10
Keg#replace_text_in_files: fix incorrect replace order

xu-cheng:
The order of the regexes in `Keg#replace_text_in_files` matters.
Otherwise, `/usr/local/Cellar` will be replaced to
`@@HOMEBREW_PREFIX@@/Cellar` instead of `@@HOMEBREW_CELLAR@@`.

Fixes Homebrew/homebrew-core#26043

-----

### DIFF
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -68,7 +68,7 @@ class Keg
         relocation.old_cellar => relocation.new_cellar,
         relocation.old_repository => relocation.new_repository,
       }
-      changed = s.gsub!(Regexp.union(replacements.keys), replacements)
+      changed = s.gsub!(Regexp.union(replacements.keys.sort_by(&:length).reverse), replacements)
       next unless changed
       changed_files += [first, *rest].map { |file| file.relative_path_from(path) }
 


sjackman:
I don't believe this PR does what you think it does. A regexp matches the longest possible pattern. The order of the options does not matter. `/a|b/` is equivalent to `/b|a/` for any string `a` and `b`. The order of the elements of `foo` in `Regexp.union(foo)` has no effect, since the longest possible match of the elements of `foo` is matched by the regex.
So
```ruby
      changed = s.gsub!(Regexp.union(replacements.keys.sort_by(&:length).reverse), replacements)
```
is equivalent to
```ruby
      changed = s.gsub!(Regexp.union(replacements.keys), replacements)
```
Sorting by length has no effect.

sjackman:
I'm wrong. It's awfully inconvenient when theory is disproven by the truth.
```
irb(main):001:0> /a|ab/.match("abc")
=> #<MatchData "a">
irb(main):002:0> /ab|a/.match("abc")
=> #<MatchData "ab">
```
So Ruby Regexes are not greedy? That's super weird. Time to read the docs…

sjackman:
Perl behaves the same as Ruby. Here's a related StackOverflow post: https://stackoverflow.com/questions/2394931/regex-is-behaving-lazy-should-be-greedy

The alternation (`|`) operator of command line `egrep` however is greedy.

```
❯❯❯ egrep -o 'a|ab' <<<abc
ab
❯❯❯ egrep -o 'ab|a' <<<abc
ab
```

Well. I learned something today.

sjackman:
Also, I believe that I introduced this bug. So, my apologies, and thanks for for fixing it, Cheng.

PR #4064, opened 2018-04-12
Support enabling auto-updates for non-GitHub taps

elyscape:

-----

By default, taps with repositories from places other than GitHub are skipped during the auto-update step that runs before some commands, the reasoning being that these might be slow. This adds the ability to mark such a tap as not to be skipped.

### DIFF
--- a/Library/Homebrew/cmd/tap.rb
+++ b/Library/Homebrew/cmd/tap.rb
@@ -1,7 +1,7 @@
 #:  * `tap`:
 #:    List all installed taps.
 #:
-#:  * `tap` [`--full`] <user>`/`<repo> [<URL>]:
+#:  * `tap` [`--full`] [`--force-auto-update`] <user>`/`<repo> [<URL>]:
 #:    Tap a formula repository.
 #:
 #:    With <URL> unspecified, taps a formula repository from GitHub using HTTPS.
@@ -17,6 +17,10 @@
 #:    By default, the repository is cloned as a shallow copy (`--depth=1`), but
 #:    if `--full` is passed, a full clone will be used. To convert a shallow copy
 #:    to a full copy, you can retap passing `--full` without first untapping.
+#:
+#:    By default, only taps hosted on GitHub are auto-updated (for performance
+#:    reasons). If `--force-auto-update` is passed, this tap will be auto-updated
+#:    even if it is not hosted on GitHub.
 #:
 #:    `tap` is re-runnable and exits successfully if there's nothing to do.
 #:    However, retapping with a different <URL> will cause an exception, so first
@@ -44,6 +48,7 @@ module Homebrew
       tap = Tap.fetch(ARGV.named[0])
       begin
         tap.install clone_target: ARGV.named[1],
+                    force_auto_update: force_auto_update?,
                     full_clone: full_clone?,
                     quiet: ARGV.quieter?
       rescue TapRemoteMismatchError => e
@@ -55,5 +60,10 @@ module Homebrew
 
   def full_clone?
     ARGV.include?("--full") || ARGV.homebrew_developer?
+  end
+
+  def force_auto_update?
+    # if no relevant flag is present, return nil, meaning "no change"
+    true if ARGV.include?("--force-auto-update")
   end
 end
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -493,8 +493,12 @@ EOS
         [[ -z "$HOMEBREW_UPDATE_FORCE" ]] && [[ "$UPSTREAM_SHA_HTTP_CODE" = "304" ]] && exit
       elif [[ -n "$HOMEBREW_UPDATE_PREINSTALL" ]]
       then
-        # Don't try to do a `git fetch` that may take longer than expected.
-        exit
+        FORCE_AUTO_UPDATE="$(git config homebrew.forceautoupdate 2>/dev/null || echo "false")"
+        if [[ "$FORCE_AUTO_UPDATE" != "true" ]]
+        then
+          # Don't try to do a `git fetch` that may take longer than expected.
+          exit
+        fi
       fi
 
       if [[ -n "$HOMEBREW_VERBOSE" ]]
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -201,7 +201,9 @@ class Tap
   # install this {Tap}.
   #
   # @param [Hash] options
-  # @option options [String]  :clone_targe If passed, it will be used as the clone remote.
+  # @option options [String] :clone_target If passed, it will be used as the clone remote.
+  # @option options [Boolean, nil] :force_auto_update If present, whether to override the
+  #   logic that skips non-GitHub repositories during auto-updates.
   # @option options [Boolean] :full_clone If set as true, full clone will be used.
   # @option options [Boolean] :quiet If set, suppress all output.
   def install(options = {})
@@ -210,12 +212,14 @@ class Tap
     full_clone = options.fetch(:full_clone, false)
     quiet = options.fetch(:quiet, false)
     requested_remote = options[:clone_target] || default_remote
+    # if :force_auto_update is unset, use nil, meaning "no change"
+    force_auto_update = options.fetch(:force_auto_update, nil)
 
     if official? && DEPRECATED_OFFICIAL_TAPS.include?(repo)
       odie "#{name} was deprecated. This tap is now empty as all its formulae were migrated."
     end
 
-    if installed?
+    if installed? && force_auto_update.nil?
       raise TapAlreadyTappedError, name unless full_clone
       raise TapAlreadyUnshallowError, name unless shallow?
     end
@@ -224,6 +228,11 @@ class Tap
     Utils.ensure_git_installed!
 
     if installed?
+      unless force_auto_update.nil?
+        config["forceautoupdate"] = force_auto_update
+        return if !full_clone || !shallow?
+      end
+
       if options[:clone_target] && requested_remote != remote
         raise TapRemoteMismatchError.new(name, @remote, requested_remote)
       end
@@ -258,6 +267,8 @@ class Tap
       end
       raise
     end
+
+    config["forceautoupdate"] = force_auto_update unless force_auto_update.nil?
 
     link_completions_and_manpages
 
--- a/Library/Homebrew/test/tap_spec.rb
+++ b/Library/Homebrew/test/tap_spec.rb
@@ -209,6 +209,31 @@ describe Tap do
       expect { already_tapped_tap.install full_clone: true }.to raise_error(TapAlreadyUnshallowError)
     end
 
+    describe "force_auto_update" do
+      before do
+        setup_git_repo
+      end
+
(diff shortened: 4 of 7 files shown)


MikeMcQuaid:
@elyscape Nice idea. I reckon you should kill the `--no-force-auto-update`; I think it's both unnecessary and a bit confusing. What do you think?

elyscape:
I added that so that it would be possible to unmark a tap. I figured that was probably a good idea in case somebody marked a tap for forced autoupdates but then it made things run slowly and they wanted to undo it. Thoughts?

MikeMcQuaid:
> I figured that was probably a good idea in case somebody marked a tap for forced autoupdates but then it made things run slowly and they wanted to undo it.

In that case they can always just untap and tap. If we get a bunch of requests for it we can consider adding it later.

elyscape:
Fair enough! I’ll keep it in a side branch for future reference

elyscape:
@MikeMcQuaid Extra flag removed.

elyscape:
Also pinging @woodruffw per discussion on IRC.

Library/Homebrew/cmd/tap.rb
```diff
@@ -56,4 +62,8 @@ def tap
   def full_clone?
     ARGV.include?("--full") || ARGV.homebrew_developer?
   end
+
+  def force_auto_update?
+    true if ARGV.include?("--force-auto-update")
```
woodruffw: This can be simplified to just:

```ruby
def force_auto_update?
  ARGV.include?("--force-auto-update")
end
```

Library/Homebrew/tap.rb
```diff
@@ -210,12 +212,13 @@ def install(options = {})
     full_clone = options.fetch(:full_clone, false)
     quiet = options.fetch(:quiet, false)
     requested_remote = options[:clone_target] || default_remote
+    force_auto_update = options.fetch(:force_auto_update, nil)
```
woodruffw: Instead of `fetch`ing with default as `nil`, maybe default to `false` like the `full_clone` variable above?

elyscape: The idea here was that, if `:force_auto_update` isn't set, we don't specifically mark the repository either way.

elyscape: To elaborate some more, I did it this way to handle the retapping. By having it default to `nil`, you don't need to specify the same value; it'll just use whatever is already in place. Additionally, it allows us to know if we need to mark/unmark the repo during a retap by checking for `nil`.

Library/Homebrew/cmd/tap.rb
elyscape: I actually did it this way on purpose so that it returns `nil` by default. More information in [this comment].

[this comment]: https://github.com/Homebrew/brew/pull/4064#discussion_r181530968

Library/Homebrew/tap.rb
woodruffw: It's possible that I'm misunderstanding, but this *should* have the same effect:

```ruby
force_auto_update = options.fetch(:force_auto_update, false)

# [... snip ...]

config["forceautoupdate"] = force_auto_update unless force_auto_update
```

In either case, it'd be good to add a comment stating the goal of a line like that (i.e., avoiding marking the repository either way).

Library/Homebrew/cmd/tap.rb
woodruffw: Does my suggestion in https://github.com/Homebrew/brew/pull/4064#discussion_r181532198 help resolve this? Having a predicate method return `nil` is a little unintuitive, especially since `:force_auto_update` gets documented below as being a Boolean.

Library/Homebrew/tap.rb
elyscape: I'm pretty sure that wouldn't work.

Scenarios:

`options[:force_auto_update]` is `true`:
- `force_auto_update` is set to `true`
- `unless force_auto_update` fails:
  - `config["forceautoupdate"] = force_auto_update` is not executed
- the repo is not marked (**undesired outcome**)

`options[:force_auto_update]` is `false`:
- `force_auto_update` is set to `false`
- `unless force_auto_update` succeeds:
  - `config["forceautoupdate"] = force_auto_update` is executed
- the repo is marked with `forceautoupdate` set as `false`

`options[:force_auto_update]` does not exist:
- `force_auto_update` is set to `false`
- `unless force_auto_update` succeeds:
  - `config["forceautoupdate"] = force_auto_update` is executed
- the repo is marked with `forceautoupdate` set as `false` (**undesired outcome**)

Library/Homebrew/cmd/tap.rb
elyscape: Unfortunately not. How would you recommend documenting it as an optional boolean?

elyscape: Pushed a minor change that adds documentation to this function and changes the description of `:force_auto_update` to read as follows:
```ruby
# @option options [Boolean] :force_auto_update If present, whether to override the logic
#  that skips non-GitHub repositories during auto-updates.
```
Thoughts?

woodruffw: That looks good, although you should turn `[Boolean]` into `[Boolean, nil]` so that YARD can mark it as nilable.

elyscape: Done.d

Library/Homebrew/tap.rb
reitermarkus: > will get blown away by some well-intentioned refactorer

There is a spec for it, but it could use a better dedcription than “defaults to nil”.

elyscape: Added a comment explaining why this defaults to `nil`.

Library/Homebrew/test/tap_spec.rb
```diff
@@ -209,6 +209,31 @@ def setup_git_repo
       expect { already_tapped_tap.install full_clone: true }.to raise_error(TapAlreadyUnshallowError)
     end
 
+    describe "force_auto_update" do
+      it "defaults to nil" do
+        setup_git_repo
+        already_tapped_tap = described_class.new("Homebrew", "foo")
```
reitermarkus: Use `let(:already_tapped_tap)` to define this, since it is used in all three specs.

Library/Homebrew/test/tap_spec.rb
```diff
@@ -209,6 +209,31 @@ def setup_git_repo
       expect { already_tapped_tap.install full_clone: true }.to raise_error(TapAlreadyUnshallowError)
     end
 
+    describe "force_auto_update" do
+      it "defaults to nil" do
+        setup_git_repo
```
reitermarkus: Use a `before` block for this, since it is used in all three specs.

elyscape:
@reitermarkus Comments addressed.

Library/Homebrew/tap.rb
```diff
@@ -201,7 +201,9 @@ def core_tap?
   # install this {Tap}.
   #
   # @param [Hash] options
-  # @option options [String]  :clone_targe If passed, it will be used as the clone remote.
```
elyscape: While I was in here, I went ahead and fixed the spelling on this option to match the code.

Library/Homebrew/test/tap_spec.rb
```diff
@@ -209,6 +209,31 @@ def setup_git_repo
...
+      let(:already_tapped_tap) { described_class.new("Homebrew", "foo") }
+
+      it "defaults to nil" do
+        expect(already_tapped_tap).to be_installed
+        expect(already_tapped_tap.config["forceautoupdate"]).to be_nil
+      end
+
+      it "enables forced auto-updates when true" do
+        expect(already_tapped_tap).to be_installed
+        expect { already_tapped_tap.install force_auto_update: true }.not_to raise_error
```
reitermarkus: `.not_to raise_error` here is pretty useless, since the spec would fail anyways in that case.

elyscape: Good point. Removed.

elyscape:
Pinging to make sure this doesn't fall through the cracks.

Library/Homebrew/cmd/tap.rb
```diff
@@ -18,6 +18,11 @@
 #:    if `--full` is passed, a full clone will be used. To convert a shallow copy
 #:    to a full copy, you can retap passing `--full` without first untapping.
 #:
+#:    By default, taps with repositories cloned from places other than GitHub
+#:    are skipped for performance reasons during the auto-update step that runs
+#:    before `brew install`, `brew upgrade` and `brew tap`. If `--force-auto-update` is
+#:    passed, this behavior will be disabled for the tap.
```
MikeMcQuaid: Let's slim this down a bit. How about:

> By default, taps with non-GitHub repositories are not auto-updated (for performance reasons). If `--force-auto-update` is passed, this tap will be auto-updated even if it is not hosted on GitHub.

Library/Homebrew/cmd/update.sh
```diff
@@ -493,8 +493,12 @@ EOS
         [[ -z "$HOMEBREW_UPDATE_FORCE" ]] && [[ "$UPSTREAM_SHA_HTTP_CODE" = "304" ]] && exit
       elif [[ -n "$HOMEBREW_UPDATE_PREINSTALL" ]]
       then
-        # Don't try to do a `git fetch` that may take longer than expected.
-        exit
+        FORCE_AUTO_UPDATE="$(git config homebrew.forceautoupdate || echo "false")"
```
MikeMcQuaid: `git config homebrew.forceautoupdate 2>/dev/null`

Library/Homebrew/tap.rb
```diff
@@ -201,7 +201,9 @@ def core_tap?
...
+  # @option options [String] :clone_target If passed, it will be used as the clone remote.
+  # @option options [Boolean, nil] :force_auto_update If present, whether to override the
+  #  logic that skips non-GitHub repositories during auto-updates.
```
MikeMcQuaid: Can you indent this line a bit more? One character seems a bit odd. Thanks!

Library/Homebrew/tap.rb
```diff
@@ -224,6 +228,11 @@ def install(options = {})
     Utils.ensure_git_installed!
 
     if installed?
+      unless force_auto_update.nil?
+        config["forceautoupdate"] = force_auto_update
+        return unless full_clone && shallow?
```
MikeMcQuaid: `return if !full_clone || !shallow?` is a bit easier to mentally parse

Library/Homebrew/cmd/tap.rb
reitermarkus: > By default, taps with non-GitHub repositories are not auto-updated (for performance reasons).

Maybe also remove the double-negation here.

> By default, only taps hosted on GitHub are auto-updated (for performance reasons).

MikeMcQuaid: @reitermarkus Agreed 👍

elyscape:
@MikeMcQuaid This time with the manpages being updated, too, so that testing won't fail. Oops.

MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @elyscape!

PR #4066, opened 2018-04-13
missing: exit nonzero if missing deps are found

scpeters:

-----

Currently `brew missing` always returns a successful error code. This changes it to return a failing error code when it detects formulae with missing dependencies.

### DIFF
--- a/Library/Homebrew/cmd/missing.rb
+++ b/Library/Homebrew/cmd/missing.rb
@@ -4,6 +4,8 @@
 #:
 #:    If `--hide=`<hidden> is passed, act as if none of <hidden> are installed.
 #:    <hidden> should be a comma-separated list of formulae.
+#:
+#:    `missing` exits with a non-zero status if any formulae are missing dependencies.
 
 require "formula"
 require "tab"
@@ -25,6 +27,7 @@ module Homebrew
       missing = f.missing_dependencies(hide: ARGV.values("hide"))
       next if missing.empty?
 
+      Homebrew.failed = true
       print "#{f}: " if ff.size > 1
       puts missing.join(" ")
     end
--- a/Library/Homebrew/test/cmd/missing_spec.rb
+++ b/Library/Homebrew/test/cmd/missing_spec.rb
@@ -14,7 +14,7 @@ describe "brew missing", :integration_test do
     expect { brew "missing" }
       .to output("foo\n").to_stdout
       .and not_to_output.to_stderr
-      .and be_a_success
+      .and be_a_failure
   end
 
   it "prints nothing if all dependencies are installed" do
@@ -35,7 +35,7 @@ describe "brew missing", :integration_test do
       expect { brew "missing", "--hide=foo" }
         .to output("bar: foo\n").to_stdout
         .and not_to_output.to_stderr
-        .and be_a_success
+        .and be_a_failure
     end
   end
 end
--- a/docs/Manpage.md
+++ b/docs/Manpage.md
@@ -340,6 +340,8 @@ With `--verbose` or `-v`, many commands print extra debugging information. Note 
     If `--hide=``hidden` is passed, act as if none of `hidden` are installed.
     `hidden` should be a comma-separated list of formulae.
 
+    `missing` exits with a non-zero status if any formulae are missing dependencies.
+
   * `options` [`--compact`] (`--all`|`--installed`|`formulae`):
     Display install options specific to `formulae`.
 
--- a/manpages/brew.1
+++ b/manpages/brew.1
@@ -346,6 +346,9 @@ Check the given \fIformulae\fR for missing dependencies\. If no \fIformulae\fR a
 .IP
 If \fB\-\-hide=\fR\fIhidden\fR is passed, act as if none of \fIhidden\fR are installed\. \fIhidden\fR should be a comma\-separated list of formulae\.
 .
+.IP
+\fBmissing\fR exits with a non\-zero status if any formulae are missing dependencies\.
+.
 .TP
 \fBoptions\fR [\fB\-\-compact\fR] (\fB\-\-all\fR|\fB\-\-installed\fR|\fIformulae\fR)
 Display install options specific to \fIformulae\fR\.


ilovezfs:
Can you explain how this is a workaround for https://github.com/Homebrew/homebrew-core/pull/26497 ?

scpeters:
The issue that https://github.com/Homebrew/homebrew-core/pull/26497 aimed to fix was a bottle that generates `brew missing` complaints (via `brew doctor`), which breaks our CI. Rather than remove `brew doctor` from our CI completely, my plan is to just try to install any packages that `brew missing` thinks are necessary (even though in this example, `gobject-introspection` is only a build depend of one of `gtkglext`'s dependencies). It makes it a little easier to implement the installation logic if `brew missing` returns a failure.

~~~
$ brew install gtkglext
...
$ brew doctor
...
Warning: Some installed formulae are missing dependencies.
You should `brew install` the missing dependencies:
  brew install gobject-introspection

Run `brew missing` for more details.
$ brew missing
gtkglext: gobject-introspection
$ brew missing || brew install $(brew missing | awk '{print $2}') && brew missing && echo success
...
==> Installing gobject-introspection
...
success
~~~

This PR enables the line, which is my current preferred workaround.

~~~
$ brew missing || brew install $(brew missing | awk '{print $2}') && brew missing
~~~

ilovezfs:
Ah, OK. I think `brew install --only-dependencies $(brew list)` does what you want.

This PR makes sense regardless, of course.

scpeters:
In most cases, `brew install --only-dependencies $(brew list)` will work, but the `gtkglext` high sierra bottle (not the other bottles) currently has `gobject-introspection` in its `INSTALL_RECEIPT.json` but it doesn't show up in `brew deps gtkglext`.

~~~
$ brew install --only-dependencies gtkglext
All dependencies for gtkglext are satisfied.
$ brew missing
gtkglext: gobject-introspection
$ brew deps gtkglext
atk
cairo
fontconfig
freetype
fribidi
gdk-pixbuf
gettext
glib
graphite2
gtk+
harfbuzz
hicolor-icon-theme
icu4c
jpeg
libffi
libpng
libtiff
pango
pcre
pixman
~~~

So this `brew missing` trick is helpful for these old bottle edge cases.

ilovezfs:
That's confusing if `brew install --only-dependencies gtkglext` can still leave `brew missing` reporting dependencies missing for `gtkglext`.

Also, failing CI over bottles we're intentionally not updating seems bogus and you shouldn't have to fill in those dependencies as a workaround.

scpeters:
Well, that is why I submitted https://github.com/Homebrew/homebrew-core/pull/26497 initially, but I don't mind this workaround if the principle expressed by @MikeMcQuaid in https://github.com/Homebrew/homebrew-core/pull/26497#issuecomment-380677042 is important.

ilovezfs:
The idea was that new bottles were no longer supposed to be needed for that scenario. The idea was not that you should have to hack around problems that cause CI failures.

MikeMcQuaid:
I'm just back home from a work trip and may take a few days to get to this but: yes, there should be no circumstance in which we're displaying output that means that the solution is to rebottle just to change an `INSTALL_RECEIPT`.

I'm still not sure I understand what breaks your CI here, can you elaborate further? Is it that you're installing a bottle from scratch and it immediately makes `brew doctor` and `brew missing` fail?

scpeters:
We use [brew install --only-dependencies](https://bitbucket.org/osrf/release-tools/src/ab2191530b1bef77cbdf451b8a8f8e2c65ba0076/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash?at=default&fileviewer=file-view-default#project-default-devel-homebrew-amd64.bash-73) to resolve dependencies, then [manually build from source](https://bitbucket.org/osrf/release-tools/src/ab2191530b1bef77cbdf451b8a8f8e2c65ba0076/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash?at=default&fileviewer=file-view-default#project-default-devel-homebrew-amd64.bash-119:131) and install into `brew --cellar` and run `brew link`. We then run `brew doctor`, I think so we could catch any errors we had with `brew link` (to be honest I can't remember). I could change it to `brew doctor || true` so we would still have the console output, but it wouldn't fail the build. Or we could suppress `check_missing_deps`, as @ilovezfs suggested.

MikeMcQuaid:
@scpeters I'm not sure I see what I'm looking at on the manually building from source. How do you have an install receipt at all when not using Homebrew to do that?

MikeMcQuaid:
So the issue is if you run `brew install gtkglext` and then `brew missing` or `brew doctor` they show missing deps?

scpeters:
> So the issue is if you run `brew install gtkglext` and then `brew missing` or `brew doctor` they show missing deps?

Yes, exactly, though only on high sierra (in case anyone tries to reproduce). I believe the older bottles were built with older versions of homebrew.

ilovezfs:
More weirdness with missing:
```
Josephs-MacBook-Pro:~ joe$ brew outdated subversion
subversion (1.9.7_3) < 1.10.0
Josephs-MacBook-Pro:~ joe$ brew missing
subversion: lz4 utf8proc
Josephs-MacBook-Pro:~ joe$
```
but lz4 and utf8proc are deps of 1.10.0 not 1.9.7_3.

MikeMcQuaid:
Will continue to talk here when I've thought about a better plan but this PR is 👍 as-is, thanks @scpeters!

MikeMcQuaid:
> More weirdness with missing:

@ilovezfs Fixed your case with https://github.com/Homebrew/brew/pull/4118.

MikeMcQuaid:
@scpeters @ilovezfs I'm still trying to figure out the best fix for that scenario. Some options I can see:
- store all build dependencies in the tab along with runtime dependencies on a formula and when the build dependencies differ from the formula: recalculate the runtime dependencies on pour
- have some `audit` or `test` or `linkage` that means that CI will fail if there's not a `revision` bump made when runtime dependencies on a formula have been removed.
- something else

Thoughts?

scpeters:
> * store all build dependencies in the tab along with runtime dependencies on a formula and when the build dependencies differ from the formula: recalculate the runtime dependencies on pour

This approach seems to try to solve the problem by making `brew` smarter.

> * have some audit or test or linkage that means that CI will fail if there's not a revision bump made when runtime dependencies on a formula have been removed.

This approach seems to try to detect problems and rebuild bottles more often.

I don't spend enough time with `brew` to feel confident in commenting on whether the first approach will work or not, though I think it is more complex. I think the second approach has simpler logic but involves a little more time spent rebuilding bottles. For my own sake, I would lean towards the second because I think it would be easier to understand, but I don't know how that impacts maintainer workload.

PR #4067, opened 2018-04-13
list.rb: use CLI::Parser and improve handling of flags

maxim-belkin:

-----

`brew list`:
* When --multiple is specified without --versions it results in the
execution of the `ls` command with `--multiple` flag, which results in a
failure. Here we assume that `--versions` flag was forgotten when  a lone`--multiple` option is provided.

* Use OpenStruct to handle arguments for `list`

**Note**: alternative solution to setting mode.versions? to true (line 39) would be to `odie` instead of `opoo` on line 38.

### DIFF
--- a/Library/Homebrew/cmd/list.rb
+++ b/Library/Homebrew/cmd/list.rb
@@ -31,7 +31,7 @@ module Homebrew
       switch "--pinned"
       switch "--versions"
       switch "--full-name"
-      switch "--multiple", required_for: "--versions"
+      switch "--multiple", depends_on: "--versions"
       switch :verbose
     end
 
--- a/Library/Homebrew/cmd/list.rb
+++ b/Library/Homebrew/cmd/list.rb
@@ -20,13 +20,28 @@
 
 require "metafiles"
 require "formula"
+require "cli_parser"
 
 module Homebrew
   module_function
 
   def list
+    @args = CLI::Parser.parse do
+      switch "--unbrewed"
+      switch "--pinned"
+      switch "--versions"
+      switch "--full-name"
+      switch "--multiple"
+      switch :verbose
+    end
+
+    if @args.multiple? && !@args.versions?
+      opoo "--multiple requires --versions"
+      @args.send("versions?=", true)
+    end
+
     # Use of exec means we don't explicitly exit
-    list_unbrewed if ARGV.flag? "--unbrewed"
+    list_unbrewed if @args.unbrewed?
 
     # Unbrewed uses the PREFIX, which will exist
     # Things below use the CELLAR, which doesn't until the first formula is installed.
@@ -35,10 +50,10 @@ module Homebrew
       return
     end
 
-    if ARGV.include?("--pinned") || ARGV.include?("--versions")
+    if @args.pinned? || @args.versions?
       filtered_list
     elsif ARGV.named.empty?
-      if ARGV.include? "--full-name"
+      if @args.full_name?
         full_names = Formula.installed.map(&:full_name).sort(&tap_and_name_comparison)
         return if full_names.empty?
         puts Formatter.columns(full_names)
@@ -46,7 +61,7 @@ module Homebrew
         ENV["CLICOLOR"] = nil
         exec "ls", *ARGV.options_only << HOMEBREW_CELLAR
       end
-    elsif ARGV.verbose? || !$stdout.tty?
+    elsif Homebrew.args.verbose? || !$stdout.tty?
       exec "find", *ARGV.kegs.map(&:to_s) + %w[-not -type d -print]
     else
       ARGV.kegs.each { |keg| PrettyListing.new keg }
@@ -87,13 +102,13 @@ module Homebrew
     dirs.delete "etc"
     dirs.delete "var"
 
-    args = dirs.sort + %w[-type f (]
-    args.concat UNBREWED_EXCLUDE_FILES.flat_map { |f| %W[! -name #{f}] }
-    args.concat UNBREWED_EXCLUDE_PATHS.flat_map { |d| %W[! -path #{d}] }
-    args.concat %w[)]
+    arguments = dirs.sort + %w[-type f (]
+    arguments.concat UNBREWED_EXCLUDE_FILES.flat_map { |f| %W[! -name #{f}] }
+    arguments.concat UNBREWED_EXCLUDE_PATHS.flat_map { |d| %W[! -path #{d}] }
+    arguments.concat %w[)]
 
     cd HOMEBREW_PREFIX
-    exec "find", *args
+    exec "find", *arguments
   end
 
   def filtered_list
@@ -106,7 +121,7 @@ module Homebrew
         rack.exist?
       end
     end
-    if ARGV.include? "--pinned"
+    if @args.pinned?
       pinned_versions = {}
       names.sort.each do |d|
         keg_pin = (HOMEBREW_PINNED_KEGS/d.basename.to_s)
@@ -115,12 +130,12 @@ module Homebrew
         end
       end
       pinned_versions.each do |d, version|
-        puts d.basename.to_s.concat(ARGV.include?("--versions") ? " #{version}" : "")
+        puts d.basename.to_s.concat(@args.versions? ? " #{version}" : "")
       end
     else # --versions without --pinned
       names.sort.each do |d|
         versions = d.subdirs.map { |pn| pn.basename.to_s }
-        next if ARGV.include?("--multiple") && versions.length < 2
+        next if @args.multiple? && versions.length < 2
         puts "#{d.basename} #{versions * " "}"
       end
     end
--- a/Library/Homebrew/cmd/list.rb
+++ b/Library/Homebrew/cmd/list.rb
@@ -31,13 +31,8 @@ module Homebrew
       switch "--pinned"
       switch "--versions"
       switch "--full-name"
-      switch "--multiple"
+      switch "--multiple", required_for: "--versions"
       switch :verbose
-    end
-
-    if @args.multiple? && !@args.versions?
-      opoo "--multiple requires --versions"
-      @args.send("versions?=", true)
     end
 
     # Use of exec means we don't explicitly exit


woodruffw:
> Note: alternative solution to setting mode.versions? to true (line 39) would be to odie instead of opoo on line 38.

:+1: on `odie` -- that `send` is a little confusing.

maxim-belkin:
well, in that case it will go away! 👍

GauthamGoli:
@maxim-belkin Please use `Homebrew::CLI::Parser` for parsing args, it uses `OpenStruct`

Library/Homebrew/cmd/list.rb
```diff
@@ -35,18 +50,18 @@ def list
...
+      if args.full_name?
         full_names = Formula.installed.map(&:full_name).sort(&tap_and_name_comparison)
         return if full_names.empty?
         puts Formatter.columns(full_names)
       else
         ENV["CLICOLOR"] = nil
         exec "ls", *ARGV.options_only << HOMEBREW_CELLAR
       end
-    elsif ARGV.verbose? || !$stdout.tty?
+    elsif args.verbose? || !$stdout.tty?
```
GauthamGoli: `Homebrew.args.verbose?`

Library/Homebrew/cmd/list.rb
```diff
@@ -35,18 +50,18 @@ def list
       return
     end
 
-    if ARGV.include?("--pinned") || ARGV.include?("--versions")
-      filtered_list
+    if args.pinned? || args.versions?
+      filtered_list(args)
```
GauthamGoli: Could make `args` -> `@args`, instance variable, then its not needed to pass it as an argument.

Library/Homebrew/cmd/list.rb
```diff
@@ -20,13 +20,28 @@
 
 require "metafiles"
 require "formula"
+require "cli_parser"
 
 module Homebrew
   module_function
 
   def list
+    args = Homebrew::CLI::Parser.parse do
```
reitermarkus: No need for the `Homebrew::` prefix, since this is already inside it.

Library/Homebrew/cmd/list.rb
maxim-belkin: what is the purpose of adding `Homebrew.` ?

GauthamGoli: Options like `verbose`, `force`, `quiet` and `debug` are project wide, hence they are stored as global vars to be accessible everywhere. 

If we have `--verbose` passed to a command and that calls method defined in some other file, it might also need `--verbose` and we don't want to be passing around them as parameters if possible.

Library/Homebrew/cmd/list.rb
```diff
@@ -20,13 +20,28 @@
...
+    @args = CLI::Parser.parse do
+      switch "--unbrewed"
+      switch "--pinned"
+      switch "--versions"
+      switch "--full-name"
+      switch "--multiple"
+      switch :verbose
+    end
+
+    if args.multiple? && !args.versions?
```
GauthamGoli: You'd have to use `@args` everywhere else

Library/Homebrew/cmd/list.rb
```diff
@@ -35,18 +50,18 @@ def list
...
+      if @args.full_name?
         full_names = Formula.installed.map(&:full_name).sort(&tap_and_name_comparison)
         return if full_names.empty?
         puts Formatter.columns(full_names)
       else
         ENV["CLICOLOR"] = nil
         exec "ls", *ARGV.options_only << HOMEBREW_CELLAR
       end
-    elsif ARGV.verbose? || !$stdout.tty?
+    elsif Homebrew.@args.verbose? || !$stdout.tty?
```
GauthamGoli: This will still be `Homebrew.args.verbose?`

GauthamGoli:
Looks good. Can you please `squash` the commits into a single/couple commit(s) ?

maxim-belkin:
Is `opoo` message fine? Or should I change it to `odie` (and remove the line after it)?

woodruffw:
I'm still :+1: on making it `odie`, but will defer to @GauthamGoli on it :slightly_smiling_face:

GauthamGoli:
`brew list` can take other undeclared arguments like `-t` which would be passed to `ls`, there's bigger problem here, it's because of the option parser itself. It never accepts undeclared options and passing some undeclared option causes it to output invalid option error.

I guess, until there's solution for this, we should keep this PR on hold.

maxim-belkin:
does `cli_parser` allow something like `flag "*"`?

maxim-belkin:
Looks like `flag "-[ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1]*"` works, no?

GauthamGoli:
We'd prefer not to declare options of `ls` inside brew

maxim-belkin:
makes sense 

I guess the solution is to define acceptable `LS_FLAGS` (different on Mac and Linux) and use them here.

GauthamGoli:
Better solution is to add an optional feature to Parser to not fail when undeclared options are passed, rather store them separately.

maxim-belkin:
1. I think making Parser not fail for undeclared options is one part of the problem, whereas limiting acceptable (undeclared) flags for every command that accepts such flags is another.

2. In the current implementation, it is not possible to pass flags that Parser knows: `-f`, `-v`, `-q`, or `-d`. How about `--lsflags` switch that would pass all the remaining args to `ls`?

GauthamGoli:
Point `1` is solved, I have a working implementation for making Parser not fail for undeclared options.

I am yet to find a way to solve point number `2` and I think its inherent to `OptionParser` hence difficult to override.

maxim-belkin:
I haven't seen your implementation, but make sure it doesn't let users pass `; /bin/rm -rf / 2>/dev/null` as an option to `ls` :)

GauthamGoli:
User can always specify `; rm -rf /` the shell itself would interpret it as a command and we wouldn't have control over it. So its not command injection. Also, as we are not offering `brew as a Service` on a server, we wouldn't have to worry about command injection :)

GauthamGoli:
Here’s snippet from `OptionParser`’s source
```ruby
begin
  sw, = complete(:short, opt)
  # short option matched.
  val = arg.sub(/\A-/, '')
  has_arg = true
rescue InvalidOption
  # if no short options match, try completion with long
  # options.
  sw, = complete(:long, opt)
  eq ||= !rest
end
```
The above code makes undeclared short options (Ex: `-f` ) to match with long options (Ex: `--filename` ) instead of raising `InvalidOption` , as a result leads to some issues.  Couple of issues mentioned by @maxim-belkin https://github.com/Homebrew/brew/pull/4067#issuecomment-381741609

None of the OptionParsers from other standard libraries have this behaviour (Even ruby’s `Getoptlong`, and python’s `argparse` ) 

I have also raised a discussion with `OptionParser`s maintainer here https://bugs.ruby-lang.org/issues/14692 , asking him if its possible for a workaround/feature implementation to suppress this behavior.

There a no way to suppress this default behaviour, I think. 
- I tried implementing this functionality outside `OptionParser` but there are many edge cases, makes the code hard to read/maintain. Basically it is almost like writing `OptionParser`'s logic.
- We could use `Getoptlong` instead of `OptionParser` but would lose(miss?) a lot of "off-the-shelf" functionality provided by `OptionParser` which we might want to use in future.
- Don't use `Homebrew::CLI::Parser` for `brew list` until there's a work around.

 What’s the best way forward / any suggestions anyone? 
cc @MikeMcQuaid @alyssais

ilovezfs:
yikes.

reitermarkus:
You could use a refinement to monkey-patch that part of `OptionParser`, but you'd have to copy-paste the whole `parse_in_order` method since the relevant part is deep inside that method.

> We could use `Getoptlong` instead of `OptionParser` but would lose(miss?) a lot of "off-the-shelf" functionality provided by `OptionParser` which we might want to use in future.

Which “off-the-shelf” functionality does `OptionParser` provide?

MikeMcQuaid:
> Point 1 is solved, I have a working implementation for making Parser not fail for undeclared options.

This should be something that is not on by default.

> I am yet to find a way to solve point number 2 and I think its inherent to OptionParser hence difficult(edit: not possible?) to override.

I think this is fine. I don't think it's essential to be able to pass those options through to `ls` and I think `brew` handling those options consistently is better.

GauthamGoli:
Yep, monkey patching would work, but we'd have to be mindful of changes in that library whenever we change Ruby version. 

> Which “off-the-shelf” functionality does OptionParser provide?

Data types support for arguments , Documentation/help text generation. 
We'd have to implement these ourselves, meaning more maintenance

MikeMcQuaid:
> Yep, monkey patching would work for now, but we'd have to be mindful of changes in that library whenever we change Ruby version.

I'm a pretty hard 👎 on any Ruby internal class monkey patching. I'd sooner deprecate/disable the "pass arbritrary options to `brew ls`" than do that.

reitermarkus:
> I'd sooner deprecate/disable the "pass arbitrary options to `brew ls`" than do that.

I for one have never used this feature.

maxim-belkin:
🍿 &nbsp;  🤔 &nbsp;  😕   &nbsp;   :trollface: 
----
 I'm +1 for disabling arbitrary options for `brew ls`.

maxim-belkin:
what about `opoo` _vs._ `odie`? 🙄 

https://github.com/Homebrew/brew/pull/4067/files#diff-a1ae5d36c5006e319d3b6d3c8d07f58fR39

reitermarkus:
> what about `opoo` _vs._ `odie`? 🙄 

Might want to wait for https://github.com/Homebrew/brew/pull/3987, which allows specifying conflicting and depending options.

MikeMcQuaid:
@GauthamGoli Any news on this?

GauthamGoli:
@MikeMcQuaid  My bad. I'll get back on this today evening.

Library/Homebrew/cmd/list.rb
```diff
@@ -20,13 +20,28 @@
...
+    @args = CLI::Parser.parse do
+      switch "--unbrewed"
+      switch "--pinned"
+      switch "--versions"
+      switch "--full-name"
+      switch "--multiple"
+      switch :verbose
+    end
+
+    if @args.multiple? && !@args.versions?
```
GauthamGoli: This if block can be removed. 
Specifying constraints on options is now possible. You can have a look at `cli_parser_spec.rb`  to look at corresponding test cases.

Library/Homebrew/cmd/list.rb
```diff
@@ -20,13 +20,28 @@
...
 module Homebrew
   module_function
 
   def list
+    @args = CLI::Parser.parse do
+      switch "--unbrewed"
+      switch "--pinned"
+      switch "--versions"
+      switch "--full-name"
+      switch "--multiple"
```
GauthamGoli: `switch "--multiple", required_for: "--versions"`

Library/Homebrew/cmd/list.rb
maxim-belkin: do I need to add anything here? If so, what is the difference between `required_for` and `depends_on` (except for expressing opposite direction of a relationship) ?

Library/Homebrew/cmd/list.rb
reitermarkus: Don't you mean `depends_on`? `--multiple` is optional.

Library/Homebrew/cmd/list.rb
```diff
@@ -20,13 +20,23 @@
...
 module Homebrew
   module_function
 
   def list
+    @args = CLI::Parser.parse do
+      switch "--unbrewed"
+      switch "--pinned"
+      switch "--versions"
+      switch "--full-name"
+      switch "--multiple", required_for: "--versions"
```
reitermarkus: Should be `depends_on`.

maxim-belkin: So, what does the phrase
```
`--multiple` requires `--versions`
```
in plain English corresponds to in this new syntax?
```
switch "--multiple", depends_on: "--versions"
```
?

reitermarkus:
```ruby
switch "--multiple", depends_on: "--versions"
```

`--multiple` is optional and can only be given as a “sub-option” for `--versions`.


```ruby
switch "--multiple", required_for: "--versions"
```

When `--versions` is given, `--multiple` also has to be specified.

Library/Homebrew/cmd/list.rb
GauthamGoli: Ops.. I got that wrong, you're right.

GauthamGoli:
@maxim-belkin Sorry. I added constraints for flags but not switches. I'll get that done very soon, and then we could proceed with this.

GauthamGoli:
@maxim-belkin Please rebase with master

GauthamGoli:
I have tested this locally. Will be merged when @MikeMcQuaid gives 👍

MikeMcQuaid:
Thanks @maxim-belkin! If people go 😭 over the death of custom flags we'll think about fiddling with this more but I'm not sure it's worth blocking this PR over.

MikeMcQuaid:
> -1 is not necessary when piping output to another command.

Regardless, given it's used by others, we should fix this.

PR #4105, opened 2018-04-24
components_order_cop: make non-strict

commitay:

-----
Also whitelist [`aspell`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/aspell.rb)

### DIFF
--- a/Library/Homebrew/rubocops/components_order_cop.rb
+++ b/Library/Homebrew/rubocops/components_order_cop.rb
@@ -2,7 +2,7 @@ require_relative "./extend/formula_cop"
 
 module RuboCop
   module Cop
-    module FormulaAuditStrict
+    module FormulaAudit
       # This cop checks for correct order of components in a Formula
       #
       # - component_precedence_list has component hierarchy in a nested list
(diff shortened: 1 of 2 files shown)


Library/Homebrew/rubocops/components_order_cop.rb
```diff
@@ -62,8 +62,13 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
           end
         end
 
+        WHITELIST = %w[
+          aspell
+        ].freeze
+
         # Method to format message for reporting component precedence violations
         def component_problem(c1, c2)
+          return if @formula_name.match?(*WHITELIST)
```
MikeMcQuaid:
```ruby
return WHITELIST.include?(@formula_name)
```

to do an exact match.

Library/Homebrew/rubocops/components_order_cop.rb
```diff
@@ -62,8 +62,13 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
           end
         end
 
+        WHITELIST = %w[
```
MikeMcQuaid: Can we get an inline code comment on why aspell is getting whitelist? Thanks!

MikeMcQuaid:
Thanks @commitay!

PR #4110, opened 2018-04-25
formula_desc_cop desclength: make non-strict

commitay:

-----

3 formulae fail `DescLength`

15 formulae fail `Desc`

### DIFF
--- a/Library/Homebrew/rubocops/formula_desc_cop.rb
+++ b/Library/Homebrew/rubocops/formula_desc_cop.rb
@@ -3,7 +3,7 @@ require_relative "../extend/string"
 
 module RuboCop
   module Cop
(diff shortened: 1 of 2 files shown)


ilovezfs:
We'll need more whitelisting if `desc`, not just `desc` length, is becoming non-strict.

MikeMcQuaid:
I'm 🆒 with this but I guess makes sense to wait until Homebrew/core is sorted (and any whitelisting added first).

ilovezfs:
Ideally the three that are over the max length should be fixed first, but there are only three. The non-length checks are still strict, and will probably need to remain that way.

ilovezfs:
🍨

PR #4113, opened 2018-04-27
components_order_cop: add post_install

commitay:

-----
https://github.com/Homebrew/homebrew-core/pull/27153

### DIFF
--- a/Library/Homebrew/rubocops/components_order_cop.rb
+++ b/Library/Homebrew/rubocops/components_order_cop.rb
@@ -31,6 +31,7 @@ module RuboCop
             [{ name: :conflicts_with, type: :method_call }],
             [{ name: :go_resource, type: :block_call }, { name: :resource, type: :block_call }],
             [{ name: :install, type: :method_definition }],
+            [{ name: :post_install, type: :method_definition }],
             [{ name: :caveats, type: :method_definition }],
             [{ name: :plist_options, type: :method_call }, { name: :plist, type: :method_definition }],
             [{ name: :test, type: :block_call }],


commitay:
Others that would be nice to have audited:

- `deprecated_option`: 36 formulae

- `fails_with`: 13 formulae

- `needs`: 33 formulae

- `patch`: 19 formulae

MikeMcQuaid:
> Others that would be nice to add to audit:
> 
> deprecated_option: 36 formulae
> 
> fails_with: 13 formulae
> 
> needs: 33 formulae
> 
> patch: 19 formulae

I'm 👍 on all of them. Could you add them to this PR?

commitay:
Adding them individually would be a bit easier to do (and review), all that once makes the diff > 1000 lines with changes to whitespace and comments.

ilovezfs:
I'm not sure there's a clear benefit to ordering those.

commitay:
Not adding them means that the current audit is less consistent and also allows components to be grouped when in the majority of formulae they are separated.

https://github.com/Homebrew/homebrew-core/compare/master...commitay:needs

ilovezfs:
Well, for example, I don't see any good coming from moving all the deprecated options around. They're already grouped. Whether they're above or below the options isn't affecting us and will just blow up the Git history. I don't think "more consistent" justifies things that aren't causing us problems.

MikeMcQuaid:
> Adding them individually would be a bit easier to do (and review), all that once makes the diff > 1000 lines with changes to whitespace and comments (that need to be fixed manually).

Seems reasonable to me. It's also easier to discuss their pros/cons one by one (as @ilovezfs has done eloquently above).

MikeMcQuaid:
Thanks @commitay!

PR #4116, opened 2018-04-28
Fix `zap --force` for uninstalled casks

commitay:

-----
Fixes https://github.com/caskroom/homebrew-cask/issues/46466


commitay:
>  if --force should be required to zap an already uninstalled formula (which I am not sure it should: the default behaviour could be to zap even if app is not there)

`zap` runs `uninstall` first and `uninstall` will fail for some casks if that cask is not installed (see https://github.com/caskroom/homebrew-cask/issues/36162#issuecomment-314075741). Using `--force` here is also consistent with other commands.

> there should be a mention of it on the error message

Yes, the error message can be improved.

PR #4125, opened 2018-04-30
Support GCC 8

fxcoudert:
Support GCC 8 as its release is nearing: https://gcc.gnu.org/ml/gcc/2018-04/msg00187.html

@MikeMcQuaid


MikeMcQuaid:
Any reason(s) to not just add `9` in here while we're at it?

fxcoudert:
No reason, except that it does not exist :)
But if you prefer it in, I'll do it.

Issue #4132, opened 2018-05-02
brew update fails when using Include in ssh config


poochiethecat:
**Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

- [X] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [X] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new. If it's a `brew cask` problem please file this issue at https://github.com/caskroom/homebrew-cask/issues/new. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [X] ran `brew update` and can still reproduce the problem?
- [X] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [X] ran `brew config` and `brew doctor` and included their output with your issue?

`brew doctor`
```
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry or file an issue; just ignore this. Thanks!

Warning: "config" scripts exist outside your system or Homebrew directories.
`./configure` scripts often look for *-config scripts to determine if
software packages are installed, and what additional flags to use when
compiling and linking.

Having additional scripts in your path can confuse software installed via
Homebrew if the config script overrides a system or Homebrew provided
script of the same name. We found the following "config" scripts:
  /Users/poochiethecat/.pyenv/shims/icu-config
  /Users/poochiethecat/.pyenv/shims/libpng16-config
  /Users/poochiethecat/.pyenv/shims/llvm-config
  /Users/poochiethecat/.pyenv/shims/ncurses5-config
  /Users/poochiethecat/.pyenv/shims/ncursesw5-config
  /Users/poochiethecat/.pyenv/shims/ncursesw6-config
  /Users/poochiethecat/.pyenv/shims/pcre-config
  /Users/poochiethecat/.pyenv/shims/python-config
  /Users/poochiethecat/.pyenv/shims/python2-config
  /Users/poochiethecat/.pyenv/shims/python2.7-config
  /Users/poochiethecat/.pyenv/shims/python3-config
[...]

PR #4133, opened 2018-05-02
add option to force use of brewed ssh in update command

poochiethecat:

-----
`openssh` added the declaration `Include` in version 7.3p1 (see [here](https://superuser.com/a/1142813)).

The system version of openssh might not support this (it does not on my system), so when I changed my config to use `Include`, I also had to change some other things like `GIT_SSH_COMMAND` to use the correct ssh version.

After that, Homebrew failed to update as I described in this issue: [brew update fails when using Include in ssh config](https://github.com/Homebrew/brew/issues/4132)

I analyzed the brew update commands code and identified the problem.
On line 397 GIT_SSH_COMMAND is set to "ssh -oBatchMode=yes". Since Homebrew replaces PATH when filtering the environment, ssh in this case is the system provided executable in /usr/bin.
As far as I could find, there is currently no way to make Homebrew use the brewed ssh (although it actually does so in pretty much every other case if I correctly set my GIT_SSH_COMMAND environment variable.

This pull-request adds an environment variable HOMEBREW_FORCE_BREWED_SSH, analogous to HOMEBREW_FORCE_BREWED_CURL, that allows to force Homebrew to use the brewed ssh executable instead of the system provided one.

I don't know how I could write a test that adequately tests if this works, but if someone can give me some pointers as to how I would approach that, I would be happy to add them as well.

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -395,8 +395,18 @@ EOS
       odie "Git must be installed and in your PATH!"
     fi
   fi
+  if [[ -n "$HOMEBREW_FORCE_BREWED_SSH" ]]
+    then
+    if [[ ! -x "$HOMEBREW_PREFIX/bin/ssh" ]]
+      then
+        brew install openssh
+    fi
+    HOMEBREW_SSH_EXECUTABLE="$HOMEBREW_PREFIX/bin/ssh"
+  else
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Sorry, we're not interested in this as forcing a Homebrew'd `curl` is necessary in some configurations (e.g. old versions of macOS) for security reasons whereas this is not. My recommendation is you adjust your SSH configuration files accordingly.

PR #4146, opened 2018-05-03
cli: Store parsed args in Homebrew.args and access @args everywhere else

GauthamGoli:

-----
https://github.com/Homebrew/brew/pull/3987#issuecomment-386190815

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -53,7 +53,7 @@ module Homebrew
   module_function
 
   def audit
-    args = Homebrew::CLI::Parser.parse do
+    Homebrew::CLI::Parser.parse do
       switch      "--strict"
       switch      "--online"
       switch      "--new-formula"
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -71,7 +71,7 @@ module Homebrew
   module_function
 
   def bottle
-    @args = Homebrew::CLI::Parser.parse do
+    Homebrew::CLI::Parser.parse do
       switch "--merge"
       switch "--skip-relocation"
       switch "--force-core-tap"
@@ -86,7 +86,7 @@ module Homebrew
       flag   "--root-url"
     end
 
-    return merge if @args.merge?
+    return merge if args.merge?
     ARGV.resolved_formulae.each do |f|
       bottle_formula f
     end
@@ -117,7 +117,7 @@ module Homebrew
       linked_libraries = Keg.file_linked_libraries(file, string)
       result ||= !linked_libraries.empty?
 
-      if Homebrew.args.verbose?
+      if args.verbose?
         print_filename.call(string, file) unless linked_libraries.empty?
         linked_libraries.each do |lib|
           puts " #{Tty.bold}-->#{Tty.reset} links to #{lib}"
@@ -140,7 +140,7 @@ module Homebrew
         end
       end
 
-      next unless Homebrew.args.verbose? && !text_matches.empty?
+      next unless args.verbose? && !text_matches.empty?
       print_filename.call(string, file)
       text_matches.first(MAXIMUM_STRING_MATCHES).each do |match, offset|
         puts " #{Tty.bold}-->#{Tty.reset} match '#{match}' at offset #{Tty.bold}0x#{offset}#{Tty.reset}"
@@ -161,7 +161,7 @@ module Homebrew
       absolute_symlinks_start_with_string << pn if link.to_s.start_with?(string)
     end
 
-    if Homebrew.args.verbose?
+    if args.verbose?
       unless absolute_symlinks_start_with_string.empty?
         opoo "Absolute symlink starting with #{string}:"
         absolute_symlinks_start_with_string.each do |pn|
@@ -184,7 +184,7 @@ module Homebrew
     end
 
     unless tap = f.tap
-      unless @args.force_core_tap?
+      unless args.force_core_tap?
         return ofail "Formula not from core or any taps: #{f.full_name}"
       end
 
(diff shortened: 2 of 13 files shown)


GauthamGoli:
We could either do it this way or rename class var `@args` in `global.rb` to something else and replace occurences of `Homebrew.args` to the new name everywhere else

MikeMcQuaid:
> We could either do it this way or rename class var @args in global.rb to something else and replace occurences of Homebrew.args to the new name everywhere else

I think that would be preferable. How about `@global_args`? 

Whatever we do: can you update `brew bump-formula-pr` in this PR to be consistent too? Thanks!

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -59,7 +59,7 @@
     <% checksums.each do |checksum_type, checksum_values| %>
     <% checksum_values.each do |checksum_value| %>
     <% checksum, macos = checksum_value.shift %>
-    <%= checksum_type %> "<%= checksum %>" => :<%= macos %><%= "_or_later" if Homebrew.args.or_later? %>
+    <%= checksum_type %> "<%= checksum %>" => :<%= macos %><%= "_or_later" if Homebrew.global_args.or_later? %>
```
GauthamGoli: What does the above code do?

MikeMcQuaid: Outputs the `brew bottle` output and appends `or_later` if that argument is specified.

Library/Homebrew/dev-cmd/bump-formula-pr.rb
```diff
@@ -48,7 +48,7 @@ module Homebrew
   module_function
 
   def bump_formula_pr
-    @bump_args = Homebrew::CLI::Parser.parse do
+    @args = Homebrew::CLI::Parser.parse do
```
MikeMcQuaid: Thinking about this more: I think it would be ideal if you could always just use `Homebrew.args` (or `args` if inside the `Homebrew` module already) and have `Homebrew::CLI::Parser.parse` add to the existing `args` for global values rather than having two sets that you need to query at separate times. Thoughts?

Library/Homebrew/dev-cmd/bottle.rb
MikeMcQuaid: Pretty sure this shouldn't be `global_args`?

Library/Homebrew/dev-cmd/bump-formula-pr.rb
GauthamGoli: Seems sensible to store everything in `Homebrew.args` I'll get this to work and if there are no issues will update the PR.

GauthamGoli: @MikeMcQuaid , the PR is updated now.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -70,13 +70,13 @@ def audit
     end
 
     Homebrew.auditing = true
-    inject_dump_stats!(FormulaAuditor, /^audit_/) if args.audit_debug?
+    inject_dump_stats!(FormulaAuditor, /^audit_/) if @args.audit_debug?
```
MikeMcQuaid: As we're in the `Homebrew` module: can this use either just `args` or `Homebrew.args`?

reitermarkus:
It feels really weird using a global variable `Homebrew.args` inside a `CLI::Parser` instance.

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: Addressed. Using `args` now

MikeMcQuaid:
> It feels really weird using a global variable Homebrew.args inside a CLI::Parser instance.

@reitermarkus What would you suggest instead?

reitermarkus:
Use 

```ruby
args = Homebrew::CLI::Parser.parse
```

instead of 

```ruby
@args = Homebrew::CLI::Parser.parse
```

everywhere. Then pass `args` to other methods that used `@args` before and use `Homebrew.args` to access global arguments.

GauthamGoli:
@MikeMcQuaid which way should this be done ? 
Both seem fine to me, with the implementation in this PR, its more like `Homebrew.args` is single source of truth, we can be sure args would be available in cross command module method calls, while with what @reitermarkus is suggesting we'd have to take care that args are passed around wherever required which I think might possibly lead to edge cases( or may be not too..)

reitermarkus:
> I think might possibly lead to edge cases( or may be not too..)

I think there will be more edge cases (this PR included) the more global state there is.

MikeMcQuaid:
@GauthamGoli Do you think if `Homebrew.args` is used only for global arguments it will be sufficient to only use e.g. `@args` (or `args`) in `cmd/*.rb` files? If so, I think that's the best solution and we can consider making things like `--build-bottle` global arguments (as they aren't specific to a single command). Thoughts?

ilovezfs:
```
brew info --build-bottle foo
```
should have interesting output.

MikeMcQuaid:
> should have interesting output.

@ilovezfs can you elaborate a bit on "interesting" 😉

GauthamGoli:
> Do you think if Homebrew.args is used only for global arguments it will be sufficient to only use e.g. @ args (or args) in cmd/*.rb files? 

@MikeMcQuaid  Yes, all files except for `update-report.rb` , it has a `Reporter` class outside `Homebrew` module, we'd have to pass `args` when instantiating that.

ilovezfs:
>@ilovezfs can you elaborate a bit on "interesting" 

Presumably we want it to say `--build-bottle` isn't a valid option for `info`. But if `--build-bottle` is elevated to global status, I'm not sure how that's supposed to happen.

reitermarkus:
> Presumably we want it to say `--build-bottle` isn't a valid option for `info`.

Exactly, otherwise what is the point in using an option parser? If it doesn't fail on invalid options, we could have just stayed with `ARGV.include?`.

MikeMcQuaid:
> Presumably we want it to say --build-bottle isn't a valid option for info. But if --build-bottle is elevated to global status, I'm not sure how that's supposed to happen.

Agreed. From that argument then `--force` and maybe even `--quiet` and `--verbose` shouldn't be global arguments either (as they don't do something for all commands).

> @MikeMcQuaid Yes, all files except for update-report.rb , it has a Reporter class defined outside Homebrew module, we'd have to pass args when instantiating that.

This is fine. Alternatively we can consider moving it into the Homebrew module instead.

ilovezfs:
Should that still be using bump_args? https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb#L292

PR #4161, opened 2018-05-12
Fix brew pull with git commit.gpgsign enabled

alyssais:

-----

Because of environment filtering, git tries to sign the commit but can't find gpg, which means that the patch won't be applied.

### DIFF
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -103,14 +103,16 @@ module Homebrew
     end
 
     # Depending on user configuration, git may try to invoke gpg.
-    begin
-      gnupg = Formula["gnupg"]
-    rescue FormulaUnavailableError # rubocop:disable Lint/HandleExceptions
-    else
-      if gnupg.installed?
-        path = PATH.new(ENV.fetch("PATH"))
-        path.prepend(gnupg.installed_prefix/"bin")
-        ENV["PATH"] = path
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -102,6 +102,18 @@ def pull
       ENV["GIT_COMMITTER_EMAIL"] = ENV["HOMEBREW_GIT_EMAIL"]
     end
 
+    # Depending on user configuration, git may try to invoke gpg.
```
MikeMcQuaid: Is there any way to identify this in advance rather than doing so unconditionally?

DomT4:
```ruby
Utils.popen_read("git config --get commit.gpgsign").chomp
```
Maybe?

MikeMcQuaid: @alyssais any thoughts?

alyssais: What @DomT4 says looks good, though I'm not sure there's any harm in doing it unconditionally. I'll try to push that change tonight if we want it.

MikeMcQuaid: Yeh, that would be good, thanks 👍

MikeMcQuaid:
@alyssais Thanks again!

PR #4170, opened 2018-05-15
audit: new formulae should not have options

commitay:

-----

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -408,6 +408,13 @@ module Homebrew
           if dep.tags.include?(:run)
             problem "Dependency '#{dep.name}' is marked as :run. Remove :run; it is a no-op."
           end
+
+          next unless @new_formula
+          next if formula.versioned_formula?
+          next unless @official_tap
+          if dep.tags.include?(:recommended) || dep.tags.include?(:optional)
+            new_formula_problem "Formulae should not have #{dep.tags} dependencies."
+          end
         end
       end
     end
--- a/Library/Homebrew/rubocops/options_cop.rb
+++ b/Library/Homebrew/rubocops/options_cop.rb
@@ -46,11 +46,14 @@ module RuboCop
 
     module NewFormulaAudit
       class Options < FormulaCop
-        MSG = "New Formula should not use `deprecated_option`".freeze
+        DEP_OPTION = "New Formula should not use `deprecated_option`".freeze
+        OPTION = "Formula should not have an `option`".freeze
 
         def audit_formula(_node, _class_node, _parent_class_node, body_node)
           return if versioned_formula?
-          problem MSG if method_called_ever?(body_node, :deprecated_option)
+          problem DEP_OPTION if method_called_ever?(body_node, :deprecated_option)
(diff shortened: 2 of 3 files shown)


commitay:
Unrelated to this PR: do we want have audit check for option deps in new formulae?

MikeMcQuaid:
This makes sense to me but I'd want @ilovezfs to 👍. Thanks @commitay!

> Unrelated to this PR: do we want to have audit check for optional deps in new formulae?

I think it kinda is related 😁 as `:optional` and `:recommended` end up being options.

Also, ideally this check would only apply to Homebrew/homebrew-core formulae.

commitay:
> ideally this check would only apply to Homebrew/homebrew-core formulae

Updated.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -381,6 +381,11 @@ def audit_deps
           if dep.tags.include?(:run)
             problem "Dependency '#{dep.name}' is marked as :run. Remove :run; it is a no-op."
           end
+
+          next unless @new_formula && @official_tap
```
MikeMcQuaid: Can you make this an `if &&`; `unless &&` is quite hard to parse.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -382,7 +382,9 @@ def audit_deps
             problem "Dependency '#{dep.name}' is marked as :run. Remove :run; it is a no-op."
           end
 
-          next unless @new_formula && @official_tap
+          next unless @new_formula
+          next if formula.versioned_formula?
```
commitay: Added this to match the `options_cop` check so it doesn't fail for deps and not options.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -532,7 +539,7 @@ def audit_specs
       end
 
       if @new_formula && formula.head
-        problem "New formulae should not have a HEAD spec"
```
MikeMcQuaid: @ilovezfs Is this what you're wanting? I wasn't sure.

MikeMcQuaid:
Just FYI @commitay https://github.com/Homebrew/brew/pull/3835 is the approach we're moving towards for new formulae-only audits so they will post a comment on the PR but not fail CI. I could do with some help testing that and getting it merged if you're interested. We can chat more about it on Slack if so (as it'll involved internal project security stuff).

Library/Homebrew/dev-cmd/audit.rb
ilovezfs: Yes. The original wording was spelling out exactly how to cheat the audit. Open a new formula, get it merged. Then immediately open a second PR to add head.

commitay:
> I could do with some help testing that and getting it merged if you're interested.

Definitely interested. 👍

commitay:
@MikeMcQuaid Do you want this merged before https://github.com/Homebrew/brew/pull/3835?

MikeMcQuaid:
I'd probably rather #3835 before this.

commitay:
Rebased on master and updated to use `new_formula_problem`

PR #4174, opened 2018-05-15
Add compatibility layer for handling Cask Tap migration.

reitermarkus:


Library/Homebrew/compat/tap.rb
```diff
@@ -0,0 +1,38 @@
+module CaskTapMigration
+  def initialize(user, repo)
+    super
+
+    return unless user == "caskroom"
```
MikeMcQuaid: Make this an if?

Library/Homebrew/compat/tap.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    # TODO: Remove this check after migration.
+    return unless repo == "tap-migration-test"
```
MikeMcQuaid: Make this an if?

Library/Homebrew/compat/tap.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    new_user = "Homebrew"
+    new_repo = (repo == "cask") ? repo : "cask-#{repo}"
```
MikeMcQuaid: If else?

Library/Homebrew/compat/tap.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    super(new_user, new_repo)
+
+    return unless old_path.git?
+
+    new_name = name
+    new_path = path
+    new_remote = default_remote
+
+    ohai "Migrating Tap #{old_name} to #{new_name} …"
```
MikeMcQuaid: Lowercase t on tap and use ... rather than Unicode thing

Library/Homebrew/extend/git_repository.rb
```diff
@@ -13,6 +13,13 @@ def git_origin
     end
   end
 
+  def git_origin=(origin)
+    return unless git? && Utils.git_available?
```
MikeMcQuaid: Make if rather than unless

Library/Homebrew/compat/tap.rb
```diff
@@ -0,0 +1,38 @@
...
+    path.dirname.mkpath
+    FileUtils.mv old_path, new_path
+
+    puts "Changing remote from #{old_remote} to #{new_remote} …"
+    new_path.git_origin = new_remote
+  end
+end
+
+class Tap
+  prepend CaskTapMigration
```
MikeMcQuaid: This isn’t how we generally do compat classes; might be worth a PR to update the others to this format if it’s more desirable?

Library/Homebrew/missing_formula.rb
```diff
@@ -101,7 +101,7 @@ def tap_migration_reason(name)
           EOS
           break if new_tap_name == CoreTap.instance.name
 
-          install_cmd = if new_tap_user == "caskroom"
+          install_cmd = if new_tap_user == "caskroom" || (new_tap_user == "homebrew" && new_tap_repo.start_with?("cask"))
```
MikeMcQuaid: Feels a bit much for one line

Library/Homebrew/tap.rb
```diff
@@ -26,7 +26,7 @@ def self.fetch(*args)
     end
 
     # We special case homebrew and linuxbrew so that users don't have to shift in a terminal.
-    user = user.capitalize if ["homebrew", "linuxbrew"].include? user
+    user = user.capitalize if ["homebrew", "linuxbrew"].include?(user)
```
MikeMcQuaid: Any reason for this change?

Library/Homebrew/compat/tap.rb
reitermarkus: How?

Library/Homebrew/compat/tap.rb
reitermarkus: How?

Library/Homebrew/missing_formula.rb
reitermarkus: For what it's worth, this is only temporary.

reitermarkus:
By the way, to test this, run:

```
brew tap caskroom/tap-migration-test --no-compat
brew untap caskroom/tap-migration-test
```

Library/Homebrew/compat/tap.rb
MikeMcQuaid: `return if repo != "tap-migration-test"`

Library/Homebrew/extend/git_repository.rb
MikeMcQuaid: @reitermarkus Could update them too; it's pretty yuck.

Library/Homebrew/missing_formula.rb
MikeMcQuaid: How temporary, roughly?

Library/Homebrew/compat/tap.rb
reitermarkus: I'm pretty sure RuboCop would complain about that.

Library/Homebrew/missing_formula.rb
reitermarkus: It can be removed after the migration.

Library/Homebrew/extend/git_repository.rb
reitermarkus: On another note, would we even get to a point where we call any of these when `Utils.git_available?` is `false`? I mean, how can there be a `.git` directory when `git` is not available?

Library/Homebrew/compat/tap.rb
MikeMcQuaid: @reitermarkus Try it; pretty sure it doesn't 😉

Library/Homebrew/missing_formula.rb
reitermarkus: Exactly.

JCount: I concur.

MikeMcQuaid:
Thanks @reitermarkus!

JCount:
🎊

PR #4178, opened 2018-05-16
Alphabetize the `brew outdated` formula listing.

dougal:
Sorts the output of `brew outdated` by `Formula#<=>`.

For me this significantly speeds up scanning the list to look for formula I consider critical to my work.

Includes updated test example.

> Have you written new tests for your changes?

Not strictly speaking. I modified an existing test.

-----

### DIFF
--- a/Library/Homebrew/cmd/outdated.rb
+++ b/Library/Homebrew/cmd/outdated.rb
@@ -41,7 +41,9 @@ module Homebrew
     verbose = ($stdout.tty? || ARGV.verbose?) && !ARGV.flag?("--quiet")
     fetch_head = ARGV.fetch_head?
 
-    outdated_formulae = formulae.select { |f| f.outdated?(fetch_head: fetch_head) }
+    outdated_formulae = formulae
+                        .select { |f| f.outdated?(fetch_head: fetch_head) }
+                        .sort
 
     outdated_formulae.each do |f|
       if verbose
--- a/Library/Homebrew/cmd/outdated.rb
+++ b/Library/Homebrew/cmd/outdated.rb
@@ -41,9 +41,8 @@ module Homebrew
     verbose = ($stdout.tty? || ARGV.verbose?) && !ARGV.flag?("--quiet")
     fetch_head = ARGV.fetch_head?
 
-    outdated_formulae = formulae
-                        .select { |f| f.outdated?(fetch_head: fetch_head) }
-                        .sort
+    outdated_formulae = formulae.select { |f| f.outdated?(fetch_head: fetch_head) }
+                                .sort
 
     outdated_formulae.each do |f|
       if verbose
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/outdated.rb
```diff
@@ -41,7 +41,9 @@ def print_outdated(formulae)
     verbose = ($stdout.tty? || ARGV.verbose?) && !ARGV.flag?("--quiet")
     fetch_head = ARGV.fetch_head?
 
-    outdated_formulae = formulae.select { |f| f.outdated?(fetch_head: fetch_head) }
+    outdated_formulae = formulae
+                        .select { |f| f.outdated?(fetch_head: fetch_head) }
+                        .sort
```
reitermarkus: Leave the first line as it was, and align `.sort` with `.select`.

Library/Homebrew/test/cmd/outdated_spec.rb
```diff
@@ -4,8 +4,11 @@
       setup_test_formula "testball"
       (HOMEBREW_CELLAR/"testball/0.0.1/foo").mkpath
 
+      setup_test_formula "foo"
+      (HOMEBREW_CELLAR/"foo/0.0.1/foo").mkpath
+
       expect { brew "outdated" }
-        .to output("testball\n").to_stdout
+        .to output("foo\ntestball\n").to_stdout
```
reitermarkus: Did you check if this specific case actually fails without `sort`?

Library/Homebrew/cmd/outdated.rb
dougal: Fixed in dougal/brew@c80842449

Library/Homebrew/test/cmd/outdated_spec.rb
dougal: Yes. Commenting out the `sort` line results in the following failure:

```
1) brew outdated quiet output prints outdated Formulae
     Failure/Error:
       expect { brew "outdated" }
         .to output("foo\ntestball\n").to_stdout
         .and not_to_output.to_stderr
         .and be_a_success
     
       expected block to output "foo\ntestball\n" to stdout, but output "testball\nfoo\n"
```

ilovezfs: I assume that's locally but not on CI.

dougal: Yes, locally.

CI should never fail as the tests were committed with the changes. I assume, unless I am misunderstanding something.

ilovezfs: I just mean that on CI, which is using HFS+, it probably always passes regardless of the sort, whereas locally it fails without the sort because you're using APFS. I'm not sure there's anything to be done about that really in terms of making the test actually test something on CI but I don't think we need to worry about `sort` not working.

dougal: Good insight on CI using HFS+, thanks.

I originally tried making the the test case more expansive, with different formulae inserted in different orders. I simplified it as I wasn't sure the expanded test case provided any value.

reitermarkus: > I don't think we need to worry about `sort` not working.

Well, the test is not for `sort` itself, but for the fact that the output has to be sorted. Anyways, as long as it fails locally on APFS, it is good enough.

reitermarkus:
Thank you for your first contribution to Homebrew, @dougal!

PR #4182, opened 2018-05-16
Move `--version` logic into `cmd/--version.rb`.

reitermarkus:

-----

`time brew --version` is the same before and after.

### DIFF
--- a/Library/Homebrew/cmd/--version.rb
+++ b/Library/Homebrew/cmd/--version.rb
@@ -5,10 +5,9 @@ module Homebrew
   module_function
 
   def __version
-    # As a special case, `--version` is implemented directly in `brew.rb`. This
-    # file merely serves as a container for the documentation. It also catches
-    # the case where running `brew --version` with additional arguments would
-    # produce a rather cryptic message about a non-existent `--version` command.
-    raise UsageError
+    odie "This command does not take arguments." if ARGV.any?
+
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
Makes sense, thanks @reitermarkus!

Library/Homebrew/cmd/--version.rb
```diff
@@ -5,10 +5,9 @@ module Homebrew
   module_function
 
   def __version
-    # As a special case, `--version` is implemented directly in `brew.rb`. This
-    # file merely serves as a container for the documentation. It also catches
-    # the case where running `brew --version` with additional arguments would
-    # produce a rather cryptic message about a non-existent `--version` command.
-    raise UsageError
+    raise UsageError, "This command does not take arguments." unless ARGV.empty?
```
MikeMcQuaid: Could maybe use the same form here as e.g. `brew pull` i.e. the new option parser stuff?

reitermarkus: I think this is better, since a `UsageError` shows the command's help along with the error message.

MikeMcQuaid: @reitermarkus I disagree; I'd rather see `Homebrew::CLI::Parser.parse` extended to output the command's help and have more consistent behaviour for now.

PR #4186, opened 2018-05-17
Clean up `compat`.

reitermarkus:

-----

Always use this style:

```ruby
module Compat
end

prepend Compat
```

### DIFF
--- a/Library/Homebrew/compat/ENV/shared.rb
+++ b/Library/Homebrew/compat/ENV/shared.rb
@@ -1,9 +1,13 @@
 module SharedEnvExtension
-  def j1
-    odisabled "ENV.j1", "ENV.deparallelize"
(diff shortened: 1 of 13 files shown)


MikeMcQuaid:
There's still e.g. `compat/formula.rb` using the "open the existing class to overwrite the method" style; you might want to read through all the files in `compat` for comparison. This style is also taken by `extend/os` too if it made sense to clean that up in the same fashion.

MikeMcQuaid:
> Well, prepend is only needed if the compatibility layer should be able to call the current implementation, i.e. super, but I guess we can use the same style anyways, just to be consistent.

👍 Gotcha. I'm fine either way. For a future PR this does apply a bunch to the `extend/os` layer so doing it there too would be 🎉

PR #4194, opened 2018-05-21
Update Homebrew-Cask references.

reitermarkus:

-----

Update URLs and refactor unnecessary checks.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/audit.rb
+++ b/Library/Homebrew/cask/lib/hbc/audit.rb
@@ -61,7 +61,7 @@ module Hbc
 
       tap = @cask.tap
       return if tap.nil?
-      return unless ["Homebrew", "caskroom"].include?(tap.user)
+      return if tap.user != "Homebrew"
 
       return unless cask.artifacts.any? { |k| k.is_a?(Hbc::Artifact::Pkg) && k.stanza_options.key?(:allow_untrusted) }
       add_warning "allow_untrusted is not permitted in official Homebrew-Cask taps"
@@ -133,7 +133,7 @@ module Hbc
         return unless previous_cask.version == cask.version
         return if previous_cask.sha256 == cask.sha256
 
-        add_error "only sha256 changed (see: https://github.com/caskroom/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md)"
+        add_error "only sha256 changed (see: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md)"
       rescue CaskError => e
         add_warning "Skipped version and checksum comparison. Reading previous version failed: #{e}"
       end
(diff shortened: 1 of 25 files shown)


Library/Homebrew/cask/lib/hbc/audit.rb
```diff
@@ -61,7 +61,7 @@ def check_untrusted_pkg
 
       tap = @cask.tap
       return if tap.nil?
-      return unless ["Homebrew", "caskroom"].include?(tap.user)
+      return unless tap.user == "Homebrew"
```
MikeMcQuaid: `return if tap.user != "Homebrew"`

Library/Homebrew/manpages/brew.1.md.erb
```diff
@@ -61,12 +61,6 @@ With `--verbose` or `-v`, many commands print extra debugging information. Note
 
   Homebrew/homebrew-bundle <https://github.com/Homebrew/homebrew-bundle>
 
-  * `cask` [`--version` | `audit` | `cat` | `cleanup` | `create` | `doctor` | `edit` | `fetch` | `home` | `info`]:
```
MikeMcQuaid: Please add this back.

reitermarkus: Added it to the `COMMANDS` section, since it's not an external command anymore.

MikeMcQuaid: It still acts very much as an external command even if it's inside Homebrew/brew and, like the other external commands, it immediately taps a tap to be functionality. I'd rather it was left as-is until the `brew-cask.1` manpage is entirely moved into `brew.1`.

MikeMcQuaid: > It's not like external commands, for external commands, you first have to tap a tap to use the command, for brew cask you don't, it just happens to install a tap as a dependency on first use

For the user this is functionally identical to `bundle` and `services` both of which happen to install a tap as a dependency on first use. 

If we get to a point where we have https://caskroom.github.io merged into https://brew.sh, `brew cask --help` resembling `brew install --help` and other similar style matching and `brew-cask.1` goes away I'll be 👍 on changing this. Until then: please put it back how it was until we're further along, thanks.

reitermarkus: > For the user this is functionally identical to `bundle` and `services` both of which happen to install a tap as a dependency on first use. 

I fail to see how the functionality of a command has anything to do with it being external. I can have an internal command printing “Hello, world!" or an external command printing "Hello, world!".

MikeMcQuaid: Please put it back to where it was and we can revisit it later (as I've stated multiple times above). This change is unrelated to the rest of the PR, your GSoC project and is not a good use of either of our time to go around in circles about it.

PR #4199, opened 2018-05-22
Doco: add a Shell Completion doc

apjanke:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). *No: it's a doco-only change; nothing to test.*

-----

How about a doco page explaining how to set up Homebrew-provided shell completions? Seems justified since this is an extra manual configuration step that users have to do after installation, and I don't see it anywhere else in the Homebrew documentation, and the Google results for it are not great.

This PR adds a "Homebrew Shell Completion" doco page and links to it from `docs/README.md`. This covers both the completions for `brew` itself, and completions supplied as part of installed formulae.

The example shell code in it I think is appropriate for just dropping in to an rc file that's not doing any other `brew`-related stuff. More complex rc files that do multiple `brew`-related things will want to just run `brew --prefix` once and store the result in a variable (because `brew --prefix` is kind of slow), but I don't know that we want to dive in to that here.


MikeMcQuaid:
Nice write-up, thanks @apjanke!

PR #4200, opened 2018-05-22
audit: Add condition check on new_formula_audits

GauthamGoli:

-----
https://github.com/Homebrew/brew/pull/3835#issuecomment-390945189

When there are just new formula audit errors, then those audit errors are not reported because of `next` statement inside the loop.


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -149,13 +149,12 @@ def audit
...
-      problem_count += new_formula_problem_lines.size
+      new_formula_problem_count += new_formula_problem_lines.size
       puts new_formula_problem_lines.map { |s| "  #{s}" }
     end
 
-    errors_summary = "#{Formatter.pluralize(problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
-    return if problem_count.zero?
-    ofail errors_summary unless strict
+    errors_summary = "#{Formatter.pluralize(problem_count + new_formula_problem_count, "problem")} in #{Formatter.pluralize(formula_count, "formula")}"
+    ofail errors_summary if problem_count > 0 || (new_formula_problem_count > 0 && !created_pr_comment)
```
MikeMcQuaid: Can the above two lines get split into several lines each; there's a lot going on here that's hard to keep track of as-is.

MikeMcQuaid:
`brew style` is failing: https://travis-ci.org/Homebrew/brew/jobs/382141707

Library/Homebrew/dev-cmd/audit.rb
GauthamGoli: Could not split above line, Rubocop wants a guard clause in the last if statement. 
The condition is  ( if regular audits exist  or ( new formula audits exist but PR comment is not made ) ) 
then exit with non zero status

PR #4204, opened 2018-05-24
audit: github notability

commitay:

-----
This changes the new formula notability audit to check if a threshold of forks+watchers+stars is reached.

This is intended to be a more general notability check rather than checking if the repo satisfies one specific criteria.

Edit: this also increases the notability threshold.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -515,9 +515,9 @@ module Homebrew
 
       new_formula_problem "GitHub fork (not canonical repository)" if metadata["fork"]
       if formula&.tap&.core_tap? &&
-         (metadata["forks_count"] < 20) && (metadata["subscribers_count"] < 20) &&
-         (metadata["stargazers_count"] < 50)
-        new_formula_problem "GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)"
+         (metadata["forks_count"] < 30) && (metadata["subscribers_count"] < 30) &&
+         (metadata["stargazers_count"] < 75)
+        new_formula_problem "GitHub repository not notable enough (<30 forks, <30 watchers and <75 stars)"
       end
 
       return if Date.parse(metadata["created_at"]) <= (Date.today - 30)


MikeMcQuaid:
> This is intended to be a more general notability check rather than checking if the repo satisfies one specific criteria.
> 
> Edit: this also increases the notability threshold.

Personally I'd rather see it stay as-is. How are the previous values causing problems?

commitay:
Changing to a sum of the counts:
Currently a repo with 25 forks, 15 watchers and 23 stars is notable while a repo with 15 forks, 15 watchers and 40 stars is not. Kind of an edge case as most of the time a formula ends up being considered notable based on the number of stars but I thought it might be better to consider all three counts at once rather than if one is satisfied. It might also help when refusing formulae as it would be refused for failing to meet the broader total rather than one specific type.

Increasing the notability threshold:
A lot of new formulae PRs pass the current audit but aren't really "notable", e.g forks+watchers+stars <100. The audit was last revised a bit over two years ago, I think the growth of GitHub in that time is something we need to consider and adjust to otherwise the threshold will be inadvertently lowered.

MikeMcQuaid:
> Currently a repo with 25 forks, 15 watchers and 23 stars is notable while a repo with 15 forks, 15 watchers and 40 stars is not. Kind of an edge case as most of the time a formula ends up being considered notable based on the number of stars but I thought it might be better to consider all three counts at once rather than if one is satisfied.

Can you link to any PRs/repos where this has been the case?

> A lot of new formulae PRs pass the current audit but aren't really "notable", e.g forks+watchers+stars <100. The audit was last revised a bit over two years ago, I think the growth of GitHub in that time is something we need to consider and adjust to otherwise the threshold will be inadvertently lowered.

I think we should be looking for reasons to not package things rather than reasons to turn them away.

commitay:
> repo with 25 forks, 15 watchers and 23 stars

Sorry, number of watchers was off by one. 

https://github.com/cisco/cjose
https://github.com/Homebrew/homebrew-core/pull/27946

MikeMcQuaid:
@commitay And just to confirm: do you think that PR should be merged or not?

ilovezfs:
>I think the growth of GitHub in that time is something we need to consider and adjust to otherwise the threshold will be inadvertently lowered.

I agree with this. There's continuous popularity grade inflation due to GitHub's growth, and so I think we should regularly revise these thresholds upwards even if the only goal were to tread water.

commitay:
> do you think that PR should be merged or not?

I think that PR doesn't really meet the intent of the current audit. It would be refused if it had 1 fewer fork and 20 more stars.

MikeMcQuaid:
> I think that PR doesn't really meet the intent of the current audit. It would be refused if it had 1 fewer fork and 20 more stars.

I think whatever numbers we end up with ultimately they will be completely arbitrary and we may want to not include things that meet the audit and include things that do not. It's more meant to be a reason for us to be able to diplomatically say no than a rule that we're enforcing. Ultimately it was mainly added to give red CI to people immediately submitting weekend projects that weren't of any use to anyone, really.

commitay:
> I think whatever numbers we end up with ultimately they will be completely arbitrary

I don't think there is a way to avoid this without complicating the review process (maintainers voting on new formulae or something similar). FWIW, I had a look through the GitHub API to see if there was anything else that could be used as a measure of notability and nothing jumped out.

> we may want to not include things that meet the audit and include things that do not. It's more meant to be a reason for us to be able to diplomatically say no than a rule that we're enforcing.

This is a subjective judgement that is difficult to make as a new maintainer unless there is readily apparent reasons to consider disregarding the audit. It seems to be pretty much consistently enforced as a rule.

> Ultimately it was mainly added to give red CI to people immediately submitting weekend projects that weren't of any use to anyone, really.

Yes, this is basically the intent of this PR as I think the bar for what is considered a weekend project/notable/useful needs to move with the popularity of GitHub itself.

MikeMcQuaid:
> I don't think there is a way to avoid this without complicating the review process (maintainers voting on new formulae or something similar).
> This is a subjective judgement that is difficult to make as a new maintainer unless there is readily apparent reasons to consider disregarding the audit.

That assumes that we 👍 all new formulae that pass this and 👎 all that don't. The review process _is_ complicated and we have a lot of uncoded and undocumented decisions that are made.

> It seems to be pretty much consistently enforced as a rule.

If this is the case I feel even more strongly we should not raise the thresholds or change the calculation.

> Yes, this is basically the intent of this PR as I think the bar for what is considered a weekend project/notable/useful needs to move with the popularity of GitHub itself.

I don't agree. The bar is meant to be low enough to avoid people trivially gaming the system (asking a few friends to star the repo, for example) and it seems to have done a good job at that. To make an argument that the popularity of GitHub means the threshold needs raised I'd like to see the current arbitrary numbers replaced with something non-arbritrary i.e. calculate the rate of change of stars/watchers/forks and the overalls %s of public repos from a dataset like the GH Archive: https://www.gharchive.org

ilovezfs:
👍 to something non-arbitrary. The perception that just about everything is passing these days isn't being invented out of thin air.

MikeMcQuaid:
To pose a larger question that underpins this: what software should we package and what software should we not package (and why)?

MikeMcQuaid:
My starter for 10 is that we should package all software that we can do without patching, a massively complicated formula (relative to the complexity of the software), maintain a high quality on all platforms we support, that will be used by anyone and we don't expect to break in the very near future. Similarly, we should only remove software that we are unable to package well and/or breaks on our supported platforms (and we or others don't have the energy to fix it).

ilovezfs:
We should package software that is commonly available in other package managers such as Debian, likely to be maintained by users beyond just the Homebrew/homebrew-core maintainers, and of interest to a substantial number of users. We shouldn't package esoteric software that could be maintained in a tap.

I'd be fine with keeping the current

```
(metadata["forks_count"] < x) &&
(metadata["subscribers_count"] < y) &&
(metadata["stargazers_count"] < z)
```

and revise those up from 20, 20, 50 to like 30, 30, 75, and see if the sense that the standard has deteriorated resolves itself.

MikeMcQuaid:
> 👍 to something non-arbitrary.

but

> and revise those up from 20, 20, 50 to like 30, 30, 75, and see if the sense that the standard has deteriorated resolves itself.

🤔

ilovezfs:
In other words, if someone wants to propose a less arbitrary approach in the medium term, that sounds like it could be a worthwhile endeavor, but I don't think the possibility that someone might work on that should block us from improving the thresholds in the current approach now.

MikeMcQuaid:
> In other words, if someone wants to propose a less arbitrary approach in the medium term, that sounds like it could be a worthwhile endeavor, but I don't think the possibility that someone might work on that should block us from improving the thresholds in the current approach now.

Ok.



> and revise those up from 20, 20, 50 to like 30, 30, 75, and see if the sense that the standard has deteriorated resolves itself.

This seems a reasonable compromise 👍

commitay:
PR updated.

PR #4209, opened 2018-05-25
Reset `repo_var` so it actually is re-computed.

reitermarkus:


Library/Homebrew/compat/tap.rb
```diff
@@ -17,6 +17,8 @@ def initialize(user, repo)
 
       super(new_user, new_repo)
 
+      @repo_var = nil
```
MikeMcQuaid: Feels odd to set this here when the instance variable isn't otherwise use. Consider adding a helper method to reset it in `Tap`.

PR #4219, opened 2018-05-26
Fix comparison with Version::NULL on RHS

alyssais:

-----

This was accounted for in `Version::NULL.<=>`, but not in `Version#<=>`, so these could only be compared if `Version::NULL` was the left hand side.

### DIFF
--- a/Library/Homebrew/version.rb
+++ b/Library/Homebrew/version.rb
@@ -387,6 +387,7 @@ class Version
     other = Version.new(other.to_s) if other.is_a? Integer
     return 1 if other.nil?
 
+    return 1 if other.respond_to?(:null?) && other.null?
     return unless other.is_a?(Version)
     return 0 if version == other.version
     return 1 if head? && !other.head?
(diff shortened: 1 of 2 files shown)


Library/Homebrew/version.rb
```diff
@@ -387,6 +387,7 @@ def <=>(other)
     other = Version.new(other.to_s) if other.is_a? Integer
     return 1 if other.nil?
 
+    return 1 if other.respond_to?(:null?) && other.null?
     return unless other.is_a?(Version)
```
alyssais: The check had to go above the check that `other` is a version because `Version::NULL`'s anonymous class does not inherit from `Version`. (The type check feels like it's probably a smell, but out of scope).

MikeMcQuaid:
When does this occur, out of interest?

alyssais:
Here, in Linuxbrew, but it definitely feels like an upstream bug: https://github.com/Linuxbrew/homebrew-core/blob/master/Formula/glibc.rb#L5

MikeMcQuaid:
Cool, great, thanks again @alyssais!

sjackman:
Thanks, Alyssa! ❤️

PR #4224, opened 2018-05-28
HBC: Make language stanza order not matter

vitorgalvao:

-----

Fixes https://github.com/Homebrew/homebrew-cask/issues/42755 by checking first for exact matches and only then moving to general matches.

I previously kept it as blocks, but that made it uglier/harder to read/edit/find mistakes.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/dsl.rb
+++ b/Library/Homebrew/cask/lib/hbc/dsl.rb
@@ -134,9 +134,7 @@ module Hbc
       end
 
       MacOS.languages.map(&Locale.method(:parse)).each do |locale|
-        key = @language_blocks.keys.detect do |strings|
-          strings.any? { |string| locale.include?(string) }
-        end
+        key = locale.detect(@language_blocks.keys)
 
         next if key.nil?
 
--- a/Library/Homebrew/locale.rb
+++ b/Library/Homebrew/locale.rb
@@ -68,6 +68,11 @@ class Locale
   end
   alias == eql?
 
+  def detect(locale_groups)
+    locale_groups.detect { |locales| locales.any? { |locale| eql?(locale) } } ||
+      locale_groups.detect { |locales| locales.any? { |locale| include?(locale) } }
+  end
+
   def to_s
     [@language, @region, @script].compact.join("-")
   end
--- a/Library/Homebrew/test/locale_spec.rb
+++ b/Library/Homebrew/test/locale_spec.rb
@@ -72,4 +72,14 @@ describe Locale do
       expect(subject.eql?("zh_CN_Hans")).to be false
     end
   end
+
+  describe "#detect" do
+    let(:locale_groups) { [["zh"], ["zh-TW"]] }
+
+    it "finds best matching language code, independent of order" do
+      expect(described_class.new("zh", "TW", nil).detect(locale_groups)).to eql(["zh-TW"])
+      expect(described_class.new("zh", "TW", nil).detect(locale_groups.reverse)).to eql(["zh-TW"])
+      expect(described_class.new("zh", "CN", "Hans").detect(locale_groups)).to eql(["zh"])
+    end
+  end
 end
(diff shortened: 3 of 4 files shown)


Library/Homebrew/cask/lib/hbc/dsl.rb
```diff
@@ -134,9 +134,8 @@ def language_eval
       end
 
       MacOS.languages.map(&Locale.method(:parse)).each do |locale|
-        key = @language_blocks.keys.detect do |strings|
-          strings.any? { |string| locale.include?(string) }
-        end
+        key = @language_blocks.keys.detect { |strings| strings.any? { |string| locale.eql?(string) } } ||
+              @language_blocks.keys.detect { |strings| strings.any? { |string| locale.include?(string) } }
```
reitermarkus: Maybe we should add a helper method to `Locale`, so we can add a simple unit test. Something like:

```ruby
key = locale.detect(*@language_blocks.keys)
```

vitorgalvao: Gave it a try, but I don’t yet understand the full workings of `Locale` to be able to do it.

I have no problem with closing this over a better solution, though.

reitermarkus: Basically just move this logic into a method in `Locale`.

vitorgalvao: That’s what I tried and failed to do. I understand the logic of what I need to do, but the code isn’t coming out right. I can’t return the correct thing and it always goes for the default language.

reitermarkus: Could you maybe push the code anyways, even if it isn't working?

vitorgalvao: Nothing like a bit of sleep to make us understand what we were doing wrong. Got it to work though I haven’t written the test for the helper method.

Library/Homebrew/locale.rb
```diff
@@ -68,6 +68,11 @@ def eql?(other)
   end
   alias == eql?
 
+  def detect(others)
+    others.detect { |other| eql?(*other) } ||
```
reitermarkus: No need for the `*` here.

Library/Homebrew/locale.rb
```diff
@@ -68,6 +68,11 @@ def eql?(other)
   end
   alias == eql?
 
+  def detect(others)
```
reitermarkus: Use `*others` here and `locale.detect(*@language_blocks.keys)` above.

vitorgalvao: That was my first try, but I got `Error: '["en"]' cannot be parsed to a Locale`.

Library/Homebrew/locale.rb
vitorgalvao: https://github.com/Homebrew/brew/pull/4224#discussion_r191364518

Library/Homebrew/locale.rb
reitermarkus: Ah right, the keys are actually arrays.

Library/Homebrew/locale.rb
reitermarkus: `eql?` and `include?` are not implemented for multiple arguments. It should be `other.any? { |o| eql?(o) }`.

reitermarkus: Maybe rename `others` to `locale_groups` and `other` to `locales`.

Library/Homebrew/locale.rb
```diff
@@ -68,6 +68,11 @@ def eql?(other)
   end
   alias == eql?
 
+  def detect(locale_groups)
```
reitermarkus: Last thing would be a test in `locale_spec.rb` for the `zh-TW`/`zh` example mentioned in https://github.com/Homebrew/homebrew-cask/issues/42755.

vitorgalvao: Done.

Issue #4059, opened 2018-04-12
Remove all `require "cmd/*"`


MikeMcQuaid:
No `require "cmd/*"` should be needed in the codebase. Where one part of Homebrew needs to access code in `cmd/*` then that's a sign that code should be moved into a separate file and, if needed, method.

aamarill:
Hello,

I am looking to make my first open source contribution. Is this a good issue to take on?
If not, feel free to point me other issues that might be good for someone that's new to open source.
Thanks!

PR #4253, opened 2018-06-01
Refactor `search`.

reitermarkus:

-----

Refactor `search` to use `CLI::Parser`.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
@@ -9,7 +9,7 @@ module Hbc
       option "--debug",         :debug,         false
       option "--verbose",       :verbose,       false
       option "--outdated",      :outdated_only, false
-      option "--require-sha",   :require_sha, false
+      option "--require-sha",   :require_sha,   false
 
       def self.command_name
         @command_name ||= name.sub(/^.*:/, "").gsub(/(.)([A-Z])/, '\1_\2').downcase
@@ -49,24 +49,18 @@ module Hbc
         casks = args.empty? ? alternative.call : args
         @casks = casks.map { |cask| CaskLoader.load(cask) }
       rescue CaskUnavailableError => e
-        reason = [e.reason, suggestion_message(e.token)].join(" ")
+        reason = [e.reason, *suggestion_message(e.token)].join(" ")
         raise e.class.new(e.token, reason)
       end
 
       def suggestion_message(cask_token)
-        exact_match, partial_matches = Search.search(cask_token)
+        matches, = Search.search(cask_token)
 
-        if exact_match.nil? && partial_matches.count == 1
-          exact_match = partial_matches.first
-        end
-
-        if exact_match
-          "Did you mean “#{exact_match}”?"
-        elsif !partial_matches.empty?
+        if matches.one?
+          "Did you mean “#{matches.first}”?"
+        elsif !matches.empty?
           "Did you mean one of these?\n"
-            .concat(Formatter.columns(partial_matches.take(20)))
-        else
-          ""
+            .concat(Formatter.columns(matches.take(20)))
         end
       end
     end
--- a/Library/Homebrew/cask/lib/hbc/cli/search.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/search.rb
@@ -1,3 +1,5 @@
+require "cmd/search"
+
 module Hbc
   class CLI
     class Search < AbstractCommand
@@ -18,26 +20,6 @@ module Hbc
         end
       end
 
-      def self.search_remote(query)
-        matches = begin
-          GitHub.search_code(
-            user: "Homebrew",
-            path: "Casks",
-            filename: query,
-            extension: "rb",
-          )
-        rescue GitHub::Error => error
-          opoo "Error searching on GitHub: #{error}\n"
-          []
-        end
-
-        matches.map do |match|
-          tap = Tap.fetch(match["repository"]["full_name"])
-          next if tap.installed?
-          "#{tap.name}/#{File.basename(match["path"], ".rb")}"
-        end.compact
-      end
-
       def self.search(*arguments)
         exact_match = nil
         partial_matches = []
@@ -55,7 +37,7 @@ module Hbc
           partial_matches.delete(exact_match)
         end
 
-        remote_matches = search_remote(search_term)
+        _, remote_matches = Homebrew.search_taps(search_term, silent: true)
 
         [exact_match, partial_matches, remote_matches, search_term]
       end
--- a/Library/Homebrew/cask/lib/hbc/cli/search.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/search.rb
@@ -36,7 +36,7 @@ module Hbc
           partial_matches = simplified_tokens.grep(/#{simplified_search_term}/i) { |t| all_tokens[simplified_tokens.index(t)] }
         end
 
-        _, remote_matches = search_taps(search_term, silent: true)
+        remote_matches = search_taps(search_term, silent: true)[:casks]
 
         [partial_matches, remote_matches, search_term]
       end
--- a/Library/Homebrew/cask/lib/hbc/cli/search.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/search.rb
@@ -1,8 +1,10 @@
-require "cmd/search"
+require "search"
 
 module Hbc
   class CLI
     class Search < AbstractCommand
+      extend Homebrew::Search
+
       def run
         if args.empty?
           puts Formatter.columns(CLI.nice_listing(Cask.map(&:qualified_token)))
@@ -34,7 +36,7 @@ module Hbc
           partial_matches = simplified_tokens.grep(/#{simplified_search_term}/i) { |t| all_tokens[simplified_tokens.index(t)] }
         end
 
-        _, remote_matches = Homebrew.search_taps(search_term, silent: true)
+        _, remote_matches = search_taps(search_term, silent: true)
 
         [partial_matches, remote_matches, search_term]
       end
--- a/Library/Homebrew/cask/lib/hbc/cli/search.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/search.rb
@@ -21,7 +21,6 @@ module Hbc
       end
 
       def self.search(*arguments)
-        exact_match = nil
         partial_matches = []
         search_term = arguments.join(" ")
         search_regexp = extract_regexp arguments.first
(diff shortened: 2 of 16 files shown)


Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
```diff
@@ -49,24 +49,18 @@ def casks(alternative: -> { [] })
...
-        if exact_match.nil? && partial_matches.count == 1
-          exact_match = partial_matches.first
-        end
-
-        if exact_match
-          "Did you mean “#{exact_match}”?"
-        elsif !partial_matches.empty?
+        if matches.one?
+          "Did you mean “#{matches.first}”?"
+        elsif matches.any?
```
MikeMcQuaid: Is `.any?` and `! .empty?` identical performance wise? I was under the impression the latter was more performant (although that could be a Railism, creeping in)

Library/Homebrew/cask/lib/hbc/cli/search.rb
```diff
@@ -1,3 +1,5 @@
+require "cmd/search"
```
MikeMcQuaid: Trying to avoid adding more of these: https://github.com/Homebrew/brew/issues/4059

Library/Homebrew/cli_parser.rb
```diff
@@ -60,17 +60,22 @@ def conflicts(*options)
         @conflicts << options.map { |option| option_to_name(option) }
       end
 
-      def option_to_name(name)
-        name.sub(/\A--?/, "").tr("-", "_").delete("=")
+      def option_to_name(option)
+        option.sub(/\A--?/, "").tr("-", "_").delete("=")
```
MikeMcQuaid: Could stick each `.` on a new line if refactoring here anyway?

Library/Homebrew/cli_parser.rb
```diff
@@ -60,17 +60,22 @@ def conflicts(*options)
...
+      end
+
+      def name_to_option(name)
+        (name.length == 1) ? "-#{name}" : "--#{name}"
```
MikeMcQuaid: `if`/`else` is a bit more readable here.

Library/Homebrew/cli_parser.rb
```diff
@@ -126,7 +131,7 @@ def check_conflicts
           violations = mutually_exclusive_options_group.select do |option|
             option_passed? option
           end
-          raise OptionConflictError, violations if violations.length > 1
+          raise OptionConflictError, violations.map(&method(:name_to_option)) if violations.length > 1
```
MikeMcQuaid: Make a block `if` for a shorter line

Library/Homebrew/cli_parser.rb
```diff
@@ -163,9 +168,9 @@ def initialize(arg1, arg2, missing: false)
 
     class OptionConflictError < RuntimeError
       def initialize(args)
-        args_list = args.join("` and `")
+        args_list = args.map(&Formatter.public_method(:option)).join(" and ")
```
MikeMcQuaid: stick the `.join` on a new line?

Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
reitermarkus: [Seems like it](https://mensfeld.pl/2015/06/empty-vs-blank-vs-any-why-you-should-not-use-any-to-test-if-theres-anything-in-the-array/), I didn't know about that.

Library/Homebrew/cmd/search.rb
```diff
@@ -92,39 +106,44 @@ def search
...
+        [formulae, casks]
+      elsif match["path"].start_with?("Casks/")
+        [formulae, [*casks, full_name].sort]
+      else
+        [[*formulae, full_name].sort, casks]
+      end
+    end
+  rescue GitHub::Error => error
+    opoo "Error searching on GitHub: #{error}\n"
+    [[], []]
```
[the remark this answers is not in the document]
reitermarkus: A Hash cannot be destructed as nicely. Eventually this should be used like this:

```ruby
remote_formulae, remote_casks = search_taps(query)
```

Library/Homebrew/cmd/search.rb
```diff
@@ -92,39 +106,44 @@ def search
...
+      /.*#{Regexp.escape(query)}.*/i
     end
   rescue RegexpError
-    odie "#{query} is not a valid regex"
+    raise "#{query} is not a valid regex."
   end
 
   def search_taps(query, silent: false)
-    return [] if ENV["HOMEBREW_NO_GITHUB_API"]
+    return [], [] if ENV["HOMEBREW_NO_GITHUB_API"]
```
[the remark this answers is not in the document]
reitermarkus: https://github.com/Homebrew/brew/pull/4253#discussion_r192566745

Library/Homebrew/cli_parser.rb
```diff
@@ -60,17 +60,28 @@ def conflicts(*options)
...
+      def name_to_option(name)
+        if name.length == 1
+          "-#{name}"
+        else
+          "--#{name}"
+        end
       end
 
       def option_to_description(*names)
         names.map { |name| name.to_s.sub(/\A--?/, "").tr("-", " ") }.max
```
JCount: I know this code isn't directly part of this refactor, but may it make sense to make it more closely match the above syntax? Any thoughts @MikeMcQuaid?

reitermarkus: @GauthamGoli, do we even need this? We are using `OptionParser` only to parse arguments and not to display descriptions.

Library/Homebrew/cmd/search.rb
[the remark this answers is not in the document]
MikeMcQuaid: @reitermarkus I think given they are arrays a `{formulae: [], casks: []}` is a nicer interface for the consumer. It's not obvious from the naming that `search_taps` provides two arrays of arrays as results.

Library/Homebrew/cli_parser.rb
MikeMcQuaid: @GauthamGoli Given we're not using descriptions anywhere yet I'd be 👍 in deleting the description code until it's used (or using it in some places now)

JCount: 👍 For removal if it's not being utilized.

MikeMcQuaid: It seems from https://github.com/Homebrew/brew/pull/4271#issuecomment-394178063 it will be used very shortly so let's leave it as-is for now.

GauthamGoli: Sorry. I missed this notification. Yeah. I'll be working on this shortly, don't remove it.

Library/Homebrew/descriptions.rb
```diff
@@ -1,7 +1,11 @@
...
 require "formula_versions"
+require "search"
+require "searchable"
 
 class Descriptions
+  extend Homebrew::Search
+
   CACHE_FILE = HOMEBREW_CACHE + "desc_cache.json"
 
   def self.cache
```
MikeMcQuaid: Feel free to punt on this in this PR but I'd love to see this use the `CacheStore` class at some point rather than the weird JSONn cache it current uses.

Library/Homebrew/search.rb
```diff
@@ -0,0 +1,78 @@
...
+        Regexp.new(m[1])
+      else
+        query
+      end
+    rescue RegexpError
+      raise "#{query} is not a valid regex."
+    end
+
+    def search_taps(query, silent: false)
+      return [], [] if ENV["HOMEBREW_NO_GITHUB_API"]
```
MikeMcQuaid: Would still like to see this not returning two arrays.

Library/Homebrew/test/cask/cli/search_spec.rb
```diff
@@ -3,6 +3,7 @@
 describe Hbc::CLI::Search, :cask do
   before do
     allow(Tty).to receive(:width).and_return(0)
+    ENV.delete("HOMEBREW_NO_GITHUB_API")
```
MikeMcQuaid: Stick this in `dev-cmd/tests.rb` itself I think.

Library/Homebrew/descriptions.rb
reitermarkus: Yeah, I think this PR is big enough already.

Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -44,10 +44,10 @@ def tests
       ENV["HOMEBREW_NO_COMPAT"] = "1" if args.no_compat?
       ENV["HOMEBREW_TEST_GENERIC_OS"] = "1" if args.generic?
 
+      ENV.delete("HOMEBREW_NO_GITHUB_API")
```
MikeMcQuaid: Pedantry but: put this with the other `ENV.delete`s?

MikeMcQuaid: Also this should still be set when `args.online?` is false.

Library/Homebrew/search.rb
```diff
@@ -13,7 +13,7 @@ def query_regexp(query)
     end
 
     def search_taps(query, silent: false)
-      return [], [] if ENV["HOMEBREW_NO_GITHUB_API"]
+      return { formulae: [], casks: [] } if ENV["HOMEBREW_NO_GITHUB_API"]
```
MikeMcQuaid: I think it would read a little cleaner to create this hash first thing in the function and return that. It then means `each` can be used below instead of `each_with_object`

Library/Homebrew/dev-cmd/tests.rb
reitermarkus: Then I have to move into the spec again, because if it is set, stubbing `GitHub` does not work because it is skipped before it even gets to the point where it calls `search_code`.

Library/Homebrew/search.rb
reitermarkus: I don't know, in the `rescue` block `{ formulae: [], casks: [] }` feels cleaner because it is more explicit than reusing a variable that is initialized at the very beginning.

MikeMcQuaid: It feels cleaner (and more DRY) to me to only have to define the `formulae` and `casks` keys in a single location.

Library/Homebrew/dev-cmd/tests.rb
reitermarkus: Seems like it is used for things outside `GitHub` as well. Should be a separate PR in any case.

Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -39,15 +39,14 @@ def tests
...
       ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = "1"
       ENV["HOMEBREW_DEVELOPER"] = "1"
       ENV["HOMEBREW_NO_COMPAT"] = "1" if args.no_compat?
       ENV["HOMEBREW_TEST_GENERIC_OS"] = "1" if args.generic?
 
       if args.online?
         ENV["HOMEBREW_TEST_ONLINE"] = "1"
-      else
-        ENV["HOMEBREW_NO_GITHUB_API"] = "1"
       end
```
JCount: Change this to one line?

MikeMcQuaid:
Thanks again @reitermarkus! 🚢 when ready.

JCount:
👍 Good work, @reitermarkus

PR #4260, opened 2018-06-01
superenv: Move macOS-specific files to shims/mac/

sjackman:
There are three superenv directories: generic, macOS, and Linux.
```
shims/super/
shims/mac/super/
shims/linux/super/
```


MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #4265, opened 2018-06-01
Fix PKG installer environment

claui:
Last year, the `dymo-label` cask [was deleted](https://github.com/Homebrew/homebrew-cask/pull/36884) because it [would no longer install successfully](https://github.com/Homebrew/homebrew-cask/issues/36883).

I have now looked into the problem, and found something that I deem a small issue with the Cask core code.

## tl;dr

When installing a pkg, we don’t set e. g. the `USER` variable quite like the graphical installer does. This causes certain packages to fail with `brew cask install`, despite working fine with the graphical installer.

## Analysis

Looking at the failing **upstream post-install script,** I find that it attempts to launch an agent in the domain of the currently logged-in user, e. g. `gui/501`. To that end, the script queries the `USER` environment variable.

This is the relevant part from the post-install script:

```
# From component-10.pkg/Scripts/postinstall
[…]
USER_ID=`id -u ${USER}`

launchctl bootstrap gui/$USER_ID /Library/LaunchAgents/com.dymo.dls.webservice.plist

[…]
```

This works fine in the graphical installer. However, when I try to install it via command line, the installer always aborts with an error.

After a little debug logging, the installer log reveals that the graphical PKG installer exports slightly **different values** for some environment variables compared to `brew cask install`:

- **The graphical PKG installer** exports the name of the actual user, e. g. `USER=alice`. Thus, the post-install script will inherit `USER=alice`, and run `launchctl bootstrap gui/501`, and everything works fine.

- **When installing a cask** though, we currently call `sudo -E -- /usr/sbin/installer …`, which by default sets `USER=root`. This is not what the post-install script expects. With `USER_ID` set to `0`, the script then attempts to use the non-existing domain `gui/0`. Subsequently, `launchctl` refuses to launch the (non-essential) `com.dymo.dls.webservice`. It also returns with a non-zero exit status, which ultimately causes the entire installation to fail.

## Root cause

Ultimately, IMHO Apple is responsible for this inconsistency; they advertise `/usr/sbin/installer` as an alternative to the PKG installer, yet they fail to give a caveat or an explanation as to what the graphical installer does differently and why. Only partially can we blame upstream for the assumptions they have made.

Given that Apple is obviously not going to do anything, I think we need to fix this on our side.

## The fix

This PR proposes to fix the issue by setting `LOGNAME`, `USER`, and `USERNAME` to the same values as the graphical PKG installer does.

The solution looks like this:

```
/usr/bin/sudo -E LOGNAME=alice USER=alice USERNAME=alice -- /usr/sbin/installer …
```

Note that for some reason, it is not sufficient to simply export the environment to `sudo`. The variables need to be explicitly assigned in `sudo`’s command line, despite the `-E` switch being present.

## Proof

To confirm that the fix works, run:

```
brew cask install https://raw.githubusercontent.com/Homebrew/homebrew-cask/7ab32a0be8a16a2ac71ef6e802f9c045e30ca216/Casks/dymo-label.rb
```

This will result in a successful installation with this PR. Otherwise, it will fail as described in #36883.

## Other solutions

One possible workaround I’ve attempted without touching the core involves patching the post-install script to pin the current user. I feel this is a rather **awkward workaround,** difficult to understand and maintain:

```
# Shitty workaround
cask 'dymo-label' do
  […]

  # Re-packaged version after preflight
  pkg 'DYMO Label v.8.fixed.pkg'

  # Expand, patch, and re-package
  preflight do
    system_command '/usr/sbin/pkgutil',
                   args: [
                           '--expand',
                           staged_path / 'DYMO Label v.8.pkg',
                           staged_path / 'pkg',
                         ]
    system_command '/usr/bin/sed',
                   args: [
                           '-i', '',
                           '-e', "s/\\${USER}/#{ENV["USER"]}/",
                           '-e', "$a\\\n\\\nexit 0\n",
                           staged_path / 'pkg/component-10.pkg/Scripts/postinstall'
                         ]
    system_command '/usr/sbin/pkgutil',
                   args: [
                           '--flatten',
                           staged_path / 'pkg',
                           staged_path / 'DYMO Label v.8.fixed.pkg',
                         ]
  end

  […]
end
```

Another drawback of this workaround is that it wouldn’t prevent similar errors with other casks in the future.

## Review

Despite very careful testing, I suggest that we take our time to review this thoroughly. Thanks in advance.


commitay:
I haven't checked yet but this should also fix the launchctl issues we hack around in [`sekey`](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/sekey.rb) and [`softu2f`](https://github.com/Homebrew/homebrew-cask/blob/master/Casks/softu2f.rb).

The alternative workaround of extracting and editing the pkg is something I've looked at before for another cask. One problem with it is we would be removing the signature information from the pkg, which would allow it to be installed when it has an expired/untrusted signature.

Library/Homebrew/cask/lib/hbc/system_command.rb
```diff
@@ -56,14 +57,29 @@ def command
 
     private
 
-    attr_reader :executable, :args, :input, :options, :processed_output, :processed_status, :path
+    attr_reader :executable, :args, :input, :options, :processed_output, :processed_status, :env
 
     attr_predicate :sudo?, :print_stdout?, :print_stderr?, :must_succeed?
 
+    def uppercase_env
```
reitermarkus: I think this is one abstraction layer too much. I'd just use the actual variable name as a string instead of a symbol. Also makes it easier to grep.

Library/Homebrew/cask/lib/hbc/system_command.rb
```diff
@@ -47,7 +47,8 @@ def initialize(executable, args: [], sudo: false, input: [], print_stdout: false
       @must_succeed = must_succeed
       options.extend(HashValidator).assert_valid_keys(:chdir)
       @options = options
-      @path = path
+      @env = { path: ENV["PATH"] }
+      yield(@env) if block_given?
```
reitermarkus: Instead of yielding this I would make this another named parameter and merge it with `{ path: ENV["PATH"] }`.

Library/Homebrew/cask/lib/hbc/system_command.rb
claui: Fair point. Done!

Library/Homebrew/cask/lib/hbc/system_command.rb
claui: Thanks for the feedback. Changed.

Library/Homebrew/cask/lib/hbc/system_command.rb
```diff
@@ -82,10 +93,11 @@ def expanded_command
     end
 
     def each_output_line(&b)
+      popen3_env = sudo? ? env : {}
       executable, *args = expanded_command
 
       raw_stdin, raw_stdout, raw_stderr, raw_wait_thr =
-        Open3.popen3({ "PATH" => path }, [executable, executable], *args, **options)
+        Open3.popen3(popen3_env, [executable, executable], *args, **options)
```
reitermarkus: Does it hurt to always pass `env` regardless of `sudo`?

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -1,4 +1,20 @@
 describe Hbc::SystemCommand, :cask do
+  describe "#initialize" do
+    describe "given some environment variables" do
+      subject do
+        described_class.new("/usr/bin/true",
+          env: { "A" => "1", "B" => "2", "C" => "3" })
+      end
+
+      it { is_expected.to be_an_instance_of described_class }
```
reitermarkus: Can you think of a less trivial expectation?

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -1,4 +1,20 @@
 describe Hbc::SystemCommand, :cask do
+  describe "#initialize" do
+    describe "given some environment variables" do
+      subject do
```
reitermarkus: Use curly braces for `subject`.

Library/Homebrew/cask/lib/hbc/system_command.rb
```diff
@@ -56,14 +56,25 @@ def command
...
 
     def sudo_prefix
       return [] unless sudo?
       askpass_flags = ENV.key?("SUDO_ASKPASS") ? ["-A"] : []
-      ["/usr/bin/sudo", *askpass_flags, "-E", "--"]
+      prefix = ["/usr/bin/sudo", *askpass_flags, "-E"]
+
+      env.each do |name, value|
+        name.scan(/^$|^\d|\W/) do
+          ofail "Invalid variable name: '#{name}'"
```
reitermarkus: Maybe just `raise` instead of `ofail` here. Also not sure if this level of error checking is necessary at all, since this will probably be used in very specific circumstances.

Library/Homebrew/cask/lib/hbc/system_command.rb
claui: I’m torn on this one. Passing `env` unconditionally probably won’t hurt; that said, I feel it would be confusing from a maintenance point of view because 1. it has no actual effect (at least with the three variables we’re introducing here); more importantly, 2. in the `sudo` case, it may obscure the presence of the real assignments, causing confusion; finally, 3. at some point, the redundant assignment may cause a future maintainer to mistakenly remove the important part. (The latter is not really unlikely given that admittedly, I haven’t made it obvious from the code why the special treatment is there after all.)

I’ll gladly change it but I’d have to add inline comments in order to mitigate 2. and 3.  
What do you think?

Library/Homebrew/test/cask/system_command_spec.rb
claui: Good point; I planned to do so initially. I ultimately fell back to the trivial expectation when I realized I had made the `env` accessor private (like all the other accessors are).

I wouldn’t recommend to make `env` public just for the sake of testing.
How about I simply remove this particular test?

Library/Homebrew/test/cask/system_command_spec.rb
claui: Will do once we have addressed the trivial expectation thing.

Library/Homebrew/cask/lib/hbc/system_command.rb
claui: I’ve put the checks there specifically to guard against possible [confused deputy attacks](https://en.wikipedia.org/wiki/Confused_deputy_problem), where a malicious cask author may abuse `system_command` to gain root.

For example, let’s suppose some attacker has already managed to plant a malicious `/usr/local/bin/​` payload (note the last character being a `\u200b ZERO WIDTH SPACE`), and is now looking into gaining root. Without additional checks in place, the attacker may find a way to somehow put the `​` character (`\u200b`) inside the hash, and then submit a cask containing `system_command '/usr/bin/some_harmless_thing'` followed by `env` and `sudo: true`. This may then give them root, and maintainers may not notice in time until the damage is done.

I admit my example may be a poor one but it’s just to illustrate my point. There are likely other, more realistic attack vectors, of which we may not be aware right now.

**tl;dr** The checks make sense for security reasons because defense in depth, and because `sudo` is a beast.

I’m fine with `raise` instead of `ofail`, though.

Library/Homebrew/cask/lib/hbc/system_command.rb
```diff
@@ -47,7 +47,7 @@ def initialize(executable, args: [], sudo: false, input: [], print_stdout: false
       @must_succeed = must_succeed
       options.extend(HashValidator).assert_valid_keys(:chdir)
       @options = options
-      @path = path
+      @env = { "PATH" => ENV["PATH"] }.merge(env)
```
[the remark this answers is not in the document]
claui: That’s correct, and I’m in favor of that because that would pose a stable environment for the installer.

Library/Homebrew/test/cask/system_command_spec.rb
reitermarkus: You could call `/usr/bin/env` instead of `true` and check if they are in fact set.

Library/Homebrew/cask/lib/hbc/system_command.rb
reitermarkus: Maybe move this into `initialize` and raise an `ArgumentError`.

Library/Homebrew/cask/lib/hbc/system_command.rb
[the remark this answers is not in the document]
MikeMcQuaid: @claui Cool, cool, just checking 👍

Library/Homebrew/test/cask/system_command_spec.rb
MikeMcQuaid: @reitermarkus @claui Now we have a standard PATH I reckon it's always worth using `env` over `/usr/bin/env`

claui: @reitermarkus Glad you suggested that – it just revealed a major bug in my code!  
I wrote `sudo? ? env : {}`, while it obviously needs to be the other way around: `sudo? ? {} : env` 😊

claui:
@reitermarkus @MikeMcQuaid I have addressed all the requested changes. 😌

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -136,4 +197,11 @@
       }.to be_a_success
     end
   end
+
+  describe "given an invalid variable name" do
+    it "returns an ArgumentError" do
```
reitermarkus: “raises” instead of ”returns”

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -1,4 +1,65 @@
...
+          args: env_args,
+          env: { "A" => "1", "B" => "2", "C" => "3" },
+          must_succeed: true,
+          sudo: true,
+        )
+      }
+
+      describe "the resulting command line" do
+        it "includes the given variables explicitly" do
+          allow(Open3)
```
reitermarkus: I think this could be written more concisely with `expect(…).to receive`.

claui: Good point about `expect`; not sure regarding conciseness though.

What would be a more concise way to intercept the `popen3` call?

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -1,4 +1,65 @@
...
+          env: { "A" => "1", "B" => "2", "C" => "3" },
+          must_succeed: true,
+        )
+      }
+
+      its("run!.stdout") { is_expected.to eq("123") }
+
+      describe "the resulting command line" do
+        it "does not include the given variables" do
+          expect(Open3)
```
reitermarkus: I think this should work:

```ruby
expect(Open3).to receive(:popen3).with(a_hash_including("PATH"), ["env"] * 2, *env_args, {})
```

claui: This breaks for me, apparently because it causes the test double to return `nil` to the caller (instead of the expected `stdin, stdout, stderr, wait_thr` tuple.)

reitermarkus: Maybe append `.and_call_original`?

claui: My rationale behind `/usr/bin/true` was to avoid spawning a `bash` we don’t need.

But then again, it’s premature optimization, and maybe you’re right.

claui: Thanks again for your feedback @reitermarkus! I have addressed it with 071152b.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -50,7 +50,13 @@ def run_installer(command: nil, verbose: false, **_options)
         end
         with_choices_file do |choices_path|
           args << "-applyChoiceChangesXML" << choices_path if choices_path
-          command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
+          logged_in_user = ENV["USER"]
```
reitermarkus: `Utils.current_user`?

Library/Homebrew/cask/lib/hbc/system_command.rb
```diff
@@ -10,6 +10,8 @@ module Hbc
   class SystemCommand
     extend Predicable
 
+    VALID_ENV_VARIABLE=/^[\w&&\D]\w*$/
```
reitermarkus: I think this can be inlined.

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -1,4 +1,59 @@
...
+      }
+
+      describe "the resulting command line" do
+        it "includes the given variables explicitly" do
+          expect(Open3)
+            .to receive(:popen3)
+            .with(an_instance_of(Hash), ["/usr/bin/sudo"] * 2,
+                "-E", a_string_starting_with("PATH="),
+                "A=1", "B=2", "C=3", "--", "env", *env_args, {})
+            .and_wrap_original do |original_popen3, *_, &block|
```
reitermarkus: I think you can add 

```ruby
allow(Open3).to receive(:popen3)
```

before `expect` and in turn leave off `and_wrap_original`.

claui: The class under test expects the `#popen3` to return an `IO`-ish object, on which it will call e. g. `#close_write`.
Losing the `and_wrap_original` would cause the test double to return `nil`. The test would then error out.

I’ve been thinking of alternatives, e. g. `and_return` with an `IO` test double; I’m still slightly in favor of `and_wrap_original` for readability.

BTW: I’m probably missing something here but how is a (matching) `allow` before `expect` not a no-op?

reitermarkus: > BTW: I’m probably missing something here but how is a (matching) allow before expect not a no-op?

No, you're not missing anything. `allow` makes it a no-op, but if the return value is used we cannot do this.

Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -50,7 +50,13 @@ def run_installer(command: nil, verbose: false, **_options)
         end
         with_choices_file do |choices_path|
           args << "-applyChoiceChangesXML" << choices_path if choices_path
-          command.run!("/usr/sbin/installer", sudo: true, args: args, print_stdout: true)
+          logged_in_user = Utils.current_user
+          env = {
+            "LOGNAME" => logged_in_user,
+            "USER" => logged_in_user,
+            "USERNAME" => logged_in_user,
```
reitermarkus: > This PR proposes to fix the issue by setting LOGNAME, USER, and USERNAME to the same values as the graphical PKG installer does.

Is `USERNAME` set by this specific installer or by `/usr/sbin/installer` in general?

claui: I honestly have no idea. I never bothered to investigate _where_ exactly those three variables are exported in the chain of parent processes of the post-install script.

Rationale: Even though I can confirm that the `dymo-label` installer works fine without `USERNAME`, let’s not assume that any installer will.

PR #4275, opened 2018-06-04
10.14: recognise Xcode 10.0 & add pkg-config files.

DomT4:

-----

Still a bit of a WIP.

Fixes https://github.com/Homebrew/brew/issues/4276

### DIFF
--- a/Library/Homebrew/os/mac/pkgconfig/10.14/libexslt.pc
+++ b/Library/Homebrew/os/mac/pkgconfig/10.14/libexslt.pc
@@ -1,0 +1,12 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+
+Name: libexslt
+Version: 0.8.17
+Description: EXSLT Extension library
+Requires: libxml-2.0
+Libs: -L${libdir} -lexslt -lxslt  -lxml2 -lz -lpthread -licucore -lm
+Cflags: -I${includedir}
--- a/Library/Homebrew/os/mac/pkgconfig/10.14/libxslt.pc
+++ b/Library/Homebrew/os/mac/pkgconfig/10.14/libxslt.pc
@@ -1,0 +1,12 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+
+Name: libxslt
+Version: 1.1.29
+Description: XSLT library version 2.
+Requires: libxml-2.0
+Libs: -L${libdir} -lxslt  -lxml2 -lz -lpthread -licucore -lm
+Cflags: -I${includedir}
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -19,16 +19,18 @@ module OS
         when "10.11" then "8.2.1"
         when "10.12" then "9.2"
         when "10.13" then "9.3"
+        when "10.14" then "10.0"
         else
           raise "macOS '#{MacOS.version}' is invalid" unless OS::Mac.prerelease?
 
           # Default to newest known version of Xcode for unreleased macOS versions.
-          "9.3"
+          "10.0"
         end
       end
 
       def minimum_version
         case MacOS.version
+        when "10.14" then "10.0"
         when "10.13" then "9.0"
         when "10.12" then "8.0"
         else "2.0"
@@ -171,7 +173,8 @@ module OS
         when 81      then "8.3"
         when 90      then "9.2"
         when 91      then "9.3"
-        else              "9.3"
+        when 100     then "10.0"
+        else              "10.0"
         end
       end
 
@@ -222,6 +225,7 @@ module OS
         # on the older supported platform for that Xcode release, i.e there's no
         # CLT package for 10.11 that contains the Clang version from Xcode 8.
         case MacOS.version
+        when "10.14" then "1000.10.25.5"
         when "10.13" then "902.0.39.1"
         when "10.12" then "900.0.39.2"
         when "10.11" then "800.0.42.1"
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -18,7 +18,7 @@ module OS
         when "10.10" then "7.2.1"
         when "10.11" then "8.2.1"
         when "10.12" then "9.2"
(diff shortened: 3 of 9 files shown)


mistydemeo:
🤔 The clang I'm seeing from a CLT-only install doesn't return a valid value for `DevelopmentTools.clang_version`; regex isn't matching, so it returns `Version::NULL`. We'll need to update the regex to allow more than one digit in the major version:

```diff
diff --git a/Library/Homebrew/development_tools.rb b/Library/Homebrew/development_tools.rb
index 1f0d77d148..2ee5169476 100644
--- a/Library/Homebrew/development_tools.rb
+++ b/Library/Homebrew/development_tools.rb
@@ -63,7 +63,7 @@ class DevelopmentTools
     def clang_version
       @clang_version ||= begin
         if (path = locate("clang")) &&
-           build_version = `#{path} --version`[/(?:clang|LLVM) version (\d\.\d)/, 1]
+           build_version = `#{path} --version`[/(?:clang|LLVM) version (\d+\.\d)/, 1]
           Version.new build_version
         else
           Version::NULL
```

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -171,7 +173,8 @@ def detect_version_from_clang_version
         when 81      then "8.3"
         when 90      then "9.2"
         when 91      then "9.3"
-        else              "9.3"
+        when 10001   then "10.0"
```
mistydemeo: With the CLT, and the patch I mentioned earlier for `clang_version`, I'm seeing a value of `100` here, not `10001`. What string are you getting for `clang --version`? Are you using Xcode, and what version?

Library/Homebrew/os/mac/pkgconfig/10.14/libxslt.pc
```diff
@@ -0,0 +1,12 @@
...
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+
+Name: libxslt
+Version: 1.1.29
+Description: XSLT library version 2.
+Requires: libxml-2.0
+Libs: -L${libdir} -lxslt  -lxml2 -lz -lpthread -licucore -lm
```
ilovezfs: looks like there are two spaces between xslt and xml2

Library/Homebrew/os/mac/pkgconfig/10.14/libexslt.pc
```diff
@@ -0,0 +1,12 @@
...
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib
+includedir=${prefix}/include
+
+
+Name: libexslt
+Version: 0.8.17
+Description: EXSLT Extension library
+Requires: libxml-2.0
+Libs: -L${libdir} -lexslt -lxslt  -lxml2 -lz -lpthread -licucore -lm
```
ilovezfs: ditto

Library/Homebrew/os/mac/pkgconfig/10.14/libxslt.pc
DomT4: Yeah, it's just a direct copy from the prior `pkg-config` set in this case since there's no version change as far as I can see. I can fix it if ya like but I think the double spacing goes back to at least 10.10 and possibly earlier.

DomT4:
> The clang I'm seeing from a CLT-only install doesn't return a valid value for 

Hey Misty! Thanks for the review here.

I'm running semi-blind on this at the moment because my developer program membership with Apple is about to expire and, well, I can't afford to renew it at the moment, so I haven't tried to spin up a box yet because I don't want to get stuck with a VM I'll only receive updates for for another week or so.

If you have more direct access & there's anything that'd be easier for you to handle please feel free to say and I'll happily surrender it to you; don't want to needlessly eat up your time working through me if you've a limited amount of it or anything.

Library/Homebrew/os/mac/xcode.rb
DomT4: > I'm seeing a value of 100 here

That would be more like what I expected, to be honest! I assumed something was going on when `10001` popped out but I had to go grab some sleep so pushed it for review as-is in case someone else had a more sensible answer.

> What string are you getting for clang --version?

```
Apple LLVM version 10.0.0 (clang-1000.10.25.5)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

> Are you using Xcode, and what version?

Aye, 10.0.

DomT4:
Nice! Thanks for the help 🙇.

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -171,7 +173,8 @@ def detect_version_from_clang_version
         when 81      then "8.3"
         when 90      then "9.2"
         when 91      then "9.3"
-        else              "9.3"
+        when 100     then "10.0"
+        else              "10.0"
```
ilovezfs: what is the value for 9.4?

Library/Homebrew/os/mac/pkgconfig/10.14/libxslt.pc
ilovezfs: 👀

Library/Homebrew/os/mac/xcode.rb
DomT4: I completely forgot 9.4 had been released. Should probably update the `when "10.13" then "9.3"` as well.

It's still 91/9.1 though:

```
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Library/Homebrew/os/mac/pkgconfig/10.14/libxslt.pc
DomT4: Amusingly, the double spacing seems to be an upstream thing originally, and we've just kept it over the years.

DomT4:
Thanks everyone ❤️

DomT4:
> I haven't tried to spin up a box yet because I don't want to get stuck with a VM I'll only receive updates for for another week or so.

As a minor update I've set up a box locally now, so for as long as it keeps receiving updates & Apple doesn't notice my developer membership is about to expire I'll hopefully be of some use on 10.14 stuff.

PR #4288, opened 2018-06-06
docs: Add notes on revendoring

woodruffw:

-----

This adds some notes for the dependency revendoring process.

Pinging @Homebrew/maintainers for style nits.

### DIFF
--- a/docs/How-to-Revendor-a-Dependency.md
+++ b/docs/How-to-Revendor-a-Dependency.md
@@ -4,43 +4,41 @@ At the time of writing, Homebrew vendors several dependencies within the
 [Library/Homebrew/vendor](https://github.com/Homebrew/brew/tree/master/Library/Homebrew/vendor)
 subdirectory.
 
-This page serves to document the process for revendoring these dependencies.
+This page serves to document the process for revendoring these dependencies, i.e. replacing the
+vendored copy with a more recent version.
 
 ## Copying the files
+
+Here's an example for the imaginary "foobar" vendored dependency:
 
 Each directory in `vendor/` contains the **contents** of the `lib/` directory for the corresponding
 dependency. Thus, `require`ing a vendored dependency looks like this:
 
 ```ruby
-require "vendor/library/file"
+require "vendor/foobar/file"
 ```
 
-Thus, the copying process is as simple as removing the old files, and copying the new files into
-place.
-
-Here's an example for the imaginary "foobar" vendored dependency:
+The copying process is almost as simple as removing the old files, and copying the new files into
+place:
 
 ```bash
-rm -rf /path/to/brew/Library/Homebrew/vendor/foobar/*
-cp -R /path/to/foobar/lib/* /path/to/brew/Library/Homebrew/vendor/foobar/
+git rm -rf --cached /path/to/brew/Library/Homebrew/vendor/foobar
+rm -rf /path/to/brew/Library/Homebrew/vendor/foobar/
+cp -R /path/to/foobar/lib/ /path/to/brew/Library/Homebrew/vendor/foobar/
 ```
 
-Note that we copy the **contents** of `foobar/lib/` to the vendored directory, not `lib/` itself.
-
-Note, also, that the `rm` step is not necessary if you know that the updated version does not
-contain fewer files than the current version. Running it will never hurt, though, so there's no
-point in skipping it.
+Note that the new vendored directory will be a copy of `foobar/lib`, not of `foobar` itself.
 
 ## Updating the README
 
-Once the actual files of the vendored dependency are updated, you should make sure that the
+Once the files are updated, make sure that the
 [README](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/vendor/README.md) within
 `vendor/` is still accurate.
 
 This includes:
 
 1. Making sure the URL listed for the dependency is still correct
-2. Updating the version listed next to the URL
+2. Updating the version number listed next to the URL
 3. Updating the quoted license text for the dependency, if necessary
 
 That's it!
--- a/docs/How-to-Revendor-a-Dependency.md
+++ b/docs/How-to-Revendor-a-Dependency.md
@@ -1,0 +1,46 @@
+# How to Revendor a Dependency
+
+At the time of writing, Homebrew vendors several dependencies within the
+[Library/Homebrew/vendor](https://github.com/Homebrew/brew/tree/master/Library/Homebrew/vendor)
+subdirectory.
+
+This page serves to document the process for revendoring these dependencies.
+
+## Copying the files
+
+Each directory in `vendor/` contains the **contents** of the `lib/` directory for the corresponding
+dependency. Thus, `require`ing a vendored dependency looks like this:
+
+```ruby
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Any thoughts/chance on making this a script/dev-cmd instead of docs?

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+## Copying the files
+
+Each directory in `vendor/` contains the **contents** of the `lib/` directory for the corresponding
+dependency. Thus, `require`ing a vendored dependency looks like this:
+
+```ruby
+require "vendor/library/file"
+```
+
+Thus, the copying process is as simple as removing the old files, and copying the new files into
```
claui: > […] process is almost as simple as […]

(Rationale: the old files need to be removed from the Git index, too.)

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+place.
+
+Here's an example for the imaginary "foobar" vendored dependency:
+
+```bash
```
claui: I’d suggest to prepend this script with the following line:

```bash
git rm -rf --cached /path/to/brew/Library/Homebrew/vendor/foobar
```

This will clean the Git index from obsolete files, i. e. files which are not present in the updated copy.
(Even though it’s an extra step, it makes `git add`ing the new copy much less confusing.)

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+rm -rf /path/to/brew/Library/Homebrew/vendor/foobar/*
```
claui: To ensure obsolete dotfiles are removed, write this instead:

```bash
rm -rf /path/to/brew/Library/Homebrew/vendor/foobar
```

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+cp -R /path/to/foobar/lib/* /path/to/brew/Library/Homebrew/vendor/foobar/
```
claui:
```bash
cp -R /path/to/foobar/lib /path/to/brew/Library/Homebrew/vendor/foobar
```

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+Note that we copy the **contents** of `foobar/lib/` to the vendored directory, not `lib/` itself.
+
+Note, also, that the `rm` step is not necessary if you know that the updated version does not
+contain fewer files than the current version. Running it will never hurt, though, so there's no
+point in skipping it.
```
claui: What do you think about removing this sentence altogether, just to simplify things?

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+place.
+
+Here's an example for the imaginary "foobar" vendored dependency:
+
+```bash
+rm -rf /path/to/brew/Library/Homebrew/vendor/foobar/*
+cp -R /path/to/foobar/lib/* /path/to/brew/Library/Homebrew/vendor/foobar/
+```
+
+Note that we copy the **contents** of `foobar/lib/` to the vendored directory, not `lib/` itself.
```
claui: Due to the changes above, rephrase this to:

> Note that the new vendored directory will be a copy of `foobar/lib`, not of `foobar` itself.

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+## Updating the README
+
+Once the actual files of the vendored dependency are updated, you should make sure that the
+[README](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/vendor/README.md) within
+`vendor/` is still accurate.
+
+This includes:
+
+1. Making sure the URL listed for the dependency is still correct
+2. Updating the version listed next to the URL
```
claui: Very minor nit:

> […] the version number listed […]

docs/How-to-Revendor-a-Dependency.md
```diff
@@ -0,0 +1,46 @@
...
+
+Note that we copy the **contents** of `foobar/lib/` to the vendored directory, not `lib/` itself.
+
+Note, also, that the `rm` step is not necessary if you know that the updated version does not
+contain fewer files than the current version. Running it will never hurt, though, so there's no
+point in skipping it.
+
+## Updating the README
+
+Once the actual files of the vendored dependency are updated, you should make sure that the
```
claui: > Once the files […] are updated, make sure that […]

claui:
> Any thoughts/chance on making this a script/dev-cmd instead of docs?

@MikeMcQuaid Even though I’m not directly addressed/involved, my two cents:

If I had to revendor something, I know for sure I’d never use a `dev-cmd` script because I’d have no idea where it is, how to use it and what its assumptions/prerequisites are. To me personally, following a nice checklist with snippets is a lot easier than remembering there is a script, figuring out how to call it, then forgetting it five minutes later.

YMMV!

docs/How-to-Revendor-a-Dependency.md
ilovezfs: second "thus" in a row is two thuses too many

woodruffw:
Thanks for the recommendations @claui! I've applied them.

Re: making it a dev-cmd: I agree that doing so would probably cause more confusion than just documenting the procedure. We don't revendor all that frequently, so the chances of any of us saving time by remembering the correct invocation are slim :slightly_smiling_face:

MikeMcQuaid:
Even if the code bits are a script rather than `dev-cmd` that would be preferable. Copy-pasting code from docs tends to break much faster than actual code. Related reading: https://githubengineering.com/runnable-documentation/

claui:
Interesting read! I particularly like the last example from the article – there’s still documentation, it’s just that the code bits have been replaced by script invocations (whose usage is still explained well).

With that being said, how would we go about wrapping things like `rm -rf` in an external script safely?

MikeMcQuaid:
The more I think about this the more I'd really like a `brew vendor-gem` (perhaps with a different name) developer command that takes the latest version of a gem from RubyGems and vendors it to a given directory. I'd like something I can run that produces a diff if any of those gems need updated. See https://github.com/github/homebrew-bootstrap/blob/master/cmd/brew-vendor-gem for some inspiration.

I think in a really ideal world we'd consider using Bundler to do some of these updates for us (but don't require the use of Bundler to actually use the relevant gems).

woodruffw:
Yeah, I agree. I'll create a PR for something like `brew vendor-gem` later this week.

Like I mentioned in https://github.com/Homebrew/brew/pull/4288#issuecomment-395496777 doing so will leave out vendored library fragments, but I think that's a relatively minor case to miss.

MikeMcQuaid:
I've created a PR for something to handle this a bit better using Bundler in https://github.com/Homebrew/brew/pull/4895.

PR #4296, opened 2018-06-06
brew tests: suppress "Run options: exclude..." output

apjanke:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). *No: `brew tests` is not itself tested.*

-----

Gets rid of that `Run options: exclude {...}` spam in the output of `brew tests`.

Before:

```
$ brew tests       
Randomized with seed 26599
8 processes for 226 specs, ~ 28 specs per process
Run options: exclude {:needs_linux=>true}
.......Run options: exclude {:needs_linux=>true}
...Run options: exclude {:needs_linux=>true}
Run options: exclude {:needs_linux=>true}
......Run options: exclude {:needs_linux=>true}
.Run options: exclude {:needs_linux=>true}
Run options: exclude {:needs_linux=>true}
.Run options: exclude {:needs_linux=>true}
..................*...........................................................
```

After:

```
$ brew tests                                               
Randomized with seed 24974
8 processes for 226 specs, ~ 28 specs per process
..........................................................................................................................................................................................................................................
```

### DIFF
--- a/Library/Homebrew/test/support/quiet_progress_formatter.rb
+++ b/Library/Homebrew/test/support/quiet_progress_formatter.rb
@@ -1,0 +1,14 @@
+require "rspec/core/formatters/progress_formatter"
+
+class QuietProgressFormatter < RSpec::Core::Formatters::ProgressFormatter
+  RSpec::Core::Formatters.register self, :seed
+
+  def seed(notification); end
+
+  def message(notification)
+    if notification.message.start_with? "Run options: exclude"
+      return
+    end
+    puts "MESSAGE: #{notification}"
+  end
+end
--- a/Library/Homebrew/dev-cmd/tests.rb
+++ b/Library/Homebrew/dev-cmd/tests.rb
@@ -103,7 +103,7 @@ module Homebrew
         --seed #{seed}
         --color
         --require spec_helper
-        --format NoSeedProgressFormatter
+        --format QuietProgressFormatter
         --format ParallelTests::RSpec::RuntimeLogger
         --out #{HOMEBREW_CACHE}/tests/parallel_runtime_rspec.log
       ]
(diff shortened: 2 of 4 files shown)


Library/Homebrew/test/support/quiet_progress_formatter.rb
```diff
@@ -0,0 +1,14 @@
+require "rspec/core/formatters/progress_formatter"
+
+class QuietProgressFormatter < RSpec::Core::Formatters::ProgressFormatter
+  RSpec::Core::Formatters.register self, :seed
```
reitermarkus: Should this include `:message` now?

reitermarkus:
Actually, this is supported by upstream:

https://github.com/rspec/rspec-core/blob/5e395e2016f1da19475e6db2817eb26dae828c4c/Filtering.md#silencing-filter-announcements

apjanke:
Do you know how to pass an `Rspec.configure` to the `parallel_rspec` we're using? I saw that, but didn't know how to make it happen inside how we use Rspec. I'm kind of an Rspec newb.

Library/Homebrew/test/support/quiet_progress_formatter.rb
apjanke: I don't think so. It works without it. From my reading of the docs I think the fact that we're explicitly specifying this `--format` on the command line makes it (and possibly this whole `register()` call) unnecessary.

reitermarkus:
`config.silence_filter_announcements = true` in `spec_helper.rb`.

apjanke:
Oh my gosh; I missed it because I got the case wrong for `RSpec` in my `grep`ing.

New PR: #4298

PR #4299, opened 2018-06-07
superenv: set effective_sysroot on 10.14 CLT

mistydemeo:

-----

I'm mostly opening this for documentation. It's probably too early in the cycle to start committing workarounds like this.

The current version of the CLT doesn't include headers in the traditional Unixy locations, which breaks some software and which means that we end up not setting some special include paths. I've filed this bug as `rdar://40833474`, but meanwhile this workaround fixes broken builds like imagemagick.

If the traditional Unix locations aren't coming back, we'll have to look at wider-reaching changes, but I'll want to let more time pass before looking at something like that.

See also https://github.com/Homebrew/homebrew-core/issues/28703 and https://github.com/Homebrew/homebrew-core/pull/28733.

### DIFF
--- a/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
@@ -102,7 +102,15 @@ module Superenv
   end
 
   def effective_sysroot
-    MacOS.sdk_path.to_s if MacOS::Xcode.without_clt?
+    if MacOS::Xcode.without_clt?
+      MacOS.sdk_path.to_s
+    # Possibly a temporary workaround, depending on how the
+    # CLT shakes out this dev cycle.
+    # We need this sysroot set because there are no headers
+    # in /usr.
+    elsif MacOS.version >= :mojave
+      "#{MacOS::CLT::PKG_PATH}/SDKs/MacOSX#{MacOS.version}.sdk"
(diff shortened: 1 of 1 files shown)


andrewpile:
FWIW, the [10.14 release notes](https://download.developer.apple.com/Documentation/Xcode_10_beta_Release_Notes/Xcode_10_Beta_Release_Notes.pdf) make mention of the missing headers and provide a package for installing them in `/usr/include`, this is an additional step after installing CLT.

> The Command Line Tools package installs the macOS system headers inside the macOS SDK.
Software that compiles with the installed tools will search for macOS headers at the standard
include path:
>
> /Applications/Xcodebeta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include
>
> For legacy software that looks for the macOS headers in the base system under /usr/include, please
install the package file located at:
>
> /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg

ilovezfs:
😭

mistydemeo:
@andrewpile Yeah, I'm planning on replacing this PR with one that diretly addresses that... thanks!

yurikoles:
>rdar://40833474
How to open that link $_%?

Issue #4058, opened 2018-04-12
Enable `brew style --rspec` by default


MikeMcQuaid:
It would be good if we could fix up all the rules that made sense in `brew style --rspec` and disable those which do not. This would mean we could enable `brew style --rspec` by default and guide future contributors and maintainers to write better tests.

aamarill:
Hello,

I am looking to make my first open source contribution. Is this a good issue to take on?
If not, feel free to point me other issues that might be good for someone that's new to open source.
Thanks!

reitermarkus:
> Is this a good issue to take on?

If you have experience with Ruby (and RSpec), go for it. Just run `brew style --rspec` and try to fix some issues. 😉

kennethlarsen:
@MikeMcQuaid I'm looking through the output of `brew style --rspec` and a bit curious to what rules wouldn't make sense to keep? Would be a bit sad to fix 100s of `Name your test subject if you need to reference it explicitly` if we don't want to keep the rule.

aamarill:
@kennethlarsen just FYI, I am not working on this. So feel free to take it on if you want/can! 😄

reitermarkus:
I didn't see any rules which aren't useful. All are meant to make the tests easier to maintain and easier to read, so I don't think we want to disable any for now.

kennethlarsen:
@MikeMcQuaid Sounds good! I'm currently looking at this rule `Start context description with 'when', 'with', or 'without'`

Basically, a lot of context descriptions in the codebase start with `should` as in 
```
        context "should not warn on an newer version revision removal" do
          before do
            formula_gsub_commit "revision 2", ""
            formula_gsub_commit "foo-1.0.tar.gz", "foo-1.1.tar.gz"
          end

          it { is_expected.to be_nil }
        end
```

Seems like a good idea to follow the suggested convention.

reitermarkus:
> Seems like a good idea to follow the suggested convention.

It's definitely a good idea to start with those which *you* think are useful.

MikeMcQuaid:
> It's definitely a good idea to start with those which you think are useful.

Agreed 👍 

> Seems like a good idea to follow the suggested convention.

Agreed. 

---

In general I think with both RuboCop RSpec and general RuboCop: other than style preferences or Weird Things We Do That We Have A Good Reason for I think it makes the codebase easier for everyone to grok (particularly new contributors) if we deviate as little from the defaults as possible.

apjanke:
[...]

PR #4304, opened 2018-06-07
brew style: assume file args are non-formulae when passed --rspec

apjanke:

-----

Modifies `brew style` to always use the non-audit, rubocop-rspec config file when the `--rspec` option is passed, even if files were specified on the command line. This makes it feasible to test individual spec definition files with `brew style --rspec` to help work on #4058. Otherwise, it won't pick up the RSpec configurations, which you may need to modify while working on that task. And if you omit a file and just do `brew style --rspec` on everything, there are so many cop violations that it spams your terminal.

### DIFF
--- a/Library/Homebrew/style.rb
+++ b/Library/Homebrew/style.rb
@@ -57,12 +57,24 @@ module Homebrew
         args << "--only" << cops_to_include.join(",")
       end
 
+      has_non_formula = Array(files).any? do |file|
+        File.expand_path(file).start_with? HOMEBREW_LIBRARY_PATH
+      end
+      config_file = if files.nil? || has_non_formula
+        if ARGV.include?("--rspec")
+          HOMEBREW_LIBRARY_PATH/".rubocop-rspec.yml"
+        else
+          HOMEBREW_LIBRARY_PATH/".rubocop.yml"
+        end
+      else
+        HOMEBREW_LIBRARY/".rubocop_audit.yml"
+      end
+
+      args << "--config" << config_file
+
       if files.nil?
-        config_file = ARGV.include?("--rspec") ? ".rubocop-rspec.yml" : ".rubocop.yml"
-        args << "--config" << HOMEBREW_LIBRARY_PATH/config_file
         args << HOMEBREW_LIBRARY_PATH
       else
-        args << "--config" << HOMEBREW_LIBRARY/".rubocop_audit.yml"
         args += files
       end
 
--- a/Library/Homebrew/test/cmd/style_spec.rb
+++ b/Library/Homebrew/test/cmd/style_spec.rb
@@ -31,4 +31,18 @@ describe "brew style" do
         .to include("Extra empty line detected at class body beginning.")
     end
   end
+
+  describe "Homebrew::check_style_and_print" do
+    let(:dir) { mktmpdir }
+
+    it "returns false for conforming file with only audit-level violations" do
+      # This file is known to use non-rocket hashes and other things that trigger audit,
+      # but not regular, cop violations
+      target_file = HOMEBREW_LIBRARY_PATH/"utils.rb"
+
+      rubocop_result = Homebrew::Style.check_style_and_print([target_file])
+
+      expect(rubocop_result).to eq false
+    end
+  end
 end


reitermarkus:
Maybe instead switch based on whether the passed files are under `HOMEBREW_LIBRARY_PATH` so it also works without `--rspec`?

Library/Homebrew/cmd/style.rb
```diff
@@ -111,12 +111,18 @@ def check_style_impl(files, output_type, options = {})
       args << "--only" << cops_to_include.join(",")
     end
 
+    if files.nil? || ARGV.include?("--rspec")
```
MikeMcQuaid: `config file = if ...`

Library/Homebrew/cmd/style.rb
```diff
@@ -111,12 +111,18 @@ def check_style_impl(files, output_type, options = {})
...
+      config_file_base = ARGV.include?("--rspec") ? ".rubocop-rspec.yml" : ".rubocop.yml"
```
MikeMcQuaid: Think you could just `if`/`else` this without variables and use `HOMEBREW_LIBRARY_PATH` in both?

Library/Homebrew/cmd/style.rb
apjanke: Amended to fix.

Library/Homebrew/cmd/style.rb
apjanke: Amended to fix.

apjanke:
How about something like this?

```
    has_non_formula = files.nil? || files.any? do |file|
      puts "FILE: #{file}"
      File.expand_path(file).start_with? HOMEBREW_LIBRARY_PATH
    end
    config_file = if has_non_formula
      HOMEBREW_LIBRARY_PATH/(ARGV.include?("--rspec") ? ".rubocop-rspec.yml" : ".rubocop.yml")
    else
      HOMEBREW_LIBRARY/".rubocop_audit.yml"
    end
```

Amended to do this.

Library/Homebrew/cmd/style.rb
```diff
@@ -111,12 +111,21 @@ def check_style_impl(files, output_type, options = {})
       args << "--only" << cops_to_include.join(",")
     end
 
+    has_non_formula = files.nil? || files.any? do |file|
```
MikeMcQuaid: `Array(files).any?` to avoid the `nil?` check.

Library/Homebrew/cmd/style.rb
```diff
@@ -111,12 +111,21 @@ def check_style_impl(files, output_type, options = {})
...
+      puts "FILE: #{file}"
```
MikeMcQuaid: Wants removed.

Library/Homebrew/cmd/style.rb
```diff
@@ -111,12 +111,21 @@ def check_style_impl(files, output_type, options = {})
...
+      File.expand_path(file).start_with? HOMEBREW_LIBRARY_PATH
```
MikeMcQuaid: Is the `expand_path` needed/warranted here?

Library/Homebrew/cmd/style.rb
```diff
@@ -111,12 +111,21 @@ def check_style_impl(files, output_type, options = {})
...
+    end
+    config_file = if has_non_formula
+      HOMEBREW_LIBRARY_PATH/(ARGV.include?("--rspec") ? ".rubocop-rspec.yml" : ".rubocop.yml")
```
MikeMcQuaid: @apjanke I don't think the ternary operator is worth having just to avoid duplicating `HOMEBREW_LIBRARY_PATH/`. `if has_formula ... elif ARGV.include?("--rspec") ... else` would be clearest, I think (but involve reversing the `has_non_formula` to `has_formula` which I think would also be clearer).

Library/Homebrew/cmd/style.rb
reitermarkus: Or `has_formula = files.none? { … }`.

Library/Homebrew/cmd/style.rb
apjanke: Yes. Consider the case where you're in `/usr/local/Homebrew/Library/Homebrew` and do `brew style ./descriptions.rb`. Since you're under the library, we want it to return true, but it won't unless the path is expanded to resolve that leading `./`.

apjanke:
Amended to address most of the comments.

Library/Homebrew/cmd/style.rb
apjanke: `has_non_formula` doesn't reverse to `has_formula`, it reverses to `doesnt_have_non_formula` or `all_are_formula`, and the `if/then` logic is not equivalent to `if has_formula ... elif ARGV.include?("--rspec") ... else`. (Because there might be a mix of formula and non-formula files, and the way it's coded, the presence of non-formula files is the determining factor.)

I removed the ternary operator, though.

apjanke:
Rebased.

MikeMcQuaid:
`brew style` 🔥

apjanke:
The CI style failures are not in the code that changed in this PR; it's raising them in a bunch of other files.

```
Inspecting 662 files
CCC..C.CCC..C...C..C.CCCC.CCC...CC.CC.C..CCC.CC..CC.C.C...CC.C.CC.C.CCC..CC..C........C.CC....C.....C.........................................................................................C......................C...CC..CC..........C.CC.............C...........C........CCC...CC...C......C.CCC.CC.......................C....C......CC...CCCC.............C........C.CCC..CCCCCC..CCCC....C..CC..C.........CCCCCCCCCCCCC.C.C..CCCCCCCCC.CCCC.CCCCCCCCC.C.CCCC...C...CC.C.CCCCC.C.CCCCCCCCCC.CCCCCCC..CCCCCCCCCCCCC.CCCCCC.CCCC..C.CCC.C.CC.C.CC.CCCCCCC.C.CCCCCCCC.CCCCC.C...CC.CCC...........CC..CC.....C.........C..C.CCC.C....C.C.C.........CCC....CCC...CC.CCC.CCC..C.C.C.
Offenses:
Library/Homebrew/brew.rb:86:30: C: Use hash rockets syntax.
    Homebrew::Help.help cmd, empty_argv: empty_argv
                             ^^^^^^^^^^^
Library/Homebrew/brew.rb:127:28: C: Use hash rockets syntax.
  Homebrew::Help.help cmd, usage_error: e.message
                           ^^^^^^^^^^^^
Library/Homebrew/build.rb:163:28: C: Use hash rockets syntax.
    keg.detect_cxx_stdlibs(skip_executables: true)
                           ^^^^^^^^^^^^^^^^^
Library/Homebrew/build.rb:166:14: C: Method parameter must be at least 3 characters long.
  def fixopt(f)
             ^
Library/Homebrew/build_environment.rb:12:10: C: Method parameter must be at least 3 characters long.
  def <<(o)
         ^
```

It looks like the CI `brew style` check is enabling the formula RSpec configuration for non-formula files. This isn't happening for me locally, though.

I'll see if I can figure out why this is happening, and whether it's due to the changes in this PR.

MikeMcQuaid:
@apjanke Master is so I think it's this PR.

apjanke:
`brew style` passed this time; it's only codecov complaints now.

Added a test.

Writing a test for this is kind of hard, because the behavior of switching between the audit and non-audit Rubocop config files is a little opaque (and it depends on the location of target files being under HOMEBREW_LIBRARY_PATH, so it's hard to mock). If we want more testing on `style`, we should probably expose its behavior controls as options to the functions, and not just `ARGV` checks.

MikeMcQuaid:
Thanks @apjanke! Just for future: I reckon this type of integration test (that will be shelling out to `rubocop`) should be actually running `brew style` and we try to keep tests not like that as unit tests (or at least "sticking in the same process" tests)

PR #4306, opened 2018-06-07
Refactor `brew cask search`.

reitermarkus:

-----

Refactor `brew cask search` once more before merging it into `brew search`.

### DIFF
--- a/Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
+++ b/Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
@@ -85,6 +85,7 @@ RSpec.shared_context "integration test" do
         "-W0",
         "-I", "#{HOMEBREW_LIBRARY_PATH}/test/support/lib",
         "-I", HOMEBREW_LIBRARY_PATH.to_s,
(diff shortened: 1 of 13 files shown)


Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb
```diff
@@ -85,6 +85,7 @@ def brew(*args)
         "-W0",
         "-I", "#{HOMEBREW_LIBRARY_PATH}/test/support/lib",
         "-I", HOMEBREW_LIBRARY_PATH.to_s,
+        "-I", "#{HOMEBREW_LIBRARY_PATH}/cask/lib",
```
MikeMcQuaid: What's this needed for? It'd be nice to instead move logic that's needed for a generalised `brew` integration test out of `cask/lib`.

reitermarkus: The generic stuff is already outside of `cask/lib`, but we cannot get around having access to the `Cask` class.

reitermarkus: It doesn't load code, it only adds this path to the `$LOAD_PATH`.

PR #4311, opened 2018-06-09
text_cop: require cargo to use `install` instead of `build`

commitay:

-----
Opened this for discussion, I'm still checking if all of the `build` formulae work as expected with `install` or if we would need to whitelist any.

Currently we prefer `cargo install --root prefix` (which doesn't need a `bin.install`) over `cargo build`. We have 27 formulae that run `cargo`, 18 have `build` and 9 have `install`.

### DIFF
--- a/Library/Homebrew/rubocops/text_cop.rb
+++ b/Library/Homebrew/rubocops/text_cop.rb
@@ -53,6 +53,10 @@ module RuboCop
             next if parameters_passed?(d, /vendor-only/)
             problem "use \"dep\", \"ensure\", \"-vendor-only\""
           end
+
+          find_method_with_args(body_node, :system, "cargo", "build") do
+            problem "use \"cargo\", \"install\", \"--root\", prefix"
+          end
         end
       end
     end
--- a/Library/Homebrew/test/rubocops/text_cop_spec.rb
+++ b/Library/Homebrew/test/rubocops/text_cop_spec.rb
@@ -191,5 +191,19 @@ describe RuboCop::Cop::FormulaAudit::Text do
         end
       RUBY
     end
+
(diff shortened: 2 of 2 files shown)


ilovezfs:
I don't think we can rule out the possibility of legitimately needing to run `cargo build` not `cargo install`.

Library/Homebrew/rubocops/text_cop.rb
```diff
@@ -53,6 +53,10 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
             next if parameters_passed?(d, /vendor-only/)
             problem "use \"dep\", \"ensure\", \"-vendor-only\""
           end
+
+          find_method_with_args(body_node, :system, "cargo", "build") do
+            problem "use \"cargo\", \"install\", \"root\", prefix"
```
ilovezfs: `--root`

commitay:
> I don't think we can rule out the possibility of legitimately needing to run `cargo build` not `cargo install`.

Yeah, I guess it's possible. I haven't found any formulae that need `build` yet but I'm sure something will come along. Mostly wanted this to nudge new formulae to the preferred style but I suppose we could change all the formulae that we can to this style and hope people copy the right one?

ilovezfs:
It looks like they're equivalent except that `install` defaults to a `--release` build whereas `build` defaults to a `--debug` build so I think we'll be OK.

commitay:
Do you want PRs changing formulae to use install or just wait til they get bumped and fail audit?

ilovezfs:
Voilà https://github.com/Homebrew/homebrew-core/pull/28846

ilovezfs:
Thanks @commitay!

PR #4316, opened 2018-06-09
Merge `brew cask search` into `brew search`.

reitermarkus:

-----

Sample output:

```
$ brew search pip
==> Formulae
aespipe                pip-completion         pipenv                 pyenv-pip-migrate
brew-pip               pipebench              pipes-sh               soundpipe
lesspipe               pipemeter              pipeviewer             spiped

==> Casks
artpip                                        secure-pipes

If you meant "pip" specifically:
Homebrew provides pip via: `brew install python`. However you will then
have two Pythons installed on your Mac, so alternatively you can install
pip via the instructions at:
  https://pip.readthedocs.io/en/stable/installing/
```


Library/Homebrew/search.rb
```diff
@@ -53,9 +53,6 @@ def search_taps(query, silent: false)
     end
 
     def search_formulae(string_or_regex)
-      # Use stderr to avoid breaking parsed output
-      $stderr.puts Formatter.headline("Searching local taps...", color: :blue)
```
MikeMcQuaid: Is this being removed just because we're already using similar headers elsewhere on stdout?

Library/Homebrew/test/cmd/search_spec.rb
```diff
@@ -26,12 +25,11 @@
       .and be_a_success
   end
 
-  it "falls back to a GitHub tap search when no formula is found", :needs_network, retry: 3 do
+  it "falls back to a GitHub tap search when no formula is found", :needs_macos, :needs_network, retry: 3 do
```
MikeMcQuaid: What happens here on Linux, out of interest?

Library/Homebrew/cask/lib/hbc/cli/search.rb
```diff
@@ -1,53 +0,0 @@
-require "search"
-
-module Hbc
-  class CLI
-    class Search < AbstractCommand
```
MikeMcQuaid: What does `brew cask search` do now? Would be good to have a `odeprecated` message telling people to use `brew search`?

reitermarkus: There is, it's in `compat`.

MikeMcQuaid: @reitermarkus Gotcha, thanks.

Library/Homebrew/cmd/search.rb
```diff
@@ -58,40 +59,67 @@ def search(argv = ARGV)
...
+        search_formulae(string_or_regex)
+      end
 
-      remote_results = search_taps(query)
-      tap_results = [*remote_results[:formulae], *remote_results[:casks]].sort
-      puts Formatter.columns(tap_results) unless tap_results.empty?
+      remote_formulae = remote_results[:formulae]
+      all_formulae = local_formulae + remote_formulae
+
+      local_casks = if !OS.mac?
```
MikeMcQuaid: Would like to see this use the `extend/os` pattern rather than an `OS.mac?` check.

Library/Homebrew/search.rb
reitermarkus: Yes, we now have 

```
==> Formulae
…

==> Casks
…
```

and I think it is obvious that we are “searching local taps”, so I don't think it's necessary to output this information.

reitermarkus:
@MikeMcQuaid and @JCount, I added the ability to search Casks by name in the latest commit, please re-review.

MikeMcQuaid:
@reitermarkus Looks good to me! Merge when ready 👍

PR #4324, opened 2018-06-10
[WIP] Add meta-formula supprt

gigaflw:

-----

Hi! This is the first commit for GSoC project [add meta-formula support](https://github.com/Homebrew/gsoc/issues/10)
Here I report my current progress.

Meta-formula is a special kind of formula which will
* Allow bulk management of several related formulae
* extend the semantics of 'dependency' to casks

## Usage
In this commit, the following use case is covered:
```ruby
# in metafoo.rb
class MetaFoo < MetaFormula
    desc "hello world"
    version "0.1.0"	# mandatory, for `url` is not
    depends_on "dep_a"
    depends_on "dep_b"
end
```
Then, in command line

	$ brew install metafoo   // dep_a and dep_b will be installed

After this, `dep_a` and `dep_b` will be added as common dependencies, but `metafoo` will still have its own keg folder (PATH/Cellar/metafoo/0.1.0/...) even if it is empty. I think it might be better to keep the behavior of formula and metaformula consistent.

## Files Changed

**`Library/Homebrew/metaformula.rb`**

`MetaFormula` is a subclass of `Formula` in which
* `Formula#spec_eval` and `Formula#validate_attributes!` are overridden to allow a formula without a url attribute
* certain DSL which is not designed to be used in MetaFormula are disabled for now to avoid undefined behavior

```ruby
class MetaFormula < Formula
    # override to allow spec without a url
    def spec_eval(name); ...; end

    # override to allow passing validation without a url
    def validate_attributes!; ...; end

    # disable undefined DSL
    def self.disable_dsl(name); ...; end
    disable_dsl :xxx
     ...
end
```

**`Library/Homebrew/download_strategy.rb`**

`NoDownloadStrategy` is a new subclass of `AbstractDownloadStrategy` to accommodate the fact that `MetaFormula` can download nothing.

```ruby
class NoDownloadStrategy < AbstractDownloadStrategy; end
```
This class is set to be the default value of  `Resouce@download_strategy`. I think this is safe because whenever `url` is called (and it must be) in regular formula DSL, `Resouce@download_strategy` will be deduced and be set to a value other than `NoDownloadStrategy`.
I still have the concern of changing the default behavior of these stable code though, let me know if I am not doing this right.

**`Library/Homebrew/resource.rb`**
And some code is added to use the class:
 
```ruby
class Resource
    def initialize
        ...
        @download_strategy = NoDownloadStrategy  # use this as default
    end
    def stage
        ...
	# do not download if NoDownloadStrategy
        verify_download_integrity(fetch) unless @download_strategy == NoDownloadStrategy
        ...
    end
    ...
end
```
## Future Work
This commit only covers a small part of the intended functions. Future work includes
* Dependency on casks
* Bulk management other than `install` (upgrade, uninstall, etc.) 

Let me know if I am doing anything improperly :pray: and I'll try my best to improve. Thanks!


Library/Homebrew/formula.rb
```diff
@@ -2460,3 +2460,7 @@ def link_overwrite_paths
     end
   end
 end
+
+require "metaformula"
+# require metaformula so that whenever {Formula} is visible, so is {MetaFormula}
+# require in the end to avoid circular requirement
```
MikeMcQuaid: That's still a circular requirement. I'd remove it here and add the require somewhere else.

MikeMcQuaid: That's still a circular requirement. I'd remove it here and add the require somewhere else.

Library/Homebrew/metaformula.rb
```diff
@@ -0,0 +1,56 @@
+require "formula"
+
+# A meta-formula is a special kind of formula which will
+# * Allow bulk management of several related formulae (only `install `now)
```
MikeMcQuaid: Fix spacing around `install`

Library/Homebrew/metaformula.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  # Ignore url error in attribute validation
+  def validate_attributes!
+    super
+  rescue FormulaValidationError => e
+    raise unless e.attr.equal? :url
+  end
+
+  # Metaformula is not designed to be bottled for now
+  def pour_bottle?
```
MikeMcQuaid: I'd consider setting `bottle_disabled` instead which will prevent it being bottled or installed from a bottle.

Library/Homebrew/metaformula.rb
```diff
@@ -0,0 +1,56 @@
...
+    false
+  end
+
+  # DSL not designed to be used in MetaFormula for now are disabled to avoid undefined behaviors
+  # This is subject to change if certain functions should be assigned to them
+  def self.disable_dsl(name)
```
MikeMcQuaid: Would be good to take multiple arguments here.

Library/Homebrew/metaformula.rb
```diff
@@ -0,0 +1,56 @@
...
+    old_method = method(name)
+    define_singleton_method(name) do |*args, &block|
+      if block
+        opoo "#{name} DSL is ignored in MetaFormula"
```
MikeMcQuaid: I'd `raise` here instead.

Library/Homebrew/metaformula.rb
```diff
@@ -0,0 +1,56 @@
...
+      else
+        old_method.call(*args)
```
MikeMcQuaid: Why is this passing through the value to the old method?

Library/Homebrew/resource.rb
```diff
@@ -47,6 +47,7 @@ def initialize(name = nil, &block)
     @checksum = nil
     @using = nil
     @patches = []
+    @download_strategy = NoDownloadStrategy
```
MikeMcQuaid: Is this set elsewhere?

Library/Homebrew/resource.rb
```diff
@@ -87,7 +88,7 @@ def stage(target = nil, &block)
       raise ArgumentError, "target directory or block is required"
     end
 
-    verify_download_integrity(fetch)
+    verify_download_integrity(fetch) unless @download_strategy == NoDownloadStrategy
```
MikeMcQuaid: Would be nice to make this a helper method or push the check into `verify_download_integrity` itself.

Library/Homebrew/download_strategy.rb
```diff
@@ -139,6 +139,8 @@ def fossilpath
   end
 end
 
+class NoDownloadStrategy < AbstractDownloadStrategy; end
```
MikeMcQuaid: How about `NullDownloadStrategy`?

Library/Homebrew/metaformula.rb
gigaflw: Because some DSL functions (like `Formula::stable`) are also called by other functions (like `Formula::url`).  I think we should only block callings when used as DSL,

Library/Homebrew/resource.rb
gigaflw: This is also set when `Formula::url` is called (which is mandatory in regular formulae), and then the strategy will be deduced from url in [`Resource#url`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/resource.rb).

Library/Homebrew/resource.rb
gigaflw: Is something like this ok?
```ruby
def verify_download_integrity(fn)  # fn is a {Pathname}
     #if fn.file?    # old check
     if fn and fn.file?  # fn is nil when using Metaformula
     ...
end
```

Library/Homebrew/metaformula.rb
MikeMcQuaid: Sure but as-is this will mean unfortunately that `head "http://..../"`, `bottle :disable` and `patch "http://.../"` in a formula will still work.

Library/Homebrew/resource.rb
MikeMcQuaid: `if fn&.file?` would be equivalent and preferable 👍

Library/Homebrew/metaformula.rb
gigaflw: Oh. That's true. Then how about using caller to detect who calls the function ?
```ruby
if come_from_formula_file(caller(0)[1]) then
   block_it
else
   pass_through
end
```

MikeMcQuaid: Feels like a bit of a mess/hack. Probably better to define each of these methods separately.

gigaflw: It seem we still need to tell the caller even if defining these methods separately. So now I block all calls with arg or block. Do you think it is ok?

MikeMcQuaid: @gigaflw I still think handling each individual DSL method by itself would be cleaner; I don't think a one-size fits all makes sense given that e.g. `head`, `bottle` and `patch` take single arguments (as above) but `stable` does not.

reitermarkus:
Just a thought:

Since `MetaFormula` is such a slimmed-down version of `Formula`, maybe it would make more sense to make it a separate class rather than sub-class it, since there are *way* too many to special-case. And then maybe sub-class both from an `AbstractFormula` class which contains the shared methods.

gigaflw:
@reitermarkus Yes I've also though of that. But I am worried about modifying the existing code too much and possibly introducing bugs. Subclassing changes things less.
And the second thing is the boundary of `MetaFormula` is yet to be very clear. I haven't come up with a detailed list on which function it should contain and which should not.
So maybe we should keep implementing its functionality first and then see if there is a chance ?

MikeMcQuaid:
> Yes I've also though of that. But I am worried about modifying the existing code too much and possibly introducing bugs. Subclassing changes things less.

I think this is a reasonable argument. Again, though, I don't think `disable_dsl` is the correct approach to doing so. Make `MetaFormula` verbose for now and override methods individually and if we can extract a better version from that then: great.

gigaflw:
Is it ok now?  @MikeMcQuaid

MikeMcQuaid:
> Is it ok now? @MikeMcQuaid

Yeh, I prefer this approach to the `disable_dsl` 👍

gigaflw:
I added cask dependency for metaformula. Now a new dsl  `depends_on_cask` is feasible
```ruby
class MetaFoo < Metaformula
    depends_on_cask "some_cask"   # this will install cask "some_cask"
end
```

Library/Homebrew/metaformula.rb
```diff
@@ -61,5 +61,24 @@ def bottle
...
+    # A new DSL introduced in {MetaFormula} is {.depends_on_cask}
+    # <pre>
+    #   depends_on_cask "some_cask"
+    # </pre>
+    #
+    # this is implicitly equivalent to
+    # `depends_on :__meta_cask_req_some_cask`
+    # And this special prefix will help the dep to be parsed into {CaskRequirement}
+    # The special prefix is defined in {.cask_req_prefix}
+    def depends_on_cask(cask)
```
mistydemeo: Instead of making this a new DSL method, I wonder whether it might be a good idea to make this a new symbol spec like the other ones that currently exist. For example, something like `depends_on :cask => "name"`. That way you would need less special handling, and it could be managed in the same case statement that the rest of `parse_symbol_spec` uses. What do you think, @GauthamGoli?

gigaflw:
Here, in order to install casks, I used `system(HOMEBREW_BREW_FILE, "cask", "install", XXX)`. Is it a good idea to use `system` function ? Or, any advice for a better solution ?

Library/Homebrew/metaformula.rb
gigaflw: True, I think that may be better solution.

ilovezfs: cask vs. requirement name collisions
```
depends_on :java
```

also cask @ formula naming would be impossible since @ isn't a valid character in ruby symbols

ilovezfs: FWIW,
```
depends_on :cask => ["name", :build]
```
strikes me as a little weird, which is a technical term.

ilovezfs: >also cask @ formula naming would be impossible since @ isn't a valid character in ruby symbols

also `-` isn't allowed, which is used extensively in cask names already

mistydemeo: > also cask @ formula naming would be impossible since @ isn't a valid character in ruby symbols

Not the prettiest syntax, but you can put whatever in a symbol via `:"@this is a real symbol!"`

But as you pointed out there are some ambiguities in using symbols this way, so we shouldn't use that for the cask specification.

ilovezfs: TIL

MikeMcQuaid:
> Here, in order to install casks, I used system(HOMEBREW_BREW_FILE, "cask", "install", XXX). Is it a good idea to use system function ? Or, any advice for a better solution ?

@reitermarkus Will advise on the best method but I think calling Ruby directly rather than a `system` call may be better.

reitermarkus:
> @reitermarkus Will advise on the best method but I think calling Ruby directly rather than a system call may be better.

Take a look at `Hbc::Installer` in `cask/lib/hbc/installer.rb`.

Library/Homebrew/formula_installer.rb
```diff
@@ -379,6 +379,7 @@ def check_conflicts
   # Compute and collect the dependencies needed by the formula currently
   # being installed.
   def compute_dependencies
+    install_cask_requirements
```
MikeMcQuaid: I don't think installing casks should happen in a method named `compute_dependencies`. Additionally, requirements already can have `cask`s so it would be good to be able to map those casks onto casks to install.

Library/Homebrew/metaformula.rb
```diff
@@ -0,0 +1,68 @@
...
+  # Metaformula is not designed to be bottled for now
+  def bottle_disabled?
+    true
+  end
+
+  class << self
+    # DSL not designed to be used in MetaFormula for now are disabled to avoid undefined behaviors
+    # This is subject to change if certain functions should be assigned to them.
+
+    # {.stable}, {.devel} and {.head} only serve as attribute readers in {MetaFormula}
```
MikeMcQuaid: Should `devel` and `head` not maybe be completely disabled for MetaFormulae?

gigaflw: These methods are called in the initializer of Formula
```ruby
```

gigaflw: These methods are called as soon as in the initializer of Formula :
```ruby
def initializer
    ...
    spec_eval :stable
    spec_eval :devel
    spec_eval :head
end
```
And the return value is supposed to have `{.url}` and thus not to be `nil` (and thus probably can not be disabled)
```ruby
def spec_eval(name)
	spec = self.class.send(name)
	return unless spec.url
end
```
So maybe best methods I can come up with is to return a default value and prohibit others from passing a block to them, I think?

Library/Homebrew/formula_installer.rb
gigaflw: Yes, I will move that the `install` method.
And I'm sorry, but does mapping the casks mean to store the name of casks to install into the variable `@cask` of a requirement instance? This is already implemented like this, in the new subclass `CaskRequirement`. @MikeMcQuaid

MikeMcQuaid: > And I'm sorry, but does mapping the casks mean to store the name of casks to install into the variable @cask of a requirement instance?

I'm not sure I totally understand but yes it means making use of the already existing `cask` in requirements e.g. https://github.com/Homebrew/brew/blob/master/Library/Homebrew/requirements/x11_requirement.rb#L8

Library/Homebrew/metaformula.rb
MikeMcQuaid: It may be worth instead having a custom initialiser?

Library/Homebrew/formula_installer.rb
gigaflw: Yes, that's similar. `CaskRequirement` will set `@cask` according to the initialization argument.

gigaflw:
Here I added an new option to limit the use of metaformula like this:
```ruby
brew install --allow-metaformula XXX  # will abort if XXX is a metaformula and the option missing
```
An alternative to this is to set a env variable, but I think maybe a new option is more explicit. May I ask for your advice? @MikeMcQuaid

MikeMcQuaid:
> An alternative to this is to set a env variable, but I think maybe a new option is more explicit. May I ask for your advice? @MikeMcQuaid
> 
> Or, maybe keep both like brew upgrade --cleanup and HOMEBREW_UPGRADE_CLEANUP ? It is effective when any of the two is presented.

I'm not sure I understand the question?

gigaflw:
Oh, sorry for that. I mean it is not good if the user carelessly install casks in metaformulae because casks are possibly large. 
So maybe we should use certain method to limit the use of metaformula.
We can do so by requiring users to pass an extra option `--allow-metaformula` from command line every time they want to upgrade/install metaformulae. 
Or, we can set a new environment variable like `HOMEBREW_ALLOW_METAFORMULA`. Metaformulae can be installed only when this variable is set.
Or, we can use both.
Which one seems to be better?
@MikeMcQuaid

MikeMcQuaid:
> Which one seems to be better?

The `--` flag is fine for now 👍

Library/Homebrew/requirements/cask_requirement.rb
```diff
@@ -0,0 +1,71 @@
+require "requirement"
+
+$LOAD_PATH.unshift("#{HOMEBREW_LIBRARY_PATH}/cask/lib")
```
reitermarkus: This is not needed anymire, this should always be in the path now.

Library/Homebrew/requirements/cask_requirement.rb
```diff
@@ -0,0 +1,71 @@
...
+require "hbc"
```
reitermarkus: Try to avoid this and require only specific files.

gigaflw: This happens because I am working on a kind of outdated fork. I will update them soon, thanks!

gigaflw:
I will divide this into several smaller commits. Close for now.

PR #4328, opened 2018-06-11
[WIP] formula_installer: Don't check devtools if pouring

woodruffw:

-----

This *should* stop Homebrew from nagging users about their outdated Xcode installation during bottle-only installs. I'll be able to test it locally in a few days (my mac is currently packed away).

Some notes:
* This fixes the current behavior by moving the development tools check out of `perform_preinstall_checks` and into `FormulaInstaller#install`. This should be fine, although it does cause us to run the check *N + 1* times for 1 formula and  *N* dependencies.
* Some users may prefer this behavior, but it may not be beneficial in the long term: users *should* be nagged about the presence of unsupported Xcode installs, as doing so lessens the amount of weird setup triaging we have to do. Warning the user instead of totally failing the install might be a workable middle ground.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -13,6 +13,7 @@ require "emoji"
 require "development_tools"
 require "cache_store"
 require "linkage_checker"
+require "install"
 
 class FormulaInstaller
   include FormulaCellarChecks
@@ -220,7 +221,7 @@ class FormulaInstaller
 
   def install
     if !formula.bottle_unneeded? && !pour_bottle? && DevelopmentTools.installed?
-      Install.check_development_tools
+      Homebrew::Install.check_development_tools
     end
 
     # not in initialize so upgrade can unlink the active keg before calling this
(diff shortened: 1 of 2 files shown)


Library/Homebrew/formula_installer.rb
```diff
@@ -219,6 +219,10 @@ def build_bottle_postinstall
   end
 
   def install
+    if DevelopmentTools.installed? && !pour_bottle?
```
ilovezfs: any appreciable difference in speed depending on the order of those conditions?

ilovezfs:
We may also be able to skip the check if it's bottle unneeded.

Library/Homebrew/formula_installer.rb
MikeMcQuaid: Yeh, the latter is likely to be a bit faster so probably preferable?

woodruffw:
This should be good to go, pinging @ilovezfs and @MikeMcQuaid for final comments (if any).

ilovezfs:
Thanks @woodruffw! 🏆

PR #4334, opened 2018-06-12
Mojave: require CLT header package

mistydemeo:

-----

As of Xcode 10, the CLT no longer installs the traditional Unix header layout by default. The CLT SDK is the only location in which headers are located. There is, however, a separate header package which can be installed which restores the traditional Unix-style headers. I've submitted two PRs to address this situation, with conflicting solutions - we should merge only one of the two.

I haven't added tests for the untested files yet, but I'll do that once we pick which of the two solutions we want to go for.

This PR attempts to retain the status quo by making the CLT header package mandatory. It makes the following changes:

* Allows `MacOS::CLT` to detect whether the header package is relevant, whether it's installed, and whether it matches the installed CLT.
* Prints header package information as a part of `brew --config`.
* Adds a doctor check, and makes the header package mandatory for installing software in a CLT-only configuration.

See #4335 for a PR which uses the CLT SDKs instead.

### DIFF
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -196,14 +196,30 @@ module OS
 
       STANDALONE_PKG_ID = "com.apple.pkg.DeveloperToolsCLILeo".freeze
       FROM_XCODE_PKG_ID = "com.apple.pkg.DeveloperToolsCLI".freeze
-      MAVERICKS_PKG_ID = "com.apple.pkg.CLTools_Executables".freeze
+      # EXECUTABLE_PKG_ID now means two things:
+      # 1. The original Mavericks CLT package ID, and
+      # 2. The additional header package included in Mojave.
+      EXECUTABLE_PKG_ID = "com.apple.pkg.CLTools_Executables".freeze
       MAVERICKS_NEW_PKG_ID = "com.apple.pkg.CLTools_Base".freeze # obsolete
       PKG_PATH = "/Library/Developer/CommandLineTools".freeze
+      HEADER_PKG_PATH = "/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_:macos_version.pkg".freeze
 
       # Returns true even if outdated tools are installed, e.g.
       # tools from Xcode 4.x on 10.9
       def installed?
         !version.null?
+      end
+
+      def separate_header_package?
+        MacOS.version >= :mojave
+      end
+
+      def headers_installed?
+        if !separate_header_package?
+          installed?
+        else
+          headers_version == version
+        end
       end
 
       def update_instructions
@@ -282,6 +298,18 @@ module OS
         end
       end
 
+      # Version string of the header package, which is a
+      # separate package as of macOS 10.14.
+      def headers_version
+        if !separate_header_package?
+          version
+        else
+          @header_version ||= MacOS.pkgutil_info(EXECUTABLE_PKG_ID)[/version: (.+)$/, 1]
+          return ::Version::NULL unless @header_version
+          ::Version.new(@header_version)
+        end
+      end
+
       def detect_version
         # CLT isn't a distinct entity pre-4.3, and pkgutil doesn't exist
         # at all on Tiger, so just count it as installed if Xcode is installed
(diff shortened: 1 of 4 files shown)


ilovezfs:
👍

mistydemeo:
> Would it be possible/practical to use /usr/sbin/installer to install this automatically?

We can do that in the installer! I'll submit a PR for the installer if we decide to go with this branch. Once we're in Homebrew land though, we're no longer able to `sudo` on behalf of the user.

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -282,6 +298,17 @@ def version
...
       end
 
+      # Version string of the header package, which is a
+      # separate package as of macOS 10.14.
+      def headers_version
+        if !separate_header_package?
+          version
+        else
+          @header_version ||= MacOS.pkgutil_info(EXECUTABLE_PKG_ID)[/version: (.+)$/, 1]
+          @header_version.nil? ? ::Version::NULL : ::Version.new(@header_version)
```
MikeMcQuaid: I'd make this an `if` block or early `return` as the ternary plus `::` makes this hard to mentally parse.

MikeMcQuaid:
Thanks for the PR and write-up @mistydemeo!

> I've submitted two PRs to address this situation, with conflicting solutions - we should merge only one of the two.

Could merge this PR minus the doctor check immediately as that part of the diff is present in both PRs.

> > Would it be possible/practical to use /usr/sbin/installer to install this automatically?
>
> We can do that in the installer! I'll submit a PR for the installer if we decide to go with this branch. Once we're in Homebrew land though, we're no longer able to sudo on behalf of the user.

This is my main concern with this approach; unless it's possible to auto-install this stuff with e.g. `xcode-select --install` or similar it's going to be pretty painful to require users to fiddle around in their browser to download and install the right package.

That said, I think given we've just had a single developer beta and this has a much slimmer diff than https://github.com/Homebrew/brew/pull/4335 it's the right approach until we're closer to a Mojave release (or at least we've got some non-developer betas); I'd rather avoid monkeying with superenv if we can avoid it at all. 

If people don't strongly disagree then I'd suggest: merge this PR pretty much as-is and then rebase, keep open and consider https://github.com/Homebrew/brew/pull/4335 further down the line (removing the doctor check in the process) if we don't have a pleasant user experience to autoinstall these headers.

ilovezfs:
The pkg file itself gets installed along with the rest of the CLT, so it should be a question of printing "Please run ..."

```
  installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_:macos_version.pkg
```
or wrap that with a `brew install-clt-headers`.

MikeMcQuaid:
> The pkg file itself gets installed along with the rest of the CLT, so it should be a question of printing "Please run ..."

Cool. I'm definitely 👍 on this approach for now but I still think it's a step back having to get users to run Yet Another Command to have Homebrew work on 10.14.

ilovezfs:
I'm sure your sympathy will be noted by the complainers.

MikeMcQuaid:
Agreed, thanks again @mistydemeo!

jeremyhu:
*Please don't do this*.

This package is going away.  We will not be supporting it in the future.  Please remove your dependency on it.  It is being provided as an optional package as a crutch for you to use if you need to while working through issues (and filing radars if yo need support from Apple).  Simply requiring your users to install that package is not a long-term solution.

MikeMcQuaid:
@jeremyhu Homebrew lead maintainer here 👋. Thanks for chiming in and providing that information. I'd really appreciate if we can get a more formal and private communication channel (e.g. private mailing list or Slack channel) with Apple for such things (as currently we have none) and to help us get access to the Mojave beta which we do not currently have access to. You can contact me at mike@mikemcquaid.com to discuss this further. Thanks!

mistydemeo:
> This package is going away. We will not be supporting it in the future.

Is there information on this and any planned schedule? The Xcode 10 release notes don't say anything about this package being removed in the future, and when I asked about the headers in a radar I was directed to install this package.

MikeMcQuaid:
@jeremyhu Would love to get some answers on the above either publicly or privately or be directed to someone who could provide them.

jeremyhu:
@MikeMcQuaid Sorry for missing this earlier.  Let's followup in email.
@mistydemeo It should be in the release notes.  I need to check.  Can you point me to the radar you filed? I'd like to make sure messaging is clear in responses to future radars and want to followup on what was communicated.

mistydemeo:
@jeremyhu It was `rdar://40833474`.

Can you please include me on the email?

jeremyhu:
The release note says,

> Command Line Tools
> The Command Line Tools package installs the macOS system headers inside the macOS SDK. Software that compiles with the installed tools will search for macOS headers at the standard include path:
> /Applications/Xcode-
> beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sd
> k/usr/include
> For legacy software that looks for the macOS headers in the base system under /usr/include, please install the package file located at:
> /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg To make sure that you are using the intended version of the command line tools, run xcode-select -s
> <path to Xcode> after installing.

which I admit is not explicitly clear on the deprecation.

PR #4335, opened 2018-06-12
Mojave: use the CLT SDK where necessary

mistydemeo:

-----

As of Xcode 10, the CLT no longer installs the traditional Unix header layout by default. The CLT SDK is the only location in which headers are located. There is, however, a separate header package which can be installed which restores the traditional Unix-style headers. I've submitted two PRs to address this situation, with conflicting solutions - we should merge only one of the two.

I haven't added tests for the untested files yet, but I'll do that once we pick which of the two solutions we want to go for.

This PR adapts to the new default CLT layout by adapting our existing Xcode SDK code. The existing `SDKLocator` has been refactored into Xcode and CLT SDK locators. The new CLT SDK locator still returns `nil` on CLT-only systems pre-10.14, but it returns the path to the CLT SDK where it's relevant. I've also updated several places which called `MacOS.sdk_path` which now incorrectly assume it's irrelevant on CLT-only systems, and made them check for the presence of `MacOS.sdk_path` instead of hardcoding a CLT check.

See #4334 for a solution which mandates having the header package installed instead

### DIFF
--- a/Library/Homebrew/extend/os/mac/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/std.rb
@@ -27,8 +27,8 @@ module Stdenv
 
     append_path "ACLOCAL_PATH", "#{MacOS::X11.share}/aclocal"
 
-    if MacOS::XQuartz.provided_by_apple? && MacOS.sdk_path
-      append_path "CMAKE_PREFIX_PATH", "#{MacOS.sdk_path}/usr/X11"
+    if MacOS::XQuartz.provided_by_apple? && MacOS.sdk_path_if_needed
+      append_path "CMAKE_PREFIX_PATH", "#{MacOS.sdk_path_if_needed}/usr/X11"
     end
 
     append "CFLAGS", "-I#{MacOS::X11.include}" unless MacOS::CLT.installed?
@@ -93,7 +93,7 @@ module Stdenv
     delete("CPATH")
     remove "LDFLAGS", "-L#{HOMEBREW_PREFIX}/lib"
 
-    return unless (sdk = MacOS.sdk_path(version))
+    return unless (sdk = MacOS.sdk_path_if_needed(version))
     delete("SDKROOT")
     remove_from_cflags "-isysroot #{sdk}"
     remove "CPPFLAGS", "-isysroot #{sdk}"
@@ -115,7 +115,7 @@ module Stdenv
     self["CPATH"] = "#{HOMEBREW_PREFIX}/include"
     prepend "LDFLAGS", "-L#{HOMEBREW_PREFIX}/lib"
 
-    return unless (sdk = MacOS.sdk_path(version))
+    return unless (sdk = MacOS.sdk_path_if_needed(version))
     # Extra setup to support Xcode 4.3+ without CLT.
     self["SDKROOT"] = sdk
     # Tell clang/gcc where system include's are:
@@ -132,7 +132,7 @@ module Stdenv
 
   # Some configure scripts won't find libxml2 without help
   def libxml2
-    if !MacOS.sdk_path
+    if !MacOS.sdk_path_if_needed
       append "CPPFLAGS", "-I/usr/include/libxml2"
     else
       # Use the includes form the sdk
--- a/Library/Homebrew/extend/os/mac/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/std.rb
@@ -27,7 +27,7 @@ module Stdenv
 
     append_path "ACLOCAL_PATH", "#{MacOS::X11.share}/aclocal"
 
-    if MacOS::XQuartz.provided_by_apple? && !MacOS::CLT.installed?
+    if MacOS::XQuartz.provided_by_apple? && MacOS.sdk_path
       append_path "CMAKE_PREFIX_PATH", "#{MacOS.sdk_path}/usr/X11"
     end
 
@@ -93,7 +93,7 @@ module Stdenv
     delete("CPATH")
     remove "LDFLAGS", "-L#{HOMEBREW_PREFIX}/lib"
 
-    return unless (sdk = MacOS.sdk_path(version)) && !MacOS::CLT.installed?
+    return unless (sdk = MacOS.sdk_path(version))
     delete("SDKROOT")
     remove_from_cflags "-isysroot #{sdk}"
     remove "CPPFLAGS", "-isysroot #{sdk}"
@@ -115,7 +115,7 @@ module Stdenv
     self["CPATH"] = "#{HOMEBREW_PREFIX}/include"
     prepend "LDFLAGS", "-L#{HOMEBREW_PREFIX}/lib"
 
-    return unless (sdk = MacOS.sdk_path(version)) && !MacOS::CLT.installed?
+    return unless (sdk = MacOS.sdk_path(version))
     # Extra setup to support Xcode 4.3+ without CLT.
     self["SDKROOT"] = sdk
     # Tell clang/gcc where system include's are:
@@ -132,7 +132,7 @@ module Stdenv
 
   # Some configure scripts won't find libxml2 without help
   def libxml2
-    if MacOS::CLT.installed?
+    if !MacOS.sdk_path
       append "CPPFLAGS", "-I/usr/include/libxml2"
     else
       # Use the includes form the sdk
--- a/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
@@ -113,7 +113,6 @@ module Superenv
   def setup_build_environment(formula = nil)
     generic_setup_build_environment(formula)
     self["HOMEBREW_SDKROOT"] = effective_sysroot
-    self["SDKROOT"] = MacOS.sdk_path if MacOS.sdk_path
 
     # Filter out symbols known not to be defined since GNU Autotools can't
     # reliably figure this out with Xcode 8 and above.
--- a/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
@@ -58,7 +58,7 @@ module Superenv
   def homebrew_extra_library_paths
     paths = []
     if compiler == :llvm_clang
-      if !MacOS.sdk_path
+      if !MacOS.sdk_path_if_needed
         paths << "/usr/lib"
       else
         paths << "#{MacOS.sdk_path}/usr/lib"
@@ -102,7 +102,7 @@ module Superenv
   end
 
   def effective_sysroot
-    MacOS.sdk_path&.to_s
+    MacOS.sdk_path_if_needed&.to_s
   end
 
   def set_x11_env_if_installed
--- a/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/mac/extend/ENV/super.rb
@@ -58,7 +58,7 @@ module Superenv
   def homebrew_extra_library_paths
     paths = []
     if compiler == :llvm_clang
-      if MacOS::CLT.installed?
+      if !MacOS.sdk_path
         paths << "/usr/lib"
       else
         paths << "#{MacOS.sdk_path}/usr/lib"
@@ -102,7 +102,7 @@ module Superenv
   end
 
   def effective_sysroot
-    MacOS.sdk_path.to_s if MacOS::Xcode.without_clt?
+    MacOS.sdk_path&.to_s
   end
 
   def set_x11_env_if_installed
@@ -113,7 +113,7 @@ module Superenv
   def setup_build_environment(formula = nil)
     generic_setup_build_environment(formula)
     self["HOMEBREW_SDKROOT"] = effective_sysroot
-    self["SDKROOT"] = MacOS.sdk_path if MacOS::Xcode.without_clt?
+    self["SDKROOT"] = MacOS.sdk_path if MacOS.sdk_path
 
     # Filter out symbols known not to be defined since GNU Autotools can't
     # reliably figure this out with Xcode 8 and above.
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -85,7 +85,7 @@ module OS
     #      specifically been requested according to the rules above.
 
     def sdk(v = nil)
-      @locator ||= if Xcode.without_clt?
+      @locator ||= if Xcode.installed?
         XcodeSDKLocator.new
       else
         CLTSDKLocator.new
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -91,17 +91,7 @@ module OS
         CLTSDKLocator.new
       end
 
-      begin
-        sdk = if v.nil?
-          (Xcode.version.to_i >= 7) ? @locator.latest_sdk : @locator.sdk_for(version)
-        else
-          @locator.sdk_for v
-        end
-      rescue BaseSDKLocator::NoSDKError
-        sdk = @locator.latest_sdk
-      end
-      # Only return an SDK older than the OS version if it was specifically requested
-      sdk if v || (!sdk.nil? && sdk.version >= version)
+      @locator.sdk_if_applicable(v)
     end
 
     # Returns the path to an SDK or nil, following the rules set by #sdk.
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -110,6 +110,11 @@ module OS
       s&.path
     end
 
+    def sdk_path_if_needed(v = nil)
+      return if !MacOS::Xcode.installed? && MacOS::CLT.separate_header_package?
+      sdk_path(v)
+    end
+
     # See these issues for some history:
     # https://github.com/Homebrew/legacy-homebrew/issues/13
     # https://github.com/Homebrew/legacy-homebrew/issues/41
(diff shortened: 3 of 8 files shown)


Library/Homebrew/extend/os/mac/extend/ENV/std.rb
```diff
@@ -27,7 +27,7 @@ def x11
 
     append_path "ACLOCAL_PATH", "#{MacOS::X11.share}/aclocal"
 
-    if MacOS::XQuartz.provided_by_apple? && !MacOS::CLT.installed?
+    if MacOS::XQuartz.provided_by_apple? && !MacOS.sdk_path.nil?
```
MikeMcQuaid: `&& MacOS.sdk_path`

Library/Homebrew/extend/os/mac/extend/ENV/std.rb
```diff
@@ -132,7 +132,7 @@ def macosxsdk(version = MacOS.version)
 
   # Some configure scripts won't find libxml2 without help
   def libxml2
-    if MacOS::CLT.installed?
+    if MacOS.sdk_path.nil?
```
MikeMcQuaid: `if !MacOS.sdk_path`

Library/Homebrew/extend/os/mac/extend/ENV/super.rb
```diff
@@ -58,7 +58,7 @@ def homebrew_extra_isystem_paths
   def homebrew_extra_library_paths
     paths = []
     if compiler == :llvm_clang
-      if MacOS::CLT.installed?
+      if MacOS.sdk_path.nil?
```
MikeMcQuaid: `if !MacOS.sdk_path`

Library/Homebrew/extend/os/mac/extend/ENV/super.rb
```diff
@@ -102,7 +102,7 @@ def determine_cccfg
   end
 
   def effective_sysroot
-    MacOS.sdk_path.to_s if MacOS::Xcode.without_clt?
+    MacOS.sdk_path.nil? ? nil : MacOS.sdk_path.to_s
```
MikeMcQuaid: `MacOS.sdk_path&.to_s`

Library/Homebrew/extend/os/mac/extend/ENV/super.rb
```diff
@@ -113,7 +113,7 @@ def set_x11_env_if_installed
   def setup_build_environment(formula = nil)
     generic_setup_build_environment(formula)
     self["HOMEBREW_SDKROOT"] = effective_sysroot
-    self["SDKROOT"] = MacOS.sdk_path if MacOS::Xcode.without_clt?
+    self["SDKROOT"] = MacOS.sdk_path unless MacOS.sdk_path.nil?
```
MikeMcQuaid: `if MacOS.sdk_path`

Library/Homebrew/os/mac/xcode.rb
```diff
@@ -282,6 +298,17 @@ def version
...
       end
 
+      # Version string of the header package, which is a
+      # separate package as of macOS 10.14.
+      def headers_version
+        if !separate_header_package?
+          version
+        else
+          @header_version ||= MacOS.pkgutil_info(EXECUTABLE_PKG_ID)[/version: (.+)$/, 1]
+          @header_version.nil? ? ::Version::NULL : ::Version.new(@header_version)
```
MikeMcQuaid: I'd make this an `if` block or early `return` as the ternary plus `::` makes this hard to mentally parse.

MikeMcQuaid:
Have posted thoughts on https://github.com/Homebrew/brew/pull/4334#issuecomment-396850508

ilovezfs:
If we end up going this direction – and, to be clear, I sure hope we don't – let's make sure this is merged _before_ bottling starts.

mistydemeo:
> let's make sure this is merged before bottling starts.

Agreed - this would have major implications for that, not to mention that some things will not build with either this or with the header package installed.

MikeMcQuaid:
> If we end up going this direction – and, to be clear, I sure hope we don't – let's make sure this is merged before bottling starts.

Also agreed here. I think we'll need to make a decision either way in that case but I also think it'll be a little while (at the very, very earliest at the first public beta).

Would be good to rebase this on `master` and have it waiting here for if/when that happens.

mistydemeo:
> Would be good to rebase this on master and have it waiting here for if/when that happens.

Rebased! Just to be clear, for when what specifically happens?

MikeMcQuaid:
> Just to be clear, for when what specifically happens?

When/if we decide that this PR is the approach we want to take. This may be because the update situation for the extra `/usr/include` package is undesirable or terrible or we just don't want to put users through having to install it longer-term.

Library/Homebrew/extend/os/mac/extend/ENV/super.rb
```diff
@@ -113,7 +113,7 @@ def set_x11_env_if_installed
   def setup_build_environment(formula = nil)
     generic_setup_build_environment(formula)
     self["HOMEBREW_SDKROOT"] = effective_sysroot
-    self["SDKROOT"] = MacOS.sdk_path if MacOS::Xcode.without_clt?
+    self["SDKROOT"] = MacOS.sdk_path if MacOS.sdk_path
```
ilovezfs: Setting SDKROOT is undesirable on older systems that have /usr/include available as we don't want to bring back the 10.11 clock_gettime and 10.12 futimens, etc. problems from the dead.

mistydemeo: Maybe `self["SDKROOT"] = MacOS.sdk_path if MacOS.sdk_path && !Mac::CLT.provides_headers?`?

ilovezfs: Yes, that should definitely be safer. I'm not sure sure if that's the behavior we'll end up wanting on Mojave, though.

ilovezfs: I think it would be great if we could avoid setting it on the backend at all, including in the no-usr-include and xcode-only cases, and leave it to formulae to set when it fixes something. When we were always setting it, conversations with upstream would often get confused because build success/failure that occurred when building within Homebrew would sometimes not occur outside of Homebrew unless SDKROOT was set.

theeternalsw0rd:
Also of note, the CLT includes the headers necessary to build perl modules in /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Perl, and the headers installer package that populates /usr/include does not install the headers into the appropriate locations under /System/Library/Perl/ either.

It seems to me that perl detects the headers based on its own install directory, so it may be the situation where it is necessary to use a replacement perl where the paths can be controlled; or maybe a way at the formula level when perl Makefile.pl gets called to tell it where to find the headers. But I could be wrong and maybe this merge alone would fix it. I would be happy to do some testing before or after merging in this or some other pull request.

If you want to know a package currently broken by this to test yourself, you can use exiftool.

MikeMcQuaid:
@mistydemeo @ilovezfs What's the latest here? Would be good to get this merged so we can get some Mojave CI up and running with this. Thanks folks ❤️

Library/Homebrew/os/mac/sdk.rb
```diff
@@ -55,5 +52,43 @@ def sdk_paths
...
+      # While CLT SDKs existed prior to Xcode 10, those packages also
+      # installed a traditional Unix-style header layout and we prefer
+      # using that
+      # As of Xcode 10, the Unix-style headers are installed via a
+      # separate package, so we can't rely on their being present.
+      # This will only look up SDKs on Xcode 10 or newer, and still
+      # return nil SDKs for Xcode 9 and older.
+      def sdk_prefix
+        @sdk_prefix ||= begin
+          if !CLT.separate_header_package?
```
ilovezfs: I don't think we should favor the header package even when installed because it's missing stuff.

ilovezfs:
What do we anticipate the replacement for usage such as
```
      sdk = MacOS::CLT.installed? ? "" : MacOS.sdk_path
      ENV["LIBEDIT_CFLAGS"] = "-I#{sdk}/usr/include -I#{sdk}/usr/include/editline"
```
will be, bearing in mind the header package is apparently missing some important items?

MikeMcQuaid:
> sdk = MacOS::CLT.installed? ? "" : MacOS.sdk_path
>       ENV["LIBEDIT_CFLAGS"] = "-I#{sdk}/usr/include -I#{sdk}/usr/include/editline"

I'd say something like `sdk = MacOS.sdk_path_if_needed` which returns a new value.

mistydemeo:
> `sdk = MacOS::CLT.installed? ? "" : MacOS.sdk_path`

This could probably just be simplified to `MacOS.sdk_path.to_s`, now that I think about it. `nil.to_s` is already an empty string.

mistydemeo:
> @mistydemeo @ilovezfs What's the latest here?

I'm planning on doing some testing next week to remove always setting `SDKROOT` like @ilovezfs suggested. I'd like to see if it works reliably.

ilovezfs:
>This could probably just be simplified to MacOS.sdk_path.to_s, now that I think about it. nil.to_s is already an empty string.

Nope. For <= 10.12, MacOS.sdk_path would point to the Xcode headers when we specifically want /usr/include. That is why we currently must use

```
sdk = MacOS::CLT.installed? ? "" : MacOS.sdk_path
```
not
```
sdk = MacOS.sdk_path.to_s
```

Library/Homebrew/os/mac/sdk.rb
ilovezfs: Sorry if that wasn't clear. Since the "header package" is only a thing for Xcode 10, I was only talking about Xcode 10.

MikeMcQuaid: > I don't think we should favor the header package even when installed because it's missing stuff.

👍 

> Pre-Xcode 10, I think we should still use the headers in /usr. That's how we've done it before, and we might as well keep consistency there.

👍 

> so I think this ends up favoring /usr/include on 10.13 with Xcode 10/CLT 10. Is that intentional?

If it is: I don't think it should be. I think that should also check the Xcode version >= 10.

Library/Homebrew/extend/os/mac/extend/ENV/super.rb
mistydemeo: Updated to experimentally remove this line entirely. I'll do some testing to see how reliable it is.

Library/Homebrew/os/mac/sdk.rb
mistydemeo: > I think that should also check the Xcode version >= 10.

Fixed. This now favours the SDK with Xcode 10 on 10.13.

> I don't think we should favor the header package even when installed because it's missing stuff.

With the log update for `CLT.separate_header_package?` this will now only use the CLT SDK for Xcode 10 onwards, and it will always use it on Xcode 10.

mistydemeo:
> I'd say something like sdk = MacOS.sdk_path_if_needed which returns a new value.

Implemented this method, and updated a few bits of stdenv/superenv to use it.

ilovezfs:
On 10.11, this PR is causing`MacOS.sdk_path` to be `nil`, which is not correct.

We set
```
ENV["SDKROOT"] =  MacOS.sdk_path
```
in some formulae in order to get the Xcode sdk and prevent build failures.

mistydemeo:
> On 10.11, this PR is causingMacOS.sdk_path to be nil, which is not correct.

10.11 Xcode-only? I'll take a look at it.

ilovezfs:
No, that's just where I tested. I assume it occurs everywhere except Xcode 10 land.

mistydemeo:
> On 10.11, this PR is causing `MacOS.sdk_path` to be nil, which is not correct.

Can't repro on 10.11 Xcode without CLT. Just checking, are you saying that 10.11 with Xcode *and* the CLT, you'll want `MacOS.sdk_path` to return the Xcode SDK?

mistydemeo:
Pushed a commit which makes `MacOS.sdk_path` return the Xcode SDK in the following two cases:

1. Xcode and the CLT are installed, and no CLT SDK exists,
2. Xcode and the CLT are installed, and the CLT SDK exists.

These match current behaviour pre-Xcode 10, and will also make Xcode 10 consistent with that.

ilovezfs:
We'll likely need some way to refer to the CLT's sdk path, at least on >= Xcode/CLT 10, when both are installed to prevent the gcc bottle from ending up needing Xcode.

mistydemeo:
> We'll likely need some way to refer to the CLT's sdk path, at least on >= Xcode/CLT 10, when both are installed to prevent the gcc bottle from ending up needing Xcode.

I'll add that. It looks like at least some things will fail if `SDKROOT` is set to the CLT SDK when Xcode is installed; for example, this message from lxml in bzt:

```
xcodebuild: error: SDK "/Library/Developer/CommandLineTools/SDKs/MacOSX10.13.sdk" cannot be located.
clang: error: unable to find utility "clang", not a developer tool or in PATH
```

Note that that path exists, xcodebuild just isn't happy using it.

In happier news, many of the formulae which need an explicit `SDKROOT` now will no longer need to do so in this branch. We should be able to simplify things pretty nicely.

mistydemeo:
OK, updated to add methods to request the Xcode and CLT SDKs specifically. This also simplified the logic of `MacOS.sdk` by encapsulating that in the parent `SDKLocator` class instead.

Library/Homebrew/os/mac.rb
```diff
@@ -105,6 +100,11 @@ def sdk_path(v = nil)
       s&.path
     end
 
+    def sdk_path_if_needed(v = nil)
+      return nil if !MacOS::Xcode.installed? && MacOS::CLT.separate_header_package?
```
MikeMcQuaid: Could just be `return if`?

Library/Homebrew/os/mac/sdk.rb
```diff
@@ -28,17 +28,32 @@ def latest_sdk
         SDK.new v, path
       end
 
+      def sdk_if_applicable(v = nil)
+        begin
+          sdk = if v.nil?
+            (source_version.to_i >= 7) ? latest_sdk : sdk_for(OS::Mac.version)
```
MikeMcQuaid: Maybe make this a block `if` or `elsif` if that looks cleaner?

Library/Homebrew/os/mac/sdk.rb
```diff
@@ -28,17 +28,32 @@ def latest_sdk
         SDK.new v, path
       end
 
+      def sdk_if_applicable(v = nil)
+        begin
```
MikeMcQuaid: could `sdk = begin` to clean this all up a little?

Library/Homebrew/os/mac/sdk.rb
```diff
@@ -28,17 +28,32 @@ def latest_sdk
...
+          sdk = if v.nil?
+            (source_version.to_i >= 7) ? latest_sdk : sdk_for(OS::Mac.version)
+          else
+            sdk_for v
+          end
+        rescue BaseSDKLocator::NoSDKError
+          sdk = latest_sdk
+        end
+        # Only return an SDK older than the OS version if it was specifically requested
+        sdk if v || (!sdk.nil? && sdk.version >= OS::Mac.version)
```
MikeMcQuaid: An early `nil` return might be easier to parse here?

Library/Homebrew/os/mac/sdk.rb
```diff
@@ -55,5 +70,51 @@ def sdk_paths
...
+      end
+
+      def sdk_prefix
+        @sdk_prefix ||= begin
+          # Xcode.prefix is pretty smart, so let's look inside to find the sdk
+          sdk_prefix = "#{Xcode.prefix}/Platforms/MacOSX.platform/Developer/SDKs"
+          # Xcode < 4.3 style
+          sdk_prefix = "/Developer/SDKs" unless File.directory? sdk_prefix
+          # Finally query Xcode itself (this is slow, so check it last)
+          sdk_prefix = File.join(Utils.popen_read(DevelopmentTools.locate("xcrun"), "--show-sdk-platform-path").chomp, "Developer", "SDKs") unless File.directory? sdk_prefix
```
MikeMcQuaid: I know it's not you but any chance of splitting this onto a few lines?

Library/Homebrew/os/mac/sdk.rb
mistydemeo: I can make those changes, but this code hasn't actually changed - this is the exact same method that used to be in `MacOS.sdk`. It's just moved.

mistydemeo:
OK, updated! Confirmed working with 10.12 with Xcode and the CLT installed, and on 10.14 with the CLT installed.

mistydemeo:
> Thanks for cleaning up other people's code!

To be honest it's probably my own code from a few years ago... 😅

mistydemeo:
Updated the logic so that Perl on 10.11 reverts to the older behaviour:

```
Directories to use for library searches?
[/usr/local/lib /usr/include/libxml2 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../lib/clang/8.0.0/lib /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib /usr/lib]
```

I've added an extensive comment to `MacOS.sdk_path_if_needed` documenting the intended behaviour. It feels like this likely needs a unit test at this point.

MikeMcQuaid:
@mistydemeo Looks good to me @mistydemeo, thanks for all your work here. Merge when ready 👍

jnozsc:
just curious,  should https://github.com/Homebrew/brew/pull/4334 gets reverted, after this one gets merged?  thanks

mistydemeo:
Yes, this branch reverts the relevant parts of that branch.

PR #4348, opened 2018-06-15
Ignore any version of Go, when it is a dependency of a bottle

errordeveloper:

-----

As discussed in https://github.com/Homebrew/homebrew-core/pull/29040#issuecomment-397704661, this is needed for bottles that pin a specific version of Go as their dependency.

I'm happy to write tests, if that's required for a trivial change like this, and if there are tests covering this code path already.

### DIFF
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -284,8 +284,13 @@ module Homebrew
         end
 
         ignores = []
-        if f.deps.any? { |dep| dep.name =~ /^go(@[\d\.]+)?$/ }
-          ignores << %r{#{Regexp.escape(HOMEBREW_CELLAR)}/go(@[\d\.]+)?/[\d\.]+/libexec}
+        any_go_deps = f.deps.any? do |dep|
+          dep.name =~ Version.formula_optionally_versioned_regex(:go)
+        end
+        if any_go_deps
+          go_regex =
+            Version.formula_optionally_versioned_regex(:go, full: false)
+          ignores << %r{#{Regexp.escape(HOMEBREW_CELLAR)}/#{go_regex}/[\d\.]+/libexec}
         end
 
         relocatable = true
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -284,8 +284,8 @@ module Homebrew
         end
 
         ignores = []
-        if f.deps.any? { |dep| dep.name == "go" }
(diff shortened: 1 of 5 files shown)


errordeveloper:
cc @ilovezfs

ilovezfs:
This looks good to me although we may want to settle on exactly what regex to use for @ formulae. See also

https://github.com/Homebrew/brew/blob/32fa773cf87c744f2d630efa1d702880d34bc01d/Library/Homebrew/rubocops/lines_cop.rb#L293
https://github.com/Homebrew/brew/blob/32fa773cf87c744f2d630efa1d702880d34bc01d/Library/Homebrew/extend/os/mac/formula_cellar_checks.rb#L10

errordeveloper:
I haven't spent too much time on the regex, was really the simplest one
that worked for 'go' and 'go@1.9', but didn't work for 'go@' etc.

I think kibana one is a good match, let me know what your preference would
be.

Also, the test coverage

errordeveloper:
@ilovezfs personally, I don't see a good reason why my version is unsatisfactory, but maybe we should check with the maintainer(s) of go formula – WDYT?

errordeveloper:
@ilovezfs thank's for the review! how does merging work here?

ilovezfs:
I'll let some other maintainers comment on what regex they prefer but then we should be good to go. @woodruffw pointed out on Slack

>I'd prefer unescaped `@`, although the `lines_cop` regex is more correct strictly speaking
since the `formula_cellar_checks` one will match a nonsense version like `php5.`

So I suspect the one from kibana will be better (though without case-insentivity) so that we don't match nonsense versions like `"go@2."` Also, it would be best if the Cellar path worked for devel versions too like go 1.10rc2 (meaning the complete version may have alpha characters), although admittedly devel bottles aren't very common.

MikeMcQuaid:
I've made these regex consistent with an accessory function.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @errordeveloper!

errordeveloper:
🎉 thanks @ilovezfs & @MikeMcQuaid, this is awesome!

errordeveloper:
Am I understanding correct that for these change to take effect the `kubernetes-cli` (and any other formulae that depend on `go@1.*`) would have to get rebottled? Is there a way to do force the CI to do this before those formulae get updated or some change to the formula would be required?

ilovezfs:
fyi https://github.com/Homebrew/homebrew-core/commit/4fe6716ea22392f682e8d457cbd8173aaa03c692

so `brew reinstall kubernetes-cli` and you're good to ... go

errordeveloper:
brilliant, it worked!

PR #4357, opened 2018-06-19
Show tapped casks.

reitermarkus:

-----

Also show cask count when tapped in addition to `0 formulae`.

```
==> Tapping homebrew/cask
Tapped 0 formulae (4,083 files, 4MB)
```

### DIFF
--- a/Library/Homebrew/cmd/tap-info.rb
+++ b/Library/Homebrew/cmd/tap-info.rb
@@ -62,10 +62,10 @@ module Homebrew
         if tap.installed?
           info += tap.pinned? ? "pinned" : "unpinned"
           info += ", private" if tap.private?
-          if (contents = tap.contents).empty?
-            info += ", no commands/casks/formulae"
+          info += if (contents = tap.contents).empty?
+            ", no commands/casks/formulae"
           else
-            info += ", #{contents.join(", ")}"
+            ", #{contents.join(", ")}"
           end
           info += "\n#{tap.path} (#{tap.path.abv})"
           info += "\nFrom: #{tap.remote.nil? ? "N/A" : tap.remote}"
--- a/Library/Homebrew/cmd/tap-info.rb
+++ b/Library/Homebrew/cmd/tap-info.rb
@@ -62,13 +62,11 @@ module Homebrew
         if tap.installed?
           info += tap.pinned? ? "pinned" : "unpinned"
           info += ", private" if tap.private?
-          if (formula_count = tap.formula_files.size).positive?
-            info += ", #{Formatter.pluralize(formula_count, "formula")}"
+          if (contents = tap.contents).empty?
+            info += ", no commands/casks/formulae"
+          else
+            info += ", #{contents.join(", ")}"
           end
-          if (command_count = tap.command_files.size).positive?
-            info += ", #{Formatter.pluralize(command_count, "command")}"
-          end
-          info += ", no formulae/commands" if (formula_count + command_count).zero?
           info += "\n#{tap.path} (#{tap.path.abv})"
           info += "\nFrom: #{tap.remote.nil? ? "N/A" : tap.remote}"
         else
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -284,7 +284,7 @@ class Tap
 
     link_completions_and_manpages
 
-    formatted_contents = Formatter.enumeration(*contents)&.prepend(" ")
+    formatted_contents = Formatter.comma_and(*contents)&.prepend(" ")
     puts "Tapped#{formatted_contents} (#{path.abv})." unless quiet
     Descriptions.cache_formulae(formula_names)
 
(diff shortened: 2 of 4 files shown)


Library/Homebrew/tap.rb
```diff
@@ -284,8 +284,9 @@ def install(options = {})
 
     link_completions_and_manpages
 
-    formula_count = formula_files.size
-    puts "Tapped #{Formatter.pluralize(formula_count, "formula")} (#{path.abv})" unless quiet
+    casks = Formatter.pluralize(cask_files.count, "cask")
+    formulae = Formatter.pluralize(formula_files.count, "formula")
+    puts "Tapped #{formulae} and #{casks} (#{path.abv})." unless quiet
```
MikeMcQuaid: How about not outputting formulae or casks if they are zero? Just an idea (and I can be talked out of it)

apjanke: I like this idea. How about adding a command count, too? Like:

```
Tapped 3 casks.
Tapped 14 casks and 2 commands.
Untapped 2349 formulae.
Untapped 2673 formulae and 5 commands.
```

MikeMcQuaid: Yeh, that would be good 👍

reitermarkus:
Also added commands.

Library/Homebrew/utils/formatter.rb
```diff
@@ -108,4 +108,11 @@ def pluralize(count, singular, plural = nil, show_count: true)
 
     show_count ? "#{count} #{words}" : words
   end
+
+  def enumeration(*items)
```
MikeMcQuaid: `array_comma_and` or something to me would read a bit more obviously; I was a bit confused as to what `enumeration` meant.

Library/Homebrew/tap.rb
```diff
@@ -284,8 +284,8 @@ def install(options = {})
 
     link_completions_and_manpages
 
-    formula_count = formula_files.size
-    puts "Tapped #{Formatter.pluralize(formula_count, "formula")} (#{path.abv})" unless quiet
+    formatted_contents = Formatter.enumeration(*contents)&.prepend(" ")
```
MikeMcQuaid: The `&.prepend(" ")` is neat; I wonder if it's worth making this an optional parameter for `Formatter.enumeration`

Library/Homebrew/utils/formatter.rb
```diff
@@ -108,4 +108,11 @@ def pluralize(count, singular, plural = nil, show_count: true)
 
     show_count ? "#{count} #{words}" : words
   end
+
+  def comma_and(*items)
+    *items, last = items.map(&:to_s) # rubocop:disable Lint/ShadowedArgument, TODO: Remove when RuboCop 0.57.3 is released.
```
MikeMcQuaid: Is there an upstream bug you can link to here?

reitermarkus: Done.

Library/Homebrew/cmd/tap-info.rb
```diff
@@ -62,13 +62,11 @@ def print_tap_info(taps)
         if tap.installed?
           info += tap.pinned? ? "pinned" : "unpinned"
           info += ", private" if tap.private?
-          if (formula_count = tap.formula_files.size).positive?
-            info += ", #{Formatter.pluralize(formula_count, "formula")}"
+          if (contents = tap.contents).empty?
```
MikeMcQuaid: `info += if`?

Library/Homebrew/tap.rb
```diff
@@ -284,8 +284,8 @@ def install(options = {})
 
     link_completions_and_manpages
 
-    formula_count = formula_files.size
-    puts "Tapped #{Formatter.pluralize(formula_count, "formula")} (#{path.abv})" unless quiet
+    formatted_contents = Formatter.comma_and(*contents)&.prepend(" ")
```
MikeMcQuaid: Thoughts about the `prepend` being part of `comma_and`?

Library/Homebrew/cmd/tap-info.rb
reitermarkus: I think I had this before, but it failed because it calls `to_str` on `nil`.

Library/Homebrew/tap.rb
reitermarkus: I don't think it's worth it.

Library/Homebrew/cmd/tap-info.rb
reitermarkus: 👍

Library/Homebrew/tap.rb
MikeMcQuaid: My thinking was it's currently an accessory function where the only use is with the `prepend`?

reitermarkus: Still, I think it‘s not worth it:

```ruby
Formatter.comma_and(*contents)&.prepend(" ")
 ```

vs.

```ruby
Formatter.comma_and(*contents, prepend: " ")
 ```
and a more complex `comma_and`.

PR #4358, opened 2018-06-20
Fix SVN downloads of versioned formula by munging directory names

apjanke:

-----

The SubversionDownloadStrategy is broken for versioned formulae, because it includes an "@" in the resulting directory names, and "@" is a special character to Subversion. This PR fixes it by munging the cache and staging directory names to use "\_AT\_" instead of "@", avoiding the special character.

There's currently no test for the SubversionDownloadStrategy, and it doesn't look like download strategies are tested in the context of a `stage` operation. Suggestions welcome as to how to do a decent test on this.

###  Details

To reproduce, create a versioned formula that uses an SVN download as its main download, for example, `netpbm`. Do a `brew install -s` on it, and it'll error during the fetch step.

```
$ brew install -s octave-app/octave-app/netpbm@10.73.19                                                                                                                                                      master
==> Installing netpbm@10.73.19 from octave-app/octave-app
==> Cloning svn://svn.code.sf.net/p/netpbm/code/stable
Updating /Users/janke/Library/Caches/Homebrew/netpbm@10.73.19--svn
==> Checking out 3244
svn: E200009: '/Users/janke/Library/Caches/Homebrew/netpbm@10.73.19--svn': a peg revision is not allowed here
Error: Failed to download resource "netpbm@10.73.19"
Failure while executing: svn up -q /Users/janke/Library/Caches/Homebrew/netpbm@10.73.19--svn -r 3244
```

###  References

* https://github.com/silverstripe/silverstripe-framework/issues/4643
* https://jira.apache.org/jira/browse/SCM-859

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -665,6 +665,10 @@ class SubversionDownloadStrategy < VCSDownloadStrategy
     @url = @url.sub("svn+http://", "")
   end
 
+  def cache_filename
+    Resource.safe_download_name("#{name}--#{cache_tag}")
+  end
+
   def fetch
     clear_cache unless @url.chomp("/") == repo_url || quiet_system("svn", "switch", @url, cached_location)
     super
--- a/Library/Homebrew/resource.rb
+++ b/Library/Homebrew/resource.rb
@@ -66,8 +66,18 @@ class Resource
     name.tr("/", "-")
   end
 
+  def self.safe_download_name(str)
+    # "@" is a special character for Subversion. Keep it out of the cache dir name.
+    str.gsub("@", "-AT-")
+  end
+
   def download_name
-    name.nil? ? owner.name : "#{owner.name}--#{escaped_name}"
+    raw_name = if name
+      "#{owner.name}--#{escaped_name}"
+    else
+      owner.name
+    end
+    Resource.safe_download_name(raw_name)
   end
 
   def cached_download


Library/Homebrew/download_strategy.rb
```diff
@@ -665,6 +665,11 @@ def initialize(name, resource)
     @url = @url.sub("svn+http://", "")
   end
 
+  def cache_filename
+    # "@" is a special character for Subversion. Keep it out of the cache dir name.
+    "#{name.gsub("@", "_AT_")}--#{cache_tag}"
```
MikeMcQuaid: Do you think we need the `_`? I also wonder about using `-` instead.

Library/Homebrew/resource.rb
```diff
@@ -67,7 +67,8 @@ def escaped_name
   end
 
   def download_name
-    name.nil? ? owner.name : "#{owner.name}--#{escaped_name}"
+    # Must not have "@" in download_name because it's a special character to Subversion
+    (name.nil? ? owner.name : "#{owner.name}--#{escaped_name}").gsub("@", "_AT_")
```
MikeMcQuaid: `if name` would be a bit nicer (I know it's not your code). Would be good to reuse the `gsub`; perhaps a shared resource function?

Library/Homebrew/download_strategy.rb
apjanke: Switched to `-`. I like having some separator there for readability.

Library/Homebrew/resource.rb
apjanke: Refactored it to use an `if name`.

apjanke:
Ping! Any outstanding changes still needed on this? Would be nice to get this merged, since I'm actively encountering this error in a project.

apjanke:
No worries. Enjoy your vacation; this stuff can wait!

When you have time, I've pushed an amendment with a refactor that addresses your latest comments, consolidating the `gsub` into a single shared method.

Library/Homebrew/download_strategy.rb
```diff
@@ -665,6 +665,11 @@ def initialize(name, resource)
     @url = @url.sub("svn+http://", "")
   end
 
+  def cache_filename
+    # "@" is a special character for Subversion. Keep it out of the cache dir name.
```
MikeMcQuaid: Feels like this comment should be for safe download name

Library/Homebrew/resource.rb
```diff
@@ -66,8 +66,19 @@ def escaped_name
...
+    str.gsub("@", "-AT-")
+  end
+
   def download_name
-    name.nil? ? owner.name : "#{owner.name}--#{escaped_name}"
+    # Must not have "@" in download_name because it's a special character to Subversion
+    Resource.safe_download_name(
+      if name
+        "#{owner.name}--#{escaped_name}"
+      else
```
MikeMcQuaid: Feels weird having an if inside a function call. I’d assign a new variable outside it.

Library/Homebrew/download_strategy.rb
apjanke: Moved comment to `safe_download_name`.

Library/Homebrew/resource.rb
apjanke: Rewrote it to use a temporary variable. Moved & merged comment to `safe_download_name`.

apjanke:
Amended to address comments; `brew style` fixed.

Library/Homebrew/formula_installer.rb
```diff
@@ -727,6 +727,7 @@ def build
 
     formula.update_head_version
 
+    puts "Testing formula prefix for emptiness: formula.prefix='#{formula.prefix}'"
```
MikeMcQuaid: Oops

apjanke: Oops. Fixed.

Sorry for all the hassle here.

MikeMcQuaid:
Thanks @apjanke!

apjanke:
Yay, thank you! And you're welcome!

ilovezfs:
This seems to have changed the URLs that `brew mirror` uses for @ formulae.

Before revert:
```
iMac-TMP:Homebrew joe$ brew mirror imagemagick@6
==> Downloading https://dl.bintray.com/homebrew/mirror/imagemagick%406-6.9.10-3.tar.xz
Already downloaded: /Users/joe/Library/Caches/Homebrew/imagemagick-AT-6-6.9.10-3.tar.xz
==> Uploading to https://dl.bintray.com/homebrew/mirror/imagemagick-AT-6-6.9.10-3.tar.xz
^C
```

After revert:
```
iMac-TMP:Homebrew joe$ brew mirror imagemagick@6
==> Downloading https://dl.bintray.com/homebrew/mirror/imagemagick%406-6.9.10-3.tar.xz
Already downloaded: /Users/joe/Library/Caches/Homebrew/imagemagick@6-6.9.10-3.tar.xz
==> Uploading to https://dl.bintray.com/homebrew/mirror/imagemagick@6-6.9.10-3.tar.xz
```
CC @apjanke

apjanke:
Darn it!

Do the file upload names to Bintray matter? If so, best revert this PR, and I'll have to come up with a more detailed PR that tracks both original filenames and local filenames separately.

MikeMcQuaid:
I’d suggest tracking or calculating the original url filename in mirror as that’ll be the only place this matters.

ilovezfs:
>Do the file upload names to Bintray matter?

```
mirror "https://dl.bintray.com/homebrew/mirror/imagemagick-AT-6-6.9.10-3.tar.xz"
```

should work if the change is intentional.

Personally, I'm really unconvinced that having the @ formula names in the cache files be munged just for the sake of svn was a desirable change. Hunting for 

gcc@7-7.3.0.tar.xz
gcc@7--patch-863957f90a934ee8f89707980473769cff47ca0663c3906992da6afb242fb220.patch

is a lot nicer than
gcc-AT-7-7.3.0.tar.xz
gcc-AT-7--patch-863957f90a934ee8f89707980473769cff47ca0663c3906992da6afb242fb220.patch

Also does this invalidate cleanup? Is it still going to know how to cleanup the tarballs with the old names or did they all just get stranded forever?

MikeMcQuaid:
I didn’t review this sufficiently. I was under the impression from the PR description that this only affected SVN resources. I think it should be reverted and scoped just to them, unfortunately.

apjanke:
Okay, sorry. I will re-submit a new PR scoped only to SVN downloads when I'm able to figure out how to properly unit test it.

PR #4359, opened 2018-06-20
Add --display-times option to `install`, `upgrade`, and `reinstall`

apjanke:
- [X] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [X] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [X] Have you successfully run `brew style` with your changes locally?
- [X] Have you successfully run `brew tests` with your changes locally?

-----

Would you consider a `--display-build-times` option for `brew install` and `brew upgrade`? This could be helpful to developers who are wrangling big installation trees.

I'm working on a big build with lots of dependencies (like 40 minutes total), and I'd like to see which of the dependencies are taking a lot of compilation time, so I know where to spend my efforts at making relocatable bottles, removing dependencies, or something else. I don't see a good way to do this with a custom `brew` command, since there aren't public API hooks into the installation process.

Adds output like this:

```
$ brew install --display-build-times -s arpack cowsay libpng                                                   display-build-times
==> Downloading https://github.com/opencollab/arpack-ng/archive/3.6.0.tar.gz
...
==> make test
==> make install
🍺  /usr/local/Cellar/libpng/1.6.34: 26 files, 1.2MB, built in 33 seconds
==> Build times
arpack               28.326 s
cowsay               3.063 s
libpng               33.318 s
```

PR #4361, opened 2018-06-20
Display collected caveats at end of `install` or `upgrade`

### LINKED ISSUE #4057: Output all caveats at the end of a build
After `brew install` or `brew upgrade` multiple formulae we should group the caveats for all the formulae and output them at the end of the build.

apjanke:

-----

Fixes #4057.

This PR collects all caveats encountered during an `install`, `upgrade`, or `reinstall`, and outputs them together at the end of the build.

It also keeps the per-formula caveats output with each individual formula, so when a user is scanning through a long build output, they notice the caveats for each formula, instead of having to skip down to the end to check if there were caveats for that build.

Uses a generic message-collection design, per @ilovezfs's suggestion at https://github.com/Homebrew/brew/pull/4359#issuecomment-398627182, so this could be combined with a build times output, or other output that combines messages from multiple formulae across a multi-step operation.

Example:

```
$ brew install vf vice
==> Downloading https://github.com/glejeune/vf/archive/0.0.1.tar.gz
Already downloaded: /Users/janke/Library/Caches/Homebrew/vf-0.0.1.tar.gz
==> Caveats
To complete installation, add the following line to your shell's rc file:
  source /usr/local/opt/vf/vf.sh
==> Summary
🍺  /usr/local/Cellar/vf/0.0.1: 6 files, 14.2KB, built in 2 seconds
==> Downloading https://homebrew.bintray.com/bottles/vice-3.1_1.high_sierra.bott
Already downloaded: /Users/janke/Library/Caches/Homebrew/vice-3.1_1.high_sierra.bottle.1.tar.gz
==> Pouring vice-3.1_1.high_sierra.bottle.1.tar.gz
==> Caveats
Apps for these emulators have been installed to /usr/local/opt/vice.
==> Summary
🍺  /usr/local/Cellar/vice/3.1_1: 1,493 files, 143.6MB
==> Caveats
==> vf
To complete installation, add the following line to your shell's rc file:
  source /usr/local/opt/vf/vf.sh
==> vice
Apps for these emulators have been installed to /usr/local/opt/vice.
```

### DIFF
--- a/Library/Homebrew/global.rb
+++ b/Library/Homebrew/global.rb
@@ -1,6 +1,7 @@
 require "pathname"
 require "English"
 require "ostruct"
+require "messages"
 
 require "pp"
 require "extend/ARGV"
@@ -47,6 +48,10 @@ module Homebrew
 
     def args
       @args ||= OpenStruct.new
+    end
+
+    def messages
+      @messages ||= Messages.new
     end
 
     def raise_deprecation_exceptions?
--- a/Library/Homebrew/messages.rb
+++ b/Library/Homebrew/messages.rb
@@ -1,0 +1,31 @@
+# A Messages object collects messages that may need to be displayed together
+# at the end of a multi-step `brew` command run
+class Messages
+  attr_reader :caveats, :formula_count
+
+  def initialize
+    @caveats = []
+    @formula_count = 0
+  end
+
+  def record_caveats(f, caveats)
+    @caveats.push(formula: f.name, caveats: caveats)
+  end
+
+  def formula_installed(_f)
+    @formula_count += 1
+  end
+
+  def display_messages
+    display_caveats
+  end
+
+  def display_caveats
+    return if @formula_count <= 1
+    return if @caveats.empty?
+    oh1 "Caveats"
+    @caveats.each do |c|
+      ohai c[:formula], c[:caveats]
+    end
+  end
+end
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -249,6 +249,7 @@ module Homebrew
         Migrator.migrate_if_needed(f)
         install_formula(f)
       end
+      Homebrew.messages.display_messages
     rescue FormulaUnreadableError, FormulaClassUnavailableError,
            TapFormulaUnreadableError, TapFormulaClassUnavailableError => e
       # Need to rescue before `FormulaUnavailableError` (superclass of this)
--- a/Library/Homebrew/cmd/reinstall.rb
+++ b/Library/Homebrew/cmd/reinstall.rb
@@ -3,6 +3,7 @@
 
 require "formula_installer"
 require "development_tools"
+require "messages"
 
 module Homebrew
   module_function
@@ -18,6 +19,7 @@ module Homebrew
       Migrator.migrate_if_needed(f)
       reinstall_formula(f)
     end
+    Homebrew.messages.display_messages
   end
 
   def reinstall_formula(f)
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -21,6 +21,7 @@ require "install"
 require "formula_installer"
 require "cleanup"
 require "development_tools"
+require "messages"
 
 module Homebrew
   module_function
@@ -102,6 +103,7 @@ module Homebrew
         onoe "#{f}: #{e}"
       end
     end
+    Homebrew.messages.display_messages
   end
 
   def upgrade_formula(f)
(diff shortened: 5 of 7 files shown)


Library/Homebrew/messages.rb
```diff
@@ -0,0 +1,36 @@
...
+    return unless @formula_count > 1
+    return if @caveats.empty?
+    oh1 "Caveats"
+    @caveats.each do |c|
+      ohai c[0], c[1]
+    end
+  end
+end
+
+module Homebrew
```
MikeMcQuaid: I'd be tempted to put this in `global.rb`

Library/Homebrew/messages.rb
```diff
@@ -0,0 +1,36 @@
+# A Messages object collects messages that may need to be displayed together
+# at the end of a multi-step `brew` command run
+class Messages
```
MikeMcQuaid: Could we get some basic unit tests for this class?

Library/Homebrew/messages.rb
```diff
@@ -0,0 +1,36 @@
...
+  def formula_installed(_f)
+    @formula_count += 1
+  end
+
+  def display_messages
+    display_caveats
+  end
+
+  def display_caveats
+    return unless @formula_count > 1
```
MikeMcQuaid: `return if @formula_count < 1`

Library/Homebrew/messages.rb
```diff
@@ -0,0 +1,36 @@
...
+class Messages
+  attr_reader :caveats, :formula_count
+
+  def initialize
+    @caveats = []
+    @formula_count = 0
+  end
+
+  def record_caveats(f, caveats)
+    @caveats.push([f, caveats])
```
MikeMcQuaid: A hash might be nicer here?

apjanke: Not sure a hash would be best — I'd like to preserve the ordering of the messages wrt how they were first encountered. (I'm thinking ahead to build times here.) Or maybe alphabetical would be better?

Library/Homebrew/messages.rb
apjanke: Sure. Give me a little while...

Library/Homebrew/messages.rb
apjanke: Amended.

Library/Homebrew/messages.rb
apjanke: Yeah, that sounds better. Amended.

Library/Homebrew/messages.rb
MikeMcQuaid: `@caveats.push({f.name => caveats})` perhaps?

Library/Homebrew/global.rb
```diff
@@ -30,6 +31,9 @@
 module Homebrew
   extend FileUtils
 
+  @messages = Messages.new
+  class << self; attr_accessor :messages; end
```
MikeMcQuaid: There's a `class << self` just below here it might be nicer for all this to live in (and perhaps a lazy initialisation of `@messages`

apjanke: Amended to address.

Library/Homebrew/messages.rb
apjanke: That gets kind of ugly when doing the reading:

```
  ohai c.keys.first, c[c.keys.first]
```

Maybe `@caveats.push({"name" => f.name, "caveats" => caveats})`?

MikeMcQuaid: Yeh, that looks nicer. I'd still be tempted to just store the formula name, though, if that's all you're using?

apjanke: Amended to store just the formula name.

Library/Homebrew/messages.rb
apjanke: Do you want to see the `Messages` class tested in isolation in-process, or an example of shelling out to a `brew` command that produces collected caveats?

MikeMcQuaid: @apjanke Tested in process/unit tests please, thanks!

Library/Homebrew/messages.rb
RandomDSdevel: Repository lurker sharp-sighting:  Shouldn't that be '`<=`,' not just '`<`' (to properly be the true reverse of '`>`?')

apjanke: Yep. The amended commit uses `return if @formula_count <= 1`.

Library/Homebrew/messages.rb
MikeMcQuaid: @apjanke Gentle ping on some unit tests, thanks!

apjanke: Added unit tests.

MikeMcQuaid:
@apjanke Can you rebase on `origin/master`? Thanks!

MikeMcQuaid:
@apjanke Ignore me, all green now, thanks!

MikeMcQuaid:
Thanks again @apjanke!

PR #4377, opened 2018-06-26
Implement HOMEBREW_FORCE_BREWED_GIT

jacktose:
Because of [this messing with the user's path](https://github.com/Homebrew/brew/blob/efc02899c851c62c9ce0d15dea9a231575d7d774/bin/brew#L68), brew uses `/usr/bin/git` over brewed git, even when the former is problematically old.
There may also be other reasons a user prefers to use brewed git.

There was already a `HOMEBREW_FORCE_BREWED_CURL` option and a `HOMEBREW_SYSTEM_CURL_TOO_OLD` check to set it. This mostly copies those to implement `HOMEBREW_FORCE_BREWED_GIT` & `HOMEBREW_SYSTEM_GIT_TOO_OLD`.

See also: https://github.com/Linuxbrew/brew/issues/736

-----

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -102,6 +102,7 @@ then
   if [[ "$HOMEBREW_MACOS_VERSION_NUMERIC" -lt "100900" ]]
   then
     HOMEBREW_SYSTEM_GIT_TOO_OLD="1"
+    HOMEBREW_FORCE_BREWED_GIT="1"
   fi
 
   if [[ -z "$HOMEBREW_CACHE" ]]
@@ -133,6 +134,15 @@ then
   HOMEBREW_CURL="$HOMEBREW_PREFIX/opt/curl/bin/curl"
 else
   HOMEBREW_CURL="curl"
+fi
+
+if [[ -n "$HOMEBREW_FORCE_BREWED_GIT" &&
+      -x "$HOMEBREW_PREFIX/opt/git/bin/git" ]] &&
+         "$HOMEBREW_PREFIX/opt/git/bin/git" --version >/dev/null
+then
+  HOMEBREW_GIT="$HOMEBREW_PREFIX/opt/git/bin/git"
+else
+  HOMEBREW_GIT="git"
 fi
 
 HOMEBREW_USER_AGENT="$HOMEBREW_PRODUCT/$HOMEBREW_USER_AGENT_VERSION ($HOMEBREW_SYSTEM; $HOMEBREW_PROCESSOR $HOMEBREW_OS_USER_AGENT_VERSION)"
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -384,7 +384,8 @@ EOS
   fi
 
   if ! git --version &>/dev/null ||
-       [[ -n "$HOMEBREW_SYSTEM_GIT_TOO_OLD" &&
+     [[ ( -n "$HOMEBREW_SYSTEM_GIT_TOO_OLD" ||
+          -n "$HOMEBREW_FORCE_BREWED_GIT" ) &&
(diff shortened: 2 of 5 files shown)


MikeMcQuaid:
I’m 👍🏻 on this but you’ll need to run ‘brew man’ and commit the result.

jacktose:
Manpages fixed, now failing codecov. Can someone give me a hand with that? I don't really know how coverage works, and I didn't touch the file that it says has changed.

jacktose:
Manpages fixed, now failing codecov. Can someone give me a hand with that? I don't really know how coverage works, and I didn't touch the file that it says has changed.

Library/Homebrew/cmd/update.sh
```diff
@@ -384,7 +384,8 @@ EOS
   fi
 
   if ! git --version &>/dev/null ||
-       [[ -n "$HOMEBREW_SYSTEM_GIT_TOO_OLD" &&
+     [[ ( -n "$HOMEBREW_SYSTEM_GIT_TOO_OLD" ||
```
MikeMcQuaid: Given you always set force brewed git above when git is too old you can just make this check for forced brewed git

MikeMcQuaid:
Don’t worry about codecov. Sorry I missed the previous comment. Will merge once that’s changed. Sorry for brevity; on my phone.

Library/Homebrew/cmd/update.sh
jacktose: Oh, above = brew.sh? We can count on that always running first?
I suppose that's what the CURL options are doing, so I'll match that.

MikeMcQuaid:
Great work, thanks a lot! Hope to see more contributions from you.

jacktose:
Thanks, me too. Trying to get my feet wet in OSS contribution.
Thanks to @sjackman too for the help & encouragement.

sjackman:
Excellent! Thanks for this work, Jack! 🏅It'll be a big help for Linuxbrew. And thanks for merging, Mike.

Library/Homebrew/brew.sh
```diff
@@ -102,6 +102,7 @@ then
   if [[ "$HOMEBREW_MACOS_VERSION_NUMERIC" -lt "100900" ]]
   then
     HOMEBREW_SYSTEM_GIT_TOO_OLD="1"
```
reitermarkus: Is this still used now?

jacktose: I hope so. There's no other check (that I've seen) for too-old-ness. Without that, it's just a manual option.
Maybe there's a better or additional check, though? Parsing `git --version`?

sjackman: Yes, parsing `git --version` would be preferable, at least on Linux.
@reitermarkus Which do you prefer on macOS? Checking the version of Git, or checking the version of macOS?

reitermarkus: I actually just meant if the variable `HOMEBREW_SYSTEM_GIT_TOO_OLD` is still used, not the check itself.

jacktose: It seems that `HOMEBREW_SYSTEM_CURL_TOO_OLD` is redundant, because `HOMEBREW_FORCE_SYSTEM_CURL` is set at the same time. The `GIT` variables are now in the same situation.
I guess the question is, is there a case for using an auto-determined `TOO_OLD` separately from a user-settable `FORCE`?

sjackman: `HOMEBREW_SYSTEM_GIT_TOO_OLD` does not appear to be used.

`HOMEBREW_SYSTEM_CURL_TOO_OLD` is used here:
https://github.com/Homebrew/brew/blob/af9ddfe899c3593a4d8168d4ade61eccd01d19b0/Library/Homebrew/extend/os/mac/development_tools.rb#L48

PR #4380, opened 2018-06-27
Properly append values to HOMEBREW_CCCFG

maxim-belkin:

-----

I don't see anything wrong with having spaces (the default separator) in `HOMEBREW_CCCFG`, but for the sake of consistency....

### DIFF
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -55,7 +55,7 @@ module SharedEnvExtension
   end
 
   def append_to_cccfg(value)
-    append(HOMEBREW_CCCFG, value, "")
+    append("HOMEBREW_CCCFG", value, "")
   end
 
   def append(keys, value, separator = " ")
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
Can you explain the motivation for this change, why this is the right solution and what (if any) user issues this fixes?

MikeMcQuaid:
Ok. Is the extra argument needed? If it's not: maybe it's worth removing them rather than adding them?

sjackman:
The third argument is the separator. The default value is a space. There's no particular harm in appending a space, other than being a tad unsightly.

maxim-belkin:
> maybe it's worth removing them rather than adding them?

either way is fine. I think consistency is more important than the actual implementation. But adding two `""` is easier :)

ilovezfs:
We do charge extra for spaces, though.

maxim-belkin:
not_a_problem_:)

MikeMcQuaid:
Suggestion: have method just to do the `HOMEBREW_CCCFG` append which takes a single argument and can be used consistently.

maxim-belkin:
ok

MikeMcQuaid:
@maxim-belkin Gentle ping :)

maxim-belkin:
done

maxim-belkin:
Ooops

MikeMcQuaid:
Thanks again @maxim-belkin!

PR #4381, opened 2018-06-27
Add make shim to Linux super env

maxim-belkin:

-----

Add a `shim` for `make` that has been missing from the super env on Linux. 

CC @sjackman and @iMichka

### DIFF
--- a/Library/Homebrew/shims/linux/super/make
+++ b/Library/Homebrew/shims/linux/super/make
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+if [[ -n "$HOMEBREW_MAKE" && "$HOMEBREW_MAKE" != "make" ]]
+then
+  export MAKE="$HOMEBREW_MAKE"
+else
+  MAKE="make"
+fi
+export HOMEBREW_CCCFG="O$HOMEBREW_CCCFG"
+exec "$MAKE" "$@"
--- a/Library/Homebrew/shims/linux/super/make
+++ b/Library/Homebrew/shims/linux/super/make
@@ -7,4 +7,20 @@ else
   MAKE="make"
 fi
 export HOMEBREW_CCCFG="O$HOMEBREW_CCCFG"
+
+pathremove () {
+        local IFS=':'
+        local NEWPATH
+        local DIR
+        local PATHVARIABLE=${2:-PATH}
+        for DIR in ${!PATHVARIABLE} ; do
+                if [ "$DIR" != "$1" ] ; then
+                  NEWPATH=${NEWPATH:+$NEWPATH:}$DIR
+                fi
+        done
+        export $PATHVARIABLE="$NEWPATH"
+}
+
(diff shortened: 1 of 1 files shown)


sjackman:
It looks to me as though `shims/linux/super/make` would call itself in an infinite loop (unconfirmed).

sjackman:
A question for the Homebrew developers. `shims/mac/super/make` adds `O` to `HOMEBREW_CCCFG` to enabled `ENV.refurbish_args` by default for any formula that uses `make`, but not for `./configure` or any non-make build system. Does someone know why that is, rather than enabling `ENV.refurbish_args` by default for all build systems?

maxim-belkin:
> It looks to me as though shims/linux/super/make would call itself in an infinite loop (unconfirmed).

Ha! interesting observation. It might be the case. I guess we have to change line 7 to`
```
MAKE=$(type -p make)
```
or 
```
MAKE=$(command -v make)
```
?

sjackman:
Perhaps, but you'll need to first exclude `…/shims/linux/super` from `PATH`.

maxim-belkin:
@sjackman done

jonchang:
> A question for the Homebrew developers. shims/mac/super/make adds O to HOMEBREW_CCCFG to enabled ENV.refurbish_args by default for any formula that uses make, but not for ./configure or any non-make build system. Does someone know why that is, rather than enabling ENV.refurbish_args by default for all build systems?

Wouldn't enabling `refurbish_args` make most checks in `./configure` pointless since e.g. autotools will call `cc` with different flags to determine what kind of compiler, etc. it's running on?

sjackman:
Yes, `ENV.refurbish_args` would probably break `./configure`. There's a check to prevent specifically that:
https://github.com/Homebrew/brew/blob/0d33aba0b7d8448e815a2eecc92417b46cb905f3/Library/Homebrew/shims/super/cc#L104

I'm more curious why `ENV.refurbish_args` is not use for other non-make build systems, or phrased another way, why it's used specifically for `make`.

sjackman:
Also interesting that `ENV.refurbish_args` is not used when`ld` is used for linking, but is used when `cc` is used for linking.

ilovezfs:
Presumably because `make` usually is separated from configure.

MikeMcQuaid:
Why is/does this need to be different from the macOS shim?

maxim-belkin:
we don't have `xcrun`

MikeMcQuaid:
I see. The path removal feels a little cludgy; is there no easier way of doing that?

MikeMcQuaid:
@maxim-belkin Ping?

maxim-belkin:
How can I improve this PR?  Rewrite `pathremove` function? 
I can make it specific to `PATH` var instead of any variable but I don't know if it will be an improvement.

MikeMcQuaid:
@maxim-belkin It's fine with me, just a question. Merging, thanks again!

maxim-belkin:
Thanks, Mike!

PR #4385, opened 2018-06-28
cask audit: check for delete/trash in uninstall and zap

commitay:

-----

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/audit.rb
+++ b/Library/Homebrew/cask/lib/hbc/audit.rb
@@ -33,7 +33,7 @@ module Hbc
       check_download
       check_single_pre_postflight
       check_single_uninstall_zap
-      check_untrusted_pkg
+      check_official_taps
       check_hosting_with_appcast
       check_latest_with_appcast
       check_stanza_requires_uninstall
@@ -54,8 +54,8 @@ module Hbc
 
     private
 
-    def check_untrusted_pkg
-      odebug "Auditing pkg stanza: allow_untrusted"
+    def check_official_taps
+      odebug "Auditing official tap only checks"
 
       return if @cask.sourcefile_path.nil?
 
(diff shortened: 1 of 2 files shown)


reitermarkus:
Not everything should be trashed, only user-generated stuff should be. Anything under `~/Library/Caches` can be deleted when zapping.

The opposite however, could be true. `uninstall` should never use `trash`.

commitay:
https://github.com/Homebrew/homebrew-cask/issues/40501

commitay:
> `uninstall` should never use `trash`.

👍  I'll add this as well.

reitermarkus:
We can have a whitelist for `delete`, starting with `~/Library/Caches` and `~/Library/Saved Application State`.

commitay:
What would whitelisting them achieve apart from complicating the zap stanza?

reitermarkus:
> What would whitelisting them achieve apart from complicating the zap stanza?

The point is that useless things such as caches should just be straight up deleted and not “clutter” your trash. So maybe this should rather be a `trash` blacklist than a `delete` whitelist.

commitay:
Yes, I understand what your point is but I think it overly complicates a non-essential part of cask for a _very_ limited benefit.

Maintainers having to sort zap stanzas by what should be deleted and what should be trashed is not a productive use of our time, which is what prompted https://github.com/Homebrew/homebrew-cask/issues/40501 to begin with.

reitermarkus:
> Maintainers having to sort zap stanzas by what should be deleted and what should be trashed is not a productive use of our time

I agree, this should be done by `brew cask style --fix`. That doesn't mean we should throw everything out the window now and use `trash` no matter what. Having a whitelist doesn't complicate anything, it just allows us to use `delete` when it is semantically more appropriate. A whitelist achieves everything you want, reducing maintainer burden, while keeping the flexibility to use `delete`.

commitay:
> it overly complicates a non-essential part of cask for a _very_ limited benefit

This is true even if it is handled automatically by style.

> That doesn't mean we should throw everything out the window now and use `trash` no matter what.

Nothing is being throw out the window, everything has _already_ been changed to use zap trash last year. Current only 5 casks fail this audit. https://github.com/Homebrew/homebrew-cask/pull/49090

commitay:
> How is having `delete` there complicating it?

Currently everything get trashed, which is very straightforward. AFAIK there have been no user requests nor any technical reason to split up zap to preserve some files with trash and delete others.

Zap is nice feature for users but it is hardly an essential part of cask, I do not see the need to complicate it just for the sake of making the trash directory a little bit less cluttered if a user runs zap.

reitermarkus:
> AFAIK there have been no user requests nor any technical reason to split up zap to preserve some files with trash and delete others.

I could argue the same thing. There have also been no user requests to remove `delete` functionality from `zap`. There also has been no technical reason for it.

> I do not see the need to complicate it

Even if you don't see the *need*, at least let me have the *option*.

vitorgalvao:
The issue here is if maintaining `delete` and `trash` and dealing with it manually is worth the burden.

I think we all agree `~/Library/Caches` and `~/Library/Saved Application State` are useless and should be deleted instead of trashed. We also agree everything else in `uninstall` should be deleted and everything else in `zap` should be trashed.

So if we all agree with what should be done in which case, why do we even *have* two options? Can’t we just get rid of them both and have a single one (let’s call it “remove”) that is context aware and does all that automatically for each case?

I’m calling it “remove” purely for differentiating reasons. I’d be fine with it becoming the default behaviour for one of the other names.

MikeMcQuaid:
What's the latest here?

PR #4397, opened 2018-06-30
brew.sh: Don't allow system tmp dirs as prefixes

woodruffw:

-----

This explicitly forbids the use of the system's temporary directory as an installation prefix, as we perform checks against the temporary directory when doing binary relocation.

See: https://github.com/Homebrew/homebrew-core/issues/28877#issuecomment-401511742

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -114,8 +114,8 @@ else
   : "${HOMEBREW_OS_VERSION:=$(uname -r)}"
   HOMEBREW_OS_USER_AGENT_VERSION="$HOMEBREW_OS_VERSION"
 
-  cache_home="${XDG_CACHE_HOME:-${HOME}/.cache}"
-  HOMEBREW_CACHE="${HOMEBREW_CACHE:-${cache_home}/Homebrew}"
+  CACHE_HOME="${XDG_CACHE_HOME:-${HOME}/.cache}"
+  HOMEBREW_CACHE="${HOMEBREW_CACHE:-${CACHE_HOME}/Homebrew}"
 
   HOMEBREW_TEMP="${HOMEBREW_TEMP:-/tmp}"
 fi
@@ -295,10 +295,10 @@ check-prefix-is-not-tmpdir() {
   if [[ "${HOMEBREW_PREFIX}" = "${HOMEBREW_TEMP}"* ]]
   then
     odie <<EOS
-Your HOMEBREW_PREFIX is in the system temporary directory, which Homebrew
+Your HOMEBREW_PREFIX is in the Homebrew temporary directory, which Homebrew
 uses to store downloads and builds. You can resolve this by installing Homebrew to
 either the standard prefix (/usr/local) or to a non-standard prefix that is not
-in the system temporary directory.
+in the Homebrew temporary directory.
 EOS
   fi
 }
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -108,6 +108,8 @@ then
   then
     HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
   fi
+
+  HOMEBREW_TEMP="${HOMEBREW_TEMP:-/private/tmp}"
 else
   HOMEBREW_PROCESSOR="$(uname -m)"
   HOMEBREW_PRODUCT="${HOMEBREW_SYSTEM}brew"
@@ -124,6 +126,8 @@ else
       HOMEBREW_CACHE="$HOME/.cache/Homebrew"
     fi
   fi
+
+  HOMEBREW_TEMP="${HOMEBREW_TEMP:-/tmp}"
 fi
 
 if [[ -n "$HOMEBREW_FORCE_BREWED_CURL" &&
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -301,7 +301,7 @@ check-prefix-is-not-tmpdir() {
 Your HOMEBREW_PREFIX is in the system temporary directory, which Homebrew
 uses to store downloads and builds. You can resolve this by installing Homebrew to
 either the standard prefix (/usr/local/) or to a non-standard prefix that is not
-the system temporary directory.
+in the system temporary directory.
 EOS
   fi
 }
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -104,10 +104,7 @@ then
     HOMEBREW_SYSTEM_GIT_TOO_OLD="1"
   fi
 
-  if [[ -z "$HOMEBREW_CACHE" ]]
-  then
-    HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
-  fi
+  HOMEBREW_CACHE="${HOMEBREW_CACHE:-${HOME}/Library/Caches/Homebrew}"
 
   HOMEBREW_TEMP="${HOMEBREW_TEMP:-/private/tmp}"
 else
@@ -117,15 +114,8 @@ else
   : "${HOMEBREW_OS_VERSION:=$(uname -r)}"
   HOMEBREW_OS_USER_AGENT_VERSION="$HOMEBREW_OS_VERSION"
 
-  if [[ -z "$HOMEBREW_CACHE" ]]
-  then
-    if [[ -n "$XDG_CACHE_HOME" ]]
-    then
-      HOMEBREW_CACHE="$XDG_CACHE_HOME/Homebrew"
-    else
-      HOMEBREW_CACHE="$HOME/.cache/Homebrew"
-    fi
-  fi
+  cache_home="${XDG_CACHE_HOME:-${HOME}/.cache}"
+  HOMEBREW_CACHE="${HOMEBREW_CACHE:-${cache_home}/Homebrew}"
 
   HOMEBREW_TEMP="${HOMEBREW_TEMP:-/tmp}"
 fi
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -294,6 +294,20 @@ EOS
 }
 check-run-command-as-root
 
+check-prefix-is-not-tmpdir() {
+  if [[ "${HOMEBREW_PREFIX}" = /tmp/* ||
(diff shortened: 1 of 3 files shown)


Library/Homebrew/brew.sh
```diff
@@ -294,6 +294,19 @@ EOS
...
 check-run-command-as-root
 
+check-prefix-is-not-tmpdir() {
+  if [[ $(realpath "${HOMEBREW_PREFIX}") = /private/tmp/* ]]
+  then
+    odie <<EOS
+Your HOMEBREW_PREFIX is in the system temporary directory, which Homebrew
+uses to store downloads and builds. You can resolve this by installing Homebrew to
+either the standard prefix (/usr/local/) or to a non-standard prefix that is not
+the system temporary directory.
```
ilovezfs: s/the/in the/

ilovezfs:
The [underlying issue](https://github.com/Homebrew/brew/blob/9fca172d0366da076295b14b4e01917919963282/Library/Homebrew/extend/os/mac/keg_relocate.rb#L64-L66) specifically affects HOMEBREW_TEMP regardless of whether it's the default /tmp or some other path (and if HOMEBREW_TEMP isn't /tmp then /tmp is unaffected) so I'm wondering if the check and the messaging should check the value of HOMEBREW_TEMP for a non-default value and handle that case correctly too.

Library/Homebrew/brew.sh
```diff
@@ -294,6 +294,19 @@ EOS
 }
 check-run-command-as-root
 
+check-prefix-is-not-tmpdir() {
+  if [[ $(realpath "${HOMEBREW_PREFIX}") = /private/tmp/* ]]
+  then
+    odie <<EOS
+Your HOMEBREW_PREFIX is in the system temporary directory, which Homebrew
```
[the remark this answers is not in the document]
woodruffw: I thought that the connection between `/tmp` and `/private/tmp` might confuse some users (Apple links `/tmp` to `/private/tmp` for FHS compliance), so I just went with "system temporary directory." I can change it/add those directories to the message if you think it'll improve the error, though.

The relocation code at fault isn't used on Linux, so Linuxbrew users might not encounter this problem at all. I'll confirm that in a bit with some local testing.

woodruffw:
> so I'm wondering if the check and the messaging should check the value of HOMEBREW_TEMP for a non-default value and handle that case correctly too.

Yeah, I'll update it to handle this. It gets a little complicated, though, thanks to the `/tmp` <-> `/private/tmp` link. Would anybody mind if I added a `realpath` to the assignment of `HOMEBREW_TEMP` in `config.rb`, so that we don't have to worry about handling all of the possible permutations of users installing in `/tmp` and `/private/tmp`?

bin/brew
```diff
@@ -72,7 +73,7 @@ then
 
   FILTERED_ENV=()
   # Filter all but the specific variables.
-  for VAR in HOME SHELL PATH TERM COLUMNS LOGNAME USER CI TRAVIS SSH_AUTH_SOCK SUDO_ASKPASS \
+  for VAR in HOME SHELL PATH TEMP TERM COLUMNS LOGNAME USER CI TRAVIS SSH_AUTH_SOCK SUDO_ASKPASS \
```
MikeMcQuaid: Why do we want to add this?

Library/Homebrew/config.rb
```diff
@@ -39,7 +39,7 @@
 HOMEBREW_LOGS = Pathname.new(ENV["HOMEBREW_LOGS"] || "~/Library/Logs/Homebrew/").expand_path
 
 # Must use /tmp instead of $TMPDIR because long paths break Unix domain sockets
-HOMEBREW_TEMP = Pathname.new(ENV.fetch("HOMEBREW_TEMP", "/tmp"))
+HOMEBREW_TEMP = Pathname.new(ENV.fetch("HOMEBREW_TEMP", "/tmp")).realpath
```
MikeMcQuaid: This will never be unset/set to `/tmp` now.

bin/brew
woodruffw: Oh, I misread this code. I thought this was explicitly whitelisting `HOMEBREW_TEMP`, but I guess it would clobber it instead.

woodruffw:
Looks like the test failure is coming from `brew update-test`, which uses `HOMEBREW_TEMP` via `mktemp` to test `brew update`.

Any thoughts on how best to resolve this? I could add another environment variable allowing the behavior in the PR to be overridden for testing purposes, or change `brew update-test` to avoid `HOMEBREW_TEMP`, off the top of my head. Neither of those is particularly clean, though :disappointed:

MikeMcQuaid:
> change brew update-test to avoid HOMEBREW_TEMP

I'd vote on doing this; no reason it can't act in the current directory (as a developer, testing command) if it's cleaned up afterwards.

bin/brew
```diff
@@ -21,6 +21,7 @@ symlink_target_directory() {
 BREW_FILE_DIRECTORY="$(quiet_cd "${0%/*}/" && pwd -P)"
 HOMEBREW_BREW_FILE="${BREW_FILE_DIRECTORY%/}/${0##*/}"
 HOMEBREW_PREFIX="${HOMEBREW_BREW_FILE%/*/*}"
+HOMEBREW_TEMP="${HOMEBREW_TEMP:-/private/tmp}"
```
MikeMcQuaid: Will this upset Linux? Worth setting in `brew.sh` to a different default for macOS and not-macOS?

woodruffw: Yeah, `/private/tmp` is macOS only. Moving it to `brew.sh` sounds good.

Library/Homebrew/brew.sh
```diff
@@ -108,6 +108,8 @@ then
   then
     HOMEBREW_CACHE="$HOME/Library/Caches/Homebrew"
   fi
+
+  HOMEBREW_TEMP="${HOMEBREW_TEMP:-/private/tmp}"
```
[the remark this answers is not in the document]
woodruffw: I'm in favor of replacing with the `:-` form, since it's the standard bashism for a default value and we use similar bashisms elsewhere in the code.

Library/Homebrew/brew.sh
```diff
@@ -294,6 +299,19 @@ EOS
 }
 check-run-command-as-root
 
+check-prefix-is-not-tmpdir() {
+  if [[ "${HOMEBREW_PREFIX}" = "${HOMEBREW_TEMP}"* ]]
```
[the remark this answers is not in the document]
woodruffw: > check against the system temp rather than just HOMEBREW_TEMP

I might be misunderstanding, but: won't this leave open the case where they set both their prefix and their temp to something weird, e.g., `$HOME`? We want to stop all situations where Homebrew is both installing to and using the same prefix as a temporary directory, not just `/private/tmp`.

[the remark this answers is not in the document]
MikeMcQuaid: @woodruffw Ok so is it fine to e.g. have the prefix in `/tmp` if `HOMEBREW_TEMP` is set elsewhere?

Library/Homebrew/brew.sh
```diff
@@ -104,26 +104,20 @@ then
...
-  if [[ -z "$HOMEBREW_CACHE" ]]
-  then
-    if [[ -n "$XDG_CACHE_HOME" ]]
-    then
-      HOMEBREW_CACHE="$XDG_CACHE_HOME/Homebrew"
-    else
-      HOMEBREW_CACHE="$HOME/.cache/Homebrew"
-    fi
-  fi
+  cache_home="${XDG_CACHE_HOME:-${HOME}/.cache}"
```
MikeMcQuaid: Use all caps for variable name here and just don't export it.

Library/Homebrew/brew.sh
```diff
@@ -294,6 +289,21 @@ EOS
 }
 check-run-command-as-root
 
+check-prefix-is-not-tmpdir() {
+  [[ -z "${HOMEBREW_MACOS}" ]] && return
+
+  if [[ "${HOMEBREW_PREFIX}" = "${HOMEBREW_TEMP}"* ]]
+  then
+    odie <<EOS
+Your HOMEBREW_PREFIX is in the system temporary directory, which Homebrew
```
MikeMcQuaid: `HOMEBREW_TEMP` isn't necessary the _system_ temporary directory. To expand on my previous comment (and avoid the autohiding): is the issue if:

- the prefix is in the system temporary directory
- the prefix is the in the HOMEBREW_TEMP temporary directory
- either

We should ensure that the wording makes that clear and the logic is as minimal as possible to the actual problems we've seen people experience.

Library/Homebrew/brew.sh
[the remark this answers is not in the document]
woodruffw: Yep, exactly.

Library/Homebrew/brew.sh
woodruffw: The issue is when the prefix is in the `HOMEBREW_TEMP` temporary directory. I'll correct the language here :slightly_smiling_face:

MikeMcQuaid:
Thanks @woodruffw and sorry for all the back and forth! Took me a while to understand what was going on here.

Library/Homebrew/config.rb
```diff
@@ -39,7 +39,7 @@
 HOMEBREW_LOGS = Pathname.new(ENV["HOMEBREW_LOGS"] || "~/Library/Logs/Homebrew/").expand_path
 
 # Must use /tmp instead of $TMPDIR because long paths break Unix domain sockets
-HOMEBREW_TEMP = Pathname.new(ENV.fetch("HOMEBREW_TEMP", "/tmp"))
+HOMEBREW_TEMP = Pathname.new(ENV["HOMEBREW_TEMP"]).realpath
```
ilovezfs: I think this will crash when the dir doesn't exist.

MikeMcQuaid: Yup. Fixing 😭

MikeMcQuaid: https://github.com/Homebrew/brew/pull/4415

ilovezfs: Thanks!

woodruffw:
No problem at all, but: I hadn't applied the `brew update-test` changes yet, so this still breaks the CI. Maybe worth a revert (or I could just fill it in with another PR)?

(It's 100% my fault -- I forgot to add the "do not merge" label).

PR #4408, opened 2018-07-02
dependency_order_cop: skip versioned formulae

commitay:

-----


MikeMcQuaid:
@commitay Gentle reminder of https://github.com/Homebrew/brew/blob/master/docs/Maintainer-Guidelines.md#communication; can you explain the motivation and background for this change in the PR body? Thanks!

ilovezfs:
We could probably skip all new-formula-only audits for @ formulae as it's pretty much universally noise given they are copied verbatim from the originals.

commitay:
"formulae should not have a HEAD spec" might be useful to keep as an @ formulae with a HEAD spec that we would want to keep is kind of the exception (erlang@20 and llvm@5) rather than the norm. I can't really think of any other new formulae audits?

ilovezfs:
We could disallow head specs in @ formulae as a standard audit. The existing `--HEAD` options in @ formulae are getting almost no use. Here's install on request for the last 90 days for @ formulae with a head spec:
```
2 | bash-completion@2 --HEAD                                                        |    17 |   0.31%
2 | erlang@17 --HEAD                                                                   |  1 |   1.11%
26 | imagemagick@6 --HEAD                                                          |     11 |   0.03%
53 | imagemagick@6 --HEAD --with-libde265                                          |      2 |   0.01%
2 | ipython@5 --HEAD                                                                  |   1 |   0.17%
2 | llvm@3.7 --HEAD                                                                   |   4 |   1.12%
 6 | llvm@4 --HEAD                                                                  |     1 |   0.02%
10 | llvm@5 --HEAD --with-python@2                                                    |   1 |   0.13%
3 | python@2 --HEAD                                                               |      29 |   0.02%
5 | python@2 --HEAD --with-tcl-tk                                                 |      16 |   0.01%
2 | rebar@3 --HEAD                                                                  |     2 |   0.13%
```

ilovezfs:
(but white list --HEAD for python@2)

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -236,6 +236,7 @@ def audit_style
       return unless @style_offenses
       @style_offenses.each do |offense|
         if offense.cop_name.start_with?("NewFormulaAudit")
+          next if formula.versioned_formula?
```
commitay: Not sure if there is a better way to disable all `NewFormulaAudit` for @ formulae?

Issue #4291, opened 2018-06-06
Add `brew cask --json=v1` for JSON output


MikeMcQuaid:
This should follow as close a format as possible to `brew info --json=v1` and will be consumed and shared [by formulae.brew.sh](https://github.com/Homebrew/formulae.brew.sh/issues/13).

albertpaulp:
Hi @MikeMcQuaid , I'd like to work on this, is there anybody already working on this ?

MikeMcQuaid:
@albertpaulp There's no open PRs so you can feel free to create a PR. Anyone else should feel free to create a PR before you have, though 😄

MikeMcQuaid:
Fixed by @alexbostock in 
https://github.com/Homebrew/brew/pull/4410.

MikeMcQuaid:
@carddamom It'll be available in the next tag which will drop on Monday UK time at the latest.

PR #4410, opened 2018-07-02
Cask json

alexbostock:

-----
Issue #4291 

`$ brew cask info --json-v1 <formula>` outputs JSON, similar to `$ brew info --json=v1 <formula>`

I had trouble with the option parsing to match the exact format `--json=v1`. This will probably need fixing.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/abstract_flight_block.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/abstract_flight_block.rb
@@ -26,10 +26,6 @@ module Hbc
         abstract_phase(self.class.uninstall_dsl_key)
       end
 
-      def summarize
-        directives.keys.map(&:to_s).join(", ")
-      end
-
       private
 
       def class_for_dsl_key(dsl_key)
@@ -40,6 +36,10 @@ module Hbc
       def abstract_phase(dsl_key)
         return if (block = directives[dsl_key]).nil?
         class_for_dsl_key(dsl_key).new(cask).instance_eval(&block)
+      end
+
+      def summarize
+        directives.keys.map(&:to_s).join(", ")
       end
     end
   end
--- a/Library/Homebrew/cask/lib/hbc/artifact/abstract_flight_block.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/abstract_flight_block.rb
@@ -26,6 +26,10 @@ module Hbc
         abstract_phase(self.class.uninstall_dsl_key)
       end
 
+      def summarize
+        directives.keys.map(&:to_s).join(", ")
+      end
+
       private
 
       def class_for_dsl_key(dsl_key)
@@ -36,10 +40,6 @@ module Hbc
       def abstract_phase(dsl_key)
         return if (block = directives[dsl_key]).nil?
         class_for_dsl_key(dsl_key).new(cask).instance_eval(&block)
-      end
-
-      def summarize
-        directives.keys.map(&:to_s).join(", ")
       end
     end
   end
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cask/lib/hbc/cli/info.rb
```diff
@@ -1,15 +1,24 @@
+require "json"
+
 module Hbc
   class CLI
     class Info < AbstractCommand
+      option "--json-v1", :json, false
```
reitermarkus:
```ruby
option "--json=VERSION", :json
```

Then access it with `json` instead of `json?`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -129,5 +129,32 @@ def dumpcask
...
+        "depends_on" => depends_on,
+        "conflicts_with" => conflicts_with,
+        "container" => container,
+        "gpg" => gpg,
+        "accessibility_access" => accessibility_access,
+        "auto_updates" => auto_updates,
+      }
+
+      artifacts.each do |a|
+        hsh["artifacts"][a.class.english_name] = a.summarize
```
reitermarkus: This should use `dsl_key` instead of `english_name`. And I think `artifacts` should be an array, not a hash. Also, this should call `to_h` or `to_a` on the artifact, not `summarize`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -129,5 +129,32 @@ def dumpcask
...
+      end
+
+      hsh["conflicts_with"] = [] if hsh["conflicts_with"].nil?
```
reitermarkus: Just call `to_a` on `conflicts_with` above, since `nil.to_a == []`.

Library/Homebrew/cask/lib/hbc/cask.rb
alexbostock: Re: call `to_h` or `to_a` on the artifact

I'm not sure how this should work, with different types of artifact having different methods. Please see my latest commit, using `methods.include?`. Is that how it should be?

MikeMcQuaid:
@reitermarkus This was force-pushed. How's it look to you now

Library/Homebrew/cask/lib/hbc/artifact/abstract_flight_block.rb
```diff
@@ -26,6 +26,10 @@ def uninstall_phase(**)
         abstract_phase(self.class.uninstall_dsl_key)
       end
 
+      def summarize
```
reitermarkus: Move this back to `private`, since this isn't actually used publicly now.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -129,5 +129,32 @@ def dumpcask
...
+        "name" => name,
+        "homepage" => homepage,
+        "url" => url,
+        "appcast" => appcast,
+        "version" => version,
+        "sha256" => sha256,
+        "artifacts" => artifacts.map do |a|
+          if a.respond_to? :to_a
+            a.to_a
+          elsif a.methods.include? :to_h
```
reitermarkus: Also use `a.respond_to?` here. And check for `to_h` first, because a `Hash` also responds to `to_a`.

Library/Homebrew/cask/lib/hbc/cask.rb
```diff
@@ -129,5 +129,32 @@ def dumpcask
         odebug "Cask instance method '#{method}':", send(method).to_yaml
       end
     end
+
+    def to_hash
```
reitermarkus: Make this `to_h`.

MikeMcQuaid:
Thanks @alexbostock!

MikeMcQuaid:
Also: thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @alexbostock!

PR #4411, opened 2018-07-02
audit: disallow unstable specs for versioned formulae

commitay:

-----
And also disallow unstable specs for new formulae.

Mentioned in https://github.com/Homebrew/brew/pull/4408#issuecomment-401748629

cc @ilovezfs

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -571,8 +571,19 @@ module Homebrew
         end
       end
 
-      if @new_formula && formula.head
-        new_formula_problem "Formulae should not have a HEAD spec"
+      if formula.head || formula.devel
+        unstable_spec_message = "Formulae should not have an unstable spec"
+        if @new_formula
+          new_formula_problem unstable_spec_message
+        elsif formula.versioned_formula?
+          versioned_unstable_spec = %w[
+            bash-completion@2
+            imagemagick@6
+            openssl@1.1
+            python@2
+          ]
+          problem unstable_spec_message unless versioned_unstable_spec.include?(formula.name)
+        end
       end
 
       throttled = %w[


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -571,8 +571,20 @@ def audit_specs
         end
       end
 
+      head_msg = "Formulae should not have a HEAD spec"
+
       if @new_formula && formula.head
-        new_formula_problem "Formulae should not have a HEAD spec"
+        return if formula.versioned_formula?
```
ilovezfs: This would exit the entire function skipping remaining checks below but this line isn't needed at all if we're going to skip all @new_formula audits for versioned formulae.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -571,8 +571,20 @@ def audit_specs
...
+        new_formula_problem head_msg
+      end
+
+      versioned_head_whitelist = %w[
+        python@2
+      ]
+
+      if formula.versioned_formula? && formula.head
+        return if versioned_head_whitelist.include?(formula.name)
```
ilovezfs: Again, don't want to return from the entire function, so
```
if formula.versioned_formula? && formula.head &&
   !versioned_head_whitelist.include?(formula.name)
```

Library/Homebrew/dev-cmd/audit.rb
commitay: > if we're going to skip all @new_formula audits for versioned formulae.

How?

ilovezfs: Something like
```
      @new_formula = options[:new_formula] && !formula.versioned_formula?
```
I suppose?

commitay:
Do we want to also disallow devel specs for versioned formulae and whitelist openssl@1.1?

ilovezfs:
`devel` on versioned is probably fine, though rare.

`devel` on new formula is sketchier (if nothing else, `brew pull` gets confused about the standardized commit message) but also rare.

commitay:
We actually can't do devel spec && versioned formula && new formula as we've disabled the new formula audit for versioned formulae.

I've updated the PR to whitelist the versioned formulae with unstable specs.

MikeMcQuaid:
Thanks again @commitay!

PR #4412, opened 2018-07-02
Re-enable tests by removing a stray rspec `:focus`

claui:
- [ ] **(Does not apply)** Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).

-----

Following in the tradition of #3006, this PR removes a stray `:focus`. I only caught this because I ran a known-bad test by pure coincidence.

Generally, I feel that those little pests are pretty hard to catch. I’m thinking of enabling a [focus Cop](https://github.com/rubocop-hq/rubocop-rspec/blob/61f1dfed47f1565d964c6fa6446a60b57d4eee99/lib/rubocop/cop/rspec/focus.rb) so it’ll happen less often in the future. I also suggest to change test-bot so it runs `brew style --rspec`.

@Homebrew/maintainers Opinions?

### DIFF
--- a/Library/Homebrew/test/cask/system_command_spec.rb
+++ b/Library/Homebrew/test/cask/system_command_spec.rb
@@ -80,7 +80,7 @@ describe Hbc::SystemCommand, :cask do
         options.merge!(print_stdout: true)
       end
 
-      it "echoes both STDOUT and STDERR", :focus do
+      it "echoes both STDOUT and STDERR" do
         expect { described_class.run(command, options) }
           .to output("1\n3\n5\n").to_stdout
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @claui!

> Generally, I feel that those little pests are pretty hard to catch. I’m thinking of enabling a focus Cop so it’ll happen less often in the future. I also suggest to change test-bot so it runs brew style --rspec.

There's an open issue to enable this by default but there's a bit of work needed to get there: https://github.com/Homebrew/brew/issues/4058

claui:
@MikeMcQuaid Thanks for the pointer. I wasn’t aware of this issue!

PR #4420, opened 2018-07-04
audit: ensure `bottle` is defined

commitay:

-----
We can't really prevent new formulae from being merged without `bottle :unneeded` but this will catch it the next time it goes through CI.

Also disallow `bottle :disabled` in core and whitelist the formulae that currently use it.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -488,10 +488,29 @@ module Homebrew
       end
     end
 
-    def audit_bottle_spec
+    def audit_bottle_disabled
       return unless formula.bottle_disabled?
-      return if formula.bottle_disable_reason.valid?
-      problem "Unrecognized bottle modifier"
+      return if formula.bottle_unneeded?
+
+      if !formula.bottle_disable_reason.valid?
+        problem "Unrecognized bottle modifier"
+      else
+        bottle_disabled_whitelist = %w[
+          cryptopp
+          leafnode
+        ]
+        return if bottle_disabled_whitelist.include?(formula.name)
+        problem "Formulae should not use `bottle :disabled`" if @official_tap
+      end
+    end
+
+    def audit_bottle_spec
+      return unless @official_tap
+      return if @new_formula
+      return unless @online
+      return if formula.bottle_defined? || formula.bottle_disabled?
+      return if formula.name == "testbottest"
+      problem "`bottle` is not defined"
     end
 
     def audit_github_repository


commitay:
```
==> brew audit testbottest --online
Error: 1 problem in 1 formula
homebrew/test-bot/testbottest:
  * `bottle` is not defined
```

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -488,12 +488,21 @@ def audit_homepage
...
-    def audit_bottle_spec
+    def audit_bottle_disabled
       return unless formula.bottle_disabled?
       return if formula.bottle_disable_reason.valid?
       problem "Unrecognized bottle modifier"
     end
 
+    def audit_bottle_spec
+      return if formula.bottle_defined? || formula.bottle_disabled?
+      return if @new_formula
```
MikeMcQuaid: Still probably want `bottle :unneeded` when needed on new formulae?

MikeMcQuaid: And don't want `bottle :disabled` on new formulae either, I guess.

MikeMcQuaid:
> dev-cmd/audit:

Weird commit message? Missing a `squash`?

Library/Homebrew/dev-cmd/audit.rb
commitay: I'm not sure I understand? New formulae won't have a bottle block so we can't check if they should be `bottle :unneeded`?

commitay:
> Missing a `squash`?

Yes, forgot to squash this time.

Library/Homebrew/dev-cmd/audit.rb
MikeMcQuaid: Yeh I’m getting in a muddle. That’s a check that might make sense in ‘brew pull’ for maintainers (either pull a bottle on a new formula or have a bottle unneeded)

MikeMcQuaid: But with bottle disabled: I think this won’t flag it on new formula in core? That seems fine but maybe weird.

ilovezfs:
`bottle :disable` should probably a hard failure in core (perhaps even non-strict but especially for new formulae). It's used in only two formulae currently, which could be whitelisted or deleted, and never allow it.

MikeMcQuaid:
> bottle :disable should probably be a hard audit failure in core (perhaps even non-strict but especially for new formulae). It's used in only two formulae currently, which could be whitelisted or deleted, and we never allow it in new formulae.

Agreed, I'd be 👍 on a non-strict whitelist.

commitay:
Updated with an audit for `bottle :disabled` with a whitelist.

PR #4424, opened 2018-07-05
Refactor staging step for download strategies.

reitermarkus:

-----

Refactor the staging step to explicitly include the destination directory in all commands.

Extracted from https://github.com/Homebrew/brew/pull/4402.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -90,20 +90,6 @@ class AbstractDownloadStrategy
   def quiet_safe_system(*args)
     safe_system(*expand_safe_system_args(args))
   end
-
-  private
-
-  def xzpath
-    "#{HOMEBREW_PREFIX}/opt/xz/bin/xz"
-  end
-
-  def lzippath
-    "#{HOMEBREW_PREFIX}/opt/lzip/bin/lzip"
-  end
-
-  def lhapath
-    "#{HOMEBREW_PREFIX}/opt/lha/bin/lha"
-  end
 end
 
 class VCSDownloadStrategy < AbstractDownloadStrategy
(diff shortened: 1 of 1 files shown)


ilovezfs:
Command output seems to be significantly uglier.
Before
```
==> Verifying gdb-8.1.tar.xz checksum
tar xJf /Users/joe/Library/Caches/Homebrew/gdb-8.1.tar.xz
```
After
```
==> Verifying gdb-8.1.tar.xz checksum
tar -x -J -f /Users/joe/Library/Caches/Homebrew/gdb-8.1.tar.xz -C /private/tmp/gdb-20180705-21040-1vk0rfk
```
I don't think specifying the output directory should have compromised aesthetics.

Also, it calls attention to the useless `-J` which doesn't actually flag any errors. You can pass -J to a gz archive or `-z` to an xz archive and tar will just do whatever it wants anyway. Maybe this isn't true on very old versions of macOS but now we're displaying the useless options with a neon sign.

reitermarkus:
> I don't think specifying the output directory should have compromised aesthetics.

I can't change anything about the `-C /private/tmp/gdb-20180705-21040-1vk0rfk` part, 
and `xJf` vs. `-x -J -f` is debatable.

ilovezfs:
When all that's actually needed is `xf`, I think `xJf` was barely passable, and `-x -J -f` is just ugly.

MikeMcQuaid:
> When all that's actually needed is xf, I think xJf was barely passable, and -x -J -f is just ugly.

Agreed here. Let's just `xf`

MikeMcQuaid:
> Agreed here. Let's just xf

Done in https://github.com/Homebrew/brew/pull/4431

PR #4427, opened 2018-07-05
Download refactoring

reitermarkus:

-----

Clean up some things related to download strategies.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -617,7 +617,7 @@ class SubversionDownloadStrategy < VCSDownloadStrategy
 
   def stage
     super
-    quiet_safe_system "svn", "export", "--force", cached_location, Dir.pwd
+    safe_system "svn", "export", "--force", cached_location, Dir.pwd
   end
 
   def source_modified_time
@@ -794,9 +794,9 @@ class GitDownloadStrategy < VCSDownloadStrategy
     return unless @ref_type == :branch || !ref?
 
     if !shallow_clone? && shallow_dir?
-      quiet_safe_system "git", "fetch", "origin", "--unshallow"
+      safe_system "git", "fetch", "origin", "--unshallow"
     else
-      quiet_safe_system "git", "fetch", "origin"
+      safe_system "git", "fetch", "origin"
     end
   end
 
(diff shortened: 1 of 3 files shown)


Library/Homebrew/download_strategy.rb
```diff
@@ -617,7 +617,7 @@ def fetch
 
   def stage
     super
-    quiet_safe_system "svn", "export", "--force", cached_location, Dir.pwd
```
MikeMcQuaid: Why are these getting unquieted? In the previous PR you only unquieted those that had `-q` flags.

Library/Homebrew/formulary.rb
```diff
@@ -112,10 +112,10 @@ def initialize(bottle_name)
         resource = Resource.new(formula_name) { url bottle_name }
         resource.specs[:bottle] = true
         downloader = CurlDownloadStrategy.new resource.name, resource
-        @bottle_filename = downloader.cached_location
-        cached = @bottle_filename.exist?
+        cached = downloader.cached_location.exist?
         downloader.fetch
         ohai "Pouring the cached bottle" if cached
+        @bottle_filename = downloader.cached_location
```
MikeMcQuaid: What's the reordering about; does this change?

Library/Homebrew/download_strategy.rb
reitermarkus: > Why are these getting unquieted?

These are not quiet currently. `quiet_safe_system` replaces i.e. `{ quiet_flag: '--quiet' }`, which only `cvs` is using.

Library/Homebrew/formulary.rb
reitermarkus: This doesn't change anything at the moment, but this ordering is necessary for what I have in mind for refactoring the download locations.

Instead of relying on a somewhat hacky way to get the proper extension from a URL *before* downloading, we use no extension when initializing the download. We then know the actual file name *after* the download is finished, so in this case `@bottle_filename` should be set only after `downloader.fetch` has been called.

ilovezfs: So currently a cancelled download would have a correct extension – whether guessed in a "hacky" manner or not – and after the change a cancelled download would be extensionless? That sounds like a regression to me.

MikeMcQuaid:
Thanks for explaining all the changes! Looks good to me.

Library/Homebrew/formulary.rb
reitermarkus: What use does a correct extension have when the download is cancelled?

ilovezfs: Most of the downloads can be resumed – please _do not_ remove that ability as it's very useful – but `foo-1.2.3.tar.bz2.incomplete` shouldn't be resumable if the extension changed from bz2 to xz. Also it's much more clear what a `foo-1.2.3.tar.bz2.incomplete` is than what a `foo-1.2.3.incomplete` is.

PR #4453, opened 2018-07-11
Fix `SystemCommand` `:path`.

reitermarkus:

-----

Use `:env` instead of `:path`.


Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -15,46 +15,45 @@
...
-            .to receive(:popen3)
-            .with(an_instance_of(Hash), ["/usr/bin/sudo"] * 2,
-                "-E", a_string_starting_with("PATH="),
-                "A=1", "B=2", "C=3", "--", "env", *env_args, {})
-            .and_wrap_original do |original_popen3, *_, &block|
-              original_popen3.call("/usr/bin/true", &block)
-            end
-
-          subject.run!
+      context "and sudo: true" do
```
claui: I feel this part of the code used to be a bit clearer and easier to read correctly when the contexts were not nested.

A condensed example:

```
Something something
  and its result
Something something with sudo
  and its result
```

is easier to read and understand correctly than:

```
Something something
  and its result
  with sudo
    and its result
```

Also, un-nesting won't cost us anything here since the code is duplicated anyway.

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -190,10 +189,23 @@
...
 
-  describe "given an invalid variable name" do
+  context "when given an invalid variable name" do
     it "raises an ArgumentError" do
       expect { described_class.run("true", env: { "1ABC" => true }) }
         .to raise_error(ArgumentError, /variable name/)
     end
   end
+
+  it "looks for executables in custom PATH" do
```
claui: `in custom PATH` → `in a custom PATH`

Library/Homebrew/test/cask/system_command_spec.rb
```diff
@@ -190,10 +189,23 @@
...
+    mktmpdir do |path|
+      (path/"tool").write <<~SH
+        #!/bin/sh
+        echo Hello, world!
+      SH
+
```
claui: Can we stick with the generic `<<~EOS` for now, and aim to replace it one day in a single swoop if we have decided on alternatives?

It feels a little confusing to get 98 % `EOS` (including lots of code snippets) and 2 % heredocs with other identifiers when I grep for `<<~[A-Z]+$`.

Library/Homebrew/cask/lib/hbc/artifact/installer.rb
```diff
@@ -23,7 +23,7 @@ module ScriptInstaller
         def install_phase(command: nil, **_)
           ohai "Running #{self.class.dsl_key} script '#{path.relative_path_from(cask.staged_path)}'"
           FileUtils.chmod "+x", path unless path.executable?
-          command.run(path, **args, path: PATH.new(HOMEBREW_PREFIX/"bin", HOMEBREW_PREFIX/"sbin", ENV["PATH"]))
+          command.run!(path, **args, env: { "PATH" => PATH.new(HOMEBREW_PREFIX/"bin", HOMEBREW_PREFIX/"sbin", ENV["PATH"]) })
```
claui: Good catch!

I wonder whether we could have caught this earlier with an `artifact/installer_spec.rb`.
I feel this change warrants creating that spec, and have the class under test start out with `PATH=/var/dummy`, then assert that the system command finds both `HOMEBREW_PREFIX/"bin"` and `/var/dummy` in its environment `PATH`.

(No full coverage needed for now – just the tiny bit that’s being changed here!)

Library/Homebrew/test/cask/system_command_spec.rb
reitermarkus: With `SH` I get nested syntax-highlighting in TextMate.

Library/Homebrew/cask/lib/hbc/system_command.rb
```diff
@@ -64,18 +64,22 @@ def command
...
-      env.each do |name, value|
+      variables = env.map do |name, value|
         sanitized_name = Shellwords.escape(name)
         sanitized_value = Shellwords.escape(value)
-        prefix << "#{sanitized_name}=#{sanitized_value}"
+        "#{sanitized_name}=#{sanitized_value}"
       end
 
-      prefix << "--"
+      ["env", *variables]
```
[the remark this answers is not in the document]
reitermarkus: Why? The environment is filtered by default, so this shouldn't matter.

claui:
Thanks again @reitermarkus!

PR #4457, opened 2018-07-12
cask audit: check for `latest` with `auto_updates`

commitay:

-----
Casks having version `:latest` _and_  `auto_updates true` is kind of confusing and doesn't make a difference to `outdated` or `upgrade`. Some of the casks that currently fail this audit are deliberately unversioned (even if they could have an appcast and be versioned) because of update frequency.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/audit.rb
+++ b/Library/Homebrew/cask/lib/hbc/audit.rb
@@ -36,6 +36,7 @@ module Hbc
       check_untrusted_pkg
       check_hosting_with_appcast
       check_latest_with_appcast
+      check_latest_with_auto_updates
       check_stanza_requires_uninstall
       self
     rescue StandardError => e
@@ -197,6 +198,13 @@ module Hbc
       return unless cask.appcast
 
       add_warning "Casks with an appcast should not use version :latest"
+    end
+
+    def check_latest_with_auto_updates
+      return unless cask.version.latest?
+      return unless cask.auto_updates
+
+      add_warning "Casks with `version :latest` should not use `auto_updates`"
     end
 
     def check_hosting_with_appcast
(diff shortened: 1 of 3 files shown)


vitorgalvao:
I agree having both options is redundant for our behaviour, but I also think they should both stay. If we remove it and ever change the behaviour or when a `:latest` cask becomes versioned, we’ll have lost the data.

commitay:
Another motivation for this is https://github.com/Homebrew/brew/pull/4343. I think it will be confusing for `auto_updates` to be displayed in `info` on versioned and latest casks.

commitay:
Some of these probably won't ever be changed to be versioned (e.g. casks in `versions`), have been unversioned because of breakage (`dropbox`) or shouldn't have `auto_updates` because they don't really auto update (`miniconda`  casks).

vitorgalvao:
> I think it will be confusing for `auto_updates` to be displayed in `info` on versioned and latest casks.

On the contrary, it will be useful! It lets us know which `latest` casks might need `--greedy` to upgrade.

> Some of these probably won't ever be changed to be versioned (e.g. casks in `versions`)

True.

> have been unversioned because of breakage (`dropbox`)

Think of the future where we already have in place a bot to auto-update casks and quarantine downloads. Suddenly those casks can be versioned again. We might even special-case them and tell the bot “always check these first” or “check these twice a day”.

> or shouldn't have `auto_updates` because they don't really auto update

Yes, those are mistakes and should be corrected.

One final consideration: in the future we might be able to convert even more casks to be versioned. This is because `appcast` is not the only method to check for updates. One thing that works (I’ve done limited tests) is to query the `url` for a modification date, which the server can return without us needing to download the file.

commitay:
Currently `auto_updates` is treated as information only stanza if it is used with `latest` and as a functional stanza that _alters the behaviour_ of HBC if it is used with a versioned cask. This aims to correct that by enforcing it as only a functional stanza.

vitorgalvao:
Alright, but before we lose the `auto_updates` info in all those casks we should check first if some can be converted from `:latest` now by checking the appcasts retrieved by @core-code.

commitay:
Just one: https://github.com/Homebrew/homebrew-cask/pull/49550

PR #4464, opened 2018-07-13
components_redundancy_cop: check for a head block with only one line

commitay:

-----
Because we don't want formulae with this:
```
  head do
    url "https://github.com/Yubico/yubikey-manager.git"
  end
```


MikeMcQuaid:
Should this not technically be checking if the `head` block only has a `url` rather than only being one line? I don't have strong preferences if you tried that and it was too much of a pain.

commitay:
Actually, we do need to check for the `url` as 
```
  head do
    url "https://github.com/Yubico/yubikey-manager.git",
         :branch => "devel"
  end
```
would be valid with this.

commitay:
Closing for now, I'll take a look at implementing audits inside non-stable blocks rather than making a audit specific for this case.

PR #4479, opened 2018-07-16
Change `Container::me?` to `Container::can_extract?`.

reitermarkus:

-----

Use same signatures for unpack strategies and containers.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/container.rb
+++ b/Library/Homebrew/cask/lib/hbc/container.rb
@@ -6,6 +6,7 @@ require "hbc/container/bzip2"
 require "hbc/container/cab"
 require "hbc/container/criteria"
 require "hbc/container/dmg"
+require "hbc/container/self_extracting_executable"
 require "hbc/container/executable"
 require "hbc/container/generic_unar"
 require "hbc/container/gpg"
@@ -32,6 +33,7 @@ module Hbc
         Ttf,
         Otf,
         Air,
+        SelfExtractingExecutable,
         Cab,
         Dmg,
         SevenZip,
--- a/Library/Homebrew/cask/lib/hbc/container.rb
+++ b/Library/Homebrew/cask/lib/hbc/container.rb
@@ -53,12 +53,18 @@ module Hbc
       # Hbc::Container::GenericUnar
     end
 
-    def self.for_path(path, command)
+    def self.for_path(path)
       odebug "Determining which containers to use based on filetype"
-      criteria = Criteria.new(path, command)
+
+      magic_number = if path.directory?
+        ""
+      else
+        File.binread(path, 262) || ""
+      end
+
       autodetect_containers.find do |c|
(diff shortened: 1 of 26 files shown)


Library/Homebrew/cask/lib/hbc/container.rb
```diff
@@ -53,12 +55,18 @@ def self.autodetect_containers
...
 
-    def self.for_path(path, command)
+    def self.for_path(path)
       odebug "Determining which containers to use based on filetype"
-      criteria = Criteria.new(path, command)
+
+      magic_number = if path.directory?
+        ""
+      else
+        File.binread(path, 262) || ""
```
MikeMcQuaid: Would be good to make `262` a constant.

Library/Homebrew/cask/lib/hbc/container/air.rb
```diff
@@ -3,8 +3,8 @@
 module Hbc
   class Container
     class Air < Base
-      def self.me?(criteria)
-        criteria.path.extname == ".air"
+      def self.can_extract?(path:, magic_number:)
```
MikeMcQuaid: I wouldn't be opposed to having more of these formats live in `unpack_strategy` if it made sense (or at least somewhere they are all shared in a single file)

Library/Homebrew/cask/lib/hbc/container/dmg.rb
```diff
@@ -5,11 +5,13 @@
 module Hbc
   class Container
     class Dmg < Base
-      def self.me?(criteria)
-        !criteria.command.run("/usr/bin/hdiutil",
-                              # realpath is a failsafe against unusual filenames
-                              args:         ["imageinfo", Pathname.new(criteria.path).realpath],
-                              print_stderr: false).stdout.empty?
+      def self.can_extract?(path:, magic_number:)
```
MikeMcQuaid: Much better method name 👍

Library/Homebrew/cask/lib/hbc/container.rb
reitermarkus: It already is in `UnpackStrategy`, this is only temporary.

Library/Homebrew/cask/lib/hbc/container/air.rb
reitermarkus: Well the goal is to move all of these into `unpack_strategy`. Just going step by step here because I think it's easier to review small gradual changes.

Library/Homebrew/cask/lib/hbc/container.rb
MikeMcQuaid: @reitermarkus 👍

Library/Homebrew/cask/lib/hbc/container/air.rb
MikeMcQuaid: @reitermarkus 👍

PR #4484, opened 2018-07-16
Add support for nested archives.

reitermarkus:

-----

Allow extracting nested archives.

### DIFF
--- a/Library/Homebrew/unpack_strategy.rb
+++ b/Library/Homebrew/unpack_strategy.rb
@@ -67,6 +67,30 @@ class UnpackStrategy
     unpack_dir.mkpath
     extract_to_dir(unpack_dir, basename: basename)
   end
+
+  def extract_nestedly(to: nil, basename: nil)
+    if is_a?(UncompressedUnpackStrategy)
+      extract(to: to, basename: basename)
+      return
+    end
+
+    Dir.mktmpdir do |tmp_unpack_dir|
+      tmp_unpack_dir = Pathname(tmp_unpack_dir)
+
+      extract(to: tmp_unpack_dir, basename: basename)
+
+      children = tmp_unpack_dir.children
+
+      if children.count == 1
+        s = self.class.detect(children.first)
+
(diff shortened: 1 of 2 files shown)


Library/Homebrew/unpack_strategy.rb
```diff
@@ -67,6 +67,30 @@ def extract(to: nil, basename: nil)
     unpack_dir.mkpath
     extract_to_dir(unpack_dir, basename: basename)
   end
+
+  def extract_nestedly(to: nil, basename: nil)
```
MikeMcQuaid: Remind me: why is `to: nil, basename: nil` better than `to:, basename:`?

Library/Homebrew/unpack_strategy.rb
```diff
@@ -67,6 +67,30 @@ def extract(to: nil, basename: nil)
...
+      tmp_unpack_dir = Pathname(tmp_unpack_dir)
+
+      extract(to: tmp_unpack_dir, basename: basename)
+
+      children = tmp_unpack_dir.children
+
+      if children.count == 1
+        s = self.class.detect(children.first)
+
+        s.extract_nestedly(to: to, basename: basename)
```
MikeMcQuaid: Any potential for infinite recursion here?

Library/Homebrew/unpack_strategy.rb
reitermarkus: With the latter, the argument are mandatory, with the former they are optional.

Library/Homebrew/unpack_strategy.rb
reitermarkus: I don't think so, since the recursion will stop when `self` is a `UncompressedUnpackStrategy`, which is the default.

Although I think we should not recurse into nested directories.

Library/Homebrew/unpack_strategy.rb
MikeMcQuaid: Gotcha, thanks.

PR #4496, opened 2018-07-17
audit: Relax versioning audits for non-official taps

apjanke:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031). *No; we don't test `audit`.*

-----

I think a couple of our audits are being applied globally, when they are only really relevant to core/"official" taps.

* Requiring versioned formulae to have non-versioned equivalents in the same tap
* Forbidding head/devel

I think both of these are still legit use cases in non-official taps. For versioned formulae, it's legit for a custom tap to make a versioned formula off a regular one from the official core tap. And I think we still allow/support `head` and `devel` usage in non-core taps.

This PR relaxes the audits on non-official taps so that these requirements are only applied to the core tap.

This will make it easier to use audit in non-core taps like Brewsci and `dpo/openblas`, which would want to use it for the style/best-practices guidelines, but have legit use cases for both the scenarios above, and shouldn't be getting false-positives for them.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -310,7 +310,7 @@ module Homebrew
 
       problem "File should end with a newline" unless text.trailing_newline?
 
-      if @versioned_formula
+      if formula.core_formula? && @versioned_formula
         unversioned_formula = begin
           # build this ourselves as we want e.g. homebrew/core to be present
           full_name = if formula.tap


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -267,7 +267,7 @@ def audit_file
 
       problem "File should end with a newline" unless text.trailing_newline?
 
-      if formula.versioned_formula?
+      if @official_tap && formula.versioned_formula?
```
MikeMcQuaid: Mind using `formula.core_formula?` instead?

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -597,7 +597,7 @@ def audit_specs
         end
       end
 
-      if formula.head || formula.devel
+      if @official_tap && (formula.head || formula.devel)
```
MikeMcQuaid: Mind using `formula.core_formula?` instead?

MikeMcQuaid:
> And I think we still allow/support head and devel usage in non-core taps.

I guess the main argument against this is they necessitate a from-source build and HEAD isn't tested by CI.

ilovezfs:
Given the devel/head rule applies only to --new-formula I don't think it makes sense to just remove it for non-core taps. If some taps want to ignore it, then they can but that doesn't make it any less useful as guidance for the majority of taps.

MikeMcQuaid:
> And I think we still allow/support head and devel usage in non-core taps.

I agree this can be removed again now 👍

MikeMcQuaid:
@apjanke What's the latest here?

scpeters:
I'd like to revive this. I'll try to address the outstanding comments.

Library/Homebrew/dev-cmd/audit.rb
scpeters: switched to `formula.core_formula?`, and there are not other instances of `@official_tap` after rebasing

Library/Homebrew/dev-cmd/audit.rb
scpeters: this change was removed

scpeters:
I've changed the title of this to just focus on the versioning audit, and skip the devel/head audit changes, since those had been removed from this PR before it was marked as stale.

MikeMcQuaid:
Makes sense, thanks @apjanke and @scpeters!

@scpeters If you're interested feel free to propose making more audits like this only apply to core 👍

PR #4505, opened 2018-07-17
Use more backports.

reitermarkus:

-----

Add more backports from Ruby 2.4.0 and 2.5.0.


Library/Homebrew/cask/lib/hbc/artifact/pkg.rb
```diff
@@ -2,7 +2,7 @@
 
 require "hbc/utils/hash_validator"
 
-require "vendor/plist/plist"
```
MikeMcQuaid: I think it’s better to use the full require part for these as you can quickly search for which files use vendor files.

MikeMcQuaid:
@reitermarkus What's the latest here?

MikeMcQuaid:
@reitermarkus Think this could be rebased and redone. I think ideally it'd be worth splitting each backport into a separate PR (or separate commit) so we can review the vendor cost vs. code cleanup. It's hard from this as-is to weigh these up because there's so many.

PR #4517, opened 2018-07-19
Rename `Hbc::SystemCommand` to `SystemCommand`.

reitermarkus:

-----

Depends on https://github.com/Homebrew/brew/pull/4488.


Library/Homebrew/readall.rb
```diff
@@ -76,6 +76,13 @@ def syntax_errors_or_warnings?(rb)
       # Retrieve messages about syntax errors/warnings printed to `$stderr`, but
       # discard a `Syntax OK` printed to `$stdout` (in absence of syntax errors).
       messages = Utils.popen_read("#{RUBY_PATH} -c -w #{rb} 2>&1 >/dev/null")
+
+      # Ignore unnecessary warning about named capture conflicts.
+      # See https://bugs.ruby-lang.org/issues/12359.
+      messages = messages.lines
+                         .reject { |line| line.include?("named capture conflicts a local variable") }
```
MikeMcQuaid: Can this use `grep_v`?

Library/Homebrew/extend/os/linux/keg_relocate.rb
```diff
@@ -20,7 +20,9 @@ def change_rpath(file, old_prefix, new_prefix)
     # patchelf requires that the ELF file have a .dynstr section.
     # Skip ELF files that do not have a .dynstr section.
     return if ["cannot find section .dynstr", "strange: no string table"].include?(old_rpath)
-    raise ErrorDuringExecution, "#{cmd_rpath}\n#{old_rpath}" unless $CHILD_STATUS.success?
+    unless $CHILD_STATUS.success?
+      raise ErrorDuringExecution.new(cmd_rpath, status: $CHILD_STATUS, output: [:stdout, old_rpath])
```
MikeMcQuaid: Is there a `safe` method this could use instead?

reitermarkus: There is a `return` between the call and the `raise`, so I don't think so.

MikeMcQuaid: @reitermarkus Yeh, good point. I guess something could be `yield`ed but definitely overkill for one example.

reitermarkus:
Tests are passing on Linux now.

Issue #4530, opened 2018-07-22
Upgrade to 1.7.0 triggers unintended Cellar directory permissions change


philn:
**Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

- [x] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [ ] ran `brew update` and can still reproduce the problem?
- [ ] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [ ] ran `brew config` and `brew doctor` and included their output with your issue?

Steps to reproduce:

- `ls -l /usr/local | grep Cellar` -> `drwxr-xr-x  155 admin  staff   4960 Jul 21 13:47 Cellar`
- `cd /usr/local/Homebrew`
- `git checkout 1.6.13`
- `brew upgrade`
- `ls -l /usr/local | grep Cellar` -> `drwx------  148 admin  staff   4736 Jul 22 14:57 Cellar`

Now applications are unavailable for users other than `admin`. See also https://discourse.brew.sh/t/cellar-restrictive-permissions/2622/7

MikeMcQuaid:
Thanks for the report @philn!

An easier reproduction here seems to be:
```bash
chmod 755 /usr/local/Cellar
ls -l /usr/local | grep Cellar
brew reinstall wget
ls -l /usr/local | grep Cellar
```

Checking out an older revision does not seem to be necessary.

philn:
Oh, interesting. Please let me know if other informations are needed. I'd be happy to help fixing this issue but I'm not very fluent in Ruby, unfortunately.

MikeMcQuaid:
This only affects installation of bottles. The relevant problematic line is:
[...]

PR #4532, opened 2018-07-22
Fix directory permissions when unpacking.

reitermarkus:

-----

Fixes https://github.com/Homebrew/brew/issues/4530.

### DIFF
--- a/Library/Homebrew/unpack_strategy.rb
+++ b/Library/Homebrew/unpack_strategy.rb
@@ -96,7 +96,9 @@ class DirectoryUnpackStrategy < UnpackStrategy
   private
 
   def extract_to_dir(unpack_dir, basename:, verbose:)
-    FileUtils.cp_r File.join(path, "."), unpack_dir, preserve: true, verbose: verbose
+    path.children.each do |child|
+      FileUtils.copy_entry child, unpack_dir/child.basename, true, false
+    end
   end
 end
 


ilovezfs:
I think it would be preferable not to flip it back and forth between modes even if it ends up correct at the end.

Library/Homebrew/unpack_strategy.rb
```diff
@@ -96,7 +96,13 @@ def self.can_extract?(path:, magic_number:)
   private
 
   def extract_to_dir(unpack_dir, basename:)
-    FileUtils.cp_r File.join(path, "."), unpack_dir, preserve: true
+    directory_permissions = File.stat(unpack_dir).mode
+
+    begin
+      FileUtils.cp_r File.join(path, "."), unpack_dir, preserve: true
+    ensure
+      FileUtils.chmod directory_permissions, unpack_dir
```
ilovezfs: Can this fail on EPERM or similar?

Should this be running the chmod in the no-op case?

MikeMcQuaid:
> I think it would be preferable not to flip it back and forth between modes even if it ends up correct at the end.

Agreed 👍. I think bottle permissions can be respected on extraction but only for the contents of the keg. The keg directories could be created manually and the recursive copy just be for the files within there.

reitermarkus:
> the recursive copy just be for the files within there.

I tried to do it this way, but I don't know how to do this.

`FileUtils.cp_r` will follow symlinks, and `FileUtils.copy_entry`, which doesn't, does not work if the destination already exists.

reitermarkus:
Nevermind, apparently `FileUtils.copy_entry` *does*  work.

PR #4543, opened 2018-07-24
Pass `verbose` to unpack strategies.

reitermarkus:

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -333,7 +333,8 @@ end
 class NoUnzipCurlDownloadStrategy < CurlDownloadStrategy
   def stage
     UnpackStrategy::Uncompressed.new(cached_location)
-                                .extract(basename: basename_without_params)
+                                .extract(basename: basename_without_params,
+                                         verbose: ARGV.verbose? && !shutup)
   end
 end
 
--- a/Library/Homebrew/system_command.rb
+++ b/Library/Homebrew/system_command.rb
@@ -28,8 +28,9 @@ class SystemCommand
   end
 
   def run!
+    puts command.shelljoin.gsub(/\\=/, "=") if verbose? || ARGV.debug?
+
     @merged_output = []
-    odebug command.shelljoin
 
     each_output_line do |type, line|
       case type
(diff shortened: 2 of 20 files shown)


Library/Homebrew/download_strategy.rb
```diff
@@ -330,7 +330,7 @@ def _fetch
 class NoUnzipCurlDownloadStrategy < CurlDownloadStrategy
   def stage
     UnpackStrategy::Uncompressed.new(cached_location)
-                                .extract(basename: basename_without_params)
+                                .extract(basename: basename_without_params, verbose: ARGV.verbose? && !shutup)
```
MikeMcQuaid: This line is a bit long. `verbose` could be extracted to be a variable, perhaps?

Library/Homebrew/system_command.rb
```diff
@@ -46,13 +47,14 @@ def run!
     result
   end
 
-  def initialize(executable, args: [], sudo: false, input: [], print_stdout: false, print_stderr: true, must_succeed: false, env: {}, **options)
+  def initialize(executable, args: [], sudo: false, input: [], print_stdout: false, print_stderr: true, verbose: false, must_succeed: false, env: {}, **options)
```
MikeMcQuaid: Can this line be shrunk so it doesn't wrap on the GitHub UI (120 chars, iirc).

Library/Homebrew/unpack_strategy/air.rb
```diff
@@ -13,10 +13,9 @@ def dependencies
     private
 
     def extract_to_dir(unpack_dir, basename:, verbose:)
-      system_command!(
-        "/Applications/Utilities/Adobe AIR Application Installer.app/Contents/MacOS/Adobe AIR Application Installer",
-        args: ["-silent", "-location", unpack_dir, path],
-      )
+      system_command! "/Applications/Utilities/Adobe AIR Application Installer.app/Contents/MacOS/Adobe AIR Application Installer",
```
MikeMcQuaid: Can this line be shrunk so it doesn't wrap on the GitHub UI (120 chars, iirc).

chdiza:
I think something that made unpacking verbose was done before, and got reverted because it was irritating.  I'm too lazy to go look through the legacy repo, but I think @ilovezfs might have been the one who reverted it.

ilovezfs:
Yeah dumping the list of every file in a source archive is not exactly helpful.

MikeMcQuaid:
> Yeah dumping the list of every file in a source archive is not exactly helpful.

Does this actually do that or does it just output the command names that are used?

reitermarkus:
I think there are two separate questions here:

 - Should `--debug` imply `--verbose`?
 - Should every command be shown when `--verbose`, or just the ones we specifically pass `verbose` to?

MikeMcQuaid:
> Should every command be shown when --verbose/--debug, or just the ones we specifically pass verbose to?

To be more explicit here: by "every command" do you mean "the executed command and arguments" or "the output from the command" (or both/neither)?

reitermarkus:
I mean “the executed command and arguments”.

MikeMcQuaid:
Cool. I think it should be output by both debug and verbose with perhaps output also being output by verbose.

reitermarkus:
> I think it should be output by both debug and verbose

Good, this is the current state:

```ruby
verbose? || ARGV.debug?
```

PR #4545, opened 2018-07-25
Make `HOMEBREW_TAP_PATH_REGEX` also match exact path.

reitermarkus:

---

This way `Tap::from_path` also works for the exact path of a tap.

### DIFF
--- a/Library/Homebrew/tap_constants.rb
+++ b/Library/Homebrew/tap_constants.rb
@@ -5,6 +5,6 @@ HOMEBREW_TAP_CASK_REGEX = %r{^([\w-]+)/([\w-]+)/([a-z0-9\-]+)$}
 # match taps' directory paths, e.g. HOMEBREW_LIBRARY/Taps/someuser/sometap
 HOMEBREW_TAP_DIR_REGEX = %r{#{Regexp.escape(HOMEBREW_LIBRARY)}/Taps/(?<user>[\w-]+)/(?<repo>[\w-]+)}
 # match taps' formula paths, e.g. HOMEBREW_LIBRARY/Taps/someuser/sometap/someformula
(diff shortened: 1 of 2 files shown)


Library/Homebrew/tap_constants.rb
```diff
@@ -5,6 +5,6 @@
 # match taps' directory paths, e.g. HOMEBREW_LIBRARY/Taps/someuser/sometap
 HOMEBREW_TAP_DIR_REGEX = %r{#{Regexp.escape(HOMEBREW_LIBRARY)}/Taps/(?<user>[\w-]+)/(?<repo>[\w-]+)}
 # match taps' formula paths, e.g. HOMEBREW_LIBRARY/Taps/someuser/sometap/someformula
-HOMEBREW_TAP_PATH_REGEX = Regexp.new(HOMEBREW_TAP_DIR_REGEX.source + %r{/(.*)}.source)
+HOMEBREW_TAP_PATH_REGEX = Regexp.new(HOMEBREW_TAP_DIR_REGEX.source + %r{(?:/(.*)|$)}.source)
```
MikeMcQuaid: Can you provide an example string that doesn't match this? These two regexes look like they should be identical to me?

reitermarkus: The first one only matches sub-paths, the second one matches the exact path, without a trailing slash.

reitermarkus: It's not the same. `/?(.*)` is the same as just `(.*)`.

MikeMcQuaid: Sorry, I mean `/?(.*)$`

reitermarkus: That's still the same as `(.*)$`. The `?` should be around the whole thing:`(?:/(.*))?$`

MikeMcQuaid: Yeh, the greediness is different to what I thought 👍

PR #4552, opened 2018-07-25
Reorder unpack strategies.

reitermarkus:

-----

Addresses https://github.com/Homebrew/brew/pull/4541#issuecomment-407681345.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -47,8 +47,11 @@ class AbstractDownloadStrategy
   # chdir into the newly-unpacked directory.
   # Unlike {Resource#stage}, this does not take a block.
   def stage
-    UnpackStrategy.detect(cached_location, ref_type: @ref_type, ref: @ref)
+    UnpackStrategy.detect(cached_location,
+                          extension_only: true,
+                          ref_type: @ref_type, ref: @ref)
                   .extract_nestedly(basename: basename_without_params,
+                                    extension_only: true,
                                     verbose: ARGV.verbose? && !shutup)
   end
 
--- a/Library/Homebrew/unpack_strategy.rb
+++ b/Library/Homebrew/unpack_strategy.rb
@@ -43,17 +43,17 @@ module UnpackStrategy
       Xar,
       Ttf,
       Otf,
-      Executable,
       Git,
       Mercurial,
       Subversion,
       Cvs,
+      SelfExtractingExecutable, # needs to be before Cab
+      Cab,
+      Executable,
       Dmg, # needs to be before Bzip2
       Bzip2,
       Fossil,
       Bazaar,
-      SelfExtractingExecutable, # needs to be before Cab
-      Cab,
       Compress,
       P7Zip,
       Sit,
@@ -77,19 +77,14 @@ module UnpackStrategy
   end
 
   def self.from_path(path)
-    strategy = strategies.detect do |s|
-      s.can_extract?(path)
-    end
+    strategy = strategies
+               .sort_by { |s| s.extensions.map(&:length).max(0) }
+               .reverse
+               .detect { |s| s.extensions.include?(path.extname) }
 
-    # This is so that bad files produce good error messages.
-    strategy ||= case path.extname
-    when ".tar", ".tar.gz", ".tgz", ".tar.bz2", ".tbz", ".tar.xz", ".txz"
-      Tar
-    when ".zip"
-      Zip
-    else
-      Uncompressed
-    end
+    strategy ||= strategies.detect { |s| s.can_extract?(path) }
+
+    strategy ||= Uncompressed
 
     strategy
   end
--- a/Library/Homebrew/unpack_strategy.rb
+++ b/Library/Homebrew/unpack_strategy.rb
@@ -10,6 +10,7 @@ module UnpackStrategy
       Otf,
       Air,
       Executable,
+      Diff,
       Jar, # needs to be before Zip
       LuaRock, # needs to be before Zip
       MicrosoftOfficeXml, # needs to be before Zip
@@ -126,6 +127,7 @@ require "unpack_strategy/bzip2"
 require "unpack_strategy/cab"
 require "unpack_strategy/compress"
 require "unpack_strategy/cvs"
+require "unpack_strategy/diff"
 require "unpack_strategy/directory"
 require "unpack_strategy/dmg"
 require "unpack_strategy/executable"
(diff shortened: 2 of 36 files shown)


MikeMcQuaid:
@reitermarkus What would be the disadvantages of making all the existing Homebrew supported download types first in this list whenever possible? Homebrew formulae are more widely used than Homebrew casks so it seems sensible to optimise when possible for the most common case (a tarball).

reitermarkus:
> What would be the disadvantages of making all the existing Homebrew supported download types first in this list whenever possible?

Reordered them again and added comments for those that need to be in a specific order.

MikeMcQuaid:
I think it’s worth stepping back and asking if the whack-a-mole is worth it. What does this strategy get us compared to the previous heuristic based on file name extension? Could that heuristic be used by default and just fall back to this method if it fails?

reitermarkus:
> What does this strategy get us compared to the previous heuristic based on file name extension?

This way we don't have to deal with file extensions altogether, since they may be wrong sometimes.

MikeMcQuaid:
@reitermarkus The vast majority of the time the file extensions aren't wrong, though. We should assume they are correct until they prove otherwise and then fall back to the heuristic here (where the ordering doesn't matter as much if the filename is already incorrect). Some logic that looks at file extension and then tries e.g. the tarball strategy first if it's a `.tar.*` would seem to solve the issues @ilovezfs has mentioned while still allowing this to work for detecting anything regardless of naming.

reitermarkus:
> Some logic that looks at file extension and then tries e.g. the tarball strategy first if it's a `.tar.*`

This wouldn't work if the extension is wrong, which is exactly the problem. It would be first detected as something which it is not, and then would fail. Only then would you know that the extension is wrong. And then you have to try again to find the correct unpack strategy. You would have to do all of this instead of doing it without extension in the first place.

ilovezfs:
If the extension is "wrong" it's acceptable to have to specify the download strategy or report the issue upstream. For example, there was that one .zip file that was actually a jar. It is fine, and perhaps even preferable, in that case to need to add `:using => :nounzip`.

MikeMcQuaid:
> This wouldn't work if the extension is wrong, which is exactly the problem. It would be first detected as something which it is not, and then would fail. Only then would you know that the extension is wrong. And then you have to try again to find the correct unpack strategy. You would have to do all of this instead of doing it without extension in the first place.

Yes. I'm proposing the failure not be exposed to end-users but instead surfaced only to `HOMEBREW_DEVELOPER` with an `opoo` so it can be reported upstream. Either way the detection only needs run once; it's just a question of whether you assume that most files are named correctly or most files are named incorrectly. The current implementation is hugely more pessimistic that reality; it optimises for the corner case and introduces additional complexity, shelling out (therefore likely runtime cost) and output to simplify the error handling on a minority of cases. To be clear: it's a pretty smart and neat implementation but it's a much better fallback than primary logic.

reitermarkus:
The current implementation works for both cases, when the extension is right, and when it is wrong, so I don't see any benefit in changing it now to use extensions.

> it optimises for the corner case and introduces additional complexity

This is less complex than handling extensions first, and only then falling back to this approach.

> therefore likely runtime cost

I think this is negligible in comparison to the actual extraction time. And with the latest reordering it will shell out three times at most, if at all.

MikeMcQuaid:
> The current implementation works for both cases, when the extension is right, and when it is wrong, so I don't see any benefit in changing it now to use extensions.

It has required a bunch of whack-a-mole to figure out the right order. I am not confident this will never have to happen in future and that further future reordering will not cause regressions.

> This is less complex than handling extensions first, and only then falling back to this approach.
> I think this is negligible in comparison to the actual extraction time. And with the latest reordering it will shell out three times at most, if at all.

It is but I do not agree that the "handle incorrect filenames" is actually necessary at all; it feels like a solution in search of a problem. It's perhaps something that's useful for Homebrew Cask (was this behaviour previously there in Cask's download strategies?) but it does not seem useful for formulae. Additionally, if it means silently allowing a e.g. `.tar.gz` actually be a more complex type which requires decompressed with a non-system tool it's a potential vulnerability.

MikeMcQuaid:
> It has required a bunch of whack-a-mole to figure out the right order. I am not confident this will never have to happen in future and that further future reordering will not cause regressions.

> I don't think it's ever acceptable to see hdiutil imageinfo running for formulae, which will never use a dmg, and yet this is still the case even in the latest iteration of the PR:

This has proved my point. Please make this go back to the behaviour where it's primarily using file extensions to identify the tool to use. This auto-detection behaviour was never discussed and at this point if it was not so involved to revert I would revert the behaviour myself. Whether there is a (non-silent) fallback to the auto-detection behaviour on an extraction failure is up to you but this PR will not be merged until this is done.

reitermarkus:
Okay, it's now checking for extensions first, then falls back to the other method.

MikeMcQuaid:
> Okay, it's now checking for extensions first, then falls back to the other method.

Thanks 🙇. What is the output like when the extension detection fails and it falls back to auto-detection?

reitermarkus:
> What is the output like when the extension detection fails and it falls back to auto-detection?

Same as it was before.

MikeMcQuaid:
@reitermarkus Can this be addressed so the fallback is non-silent e.g. for `ARGV.homebrew_developer?` it uses `opoo` to output a warning about the incorrect file extension (or even a `raise`).

reitermarkus:
No, since this would break hundreds of casks which have no extension at all. This is why I don't want to rely on extensions.

MikeMcQuaid:
@reitermarkus That behaviour should be in place for formulae only, in that case. Can you provide an example of some casks without extensions?

reitermarkus:
Basically any cask with a URL ending with `/download`, for example.

MikeMcQuaid:
@reitermarkus Presumably at least some of those provide an `Content-Disposition: attachment; filename=` header? Is there a reason that information cannot be used?

reitermarkus:
That's the goal, but this doesn't work with the current download strategies.

MikeMcQuaid:
I'm 👍 when @ilovezfs is too.

ilovezfs:
Still doing the same even after the latest commits.

reitermarkus:
Now, finally.

MikeMcQuaid:
@reitermarkus Merge when ready.

philipjameson:
So, is there any way to specify an unpack strategy now? We have a formula in a tap that downloads from a rest endpoint that we don't control and that doesn't include content-disposition information, and now when we try to build, it just downloads the tarball and puts it in the working dir. This looks to be because there is now the 'extension_only' option set to true in download_strategy. Removing that in download_strategy.rb led to things working properly. Example formula is here: https://github.com/facebook/homebrew-fb/blob/master/buck.rb

I'm just going to hack in a tar command for now, but removing this logic makes it so we can't actually build our bottles now.

PR #4567, opened 2018-07-30
Revert "Use `SystemCommand` for `curl`."

ilovezfs:
Reverts Homebrew/brew#4561


ilovezfs:
Reverting since https://github.com/Homebrew/brew/pull/4561 broke `brew test-bot --ci-upload`, broke installation of at least `ant` and `vagrant`, and caused unexpected `brew fetch` output.

PR #4570, opened 2018-07-30
Fix for Linux make shim

maxim-belkin:

-----

The `make` shim for Linux had a bug in that it was removing `Library/Homebrew/shims/linux/super` directory from `PATH` for the duration of execution of `make`. This caused some undesired results. 

Here, I moved the code that removes `$HOMEBREW_LIBRARY/Homebrew/shims/linux/super` directory from PATH to within the `if/else` clause to emphasize the fact that we have to do it only to detect the "real" `make` executable.

Props for identifying the issue go to @xu-cheng!

### DIFF
--- a/Library/Homebrew/shims/linux/super/make
+++ b/Library/Homebrew/shims/linux/super/make
@@ -16,7 +16,7 @@ then
 else
   SAVED_PATH="$PATH"
   pathremove "$HOMEBREW_LIBRARY/Homebrew/shims/linux/super"
-  export MAKE="$(command -v make)"
+  export MAKE="$(which make)"
   export PATH="$SAVED_PATH"
 fi
 


MikeMcQuaid:
Has this already been merged to Linuxbrew? I think it would be good on Linux-specific PRs to identify which of them have been already merged and tagged in Linuxbrew and for which this is the first time the code will be used. I was a bit disappointed that the previous `make` shim PR had not been sufficiently tested on Linux.

iMichka:
It never got a chance to get tested on the CI because we had almost 2 weeks without bottling, due to various broken things, which took me forever to fix. I am currently building some bottles with this change right now, I'll come back to you once I am confident enough this works correctly.

MikeMcQuaid:
> Just one small comment, should we not use command -v instead of which?

Agreed 👍

maxim-belkin:
> I was a bit disappointed that the previous make shim PR had not been sufficiently tested on Linux.

My fault, sorry.

> > should we not use `command -v` instead of `which`
>
> Agreed 👍 

`command -v` catches aliases. `type -pP`?

iMichka:
With zsh: `type -pP xxx` gives: `type: bad option: -P`.

maxim-belkin:
:) You’re either joking or missed the shebang on line 1

iMichka:
Woups sorry, didn't see the shebang. I was playing around with the `type` options in my terminal, outside of the brew context :)

I am OK with `type -pP` then.

jonchang:
I think this is probably good to be re-reviewed? Bottles have been building on Linux for a while with this change.

Library/Homebrew/shims/linux/super/make
```diff
@@ -21,6 +10,16 @@ pathremove () {
...
 }
 
-pathremove "$HOMEBREW_LIBRARY/Homebrew/shims/linux/super"
+if [[ -n "$HOMEBREW_MAKE" && "$HOMEBREW_MAKE" != "make" ]]
+then
+  export MAKE="$HOMEBREW_MAKE"
+else
+  SAVED_PATH="$PATH"
+  pathremove "$HOMEBREW_LIBRARY/Homebrew/shims/linux/super"
+  export MAKE="$(which make)"
```
sjackman: Please change `which` to `command -v` as suggested by @iMichka.

sjackman: I'm happy with `command -v` (rather than `type -pP`).

maxim-belkin: See my comment above https://github.com/Homebrew/brew/pull/4570#issuecomment-409275957

sjackman: I don't believe aliases affect non-interactive shells. I'm okay with using the more typical idiom `command -v`.

maxim-belkin: true! forgot we're doing this in a script

jonchang:
I restarted travis, seems like a network error

MikeMcQuaid:
Leaving this as `which` for now until the change has been made/tested in Linuxbrew/brew. I'd like for anything like this to go through testing in Linuxbrew/brew until (hopefully relatively soon!) Homebrew/brew becomes the default for Linuxbrew users.

PR #4586, opened 2018-08-01
Fix `ErrorDuringExecution` initialization.

reitermarkus:

### DIFF
--- a/Library/Homebrew/utils/popen.rb
+++ b/Library/Homebrew/utils/popen.rb
@@ -5,8 +5,8 @@ module Utils
 
   def self.safe_popen_read(*args, **options, &block)
     output = popen_read(*args, **options, &block)
-    raise ErrorDuringExecution(args, stdout: output, status: $CHILD_STATUS) unless $CHILD_STATUS.success?
-    output
+    return output if $CHILD_STATUS.success?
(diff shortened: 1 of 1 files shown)


Library/Homebrew/utils/popen.rb
```diff
@@ -5,7 +5,7 @@ def self.popen_read(*args, **options, &block)
 
   def self.safe_popen_read(*args, **options, &block)
     output = popen_read(*args, **options, &block)
-    raise ErrorDuringExecution(args, stdout: output, status: $CHILD_STATUS) unless $CHILD_STATUS.success?
+    raise ErrorDuringExecution.new(args, status: $CHILD_STATUS, output: [[:stdout, output]]) unless $CHILD_STATUS.success?
```
MikeMcQuaid: Why is the double array needed? Feels a bit weird. Also, can this use a block `unless` (and below) because it's longer than the GitHub max width now (would be good to get RuboCop to enforce that 🔜)

reitermarkus: It's an array of lines, which are either labelled `:stdout` or `:stderr`.

reitermarkus: It's because `SystemCommand` collects `stdout` and `stderr` separately, which either ends up in an `ErrorDuringExecution` (which outputs the combined output), or in a `Result` (which has `#stdout` and `#stderr` methods).

PR #4587, opened 2018-08-01
Use `ditto` instead of `unzip`.

reitermarkus:

-----

There had to be a reason we were using `ditto` before, here it is:

https://github.com/Homebrew/homebrew-cask-drivers/pull/453.

### DIFF
--- a/Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
+++ b/Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
@@ -1,0 +1,9 @@
+module UnpackStrategy
+  class Zip
+    def extract_to_dir(unpack_dir, basename:, verbose:)
+      # `ditto` keeps Finder attributes intact and does not skip volume labels
+      # like `unzip` does, which can prevent disk images from being unzipped.
+      system_command! "ditto", args: ["-x", "-k", path, unpack_dir]
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


commitay:
> Fixes https://github.com/Homebrew/homebrew-cask-drivers/pull/453.

This will close the linked PR.

reitermarkus:
> This will close the linked PR.

Right, fixed.

commitay:
Is `ditto` okay for formulae?

reitermarkus:
> Is `ditto` okay for formulae?

I don't see why not, it does the same thing.

MikeMcQuaid:
> There had to be a reason we were using ditto before, here it is:
> 
> Homebrew/homebrew-cask-drivers#453.

@reitermarkus Can you summarise this a bit? I'm a bit lost.

reitermarkus:
Apparently `unzip` skips files which have a volume label by default. This seems to be related to this option (from `man unzip`):

```
       -$     [MS-DOS, OS/2, NT] restore the volume label  if  the  extraction
              medium  is  removable  (e.g.,  a diskette).  Doubling the option
              (-$$) allows fixed media (hard disks) to be  labelled  as  well.
              By default, volume labels are ignored.
```

However, `unzip -\$ <path>` doesn't work. `ditto` works as expected and extracts all files.

PR #4588, opened 2018-08-01
Don’t truncate headlines.

reitermarkus:

-----

I never understood why this is a thing. We only do this for headlines, which are supposed to be short to begin with, and if they are not, it is probably because they contain a URL, which should still always be visible in full.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -548,8 +548,7 @@ class FormulaInstaller
     if deps.empty? && only_deps?
       puts "All dependencies for #{formula.full_name} are satisfied."
     elsif !deps.empty?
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
I'd rather this is left as-is. These are intentionally truncated and used for things where you don't need to see the whole output (e.g. it's repeating what you already typed on the terminal). If you need the non-truncated behaviour you should avoid using `oh1`.

reitermarkus:
I disagree, headlines are supposed to be used to convey important information, so it is counter-intuitive to hide part of that information.

> These are intentionally truncated

Why though? There is absolutely no reason for this, given that every other output isn't.

> things where you don't need to see the whole output (e.g. it's repeating what you already typed on the terminal)

Where is this the case? If I don't need to see the whole output, I might as well not need to see any.

PR #4593, opened 2018-08-02
Merge `curl` options used by casks into download strategies.

reitermarkus:

-----

Add `curl` options from cask download strategy to `CurlDownloadStrategy`.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -1,3 +1,4 @@
+require "cgi"
 require "json"
 require "rexml/document"
 require "time"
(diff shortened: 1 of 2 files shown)


Library/Homebrew/download_strategy.rb
```diff
@@ -323,7 +342,13 @@ def _fetch
 # Query parameters on the URL are converted into POST parameters
 class CurlPostDownloadStrategy < CurlDownloadStrategy
   def _fetch
-    base_url, data = @url.split("?")
+    base_url, data = if meta.key?(:data)
+      escape_data = ->(k, v) { ["-d", "#{CGI.escape(k.to_s)}=#{CGI.escape(v.to_s)}"] }
```
MikeMcQuaid: Can the `"#{CGI.escape(k.to_s)}=#{CGI.escape(v.to_s)}"` bit maybe be a shared function where it's used above too?

reitermarkus: I'd say let's go with the rule of three on this one.

reitermarkus: Made me think, that there should be something like this for `URI`. And there is: `URI.encode_www_form`

MikeMcQuaid: Nice!

PR #4601, opened 2018-08-03
Use `--` to separate download name and version.

reitermarkus:

-----

Use `--` to separate the name and version for cached downloads. Cask download strategies and VCS strategies are already using this, as well as patches.


MikeMcQuaid:
How will this effect existing entries in the cache?

reitermarkus:
The compatibility layer will rename them if accessed, and `brew cleanup` already uses a regex which also matches the old pattern.

MikeMcQuaid:
@reitermarkus I think it would be good to have some sort of migration which avoids having to redownload anything (particularly SCM checkouts which can be a bit slow).

reitermarkus:
> particularly SCM checkouts which can be a bit slow

The SCM strategies are already using `--` as separator, so these aren't affected.

MikeMcQuaid:
Gotcha 👍. Still think a `brew update` migration might be nice if it's possible.

reitermarkus:
Added the migration to `brew update` instead of `compat`. There is also the `migrate_legacy_cache_if_necessary` method, should we reduce this to just delete the legacy cache? I don't think the files in the legacy cache are relevant anymore.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -183,6 +189,23 @@ def migrate_legacy_cache_if_necessary
...
+    HOMEBREW_CACHE.children.each do |child|
+      next unless child.file?
+
+      next unless /^(?<prefix>[^\.]+[^\-])\-(?<suffix>[^\-].*)/ =~ child.basename.to_s
+      target = HOMEBREW_CACHE/"#{prefix}--#{suffix}"
+
+      if target.exist?
+        FileUtils.rm_r child
+      else
+        FileUtils.mv child, target
```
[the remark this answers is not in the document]
reitermarkus: Changed it to use the `--force` versions of these.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -183,6 +189,23 @@ def migrate_legacy_cache_if_necessary
...
+    HOMEBREW_CACHE.children.each do |child|
+      next unless child.file?
+
+      next unless /^(?<prefix>[^\.]+[^\-])\-(?<suffix>[^\-].*)/ =~ child.basename.to_s
+      target = HOMEBREW_CACHE/"#{prefix}--#{suffix}"
+
+      if target.exist?
+        FileUtils.rm_rf child
+      else
+        FileUtils.mv child, target, force: true
```
MikeMcQuaid: I still reckon it's worth handling permission exceptions here a bit more nicely e.g. from below:

```ruby
     rescue Errno::EACCES, Errno::ENOENT
       ofail <<~EOS
         Could not create symlink at #{dst}!
         Please do this manually with:
           sudo ln -sf #{src} #{dst}
           sudo chown $(whoami) #{dst}
       EOS
     end
```

reitermarkus: Added warnings for permissions exceptions.

MikeMcQuaid:
Last thing: can you have a read of https://github.com/Homebrew/brew/issues/568 before merging and just make sure there's no way that this will or can help with any of the issues described there. Thanks again and 🚢 when ready!

ilovezfs:
The migration seems to have resulted in double `--` when `-` is part of the name. For example, I now have a file
```
reattach--to--user--namespace--2.7.tar.gz
```

But fetching it again with `brew fetch -s reattach-to-user-namespace` results in
```
reattach-to-user-namespace--2.7.tar.gz
```
and the previously cached item is not found.

Also, `bump-formula-pr` seems to be downloading with double hyphen. For example, I just ended up with `firebase--cli--4.0.3.tgz` in my cache when opening https://github.com/Homebrew/homebrew-core/pull/30824

ilovezfs:
I think this may have broken the use of hard links in test-bot https://github.com/Homebrew/homebrew-test-bot/blob/59c69a1bbc8f29daacd8b2d58c851a062a35c88c/cmd/brew-test-bot.rb#L781

The bottle file name hard linked into the cache won't be the same as what you'd get from a download so I don't think there will be a cache hit.

PR #4610, opened 2018-08-06
Fix `update-report` migration.

reitermarkus:

-----

Fixes https://github.com/Homebrew/brew/pull/4601#issuecomment-410592455.

### DIFF
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -198,6 +198,8 @@ module Homebrew
       next unless /^(?<prefix>[^\.]+[^\-])\-(?<suffix>[^\-].*)/ =~ child.basename.to_s
       target = HOMEBREW_CACHE/"#{prefix}--#{suffix}"
 
+      next if suffix.include?("--") && !suffix.start_with?("patch")
+
       if target.exist?
         begin
           FileUtils.rm_rf child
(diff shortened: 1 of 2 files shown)


ilovezfs:
I think already downloaded bottles will still be migrated to double dash inadvertently.

reitermarkus:
Should we revert to using the `Compat` module? This will now miss resources, because they are ambiguous with formula names with dashes.

ilovezfs:
```
-rw-r--r--    1 joe  staff    5102264 Jun 21 10:34 git--2.18.0.tar.xz
-rw-r--r--    1 joe  staff    1201076 Jun 21 10:34 git--html--2.18.0.tar.xz
-rw-r--r--    1 joe  staff     425516 Jun 21 10:34 git--man--2.18.0.tar.xz
-rw-r--r--    1 joe  staff   13894061 Jun 21 11:39 git--2.18.0.high_sierra.bottle.tar.gz
```
Are those all correct?

reitermarkus:
Yes, if `html` and `man` are resources. But if you update with this PR, I think it will miss the resources and leave them as `git-html` and `git-man`.

ilovezfs:
Odd ... it seems to have converted them even with the PR.

ilovezfs:
So we're going to have bottle files with one name on Bintray and a different name in the cache?

ilovezfs:
For instance, the firebase-cli 4.0.3 bottles are still single hyphen on Bintray even though they were built after the original PR: https://bintray.com/homebrew/bottles/firebase-cli#files

reitermarkus:
Yes, this only pertains to downloads, and we don't rely on the original filename for downloads. It's much easier to manage the cache cleanup with this format.

ilovezfs:
That's pretty unfortunate as it will make testing bottles by downloading them from Jenkins before they're published more difficult.

PR #4613, opened 2018-08-06
Change migration to loop through formulae.

reitermarkus:

-----

The regex in `migrate_cache_entries_to_double_dashes` will never catch all edge cases.

### DIFF
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -192,25 +192,41 @@ module Homebrew
   def migrate_cache_entries_to_double_dashes(initial_version)
     return if initial_version > "1.7.1"
 
-    HOMEBREW_CACHE.children.each do |child|
-      next unless child.file?
+    Formula.each do |formula|
+      specs = [*formula.stable, *formula.devel, *formula.head]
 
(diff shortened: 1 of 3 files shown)


Library/Homebrew/compat/download_strategy.rb
```diff
@@ -0,0 +1,16 @@
+class AbstractFileDownloadStrategy
+  # TODO: This can be removed after a month because downloads
```
MikeMcQuaid: I don't agree with this; I'd rather we kept around something indefinitely as there's plenty of things that get updated less than once a month. Can the same logic not live in `brew update` indefinitely?

reitermarkus:
Another solution would be to loop through all formulae in the migration and to the same thing the `Compat` module does.

MikeMcQuaid:
> Another solution would be to loop through all formulae in the migration and to the same thing the Compat module does. This way it doesn't rely on a regex.

@reitermarkus Sounds good to me as long as the no-op case is fast 👍

reitermarkus:
> Sounds good to me as long as the no-op case is fast

Not sure about the performance, but it will be a no-op when updating from `> 1.7.1`.

MikeMcQuaid:
@reitermarkus Sounds good.

PR #4620, opened 2018-08-06
Add analytics to cask installs

brianmorton:

-----
per https://github.com/Homebrew/homebrew-cask/issues/47720

seems odd that it might be this simple... I welcome all feedback and help.

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -90,6 +90,8 @@ module Hbc
       install_artifacts
       enable_accessibility_access
 
+      Utils::Analytics.report_event("cask_install", @cask.token)
+
       puts summary
     end
 
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -90,7 +90,9 @@ module Hbc
       install_artifacts
       enable_accessibility_access
 
-      ::Utils::Analytics.report_event("cask_install", @cask.token)
+      unless @cask.tap&.private?
+        ::Utils::Analytics.report_event("cask_install", @cask.token)
+      end
 
       puts summary
     end
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -90,7 +90,7 @@ module Hbc
       install_artifacts
       enable_accessibility_access
 
-      Utils::Analytics.report_event("cask_install", @cask.token)
+      ::Utils::Analytics.report_event("cask_install", @cask.token)
 
       puts summary
     end
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -22,18 +22,22 @@ module Homebrew
     HOMEBREW_REPOSITORY.cd do
       analytics_message_displayed =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsmessage").chuzzle
+      cask_analytics_message_displayed =
+        Utils.popen_read("git", "config", "--local", "--get", "homebrew.cask.analyticsmessage").chuzzle
       analytics_disabled =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsdisabled").chuzzle
       if analytics_message_displayed != "true" &&
-         analytics_disabled != "true" &&
-         !ENV["HOMEBREW_NO_ANALYTICS"] &&
-         !ENV["HOMEBREW_NO_ANALYTICS_MESSAGE_OUTPUT"]
+        cask_analytics_message_displayed != "true" && 
+        analytics_disabled != "true" &&
+        !ENV["HOMEBREW_NO_ANALYTICS"] &&
+        !ENV["HOMEBREW_NO_ANALYTICS_MESSAGE_OUTPUT"]
+
         ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = "1"
         # Use the shell's audible bell.
         print "\a"
 
         # Use an extra newline and bold to avoid this being missed.
-        ohai "Homebrew has enabled anonymous aggregate user behaviour analytics."
+        ohai "Homebrew and Homebrew Cask have enabled anonymous aggregate user behaviour analytics."
         puts <<~EOS
           #{Tty.bold}Read the analytics documentation (and how to opt-out) here:
             #{Formatter.url("https://docs.brew.sh/Analytics")}#{Tty.reset}
@@ -43,6 +47,7 @@ module Homebrew
         # Consider the message possibly missed if not a TTY.
         if $stdout.tty?
           safe_system "git", "config", "--local", "--replace-all", "homebrew.analyticsmessage", "true"
+          safe_system "git", "config", "--local", "--replace-all", "homebrew.cask.analyticsmessage", "true"
         end
       end
 
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -22,12 +22,12 @@ module Homebrew
     HOMEBREW_REPOSITORY.cd do
       analytics_message_displayed =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsmessage").chuzzle
-      cask_analytics_message_displayed =
+      caskanalyticsmessage =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.cask.analyticsmessage").chuzzle
       analytics_disabled =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsdisabled").chuzzle
       if analytics_message_displayed != "true" &&
-        cask_analytics_message_displayed != "true" && 
+        caskanalyticsmessage != "true" && 
         analytics_disabled != "true" &&
         !ENV["HOMEBREW_NO_ANALYTICS"] &&
         !ENV["HOMEBREW_NO_ANALYTICS_MESSAGE_OUTPUT"]
@@ -37,7 +37,7 @@ module Homebrew
         print "\a"
 
         # Use an extra newline and bold to avoid this being missed.
-        ohai "Homebrew and Homebrew Cask have enabled anonymous aggregate user behaviour analytics."
+        ohai "Homebrew (and Cask) have enabled anonymous aggregate user behaviour analytics."
         puts <<~EOS
           #{Tty.bold}Read the analytics documentation (and how to opt-out) here:
             #{Formatter.url("https://docs.brew.sh/Analytics")}#{Tty.reset}
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -22,12 +22,12 @@ module Homebrew
     HOMEBREW_REPOSITORY.cd do
       analytics_message_displayed =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsmessage").chuzzle
-      caskanalyticsmessage =
+      cask_analytics_message_displayed =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.caskanalyticsmessage").chuzzle
       analytics_disabled =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsdisabled").chuzzle
       if analytics_message_displayed != "true" &&
-         caskanalyticsmessage != "true" &&
+         cask_analytics_message_displayed != "true" &&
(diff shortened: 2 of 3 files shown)


MikeMcQuaid:
I can confirm this will work as expected but I'd suggest @Homebrew/cask give final sign-off. I'd also suggest adding a build error report, too, so casks that are broken can be identified.

brianmorton:
Thanks for the feedback @MikeMcQuaid , the HBC team's suggestions would be helpful.

First, let me say I am super inexperienced with this repo.

I looked at adding stuff for borked cask installs, however it seems because of the modular nature of most of the checks (shasum, downloads) that it would also be triggered by commands other than just failed installs, conflating numbers. The logic checks to see if it is borked is in the modules, so I can't see an elegant way to get around that. If there is a way to do this (or if just including bad numbers is acceptable) I would welcome the help.

I also looked at trying to include whether or not something was installed as a dependency (perhaps as `cask_install_dependency`?) but without explicit feedback on wanting this I was reluctant to include it form the get-go.

vitorgalvao:
> I also looked at trying to include whether or not something was installed as a dependency

I don’t see that as necessary. If a cask has low download numbers the dependency installs will also be low; if a cask has high download numbers, the dependency needs to remain to support it. Dependency statistics will not be meaningful towards the goal of deciding what should be prioritised.

brianmorton:
Any ideas around reporting for broken casks? Or shall we just skip, at least for now?

MikeMcQuaid:
I think this is sufficient for now (or at least useful enough to be mergeable as-is).

MikeMcQuaid:
@Homebrew/cask any objections to the code as-is being merged?

MikeMcQuaid:
> For good measure, a one-time user-facing message would have been nice (for perfect transparency, taking into account the public sensitivity for the topic and the popular misconception of anonymous analytics == tracking my every move).

Yes, agreed with this. I think it'd be good to put something that's displayed (once) on the first `brew cask` run  after this PR has been run. That's what we did with our other analytics. Then again, we've told people for formulae so it may not be necessary.

brianmorton:
If you point me to an example of how it was done in the past I would be happy to update the PR.

scpeters:
@brianmorton I think https://github.com/Homebrew/brew/pull/143 contains some relevant code for printing a message about analytics during `brew update`

MikeMcQuaid:
Yeh, I'd suggest something like `homebrew.cask.analyticsmessage` with the similar `git config` approach.

brianmorton:
I believe I have added the message.

Not the most graceful integration, but I think it should work.

As always, feedback is welcome/encouraged.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -22,16 +22,23 @@ def update_report
...
+        !ENV["HOMEBREW_NO_ANALYTICS"] &&
+        !ENV["HOMEBREW_NO_ANALYTICS_MESSAGE_OUTPUT"]
+
         ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = "1"
         # Use the shell's audible bell.
         print "\a"
 
         # Use an extra newline and bold to avoid this being missed.
-        ohai "Homebrew has enabled anonymous aggregate user behaviour analytics."
+        ohai "Homebrew and Homebrew Cask have enabled anonymous aggregate user behaviour analytics."
```
MikeMcQuaid: `Homebrew (and Cask)` would be nicer; we need something < 81 characters.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -22,16 +22,23 @@ def update_report
     HOMEBREW_REPOSITORY.cd do
       analytics_message_displayed = \
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsmessage").chuzzle
+      cask_analytics_message_displayed = \
+        Utils.popen_read("git", "config", "--local", "--get", "homebrew.cask.analyticsmessage").chuzzle
```
MikeMcQuaid: How about just `caskanalyticsmessage`

brianmorton: Sure, I was just prepending the other variable with cask to keep some consistency.

brianmorton:
I have updated per PR per the requested changes.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -22,16 +22,23 @@ def update_report
     HOMEBREW_REPOSITORY.cd do
       analytics_message_displayed = \
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsmessage").chuzzle
+      caskanalyticsmessage = \
+        Utils.popen_read("git", "config", "--local", "--get", "homebrew.cask.analyticsmessage").chuzzle
```
MikeMcQuaid: I'd suggest `homebrew.caskanalyticsmessage` here

Library/Homebrew/cmd/update-report.rb
```diff
@@ -41,6 +48,7 @@ def update_report
         # Consider the message possibly missed if not a TTY.
         if $stdout.tty?
           safe_system "git", "config", "--local", "--replace-all", "homebrew.analyticsmessage", "true"
+          safe_system "git", "config", "--local", "--replace-all", "homebrew.cask.analyticsmessage", "true"
```
MikeMcQuaid: I'd suggest `homebrew.caskanalyticsmessage` here

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -89,6 +89,8 @@ def install
       install_artifacts
       enable_accessibility_access
 
+      Utils::Analytics.report_event("cask_install", @cask.token)
```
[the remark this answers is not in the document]
claui: For a quick test case that reproduces the error reliably, paste this into your terminal:

```
(
  set -e
  cd "$(mktemp -d)"
  git clone -b hbc_analytics --depth 1 https://github.com/brianmorton/brew
  cd brew
  mkdir -p Library/Taps/homebrew/homebrew-{core,cask}
  cat > empty.rb <<< \
    'cask("empty") { version :latest; sha256 :no_check; url "file:///dev/null"; container type: :naked }'
  unset HOMEBREW_NO_ANALYTICS
  export HOMEBREW_NO_AUTO_UPDATE=1
  pwd
  bin/brew cask reinstall empty.rb
)
```

This results in:

> Error: uninitialized constant Hbc::Utils::Analytics

Library/Homebrew/cmd/update-report.rb
brianmorton: Earlier ( https://github.com/Homebrew/brew/pull/4620#issuecomment-411760286 ) you suggested the current implementation. Could you confirm that you would now like it changed?

Library/Homebrew/cmd/update-report.rb
brianmorton: Earlier ( https://github.com/Homebrew/brew/pull/4620#issuecomment-411760286 ) you suggested the current implementation. Could you confirm that you would now like it changed?

Library/Homebrew/cmd/update-report.rb
MikeMcQuaid: Yes, please, sorry. I wasn't familiar with what the existing keys were at the time.

reitermarkus:
This needs a rebase.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -22,18 +22,22 @@ def update_report
     HOMEBREW_REPOSITORY.cd do
       analytics_message_displayed =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsmessage").chuzzle
+      caskanalyticsmessage =
+        Utils.popen_read("git", "config", "--local", "--get", "homebrew.caskanalyticsmessage").chuzzle
       analytics_disabled =
         Utils.popen_read("git", "config", "--local", "--get", "homebrew.analyticsdisabled").chuzzle
       if analytics_message_displayed != "true" &&
+         caskanalyticsmessage != "true" &&
```
reitermarkus: Name this `cask_analytics_message_displayed`.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -22,18 +22,22 @@ def update_report
...
          !ENV["HOMEBREW_NO_ANALYTICS"] &&
          !ENV["HOMEBREW_NO_ANALYTICS_MESSAGE_OUTPUT"]
+
         ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = "1"
         # Use the shell's audible bell.
         print "\a"
 
         # Use an extra newline and bold to avoid this being missed.
-        ohai "Homebrew has enabled anonymous aggregate user behaviour analytics."
+        ohai "Homebrew (and Cask) have enabled anonymous aggregate user behaviour analytics."
```
reitermarkus: Should this stay “has” since “and Cask” is in parentheses?

reitermarkus: This feels weird anyways since everything is in the Homebrew org. Maybe “Homebrew has enabled anonymous aggregate user behaviour analytics for formulae and casks.”?

reitermarkus:
This also needs a matching PR here:

https://github.com/Homebrew/install/blob/bfe2f046fb5dfe943fb936f6de8eaac7de5c483a/install#L387

MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @brianmorton!

brianmorton:
Now, for the whole point of this PR - how do I get to see these analytics?

MikeMcQuaid:
Only Homebrew maintainers have access to them. You would need to build something like https://github.com/homebrew/homebrew-formula-analytics and https://github.com/homebrew/formulae.brew.sh to upload the cask analytics publicly somewhere.

Library/Homebrew/cmd/update-report.rb
reitermarkus: That's exactly why `ohai` shouldn't truncate text.

MikeMcQuaid: @reitermarkus That's unrelated to `ohai`; even a `puts` that's wrapped by the terminal at >80 characters is uglier than one that's been done by humans (e.g. is it wrapped in the middle of a word or not),

brianmorton:
Can you all not just add a row or two to the table at https://formulae.brew.sh/analytics/ ?

MikeMcQuaid:
> Can you all not just

https://cantyoujust.no 😉 

What I've mentioned above are the requirements. It's not as simple as just adding a line because then it would need to be downloaded manually, entered manually and updated manually.

brianmorton:
Forgive me, but since you seem to guard the analytics and workflow more closely than I would have thought, the process becomes somewhat opaque.

To me it would have seemed there was an automated process for providing the existing analytics. In that line of reasoning, since the cask analytics are feeding in to the same system, it would seem to trivial to simply output a table row.

Based on your linking of the `homebrew-formula-analytics` repo, I am now guessing that somebody manually runs a CLI command or series of commands and manually uploads a file to that repo/site?

Also, respectfully, I think it would be difficult for me to update [brew-formula-analytics.rb](https://github.com/Homebrew/homebrew-formula-analytics/blob/master/cmd/brew-formula-analytics.rb) since you have indicated the analytics are restricted.

To that end, I am happy to do what I can to move this along so there are some type of publicly accessible cask analytics.

Please remember that this is all tangential to *my* original purpose of having cask analytics available to guide automation efforts. I am happy to continue submitting PRs and Getting Things Done as long as there is light at the end of this tunnel.

MikeMcQuaid:
> Based on your linking of the homebrew-formula-analytics repo, I am now guessing that somebody manually runs a CLI command or series of commands and manually uploads a file to that repo/site?

It's manually run for maintainers to query the values. More recently a job has been added to homebrew-core (https://github.com/Homebrew/homebrew-core/blob/master/.travis.yml) which commits the values to formulae.brew.sh.



> Also, respectfully, I think it would be difficult for me to update brew-formula-analytics.rb since you have indicated the analytics are restricted.

This is tricky, I agree. If the @Homebrew/cask maintainers are happy with you being temporarily granted access to the Homebrew Cask analytics to build the necessary functionality I would be OK with that too. I would say that given the naming `homebrew-formula-analytics` and `formulae.brew.sh` may not entirely lend themselves to just having Casks bolted on but I defer that to whoever does the work.

vitorgalvao:
> If the @Homebrew/cask maintainers are happy with you being temporarily granted access to the Homebrew Cask analytics to build the necessary functionality I would be OK with that too.

I am.

claui:
I’d be fine with that, too.

MikeMcQuaid:
Cool, thanks folks. @brianmorton can you send me your Google account email address?

brianmorton:
Sent.

PR #4622, opened 2018-08-06
formula_installer: avoid cyclic dependency on Linux

iMichka:

-----

Fixes:
==> Installing patchelf dependency: patchelf
==> Installing dependencies for patchelf: patchelf
==> Installing patchelf dependency: patchelf
==> Installing dependencies for patchelf: patchelf
==> Installing patchelf dependency: patchelf
==> Installing dependencies for patchelf: patchelf
...

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -502,7 +502,7 @@ class FormulaInstaller
       end
     end
 
-    if pour_bottle
+    if pour_bottle && !Keg.relocation_formulae.include?(formula.name)
       bottle_deps = Keg.relocation_formulae
                        .map { |formula| Dependency.new(formula) }
                        .reject do |dep|


iMichka:
CC @maxim-belkin @sjackman

maxim-belkin:
Just to clarify:

1. There is nothing specific to Linux in this change.
2. This is just a last resort check to prevent cyclic dependencies.

It would be nice to see a message when a cyclic dependency is detected/prevented. 
It would also be nice to check that the dependency is actually installed :)

Alternative way to silently remove cyclic dependencies would be to add
```
# Remove cyclic dependencies
deps.reject! { |dep| dep.to_formula.full_name == formula.full_name }
```
before the `deps.each` loop.

sjackman:
Note that this issue arrises if `patchelf` were to have a bottle for Linux. It doesn't currently in Homebrew/core, but does in Linuxbrew/core. I'd suggest this fix:

```diff
diff --git a/Library/Homebrew/formula_installer.rb b/Library/Homebrew/formula_installer.rb
index 7b4531e12..94a1cce4f 100644
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -499,6 +499,7 @@ class FormulaInstaller
     if pour_bottle
       bottle_deps = Keg.relocation_formulae.map { |formula| Dependency.new(formula) }
                        .reject do |dep|
+        next true if dep.name == formula.name
         inherited_options[dep.name] |= inherited_options_for(dep)
         dep.satisfied? inherited_options[dep.name]
       end
```

@iMichka If you agree, would you update this PR?

iMichka:
Yes, my first attempt was to safeguard this only for the "patchelf" formula, because I think it will only be needed for the case.

Something like this (untested) ?:
```
if formula.full_name == dep.to_formula.full_name
   next if formula.full_name == "patchelf"
   raise/odie + error message
```

reitermarkus:
Take a look at

https://github.com/Homebrew/brew/blob/6a1fa87191bfef31ff1b2d47d3ebf281398a210f/Library/Homebrew/cask/lib/hbc/cask_dependencies.rb

It uses `TSort` to resolve the dependencies, which raises an exception if there is a cyclic dependency.

MikeMcQuaid:
I think something like `odie "Recursive dependency on #{formula.name}!" if dep.name == formula.name` would be sufficient.

sjackman:
This special case occurs only when the requested formula to be installed (`formula.name`) also occurs in `Keg.relocation_formulae`. I suggest either the patch I have above (https://github.com/Homebrew/brew/pull/4622#issuecomment-411249434), or alternatively…

```diff
diff --git a/Library/Homebrew/formula_installer.rb b/Library/Homebrew/formula_installer.rb
index d9565ddd7..23380d043 100644
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -503,7 +503,7 @@ class FormulaInstaller
     end
 
     if pour_bottle
-      bottle_deps = Keg.relocation_formulae
+      bottle_deps = (Keg.relocation_formulae - [formula.name])
                        .map { |formula| Dependency.new(formula) }
                        .reject do |dep|
         inherited_options[dep.name] |= inherited_options_for(dep)
```

In my opinion, the patch should not specifically mention `patchelf`, as the value of `Keg.relocation_formulae` differs on Linux and macOS, and may change.

@MikeMcQuaid Do you have a preference between these two patches? As above or https://github.com/Homebrew/brew/pull/4622#issuecomment-411249434

MikeMcQuaid:
I'd suggest something that avoids that logic entirely if `Keg.relocation_formulae.include?(formula.name)`

sjackman:
```diff
diff --git a/Library/Homebrew/formula_installer.rb b/Library/Homebrew/formula_installer.rb
index d9565ddd7..66a44a9cd 100644
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -502,7 +502,7 @@ class FormulaInstaller
       end
     end
 
-    if pour_bottle
+    if pour_bottle && !Keg.relocation_formulae.include?(formula.name)
       bottle_deps = Keg.relocation_formulae
                        .map { |formula| Dependency.new(formula) }
                        .reject do |dep|
```

@iMichka Could you please update this PR?

MikeMcQuaid:
@sjackman Looks good. You can edit this PR yourself though, too 😎

iMichka:
Good to go for me.

maxim-belkin:
I have a feeling that this is not the right fix (it probably works but in a weird way).
The whole infinite loop is caused by the incorrect value of `deps` variable in the `install_dependencies` function/method:

https://github.com/Homebrew/brew/blob/bb49b50f511aef7709aad5f7d0ff8b13837bb043/Library/Homebrew/formula_installer.rb#L547-L557

If we look at line 413,

https://github.com/Homebrew/brew/blob/bb49b50f511aef7709aad5f7d0ff8b13837bb043/Library/Homebrew/formula_installer.rb#L412-L415

we'll see that `deps` is computed by the `compute_dependencies` method:

https://github.com/Homebrew/brew/blob/bb49b50f511aef7709aad5f7d0ff8b13837bb043/Library/Homebrew/formula_installer.rb#L392-L398

Assuming that `formula.deps` is empty/nil for `patchelf`, it is more likely the fault of non-empty `req_deps` variable and, therefore, of the `expand_requirements` method than of `expand_dependencies`. I think there is a mistake in the `expand_requirements` method that we're masking by "fixing" the `expand_dependencies` method.

sjackman:
@maxim-belkin I believe the problem is this line:
https://github.com/Homebrew/brew/blob/7ed3be53e007f608ffc38336ba92973dfe9b13f8/Library/Homebrew/formula_installer.rb#L512

On Linux `bottle_deps` is `["patchelf"]` which causes `expanded_deps` to contain `patchelf`, but the formula that we're installing is `patchelf`, which causes the infinite loop.

maxim-belkin:
You're right. `req_deps` is empty when we install `patchelf`, but`patchelf` is still added _via_ `bottle_deps` because of the `Keg.relocation_formulae`: it is hard-coded as`["patchelf"]` on Linux and an empty list for Mac. So, proposed changes look good then. ☮️

MikeMcQuaid:
Thanks all!

Issue #4624, opened 2018-08-07
Cache migration to `--` format is broken.


chdiza:
- [x] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew update` and can still reproduce the problem?
- [x] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [x] ran `brew config` and `brew doctor` and included their output with your issue?

The migration of tarballs in the cache to the double-dash form is broken.  Instead of migrating a previously-existing tarball, a brand-new one is created and the old one is still there.

For instance, the tarball `markdown-1.0.1.zip` is already in my cache because `markdown` is already installed.  Then I uninstall `markdown`, and then do `brew install markdown`.  Instead of the old tarball being renamed as `markdown--1.0.1.zip`, I now have two markdown tarballs: one called `markdown-1.0.1.zip` and the other called `markdown--1.0.1.zip`.

(I confess I can't see what purpose this double-dash stuff is serving anyway.)

```
$ brew config
HOMEBREW_VERSION: 1.7.1-154-g9f370a3
ORIGIN: https://github.com/Homebrew/brew
HEAD: 9f370a35767f9af94ab2c519770754421a82bcd6
Last commit: 7 hours ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 919c57196e6964ad8370e2d84c96db644c4b9512
Core tap last commit: 5 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/hb
HOMEBREW_CACHE: /stuff/brache
HOMEBREW_TEMP: /Users/chdiza/.tmp/tmpdir
HOMEBREW_BUILD_FROM_SOURCE: 1
HOMEBREW_DEV_CMD_RUN: 1
HOMEBREW_LOGS: /Users/chdiza/Library/Homebrew
HOMEBREW_NO_ANALYTICS: 1
[...]

PR #4632, opened 2018-08-08
Output message when migrating cache entries.

reitermarkus:

-----

Output message when migrating cache entries, since this may take longer than a usual `brew update`.

Closes https://github.com/Homebrew/brew/issues/4624.


DomT4:
I've started getting that `==> Migrating cache entries...` notification every time I run `brew update`, even when the cache is completely empty. I assume that isn't expected?

DomT4:
Ah, thanks. Missed the initial `return` with the diff here when I glanced pre-comment.

PR #4633, opened 2018-08-08
Refactor `Cleanup`.

reitermarkus:

-----

Start refactoring `Cleanup`.

### DIFF
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -1,6 +1,71 @@
 require "utils/bottles"
 require "formula"
 
+module CleanupRefinement
+  refine Pathname do
+    def incomplete?
+      extname.end_with?(".incomplete")
+    end
+
+    def nested_cache?
+      directory? && ["glide_home", "java_cache", "npm_cache"].include?(basename.to_s)
+    end
+
+    def prune?(days)
+      return false unless days
+      return true if days.zero?
+
+      # TODO: Replace with ActiveSupport's `.days.ago`.
+      mtime < ((@time ||= Time.now) - days * 60 * 60 * 24)
+    end
+
+    def stale?(scrub = false)
+      return false unless file?
+
+      stale_formula?(scrub)
+    end
+
+    private
+
+    def stale_formula?(scrub)
+      return false unless HOMEBREW_CELLAR.directory?
+
+      version = if to_s.match?(Pathname::BOTTLE_EXTNAME_RX)
+        begin
+          Utils::Bottles.resolve_version(self)
+        rescue
+          self.version
+        end
+      else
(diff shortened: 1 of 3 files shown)


Library/Homebrew/cleanup.rb
```diff
@@ -1,6 +1,71 @@
 require "utils/bottles"
 require "formula"
 
+module CleanupRefinement
```
MikeMcQuaid: Is this replacing monkeypatching or just makes the code below a bit nicer?

Library/Homebrew/cleanup.rb
```diff
@@ -46,22 +111,17 @@ def cleanup_keg(keg)
     def cleanup_logs
       return unless HOMEBREW_LOGS.directory?
       HOMEBREW_LOGS.subdirs.each do |dir|
-        cleanup_path(dir) { dir.rmtree } if prune?(dir, days_default: 14)
+        cleanup_path(dir) { dir.rmtree } if dir.prune?(ARGV.value("prune")&.to_i || 14)
```
MikeMcQuaid: I think the `14` should become a constant as it's no longer documented by the `days_default` now

Library/Homebrew/cleanup.rb
```diff
@@ -1,6 +1,71 @@
...
+
+    def nested_cache?
+      directory? && ["glide_home", "java_cache", "npm_cache"].include?(basename.to_s)
+    end
+
+    def prune?(days)
+      return false unless days
+      return true if days.zero?
+
+      # TODO: Replace with ActiveSupport's `.days.ago`.
```
MikeMcQuaid: Do you want this to block the PR?

MikeMcQuaid:
@reitermarkus Nice work here. Just to elaborate a bit on some of my intentions here so you might be able to do some stuff to help: I was hoping to make `brew upgrade --cleanup` (and `HOMEBREW_UPGRADE_CLEANUP`) cleanup more stuff related to a formula e.g. the `cache` and perhaps prune old symlinks (`brew prune`) too. On top of this I was hoping eventually to get to the point where `brew linkage` data (now that it's cached) could be used to run `brew cleanup` automatically on every upgrade/install (guarded by `HOMEBREW_UPGRADE_CLEANUP` to begin with) if a keg isn't being linked to by any other active keg.

Library/Homebrew/cleanup.rb
reitermarkus: Just to make the actual cleanup logic easier to read.

Library/Homebrew/cleanup.rb
reitermarkus: No, this is just so we don't forget to use it when we have it.

PR #4642, opened 2018-08-08
Mac.sdk_path_if_needed tests: stub SDK check

mistydemeo:

-----

@reitermarkus noticed that the "calls sdk_path on CLT-only systems with a CLT SDK if the system does not provide headers" test was failing on a system which was missing the CLT; I believe this is because I didn't stub `OS::Mac::CLT.provides_sdk?`. This PR adds that stub to all appropriate tests, including the one that was broken.


reitermarkus:
Too bad I can't reproduce it now that I have the CLT installed.

Just one nit, change the `allow`s to use `.and_return` instead of using the block form.

mistydemeo:
Updated to rewrite all of the `sdk_path_if_needed` tests to use `and_return`.

PR #4646, opened 2018-08-09
Allow Homebrew/brew(.git)? to ignore the case.

envygeeks:
When people opt to install in their home directory this can be all lowercase (as Github doesn't enforce case on their URL's for anything but file names), and this will popup as a warning, even though the repository is perfectly valid.

-----


MikeMcQuaid:
Thanks for the PR but I'd rather we are still notified on this in `brew doctor`.

Issue #4650, opened 2018-08-09
Latest `cleanup` is broken


chdiza:
- [x] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew update` and can still reproduce the problem?
- [x] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [x] ran `brew config` and `brew doctor` and included their output with your issue?

Sometime within the last 24 hours, `cleanup -n` is now (a) listing stuff twice, and (b) listing stuff that (i) it didn't use to list and (ii) should not be cleaned up.

For instance, in the output below you can see the duplication.  Also,  every single thing listed there is both installed and current.  (Some of it is from my own custom tap but e.g., `jpeg` is not.

```
$ brew cleanup -n
Would remove: /stuff/brache/browser--7 (419B)
Would remove: /stuff/brache/jpeg--9c.high_sierra.bottle.tar.gz (297.6KB)
Would remove: /stuff/brache/jpeg--9c.tar.gz (1004.1KB)
Would remove: /stuff/brache/jq--1.5.d2.tar.xz (3.6MB)
Would remove: /stuff/brache/lynx--2.8.9rel1.tar.xz (3.5MB)
Would remove: /stuff/brache/screenresolution--1.6i.txz (34KB)
Would remove: /stuff/brache/browser--7 (419B)
Would remove: /stuff/brache/jpeg--9c.high_sierra.bottle.tar.gz (297.6KB)
Would remove: /stuff/brache/jpeg--9c.tar.gz (1004.1KB)
Would remove: /stuff/brache/jq--1.5.d2.tar.xz (3.6MB)
Would remove: /stuff/brache/lynx--2.8.9rel1.tar.xz (3.5MB)
Would remove: /stuff/brache/screenresolution--1.6i.txz (34KB)
```

None of that was listed yesterday.

```
$ brew config
HOMEBREW_VERSION: 1.7.1-204-g8e703e3
[...]

PR #4651, opened 2018-08-09
Fix `brew cleanup`.

reitermarkus:

-----

This fixes the regex to properly parse the version and to also match resources and patches. 

Fixes https://github.com/Homebrew/brew/issues/4650.

### DIFF
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -153,8 +153,8 @@ module Homebrew
         end
         cleanup_cache
         cleanup_logs
-        return if dry_run?
         cleanup_lockfiles
+        return if dry_run?
         rm_ds_store
       else
         args.each do |arg|
@@ -254,7 +254,12 @@ module Homebrew
       lockfiles.each do |file|
         next unless file.readable?
         next unless file.open(File::RDWR).flock(File::LOCK_EX | File::LOCK_NB)
-        cleanup_path(file) { file.unlink }
+
+        begin
+          cleanup_path(file) { file.unlink }
+        ensure
+          file.open(File::RDWR).flock(File::LOCK_UN) if file.exist?
+        end
       end
     end
 
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -148,12 +148,12 @@ module Homebrew
 
     def clean!
       if args.empty?
-        Formula.installed.each do |formula|
+        cleanup_lockfiles
+        Formula.installed.sort_by(&:name).each do |formula|
           cleanup_formula(formula)
         end
         cleanup_cache
         cleanup_logs
-        cleanup_lockfiles
         return if dry_run?
         rm_ds_store
       else
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -66,13 +66,14 @@ module CleanupRefinement
         end
       end
 
+      version ||= basename.to_s[/\A.*(?:\-\-.*?)*\-\-(.*?)#{Regexp.escape(extname)}\Z/, 1]
       version ||= basename.to_s[/\A.*\-\-?(.*?)#{Regexp.escape(extname)}\Z/, 1]
 
       return false unless version
 
       version = Version.new(version)
 
-      return false unless formula_name = basename.to_s[/\A(.*?)(\-\-.*)*\-\-?(?:#{Regexp.escape(version)})/, 1]
+      return false unless formula_name = basename.to_s[/\A(.*?)(?:\-\-.*?)*\-\-?(?:#{Regexp.escape(version)})/, 1]
 
       formula = begin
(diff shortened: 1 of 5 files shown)


chdiza:
I just tried this, and it doesn't fix all of the problem, and also introduces a new one.

(1) The list of things output by `cleanup -n` is still duplicated.

(2) It now lists resources for formulas that I have installed and are not outdated.  For example, it wants to delete the patch associated with `libtiff` and also "ghostscript fonts".

chdiza:
OK, that fixed what was mentioned in the last comment, but now `cleanup` wants to delete a non-outdated source tarball that it hadn't previously complained about.

Prior to the most recent commit in this PR, `cleanup` erroneously wanted to delete `foo--resource1--3.2.tar.bz2`, `foo--resource2--1.0.tar.bz2`, and so forth.  Now this is fixed.  `cleanup` no longer wants to delete those.  But instead, it now wants to delete the main `foo` source tarball.  If it matters, the tarball name is of the form `foo--2.9.1-1.tar.bz2`.

reitermarkus:
Which formula has the `foo--2.9.1-1.tar.bz2` format?

chdiza:
> Which formula has the foo--2.9.1-1.tar.bz2 format?

One in my custom tap.

reitermarkus:
I can't really debug this without an actual formula to test this with.

reitermarkus:
@chdiza, can you try again?

chdiza:
I'll check it it an hour or so when I'm at the relevant machine.

Might be worth noting that the formula `foo` has an explicit `version` declared.   Also that this declared version ("2.2.9-1") is different from the version that's embedded in the name of the tarball as it appears in the `url` argument (`foo--2.2.9.tar.bz2`).

chdiza:
> Might be worth noting that the formula foo has an explicit version declared. Also that this declared version ("2.2.9-1") is different from the version that's embedded in the name of the tarball as it appears in the url argument (foo--2.2.9.tar.bz2).

This is exactly what was going on.  I made up a different silly formula that has this property, and installed it.  `cleanup` then tried to delete the main tarball but correctly left the `resource` alone.  The formula is pasted below.


```ruby
class Foo < Formula
  desc "Fzf"
  homepage "https://github.com/junegunn/fzf"
  url "https://github.com/junegunn/fzf-bin/releases/download/0.17.4/fzf-0.17.4-darwin_386.tgz"
  version "2.2.9-1"
  sha256 "e3147b1e471c5d4fcab774cb79d880f4dc6fb88ec7217cab6c397f85af10f541"

  # The real package requires `go` and other crap to build.

  resource "sourceball" do
    url "https://github.com/junegunn/fzf/archive/0.17.4.tar.gz"
    sha256 "a4b009638266b116f422d159cd1e09df64112e6ae3490964db2cd46636981ff0"
  end

  def install
    resource('sourceball').stage do
      man1.install "man/man1/fzf.1"
    end

    bin.install "fzf"
  end
end
```

chdiza:
> @chdiza, can you try again?

With the latest change, the problem has disappeared.

However, there is still one thing that I think should be fixed.  `cleanup` now is deleting lock files, which I guess is fine (but what happens if the user runs `cleanup` while a build is underway?).  The problem is that the output of `cleanup` has the lockfiles and the tarballs/old kegs all mixed together in no real order.  E.g., 

```
Would remove: /stuff/brache/browser--7 (419B)
Would remove: /usr/local/var/homebrew/locks/foo.formula.lock (0B)
Would remove: /stuff/brache/jpeg--9c.high_sierra.bottle.tar.gz (297.6KB)
```

This is just a snippet of longer output.  What *should* happen is that `cleanup` displays all the lockfiles first (or last).   Otherwise, it is VERY easy to visually overlook a tarball or old keg in the list, because they are obscured by the noise of the lockfiles.  This makes it difficult for the user to see which tarballs and old kegs are going to be deleted when she runs `cleanup -n`.

reitermarkus:
> `cleanup` now is deleting lock files

It did that before too, but it wasn't shown in the output. `cleanup` will try to lock the lockfile, and only then delete it.

chdiza:
> They are ordered alphabetically by the formula name: browser, foo, jpeg

That's a bad idea, for the reason described above.

chdiza:
The output should look like this:
```
...browser.formula.lock
...foo.formula.lock
...jpeg.formula.lock
...browser_tarball
...foo_tarball
...jpeg_old-keg
...jpeg_tarball
```
or else show all the lockfiles at the end.

Otherwise (especially if the list is long) it's hard to see which tarballs or old kegs are going to be deleted, because they get lost in the lockfile noise.

reitermarkus:
@chdiza, I moved the lockfile cleanup to the top now.

chdiza:
> I moved the lockfile cleanup to the top now.

OK, seems to be working well now, thanks.  I'll keep an eye on things in the coming weeks to see if there are any other oddball cases.

Issue #4401, opened 2018-07-01
Tracking Issue: Merging download strategies.


reitermarkus:
Checklist:

- [x] Separate staging from downloads. (https://github.com/Homebrew/brew/pull/4402)
- [x] Refactor cask staging strategies. (https://github.com/Homebrew/brew/pull/4474, https://github.com/Homebrew/brew/pull/4479)
- [x] Merge staging strategies (https://github.com/Homebrew/brew/pull/4541).
- [x] Refactor download strategies. (https://github.com/Homebrew/brew/pull/4592, https://github.com/Homebrew/brew/pull/4593, https://github.com/Homebrew/brew/pull/4595, https://github.com/Homebrew/brew/pull/4596, https://github.com/Homebrew/brew/pull/4662)
- [x] Merge download strategies. (https://github.com/Homebrew/brew/pull/4607, https://github.com/Homebrew/brew/pull/4608, https://github.com/Homebrew/brew/pull/4616)

MikeMcQuaid:
I think this is now all done? Congrats @reitermarkus 👏

PR #4656, opened 2018-08-10
Enable quarantining of Homebrew-Cask's downloads

amyspark:

-----

This pull requests adds an additional step to the `Hbc::Download` class that quarantines all downloaded files. 

The original proposal by @reitermarkus tinkered with Gatekeeper's download database (`~/Library/Preferences/com.apple.LaunchServices.QuarantineEventsV2`, see #3526), which I found *really* prone to breaking the main Gatekeeper policy database (i.e. `/private/var/db/SystemPolicy`).

[Inspired in Transmission's source code](https://github.com/transmission/transmission/blob/c41b957b7c9623f5360807a34bae1120427c369f/macosx/Torrent.m#L1842-L1863), I implemented download quarantining through the use of a Swift script. In this way, we can add Gatekeeper support while keeping the code completely visible to the user.

As regards the metadata, I added a modicum of fields:

- Agent: `Homebrew-Cask`
- Download type: `WEB_DOWNLOAD` (see a list of the possible types [here](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.13.sdk/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Headers/LSQuarantine.h))
- Download URL: this is obtained from the Cask `url` field
- Originl URL: the Cask's homepage

This also includes:
- tests for `install`, `fetch`, and `audit`, across multiple types of containers
- the ability to disable quarantining either when Swift is not found, or when using the special flag `--no-quarantine`

If merged, this pull request closes homebrew/homebrew-cask#22388.

Please review, and thanks for your comments!

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
+++ b/Library/Homebrew/cask/lib/hbc/artifact/moved.rb
@@ -65,11 +65,8 @@ module Hbc
         ohai "Backing #{self.class.english_name} '#{target.basename}' up to '#{source}'."
         source.dirname.mkpath
 
-        if target.parent.writable?
-          FileUtils.cp_r(target, source)
-        else
-          command.run!("/bin/cp", args: ["-r", target, source], sudo: true)
-        end
+        # We need to preserve extended attributes between copies.
+        command.run!("/bin/cp", args: ["-pR", target, source], sudo: !target.parent.writable?)
 
         delete(target, force: force, command: command, **options)
       end
--- a/Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
@@ -7,11 +7,12 @@ module Hbc
       include Options
       include Homebrew::Search
 
-      option "--[no-]binaries", :binaries,      true
-      option "--debug",         :debug,         false
-      option "--verbose",       :verbose,       false
-      option "--outdated",      :outdated_only, false
-      option "--require-sha",   :require_sha,   false
+      option "--[no-]binaries",   :binaries,      true
+      option "--debug",           :debug,         false
+      option "--verbose",         :verbose,       false
+      option "--outdated",        :outdated_only, false
+      option "--require-sha",     :require_sha,   false
+      option "--[no-]quarantine", :quarantine,    true
 
       def self.command_name
         @command_name ||= name.sub(/^.*:/, "").gsub(/(.)([A-Z])/, '\1_\2').downcase
--- a/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
+++ b/Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
@@ -36,7 +36,11 @@ module Hbc
 
           old_cask = CaskLoader.load(old_cask.installed_caskfile)
 
-          old_cask_installer = Installer.new(old_cask, binaries: binaries?, verbose: verbose?, force: force?, upgrade: true)
+          old_cask_installer =
+            Installer.new(old_cask, binaries:  binaries?,
+                                    verbose:   verbose?,
+                                    force:     force?,
+                                    upgrade:   true)
 
           new_cask = CaskLoader.load(old_cask.to_s)
 
@@ -46,7 +50,8 @@ module Hbc
                                     force:          force?,
                                     skip_cask_deps: skip_cask_deps?,
                                     require_sha:    require_sha?,
-                                    upgrade: true)
+                                    upgrade:        true,
+                                    quarantine:     quarantine?)
 
           started_upgrade = false
           new_artifacts_installed = false
--- a/Library/Homebrew/cask/lib/hbc/exceptions.rb
+++ b/Library/Homebrew/cask/lib/hbc/exceptions.rb
@@ -32,13 +32,13 @@ module Hbc
 
   class CaskUnavailableError < AbstractCaskErrorWithToken
     def to_s
-      "Cask '#{token}' is unavailable" << (reason.empty? ? "." : ": #{reason}")
+      "Cask '#{token}' is unavailable#{reason.empty? ? "." : ": #{reason}"}"
     end
   end
 
   class CaskUnreadableError < CaskUnavailableError
     def to_s
-      "Cask '#{token}' is unreadable" << (reason.empty? ? "." : ": #{reason}")
+      "Cask '#{token}' is unreadable#{reason.empty? ? "." : ": #{reason}"}"
     end
   end
 
@@ -73,7 +73,7 @@ module Hbc
 
   class CaskCyclicDependencyError < AbstractCaskErrorWithToken
     def to_s
-      "Cask '#{token}' includes cyclic dependencies on other Casks" << (reason.empty? ? "." : ": #{reason}")
+      "Cask '#{token}' includes cyclic dependencies on other Casks#{reason.empty? ? "." : ": #{reason}"}"
     end
   end
 
@@ -91,7 +91,7 @@ module Hbc
 
   class CaskInvalidError < AbstractCaskErrorWithToken
     def to_s
-      "Cask '#{token}' definition is invalid" << (reason.empty? ? "." : ": #{reason}")
+      "Cask '#{token}' definition is invalid#{reason.empty? ? "." : ": #{reason}"}"
     end
   end
 
@@ -147,6 +147,41 @@ module Hbc
         Cask '#{token}' does not have a sha256 checksum defined and was not installed.
         This means you have the #{Formatter.identifier("--require-sha")} option set, perhaps in your HOMEBREW_CASK_OPTS.
       EOS
+    end
+  end
+
+  class CaskQuarantineError < CaskError
+    attr_reader :path, :reason
+
+    def initialize(path, reason)
+      @path = path
+      @reason = reason
+    end
+
+    def to_s
+      s = "Failed to quarantine #{path}."
+
+      unless reason.empty?
+        s << " Here's the reason:\n"
+        s << Formatter.error(reason)
+        s << "\n" unless reason.end_with?("\n")
+      end
+
+      s
+    end
+  end
+
+  class CaskQuarantinePropagationError < CaskQuarantineError
+    def to_s
+      s = "Failed to quarantine one or more files within #{path}."
+
+      unless reason.empty?
+        s << " Here's the reason:\n"
+        s << Formatter.error(reason)
+        s << "\n" unless reason.end_with?("\n")
+      end
+
+      s
     end
   end
 end
--- a/Library/Homebrew/cask/lib/hbc/installer.rb
+++ b/Library/Homebrew/cask/lib/hbc/installer.rb
@@ -7,6 +7,7 @@ require "hbc/cask_dependencies"
 require "hbc/download"
 require "hbc/staged"
 require "hbc/verify"
+require "hbc/quarantine"
 
 require "cgi"
 
@@ -23,7 +24,10 @@ module Hbc
 
     PERSISTENT_METADATA_SUBDIRS = ["gpg"].freeze
 
-    def initialize(cask, command: SystemCommand, force: false, skip_cask_deps: false, binaries: true, verbose: false, require_sha: false, upgrade: false, installed_as_dependency: false)
+    def initialize(cask, command: SystemCommand, force: false,
+                   skip_cask_deps: false, binaries: true, verbose: false,
+                   require_sha: false, upgrade: false,
+                   installed_as_dependency: false, quarantine: true)
       @cask = cask
       @command = command
       @force = force
@@ -34,9 +38,12 @@ module Hbc
       @reinstall = false
       @upgrade = upgrade
       @installed_as_dependency = installed_as_dependency
+      @quarantine = quarantine
     end
 
-    attr_predicate :binaries?, :force?, :skip_cask_deps?, :require_sha?, :upgrade?, :verbose?, :installed_as_dependency?
+    attr_predicate :binaries?, :force?, :skip_cask_deps?, :require_sha?,
+                   :upgrade?, :verbose?, :installed_as_dependency?,
+                   :quarantine?
 
     def self.print_caveats(cask)
       odebug "Printing caveats"
@@ -86,6 +93,7 @@ module Hbc
       uninstall_existing_cask if @reinstall
 
       oh1 "Installing Cask #{Formatter.identifier(@cask)}"
+      opoo "macOS's Gatekeeper has been disabled for this Cask" unless quarantine?
       stage
       install_artifacts
       enable_accessibility_access
@@ -137,7 +145,7 @@ module Hbc
 
     def download
       odebug "Downloading"
-      @downloaded_path = Download.new(@cask, force: false).perform
+      @downloaded_path = Download.new(@cask, force: false, quarantine: quarantine?).perform
       odebug "Downloaded to -> #{@downloaded_path}"
       @downloaded_path
     end
@@ -176,6 +184,11 @@ module Hbc
       else
         primary_container.extract_nestedly(to: @cask.staged_path, basename: basename, verbose: verbose?)
       end
+
+      return unless quarantine?
+      return unless Quarantine.available?
+
+      Quarantine.propagate(from: @downloaded_path, to: @cask.staged_path, command: @command)
     end
 
     def install_artifacts
--- a/Library/Homebrew/cask/lib/hbc/quarantine.rb
+++ b/Library/Homebrew/cask/lib/hbc/quarantine.rb
@@ -1,0 +1,80 @@
+require "development_tools"
+module Hbc
+  module Quarantine
+    module_function
+
+    QUARANTINE_ATTRIBUTE = "com.apple.quarantine".freeze
+
+    QUARANTINE_SCRIPT = (HOMEBREW_LIBRARY_PATH/"cask/lib/hbc/utils/quarantine.swift").freeze
+
+    # @private
+    def swift
+      @swift ||= DevelopmentTools.locate("swift")
+    end
+
+    def available?
+      status = !swift.nil?
+      odebug "Quarantine is #{status ? "available" : "not available"}."
+      status
+    end
+
+    def detect(file)
+      return if file.nil?
+
+      odebug "Verifying Gatekeeper status of #{file}"
+
+      quarantine_status = !status(file).empty?
+
+      odebug "#{file} is #{quarantine_status ? "quarantined" : "not quarantined"}"
+
+      quarantine_status
+    end
+
+    def status(file, command: SystemCommand)
+      command.run("/usr/bin/xattr",
+                  args:        ["-p", QUARANTINE_ATTRIBUTE, file],
+                  print_stderr: false).stdout.rstrip
+    end
+
+    def cask(cask: nil, download_path: nil, command: SystemCommand)
+      return if cask.nil? || download_path.nil?
+
+      odebug "Quarantining #{download_path}"
+
+      quarantiner = command.run(swift,
+                                args: [
+                                  QUARANTINE_SCRIPT,
+                                  download_path,
+                                  cask.url.to_s,
+                                  cask.homepage.to_s,
+                                ])
+
+      return if quarantiner.success?
+
+      case quarantiner.exit_status
+      when 2
+        raise CaskQuarantineError.new(download_path, "Insufficient parameters")
+      else
+        raise CaskQuarantineError.new(download_path, quarantiner.stderr)
+      end
+    end
+
+    def propagate(from: nil, to: nil, command: SystemCommand)
+      return if from.nil? || to.nil?
+
+      raise CaskError, "#{from} was not quarantined properly." unless detect(from)
+
+      odebug "Propagating quarantine from #{from} to #{to}"
+
+      quarantine_status = status(from, command: command)
+
+      quarantiner = command.run("/usr/bin/xattr",
+                                args: ["-w", "-r", QUARANTINE_ATTRIBUTE, quarantine_status, to],
+                                print_stderr: false)
+
+      return if quarantiner.success?
+
+      raise CaskQuarantinePropagationError.new(to, quarantiner.stderr)
+    end
+  end
+end
--- a/Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
+++ b/Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
@@ -1,0 +1,42 @@
+#!/usr/bin/swift
+
+import Foundation
+
+struct swifterr: TextOutputStream {
+  public static var stream = swifterr()
+  mutating func write(_ string: String) { fputs(string, stderr) }
+}
+
+if (CommandLine.arguments.count < 4) {
+  exit(2)
+}
+
+let dataLocationUrl: NSURL = NSURL.init(fileURLWithPath: CommandLine.arguments[1])
+
+var errorBag: NSError?
+
+let quarantineProperties: [String: Any] = [
+  kLSQuarantineAgentNameKey as String: "Homebrew-Cask",
+  kLSQuarantineTypeKey as String: kLSQuarantineTypeWebDownload,
+  kLSQuarantineDataURLKey as String: CommandLine.arguments[2],
+  kLSQuarantineOriginURLKey as String: CommandLine.arguments[3]
+]
+
+if (dataLocationUrl.checkResourceIsReachableAndReturnError(&errorBag)) {
+  do {
+    try dataLocationUrl.setResourceValue(
+      quarantineProperties as NSDictionary,
+      forKey: URLResourceKey.quarantinePropertiesKey
+      )
+  }
+  catch {
+    print(error.localizedDescription, to: &swifterr.stream)
+    exit(1)
+  }
+}
+else {
+  print(errorBag!.localizedDescription, to: &swifterr.stream)
+  exit(3)
+}
+
+exit(0)
--- a/Library/Homebrew/test/cask/cli/quarantine_spec.rb
+++ b/Library/Homebrew/test/cask/cli/quarantine_spec.rb
@@ -1,0 +1,213 @@
+describe Hbc::Quarantine, :cask do
+  matcher :be_quarantined do
+    match do |path|
+      expect(
+        described_class.detect(path),
+      ).to be true
+    end
+  end
+
+  describe "by default" do
+    it "quarantines a nice fresh Cask" do
+      Hbc::CLI::Install.run("local-transmission")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("local-transmission")),
+      ).to be_installed
+
+      expect(
+        Hbc::Config.global.appdir.join("Transmission.app"),
+      ).to be_quarantined
+    end
+
+    it "quarantines Cask fetches" do
+      Hbc::CLI::Fetch.run("local-transmission")
+      local_transmission = Hbc::CaskLoader.load(cask_path("local-transmission"))
+      cached_location = Hbc::Download.new(local_transmission, force: false, quarantine: false).perform
+
+      expect(cached_location).to be_quarantined
+    end
+
+    it "quarantines Cask audits" do
+      Hbc::CLI::Audit.run("local-transmission", "--download")
+
+      local_transmission = Hbc::CaskLoader.load(cask_path("local-transmission"))
+      cached_location = Hbc::Download.new(local_transmission, force: false, quarantine: false).perform
+
+      expect(cached_location).to be_quarantined
+    end
+
+    it "quarantines dmg-based Casks" do
+      Hbc::CLI::Install.run("container-dmg")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("container-dmg")),
+      ).to be_installed
+
+      expect(Hbc::Config.global.appdir.join("container")).to be_quarantined
+    end
+
+    it "quarantines tar-gz-based Casks" do
+      Hbc::CLI::Install.run("container-tar-gz")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("container-tar-gz")),
+      ).to be_installed
+
+      expect(Hbc::Config.global.appdir.join("container")).to be_quarantined
+    end
+
+    it "quarantines xar-based Casks" do
+      Hbc::CLI::Install.run("container-xar")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("container-xar")),
+      ).to be_installed
+
+      expect(Hbc::Config.global.appdir.join("container")).to be_quarantined
+    end
+
+    it "quarantines pure bzip2-based Casks" do
+      Hbc::CLI::Install.run("container-bzip2")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("container-bzip2")),
+      ).to be_installed
+
+      expect(Hbc::Config.global.appdir.join("container")).to be_quarantined
+    end
+
+    it "quarantines pure gzip-based Casks" do
+      Hbc::CLI::Install.run("container-gzip")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("container-gzip")),
+      ).to be_installed
+
+      expect(Hbc::Config.global.appdir.join("container")).to be_quarantined
+    end
+
+    it "quarantines the pkg in naked-pkg-based Casks" do
+      Hbc::CLI::Install.run("container-pkg")
+
+      naked_pkg = Hbc::CaskLoader.load(cask_path("container-pkg"))
+
+      expect(naked_pkg).to be_installed
+
+      expect(
+        Hbc::Caskroom.path.join("container-pkg", naked_pkg.version, "container.pkg"),
+      ).to be_quarantined
+    end
+
+    it "quarantines a nested container" do
+      Hbc::CLI::Install.run("nested-app")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("nested-app")),
+      ).to be_installed
+
+      expect(Hbc::Config.global.appdir.join("MyNestedApp.app")).to be_quarantined
+    end
+  end
+
+  describe "when disabled" do
+    it "does not quarantine even a nice, fresh Cask" do
+      Hbc::CLI::Install.run("local-transmission", "--no-quarantine")
+
+      expect(
+        Hbc::CaskLoader.load(cask_path("local-transmission")),
+      ).to be_installed
+
+      expect(Hbc::Config.global.appdir.join("Transmission.app")).to_not be_quarantined
+    end
+
+    it "does not quarantine Cask fetches" do
+      Hbc::CLI::Fetch.run("local-transmission", "--no-quarantine")
+      local_transmission = Hbc::CaskLoader.load(cask_path("local-transmission"))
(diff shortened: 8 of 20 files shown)


vitorgalvao:
> If merged, this pull request closes Homebrew/homebrew-cask#22388, Homebrew/homebrew-cask#48862.

Not quite, as it’d still need the `--no-quarantine` flag to close the issues. But that is definitely the least important part.

reitermarkus:
This needs some tests.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/swift
+
+import Foundation
+import CoreServices
+
+let QUARANTINE_SUCCESS: Int = 0
+let QUARANTINE_FAILURE: Int = 255
```
reitermarkus: Feels weird not to just use `0/1`.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,45 @@
...
+
+if (CommandLine.arguments.count < 3) {
+  print("Insufficient parameters: this script needs file_path and data_url")
+  exit(Int32(QUARANTINE_FAILURE))
+}
+
+let filepath = CommandLine.arguments[1]
+let agent: String = "Homebrew-Cask"
+let bundle: String = "sh.brew.cask"
```
reitermarkus: Could this be a path instead? I think I remember macOS using the path to the executable instead of a bundle ID if it's not an app bundle.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,45 @@
...
+  }
+  catch {
+    let errorString: String = String.init(format: "Homebrew-Cask quarantiner: unable to quarantine %@: %@", (dataLocationUrl.absoluteString)!, error.localizedDescription)
+    NSLog(errorString)
+    exit(Int32(QUARANTINE_FAILURE))
+  }
+}
+else {
+  let errorString: String = String.init(format: "Homebrew-Cask quarantiner: unable to find and quarantine %@: %@", (dataLocationUrl.absoluteString)!, errorBag!.localizedDescription)
+  NSLog(errorString)
```
reitermarkus: Print to `stderr` instead of `NSLog`.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,45 @@
...
+
+let dataLocationUrl: NSURL = NSURL.init(fileURLWithPath: filepath)
+var errorBag: NSError?
+
+if (dataLocationUrl.checkResourceIsReachableAndReturnError(&errorBag)) {
+  do {
+    try dataLocationUrl.setResourceValue(quarantineProperties, forKey: URLResourceKey.quarantinePropertiesKey)
+  }
+  catch {
+    let errorString: String = String.init(format: "Homebrew-Cask quarantiner: unable to quarantine %@: %@", (dataLocationUrl.absoluteString)!, error.localizedDescription)
```
reitermarkus: Use interpolation for this.

```swift
"Homebrew-Cask quarantiner: unable to quarantine \(dataLocationUrl.absoluteString!): \(error.localizedDescription)"
```

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/swift
+
+import Foundation
+import CoreServices
+
+let QUARANTINE_SUCCESS: Int = 0
+let QUARANTINE_FAILURE: Int = 255
+
+if (CommandLine.arguments.count < 3) {
+  print("Insufficient parameters: this script needs file_path and data_url")
```
reitermarkus: Print to `stderr` instead.

amyspark:
@vitorgalvao: which commands should `--no-quarantine` affect?
@reitermarkus: I've replied to your comments inline. About tests - these would need more logic to detect the `com.apple.quarantine` extended attribute, and besides, every time we quarantine in the tests, we'll pollute the system database.

reitermarkus:
> these would need more logic to detect the `com.apple.quarantine` extended attribute

Yes, that's what the tests should be checking.

> we'll pollute the system database

How? Everytime you download something with a browser it will also “pollute” the system database, so I don't think that is a reason not to have tests.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
amyspark: This one is inferred from [the parent process's bundle identifier](https://github.com/phracker/MacOSX-SDKs/blob/9fc3ed0ad0345950ac25c28695b0427846eea966/MacOSX10.13.sdk/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Headers/LSQuarantine.h#L50-L54); since this is Swift's interpreter, it'd be set to `nil`.

vitorgalvao:
> which commands should `--no-quarantine` affect?

Anything that downloads, so `fetch`, `install`, `audit --download`.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
amyspark: In Apple's docs [here](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW1) a bundle identifier is a reverse-DNS [Uniform Type Identifier](https://developer.apple.com/library/archive/documentation/Miscellaneous/Reference/UTIRef/Articles/System-DeclaredUniformTypeIdentifiers.html); I don't know if Gatekeeper enforces this, though.

reitermarkus:
> Anything that downloads, so `fetch`, `install`, `audit --download`.

Although this should probably be a global option, so you can set it in `HOMEBREW_CASK_OPTS`.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
reitermarkus: I think the `TCC.db` uses paths for executables and the bundle ID for apps, and this seems similar to that.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,45 @@
...
+  exit(Int32(QUARANTINE_FAILURE))
+}
+
+let filepath = CommandLine.arguments[1]
+let agent: String = "Homebrew-Cask"
+let bundle: String = "sh.brew.cask"
+let dataUrl: String = CommandLine.arguments[2]
+
+let quarantineProperties: NSMutableDictionary = NSMutableDictionary()
```
reitermarkus: Can this be a normal Swift dictionary which is casted afterwards?

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
amyspark: I've removed this altogether in an upcoming commit (it's not needed at all).

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
amyspark: Yes, will do.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/swift
+
+import Foundation
+import CoreServices
+
+let QUARANTINE_SUCCESS: Int = 0
+let QUARANTINE_FAILURE: Int = 1
```
reitermarkus: With `0/1`, I meant just using `exit(1)` instead of `exit(Int32(QUARANTINE_FAILURE))`.

amyspark:
I've found that after #4401's changes, the extended attributes are not preserved between `ditto`'s unpacking and the move to the Caskroom. Upon further inspection (`xattr` over `path` and `unpack_dir`), [`FileUtils`'s `copy_entry`](https://github.com/Homebrew/brew/blob/8f3a80b562e911ec34cde3fcb8dcd8789c0affaf/Library/Homebrew/unpack_strategy/directory.rb#L19) is not preserving the extended attributes necessary to keep the app quarantined.

chdiza:
I thought it had been decided that there will not be any swift code added to Homebrew.  Someone added a swift script once and it was quickly (and correctly) removed.  @MikeMcQuaid is this ringing a bell?

amyspark:
@MikeMcQuaid , @chdiza: as I explained in homebrew/homebrew-cask#48862, the Swift approach enables us to quickly access MacOS's APIs without having to worry about compilation, FFI, etc. The only drawback comes (as you noticed) from having to detect errors based on the exit code.

If you want to use pure Ruby, you can, but the drawbacks outweigh the benefits IMO. There are two approaches:
 1. Manipulate the SQLite database and the extended attributes by hand. This is really prone to breaking Gatekeeper's policy database, which is why I abandoned this approach.
 2. Program the logic in a Objective-C library which exposes a C API, and then use FFI to call them from within Ruby. You'd get the ease of access that comes from using Ruby, but you'd have to worry about how to compile and install the library when installing or upgrading Homebrew.

claui:
(Just a PSA, I’m only catching up on previous discussions, no strong opinion.)

Points I’ve seen made against using Swift in Homebrew:

- Swift’s [bus factor](https://github.com/Homebrew/brew/pull/2793#issuecomment-309882360) among maintainers,

- [giving up monolinguality](https://github.com/Homebrew/brew/pull/2793#issuecomment-310322889)

- and Swift’s [track record in language stability](https://github.com/Homebrew/brew/pull/2793#issuecomment-310360819).

chdiza:
Also, and correct me if I'm wrong here, but doesn't the user have to have Xcode or the CLT installed in order to run a swift script?  My impression is that lots of cask users won't have those installed because they're only looking to install `.app`s and not cmdline stuff.

vitorgalvao:
> and Swift’s [track record in language stability](https://github.com/Homebrew/brew/pull/2793#issuecomment-310360819).

> but doesn't the user have to have Xcode or the CLT installed in order to run a swift script?

It’s my understanding that from Mojave on, Swift will be (finally) included as part of macOS. That will remove the need for further software installation (Xcode/CLT) and (presumably, I’m guessing on this one) backward-incompatibility will start to be handled more carefully.

reitermarkus:
> Swift’s bus factor among maintainers

I still don't understand that argument when AppleScript is the alternative. There is much more documentation for Swift than for AppleScript, whose syntax is also much more convoluted.

claui:
> I still don't understand that argument when AppleScript is the alternative. There is much more documentation for Swift than for AppleScript

Well documented or not, people need to practise for a language to be useful to them.

I could probably churn out an IOKit driver in AppleScript for you in a day.<sup>1</sup> I’d still barely manage to write a Hello World in Swift to save my life.

> whose syntax is also much more convoluted.

Absolutely.

___

<sup>1</sup> Dramatization

vitorgalvao:
> Well documented or not, people need to practise for a language to be useful to them.

True. But as someone who writes a bunch of AppleScript regularly (and loathes every second of it) and close to no Swift, I still feel that if I needed to do something a bit more advanced, I’d more easily find information on the latter than the former.

AppleScript’s (and JXA’s) documentation is terrible and code you find online is typically bad (as in, it does the job but it’s longer and less optimal than it could be).

All that said, if we’re going to try to rewrite the scripts in AppleScript, we should at least do it in JXA. Not only is the syntax saner, it can do some things AppleScript can’t, and it should be more auditable by more people (being JavaScript).

reitermarkus:
> Well documented or not, people need to practise for a language to be useful to them.

I'd argue it's harder to practise without good documentation.

claui:
> I'd argue it's harder to practise without good documentation.

Very much so! Learning AppleScript was a royal pain.

amyspark:
@MikeMcQuaid, I'll try and port the Swift script to AppleScript (I'll need to switch to the original Objective-C classes first).
In the meanwhile, I'll change the current script according to your comments.

claui:
> I'll try and port the Swift script to AppleScript

@amyspark Already done. PR incoming.

MikeMcQuaid:
@amyspark Thanks! We can iterate on the Swift code first if that's easier.

@claui Nice work.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,42 @@
...
+import Foundation
+import CoreServices
+
+if (CommandLine.arguments.count < 4) {
+  print("Insufficient parameters: this script needs file_path, data_url, and origin_url")
+  exit(1)
+}
+
+let filepath = CommandLine.arguments[1]
+let agent: String = "Homebrew-Cask"
```
MikeMcQuaid: This could be hardcoded, right?

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,42 @@
...
+
+let dataLocationUrl: NSURL = NSURL.init(fileURLWithPath: filepath)
+var errorBag: NSError?
+
+if (dataLocationUrl.checkResourceIsReachableAndReturnError(&errorBag)) {
+  do {
+    try dataLocationUrl.setResourceValue(quarantineProperties as NSDictionary, forKey: URLResourceKey.quarantinePropertiesKey)
+  }
+  catch {
+    let errorString: String = "Homebrew-Cask quarantiner: unable to quarantine \(dataLocationUrl.absoluteString!): \(error.localizedDescription)"
```
MikeMcQuaid: This could just `exit(1)` and let the Ruby code handle the messaging.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,42 @@
...
+    NSLog(errorString)
+    exit(1)
+  }
+}
+else {
+  let errorString: String = "Homebrew-Cask quarantiner: unable to quarantine \(dataLocationUrl.absoluteString!): \(errorBag!.localizedDescription)"
```
MikeMcQuaid: This could just `exit(2)` and let the Ruby code handle the messaging.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
amyspark: I'll use `exit(1)` and use stderr to recover the error message.

amyspark:
@MikeMcQuaid, error handling has been moved to Ruby (courtesy of [Erica Sadun's stderr stream](https://ericasadun.com/2015/05/22/swift-logging/)). If @claui needs it, I can streamline error detection so that it depends only on `stderr`'s contents.

MikeMcQuaid:
Looking at 
https://github.com/Homebrew/brew/pull/4665 the Swift code here doesn't seem significantly less complex to warrant the Xcode/CLT/Mojave/newer versions of macOS dependency. That said, I don't feel strongly enough about this that if @Homebrew/cask does feel strongly (and this is handled appropriately and throws user-friendly errors on old versions of macOS/Swift not being present) that the Swift code should not be used.

claui:
I’d be fine with either – especially in light of what @vitorgalvao wrote on Swift probably shipping with Mojave, and @reitermarkus’s point on more accessible documentation.

vitorgalvao:
> That said, I don't feel strongly enough about this that if @Homebrew/cask does feel strongly

I don’t. If this can be adequately handled in AppleScript, I have no objection to using it over Swift<sup>1</sup>. I still think JXA would be a stronger choice, but I’m not going to push the point further.

I echo @claui’s sentiment. I’m fine with either.

---

1. Small caveat that AppleScript has given some issues on headless cask uninstalls.

amyspark:
@claui, @vitorgalvao, @MikeMcQuaid, thanks for your comments!

About the extra dependencies - If there's no `swift` available, quarantining is silently disabled (see `Quarantine.available?`); in this way, we don't break tests which depend on CLI output. The only (warning) message appears when using `--no-quarantine`, which has its own tests.

If you feel like it -but it'll probably be overkill- I can add @claui's script as a fallback (crediting you, of course!).

claui:
> I can add @claui's script as a fallback (crediting you, of course!).

That’s very considerate of you but I’d prefer that we stick with one single script for maintainability’s sake. Even if that means no quarantine at all for some users.

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def status(file, command = SystemCommand)
+      command.run("/usr/bin/xattr",
+                    args:        ["-p", QUARANTINE_ATTRIBUTE, file],
+                    print_stderr: false).stdout
+    end
+
+    def cask(cask, downloaded_path, command = SystemCommand)
+      odebug "Quarantining #{downloaded_path}"
+      quarantiner = command.run(swift, args: ["#{HOMEBREW_LIBRARY_PATH}/cask/lib/hbc/utils/quarantine.swift", downloaded_path, cask.url.to_s, cask.homepage.to_s])
```
claui: This line is over 163 chars long. Can you reformat this for readability?

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      raise CaskError, "when quarantining #{downloaded_path}: #{quarantiner.stderr}" unless quarantiner.success?
```
claui: I may be overly pedantic here but to me, guard clauses that don’t start before column 80 are confusing and super easy to overlook.

Consider rewriting this in the if-return-else-raise form.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/swift
+
+import Foundation
+import CoreServices
```
claui: Is the `CoreServices` framework actually used?

Library/Homebrew/cask/lib/hbc/quarantine.rb
amyspark: I'll pull the script path into a constant and reformat the rest into a new line.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
amyspark: No, it's a leftover from the full-featured script I used as basis. Will remove.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -172,6 +175,17 @@ def extract_primary_container
...
         primary_container.extract_nestedly(to: @cask.staged_path, basename: basename, verbose: verbose?)
       end
+
+      return unless quarantine? && Quarantine.available?
+
+      unless Quarantine.detect(@downloaded_path)
+        raise CaskError, "#{@downloaded_path} was not quarantined properly."
+      end
+
+      @cask.staged_path.each_child do |path|
```
reitermarkus: Why not just call `Quarantine.all` on `@cask.staged_path`?

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      raise CaskError, "when quarantining #{downloaded_path}: #{quarantiner.stderr}"
+    end
+
+    def all(downloaded_path, base_path, command = SystemCommand)
+      odebug "Propagating quarantine status of #{downloaded_path}"
+
+      quarantine_status = Quarantine.status(downloaded_path)
+
+      list_of_artifacts = Dir.glob(base_path/"**/*")
```
reitermarkus: For an application, this will actually loop through each and every file it contains, and not just add the attribute to the enclosing `.app` bundle.

Library/Homebrew/cask/lib/hbc/installer.rb
amyspark: Unpacking non-DMG containers (`.tar.gz` and other Unix-based stuff) does not propagate the container's extended attributes; instead, the *contents* of `@cask.staged_path` themselves have to be quarantined after extraction.

Library/Homebrew/cask/lib/hbc/quarantine.rb
amyspark: What you say is correct; although Gatekeeper only needs `com.apple.quarantine` on the app bundle, manually installing an app from the Finder quarantines the whole directory tree. This replicates that behavior.

Library/Homebrew/cask/lib/hbc/installer.rb
amyspark: (I'll simplify and remove this logic to `Quarantine.all`.)

Library/Homebrew/cask/lib/hbc/quarantine.rb
reitermarkus: It's just very inefficient to shell out to `xattr` for every file individually. `xattr` seems to support multiple files as arguments. There are some similar things where we use `xargs` to split the arguments.

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -0,0 +1,70 @@
...
+      quarantine_status = status(downloaded_path)
+
+      base_path.each_child do |child|
+        next if detect(child)
+
+        quarantiner = command.run("/usr/bin/xattr",
+                                  args: ["-w", "-r", QUARANTINE_ATTRIBUTE, quarantine_status, child])
+
+        next if quarantiner.success?
+        raise CaskError, "when quarantining files in #{child}: #{quarantiner.stderr}"
```
reitermarkus: `"Failed to quarantine #{child}: #{quarantiner.stderr}"`

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -0,0 +1,70 @@
...
+                                  cask.url.to_s,
+                                  cask.homepage.to_s,
+                                ])
+
+      return if quarantiner.success?
+
+      raise CaskError, "when quarantining #{downloaded_path}: #{quarantiner.stderr}"
+    end
+
+    def all(downloaded_path, base_path, command = SystemCommand)
```
reitermarkus: Maybe call this `propagate` instead of `all` and use keyword arguments `from/to`.

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -0,0 +1,70 @@
...
+                                args: [
+                                  QUARANTINE_SCRIPT,
+                                  downloaded_path,
+                                  cask.url.to_s,
+                                  cask.homepage.to_s,
+                                ])
+
+      return if quarantiner.success?
+
+      raise CaskError, "when quarantining #{downloaded_path}: #{quarantiner.stderr}"
```
reitermarkus: `"Failed to quarantine #{downloaded_path}: #{quarantiner.stderr}"`

Or maybe create a `QuarantineError` class, since this format is already used multiple times.

Library/Homebrew/cask/lib/hbc/quarantine.rb
amyspark: Moved to its own `CaskQuarantineError` class.

Library/Homebrew/cask/lib/hbc/quarantine.rb
amyspark: Done. I also replaced all parameters with keyword-based ones.

Library/Homebrew/cask/lib/hbc/quarantine.rb
amyspark: > Or maybe create a QuarantineError class, since this format is already used multiple times.

Done :+1:

amyspark:
Ping @reitermarkus @MikeMcQuaid for further review.

Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
```diff
@@ -7,11 +7,12 @@ class AbstractCommand
...
-      option "--debug",         :debug,         false
-      option "--verbose",       :verbose,       false
-      option "--outdated",      :outdated_only, false
-      option "--require-sha",   :require_sha,   false
+      option "--[no-]binaries",   :binaries,      true
+      option "--debug",           :debug,         false
+      option "--verbose",         :verbose,       false
+      option "--outdated",        :outdated_only, false
+      option "--require-sha",     :require_sha,   false
+      option "--[no-]quarantine", :quarantine,    true
```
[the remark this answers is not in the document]
vitorgalvao: > When would you not want to quarantine?

When you quarantine but want to bypass it for a single single app, the way to do it is to right-click the app and select “Open”.

I’m sure there are users that have the following method when installing with HBC:

1. `brew cask install {whatever}`.
2. Invoke app launcher (Alfred, Spotlight).
3. Call app.

In this situation you’ll be greeted with a “this app can’t be opened” and will have to manually go into `/Applications` and perform the procedure. `--no-quarantine` eases that step for apps that you know are not signed but trust.

Realistically, I predict quarantining will be one of those features we will get complains about (think dev shops that have automated setups for their machines) and will get feature requests to add option to not quarantine, so might as well do it preemptively.

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -86,6 +88,7 @@ def install
       uninstall_existing_cask if @reinstall
 
       oh1 "Installing Cask #{Formatter.identifier(@cask)}"
+      opoo "MacOS's Gatekeeper has been disabled for this Cask" unless quarantine?
```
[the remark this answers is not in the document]
amyspark: This is a warning message for @vitorgalvao's case. I'll remove it if you want.

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -0,0 +1,39 @@
...
+
+let dataLocationUrl: NSURL = NSURL.init(fileURLWithPath: CommandLine.arguments[1])
+var errorBag: NSError?
+
+if (dataLocationUrl.checkResourceIsReachableAndReturnError(&errorBag)) {
+  do {
+    try dataLocationUrl.setResourceValue(quarantineProperties as NSDictionary, forKey: URLResourceKey.quarantinePropertiesKey)
+  }
+  catch {
+    print(error.localizedDescription, to: &swifterr.stream)
```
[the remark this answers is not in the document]
amyspark: `checkResourceIsReachableAndReturnError` and `setResourceValue` return errors in different ways (the first uses NSError, the second uses Swift exceptions).

Library/Homebrew/cask/lib/hbc/installer.rb
```diff
@@ -172,6 +175,10 @@ def extract_primary_container
       else
         primary_container.extract_nestedly(to: @cask.staged_path, basename: basename, verbose: verbose?)
       end
+
+      return unless quarantine? && Quarantine.available?
```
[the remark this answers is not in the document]
amyspark: I can separate into two `return unless`; `brew style` complains when using `if`:

```
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/installer.rb:184:7: C: [Corrected] Favor unless over if for negative conditions.
      return if !quarantine?
      ^^^^^^^^^^^^^^^^^^^^^^
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/installer.rb:185:7: C: [Corrected] Favor unless over if for negative conditions.
      return if !Quarantine.available?
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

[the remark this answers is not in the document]
MikeMcQuaid: Either two or a single `if ||`

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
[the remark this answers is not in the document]
MikeMcQuaid: Ok

amyspark:
Thanks for your review @MikeMcQuaid! Please let me know if everything is OK for you.

Library/Homebrew/cask/lib/hbc/installer.rb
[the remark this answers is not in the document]
MikeMcQuaid: I'm fine either way but this should be `macOS` rather than `MacOS`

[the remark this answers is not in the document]
MikeMcQuaid: And also technically `macOS'` rather than `macOS's`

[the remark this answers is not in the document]
chdiza: No, because "macOS" isn't plural.

amyspark:
Every nitpick is an opportunity to improve. Thanks again for your comments @MikeMcQuaid.
@vitorgalvao, did I forget any Cask issues to close?
@reitermarkus, ping again 😅

Library/Homebrew/cask/lib/hbc/artifact/moved.rb
```diff
@@ -65,10 +65,11 @@ def move_back(skip: false, force: false, command: nil, **options)
         ohai "Backing #{self.class.english_name} '#{target.basename}' up to '#{source}'."
         source.dirname.mkpath
 
+        # We need to preserve extended attributes between copies.
         if target.parent.writable?
-          FileUtils.cp_r(target, source)
+          command.run!("/bin/cp", args: ["-pR", target, source])
```
reitermarkus: The only difference here is `sudo`, so maybe make this a variable which is set by `target.parent.writable?`.

Library/Homebrew/cask/lib/hbc/exceptions.rb
```diff
@@ -149,4 +149,23 @@ def to_s
...
+  class CaskQuarantineError < CaskError
+    attr_reader :path, :reason
+
+    def initialize(path, reason)
+      @path = path
+      @reason = reason
+    end
+
+    def to_s
+      "Failed to quarantine #{path}" << (reason.empty? ? "." : ": #{reason}")
```
reitermarkus: Interpolate this instead.

Library/Homebrew/cask/lib/hbc/exceptions.rb
```diff
@@ -149,4 +149,23 @@ def to_s
...
+    end
+  end
+
+  class CaskQuarantinePropagationError < CaskQuarantineError
+    def to_s
+      "Failed to quarantine one or more files within #{path}" << (reason.empty? ? "." : ": #{reason}")
```
reitermarkus: Interpolate this instead.

Library/Homebrew/test/cask/cli/quarantine_spec.rb
```diff
@@ -0,0 +1,91 @@
+describe Hbc::Quarantine, :cask do
+  describe "by default" do
+    it "quarantines Cask installs" do
+      Hbc::CLI::Install.run("local-transmission")
+
+      expect(Hbc::CaskLoader.load(cask_path("local-transmission"))).to be_installed
+      expect(Hbc::Config.global.appdir.join("Transmission.app")).to be_a_directory
+
+      expect(
+        described_class.detect(Hbc::Config.global.appdir.join("Transmission.app")),
```
reitermarkus: Add a custom `be_quarantined` matcher for this. Here's an example:

https://github.com/Homebrew/brew/blob/459e2bfa48ebed4bf49d6aebb68df1b416e9b0f3/Library/Homebrew/test/missing_formula_spec.rb#L11-L15

Library/Homebrew/cask/lib/hbc/exceptions.rb
amyspark: I've copied these from `CaskInvalidError` (earlier in the file). Would you like me to change all of them to interpolation or just the quarantine ones?

vitorgalvao:
> @vitorgalvao, did I forget any Cask issues to close?

Don’t auto-close https://github.com/Homebrew/homebrew-cask/issues/48862, since we may have to revise the policy depending on the installed artifact instead of scraping it completely. After this change is live, I’ll likely replace that issue with a new one.

Library/Homebrew/cask/lib/hbc/exceptions.rb
amyspark: Done and done 😆

reitermarkus:
> However, if one extends #4731 to `uncompressed.rb`, naked `dmg`, `tar.gz` and `zip` are fixed; the rest still need `propagate`.

I don't think this matters, because the only time we use naked archives is as a workaround.

> Would it be a better option to issue a single `xattr` and quarantine everything in `@cask.staged_path` instead of forking for each `child`?

Forking is definitely the bottleneck here, so yes.

amyspark:
@reitermarkus, done. Thanks for the review.

amyspark:
Heads up: I found that some Casks ship with the wrong permissions (tried to install Fritzing for a course and quarantine failed on an internal `.git` folder). I've added a line that ensures *all* extracted files are writable previous to quarantining, but I'm not sure if this should rather go in `dmg.rb`.

amyspark:
Everyone, please let me know if this is ready for merging. Let me know if I should merge/rebase on master, as I forked this over two weeks ago.

Library/Homebrew/cask/lib/hbc/cli/upgrade.rb
```diff
@@ -36,7 +36,11 @@ def run
 
           old_cask = CaskLoader.load(old_cask.installed_caskfile)
 
-          old_cask_installer = Installer.new(old_cask, binaries: binaries?, verbose: verbose?, force: force?, upgrade: true)
+          old_cask_installer =
+            Installer.new(old_cask, binaries:  binaries?,
+                                    verbose:   verbose?,
+                                    force:     force?,
+                                    upgrade:    true)
```
claui: Funny how RuboCop fails to notice this alignment error.

Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb
claui: I know this PR has been nitpicked to pieces but can you please add a line for the `--no-quarantine` option to `Library/Homebrew/manpages/brew-cask.1.md`?

amyspark: No problem! Let me know if the next commit is OK for you.

amyspark:
@claui, @reitermarkus -- I've rebased+squashed on my side. Before cleaning this pull, let me know if the current state is OK with you.

claui:
I ran a final smoke test on a clean VM (30-ish casks). Everything seems to work great so far.
Some PKG installers didn’t quarantine their payload but that’s expected.
Three Casks failed to install but I could confirm those errors are unrelated.
LGTM :shipit:

claui:
@amyspark Ready for cleanup 😉

amyspark:
Ok everyone, incoming rebase+squash! 🚀

claui:
Thanks a lot @amyspark ✨ for tackling this! This feature was much needed.

claui:
[BONAPARTE – Quarantine](https://youtu.be/2L1cbTFJ0iI)

PR #4662, opened 2018-08-11
Resolve URL to get real file extension.

reitermarkus:

-----

This resolves the URL to get the correct file extension, and symlinks the cache entry with the guessed file extension to the one with the real one, should they differ.

Fixes https://github.com/Homebrew/brew/issues/4648.


reitermarkus:
@JCount and @MikeMcQuaid, there is a problem with using symlinks. We also need the resolved URL for `basename_without_params`, so in case an unpack strategy needs the basename, it will need a network call, even when the download was cached. Should we switch to using `xattr`, which makes both the cached download and the basename work offline?

Library/Homebrew/cleanup.rb
```diff
@@ -47,7 +47,7 @@ def prune?(days)
     end
 
     def stale?(scrub = false)
-      return false unless file?
+      return false unless file? || symlink?
```
MikeMcQuaid: Can this be an `if`; `unless ||` is hard to parse.

MikeMcQuaid:
> @JCount and @MikeMcQuaid, there is a problem with using symlinks. We also need the resolved URL for basename_without_params, so in case an unpack strategy needs the basename, it will do a network call, even when the download was cached. Should we switch to using xattr, which makes both the cached download and the basename work offline?

@reitermarkus Having multiple symlinks might be better here. Does `xattr` work on Linux and on all filesystems (e.g. NFS)?

JCount:
NFS appears to not support extended attributes in versions other than the most recent NFSv4, and even then perhaps not robustly enough.

- Also, according to macOS High Sierra's `man nfs`:
```
MacOS makes heavy use of extended attributes, resource forks, and possibly ACLs.
NFSv2 and NFSv3 do not support these operations over the wire.
```

- Since we cannot? break NFS support, this paints a bit of a grim picture for using extended attributes.

Any thoughts or points of agreement/disagreement, @MikeMcQuaid?

MikeMcQuaid:
Linux is a more pressing concern here, I think, but good to know about NFS.

JCount:
Just to provide a bit more info about NFSv4 support, the relevant documentation is: https://tools.ietf.org/html/rfc8276

reitermarkus:
I'm not sure where to put the file with the original basename then.

JCount:
This is less than ideal, but could the strategy for doing this be split by OS type? I can't think of a situation off of the top of my head in which the local cache would need to be portable to or from Linux.

reitermarkus:
> could the strategy for doing this be split by OS type?

Not sure what that would accomplish. It's not really an OS issue but a file system issue.

JCount:
Well, the subset of supported file systems are different

JCount:
Whoops, sorry.

reitermarkus:
> the subset of supported file systems are different, especially for macOS

You can set `HOMEBREW_CACHE` to point to an NFS mount on macOS, right?

reitermarkus:
If we extend the current approach with symlinks again, we would end up with symlinks to symlinks. It's probably best to switch to an URL-based approach, which also includes the original basename.

Here's why I'm thinking:


```
HOMEBREW_CACHE
├── downloads
│   └── #{sha256}-#{real_basename}
└── foo--1.2.3 -> downloads/#{sha256}-#{real_basename}
```

JCount:
Hmmm. Your idea has merit. I haven't thought of any major disadvantages there'd be in that approach, and so far it seems to hold water. I wonder what Mr. McQuaid thinks.

iMichka:
Just a small comment on NFS: there are many users who run Linuxbrew on a NFS drive. Some of them are even on Windows ... not that we support that kind of scenario, but if the cache could be file-system agnostic that would be great :)

MikeMcQuaid:
> Here's why I'm thinking:

Can you give an example with the variables filled out with those of an actual formula e.g. `wget`? Thanks!

reitermarkus:
```
HOMEBREW_CACHE
├── downloads
│   └── 58981f4995520d916d686bb8745780b8e27b7f8249daf3e5d733e84b12ba1a9a-wget-1.19.5.tar.gz
└── wget--1.19.5 -> downloads/58981f4995520d916d686bb8745780b8e27b7f8249daf3e5d733e84b12ba1a9a-wget-1.19.5.tar.gz
```

reitermarkus:
Note that the `sha256` in this case is for the URL, otherwise it doesn't work for unversioned casks.

MikeMcQuaid:
@reitermarkus Is that basename/filename the final target or the original? Assuming the prior I think the `sha256` being for the URL makes sense 👍

chdiza:
> HOMEBREW_CACHE
> ├── downloads
> │   └── #{sha256}-#{real_basename}
> └── foo--1.2.3 -> downloads/#{sha256}-#{real_basename}

Please, no!  That makes manual manipulation of the cache a giant PITA.

sjackman:
I feel like this situation comes up relatively infrequently. Code we add a DSL to specify the true extension, for cases when it's guessed incorrect from the URL?

reitermarkus:
> Please, no! That makes manual manipulation of the cache a giant PITA.

There should be no need to ever manipulate the cache manually, otherwise this should be considered a bug.

> I feel like this situation comes up relatively infrequently. Code we add a DSL to specify the true extension, for cases when it's guessed incorrect from the URL?

It's very frequent for casks, so a DSL just for this is a maintenance no-go. Also, this is not just for the extension, it's also for the basename.

reitermarkus:
> Is that basename/filename the final target or the original?

It's the final target.

MikeMcQuaid:
@reitermarkus Cool. Yeh, I think in that case the deduplication added by this would be good.

chdiza:
> There should be no need to ever manipulate the cache manually, otherwise this should be considered a bug.

(a) Why should there be no need?  Why should the cache be some opaque blob?

(b) But there often is such a need.  E.g., when `cleanup` fails to clean something up in a way that was declared non-buggy.

Entirely different reasons to do this were mentioned in the cited thread.  Also, in that thread, @MikeMcQuaid [agreed](https://github.com/Homebrew/brew/issues/568#issuecomment-235399114).

MikeMcQuaid:
I've changed my mind in the last two years.

reitermarkus:
> manual extraction from and insertion into the cache should be as easy as possible

With deduplication and getting the real basename in mind, this is as easy is possible, unless you have an easier solution which also fixes both of these problems.

sjackman:
> This resolves the URL to get the correct file extension, and symlinks the cache entry with the guessed file extension to the one with the real one, **should they differ**.

Do all cache entries use this directory structure, or only those where the guessed and real file extension differ?

sjackman:
Which is the "guessed" extension and which is the "real" extension? Could you please provide an example?

reitermarkus:
> Do all cache entries use this directory structure, or only those where the guessed and real file extension differ?

With the current implementation (which is missing the basename), only the differing ones.

> Which is the "guessed" extension and which is the "real" extension? Could you please provide an example?

The URL is `https://example.org/1.2.3/download`, for example. Then the guessed extension is `""`. This URL redirects to `https://example.org/foo-1.2.3.tar.gz`, so the real extension is `.tar.gz`. Same goes for the basename.

sjackman:
I'd be inclined to add to the DSL something like `extension "tar.gz"` for this case, so that the extension is known offline and without having to resolve the URL. My opinion though, and you've expressed that you're not a fan of this option.

Why is the basename of the URL needed? For Homebrew's purposes, can the basename be `#{formula.name}--#{version}.#{extension}` rather than the basename of the URL?

Library/Homebrew/cmd/update-report.rb
```diff
@@ -253,6 +271,85 @@ def migrate_cache_entries_to_double_dashes(initial_version)
...
+    return if ENV.key?("HOMEBREW_DISABLE_LOAD_FORMULA")
+
+    ohai "Migrating cache entries..."
+
+    formula_downloaders = if HOMEBREW_CACHE.directory?
+      HOMEBREW_CACHE.children
+                    .select(&:file?)
+                    .map { |child| child.basename.to_s.sub(/\-\-.*/, "") }
+                    .uniq
+                    .map { |formula| Formula[formula] rescue nil } # rubocop:disable Style/RescueModifier
```
MikeMcQuaid: Can we make this into a bigger method with a proper rescue rather than needing the `rescue nil` here?

Library/Homebrew/cmd/update-report.rb
```diff
@@ -253,6 +271,85 @@ def migrate_cache_entries_to_double_dashes(initial_version)
...
+                    .map { |resource| [resource.downloader, resource.download_name, resource.version] }
+    else
+      []
+    end
+
+    cask_downloaders = if (HOMEBREW_CACHE/"Cask").directory?
+      (HOMEBREW_CACHE/"Cask").children
+                             .map { |child| child.basename.to_s.sub(/\-\-.*/, "") }
+                             .uniq
+                             .map { |cask| Hbc::CaskLoader.load(cask) rescue nil } # rubocop:disable Style/RescueModifier
```
MikeMcQuaid: Can we make this into a bigger method with a proper rescue rather than needing the `rescue nil` here? This line is also too long.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -253,6 +271,85 @@ def migrate_cache_entries_to_double_dashes(initial_version)
     end
   end
 
+  def migrate_cache_entries_to_symlinks(initial_version)
```
MikeMcQuaid: I'm wondering about getting some of the `update-report` logic into another file; it's getting pretty huge.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -206,6 +208,18 @@ def migrate_legacy_cache_if_necessary
     end
   end
 
+  def formula_resources(formula)
+    specs = [*formula.stable, *formula.devel, *formula.head]
```
MikeMcQuaid: What's with all the `*`?

Library/Homebrew/cmd/update-report.rb
```diff
@@ -214,25 +228,29 @@ def migrate_cache_entries_to_double_dashes(initial_version)
...
+          uri = URI(url)
+          uri.query ? "#{uri.path}?#{uri.query}" : uri.path
+        else
+          url
+        end
+
+        extname = nil
+
+        Pathname.new(uri_path).ascend do |path|
+          ext = path.extname[/[^?&]+/]
```
MikeMcQuaid: Not clear what this regex is doing

Library/Homebrew/cmd/update-report.rb
```diff
@@ -253,6 +271,85 @@ def migrate_cache_entries_to_double_dashes(initial_version)
...
+    end
+
+    downloaders = formula_downloaders + cask_downloaders
+
+    downloaders.each do |downloader, name, version|
+      next unless downloader.respond_to?(:symlink_location)
+
+      url = downloader.url
+
+      uri_path = if URI::DEFAULT_PARSER.make_regexp =~ url
```
MikeMcQuaid: This is identical to code above.

Library/Homebrew/cmd/update-report.rb
```diff
@@ -253,6 +271,85 @@ def migrate_cache_entries_to_double_dashes(initial_version)
...
+        uri = URI(url)
+        uri.query ? "#{uri.path}?#{uri.query}" : uri.path
+      else
+        url
+      end
+
+      extname = nil
+
+      Pathname.new(uri_path).ascend do |path|
```
MikeMcQuaid: This is identical to code above.

Library/Homebrew/test/cmd/update-report_spec.rb
```diff
@@ -43,4 +43,42 @@
...
+    let(:formula_name) { "foo" }
+    let(:f) {
+      formula formula_name do
+        url "https://example.com/foo-1.2.3.tar.gz"
+        version "1.2.3"
+      end
+    }
+    let(:old_cache_file) { HOMEBREW_CACHE/"#{formula_name}--1.2.3.tar.gz" }
+    let(:new_cache_file) { HOMEBREW_CACHE/"downloads/5994e3a27baa3f448a001fb071ab1f0bf25c87aebcb254d91a6d0b02f46eef86--foo-1.2.3.tar.gz" }
+    let(:new_cache_symlink) { HOMEBREW_CACHE/"#{formula_name}--1.2.3" }
```
MikeMcQuaid: Why are there no file extensions now?

Library/Homebrew/cmd/update-report.rb
reitermarkus: Yep.

Library/Homebrew/test/cmd/update-report_spec.rb
reitermarkus: Why should there be? We are only using the real basename now.

Library/Homebrew/cmd/update-report.rb
reitermarkus: Any of these could be `nil`, so `[*nil, *nil, *nil]` would be `[]`.

Library/Homebrew/test/cmd/update-report_spec.rb
MikeMcQuaid: It feels strange if we're going to expose these to people to not have them have extensions for ease of extraction etc.

Library/Homebrew/cmd/update-report.rb
MikeMcQuaid: Can you use `.compact` instead? That makes the intent a bit clearer.

Library/Homebrew/cmd/update-report.rb
MikeMcQuaid: @reitermarkus Cool, that'd be good, thanks 👍

MikeMcQuaid: @reitermarkus Maybe just all the various migrations.

Library/Homebrew/test/cmd/update-report_spec.rb
reitermarkus: Where do we expose this to people?

MikeMcQuaid: People expect (not unreasonably as it's been the case for all Homebrew's history) for the cache to contain a list of filenames that they can extract.

chdiza: It is harder than it should be when the target of the symlink is prefixed with a long bunch of hex gibberish.  The source tarballs don't need to have the sha256 in their very filenames, surely.

MikeMcQuaid: @reitermarkus The "guessed" extension is correct the majority of the times. A gentle reminder that formulae account for >95% of installs (comparing with the new cask analytics) and until recently we only used the formulae url's extension for performing downloads anyway. To optimise for the "what if the guessed extension is wrong" case is to reduce the overall aesthetics and usability of these cache files for a minority corner case and that's not a good design. 

Consistency is not king; usability and doing what people expect is.

Library/Homebrew/cleanup.rb
```diff
@@ -209,6 +209,19 @@ def cleanup_logs
       end
     end
 
+    def cleanup_unreferenced_downloads
+      return if dry_run?
+      return unless (cache/"downloads").directory?
+
+      downloads = (cache/"downloads").children.reject { |path| path.incomplete? } # rubocop:disable Style/SymbolProc
```
MikeMcQuaid: Why can't this use `:incomplete?`

Library/Homebrew/cleanup.rb
```diff
@@ -48,7 +48,7 @@ def prune?(days)
     end
 
     def stale?(scrub = false)
-      return false unless file?
+      return false unless file? || (symlink? && resolved_path.file?)
```
MikeMcQuaid: Can this be an `if` instead? Better still, make it a separate `return` if possible.

Library/Homebrew/update_migrator.rb
```diff
@@ -0,0 +1,39 @@
...
+  module_function
+
+  def formula_resources(formula)
+    specs = [formula.stable, formula.devel, formula.head].compact
+
+    [*formula.bottle&.resource] + specs.flat_map do |spec|
+      [
+        spec,
+        *spec.resources.values,
+        *spec.patches.select(&:external?).map(&:resource),
```
MikeMcQuaid: How about `compact` and/or `flatten` instead of `*` for these?

Library/Homebrew/update_migrator/cache_entries_to_symlinks.rb
```diff
@@ -0,0 +1,93 @@
...
+    else
+      []
+    end
+
+    cask_downloaders = if (HOMEBREW_CACHE/"Cask").directory?
+      (HOMEBREW_CACHE/"Cask").children
+                             .reject(&:symlink?)
+                             .select(&:file?)
+                             .map { |child| child.basename.to_s.sub(/\-\-.*/, "") }
+                             .uniq
```
MikeMcQuaid: Can the logic above maybe be shared between the above two variable definitions into a separate method?

Library/Homebrew/update_migrator.rb
```diff
@@ -0,0 +1,39 @@
+require "update_migrator/cache_entries_to_double_dashes"
+require "update_migrator/cache_entries_to_symlinks"
+require "update_migrator/legacy_cache"
+require "update_migrator/legacy_keg_symlinks"
+require "update_migrator/legacy_repository"
```
MikeMcQuaid: I think it would be nicer to have all the `UpdateMigrator` logic in a single file.

Library/Homebrew/brew.rb
```diff
@@ -76,7 +78,7 @@
   end
 
   # Migrate LinkedKegs/PinnedKegs if update didn't already do so
-  migrate_legacy_keg_symlinks_if_necessary
```
MikeMcQuaid: Nice catch.

Library/Homebrew/cleanup.rb
reitermarkus: Because the refinement scope doesn't work like that, sadly. It would call `.incomplete?` in the file where `reject` is defined rather than here.

MikeMcQuaid: Gotcha. Mind adding a comment above this pointing this out so people don't think it's a smell in future? Thanks!

Library/Homebrew/update_migrator.rb
reitermarkus: Why? That would solve the problem of having a big file by moving it to another big file.

Library/Homebrew/cleanup.rb
reitermarkus: Can't be separate because it is an OR.  And `return if !file? && !(symlink? && resolved_path.file?)` isn't more readable.

Library/Homebrew/update_migrator.rb
reitermarkus: `*` is more concise here, I think.

Library/Homebrew/update_migrator.rb
MikeMcQuaid: That's how we define modules in the Homebrew codebase. I think having two bigger files is still an improvement over one. Also, I think the migration logic is sensible to keep separate from the update logic itself

reitermarkus: > I think having two bigger files is still an improvement over one.

And having separate, easily readable, files for each migration is even more of an improvement.

> Also,  I think the migration logic is sensible to keep separate from the update logic itself.

What do you mean by that? That's what this is doing.

MikeMcQuaid: > > Also, I think the migration logic is sensible to keep separate from the update logic itself.
>
> What do you mean by that? That's what this is doing.

I mean moving to a separate file is a sufficient improvement to still be worth doing.

> And having separate, easily readable, files for each migration is even more of an improvement.

When it's different to the majority of the Homebrew codebase: it makes things harder to follow rather than easier. For example, if I want to find the definition of `UpdateMigrator` I need to look at multiple files rather than one and there's no longer a mapping between the test filename and the implementation filename.

Library/Homebrew/cleanup.rb
MikeMcQuaid: Sorry, missed this. I do genuinely find that more readable. An alternative would be to split the `symlink? && resolved_path.file?` into a separate variable.

MikeMcQuaid:
@reitermarkus 🚢 when ready. Will plan on tagging a day or so after the dust has settled and any issues have been ironed out.

sjackman:
Can I confirm that the contents of the cache directory (after this PR) are the same in most cases, and only differ for formulae where the guessed and real file extension differ?

reitermarkus:
@sjackman, downloads are now stored as `HOMEBREW_CACHE/downloads/#{url_sha256}--#{base name}` and a symlink with the guessed extension pointing to it.

sjackman:
Is that true of bottles as well? Or only non-bottle resources? Is it true of patches?

reitermarkus:
All file downloads will use this naming.

chdiza:
I am still wondering why the SHA has to be part of the filename.

reitermarkus:
@chdiza, it's not the file's SHA256, it's the URL's SHA256, which is used as an ID for deduplication.

DomT4:
Keep whacking into these sorts of messages on `update`, FYI:
```
==> Migrating cache entries...
Error: No such file or directory @ realpath_rec - /usr/local/Homebrew/freetds-1.00.97.mojave.bottle.tar.gz
```
Am up to date with master, so the latest commit doesn't seem to have resolved the hiccup.

reitermarkus:
@DomT4, what's the backtrace with `brew update --debug`?

reitermarkus:
@DomT4, no worries. 😉

chdiza:
>  it's not the file's SHA256, it's the URL's SHA256, which is used as an ID for deduplication.

I still don't see why this is needed.  What "duplication" might occur?

reitermarkus:
Using the same URL in two different formulae/casks. And even if there is no duplication, we still need to somehow map the URL to the cache name with the real basename.

chdiza:
This has made the cache appreciably and predictably worse.  All in order to fix #4648, a rare and niche case with an already-implemented workaround.

> Using the same URL in two different formulae/casks

Never before has this demanded a SHA in a tarball filename.

> we still need to somehow map the URL to the cache name with the real basename.

The real basename can be auto-renamed to start with `foo--3.2` just after it's downloaded.  Then it can just live in the cache like normal and not be tucked away and symlinked to.  That is, it downloads as `downloads/SHA-something.ext`, and HB will compare `SHA` to the url in the `foo` formula and say "OK, that tarball should be named `foo--3.2.ext`", and rename it while moving it out of `downloads`.  Maybe there will be occasional duplication, like if some other formula `bar` uses `foo` as a resource, but it's been like this already for a long time and I don't think there've been too many complaints.

At the very  least one could make the SHA not be the initial segment of the filename.  Make it be `foo--3.2_THEHUGEGIGANTICSHATHATMAKESDIRLISTINGSNEARLYIMPOSSIBLETOREADINBOTHTHETERMINALANDTHEFINDER.tar.bz2`.  That way they'll sort in a helpful order, and when the name is truncated for display by the Finder or whatever file manager one uses, one won't just be looking at a big list of SHAs, or a big list of "`THEHUGEGIGANTICSHATHATMAKESDIRL...o--3.2.tar.bz2`"

commitay:
This seems to have broken the apache mirror download strategy.

https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/29803/

```console
$ brew fetch -f ant
==> Downloading https://www.apache.org/dyn/closer.cgi?path=ant/binaries/apache-ant-1.10.5-bin.tar.xz
Downloaded to: /Users/commitay/Library/Caches/Homebrew/downloads/2d9de475d484b0b09e19cf0240d5861788e0db04cd088ef94b4af0086735c1a7--apache-ant-1.10.5-bin.tar.xz
SHA256: 29dbd20376b07647bbbbc57b622336781e62c2fd364658c5cf877946fdb89a13
Warning: Formula reports different SHA256: cebb705dbbe26a41d359b8be08ec066caba4e8686670070ce44bbf2b57ae113f
Resource: ivy
==> Downloading https://www.apache.org/dyn/closer.cgi?path=ant/ivy/2.4.0/apache-ivy-2.4.0-bin.tar.gz
Downloaded to: /Users/commitay/Library/Caches/Homebrew/downloads/d7c4a713694cabcfdaa178db886b9f04febb4655388accfcdacfb4dc509c4205--apache-ivy-2.4.0-bin.tar.gz
SHA256: f2974e3782a1db7337532e9d313e73dcabc7564286d3c97af7f51b37d03a11e8
Warning: Resource ivy reports different SHA256: 7a3d13a80b69d71608191463dfc2a74fff8ef638ce0208e70d54d28ba9785ee9
Resource: bcel
==> Downloading https://www.apache.org/dyn/closer.cgi?path=commons/bcel/binaries/bcel-6.2-bin.tar.gz
Downloaded to: /Users/commitay/Library/Caches/Homebrew/downloads/5a4f007849f7038f2d35716ba56a2e59f735c8db9c5fbb05931d3a448ce3778e--bcel-6.2-bin.tar.gz
SHA256: ced73c575210b3201f44057531514ffa31b00995793ea835ca2f7964a9c7213e
Warning: Resource bcel reports different SHA256: e5963ed50ef1f243f852a27efaf898c050a3b39721d147ccda8418c0b7255955
```
```console
$ brew fetch -f maven
==> Downloading https://www.apache.org/dyn/closer.cgi?path=maven/maven-3/3.5.4/binaries/apache-maven-3.5.4-bin.tar.gz
Downloaded to: /Users/commitay/Library/Caches/Homebrew/downloads/573a40954b5757e2dabd96fac6dd460eb754551fc48517d6624069b2274eb197--apache-maven-3.5.4-bin.tar.gz
SHA256: 6bfe8fb71b8a5331400f7dd00e915006b5604babe09fcb9fc31068a3ac3156d0
Warning: Formula reports different SHA256: ce50b1c91364cb77efe3776f756a6d92b76d9038b0a0782f7d53acf1e997a14d
```
```console
$ brew fetch -f tomcat
==> Downloading https://www.apache.org/dyn/closer.cgi?path=/tomcat/tomcat-9/v9.0.11/bin/apache-tomcat-9.0.11.tar.gz
Downloaded to: /Users/commitay/Library/Caches/Homebrew/downloads/bff80dbdaafc2073ebc77ac13a479a4a0f02ffbf58ac6413ca1948b480a17b87--apache-tomcat-9.0.11.tar.gz
SHA256: 5a40da1bdcb205cfd8ffdd03671c9969e6b3d5064ff48765b10ab4a303a56105
Warning: Formula reports different SHA256: baa62b93404d0e795657adb1c0cd41e83a415f10870e657d4441c104bb176471
Resource: fulldocs
==> Downloading https://www.apache.org/dyn/closer.cgi?path=/tomcat/tomcat-9/v9.0.11/bin/apache-tomcat-9.0.11-fulldocs.tar.gz
Downloaded to: /Users/commitay/Library/Caches/Homebrew/downloads/37352cbb473ac3bed811b43f3d473bc7b60efa4077c9d231cdde14eda00f00f9--apache-tomcat-9.0.11-fulldocs.tar.gz
SHA256: ba7c22e9371ddf90160bc8a18438ea8d3983b7d6edd2bcc3e53fc678ef58873b
Warning: Resource fulldocs reports different SHA256: 96c21710e039ddbd631c1d3dbefee73cc40a769e04d035581c3f0c241925c23b
```

```
fbcaa8c85ae42f127fd94a5d82f86a3eafb34848 is the first bad commit
commit fbcaa8c85ae42f127fd94a5d82f86a3eafb34848
Author: Markus Reiter <me@reitermark.us>
Date:   Fri Aug 10 04:11:54 2018 +0200

    Resolve URL to get real file extension.
```

bergman:
I believe this PR introduced a bug with handling of `Content-Disposition` headers that supply both `filename=` and `filename*=`.

Example:
`Content-Disposition: attachment; filename="foo.bar"; filename*=UTF-8''foo.bar`

```ruby
irb(main):013:0> header = "Content-Disposition: attachment; filename=\"foo.bar\"; filename*=UTF-8''foo.bar"
=> "Content-Disposition: attachment; filename=\"foo.bar\"; filename*=UTF-8''foo.bar"
irb(main):014:0> header[/^Content\-Disposition:\s*attachment;\s*filename=(["']?)(.+)\1$/i, 2]
=> "\"foo.bar\"; filename*=UTF-8''foo.bar"
```

Expected output would be `"foo.bar"`

bergman:
This regex seems to fare better. Using regex for this feels a bit shaky. Also it doesn't even try to read the `filename*` parameter which can specify an encoding.

```ruby
[/^Content\-Disposition:\s*attachment;\s*filename=("|')?([^\s'";]+)(?(1)\1|);?/i, 2]
```

sjackman:
Why is the extension of the resolved URL expected to be more sensible than the unresolved URL? It seems just as likely the other way around, that the resolved URL extension is incorrect.

reitermarkus:
> Why is the extension of the resolved URL expected to be more sensible than the unresolved URL?

Because some unresolved URLs have no extension at all and only after following all redirects do you have the actual filename.

> It seems just as likely the other way around, that the resolved URL extension is incorrect.

Do you have any example where this happens?

sjackman:
>> It seems just as likely the other way around, that the resolved URL extension is incorrect.

> Do you have any example where this happens?

No, sorry. I'll keep quiet unless I find one.

bergman:
> @bergman, do you have a real URL to test this with?

@reitermarkus, no, unfortunately the one that prompted my investigation into this was an internal artifactory URL. I'm sure there are examples in the wild since it seems to be the recommended way of supporting the non-ISO-8859-1  `filename*` while being able to fall back on `filename`.

These are the full list of headers from our internal artifactory:
```
$ curl -I https://artifactory.example.org/some/path/foobar.jar
HTTP/1.1 200 OK
Server: nginx
Date: Wed, 05 Sep 2018 08:21:13 GMT
Content-Type: application/java-archive
Content-Length: 20198
Connection: keep-alive
Keep-Alive: timeout=10
X-Artifactory-Id: 11929058f2a4bc9a:-3c35a1ff:165131a473a:-7ffd
Last-Modified: Tue, 28 Aug 2018 05:30:48 GMT
ETag: 7d45fe4ac13d3848e9a662e547bd9baaf8f93e8e
X-Checksum-Sha1: 7d45fe4ac13d3848e9a662e547bd9baaf8f93e8e
X-Checksum-Sha256: 9ad8532360a5ad0618e6b594ec29d917f39863613c4311200f766aa5ae22df68
X-Checksum-Md5: 584418ea5d7ce3c7ae4c62314aee1367
Accept-Ranges: bytes
X-Artifactory-Filename: foobar.jar
Content-Disposition: attachment; filename="foobar.jar"; filename*=UTF-8''foobar.jar
```

bergman:
@pviniciusfm, it looks like this was fixed in 17c3a66265b38084a16b3ebaf057d1f9c8b4f805

Thanks @reitermarkus 🎉

pviniciusfm:
Yeap it's working 🎉

PR #4665, opened 2018-08-12
Sample AppleScript implementation of quarantining

claui:
- [ ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change? **See discussion**
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb). **amyspark has**

-----

See [discussion](https://github.com/Homebrew/brew/pull/4656#pullrequestreview-145473511).

~~Does not yet contain latest availability checks.~~ **Edit: Rebased.**

### DIFF
--- a/Library/Homebrew/cask/lib/hbc/utils/quarantine.applescript
+++ b/Library/Homebrew/cask/lib/hbc/utils/quarantine.applescript
@@ -1,0 +1,34 @@
+#!/usr/bin/osascript --
+
+use framework "Foundation"
+
+on run(argv)
+  if argv's length < 3 then
+    error "Insufficient parameters"
+  end if
+
+  set quarantineProperties to { ¬
+    kLSQuarantineAgentNameKey: "Homebrew-Cask", ¬
+    kLSQuarantineTypeKey: my kLSQuarantineTypeWebDownload, ¬
+    kLSQuarantineDataURLKey: argv's item 2, ¬
+    kLSQuarantineOriginURLKey: argv's item 3 ¬
+  }
+
+  set dataLocationUrl to my NSURL's fileURLWithPath: argv's item 1
+
+  set {resourceReachable, errorBag} to dataLocationUrl's ¬
+    checkResourceIsReachableAndReturnError: reference
+
+  if resourceReachable then
+    set {quarantineSuccess, errorBag} to dataLocationUrl's ¬
+      setResourceValue: quarantineProperties ¬
+      forKey: my NSURLQuarantinePropertiesKey ¬
+      |error|: reference
+
+    if not quarantineSuccess then
+      error errorBag's localizedDescription as text
+    end if
+  else
+    error errorBag's localizedDescription as text
+  end if
+end run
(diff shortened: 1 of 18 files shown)


Library/Homebrew/cask/lib/hbc/utils/quarantine.applescript
```diff
@@ -0,0 +1,39 @@
...
+
+    if not dataLocationUrl's setResourceValue: quarantineProperties ¬
+      forKey: my NSURLQuarantinePropertiesKey ¬
+      |error|: missing value then
+
+      error "Homebrew-Cask quarantiner: unable to quarantine " ¬
+        & dataLocationUrl's absoluteString as text
+    end if
+  else
+    error "Homebrew-Cask quarantiner: unable to quarantine " ¬
```
MikeMcQuaid: Can this use a custom exit code?

claui: That’s an annoying limitation of AppleScript. We can choose an error number to be displayed on stderr but we have no influence on the exit code.

claui: (My usual workaround for that is to output JSON and have that unwrapped by a shell script wrapper. I don’t feel that would be worth the effort here though.)

amyspark: I've updated the Swift script to just output the error message; then, we could use `command.run(...).stderr.empty?` instead of `success?` to check for errors.

Library/Homebrew/cask/lib/hbc/utils/quarantine.applescript
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  set dataLocationUrl to my NSURL's fileURLWithPath: filepath
+
+  if dataLocationUrl's checkResourceIsReachableAndReturnError: ¬
+    missing value then
+
+    if not dataLocationUrl's setResourceValue: quarantineProperties ¬
+      forKey: my NSURLQuarantinePropertiesKey ¬
+      |error|: missing value then
```
amyspark: Isn't it possible in AppleScript to get the `NSError` if this throws?

Library/Homebrew/cask/lib/hbc/utils/quarantine.applescript
MikeMcQuaid: Yeh, outputting just the `dataLocationUrl's absoluteString` would be preferable, I think.

Library/Homebrew/cask/lib/hbc/utils/quarantine.applescript
claui: I thought it wasn’t possible but now that you’ve mentioned it, I found out [there’s a convoluted, poorly documented way](https://www.macosxautomation.com/applescript/apps/BridgePlus_Manual/Pages/runSpotlightQuery_inFolders_error_.html) to do it!

Updated.

Library/Homebrew/cask/lib/hbc/utils/quarantine.applescript
claui: Updated the script so it only outputs the error message.

MikeMcQuaid:
Thanks for this @claui but think we're 👍 on the Swift version now.

PR #4677, opened 2018-08-14
Update docs for artifact and bottle variables

jonchang:

-----

Currently the manpages are very confusing about `HOMEBREW_ARTIFACT_DOMAIN` and `HOMEBREW_BOTTLE_DOMAIN`. They are described as such:

```
HOMEBREW_ARTIFACT_DOMAIN
       If set, instructs Homebrew to use the given URL as a download mirror
       for bottles and binaries.

HOMEBREW_BOTTLE_DOMAIN
       If set, instructs Homebrew to use the given URL as a download mirror
       for bottles.
```

One might be lulled into believing these variables do essentially the same thing, but this is not the case. `HOMEBREW_ARTIFACT_DOMAIN` will prepend the its value to all downloads, while `HOMEBREW_BOTTLE_DOMAIN` has the effect of setting `root_url` with its value for all bottle downloads, with no effect on source downloads.

I'm agnostic with respect to the exact wording but it definitely should be changed from what it is now.


MikeMcQuaid:
Nice work as usual @jonchang!

PR #4685, opened 2018-08-14
check_non_libraries: *.so.* is a valid library

sjackman:

-----

`*.so.*` is a valid library file name on Linux. This change could be made specific to Linux, but it's simpler to apply it to both platforms. The list of valid library extensions already includes `*.so` on both platforms.

https://github.com/Homebrew/brew/blob/0d81ad059dd233e2780601e0ee525dc45426657f/Library/Homebrew/formula_cellar_checks.rb#L62-L63

### DIFF
--- a/Library/Homebrew/formula_cellar_checks.rb
+++ b/Library/Homebrew/formula_cellar_checks.rb
@@ -56,14 +56,19 @@ module FormulaCellarChecks
     EOS
   end
 
+  VALID_LIBRARY_EXTENSIONS = %w[.a .dylib .framework .jnilib .la .o .so .jar .prl .pm .sh].freeze
+
+  def valid_library_extension?(filename)
+    VALID_LIBRARY_EXTENSIONS.include? filename.extname
+  end
+  alias generic_valid_library_extension? valid_library_extension?
+
   def check_non_libraries
     return unless formula.lib.directory?
 
-    valid_extensions = %w[.a .dylib .framework .jnilib .la .o .so
-                          .jar .prl .pm .sh]
     non_libraries = formula.lib.children.reject do |g|
       next true if g.directory?
-      valid_extensions.include? g.extname
+      valid_library_extension? g
     end
     return if non_libraries.empty?
 
--- a/Library/Homebrew/extend/os/formula_cellar_checks.rb
+++ b/Library/Homebrew/extend/os/formula_cellar_checks.rb
@@ -1,1 +1,5 @@
-require "extend/os/mac/formula_cellar_checks" if OS.mac?
+if OS.mac?
+  require "extend/os/mac/formula_cellar_checks"
+elsif OS.linux?
+  require "extend/os/linux/formula_cellar_checks"
+end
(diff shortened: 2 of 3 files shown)


MikeMcQuaid:
> The list of valid library extensions already includes *.so on both platforms.

Can you elaborate on this?

DomT4:
I'm not sure I agree with making this platform-agnostic.

I know there's plenty of core formulae on the Homebrew side that produce `blah.so`, thinking especially of things that create Python bindings, but I can't remember a time where we have explicitly considered `blah.so.1.2.3` anything other than an upstream issue that needs resolution.

sjackman:
>> The list of valid library extensions already includes *.so on both platforms.

> Can you elaborate on this?

The current list of `valid_extensions` includes `*.so`.

https://github.com/Homebrew/brew/blob/0d81ad059dd233e2780601e0ee525dc45426657f/Library/Homebrew/formula_cellar_checks.rb#L62-L63

iMichka:
> but I can't remember a time where we have explicitly considered blah.so.1.2.3 anything other than an upstream issue that needs resolution.

I think I had the problem with `lua`, never contacted them as I saw that the discussion about their shared libraries has been ongoing for years ...

MikeMcQuaid:
> The current list of valid_extensions includes *.so.

I see. Yeh, that's a valid plugin on macOS but I agree with @DomT4: `*.so.*` should only be whitelisted on Linux.

sjackman:
I've reworked this PR to make the change only on Linux.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for the review, and thanks for merging!

Library/Homebrew/formula_cellar_checks.rb
```diff
@@ -56,14 +56,19 @@ def check_jars
     EOS
   end
 
+  VALID_LIBRARY_EXTENSIONS = %w[.a .dylib .framework .jnilib .la .o .so .jar .prl .pm .sh].freeze
+
+  def valid_library_extension?(filename)
+    VALID_LIBRARY_EXTENSIONS.include? filename.extname
```
maxim-belkin: Apologies for a late comment. Would it make sense to define `VALID_OS_SPECIFIC_LIBRARY_EXTENSIONS` variable and then do 
```ruby
(VALID_LIBRARY_EXTENSIONS + VALID_OS_SPECIFIC_LIBRARY_EXTENSIONS).include? filename.extname
```
?

Reason: `.dylib` and `.framework` are not used as library extensions on Linux.

sjackman: Good point. Within the current framework, `extend/os/mac/formula_cellar_checks.rb` should override `valid_library_extension?` and `.dylib .framework` should be moved to this macOS-specific function, similar to how Linux handles `*.so.*` in `extend/os/linux/formula_cellar_checks.rb`. Would you like to open that PR?

maxim-belkin: > Would you like to open that PR?

1. PR against Linuxbrew first?
1. Can I get couple of extra hours in a day to go please? :)

sjackman: Hah! For tricky PRs, I'd suggest Linuxbrew first, but this one is pretty straight forward, so a PR direct to Homebrew/brew ought to be fine. Tag me, and I'll review.

PR #4686, opened 2018-08-14
X11Requirement: Recent X11 client is sufficient

sjackman:

-----

Building and running X11 programs requires sufficiently recent X11 client tools. The X11Requirement is satisfied if the X11 client tools (represented here by the version of xdpyinfo) are sufficiently recent, regarldess of the X server version (which can also change at run time by changing the DISPLAY environment variable).

On Linuxbrew, installing `linuxbrew/xorg/xorg` installs a recent version of the X11 client tools, but does not install an X server.

### DIFF
--- a/Library/Homebrew/requirements/x11_requirement.rb
+++ b/Library/Homebrew/requirements/x11_requirement.rb
@@ -27,20 +27,16 @@ class X11Requirement < Requirement
 
   satisfy build_env: false do
     if which_xorg = which("Xorg")
-      version = Utils.popen_read which_xorg, "-version", err: :out
-      next false unless $CHILD_STATUS.success?
-      version = version[/X Server (\d+\.\d+\.\d+)/, 1]
-      next false unless version
-      Version.new(version) >= min_version
-    elsif which_xdpyinfo = which("xdpyinfo")
-      version = Utils.popen_read which_xdpyinfo, "-version"
-      next false unless $CHILD_STATUS.success?
-      version = version[/^xdpyinfo (\d+\.\d+\.\d+)/, 1]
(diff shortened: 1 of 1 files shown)


Library/Homebrew/requirements/x11_requirement.rb
```diff
@@ -27,20 +27,16 @@ def min_xdpyinfo_version
...
-      Version.new(version) >= min_version
-    elsif which_xdpyinfo = which("xdpyinfo")
-      version = Utils.popen_read which_xdpyinfo, "-version"
-      next false unless $CHILD_STATUS.success?
-      version = version[/^xdpyinfo (\d+\.\d+\.\d+)/, 1]
-      next false unless version
-      Version.new(version) >= min_xdpyinfo_version
-    else
-      false
+      version = Utils.popen_read(which_xorg, "-version", err: :out)[/X Server (\d+\.\d+\.\d+)/, 1]
```
MikeMcQuaid: If this doesn't return a string then the `[/X Server` will fail. It'd be better to keep the version separate like before with `version = version[/X Server (\d+\.\d+\.\d+)/, 1]`

sjackman: `Utils.popen_read` always returns a string, even when it fails.

```
❯❯❯ brew irb <<<'Utils.popen_read("foo")'
Switch to inspect mode.
Utils.popen_read("foo")
""
```

MikeMcQuaid: Cool 👍

sjackman:
Merged. Thanks for the review, Mike!

PR #4688, opened 2018-08-14
setup-analytics: Ensure that git is available

sjackman:
Return and do nothing when git is not available.


MikeMcQuaid:
What's the current behaviour?

sjackman:
The current behaviour seems fine. It behave as though `HOMEBREW_NO_ANALYTICS_THIS_RUN=1`.
I'm reverting the corresponding change in Linuxbrew/brew with PR https://github.com/Linuxbrew/brew/pull/810

sjackman:
Closed without merging.

Issue #4641, opened 2018-08-08
bump-formula-pr fails with key not found: "clone_url"


alexanderharm:
- [ ] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [ ] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [X] ran `brew update` and can still reproduce the problem?
- [X] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [X] ran `brew config` and `brew doctor` and included their output with your issue?

Using the `bump-formula-pr` command trying to update `s-nail` formula fails with

```
brew bump-formula-pr --strict --url=http://www.sdaoden.eu/downloads/s-nail-14.9.11.tar.gz --sha256=279202687409b8e7b4f267e178aed1bd4c68b79c01c10b80f07197f2f73d6695 s-nail
Already up-to-date.
==> replace /^  revision \d+\n(\n(  head "))?/m with "\\2"
==> replace /https:\/\/www\.sdaoden\.eu\/downloads\/s\-nail\-14\.9\.10\.tar\.gz/ with "http://www.sdaoden.eu/downloads/s-nail-14.9.11.tar.gz"
==> replace "87ade7bbeb110250afce8221b2bb553a3d9661327dfca3f622994fb017c49ccb" with "279202687409b8e7b4f267e178aed1bd4c68b79c01c10b80f07197f2f73d6695"
Error: key not found: "clone_url"
Please report this bug:
  https://docs.brew.sh/Troubleshooting
/usr/local/Homebrew/Library/Homebrew/dev-cmd/bump-formula-pr.rb:288:in `fetch'
/usr/local/Homebrew/Library/Homebrew/dev-cmd/bump-formula-pr.rb:288:in `block in bump_formula_pr'
/usr/local/Homebrew/Library/Homebrew/extend/pathname.rb:271:in `block in cd'
/usr/local/Homebrew/Library/Homebrew/extend/pathname.rb:271:in `chdir'
[...]

PR #4692, opened 2018-08-15
utils/github: Exception handling

GauthamGoli:

-----
Fixes https://github.com/Homebrew/brew/issues/4641
and 
https://github.com/Homebrew/brew/issues/4641#issuecomment-413188584 
Rate limit exceptions are already being handled so, I noticed that curl command's return code is 0 even when http response code is not 200


MikeMcQuaid:
Good catch @GauthamGoli!

PR #4696, opened 2018-08-15
BottleSpecification: Do not skip bottle relocation on Linux

sjackman:

-----

Bottles must be relocated on Linux to find the brewed glibc and gcc.

### DIFF
--- a/Library/Homebrew/extend/os/linux/software_spec.rb
+++ b/Library/Homebrew/extend/os/linux/software_spec.rb
@@ -1,0 +1,5 @@
+class BottleSpecification
+  def skip_relocation?
+    false
+  end
+end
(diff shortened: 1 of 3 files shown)


MikeMcQuaid:
Alternatively: what about never generating `:any_no_relocation` bottles on Linux?

sjackman:
In Brewsci/bio we build bottles for both macOS and Linux. The bottles may be detected as `:any_skip_relocation`, which may be true on macOS, though it's not true on Linux, due to using a brewed `glibc` and/or `gcc`.

I could change all the bottles to `:any` in Brewsci/bio. That loses the benefit of `:any_skip_relocation` on macOS. Thoughts?

sjackman:
Thanks for the review, Mike.

PR #4716, opened 2018-08-17
brew.sh: Allow developers to specify curl/git paths

woodruffw:

-----

This introduces the `HOMEBREW_CURL_PATH` and `HOMEBREW_GIT_PATH`
variables, which are used to populate `HOMEBREW_CURL` and `HOMEBREW_GIT`,
respectively, when in developer mode (`HOMEBREW_DEVELOPER`).

Both paths are expected to be valid executables and must
have the executable bit set.

I split the two checks apart for readability (each is right below the normal assignment for `HOMEBREW_CURL`/`HOMEBREW_GIT`), but I could merge the `HOMEBREW_DEVELOPER` check if anybody prefers.


Library/Homebrew/brew.sh
```diff
@@ -129,6 +129,11 @@ else
   HOMEBREW_CURL="curl"
 fi
 
+if [[ -n "$HOMEBREW_DEVELOPER" && -x "$HOMEBREW_CURL_PATH" ]]
```
MikeMcQuaid: What about putting this as an `elif` above; it feels like if `HOMEBREW_FORCE_BREWED_CURL` is set this shouldn't override it?

Library/Homebrew/brew.sh
```diff
@@ -138,6 +143,11 @@ else
   HOMEBREW_GIT="git"
 fi
 
+if [[ -n "$HOMEBREW_DEVELOPER" && -x "$HOMEBREW_GIT_PATH" ]]
```
MikeMcQuaid: What about putting this as an `elif` above; it feels like if `HOMEBREW_FORCE_BREWED_GIT ` is set this shouldn't override it?

MikeMcQuaid:
Thanks again @woodruffw!

PR #4722, opened 2018-08-18
Remove the chdir call that was already moved to the ancestor

kabel:

-----

528b4b367ed2a4026c321e2f2e503ace4548b9d2 moved the chdir definition to the parent class and call it in the parent's stage. chdir was getting called twice for archives with two nested directories.

Example of problem: 

* Formula resource to download test.zip
* test.zip only contains: `top_dir/middle_dir/leaf_dir/file.h`
* the current directory after staging the resource is `middle_dir` instead of `top_dir`


reitermarkus:
That doesn't include `NoUnzipCurlDownloadStrategy`. Then again, if we don't unzip, there's nothing to `chdir` to.

PR #4729, opened 2018-08-20
formula_installer: permit multi-lua dependency trees

DomT4:

-----

We've sufficiently installed `lua` and `lua@5.1` in a way that makes it very very difficult for them to bump into each other without significant nudging. If `python` is safe enough `lua` should be as well.

### DIFF
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -187,7 +187,7 @@ class FormulaInstaller
     recursive_runtime_formulae.each do |f|
       name = f.name
       unversioned_name, = name.split("@")
-      next if unversioned_name == "python"
+      next if ["python", "lua"].include?(unversioned_name)
       version_hash[unversioned_name] ||= Set.new
       version_hash[unversioned_name] << name
       next if version_hash[unversioned_name].length < 2


MikeMcQuaid:
> If python is safe enough lua should be as well.

I'm afraid I'm not sure I agree with that. I do think, though, that this check could be replaced with a check in `brew linkage --test` as the issue is multiple linkage in the same formula rather than multiple recursive dependencies, necessarily.

DomT4:
Post-https://github.com/Homebrew/homebrew-core/pull/28143 it should be essentially impossible for anything to trip over dual-`lua`s. I've yet to see any evidence counter to this in the few months since in version bumps, revision bumps, etc.

> as the issue is multiple linkage in the same formula rather than multiple recursive dependencies, necessarily.

I think that has the potential to be a fairly hefty "necessarily", though, especially around common symbols sharing a namespace, I'd imagine? _(Could be wildly wrong; hardly my area of "expertise")_. It feels to me like the current check isn't undesirable but can be a little inflexible.

DomT4:
Still think this is a reasonable step to take but don't have the energy to pointlessly argue it out, so closing.

MikeMcQuaid:
@DomT4 FYI you won the argument 😂 I should have said but I’ve added to my list to remove all this logic and rely on ‘brew linkage’ data instead. Work for you?

PR #4730, opened 2018-08-20
Update installed caskfiles of modified casks.

reitermarkus:

-----

Often casks get updated without a version bump to fix a broken `uninstall` or add something to `zap`. This replaces the installed caskfile if it differs from the current one and the version hasn't changed.

### DIFF
--- a/Library/Homebrew/cmd/update-report.rb
+++ b/Library/Homebrew/cmd/update-report.rb
@@ -7,6 +7,7 @@ require "migrator"
 require "formulary"
 require "descriptions"
 require "cleanup"
+require "hbc/cask_loader"
 
 module Homebrew
   module_function
@@ -125,6 +126,7 @@ module Homebrew
         hub.dump
         hub.reporters.each(&:migrate_tap_migration)
         hub.reporters.each(&:migrate_formula_rename)
+        hub.reporters.each(&:migrate_modified_casks)
         Descriptions.update_cache(hub)
       end
       puts if ARGV.include?("--preinstall")
@@ -420,9 +422,10 @@ class Reporter
       next unless dst.extname == ".rb"
 
       if paths.any? { |p| tap.cask_file?(p) }
-        # Currently only need to handle Cask deletion/migration.
-        if status == "D"
-          # Have a dedicated report array for deleted casks.
+        case status
+        when "M"
+          @report[:MC] << [tap, src]
+        when "D"
           @report[:DC] << tap.formula_file_to_name(src)
         end
       end
@@ -569,6 +572,24 @@ class Reporter
         tabs.each { |tab| tab.tap = new_tap }
         tabs.each(&:write)
       end
+    end
+  end
+
+  def migrate_modified_casks
+    report[:MC].each do |tap, path|
+      cask = Hbc::CaskLoader.load(tap.path/path)
+      next unless cask.installed?
+
+      current_version = cask.version
+      installed_version = cask.versions.last
+
+      current_caskfile = cask.sourcefile_path
+      installed_caskfile = cask.installed_caskfile
+
+      next unless current_version == installed_version
+      next if current_caskfile.sha256 == installed_caskfile.sha256
+
+      installed_caskfile.atomic_write File.read(current_caskfile)
     end
   end
 


Library/Homebrew/cmd/update-report.rb
```diff
@@ -572,6 +575,24 @@ def migrate_tap_migration
...
+      current_version = cask.version
+      installed_version = cask.versions.last
+
+      current_caskfile = cask.sourcefile_path
+      installed_caskfile = cask.installed_caskfile
+
+      next unless current_version == installed_version
+      next if current_caskfile.sha256 == installed_caskfile.sha256
+
+      installed_caskfile.atomic_write File.read(current_caskfile)
```
MikeMcQuaid: Should handle if the write (or perhaps even read) fails here.

MikeMcQuaid:
What's the overhead on doing this on every `brew update`?

vitorgalvao:
Big thumbs up to this. Not only will it fix issues with uninstalls, it should help prevent problems like the `undent` deprecation (closes https://github.com/Homebrew/homebrew-cask/issues/49716).

MikeMcQuaid:
> it should help prevent problems like the undent deprecation (closes Homebrew/homebrew-cask#49716).

It won't prevent that because if the `version` of a Cask is updated it'll still have the same problem.

vitorgalvao:
> It won't prevent that because if the `version` of a Cask is updated it'll still have the same problem.

On the HBC repos, it’s customary that when we introduce a change like the `undent` deprecation that affects a lot of casks, we make a PR that introduces the change (and nothing else) to all relevant casks at once.

MikeMcQuaid:
> On the HBC repos, it’s customary that when we introduce a change like the undent deprecation that affects a lot of casks, we make a PR that introduces the change (and nothing else) to all relevant casks at once.

Even when you do that: it's still a problem. If you're relying on having ambiguously old Cask files which use Homebrew APIs on them: eventually some of those APIs will be deprecated, disabled and removed and you'll have the same problem again. Assuming that the current version of Cask has infinite backwards compatibility is the root issue here.

reitermarkus:
Yes, I am aware that this won't fix the problem. There really isn't a fix for this, other than never deprecating anything. However, this will prevent deprecation errors for people who run `brew update` regularly.

Even I had one cask still using `.undent`, and I run `brew update` daily.

MikeMcQuaid:
> There really isn't a fix for this, other than never deprecating anything.

I disagree. Several options off the top of my head:

- Catch these exceptions and fall back to the current Cask file's `uninstall`
- Catch these exceptions and print a warning saying uninstall hooks are not run
- Instead of storing the entire Cask file in the Caskroom store just the part that's needed e.g. the uninstall hook method
- Instead of storing the entire Cask file in the Caskroom store instead store something to define uninstall logic that doesn't rely on Homebrew internals e.g. a Bash script or a JSON file that defines logic

commitay:
> Catch these exceptions and fall back to the current Cask file's uninstall

https://github.com/Homebrew/homebrew-cask/issues/49716#issuecomment-406605565

MikeMcQuaid:
@commitay I replied to that comment there. Can you be more explicit on what you're trying to say rather than just dropping a single link to a comment you'd made previously? Would you care to address any of my other options?

To be more explicit on my viewpoint: no-one except me seems to be suggesting solutions to this problem which will not recur next time we remove formerly deprecated then disabled Homebrew APIs. Homebrew formulae have solutions to this problem. I'm not pretending I have all the answers but the current situation of "it'll happen again and there's nothing we can do about it" is defeatist and makes it hard for us to remove any APIs in future.

vitorgalvao:
> Catch these exceptions and fall back to the current Cask file's `uninstall`

A bunch of casks have changes that make this impractical. E.g., software that was a `.app` is now a `.pkg`. Or something that didn’t have an uninstall script, now has one. I don’t have numbers on how often this happens, but it’s not incredibly rare.

> Catch these exceptions and print a warning saying uninstall hooks are not run

Worse experience for the user, but I’m not opposed to it. We’d need to document why this happens and explain how to uninstall manually, but it might work.

> Instead of storing the entire Cask file in the Caskroom store just the part that's needed e.g. the uninstall hook method

Impractical if the deprecation is on a part needed for the uninstall.

> Instead of storing the entire Cask file in the Caskroom store instead store something to define uninstall logic that doesn't rely on Homebrew internals e.g. a Bash script or a JSON file that defines logic

I like this idea.

MikeMcQuaid:
> Worse experience for the user, but I’m not opposed to it. We’d need to document why this happens and explain how to uninstall manually, but it might work.

I think it's a better experience than the current situation of an unhandled but expected error.

> Impractical if the deprecation is on a part needed for the uninstall.

Yep but it still reduces the "surface" somewhat.



> I like this idea.

🎉

reitermarkus:
> - Catch these exceptions and fall back to the current Cask file's `uninstall`
>  - Catch these exceptions and print a warning saying uninstall hooks are not run

Those are not fixes, those are fallbacks.

> - Instead of storing the entire Cask file in the Caskroom store just the part that's needed e.g. the uninstall hook method
> - Instead of storing the entire Cask file in the Caskroom store instead store something to define uninstall logic that doesn't rely on Homebrew internals e.g. a Bash script or a JSON file that defines logic

This sounds good in theory, but I'm not sure if this isn't just shifting the problem elsewhere and additionally making the uninstall less flexible.

MikeMcQuaid:
> Those are not fixes, those are fallbacks.

Indeed but they are an improvement on the current situation of a known, unhandled exception.

> This sounds good in theory, but I'm not sure if this isn't just shifting the problem elsewhere and additionally making the uninstall less flexible.

It will make uninstall less flexible but it's not going to increase the surface area of problems. In the `undent` case: most of these issues seem to be coming from places other than the uninstall block. Additionally, a bash script would isolate entirely from Homebrew Cask's Ruby (or even Bash) logic as Apple aren't going to be breaking Bash's backwards compatibility any time soon.

MikeMcQuaid:
@Homebrew/cask any thoughts or updates on this? Please note as above these DSLs will be removed in Homebrew 2.0 around February regardless of a PR like this being merged.

vitorgalvao:
I’m still fine with the idea behind this.

PR #4752, opened 2018-08-25
Use JSON to marshal errors from children

woodruffw:

-----

Take two on #4717:

This replaces our use of Marshal to send exceptions from child processes to their controlling parents with a JSON representation of the exception state. This JSON representation is then wrapped in a new exception class, ChildProcessError. For particularly interesting exceptions in child processes (namely `BuildError`s and `ErrorDuringExecution`s), we reify the ChildProcessError back into its original exception (which keeps analytics and other parts of the codebase that use it directly happy).

Adds tests for `FormulaInstaller` and `Utils.safe_fork`.

### DIFF
--- a/Library/Homebrew/build.rb
+++ b/Library/Homebrew/build.rb
@@ -190,7 +190,20 @@ begin
   build   = Build.new(formula, options)
   build.install
 rescue Exception => e # rubocop:disable Lint/RescueException
-  Marshal.dump(e, error_pipe)
+  error_hash = JSON.parse e.to_json
+
+  # Special case: need to recreate BuildErrors in full
+  # for proper analytics reporting and error messages.
+  # BuildErrors are specific to build processses and not other
+  # children, which is why we create the necessary state here
+  # and not in Utils.safe_fork.
+  if error_hash["json_class"] == "BuildError"
+    error_hash["cmd"] = e.cmd
+    error_hash["args"] = e.args
+    error_hash["env"] = e.env
+  end
+
+  error_pipe.puts error_hash.to_json
   error_pipe.close
   exit! 1
 end
--- a/Library/Homebrew/dev-cmd/test.rb
+++ b/Library/Homebrew/dev-cmd/test.rb
@@ -93,9 +93,6 @@ module Homebrew
             exec(*args)
           end
         end
-      rescue ::Test::Unit::AssertionFailedError => e
-        ofail "#{f.full_name}: failed"
-        puts e.message
       rescue Exception => e # rubocop:disable Lint/RescueException
         ofail "#{f.full_name}: failed"
         puts e, e.backtrace
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -744,14 +744,18 @@ class FormulaInstaller
       raise "Empty installation"
     end
   rescue Exception => e # rubocop:disable Lint/RescueException
-    e.options = display_options(formula) if e.is_a?(BuildError)
+    if e.is_a? BuildError
+      e.formula = formula
+      e.options = options
+    end
+
     ignore_interrupts do
       # any exceptions must leave us with nothing installed
       formula.update_head_version
       formula.prefix.rmtree if formula.prefix.directory?
       formula.rack.rmdir_if_possible
     end
-    raise
+    raise e
   end
 
   def link(keg)
--- a/Library/Homebrew/test/support/fixtures/failball.rb
+++ b/Library/Homebrew/test/support/fixtures/failball.rb
@@ -1,0 +1,20 @@
+class Failball < Formula
+  def initialize(name = "failball", path = Pathname.new(__FILE__).expand_path, spec = :stable, alias_path: nil)
+    self.class.instance_eval do
+      stable.url "file://#{TEST_FIXTURE_DIR}/tarballs/testball-0.1.tbz"
+      stable.sha256 TESTBALL_SHA256
+    end
+    super
+  end
+
+  def install
+    prefix.install "bin"
+    prefix.install "libexec"
+
+    # This should get marshalled into a BuildError.
+    system "/usr/bin/false" if ENV["FAILBALL_BUILD_ERROR"]
+
+    # This should get marshalled into a RuntimeError.
+    raise "something that isn't a build error happened!"
+  end
+end
--- a/Library/Homebrew/utils/fork.rb
+++ b/Library/Homebrew/utils/fork.rb
@@ -2,6 +2,26 @@ require "fcntl"
 require "socket"
 
 module Utils
+  def self.rewrite_child_error(child_error)
+    error = if child_error.inner_class == ErrorDuringExecution
+      ErrorDuringExecution.new(child_error.inner["cmd"],
+                               status: child_error.inner["status"],
+                               output: child_error.inner["output"])
+    elsif child_error.inner_class == BuildError
+      # We fill `BuildError#formula` and `BuildError#options` in later,
+      # when we rescue this in `FormulaInstaller#build`.
+      BuildError.new(nil, child_error.inner["cmd"],
+                     child_error.inner["args"], child_error.inner["env"])
+    else
+      # Everything other error in the child just becomes a RuntimeError.
+      RuntimeError.new(child_error.message)
+    end
+
+    error.set_backtrace child_error.backtrace
+
+    error
+  end
+
   def self.safe_fork(&_block)
     Dir.mktmpdir("homebrew", HOMEBREW_TEMP) do |tmpdir|
       UNIXServer.open("#{tmpdir}/socket") do |server|
@@ -15,7 +35,18 @@ module Utils
             write.fcntl(Fcntl::F_SETFD, Fcntl::FD_CLOEXEC)
             yield
           rescue Exception => e # rubocop:disable Lint/RescueException
-            Marshal.dump(e, write)
+            error_hash = JSON.parse e.to_json
+
+            # Special case: We need to recreate ErrorDuringExecutions
+            # for proper error messages and because other code expects
+            # to rescue them further down.
+            if e.is_a?(ErrorDuringExecution)
+              error_hash["cmd"] = e.cmd
+              error_hash["status"] = e.status.exitstatus
+              error_hash["output"] = e.output
+            end
+
+            write.puts error_hash.to_json
             write.close
             exit!
           else
@@ -33,10 +64,20 @@ module Utils
             socket.close
           end
           write.close
-          data = read.read
+          # Each line on the error pipe contains a JSON-serialized exception.
+          # We read the first, since only one is necessary for a failure.
+          data = read.gets
           read.close
           Process.wait(pid) unless socket.nil?
-          raise Marshal.load(data) unless data.nil? || data.empty? # rubocop:disable Security/MarshalLoad
+
+          if data && !data.empty?
+            error_hash = JSON.parse(data) unless data.nil? || data.empty?
+
+            e = ChildProcessError.new(error_hash)
+
+            raise rewrite_child_error(e)
+          end
+
           raise Interrupt if $CHILD_STATUS.exitstatus == 130
           raise "Forked child process failed: #{$CHILD_STATUS}" unless $CHILD_STATUS.success?
         end
--- a/Library/Homebrew/exceptions.rb
+++ b/Library/Homebrew/exceptions.rb
@@ -352,14 +352,16 @@ class FormulaAmbiguousPythonError < RuntimeError
 end
 
 class BuildError < RuntimeError
-  attr_reader :formula, :env
-  attr_accessor :options
+  attr_reader :cmd, :args, :env
+  attr_accessor :formula, :options
 
   def initialize(formula, cmd, args, env)
     @formula = formula
+    @cmd = cmd
+    @args = args
     @env = env
-    args = args.map { |arg| arg.to_s.gsub " ", "\\ " }.join(" ")
-    super "Failed executing: #{cmd} #{args}"
+    pretty_args = args.map { |arg| arg.to_s.gsub " ", "\\ " }.join(" ")
+    super "Failed executing: #{cmd} #{pretty_args}"
   end
 
   def issues
@@ -526,12 +528,22 @@ end
 
 # raised by safe_system in utils.rb
 class ErrorDuringExecution < RuntimeError
+  attr_reader :cmd
   attr_reader :status
+  attr_reader :output
 
   def initialize(cmd, status:, output: nil)
+    @cmd = cmd
     @status = status
+    @output = output
 
-    s = "Failure while executing; `#{cmd.shelljoin.gsub(/\\=/, "=")}` exited with #{status.exitstatus}."
+    exitstatus = if status.respond_to?(:exitstatus)
+      status.exitstatus
+    else
+      status
+    end
+
+    s = "Failure while executing; `#{cmd.shelljoin.gsub(/\\=/, "=")}` exited with #{exitstatus}."
 
     unless [*output].empty?
       format_output_line = lambda do |type, line|
@@ -594,5 +606,24 @@ class BottleFormulaUnavailableError < RuntimeError
         #{bottle_path}
         #{formula_path}
     EOS
+  end
+end
+
+# Raised when a child process sends us an exception over its error pipe.
+class ChildProcessError < RuntimeError
+  attr_reader :inner
+  attr_reader :inner_class
+
+  def initialize(inner)
+    @inner = inner
+    @inner_class = Object.const_get inner["json_class"]
+
+    super <<~EOS
+      An exception occured within a build process:
+        #{inner_class}: #{inner["m"]}
+    EOS
+
+    # Clobber our real (but irrelevant) backtrace with that of the inner exception.
+    set_backtrace inner["b"]
   end
 end
--- a/Library/Homebrew/global.rb
+++ b/Library/Homebrew/global.rb
@@ -1,5 +1,7 @@
 require "pathname"
 require "English"
+require "json"
+require "json/add/core"
 
 HOMEBREW_LIBRARY_PATH = Pathname.new(__FILE__).realpath.parent
 
--- a/Library/Homebrew/postinstall.rb
+++ b/Library/Homebrew/postinstall.rb
@@ -15,7 +15,7 @@ begin
   formula.extend(Debrew::Formula) if ARGV.debug?
   formula.run_post_install
 rescue Exception => e # rubocop:disable Lint/RescueException
-  Marshal.dump(e, error_pipe)
+  error_pipe.puts e.to_json
   error_pipe.close
   exit! 1
 end
--- a/Library/Homebrew/test.rb
+++ b/Library/Homebrew/test.rb
@@ -28,7 +28,7 @@ begin
     raise "test returned false" if formula.run_test == false
   end
 rescue Exception => e # rubocop:disable Lint/RescueException
-  Marshal.dump(e, error_pipe)
+  error_pipe.puts e.to_json
   error_pipe.close
   exit! 1
 end
--- a/Library/Homebrew/test/formula_installer_spec.rb
+++ b/Library/Homebrew/test/formula_installer_spec.rb
@@ -4,6 +4,7 @@ require "keg"
 require "tab"
 require "test/support/fixtures/testball"
 require "test/support/fixtures/testball_bottle"
+require "test/support/fixtures/failball"
 
 describe FormulaInstaller do
   define_negated_matcher :need_bottle, :be_bottle_unneeded
@@ -28,7 +29,7 @@ describe FormulaInstaller do
       Tab.clear_cache
       expect(Tab.for_keg(keg)).not_to be_poured_from_bottle
 
-      yield formula
+      yield formula if block_given?
     ensure
       Tab.clear_cache
       keg.unlink
@@ -131,5 +132,22 @@ describe FormulaInstaller do
     expect {
       fi.check_install_sanity
     }.to raise_error(CannotInstallFormulaError)
+  end
+
+  specify "install fails with BuildError when a system() call fails" do
+    ENV["HOMEBREW_TEST_NO_EXIT_CLEANUP"] = "1"
+    ENV["FAILBALL_BUILD_ERROR"] = "1"
+
+    expect {
+      temporary_install(Failball.new)
+    }.to raise_error(BuildError)
+  end
+
+  specify "install fails with a RuntimeError when #install raises" do
+    ENV["HOMEBREW_TEST_NO_EXIT_CLEANUP"] = "1"
+
+    expect {
+      temporary_install(Failball.new)
+    }.to raise_error(RuntimeError)
   end
 end
--- a/Library/Homebrew/test/support/lib/config.rb
+++ b/Library/Homebrew/test/support/lib/config.rb
@@ -8,7 +8,11 @@ HOMEBREW_BREW_FILE = Pathname.new(ENV["HOMEBREW_BREW_FILE"])
 
 TEST_TMPDIR = ENV.fetch("HOMEBREW_TEST_TMPDIR") do |k|
   dir = Dir.mktmpdir("homebrew-tests-", ENV["HOMEBREW_TEMP"] || "/tmp")
-  at_exit { FileUtils.remove_entry(dir) }
+  at_exit do
+    # Child processes inherit this at_exit handler, but we don't want them
+    # to clean TEST_TMPDIR up prematurely (i.e., when they exit early for a test).
+    FileUtils.remove_entry(dir) unless ENV["HOMEBREW_TEST_NO_EXIT_CLEANUP"]
+  end
   ENV[k] = dir
 end
 
--- a/Library/Homebrew/test/utils/fork_spec.rb
+++ b/Library/Homebrew/test/utils/fork_spec.rb
@@ -1,0 +1,21 @@
+require "utils/fork"
+
+describe Utils do
+  describe "#safe_fork" do
+    it "raises a RuntimeError on an error that isn't ErrorDuringExecution" do
+      expect {
+        described_class.safe_fork do
+          raise "this is an exception in the child"
+        end
+      }.to raise_error(RuntimeError)
+    end
+
+    it "raises an ErrorDuringExecution on one in the child" do
+      expect {
+        described_class.safe_fork do
+          safe_system "/usr/bin/false"
+        end
+      }.to raise_error(ErrorDuringExecution)
+    end
+  end
+end


Library/Homebrew/build.rb
```diff
@@ -190,7 +190,20 @@ def fixopt(f)
   build   = Build.new(formula, options)
   build.install
 rescue Exception => e # rubocop:disable Lint/RescueException
-  Marshal.dump(e, error_pipe)
+  error_hash = JSON.parse e.to_json
```
MikeMcQuaid: Why convert this to JSON and then to a hash?

Library/Homebrew/dev-cmd/test.rb
```diff
@@ -93,9 +93,14 @@ def test
...
         end
-      rescue ::Test::Unit::AssertionFailedError => e
+      rescue ChildProcessError => e
         ofail "#{f.full_name}: failed"
-        puts e.message
+        case e.inner["json_class"]
+        when "Test::Unit::AssertionFailedError"
+          puts e.inner["m"]
+        else
+          puts e.inner["json_class"], e.backtrace
```
MikeMcQuaid: What happens to the exception message?

Library/Homebrew/test/support/fixtures/failball.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+    super
+  end
+
+  def install
+    prefix.install "bin"
+    prefix.install "libexec"
+
+    # This should get marshalled into a BuildError.
+    system "this_command_must_never_execute" if ENV["FAILBALL_BUILD_ERROR"]
```
MikeMcQuaid: Can you explain this line a bit?

Library/Homebrew/utils/fork.rb
```diff
@@ -33,10 +44,27 @@ def self.safe_fork(&_block)
...
           write.close
-          data = read.read
+          # Each line on the error pipe contains a JSON-serialized exception.
+          # We read the first, since only one is necessary for a failure.
+          data = read.gets
           read.close
           Process.wait(pid) unless socket.nil?
-          raise Marshal.load(data) unless data.nil? || data.empty? # rubocop:disable Security/MarshalLoad
+
+          error_hash = JSON.parse(data) unless data.nil? || data.empty?
```
MikeMcQuaid: Can you refactor this into an `if`

Library/Homebrew/build.rb
```diff
@@ -190,7 +190,20 @@ def fixopt(f)
...
+
+  # Special case: need to recreate BuildErrors in full
+  # for proper analytics reporting and error messages.
+  # BuildErrors are specific to build processses and not other
+  # children, which is why we create the necessary state here
+  # and not in Utils.safe_fork.
+  if e.is_a?(BuildError)
```
MikeMcQuaid: Is there any way to avoid all the `is_a?` Maybe looking at the keys instead?

Library/Homebrew/utils/fork.rb
```diff
@@ -33,10 +44,27 @@ def self.safe_fork(&_block)
...
+
+          if error_hash
+            e = case error_hash["json_class"]
+            when "ErrorDuringExecution"
```
MikeMcQuaid: Any way of turning this into an actual class comparison?

Library/Homebrew/build.rb
woodruffw: So that we can handle adding extra information in the `BuildError` case below -- `Exception` doesn't have a `#to_h` or a `#to_hash` method, so I figured it was slightly cleaner to go from JSON to a hash than to duplicate write a monkeypatched method or `Utils` method that does the same.

Library/Homebrew/build.rb
woodruffw: Yeah, `error_hash["json_class"] == "BuildError"` should do the trick. I'll change it.

Library/Homebrew/dev-cmd/test.rb
woodruffw: `e.inner["m"]` contains the message, so it still gets printed in the `Test::Unit:AssertionFailedError` case.

Library/Homebrew/test/support/fixtures/failball.rb
woodruffw: Yup!

We're calling `system` with a bogus command to induce a failure, which causes it to raise a `BuildError` in the build process, per https://github.com/Homebrew/brew/blob/9bc17c8a82a36d9676528f761eaec3961f240cc8/Library/Homebrew/formula.rb#L1878

That `BuildError` contains a bunch of state that we use later on in analytics (and to tell the user what exactly went wrong), so we make a special case when catching that exception to re-add most of that state to the JSON blob that goes over the error pipe. The parent process reads that blob, realizes that it's a special case, and reconstructs a `BuildError` instead of a more generic `ChildProcessError`.

We do all of that only when `ENV["FAILBALL_BUILD_ERROR"]` is set in the test harness, so that we can fall through to the more generic test case below (when something other than a command fails in the build process, and a generic `ChildProcessError` gets reported back).

Library/Homebrew/utils/fork.rb
woodruffw: Probably not at this point in the program -- we're comparing with strings because we're about to (re)create the classes that are coming from the error pipe in JSON form.

woodruffw:
> brew tests is failing.

Yeah looks like there's something funny going on with the mocks for `Process::Status`:

```
  4) ErrorDuringExecution#to_s when only given a command and a status to_s should eq "Failure while executing; `false` exited with 1."
     Failure/Error: its(:to_s) { is_expected.to eq "Failure while executing; `false` exited with 1." }
       expected: "Failure while executing; `false` exited with 1."
            got: "Failure while executing; `false` exited with #[InstanceDouble(Process::Status) (anonymous)]."
```

I'll investigate.

Library/Homebrew/dev-cmd/test.rb
```diff
@@ -93,9 +93,14 @@ def test
             exec(*args)
           end
         end
-      rescue ::Test::Unit::AssertionFailedError => e
+      rescue ChildProcessError => e
         ofail "#{f.full_name}: failed"
-        puts e.message
+        case e.inner["json_class"]
```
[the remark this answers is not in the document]
woodruffw: I could add `ChildProcessError.inner_class` that just calls `const_get` on `inner["json_class"]`, resulting in something like this:

```ruby

if e.inner_class == ::Test::Unit::AssertionFailedError
  puts e.inner["m"]
else
  puts e.inner_class, e.backtrace
end
```

How does that sound?

Library/Homebrew/dev-cmd/test.rb
woodruffw: I duplicated the logic from below (the `rescue Exception => e`), since it's basically the same failure mode (non-assertion failures, it's just that we're distinguishing between non-assertion failures in the child and parent now).

I can make it print the message explicitly if you'd like, in which case I think it makes sense to update the original fallthrough too.

Library/Homebrew/formula_installer.rb
```diff
@@ -763,14 +763,23 @@ def build
       raise "Empty installation"
     end
   rescue Exception => e # rubocop:disable Lint/RescueException
-    e.options = display_options(formula) if e.is_a?(BuildError)
+    # Utils.safe_fork rebuilds ErrorDuringExecutions for us, but doesn't have access
+    # to the Formula object that we need to rebuild BuildErrors.
+    if e.is_a?(ChildProcessError) && e.inner["json_class"] == "BuildError"
```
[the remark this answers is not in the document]
woodruffw: Yeah, I could make this do a `respond_to? :inner_class` if we're good on the `inner_class` solution above.

Library/Homebrew/dev-cmd/test.rb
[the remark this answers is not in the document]
MikeMcQuaid: Yeh, I think that'd be nicer 👍. I do wonder if this couldn't be done at a higher level somewhere (`Utils.safe_fork`?) so that it doesn't need to be handled as JSON in so many places.

Library/Homebrew/dev-cmd/test.rb
MikeMcQuaid: Like above I think it might be nice to fiddle with handling `ChildProcessError` elsewhere so this special handling isn't required at all here.

Library/Homebrew/formula_installer.rb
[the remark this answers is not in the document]
MikeMcQuaid: Actually: like above I think it might be nice to fiddle with handling `ChildProcessError` elsewhere so this special handling isn't required at all here.

Library/Homebrew/test/support/fixtures/failball.rb
MikeMcQuaid: Can this use `/usr/bin/false` instead?

Library/Homebrew/utils/fork.rb
```diff
@@ -33,10 +44,27 @@ def self.safe_fork(&_block)
...
+          if data && !data.empty?
+            error_hash = JSON.parse(data) unless data.nil? || data.empty?
+
+            e = case error_hash["json_class"]
+            when "ErrorDuringExecution"
+              ErrorDuringExecution.new(error_hash["cmd"],
+                                       status: error_hash["status"],
+                                       output: error_hash["output"])
+            else
+              ChildProcessError.new(error_hash)
```
MikeMcQuaid: Could this be caught and handled somewhere elsewhere so the fact it's come from a child process doesn't need to be handled in each and every case above?

woodruffw: Where else were you thinking? I'm not sure where else to put it, since `fork.rb` is where the parent and child processes converge.

woodruffw: Hmm, I could do something like this:

```ruby
e = ChildProcessError.new(error_hash)

raise e unless e.inner_class == ErrorDuringExecution

# Rebuild and throw the ErrorDuringExecution here
```

That flattens the logic a bit and gets rid of the ugly string comparison above, but it also creates an extra `ChildProcessError` that we never throw.

Library/Homebrew/utils/fork.rb
```diff
@@ -33,10 +44,27 @@ def self.safe_fork(&_block)
...
+          data = read.gets
           read.close
           Process.wait(pid) unless socket.nil?
-          raise Marshal.load(data) unless data.nil? || data.empty? # rubocop:disable Security/MarshalLoad
+
+          if data && !data.empty?
+            error_hash = JSON.parse(data) unless data.nil? || data.empty?
+
+            e = case error_hash["json_class"]
+            when "ErrorDuringExecution"
```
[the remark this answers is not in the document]
woodruffw: I could probably do the same trick as mentioned in https://github.com/Homebrew/brew/pull/4752#discussion_r213154037, although I'm not sure how to ducktype it per se.

Library/Homebrew/utils/fork.rb
MikeMcQuaid: @woodruffw Yeh, something like that would be good, I think. It'd be good to recreate the original exception and throw it as "early" as possible so that things that are "later"/higher up the call stack can avoid needing to know or care about `ChildProcessError` being a thing. It's possible I've misunderstood here, though, so please feel free to argue and disagree 😆

Library/Homebrew/utils/fork.rb
```diff
@@ -15,7 +15,18 @@ def self.safe_fork(&_block)
             write.fcntl(Fcntl::F_SETFD, Fcntl::FD_CLOEXEC)
             yield
           rescue Exception => e # rubocop:disable Lint/RescueException
-            Marshal.dump(e, write)
+            error_hash = JSON.parse e.to_json
+
+            # Special case: We need to recreate ErrorDuringExecutions
+            # for proper error messages and because other code expects
+            # to rescue them further down.
+            if e.is_a?(ErrorDuringExecution)
```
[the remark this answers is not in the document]
woodruffw: I'm a little stuck on the ducktyping here -- I don't think we want to test the exception with `responds_to?` here, since there are probably other exceptions that have `cmd`/`status`/`output` attributes that we don't want to handle specially. I flattened the exception recreation below, though.

MikeMcQuaid:
@woodruffw I'm doing a bad job explaining but I think what would be nice would be a function-level handling of `ChildProcessError` in `safe_fork` which converted it into a "normal" exception so none of the callers need to care/change. Is that possible or am I missing something here?

woodruffw:
> I'm doing a bad job explaining but I think what would be nice would be a function-level handling of ChildProcessError in safe_fork which converted it into a "normal" exception so none of the callers need to care/change. Is that possible or am I missing something here?

If you mean a single method that converts `ChildProcessError`s into `BuildError`s and `ErrorDuringExecution`s (the two special cases): yeah, it's possible. The reason I didn't do it is because I thought it muddled the distinction between the individual callers/children a bit -- only `build.rb` can raise `BuildError`s, for example, but sharing a single exception reconstruction method between `build.rb`, `test.rb`, and `postinstall.rb` makes it look like all three have the same exception behavior when they don't.

There is also at least one case where we can't rebuild the exception directly inside of `safe_fork`: `BuildError`s need to be instantiated with the `Formula` that failed, but that isn't in scope until we `rescue` back in `formula_installer.rb`. The `Test::Unit::AssertionFailedError`, since it's a third-party library's class and takes a mystery `options` parameter that we don't have access to: https://www.rubydoc.info/github/test-unit/test-unit/Test/Unit/AssertionFailedError#initialize-instance_method

In sum: I could probably do something close to what you want, but I don't think it'll remove many of the caller changes I've had to make: we're still going to need to recreate the `BuildError` outside of `safe_fork`, and `ChildProcessError`s are going to leak *somewhere* eventually (since all `Exception`s in the children are made into them). It's the small price we're paying for not deserializing arbitrary classes :slightly_smiling_face:.

MikeMcQuaid:
> If you mean a single method that converts ChildProcessErrors into BuildErrors and ErrorDuringExecutions (the two special cases): yeah, it's possible. 

Cool. I hate to be a pain but: could we see that? Feel free to make it a bit hacky because if it's awful I may decide the current way is better?

> The reason I didn't do it is because I thought it muddled the distinction between the individual callers/children a bit -- only build.rb can raise BuildErrors, for example, but sharing a single exception reconstruction method between build.rb, test.rb, and postinstall.rb makes it look like all three have the same exception behavior when they don't.

My take is that the extracting of the "real" exception from `ChildProcessError` feels like an implementation detail that should be handled by the methods that actually create a child process; their callers shouldn't need to care about the implementation details of the methods they call.

> There is also at least one case where we can't rebuild the exception directly inside of safe_fork: BuildErrors need to be instantiated with the Formula that failed, but that isn't in scope until we rescue back in formula_installer.rb.

Could the formula object that failed be passed through instead? Alternatively, could a stub value be used here that's populated later?

> ChildProcessErrors are going to leak somewhere eventually (since all Exceptions in the children are made into them). It's the small price we're paying for not deserializing arbitrary classes 🙂.

Yeh, I just think they should have a layer containing them as far into the stack as possible and they don't leak beyond that point (even if that means refactoring some existing code so everywhere they could leak from goes through a single method/class).

woodruffw:
@MikeMcQuaid https://github.com/Homebrew/brew/pull/4752/commits/98a43effade057c1b24310d5384abb76e467b344 has a version of your idea; I'm breaking the pros and cons out below:

Pros:
* Conceptually cleaner -- `ChildProcessError`s no longer occur any further up the stack than `Utils.safe_fork`, so individual callers don't need to handle them specially.
* DRY -- all exception rewriting is now done in `Utils.rewrite_child_error`, which will make adding new special cases (if we ever have to) into a single-file affair. Having a single method also allows us to rewrite the exception backtraces uniformly.

Cons:
* We lose access to `ChildProcessError#inner_class`, which we used previously to handle the `AssertionFailedError` case in `test.rb`. This isn't a huge deal since the more general (RuntimeError) case will still produce an acceptable error message and backtrace, but it makes the distinction between a test case failure and a logic error in the test code itself a little less clear.
* We have to add `BuildError#formula=`, as the `Formula` that caused the `BuildError` isn't in scope at the time of creation within `rewrite_child_error`. This gets handled the same as `BuildError#options=`, and is just a little bit more state that adds to the conceptual load of correctly creating a `BuildError`.
* We lose any distinction between errors that happen in child processes and those that happen in the parent. This is essentially just a rollback to the way things were when this code used `Marshal` so it's also not a huge deal, although I think being able to distinguish between parent and child errors could aid in debugging.

Ultimately, I writing this has changed my mind: I think the pros *do* outweigh the cons here, especially given that my original approach of handling `ChildProcessError`s manually in each parent was pretty ugly and non-DRY.

Library/Homebrew/utils/fork.rb
```diff
@@ -2,6 +2,26 @@
 require "socket"
 
 module Utils
+  def self.rewrite_child_error(child_error)
+    error = if child_error.inner_class == ErrorDuringExecution
```
MikeMcQuaid: Maybe a `case` here?

MikeMcQuaid:
> Ultimately, I writing this has changed my mind: I think the pros do outweigh the cons here, especially given that my original approach of handling ChildProcessErrors manually in each parent was pretty ugly and non-DRY.

I agree 👍 

> We lose access to ChildProcessError#inner_class, which we used previously to handle the AssertionFailedError case in test.rb.

Would it be possible to create the new exception object based on that `inner_class`? No worries if not but I think if that's done there's no doubt this is a better approach anyway.

Library/Homebrew/utils/fork.rb
woodruffw: `case` uses `#===` for comparison, so comparing a class directly doesn't work :disappointed::

```ruby
case String
when String then :foo
else :bar
end # => :bar
```

woodruffw:
> Would it be possible to create the new exception object based on that inner_class? No worries if not but I think if that's done there's no doubt this is a better approach anyway.

Probably not, since that would require arbitrary exception class instantiation (and that's what we're trying to eliminate by removing `Marshal`). I could add `AssertionFailedError` as another explicit special case, but I figured that it was safer to not try instantiating a 3rd-party exception and wouldn't add much to the error reporting anyways (since we place the `ChildProcessError`'s message inside the `RuntimeError` anyways).

woodruffw:
CI's failing, although it looks unrelated to these changes:

```
  1) brew style Homebrew::check_style_json corrected offense output format
     Failure/Error: offense_string = rubocop_result.file_offenses(formula.realpath).first.to_s
     NoMethodError:
       undefined method `upcase' for nil:NilClass
     # ./style.rb:140:in `severity_code'
     # ./style.rb:156:in `to_s'
     # ./test/cmd/style_spec.rb:54:in `block (3 levels) in <top (required)>'
     # ./test/cmd/style_spec.rb:9:in `block (2 levels) in <top (required)>'
     # ./test/spec_helper.rb:121:in `block (2 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:123:in `block in run'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `loop'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `run'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:37:in `block (2 levels) in setup'
```

MikeMcQuaid:
> CI's failing, although it looks unrelated to these changes:

`master` is not failing. Try to rebase and/or rerun the build. If it still fails: it may be related to these changes 😉

woodruffw:
Rebase did the trick :tada:

MikeMcQuaid:
I'm happy with this now. Any other tweaks you want before merging this @woodruffw? If not: 🚢!

woodruffw:
Yeah, I'm happy with it too. I'll squash down into a few manageable/feature commits and merge in a bit.

Thanks for the review!

commitay:
This seems to have broken analytics.

```console
$ brew formula-analytics go
Error: badRequest: Invalid JSON payload received. Unknown name "json_class" at 'report_requests[0].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[0].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "json_class" at 'report_requests[0].sampling_level': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[0].sampling_level': Cannot find field.
Invalid JSON payload received. Unknown name "json_class" at 'report_requests[1].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[1].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "json_class" at 'report_requests[1].sampling_level': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[1].sampling_level': Cannot find field.
Invalid JSON payload received. Unknown name "json_class" at 'report_requests[2].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[2].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "json_class" at 'report_requests[2].sampling_level': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[2].sampling_level': Cannot find field.
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/http_command.rb:218:in `check_status'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/api_command.rb:116:in `check_status'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/http_command.rb:183:in `process_response'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/batch.rb:87:in `block in decode_response_body'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/batch.rb:81:in `each_index'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/batch.rb:81:in `decode_response_body'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/http_command.rb:184:in `process_response'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/http_command.rb:299:in `execute_once'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/http_command.rb:104:in `block (2 levels) in execute'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:61:in `block in retriable'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `times'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `retriable'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/http_command.rb:101:in `block in execute'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:61:in `block in retriable'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `times'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `retriable'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/http_command.rb:93:in `execute'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.6/lib/google/apis/core/base_service.rb:178:in `batch'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:212:in `block in <top (required)>'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:209:in `each'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:209:in `each_slice'
/usr/local/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:209:in `<top (required)>'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/usr/local/Homebrew/Library/Homebrew/utils.rb:19:in `require?'
/usr/local/Homebrew/Library/Homebrew/brew.rb:95:in `<main>'
```

woodruffw:
Ugh, thought I had fixed that. I'm guessing (since I don't know much about GA or our analytics code) that the issue here is some new keys (`s` and `json_class`) getting added to the schema.

lembacon:
The core's travis test is also failing due to the breakage of analytics.

https://travis-ci.org/Homebrew/homebrew-core/builds/424184064

```
brew formula-analytics --days-ago=30 --json --build-error > _data/analytics/build-error/30d.json
Error: badRequest: Invalid JSON payload received. Unknown name "json_class" at 'report_requests[0].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[0].dimension_filter_clauses[0].filters[0].expressions[0]': Cannot find field.
Invalid JSON payload received. Unknown name "json_class" at 'report_requests[0].sampling_level': Cannot find field.
Invalid JSON payload received. Unknown name "s" at 'report_requests[0].sampling_level': Cannot find field.
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/http_command.rb:218:in `check_status'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/api_command.rb:116:in `check_status'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/http_command.rb:183:in `process_response'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/batch.rb:87:in `block in decode_response_body'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/batch.rb:81:in `each_index'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/batch.rb:81:in `decode_response_body'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/http_command.rb:184:in `process_response'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/http_command.rb:299:in `execute_once'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/http_command.rb:104:in `block (2 levels) in execute'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:61:in `block in retriable'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `times'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `retriable'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/http_command.rb:101:in `block in execute'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:61:in `block in retriable'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `times'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/retriable-3.1.2/lib/retriable.rb:56:in `retriable'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/http_command.rb:93:in `execute'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/vendor/ruby/ruby/2.3.0/gems/google-api-client-0.23.8/lib/google/apis/core/base_service.rb:178:in `batch'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:212:in `block in <top (required)>'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:209:in `each'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:209:in `each_slice'
/home/travis/Homebrew/Library/Taps/homebrew/homebrew-formula-analytics/cmd/brew-formula-analytics.rb:209:in `<top (required)>'
/home/travis/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.7/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/home/travis/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.7/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
/home/travis/Homebrew/Library/Homebrew/utils.rb:19:in `require?'
/home/travis/Homebrew/Library/Homebrew/brew.rb:95:in `<main>'
```

PR #4765, opened 2018-08-27
audit: Use `to_i`during assignment to prevent NPE

GauthamGoli:

-----
https://github.com/Homebrew/homebrew-core/pull/31514#issuecomment-416292746

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -130,7 +130,7 @@ module Homebrew
       formula_count += 1
       problem_count += fa.problems.size
       problem_lines = format_problem_lines(fa.problems)
-      corrected_problem_count = options[:style_offenses]&.count(&:corrected?)
+      corrected_problem_count = options.fetch(:style_offenses, []).count(&:corrected?)
       new_formula_problem_lines = format_problem_lines(fa.new_formula_problems)
       if args.display_filename?
         puts problem_lines.map { |s| "#{f.path}: #{s}" }


Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -130,7 +130,7 @@ def audit
       formula_count += 1
       problem_count += fa.problems.size
       problem_lines = format_problem_lines(fa.problems)
-      corrected_problem_count = options[:style_offenses]&.count(&:corrected?)
+      corrected_problem_count ||= options[:style_offenses]&.count(&:corrected?)
```
MikeMcQuaid: Maybe this should always be set to an integer of some sort?

GauthamGoli: https://github.com/Homebrew/brew/pull/4765/files#diff-bd0ba063e0f8aeb1f4a3997d34914c0cR77

It is being initialized as 0. Without this conditional assignment, its value is being changed to `nil`

MikeMcQuaid: `0` isn't falsey so this means it'll never be set in this case. Instead use a different variable which can be the `count` or `nil` and use `.to_i` which will convert `nil` to `0.

GauthamGoli: Fixed.

Library/Homebrew/dev-cmd/audit.rb
```diff
@@ -130,7 +130,7 @@ def audit
       formula_count += 1
       problem_count += fa.problems.size
       problem_lines = format_problem_lines(fa.problems)
-      corrected_problem_count = options[:style_offenses]&.count(&:corrected?)
+      corrected_problem_count = options[:style_offenses]&.count(&:corrected?).to_i
```
reitermarkus: How about `options.fetch(:style_offenses, []).count(&:corrected?)`?

MikeMcQuaid: I like that, yeh 👍

GauthamGoli: Updated. Reads much better, thanks!

MikeMcQuaid:
Nice work @GauthamGoli :shipit:

MikeMcQuaid:
> 0% of diff hit (target 71.15%)

Doesn't need to be this PR but this might be a good thing to address for future.

PR #4767, opened 2018-08-27
Upgrade: implement linkage repair

### LINKED ISSUE #4761: Reinstall formulae with broken linkage on `brew upgrade`
We `revision` bump things that have mandatory dependencies to other formulae that would otherwise be broken on a `brew upgrade`. Unfortunately this does not cover optionally or opportunistically linked dependencies which will be broken by a `brew upgrade` and require a manual `brew reinstall`. 

Thankfully we now have a fast cache for `brew linkage` so it's possible to quickly query on every upgrade whether there's any broken linkage and reinstall applications accordingly.

amyspark:

-----

After upgrade, verify that existing kegs are still linked properly. If not, reinstall them.

Fixes #4761.

### DIFF
--- a/Library/Homebrew/cmd/reinstall.rb
+++ b/Library/Homebrew/cmd/reinstall.rb
@@ -7,6 +7,7 @@
 require "formula_installer"
 require "development_tools"
 require "messages"
+require "reinstall"
 
 module Homebrew
   module_function
@@ -25,60 +26,5 @@ module Homebrew
       reinstall_formula(f)
     end
     Homebrew.messages.display_messages
-  end
-
-  def reinstall_formula(f)
-    if f.opt_prefix.directory?
-      keg = Keg.new(f.opt_prefix.resolved_path)
-      keg_had_linked_opt = true
-      keg_was_linked = keg.linked?
-      backup keg
-    end
-
-    build_options = BuildOptions.new(Options.create(ARGV.flags_only), f.options)
-    options = build_options.used_options
-    options |= f.build.used_options
-    options &= f.options
-
-    fi = FormulaInstaller.new(f)
-    fi.options              = options
-    fi.invalid_option_names = build_options.invalid_option_names
-    fi.build_bottle         = ARGV.build_bottle? || (!f.bottled? && f.build.bottle?)
-    fi.interactive          = ARGV.interactive?
-    fi.git                  = ARGV.git?
-    fi.link_keg           ||= keg_was_linked if keg_had_linked_opt
-    fi.prelude
-
-    oh1 "Reinstalling #{Formatter.identifier(f.full_name)} #{options.to_a.join " "}"
-
-    fi.install
-    fi.finish
-  rescue FormulaInstallationAlreadyAttemptedError
-    nil
-  rescue Exception # rubocop:disable Lint/RescueException
-    ignore_interrupts { restore_backup(keg, keg_was_linked) }
-    raise
-  else
-    backup_path(keg).rmtree if backup_path(keg).exist?
-  end
-
-  def backup(keg)
-    keg.unlink
-    keg.rename backup_path(keg)
-  end
-
-  def restore_backup(keg, keg_was_linked)
-    path = backup_path(keg)
-
-    return unless path.directory?
-
-    Pathname.new(keg).rmtree if keg.exist?
-
-    path.rename keg
-    keg.link if keg_was_linked
-  end
-
-  def backup_path(path)
-    Pathname.new "#{path}.reinstall"
   end
 end
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -21,6 +21,7 @@
 #:    are pinned; see `pin`, `unpin`).
 
 require "install"
+require "reinstall"
 require "formula_installer"
 require "cleanup"
 require "development_tools"
@@ -80,6 +81,16 @@ module Homebrew
       puts formulae_upgrades.join(", ")
     end
 
+    upgrade_formulae(formulae_to_install)
+
+    check_dependents(formulae_to_install)
+
+    Homebrew.messages.display_messages
+  end
+
+  def upgrade_formulae(formulae_to_install)
+    return if formulae_to_install.empty?
+
     # Sort keg_only before non-keg_only formulae to avoid any needless conflicts
     # with outdated, non-keg_only versions of formulae being upgraded.
     formulae_to_install.sort! do |a, b|
@@ -104,7 +115,6 @@ module Homebrew
         onoe "#{f}: #{e}"
       end
     end
-    Homebrew.messages.display_messages
   end
 
   def upgrade_formula(f)
@@ -169,6 +179,191 @@ module Homebrew
       linked_kegs.each(&:link) unless f.installed?
     rescue
       nil
+    end
+  end
+
+  def upgradable_dependents(kegs, formulae)
+    formulae_to_upgrade = Set.new
+    formulae_pinned = Set.new
+
+    formulae.each do |formula|
+      descendants = Set.new
+
+      dependents = kegs.select do |keg|
+        keg.runtime_dependencies
+           .any? { |d| d["full_name"] == formula.full_name }
+      end
+
+      next if dependents.empty?
+
+      dependent_formulae = dependents.map(&:to_formula)
+
+      dependent_formulae.each do |f|
+        next if formulae_to_upgrade.include?(f)
+        next if formulae_pinned.include?(f)
+
+        if f.outdated?(fetch_head: ARGV.fetch_head?)
+          if f.pinned?
+            formulae_pinned << f
+          else
+            formulae_to_upgrade << f
+          end
+        end
+
+        descendants << f
+      end
+
+      upgradable_descendants, pinned_descendants = upgradable_dependents(kegs, descendants)
+
+      formulae_to_upgrade.merge upgradable_descendants
+      formulae_pinned.merge pinned_descendants
+    end
+
+    [formulae_to_upgrade, formulae_pinned]
+  end
+
+  def broken_dependents(kegs, formulae)
+    formulae_to_reinstall = Set.new
+    formulae_pinned_and_outdated = Set.new
+
+    CacheStoreDatabase.use(:linkage) do |db|
+      formulae.each do |formula|
+        descendants = Set.new
+
+        dependents = kegs.select do |keg|
+          keg.runtime_dependencies
+             .any? { |d| d["full_name"] == formula.full_name }
+        end
+
+        next if dependents.empty?
+
+        dependents.each do |keg|
+          f = keg.to_formula
+
+          next if formulae_to_reinstall.include?(f)
+          next if formulae_pinned_and_outdated.include?(f)
+
+          checker = LinkageChecker.new(keg, cache_db: db)
+
+          if checker.broken_library_linkage?
+            if f.outdated?(fetch_head: ARGV.fetch_head?)
+              # Outdated formulae = pinned formulae (see function above)
+              formulae_pinned_and_outdated << f
+            else
+              formulae_to_reinstall << f
+            end
+          end
+
+          descendants << f
+        end
+
+        descendants_to_reinstall, descendants_pinned = broken_dependents(kegs, descendants)
+
+        formulae_to_reinstall.merge descendants_to_reinstall
+        formulae_pinned_and_outdated.merge descendants_pinned
+      end
+    end
+
+    [formulae_to_reinstall, formulae_pinned_and_outdated]
+  end
+
+  # @private
+  def depends_on(a, b)
+    if a.opt_or_installed_prefix_keg
+        .runtime_dependencies
+        .any? { |d| d["full_name"] == b.full_name }
+      1
+    else
+      a <=> b
+    end
+  end
+
+  # @private
+  def formulae_with_runtime_dependencies
+    Formula.installed
+           .map(&:opt_or_installed_prefix_keg)
+           .reject(&:nil?)
+           .reject { |f| f.runtime_dependencies.to_a.empty? }
+  end
+
+  def check_dependents(formulae)
+    return if formulae.empty?
+
+    # First find all the outdated dependents.
+    kegs = formulae_with_runtime_dependencies
+
+    return if kegs.empty?
+
+    oh1 "Checking dependents for outdated formulae" if ARGV.verbose?
+    upgradable, pinned = upgradable_dependents(kegs, formulae).map(&:to_a)
+
+    upgradable.sort! { |a, b| depends_on(a, b) }
+
+    pinned.sort! { |a, b| depends_on(a, b) }
+
+    # Print the pinned dependents.
+    unless pinned.empty?
+      ohai "Not upgrading #{Formatter.pluralize(pinned.length, "pinned dependent")}:"
+      puts pinned.map { |f| "#{f.full_specified_name} #{f.pkg_version}" } * ", "
+    end
+
+    # Print the upgradable dependents.
+    if upgradable.empty?
+      ohai "No dependents to upgrade" if ARGV.verbose?
+    else
+      ohai "Upgrading #{Formatter.pluralize(upgradable.length, "dependent")}:"
+      formulae_upgrades = upgradable.map do |f|
+        if f.optlinked?
+          "#{f.full_specified_name} #{Keg.new(f.opt_prefix).version} -> #{f.pkg_version}"
+        else
+          "#{f.full_specified_name} #{f.pkg_version}"
+        end
+      end
+      puts formulae_upgrades.join(", ")
+    end
+
+    upgrade_formulae(upgradable)
+
+    # Assess the dependents tree again.
+    kegs = formulae_with_runtime_dependencies
+
+    oh1 "Checking dependents for broken library links" if ARGV.verbose?
+    reinstallable, pinned = broken_dependents(kegs, formulae).map(&:to_a)
+
+    reinstallable.sort! { |a, b| depends_on(a, b) }
+
+    pinned.sort! { |a, b| depends_on(a, b) }
+
+    # Print the pinned dependents.
+    unless pinned.empty?
+      onoe "Not reinstalling #{Formatter.pluralize(pinned.length, "broken and outdated, but pinned dependent")}:"
+      $stderr.puts pinned.map { |f| "#{f.full_specified_name} #{f.pkg_version}" } * ", "
+    end
+
+    # Print the broken dependents.
+    if reinstallable.empty?
+      ohai "No broken dependents to reinstall" if ARGV.verbose?
+    else
+      ohai "Reinstalling #{Formatter.pluralize(reinstallable.length, "broken dependent")} from source:"
+      puts reinstallable.map(&:full_specified_name).join(", ")
+    end
+
+    reinstallable.each do |f|
+      begin
+        reinstall_formula(f, build_from_source: true)
+      rescue FormulaInstallationAlreadyAttemptedError
+        # We already attempted to reinstall f as part of the dependency tree of
+        # another formula. In that case, don't generate an error, just move on.
+        nil
+      rescue CannotInstallFormulaError => e
+        ofail e
+      rescue BuildError => e
+        e.dump
+        puts
+        Homebrew.failed = true
+      rescue DownloadError => e
+        ofail e
+      end
     end
   end
 end
--- a/Library/Homebrew/reinstall.rb
+++ b/Library/Homebrew/reinstall.rb
@@ -1,0 +1,63 @@
+require "formula_installer"
+require "development_tools"
+require "messages"
+
+module Homebrew
+  module_function
+
+  def reinstall_formula(f, build_from_source: false)
+    if f.opt_prefix.directory?
+      keg = Keg.new(f.opt_prefix.resolved_path)
+      keg_had_linked_opt = true
+      keg_was_linked = keg.linked?
+      backup keg
+    end
+
+    build_options = BuildOptions.new(Options.create(ARGV.flags_only), f.options)
+    options = build_options.used_options
+    options |= f.build.used_options
+    options &= f.options
+
+    fi = FormulaInstaller.new(f)
+    fi.options              = options
+    fi.invalid_option_names = build_options.invalid_option_names
+    fi.build_bottle         = ARGV.build_bottle? || (!f.bottled? && f.build.bottle?)
+    fi.interactive          = ARGV.interactive?
+    fi.git                  = ARGV.git?
+    fi.link_keg           ||= keg_was_linked if keg_had_linked_opt
+    fi.build_from_source    = true if build_from_source
+    fi.prelude
+
+    oh1 "Reinstalling #{Formatter.identifier(f.full_name)} #{options.to_a.join " "}"
+
+    fi.install
+    fi.finish
+  rescue FormulaInstallationAlreadyAttemptedError
+    nil
+  rescue Exception # rubocop:disable Lint/RescueException
+    ignore_interrupts { restore_backup(keg, keg_was_linked) }
+    raise
+  else
+    backup_path(keg).rmtree if backup_path(keg).exist?
+  end
+
+  def backup(keg)
+    keg.unlink
+    keg.rename backup_path(keg)
+  end
+
+  def restore_backup(keg, keg_was_linked)
+    path = backup_path(keg)
+
+    return unless path.directory?
+
+    Pathname.new(keg).rmtree if keg.exist?
+
+    path.rename keg
+    keg.link if keg_was_linked
+  end
+
+  def backup_path(path)
+    Pathname.new "#{path}.reinstall"
+  end
+end
--- a/Library/Homebrew/test/cmd/upgrade_spec.rb
+++ b/Library/Homebrew/test/cmd/upgrade_spec.rb
@@ -7,4 +7,14 @@ describe "brew upgrade", :integration_test do
 
     expect(HOMEBREW_CELLAR/"testball/0.1").to be_a_directory
   end
+
+  it "upgrades a Formula and cleans up old versions" do
+    setup_test_formula "testball"
+    (HOMEBREW_CELLAR/"testball/0.0.1/foo").mkpath
+
+    expect { brew "upgrade", "--cleanup" }.to be_a_success
+
+    expect(HOMEBREW_CELLAR/"testball/0.1").to be_a_directory
+    expect(HOMEBREW_CELLAR/"testball/0.0.1").not_to exist
+  end
 end


Library/Homebrew/cmd/upgrade.rb
```diff
@@ -104,6 +104,32 @@ def upgrade
...
+
+      CacheStoreDatabase.use(:linkage) do |db|
+        kegs = Formula.installed.collect(&:opt_or_installed_prefix_keg).reject(&:nil?)
+
+        odebug "Assessing keg linkage status" if kegs.size > 1
+
+        kegs.each do |keg|
+          result = LinkageChecker.new(keg, cache_db: db)
+
+          broken_formulae << resolve_formula(keg) if result.broken_library_linkage?
```
scpeters: I get the following error when testing this:

~~~
Error: undefined method `resolve_formula' for Homebrew:Module
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:120:in `block (2 levels) in upgrade'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:117:in `each'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:117:in `block in upgrade'
/usr/local/Homebrew/Library/Homebrew/cache_store.rb:16:in `use'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:112:in `upgrade'
/usr/local/Homebrew/Library/Homebrew/brew.rb:87:in `<main>'
~~~

I believe `linkage_checker.rb` has this function, but `cmd/upgrade.rb` does not

amyspark: It should be fixed in the next commit. Thanks for testing!

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -104,6 +105,35 @@ def upgrade
...
+      CacheStoreDatabase.use(:linkage) do |db|
+        kegs = Formula.installed.collect(&:opt_or_installed_prefix_keg).reject(&:nil?)
+
+        odebug "Assessing keg linkage status" if kegs.size > 1
+
+        kegs.each do |keg|
+          checker = LinkageChecker.new(keg, cache_db: db)
+          next unless checker.broken_library_linkage?
+
+          odebug "#{formula} has broken dynamic library links"
```
scpeters:
~~~
Error: undefined local variable or method `formula' for Homebrew:Module
Did you mean?  format
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:122:in `block (2 levels) in upgrade'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:118:in `each'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:118:in `block in upgrade'
/usr/local/Homebrew/Library/Homebrew/cache_store.rb:16:in `use'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:113:in `upgrade'
/usr/local/Homebrew/Library/Homebrew/brew.rb:87:in `<main>'
~~~

scpeters: the following lets it continue:

~~~
diff --git a/Library/Homebrew/cmd/upgrade.rb b/Library/Homebrew/cmd/upgrade.rb
index 9256e6e..dbc05b2 100644
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -119,9 +119,9 @@ module Homebrew
           checker = LinkageChecker.new(keg, cache_db: db)
           next unless checker.broken_library_linkage?
 
-          odebug "#{formula} has broken dynamic library links"
+          odebug "#{keg} has broken dynamic library links"
 
-          broken_formulae << resolve_formula(keg)
+          broken_formulae << Formulary.from_keg(keg)
         end
       end
 
~~~

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -104,6 +105,35 @@ def upgrade
...
+          broken_formulae << resolve_formula(keg)
+        end
+      end
+
+      return if broken_formulae.empty?
+
+      opoo "Reinstalling #{Formatter.pluralize(broken_formulae.length, "keg")} with broken dynamic library links"
+
+      broken_formulae.each do |f|
+        reinstall_formula(f)
```
scpeters: do we want to consider a reinstall with `--build-from-source`? if it is just going to pour a broken bottle again, then it won't fix anything

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -104,6 +105,35 @@ def upgrade
         onoe "#{f}: #{e}"
       end
     end
+
+    unless Homebrew.failed?
```
MikeMcQuaid: `return if Homebrew.failed?`

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -104,6 +105,35 @@ def upgrade
...
+      #  Assess existing kegs and reinstall any broken.
+      broken_formulae = []
+
+      CacheStoreDatabase.use(:linkage) do |db|
+        kegs = Formula.installed.collect(&:opt_or_installed_prefix_keg).reject(&:nil?)
```
MikeMcQuaid: Stick each `.` on a new line.

Library/Homebrew/cmd/upgrade.rb
MikeMcQuaid: Agreed to building from source here 👍

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -104,6 +105,35 @@ def upgrade
...
+
+        odebug "Assessing keg linkage status" if kegs.size > 1
+
+        kegs.each do |keg|
```
MikeMcQuaid: Could consider slimming this down to only those formulae with `runtime_dependencies` on this formula (which indicates linkage already and is much faster than reading linkage on all kegs). Thoughts?

Library/Homebrew/cmd/upgrade.rb
MikeMcQuaid: This should use `fi = FormulaInstaller.new(f)` rather than including `cmd/reinstall` here.

Library/Homebrew/cmd/upgrade.rb
amyspark: We cannot return yet, as we have not printed the caveats of the upgraded casks yet.

amyspark:
In general @MikeMcQuaid -- how does one use `FormulaInstaller` for rebuilding from source? Should I duplicate the logic from `cmd/reinstall` and just toggle `build_from_source`?

MikeMcQuaid:
> In general @MikeMcQuaid -- how does one use FormulaInstaller for rebuilding from source? Should I duplicate the logic from cmd/reinstall and just toggle build_from_source?

@amyspark Yeh, pretty much, thanks!

Library/Homebrew/cmd/upgrade.rb
MikeMcQuaid: @amyspark sorry, misread, you're right!

Library/Homebrew/cmd/upgrade.rb
amyspark:
```
Formula.installed
  .collect(&:opt_or_installed_prefix_keg)
  .keep_if {...}
```
Two options for checking here:
- `|k| k.runtime_dependencies.contains?(f)`;
- `|k| k.runtime_dependencies.map(&:name).include?(f.full_alias_name)`.

I went for the first one, but let me know which is correct!

amyspark:
Done @MikeMcQuaid, thanks for your comments! Also, Codecov thinks this may need some tests 👇 -- does anyone know where to get or break test formulae for this purpose?

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -92,18 +93,46 @@ def upgrade
       end
     end
 
+    broken_formulae = []
+
     formulae_to_install.each do |f|
       Migrator.migrate_if_needed(f)
       begin
         upgrade_formula(f)
         next if !ARGV.include?("--cleanup") && !ENV["HOMEBREW_UPGRADE_CLEANUP"]
```
scpeters: this cleanup logic is preventing the new linkage-checking code from being run in my test (since I haven't been passing --cleanup)

we aren't going to require `--cleanup` for the linkage checking are we?

MikeMcQuaid: > we aren't going to require --cleanup for the linkage checking are we?

Yeh, we shouldn't 👍

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +200,60 @@ def upgrade_formula(f)
       nil
     end
   end
+
+  def reinstall_formula_from_source(f)
```
MikeMcQuaid: Eek, there's rather a bit more here than I expected. I wonder how best to slim this down.

amyspark: Thought you'd say that 🤣 
I'd say including `cmd/reinstall` and add a parameter to `reinstall_formula` would be a better choice. Factoring out the reinstall logic into a new module like `LinkageChecker`-`cmd/linkage` and calling it from there would also work.

Library/Homebrew/cmd/upgrade.rb
amyspark: Refactored this part into a new function.

Library/Homebrew/cmd/upgrade.rb
MikeMcQuaid: > I'd say including cmd/reinstall and add a parameter to reinstall_formula would be a better choice. Factoring out the reinstall logic into a new module like LinkageChecker-cmd/linkage and calling it from there would also work.

@amyspark Yeh, I think so; having a new `Reinstall` class which both `cmd/reinstall` and `cmd/upgrade` can call would be ideal given `build_from_source` seems to be the only required new argument there.

MikeMcQuaid:
> Done @MikeMcQuaid, thanks for your comments! Also, Codecov thinks this may need some tests 👇 -- does anyone know where to get or break test formulae for this purpose?

@amyspark Check the existing tests; you can use `formula do` to specify new formulae for such things.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +174,93 @@ def upgrade_formula(f)
...
   end
+
+  def relink_dependents(formula)
+    broken_formulae = []
+
+    CacheStoreDatabase.use(:linkage) do |db|
+      kegs = Formula.installed
+                    .collect(&:opt_or_installed_prefix_keg)
+                    .reject { |f| f.runtime_dependencies.nil? }
+                    .keep_if { |f| f.runtime_dependencies.include?(formula) }
```
scpeters: thanks; it's getting closer. the following makes it work for me in my testing:

~~~
diff --git a/Library/Homebrew/cmd/upgrade.rb b/Library/Homebrew/cmd/upgrade.rb
index 94a99c0..1f86258 100644
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -182,7 +182,7 @@ module Homebrew
       kegs = Formula.installed
                     .collect(&:opt_or_installed_prefix_keg)
                     .reject { |f| f.runtime_dependencies.nil? }
-                    .keep_if { |f| f.runtime_dependencies.include?(formula) }
+                    .keep_if { |f| f.runtime_dependencies.collect{ |d| d["full_name"] == formula.full_name }.any? }
 
       odebug "Checking dynamic library links of #{formula}'s dependents" unless kegs.empty?
 
@@ -192,7 +192,7 @@ module Homebrew
 
         odebug "#{keg} has broken dynamic library links and will be reinstalled from source"
 
-        broken_formulae << Formulary.from_keg(f)
+        broken_formulae << Formulary.from_keg(keg)
       end
     end
 
~~~

Library/Homebrew/cmd/upgrade.rb
amyspark: I've just sent in the changes, please check if it's OK for you.

Library/Homebrew/reinstall.rb
```diff
@@ -0,0 +1,63 @@
...
+    options &= f.options
+
+    fi = FormulaInstaller.new(f)
+    fi.options              = options
+    fi.invalid_option_names = build_options.invalid_option_names
+    fi.build_bottle         = ARGV.build_bottle? || (!f.bottled? && f.build.bottle?)
+    fi.interactive          = ARGV.interactive?
+    fi.git                  = ARGV.git?
+    fi.link_keg           ||= keg_was_linked if keg_had_linked_opt
+    fi.build_from_source    = true if build_from_source
```
scpeters: could the right-hand side just be `build_from_source`?

amyspark: No.`FormulaInstaller` internally sets `build_from_source` from `ARGV`, e.g. as in `brew reinstall <keg> --build-from-source`. If the relinkage detects broken kegs, this line is triggered to force the reinstall from source; otherwise, the parameter is left alone.

scpeters: sounds good; thanks for the explanation

amyspark:
@MikeMcQuaid:
- Messaging was polished. `odebug` wasn't used anywhere in the Homebrew side, so I copied how `upgrade` lists the formulae.
- The reinstallation logic was refactored into `reinstall.rb`, which `upgrade` and `reinstall` now require.
- Build from source is only forced for poured kegs (`k.tab.poured_from_bottle`). Forcing this on kegs that were already built from source causes to unnecessarily download and build **the whole dependency tree** from source.

Library/Homebrew/cmd/reinstall.rb
```diff
@@ -7,7 +7,7 @@
 require "formula_installer"
 require "development_tools"
 require "messages"
-
+require "reinstall"
```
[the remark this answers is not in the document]
amyspark: Yes, will do.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +174,41 @@ def upgrade_formula(f)
...
+      kegs.each do |keg|
+        checker = LinkageChecker.new(keg, cache_db: db)
+        next unless checker.broken_library_linkage?
+        broken_kegs << keg
+      end
+    end
+
+    return if broken_kegs.empty?
+
+    ohai "#{Formatter.pluralize(broken_kegs.length, "keg")} with broken dynamic library links will be reinstalled from source:"
```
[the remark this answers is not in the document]
amyspark: Done, please tell me if it's OK for you.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +174,43 @@ def upgrade_formula(f)
...
+    broken_kegs.each do |k|
+      f = Formulary.from_keg(k)
+      ohai "#{formula}'s dependent #{f} has broken dynamic library links!"
+
+      # Only *force* rebuild the formula if it was poured.
+      reinstall_formula(f, build_from_source: !k.tab.poured_from_bottle)
+
+      next if !ARGV.include?("--cleanup") && !ENV["HOMEBREW_UPGRADE_CLEANUP"]
+      next unless f.installed?
+      Cleanup.new.cleanup_formula(f)
```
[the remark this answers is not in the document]
amyspark: I'll remove this part to be consistent with `brew reinstall`'s behaviour.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +174,43 @@ def upgrade_formula(f)
...
+      end
+    end
+
+    return if broken_kegs.empty?
+
+    broken_kegs.each do |k|
+      f = Formulary.from_keg(k)
+      ohai "#{formula}'s dependent #{f} has broken dynamic library links!"
+
+      # Only *force* rebuild the formula if it was poured.
```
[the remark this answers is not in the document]
amyspark: Sorry, I misinterpreted my tests' results. I used `knot` and `p11-kit` to test this part and I got extra dependencies installed from source (`Sphinx-doc`). I checked now, and the existing behaviour is OK. Will remove this part.

amyspark:
Ty @MikeMcQuaid! Shall I squash these too?

MikeMcQuaid:
@amyspark Up to you. I'd still like to wait and see what @reitermarkus thinks first but if he has no comments by e.g. Monday we could merge anyway.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +174,35 @@ def upgrade_formula(f)
...
+         .any? { |d| d["full_name"] == formula.full_name }
+      end
+
+      break if dependents.empty?
+
+      ohai "Checking dependents of #{formula} for broken dynamic library links" if ARGV.verbose?
+
+      dependents.each do |keg|
+        checker = LinkageChecker.new(keg, cache_db: db)
+        broken_formulae << Formulary.from_keg(k) if checker.broken_library_linkage?
```
scpeters: one more quick fix (`k` -> `keg`)

~~~
diff --git a/Library/Homebrew/cmd/upgrade.rb b/Library/Homebrew/cmd/upgrade.rb
index d967e09..057ae9c 100644
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -194,7 +194,7 @@ module Homebrew
 
       dependents.each do |keg|
         checker = LinkageChecker.new(keg, cache_db: db)
-        broken_formulae << Formulary.from_keg(k) if checker.broken_library_linkage?
+        broken_formulae << Formulary.from_keg(keg) if checker.broken_library_linkage?
       end
     end
 
~~~

DomT4:
I have some mild concerns here around this squashing our ability to do things piecemeal. I merged in a `hdf5` update the other day, for example, because it had a fairly weighty number of security fixes & everything except [one of the](https://github.com/Homebrew/homebrew-core/issues/31500) `brew uses` was building okay. I essentially temporarily sacrificed the one formula for the "greater good".

Presumably, unless I'm misunderstanding _(which is possible based on the amount of time I've had to look this over properly, sorry)_ if I do that after this everyone's going to get that one forced reinstallation from source locally and a good chunk of those will end up reporting it to us? Essentially my choice would become sit on security fixes for days or weeks after they've been publicly disclosed, or face being whomped by people filing issues for a known problem; is that the case?

scpeters:
I was initially testing this PR by checking out an old version of both `homebrew/core` and my `osrf/simulation` and then install the old protobuf 3.5.1 and something that uses protobuf like `ignition-msgs1`. Then I checkout the HEAD of core and `brew upgrade protobuf`. This pours the protobuf 3.6.0 bottle and then detects a linkage problem with ignition-msgs1 and rebuilds ignition-msgs1 from source. In this case, that works fine because the latest bottle of `ignition-msgs1` had already been installed and it was bad, so a rebuild from source fixed things.

I think the logic in this PR as of cf9d84e isn't ideal for other scenarios though. For example, when boost 1.68.0 eventually lands ( https://github.com/Homebrew/homebrew-core/pull/30914 ), it will introduce new bottles for `boost` and its dependents, such as `boost-bcp`. With this PR, running `brew upgrade boost boost-bcp` would pour boost from a bottle, then detect a linkage problem with `boost-bcp` while still processing `boost` and then build `boost-bcp` from source even though it has a good bottle available.

MikeMcQuaid:
@DomT4 This will only try to `brew reinstall` currently installed, newest version kegs with broken linkage on upgrade. I see this as improving rather than detracting from the situation you mention because it means those people won't have their stuff broken, need to notice and reinstall manually. I can't see why anyone will be filing issues for that as it is intended behaviour and will result in a `brew upgrade` or `brew upgrade foo` that doesn't end up randomly breaking all packages with opportunistic or optional linkage. This is a change of behaviour and may result in `brew upgrade` taking longer in some cases but, like when `brew install` fails to pour a bottle so tries to build from source, I see this as introducing a more correct behaviour. Please feel free to disagree though; I may have misunderstood your use-case.

> With this PR, running brew upgrade boost boost-bcp would pour boost from a bottle, then detect a linkage problem with boost-bcp while still processing boost and then build boost-bcp from source even though it has a good bottle available.

@scpeters Very good spot/catch, here. That's indeed a problem that should be addressed before this is merged but shouldn't be too hard; we've already calculated the `formulae_to_install` so we could either check the one we're reinstalling isn't in that list or (my preference) only check the linkage after all the installs have been completed (so a separate loop through `formulae_to_install` after the upgrades have been completed).

amyspark:
@scpeters -- good catch! Fix incoming.

DomT4:
> I see this as improving rather than detracting from the situation you mention because it means those people won't have their stuff broken, need to notice and reinstall manually.

Well, my point was more that sometimes we intentionally break stuff for the greater good. `shogun` currently has broken linkage, for example, but that can't be resolved until we verify & fix the checksum mismatch, which users would also see if they tried to `brew reinstall shogun`. That's the sort of situation I was vaguely trying to ask about.

MikeMcQuaid:
> Well, my point was more that sometimes we intentionally break stuff for the greater good. shogun currently has broken linkage, for example, but that can't be resolved until we verify & fix the checksum mismatch, which users would also see if they tried to brew reinstall shogun. That's the sort of situation I was vaguely trying to ask about.

Hmm, yeh, I'm not sure how best to deal with this. I don't think it's worth blocking the feature on but I would like to think about something to deal with that. Even with `shogun` I think there's other options to consider other than breaking it; can we upload a mirror from the cache or something?

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +177,35 @@ def upgrade_formula(f)
...
+        checker = LinkageChecker.new(keg, cache_db: db)
+        broken_formulae << Formulary.from_keg(keg) if checker.broken_library_linkage?
+      end
+    end
+
+    return if broken_formulae.empty?
+
+    broken_formulae.each do |f|
+      ohai "#{formula}'s dependent #{f} has broken dynamic library links!"
+      reinstall_formula(f, build_from_source: true)
```
MikeMcQuaid: What happens if one of these fail to install?

amyspark: Just as in `reinstall.rb`, the error bubbles up the call stack. Should I do something like this (extracted from `upgrade`)?
```ruby
    broken_formulae.each do |f|
      ohai "#{formula}'s dependent #{f} has broken dynamic library links!"
      reinstall_formula(f, build_from_source: true)
    rescue FormulaInstallationAlreadyAttemptedError
      # We already attempted to reinstall f as part of the dependency tree of
      # another formula. In that case, don't generate an error, just move on.
      nil
    rescue CannotInstallFormulaError => e
      ofail e
    rescue BuildError => e
      e.dump
      puts
      Homebrew.failed = true
    rescue DownloadError => e
      ofail e
    end
```

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +177,35 @@ def upgrade_formula(f)
...
+
+      dependents.each do |keg|
+        checker = LinkageChecker.new(keg, cache_db: db)
+        broken_formulae << Formulary.from_keg(keg) if checker.broken_library_linkage?
+      end
+    end
+
+    return if broken_formulae.empty?
+
+    broken_formulae.each do |f|
```
amyspark: Should I deduplicate formulae here (e.g. in case one has multiple broken links)?

DomT4:
> I don't think it's worth blocking the feature on but I would like to think about something to deal with that

Yeah, agreed. I'm a firm 👍 on the overall feature; just slightly concerned about either a) loss of flexibility in core or b) a hundred people filing or commenting on issues to tell me/us something I/we already know to be the case.

> can we upload a mirror from the cache or something?

`shogun` is actually a pretty weird case. The thing that broke was a patch URL, using `?full_index=1`. I haven't talked to upstream yet because there's a lot going on with brew at the moment & it all needs a lot of attention heh, but the only thing that I can think of that could _possibly_ cause that would be if the repo was force pushed over, but I'm not 100% on that even.

I don't think it has been through CI recently enough for the prior known-good patch to have been retained in the cache.

MikeMcQuaid:
> a hundred people filing or commenting on issues to tell me/us something I/we already know to be the case.

In this case it's only on 318 installs in 365 days so I think it's pretty safe. If loads of people comment on this stuff I think it's just a sign we need to prioritise it or remove the formula entirely. Ultimately this is only automating what people will do when they notice `shogun` is broken anyway.

> the only thing that I can think of that could possibly cause that would be if the repo was force pushed over, but I'm not 100% on that even.

Nope, the SHA-1 would have changed then. It's definitely a change on GitHub's end. Honestly I'd just confirm the upstream commit still applies, update the checksum and move on. This isn't in upstream's control (it's effectively an undocumented/unsupported GitHub API) and a malicious actor couldn't change it even if they wanted to. It could be reported to GitHub at most but I'm pretty sure you'll be told "this isn't a supported API".

Library/Homebrew/cmd/upgrade.rb
MikeMcQuaid: @amyspark Yeh, I think it's worth doing 👍

Library/Homebrew/cmd/upgrade.rb
MikeMcQuaid: Yeh, I think something like that would be ideal 👍. We want to ensure this will continue to work through formulae that fail to install. Additionally it might be nice to have a message printing the list of all formulae that we'll attempt to reinstall before we do so 👍

MikeMcQuaid:
@amyspark I've just thought of another edge case that may want addressed if it's not already: the broken linkage may be fixed by an `upgrade` so we should check that the formula isn't outdated before we do a `reinstall` (because then you could do an `upgrade` instead).

Sorry there's so much back and forth here; this is a pretty complicated feature but it will be one that people will be extremely grateful for!

amyspark:
@MikeMcQuaid, it's no problem! I'm glad to be of help.

In the next commit, all your comments should've been addressed:
- Linkage assessment is now performed after all upgrades are done.
- For the sake of efficiency, the totality of kegs are fetched only once per run (instead of once per formula).
- The same error handling for `upgrade` is applied for the reinstall process.
- We now print a message for `broken_dependents` like we do for `formulae_to_install`.

MikeMcQuaid:
I'm now happy with this, nice work @amyspark! When @scpeters has given this a final look/test: let's 🚢!

scpeters:
> @MikeMcQuaid said: @amyspark I've just thought of another edge case that may want addressed if it's not already: the broken linkage may be fixed by an upgrade so we should check that the formula isn't outdated before we do a reinstall (because then you could do an upgrade instead).

I don't see where this comment has been addressed. I don't see any place where we check for outdated formulae before trying to reinstall. I'm not sure if this is necessary for this PR or can wait for a follow-up.

MikeMcQuaid:
> I don't see where this comment has been addressed. I don't see any place where we check for outdated formulae before trying to reinstall. I'm not sure if this is necessary for this PR or can wait for a follow-up.

@scpeters you're right, thanks!

amyspark:
@MikeMcQuaid: if I understood this correctly, the only case where we would end up with outdated *and* broken formulae would be in `brew upgrade X Y Z...`. How should I check for outdated dependents, especially if e.g. a broken dependent is also pinned?

MikeMcQuaid:
@amyspark It could be I `brew upgrade libyaml` which breaks `ruby` but `ruby` has had a `revision` so `brew upgrade ruby` would fix it. In that case a `brew reinstall` will attempt to install the newer version so is undesirable so we should `brew upgrade` instead. If I've `brew pin`d `ruby` we should print out an error message stating it could not be upgraded because of the pin.

scpeters:
I think that makes sense to me. I think point 2.3.1 is the only missing piece.

Thanks for the great work you've been doing here @amyspark !

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -80,6 +81,16 @@ def upgrade
...
     end
 
+    start_upgrading(formulae_to_install)
+
+    relink_dependents(formulae_to_install)
+
+    Homebrew.messages.display_messages
+  end
+
+  def start_upgrading(formulae_to_install)
```
MikeMcQuaid: `upgrade_formulae` maybe?

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,100 @@ def upgrade_formula(f)
...
+                "#{f.full_specified_name} #{f.pkg_version}"
+              end
+            end
+            puts formulae_upgrades.join(", ")
+            formulae_to_upgrade << upgradable
+          end
+
+          next if from_source.empty?
+
+          oh1 "Reinstalling #{Formatter.pluralize(from_source.length, "dependent")} from source:"
```
MikeMcQuaid: Should this be maybe further down?

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,100 @@ def upgrade_formula(f)
...
+
+    kegs = Formula.installed
+                  .map(&:opt_or_installed_prefix_keg)
+                  .reject { |f| f.runtime_dependencies.to_a.empty? }
+
+    formulae_to_assess = formulae
+    wave = 0
+
+    until formulae_to_assess.empty?
+      wave += 1
```
MikeMcQuaid: I'm not sure I understand why the `until`/`wave` is necessary, can you explain?

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,100 @@ def upgrade_formula(f)
...
+
+          upgradable, pinned = upgrade_possible.partition(&:pinned?)
+
+          if !pinned.empty? && !ARGV.include?("--ignore-pinned")
+            oh1 "Skipping #{Formatter.pluralize(pinned.length, "pinned dependent")}:"
+            puts pinned.map { |f| "#{f.full_specified_name} #{f.pkg_version}" } * ", "
+          end
+
+          unless upgradable.empty?
+            oh1 "Upgrading #{Formatter.pluralize(upgradable.length, "dependent")}:"
```
MikeMcQuaid: I reckon this should be printed just before the upgrade happens

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,100 @@ def upgrade_formula(f)
...
+          dependents.each do |keg|
+            checker = LinkageChecker.new(keg, cache_db: db)
+            broken_dependents << Formulary.from_keg(keg) if checker.broken_library_linkage?
+          end
+
+          next if broken_dependents.empty?
+
+          ohai "Dependents of #{formula} have broken dynamic library links!"
+
+          upgrade_possible, from_source = broken_dependents.partition do |f|
```
MikeMcQuaid: `upgrade_needed, reinstall_needed`?

Library/Homebrew/cmd/upgrade.rb
amyspark: Please see the explanation below.

Library/Homebrew/cmd/upgrade.rb
amyspark: Please see the explanation above.

Library/Homebrew/cmd/upgrade.rb
amyspark: The initial version performs the assessment in a depth-first fashion (libyaml->ruby->...). Imagine libyaml broke a dependent X, which can be upgraded, but ruby broke X too; it will be upgraded twice in that version (once for ruby and another for libyaml).

In this version, I check all formulae in that dependency level *in a single pass*, reinstall/upgrade them, and then loop again to check their dependents, until this results in no broken formulae. This is why I print all messages at once instead of in each loop, to give a complete context of what will be done in this step.

Finally, the wave stuff serves to mark which depth level we're in regarding the `brew upgrade`d formulae.

MikeMcQuaid: Gotcha, I see. Whenever I see an `until`, though, I'm worried about the potential for an infinite loop. I think given you're storing the `wave` here you should `break` when `wave` hits a certain level?

MikeMcQuaid: @amyspark I think the lowest level you think would still be useful. From your description that could be `2` or `3`? What do you think?



> Should we also add a --just-fix-them-all flag in case someone still wants to jump down this rabbit hole?

Let's pass on that unless we get a request for it.

amyspark: In my Mac the deepest level is 5, starting with `poppler`. I'll use that.

MikeMcQuaid: @amyspark Can you explain; is that the number of levels of recursive dependencies?

amyspark: Yes. Suppose I upgrade `poppler` and it breaks a dependent, that's how deep I can go in the dependency tree before I run out of formulae to keep assessing and reinstalling/upgrading.

amyspark: Of course, this is assuming the very worst case!

MikeMcQuaid: @amyspark Given this is going down/up a tree structure I think that a recursive solution might be more natural. What do you think?

amyspark: It is more natural, you're correct. However:
- we'll have to fetch all installed kegs *in each recursion*
- it's not easy to return the list of upgraded/reinstalled formulae
- if it were (perhaps a `[upgrade, reinstalled].flatten`), we have to deduplicate formulae so as to act only on the remaining ones.

MikeMcQuaid: @amyspark I was thinking about this a bit more and I think the problem is putting `upgrade` and `reinstall` in the same loop. What's I'd propose:

- find all the outdated dependents (if `brew upgrade` is specified without arguments) or all the named formulae
- for each of those dependents, look at the formulae that depend directly on them and are outdated
- for each of those second-level formulae (dependents): recursively look at the formulae that depend on them, finding those that are outdated etc.
- when all formulae in the tree have been found (but not yet upgraded): upgrade them all
- when all formulae have been upgraded (or any upgrades failed): check the linkage on all formulae and reinstall those necessary

How does that sound?

amyspark: That sounds... so... sweeeet 😭 

This only leaves the messaging stuff. When a dependent is detected to be broken, how would you like it to be reported? Is the current version fine?

MikeMcQuaid: @amyspark Current version is fine, yeh 👍 .



> when brew upgrade X Y Z, even if they are not X, Y or Z?

@amyspark Yes, I think so. I realise that's a behaviour change but it's one that avoids unnecessary complexity in the code while still respecting pins.

amyspark:
@MikeMcQuaid I've updated this with your comments in re recursive traversal. Please let me know if it's OK.
BTW, do I need to retrieve again the list of kegs after the dependents' upgrade?

scpeters:
I just tested this after being absent from this discussion for a few days, and it doesn't seem to be detecting the broken linkage after the upgrade in my latest testing. Investigating...

scpeters:
the following helps with the recursion logic in my testing:

~~~
diff --git a/Library/Homebrew/cmd/upgrade.rb b/Library/Homebrew/cmd/upgrade.rb
index 53a300283..84844f9ab 100644
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -239,6 +239,7 @@ module Homebrew
           checker = LinkageChecker.new(keg, cache_db: db)
           next unless checker.broken_library_linkage?
 
+          formulae_to_reinstall[keg] = f
           dependents_to_reinstall << f
         end
 
~~~

scpeters:
If there are unlinked kegs, the command crashes for me:

~~~
Error: undefined method `runtime_dependencies' for nil:NilClass
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:260:in `block in check_dependents'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:260:in `reject'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:260:in `check_dependents'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:86:in `upgrade'
/usr/local/Homebrew/Library/Homebrew/brew.rb:89:in `<main>'
~~~

I think the following will fix it:

~~~
diff --git a/Library/Homebrew/cmd/upgrade.rb b/Library/Homebrew/cmd/upgrade.rb
index 53a300283..84844f9ab 100644
--- a/Library/Homebrew/cmd/upgrade.rb
+++ b/Library/Homebrew/cmd/upgrade.rb
@@ -256,6 +257,7 @@ module Homebrew
 
     kegs = Formula.installed
                   .map(&:opt_or_installed_prefix_keg)
+                  .reject(&:nil?)
                   .reject { |f| f.runtime_dependencies.to_a.empty? }
 
     return if kegs.empty?
@@ -291,6 +293,7 @@ module Homebrew
 
     kegs = Formula.installed
                   .map(&:opt_or_installed_prefix_keg)
+                  .reject(&:nil?)
                   .reject { |f| f.runtime_dependencies.to_a.empty? }
 
     oh1 "Checking dependents for broken library links" if ARGV.verbose?
~~~

scpeters:
I suggested a patch above for fixing the logic for detecting broken linkage, but the logic for detecting outdated kegs isn't working for me either. I'll take another look at that tomorrow

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,145 @@ def upgrade_formula(f)
...
+        next if formulae_pinned[f.full_name]
+
+        if f.pinned?
+          formulae_pinned[f.full_name] = f
+        else
+          dependents_to_upgrade << f
+        end
+      end
+
+      u, p = upgradable_dependents(kegs, dependents_to_upgrade)
```
MikeMcQuaid: Can you use longer variable names for these?

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,145 @@ def upgrade_formula(f)
...
+      dependents_to_upgrade = []
+
+      dependents = kegs.select do |k|
+        k.runtime_dependencies
+         .any? { |d| d["full_name"] == formula.full_name }
+      end
+
+      next if dependents.empty?
+
+      dependents.each do |k|
```
MikeMcQuaid: `keg` would be nicer

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,145 @@ def upgrade_formula(f)
...
   end
+
+  def upgradable_dependents(kegs, formulae)
+    formulae_to_upgrade = {}
+    formulae_pinned = {}
+
+    formulae.each do |formula|
+      dependents_to_upgrade = []
+
+      dependents = kegs.select do |k|
```
MikeMcQuaid: `keg` would be nicer

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,145 @@ def upgrade_formula(f)
       nil
     end
   end
+
+  def upgradable_dependents(kegs, formulae)
+    formulae_to_upgrade = {}
+    formulae_pinned = {}
```
MikeMcQuaid: Might be worth making these `Set`s instead; `Formulary` does its own caching and it'll ease correct merging.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,145 @@ def upgrade_formula(f)
...
+  end
+
+  def broken_dependents(kegs, formulae)
+    formulae_to_reinstall = {}
+
+    CacheStoreDatabase.use(:linkage) do |db|
+      formulae.each do |formula|
+        dependents_to_reinstall = []
+
+        dependents = kegs.select do |f|
```
MikeMcQuaid: Use `keg` here

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,145 @@ def upgrade_formula(f)
...
+
+      formulae_to_upgrade.merge u
+      formulae_pinned.merge p
+    end
+
+    [formulae_to_upgrade, formulae_pinned]
+  end
+
+  def broken_dependents(kegs, formulae)
+    formulae_to_reinstall = {}
```
MikeMcQuaid: Might be worth making this a `Set` instead; `Formulary` does its own caching and it'll ease correct merging.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,145 @@ def upgrade_formula(f)
...
+
+    oh1 "Checking dependents for broken library links" if ARGV.verbose?
+    reinstallable = broken_dependents(kegs, formulae)
+
+    # Print the broken dependents.
+    if reinstallable.empty?
+      ohai "No broken dependents" if ARGV.verbose?
+    else
+      ohai "Reinstalling #{Formatter.pluralize(reinstallable.length, "broken dependent")} from source:"
+      puts reinstallable.values.map(&:full_specified_name).join(", ")
```
MikeMcQuaid: Just a thought as I'm not sure if this edge case is handled: if something is pinned and outdated and has broken linkage: if you try to `reinstall` it'll try to install the current version, I think. That case may warrant a special message about it being broken but because it's pinned we can't upgrade it.

amyspark: That's exactly the intention, reinstalling it since we cannot upgrade it. I'll restore the old messaging.

MikeMcQuaid: @amyspark Just to make sure I'm being clear (I'm not sure I am 😆) I'm proposing that we don't try to `reinstall` pinned, outdated packages with broken linkage but just display an `Error` instead for them.

amyspark: You were not, I apologize 😄 I'll add the partition logic again and print the corresponding error messages.

amyspark:
Thanks @MikeMcQuaid - @scpeters, do tell me when this is ready, so I can flatten this to a nice single commit 😃

scpeters:
We're getting closer! The logic for upgrading outdated dependents is working for me now; though there's still an issue with linkage checking, and I made an inline comment for that

amyspark:
@scpeters found and fixed! Thanks for your review!

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,167 @@ def upgrade_formula(f)
...
+    upgrade_formulae(upgradable)
+
+    # Assess the dependents tree again.
+    kegs = Formula.installed
+                  .map(&:opt_or_installed_prefix_keg)
+                  .reject(&:nil?)
+                  .reject { |f| f.runtime_dependencies.to_a.empty? }
+
+    oh1 "Checking dependents for broken library links" if ARGV.verbose?
+    reinstallable, pinned = broken_dependents(kegs, formulae).map(&:to_a)
```
scpeters: I think `broken_dependents` is not returning anything about pinned formulae; it's just returning `formulae_to_reinstall`

also, I get the following error at this line:

~~~
Error: undefined method `to_a' for #<Formulary::FormulaNamespace85f259522d1d32129f89b1cc281316cc::IgnitionMsgs1:0x0000000106016538>
Did you mean?  to_hash
               to_s
/usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.7/lib/ruby/2.3.0/set.rb:306:in `each_key'
/usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.7/lib/ruby/2.3.0/set.rb:306:in `each'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:313:in `map'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:313:in `check_dependents'
/usr/local/Homebrew/Library/Homebrew/cmd/upgrade.rb:86:in `upgrade'
/usr/local/Homebrew/Library/Homebrew/brew.rb:89:in `<main>'
~~~

amyspark: Those errors are one and the same; `broken_dependents` is supposed to return an `Array` (`.map`) containing two `Set`s (`&:to_a`) of `Formula`e. Fixed now.

scpeters:
My first test is successful; I'm running one more and then I'll be ready to approve.

scpeters:
Based on my second test, I think there is an edge case that is not handled properly. I think it's worth discussing, but I'm not sure if it's worth blocking over.

I think that the reinstallable broken dependents will be reinstalled in alphabetical order rather than topological order. For example, if I have 2 packages named `a`, `b` in a tap that both need to be reinstalled due to broken linkage, in my testing `a` will always be reinstalled before `b` even if `a` has `depends_on "b"` in its metadata.

Outside of that edge case, I'm ready to approve.

amyspark:
I think this is `Set` storing them in that order. Do you want me to whip up a `sort` code block so that we get alphabetical order for non-dependents, and then topological?

amyspark:
@scpeters There you go 👍

scpeters:
quick work! testing now...

scpeters:
Hm... the upgrade command is aborting during the `reinstallable.sort!` block and it fails to reinstall any of the broken formulae. So it's a bit broken by 8135d06.

Unless @MikeMcQuaid or someone else has a good idea on how to sort these, then I think we should revert 
8135d06 and open an issue describing the edge case.

amyspark:
@scpeters, I've reverted the sort stuff so that it does not break your testing. Could you please send me the stacktrace so I can have a look?

scpeters:
I can't get a stacktrace! It just exits, and strangely prints some Caveats again. I pasted the results of my hacking session into this gist:

* https://gist.github.com/scpeters/36c0ccc3ea016d88e0d49613871bedb1

amyspark:
@scpeters It was the use of `return` inside the block (I don't seem to remember that keyword works differently in Ruby). You can try again now.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -171,4 +181,208 @@ def upgrade_formula(f)
...
+                  .map(&:opt_or_installed_prefix_keg)
+                  .reject(&:nil?)
+                  .reject { |f| f.runtime_dependencies.to_a.empty? }
+
+    return if kegs.empty?
+
+    oh1 "Checking dependents for outdated formulae" if ARGV.verbose?
+    upgradable, pinned = upgradable_dependents(kegs, formulae).map(&:to_a)
+
+    upgradable.sort! do |a, b|
```
MikeMcQuaid: Would be good to put these in a function as there's multiple identical versions.

MikeMcQuaid:
Thanks for all the testing @scpeters and coding @amyspark!

scpeters:
@amyspark you've done it! This looks great to me now. Thanks for all your hard work!

Thanks for the extra reviews @MikeMcQuaid !

scpeters:
> I've not included tests as Homebrew only uses dummy formulae in its test suite.

If I could find the time, I would try to create some dummy formulae with dependencies on each other. I'm not sure how to mock broken linkage in the dummy formulae, but I imagine that could be done as well.

MikeMcQuaid:
I think this can go in without more test coverage; I trust @scpeters manual testing on this. Once we've got a final revision that both @amyspark and @scpeters are 👍 on being merged I'll 🚢 this!

scpeters:
I'm :+1: for :ship:

amyspark:
@MikeMcQuaid 🚀

MikeMcQuaid:
Great work here @amyspark and nice work @scpeters for all the testing here 👏

PR #4773, opened 2018-08-29
superenv: Use 02 optimization flag for Linux builds

iMichka:
`-Os` produces sometimes bigger binaries on Linux.
Also, llvm built with `-Os` is really slow at runtime for Linux.

Using `-02` aligns us with what Debian does, and as we are compiling most of our stuff with gcc (and not clang), it makes sense to use `-02` on Linux.
`-Os` does probably slightly different things when used on mac with llvm, compared to when it is used with gcc on Linux.

Related to #4758

-----

### DIFF
--- a/Library/Homebrew/extend/os/linux/extend/ENV/super.rb
+++ b/Library/Homebrew/extend/os/linux/extend/ENV/super.rb
@@ -7,6 +7,7 @@ module Superenv
   # @private
   def setup_build_environment(formula = nil)
     generic_setup_build_environment(formula)
+    self["HOMEBREW_OPTIMIZATION_LEVEL"] = "O2"
     self["HOMEBREW_DYNAMIC_LINKER"] = determine_dynamic_linker_path
     self["HOMEBREW_RPATH_PATHS"] = determine_rpath_paths(formula)
   end


Library/Homebrew/extend/os/linux/extend/ENV/super.rb
```diff
@@ -7,6 +7,7 @@ def self.bin
   # @private
   def setup_build_environment(formula = nil)
     generic_setup_build_environment(formula)
+    self["HOMEBREW_OPTIMIZATION_LEVEL"] = "O2"
```
scpeters: this might get overwritten by the generic `setup_build_environment` since it calls `super` before setting this variable to `"Os"`:

https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/ENV/super.rb#L39-L48

sjackman: Our `setup_build_environment`, which calls`generic_setup_build_environment`, where the `super` calls `setup_build_environment` in `extend/ENV/shared.rb`, so it behaves as desired.
https://github.com/Homebrew/brew/blob/0af24cf8bc6ae9572de19c1f1209320f81cf31e2/Library/Homebrew/extend/ENV/shared.rb#L30

MikeMcQuaid:
To confirm: is this the code that's currently present in Linuxbrew? If so: 👍, if not: I'd like it to get some testing there first.

sjackman:
Yes, this code is already being used in Linuxbrew.

PR #4778, opened 2018-08-30
cleanup: fix go_cache cleanup

DomT4:

-----

Oops 🙃. It's possible people want me to write this a different way, which I'm open to.

### DIFF
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -39,6 +39,12 @@ module CleanupRefinement
       directory? && %w[go_cache glide_home java_cache npm_cache gclient_cache].include?(basename.to_s)
     end
 
+    def go_cache_directory?
+      # Go makes its cache contents read-only to ensure cache integrity,
+      # which makes sense but is something we need to undo for cleanup.
+      directory? && %w[go_cache].include?(basename.to_s)
+    end
+
     def prune?(days)
       return false unless days
       return true if days.zero?
@@ -228,6 +234,7 @@ module Homebrew
       entries ||= [cache, cache/"Cask"].select(&:directory?).flat_map(&:children)
 
       entries.each do |path|
+        FileUtils.chmod_R 0755, path if path.go_cache_directory? && !dry_run?
         next cleanup_path(path) { path.unlink } if path.incomplete?
         next cleanup_path(path) { FileUtils.rm_rf path } if path.nested_cache?
 


Library/Homebrew/cleanup.rb
```diff
@@ -237,6 +237,10 @@ def cleanup_path(path)
       if dry_run?
         puts "Would remove: #{path} (#{path.abv})"
       else
+        # Go makes its cache contents read-only to ensure cache integrity,
+        # which makes sense but is something we need to undo for cleanup.
+        FileUtils.chmod_R 0755, path if path.to_s.include?(cache/"go_cache")
```
reitermarkus: I think this would fit better above where we check for `path.nested_cache?`.

DomT4:
Made some changes to the way this was written.

Library/Homebrew/cleanup.rb
```diff
@@ -213,6 +219,7 @@ def cleanup_cache(entries = nil)
       entries ||= [cache, cache/"Cask"].select(&:directory?).flat_map(&:children)
 
       entries.each do |path|
+        FileUtils.chmod_R 0755, path if path.needs_chowning? && !dry_run?
```
MikeMcQuaid: Does this need to be recursive? Could `+r` be used instead? Otherwise this may overwrite a decision to e.g. make this group writable.

Library/Homebrew/cleanup.rb
```diff
@@ -39,6 +39,12 @@ def nested_cache?
       directory? && %w[go_cache glide_home java_cache npm_cache gclient_cache].include?(basename.to_s)
     end
 
+    def needs_chowning?
+      # Go makes its cache contents read-only to ensure cache integrity,
+      # which makes sense but is something we need to undo for cleanup.
+      directory? && %w[go_cache].include?(basename.to_s)
```
MikeMcQuaid: This doesn't check if it needs chowning or not? Maybe worth checking if it's readable?

Library/Homebrew/cleanup.rb
DomT4: The cache wouldn't remove locally unless I ran recursively through everything; you end up with the build cache removed but the module cache being persistent. Which isn't the worst thing in the world, a longer-lasting cache is the most effective way for the Go changes to be useful, but I was planning to achieve that a little more legitimately than simply making it impossible for cleanup to remove without user intervention, heh.

> Otherwise this may overwrite a decision to e.g. make this group writable.

Where would that decision come from? I didn't think we especially supported user intervention in cache management?

Library/Homebrew/cleanup.rb
MikeMcQuaid: Could just tweak the name; `is_go_cache_directory?` or similar I think signals what it does a bit more?

DomT4: Sure, if you're happy with that it works for me.

I generalised it in case we needed to expand it to other tools/formulae/etc for whatever reason in the future, but we can always handle that in the future rather than me trying to peer into a 🔮, heh.

MikeMcQuaid: Aye, can refactor if and when we need it for more 👍

DomT4:
Change pushed.

MikeMcQuaid:
Perfect, thanks @DomT4!

DomT4:
Thanks Mike!

PR #4781, opened 2018-08-31
Allow dmg files with no extension to be converted

MTCoster:
Fixes the underlying problem discussed in https://github.com/Homebrew/homebrew-cask/issues/50622.

Currently, dmg archives with no extension will not be parsed correctly by `hdiutil`.  The added flag (`-srcimagekey diskimage-class=CRawDiskImage`) is a hint to `hdiutil` that the input file can be read like a dmg archive.

-----

### DIFF
--- a/Library/Homebrew/unpack_strategy/dmg.rb
+++ b/Library/Homebrew/unpack_strategy/dmg.rb
@@ -137,7 +137,8 @@ module UnpackStrategy
           cdr_path = mount_dir/path.basename.sub_ext(".cdr")
 
           system_command! "hdiutil",
-                          args: ["convert", "-quiet", "-format", "UDTO", "-srcimagekey", "diskimage-class=CRawDiskImage", "-o", cdr_path, path],
+                          args: ["convert", "-quiet", "-format", "UDTO", "-srcimagekey", "diskimage-class=CRawDiskImage",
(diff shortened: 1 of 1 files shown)


MTCoster:
@colindean What would you suggest to use as a source file for a test?

MTCoster:
Yes and no.  Without its extension, it succeeds before this patch but fails after.

I think the issue here is more complex than I'd originally suspected.  Your test image is of type `CUDIFDiskImage`, which makes sense because UDIF is the proper format for any file with a `.dmg` extension.  The Intel-provided image causing the original issue is `CRawDiskImage` for some reason, and is therefore not processed correctly by `hdiutil` without an extension.

Since this is apparently not a general issue, I propose withdrawing this PR in favour of a one-off patch on the cask file to add the extension.

A proper fix here would effectively involve re-implementing `hdiutil`'s format detection algorithm (which based on the output with `-verbose` accounts for far more than just the two mentioned here).

reitermarkus:
The root of the problem was also fixed by https://github.com/Homebrew/brew/pull/4662, so [the `#?.dmg` workaround isn't needed any longer](https://github.com/Homebrew/homebrew-cask/pull/51547).

MTCoster:
Wow, I totally missed that! Okay, definitely closing this PR now.

PR #4796, opened 2018-09-02
Cask: check that the tools for quarantining are available

amyspark:

-----

The Gatekeeper API I've written relies on an up-to-date CLT (or Xcode) install, and `xattr`'s `-r` flag for native recursion. There are cases where Swift is too old or is set to a deployment target earlier than MacOS 10.10, and thus cannot use the `URLResourceKey.quarantinePropertiesKey` constant. And in the current Mojave beta, `xattr` does not have Apple's `-r` extension for doing native filesystem traversal.

This pull request inserts an additional check in the Swift script, the `Quarantine.available?` function and `brew cask doctor`, and changes `propagate` to use `xargs` for recursion.

Fixes Homebrew/homebrew-cask#51554, and fixes Homebrew/homebrew-cask#51538.

### DIFF
--- a/Library/Homebrew/cask/cmd/doctor.rb
+++ b/Library/Homebrew/cask/cmd/doctor.rb
@@ -22,6 +22,7 @@ module Hbc
 
       def run
         check_software_versions
+        check_quarantine_support
         check_install_location
         check_staging_location
         check_taps
@@ -114,6 +115,23 @@ module Hbc
         locale_variables = ENV.keys.grep(/^(?:LC_\S+|LANG|LANGUAGE)\Z/).sort
 
         (locale_variables + environment_variables).sort.each(&method(:render_env_var))
+      end
+
+      def check_quarantine_support
+        ohai "Gatekeeper support"
+
+        status = Quarantine.check_quarantine_support
+
+        case status
+        when :quarantine_available
+          puts "Enabled"
+        when :no_swift
+          add_error "Swift is not available on this system."
+        when :no_quarantine
+          add_error "This feature requires the macOS 10.10 SDK or higher."
+        else
+          onoe "Unknown support status"
+        end
       end
 
       def user_tilde(path)
--- a/Library/Homebrew/cask/installer.rb
+++ b/Library/Homebrew/cask/installer.rb
@@ -186,7 +186,7 @@ module Hbc
       return unless quarantine?
       return unless Quarantine.available?
 
-      Quarantine.propagate(from: @downloaded_path, to: @cask.staged_path, command: @command)
+      Quarantine.propagate(from: @downloaded_path, to: @cask.staged_path)
     end
 
     def install_artifacts
--- a/Library/Homebrew/cask/quarantine.rb
+++ b/Library/Homebrew/cask/quarantine.rb
@@ -12,10 +12,29 @@ module Hbc
       @swift ||= DevelopmentTools.locate("swift")
     end
 
+    def check_quarantine_support
+      odebug "Checking quarantine support"
+
+      if swift.nil?
+        odebug "Swift is not available on this system."
+        return :no_swift
+      end
+
+      api_check = system_command(swift, args: [QUARANTINE_SCRIPT])
+
+      if api_check.exit_status == 5
+        odebug "This feature requires the macOS 10.10 SDK or higher."
+        return :no_quarantine
+      end
+
+      odebug "Quarantine is available."
+      :quarantine_available
+    end
+
     def available?
-      status = !swift.nil?
-      odebug "Quarantine is #{status ? "available" : "not available"}."
-      status
+      @status ||= check_quarantine_support
+
+      @status == :quarantine_available
     end
 
     def detect(file)
@@ -30,24 +49,24 @@ module Hbc
       quarantine_status
     end
 
-    def status(file, command: SystemCommand)
-      command.run("/usr/bin/xattr",
-                  args:        ["-p", QUARANTINE_ATTRIBUTE, file],
-                  print_stderr: false).stdout.rstrip
+    def status(file)
+      system_command("/usr/bin/xattr",
+                     args:         ["-p", QUARANTINE_ATTRIBUTE, file],
+                     print_stderr: false).stdout.rstrip
     end
 
-    def cask(cask: nil, download_path: nil, command: SystemCommand)
+    def cask(cask: nil, download_path: nil)
       return if cask.nil? || download_path.nil?
 
       odebug "Quarantining #{download_path}"
 
-      quarantiner = command.run(swift,
-                                args: [
-                                  QUARANTINE_SCRIPT,
-                                  download_path,
-                                  cask.url.to_s,
-                                  cask.homepage.to_s,
-                                ])
+      quarantiner = system_command(swift,
+                                   args: [
+                                     QUARANTINE_SCRIPT,
+                                     download_path,
+                                     cask.url.to_s,
+                                     cask.homepage.to_s,
+                                   ])
 
       return if quarantiner.success?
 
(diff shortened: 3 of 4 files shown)


Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -12,10 +12,28 @@ def swift
...
-    def available?
-      status = !swift.nil?
-      odebug "Quarantine is #{status ? "available" : "not available"}."
-      status
+    def available?(command: SystemCommand)
+      # We use Strings to force caching of the availability status.
+      @status ||= begin
+        if swift.nil?
+          odebug "Swift is not available on this system."
+          "N/A"
```
colindean: On first read, I ask: Why `"N/A"` here instead of "`false`"? 

Or, a symbol like `:no_swift` and the output of the method is trinary: `:no_swift`, `:no_quarantine`, `:quarantine_available` or something like that, with the method called something like `check_quarantine_availability` and `available?` being a check of `check_quarantine_availability == :quarantine_available`?

But then I see: 

> \# We use Strings to force caching of the availability status.

Ah, I've not heard of this. Why is this?

amyspark: Didn't know about the symbol stuff, and it sounds way cleaner! I'll redo this.

reitermarkus: If you want to cache a boolean, you should use 

```ruby
return @status if defined?(@status)
```

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -12,10 +12,32 @@ def swift
...
+      :quarantine_available
+    end
+
     def available?
-      status = !swift.nil?
-      odebug "Quarantine is #{status ? "available" : "not available"}."
-      status
+      @status ||= check_availability_status
+
+      odebug "Quarantine is #{(@status == :quarantine_available) ? "available" : "not available"}."
```
reitermarkus: Remove this, or move it to `check_availability_status`.

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -12,10 +12,32 @@ def swift
       @swift ||= DevelopmentTools.locate("swift")
     end
 
+    # @private
```
reitermarkus: No need to mark this as private, since everything except the explicitly documented parts of `Formula` should be considered private.

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -12,10 +12,30 @@ def swift
       @swift ||= DevelopmentTools.locate("swift")
     end
 
+    def check_availability_status(command: SystemCommand)
```
reitermarkus: Use `system_command` directly. We should try to get everything to use `system_command` across the board.

amyspark: 👍  While we're at it, may I also update the rest of `quarantine` as well?

reitermarkus: Sure, go ahead! 😉

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -12,10 +12,43 @@ def swift
       @swift ||= DevelopmentTools.locate("swift")
     end
 
+    def check_recursion_support
```
reitermarkus: Let's remove this and just default to the Mojave behaviour.

amyspark:
@reitermarkus, @MikeMcQuaid: I've tossed in a support check in `brew cask doctor`, just in case.

amyspark:
Oh, I can also force Swift to change its deployment target with `swift -target x86_64-apple-macosx10.10`. This should help users of 10.11 and earlier. Do you want me to add this flag to the existing calls?

jabenninghoff:
Hi @amyspark! A couple of things...

1. I'd be happy to help test with deployment targets ... as I mentioned in #4809, for some reason the default target on 10.10 is actually 10.9...
1. Adding checks to `brew doctor` seems to go against the intent of the command, which direct the user to fix a problem with their configuration. In the case of missing swift or having the wrong deployment target, I don't see how someone could fix either issue without upgrading the OS. Would it be better to leave out the quarantine check in `brew doctor`?

Library/Homebrew/cask/lib/hbc/cli/doctor.rb
```diff
@@ -27,6 +27,7 @@ def run
         check_taps
         check_load_path
         check_environment_variables
+        check_quarantine_support
```
commitay: Maybe move this into `check_software_versions` so it is listed with `macOS` and `SIP` rather than at the end?

amyspark:
> I'd be happy to help test with deployment targets ... as I mentioned in #4809, for some reason the default target on 10.10 is actually 10.9...

Thanks for the help @jabenninghoff ! Please test again with the next commit, it should now work for your SDK version.

> Adding checks to brew doctor seems to go against the intent of the command, which direct the user to fix a problem with their configuration. In the case of missing swift or having the wrong deployment target, I don't see how someone could fix either issue without upgrading the OS. Would it be better to leave out the quarantine check in brew doctor?

I'd like to hear @Homebrew/cask opinion about this; in any case, it should warn users if some features are not supported.

Library/Homebrew/cask/lib/hbc/cli/doctor.rb
amyspark: This adds an error message if there are missing tools; I can move it up the chain instead, if you want.

jabenninghoff:
Hi @amyspark, I tested with this PR and can confirm it's working now! Thanks!

FWIW, I can appreciate the desire to warn users when some features are not supported, but that's not what I would expect from `brew doctor`. I suspect people will issue reports if you add warnings that impact all users of older OSes (I would).

Library/Homebrew/cask/lib/hbc/cli/doctor.rb
```diff
@@ -116,6 +117,21 @@ def check_environment_variables
...
+        ohai "Gatekeeper support"
+
+        status = Quarantine.check_quarantine_support
+
+        case status
+        when :no_swift
+          add_error "Swift is not available on this system."
+        when :no_quarantine
+          add_error "This feature requires the macOS 10.10 SDK or higher."
+        else
```
colindean: I'd like to see:

* a `when :quarantine_available` for enabled instead of a default case
* a `fail "unknown quarantine status"` in the default case

This guards against `Quarantine.check_quarantine_support` changing its return.

Library/Homebrew/cask/lib/hbc/quarantine.rb
```diff
@@ -12,10 +12,29 @@ def swift
       @swift ||= DevelopmentTools.locate("swift")
     end
 
+    def check_quarantine_support
+      odebug "Checking quarantine support"
+
+      if swift.nil?
+        odebug "Swift is not available on this system."
+        return :no_swift
```
colindean: This is much better with symbol return values 👍

Library/Homebrew/cask/lib/hbc/cli/doctor.rb
```diff
@@ -116,6 +117,23 @@ def check_environment_variables
...
       end
 
+      def check_quarantine_support
+        ohai "Gatekeeper support"
+
+        status = Quarantine.check_quarantine_support
+
+        case status
+        when :quarantine_available
+          puts "Enable"
```
reitermarkus: Enable what?

Library/Homebrew/cask/lib/hbc/utils/quarantine.swift
```diff
@@ -7,36 +7,41 @@ struct swifterr: TextOutputStream {
...
+  exit(0)
 }
 else {
-  print(errorBag!.localizedDescription, to: &swifterr.stream)
-  exit(3)
-}
-
-exit(0)
+  exit(5)
+}
```
MikeMcQuaid: needs a newline

colindean:
To save a clickthrough, the error:

```
Error: Your HOMEBREW_GITHUB_API_TOKEN does not have sufficient scope!
Scopes they need: none
Scopes it has:
Create a new personal access token: https://github.com/settings/tokens/new?scopes=gist,public_repo&description=Homebrew
and then set the new HOMEBREW_GITHUB_API_TOKEN as the authentication method instead.
Warning: Error searching on GitHub: GitHub {
  "documentation_url": "https://developer.github.com/v3/#abuse-rate-limits",
  "message": "You have triggered an abuse detection mechanism. Please wait a few minutes before you try again."
}

HOMEBREW_GITHUB_API_TOKEN may be invalid or expired; check:
https://github.com/settings/tokens

```

amyspark:
@MikeMcQuaid restarted Travis build and everything is 👍  Can you review?

MikeMcQuaid:
Looks good, thanks @amyspark!

PR #4798, opened 2018-09-02
Revert "Add `$LOAD_PATH` for RuboCop cops."

MikeMcQuaid:
Reverts Homebrew/brew#4771

@reitermarkus Sorry, I didn't think about this before but this breaks one of the primary use-cases of our `rubocop` integration: being able to work with just `rubocop` loading just the cops outside of a `brew` command. A plain `rubocop` outside of Homebrew should work on Homebrew formulae inside the taps directory without `brew style`. I'm open to having a test added for this; I'm not quite sure what it would look like.


reitermarkus:
> being able to work with just `rubocop` loading just the cops outside of a `brew` command.

What? Can you give an example of what exactly should work?

MikeMcQuaid:
Install rubocop outside of Homebrew and run it on a formula.

PR #4799, opened 2018-09-02
Add `$LOAD_PATH` for RuboCop cops.

reitermarkus:

-----

Moved the `$LOAD_PATH` setup to a separate file, since this is the only thing we need to share.


MikeMcQuaid:
Can we add a test to verify that `rubocop` works outside of Homebrew as part of this? My concern with merging as-is is if someone `require`s something that does a `require "global"` then it will silently break.

PR #4801, opened 2018-09-03
class_cop: flag simple tests

DomT4:

-----

We have too many of these in core, they aren't an effective test, things _alarmingly often_ break in some way or another despite this being the formula's test, and I am tired of seeing my own quick reply asking for a better test on new formulae around core.

Note: If people see the `single_line?` element here & try to work around that by splitting lines I will see, and I will thwack that as well 👀 😈.

### DIFF
--- a/Library/Homebrew/rubocops/class_cop.rb
+++ b/Library/Homebrew/rubocops/class_cop.rb
@@ -82,6 +82,14 @@ module RuboCop
             return
           end
 
+          if test.body.single_line? &&
+             test.body.source.to_s.include?("version") ||
+             test.body.source.to_s.include?("help")
+            problem "Simple tests should be a last resort. " \
+                    "Please try to write a test that exercises a " \
+                    "deeper level of functionality"
+          end
+
           return unless test.body.single_line? &&
                         test.body.source.to_s == "true"
           problem "`test do` should contain a real test"
--- a/Library/Homebrew/test/rubocops/class_cop_spec.rb
+++ b/Library/Homebrew/test/rubocops/class_cop_spec.rb
@@ -139,4 +139,17 @@ describe RuboCop::Cop::FormulaAuditStrict::Test do
       end
     RUBY
   end
+
+  it "reports an offense when test is overly simple" do
+    expect_offense(<<~RUBY)
+      class Foo < Formula
+        url 'https://example.com/foo-1.0.tgz'
+
+        test do
+        ^^^^^^^ Simple tests should be a last resort. Please try to write a test that exercises a deeper level of functionality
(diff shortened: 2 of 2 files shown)


Library/Homebrew/rubocops/class_cop.rb
```diff
@@ -82,6 +82,14 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
             return
           end
 
+          if test.body.single_line? &&
+             test.body.source.to_s.include?("version") ||
+             test.body.source.to_s.include?("help")
```
reitermarkus: Maybe also `-v` and `-h`?

Library/Homebrew/rubocops/class_cop.rb
```diff
@@ -82,6 +82,14 @@ def audit_formula(_node, _class_node, _parent_class_node, body_node)
             return
           end
 
+          if test.body.single_line? &&
```
reitermarkus: Does something like `single_statement?` exist?

Library/Homebrew/rubocops/class_cop.rb
DomT4: The `version` here is intended mostly to match `version.to_s` and `/#{version}/`. I wasn't sure on `-v` because sometimes people seem to map it to `--verbose`, which could have legitimate usage in a test, but I'm happy to go for it and see how things go if you/others don't think it unwise.

`-h` definitely sounds like a 👍.

Library/Homebrew/rubocops/class_cop.rb
DomT4: I'm not aware of one, sadly. Could be wrong because Rubocop is a labyrinth heh, but not to my knowledge.

MikeMcQuaid: CC @GauthamGoli who may have thoughts. I also think `single_line` is a bit shaky for this.

Library/Homebrew/rubocops/class_cop.rb
MikeMcQuaid: This will make `subversion` and `whelp` both raise an error. Can you make it a bit more specific to those cases?

reitermarkus: You can match for `\bversion\b`.

Library/Homebrew/rubocops/class_cop.rb
GauthamGoli: https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/node_pattern.rb
I think this can be used.

PR #4815, opened 2018-09-04
rust/cargo: enable caching

commitay:

-----
The primary intention of this PR is to avoid having to download the complete `crates.io-index` package registry (currently a 125MB download) for each build. The formulae crate dependencies are also cached and can be reused between formulae. This will work without needing any changes to be made in formulae.

Similar to https://github.com/Homebrew/brew/pull/4774 one downside is increased disk usage. Also, `rust` formulae aren't as common (currently 35), but we do seem to be adding them quickly (16 this year so far.)

One possible problem that I've encountered with setting `CARGO_HOME` occurs if the install target (i.e. `"--root", prefix`) isn't set as it will install by default into `CARGO_HOME`.
```
  Installing /Users/commitay/Library/Caches/Homebrew/cargo_cache/bin/bat
warning: be sure to add `/Users/commitay/Library/Caches/Homebrew/cargo_cache/bin` to your PATH to be able to run the installed binaries
Error: Empty installation
```

Disclaimer: My internet is terrible, so while this is a decent improvement for me locally it might not actually be that useful.

Removing `cargo_cache` for each formula:
```
/usr/local/Cellar/bat/0.6.1: 5 files, 5.2MB, built in 11 minutes 45 seconds

/usr/local/Cellar/bat/0.6.1: 5 files, 5.2MB, built in 5 minutes 34 seconds
```
```
/usr/local/Cellar/exa/0.8.0: 9 files, 1.4MB, built in 7 minutes 17 seconds

/usr/local/Cellar/exa/0.8.0: 9 files, 1.4MB, built in 2 minutes 38 seconds
```
```
/usr/local/Cellar/fd/7.1.0: 9 files, 3.1MB, built in 6 minutes 43 seconds

/usr/local/Cellar/fd/7.1.0: 9 files, 3.1MB, built in 1 minute 56 seconds
```
```
/usr/local/Cellar/ripgrep/0.9.0: 11 files, 4.9MB, built in 6 minutes 54 seconds

/usr/local/Cellar/ripgrep/0.9.0: 11 files, 4.9MB, built in 2 minutes 45 seconds
```
```
/usr/local/Cellar/watchexec/1.9.0: 7 files, 2.3MB, built in 6 minutes 52 seconds

/usr/local/Cellar/watchexec/1.9.0: 7 files, 2.3MB, built in 2 minutes 28 seconds
```

The `cargo_cache` size after building all of the above formulae again was 247MB (including the 125MB package registry.)

### DIFF
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -36,7 +36,7 @@ module CleanupRefinement
     end
 
     def nested_cache?
-      directory? && %w[go_cache glide_home java_cache npm_cache gclient_cache].include?(basename.to_s)
+      directory? && %w[cargo_cache go_cache glide_home java_cache npm_cache gclient_cache].include?(basename.to_s)
     end
 
     def go_cache_directory?
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -1680,6 +1680,7 @@ class Formula
       HOMEBREW_PATH: nil,
       _JAVA_OPTIONS: "#{ENV["_JAVA_OPTIONS"]} -Duser.home=#{HOMEBREW_CACHE}/java_cache",
       GOCACHE: "#{HOMEBREW_CACHE}/go_cache",
+      CARGO_HOME: "#{HOMEBREW_CACHE}/cargo_cache",
     }
 
     ENV.clear_sensitive_environment!
@@ -2034,6 +2035,7 @@ class Formula
         stage_env[:_JAVA_OPTIONS] =
           "#{ENV["_JAVA_OPTIONS"]} -Duser.home=#{HOMEBREW_CACHE}/java_cache"
         stage_env[:GOCACHE] = "#{HOMEBREW_CACHE}/go_cache"
+        stage_env[:CARGO_HOME] = "#{HOMEBREW_CACHE}/cargo_cache"
         stage_env[:CURL_HOME] = ENV["CURL_HOME"] || ENV["HOME"]
       end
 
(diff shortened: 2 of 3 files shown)


commitay:
We could probably do the same for `cargo install` and `"--root", prefix` as we do here:

https://github.com/Homebrew/brew/blob/bfb2d53f98b9055b2fce53830f5b74a1f99ffb26/Library/Homebrew/rubocops/text_cop.rb#L52-L55

colindean:
Does it make sense for an enhancement to allow a user to specify their own cache path? As in, if I'm working in Rust regularly, would it be worthwhile for me to be able to set `HOMEBREW_CACHE_PATH_CARGO` or something like that in order to have Homebrew use my user's CARGO_HOME path in `~/.cargo` instead? 

This optional centralization might save significant disk space, but we'd probably want to determine the benefit before spending the time to implement it.

commitay:
> Does it make sense for an enhancement to allow a user to specify their own cache path? As in, if I'm working in Rust regularly, would it be worthwhile for me to be able to set `HOMEBREW_CACHE_PATH_CARGO` or something like that in order to have Homebrew use my user's `CARGO_HOME` path in `~/.cargo` instead? 

I suppose we could in theory but we probably wouldn't want to go down that road as people would want similar support for go, npm and everything else. We'd also have to handle any potential conflicts that might arise from the user and Homebrew attempting to access the cache at the same time.

MikeMcQuaid:
> Does it make sense for an enhancement to allow a user to specify their own cache path

@colindean FYI you can do that with `HOMEBREW_CACHE` for the Homebrew cache more generally but I agree with @commitay that it's not worth allowing each to be configurable. One can always use symlinks if needed.

PR #4816, opened 2018-09-04
Revert "Use JSON to marshal errors from children"

MikeMcQuaid:
Reverts Homebrew/brew#4752

This breaks `brew formula-analytics` and in turn https://formulae.brew.sh.


MikeMcQuaid:
@woodruffw Can reapply this when it's tested and working nicely with `brew formula-analytics`. I'm a bit concerned about how this blew up, though; is it monkeypatching all uses of JSON or something?

woodruffw:
> is it monkeypatching all uses of JSON or something?

I think the issue is the extra context being added to `#to_json` by `require "json/add/core"`: there are probably a few places where we're expecting `#to_json` to return just a JSON string object and now returning something more complicated instead.

In particular, I'm guessing that the `'s'` keys are coming from Symbol: https://github.com/flori/json/blob/master/lib/json/add/symbol.rb, and that the `json_class` keys are coming from `JSON.create_id` (which get thrown into every one of these extended serializations).

We can probably avoid this by just doing `require 'json/add/exception'` instead of `json/add/core`, since Exception serialization is the only thing we need. I'll open a new PR with the fixes (and hopefully a test).

PR #4817, opened 2018-09-04
docs/Formula-Cookbook: preach what we practise

claui:

-----

In both core and formula PRs, we tend to request from contributors more often that they inline method arguments rather than using `*args`:

```
system "make", "CC=#{ENV.cc}", "PREFIX=#{prefix}"
```

I feel we should codify that style in the docs so we preach what we practise.
Thanks @r4sas for the [heads up](https://github.com/Homebrew/homebrew-core/pull/31772#discussion_r214862652)!


MikeMcQuaid:
Thanks @claui!

PR #4828, opened 2018-09-05
scm shim: handle edge case

### LINKED ISSUE #4825: #4748 breaks git shim

- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew config` and `brew doctor` and included their output with your issue?

---

Since #4748 `HOMEBREW_GIT` is set by the `brew.sh`, whose value is default to be `git` . As a result, it completely bypasses the logic of the `brew/Library/Homebrew/shims/scm/git` with the following code. 

https://github.com/Homebrew/brew/blob/3a4914664e0b2ce83d9ceb029605c5a40e2f6bc7/Library/Homebrew/shims/scm/git#L88

This has the following direct consequences. 

* For users on macOS, invoking git no longer checks whether the Xcode license is accepted. 
https://github.com/Homebrew/brew/blob/3a4914664e0b2ce83d9ceb029605c5a40e2f6bc7/Library/Homebrew/shims/scm/git#L111-L137
* For users on old system for both macOS and Linux, `brew update` fails as it no longer uses brew git as it used to be. For example on Linuxbrew with old system, I got:
```
$ brew update
error: unknown option `list'
usage: git tag [-a|-s|-u <key-id>] [-f] [-m <msg>|-F <file>] <tagname> [<head>]
   or: git tag -d <tagname>...
   or: git tag -l [-n[<num>]] [<pattern>]
   or: git tag -v <tagname>...

    -l                    list tag names
    -n[<n>]               print <n> lines of each tag message
    -d                    delete tags
    -v                    verify tags

Tag creation options
    -a                    annotated tag, needs a message
    -m <msg>              message for the tag
    -F <FILE>             message in a file
    -s                    annotated and GPG-signed tag
    -u <key-id>           use another key to sign the tag
    -f, --force           replace the tag if exists

Tag listing options
    --contains <commit>   print only tags that contain the commit

Unknown option: -c
usage: git [--version] [--exec-path[=GIT_EXEC_PATH]] [--html-path]
           [-p|--paginate|--no-pager] [--no-replace-objects]
           [--bare] [--git-dir=GIT_DIR] [--work-tree=GIT_WORK_TREE]
           [--help] COMMAND [ARGS]
Error: Could not 'git stash' in /home/comp/chengxu/usr!
Please stash/commit manually if you need to keep your changes or, if not, run:
  cd /home/comp/chengxu/usr
  git reset --hard origin/master
```

My brew config:
```
$ brew config
HOMEBREW_VERSION: 1.7.3
ORIGIN: https://github.com/Linuxbrew/brew.git
HEAD: 76878ffd7882076d23227eff8aae6bdf500e582b
Last commit: 10 hours ago
Core tap ORIGIN: https://github.com/Linuxbrew/homebrew-core
Core tap HEAD: d57f2b8664b0cf8ed5b67023b4b1d1d356c65307
Core tap last commit: 63 minutes ago
HOMEBREW_PREFIX: /home/comp/chengxu/usr
HOMEBREW_REPOSITORY: /home/comp/chengxu/usr
HOMEBREW_CELLAR: /home/comp/chengxu/usr/Cellar
HOMEBREW_CACHE: /tmp/chengxu/cache/homebrew
HOMEBREW_BOTTLE_DEFAULT_DOMAIN_LINUX: https://linuxbrew.bintray.com
HOMEBREW_BOTTLE_DEFAULT_DOMAIN_MACOS: https://homebrew.bintray.com
HOMEBREW_DEVELOPER: true
HOMEBREW_FORCE_VENDOR_RUBY: true
HOMEBREW_GIT: git
HOMEBREW_LOGS: /tmp/chengxu/cache/homebrew/logs
HOMEBREW_NO_ANALYTICS: true
HOMEBREW_NO_AUTO_UPDATE: true
HOMEBREW_SANDBOX: true
CPU: 24-core 64-bit westmere
Homebrew Ruby: 2.3.7 => /home/comp/chengxu/usr/Library/Homebrew/vendor/portable-ruby/2.3.7/bin/ruby
Clang: 6.0 build 601
Git: 1.7.1 => /usr/bin/git
Curl: 7.19.7 => /usr/bin/curl
Java: N/A
Kernel: Linux 2.6.32-754.el6.x86_64 x86_64 GNU/Linux
OS: CentOS release 6.10 (Final) (Final)
Host glibc: 2.14
/usr/bin/gcc: 4.4.7
glibc: 2.23
gcc: 5.5.0_4
xorg: 20170115
```

Noted that although I find this bug in the Linux land, it will affect Mac users as well in certain cases.

cc @sjackman

xu-cheng:

-----

Since #4748 `HOMEBREW_GIT` is set by the `brew.sh`, whose value is default to be `git`.
As a result, it completely bypasses the logic of the shims/scm/git.

This fixes the issue by checking whether `HOMEBREW_GIT` and
`HOMEBREW_SVN` are set to be `git` and `svn` respectively.

Fixes #4825.

### DIFF
--- a/Library/Homebrew/shims/scm/git
+++ b/Library/Homebrew/shims/scm/git
@@ -85,10 +85,16 @@ fi
 
 case "$(lowercase "$SCM_FILE")" in
   git)
-    [[ -n "$HOMEBREW_GIT" ]] && safe_exec "$(type -P "$HOMEBREW_GIT")" "$@"
+    if [[ -n "$HOMEBREW_GIT" && "$HOMEBREW_GIT" != git ]]
+    then
+      safe_exec "$(type -P "$HOMEBREW_GIT")" "$@"
+    fi
     ;;
   svn)
-    [[ -n "$HOMEBREW_SVN" ]] && safe_exec "$(type -P "$HOMEBREW_SVN")" "$@"
+    if [[ -n "$HOMEBREW_SVN" && "$HOMEBREW_SVN" != svn ]]
+    then
+      safe_exec "$(type -P "$HOMEBREW_SVN")" "$@"
+    fi
     ;;
 esac
 


MikeMcQuaid:
Thanks @xu-cheng!

sjackman:
Thanks, @xu-cheng !

PR #4833, opened 2018-09-06
language/python: support pypy

jonchang:

-----

Now that PyPy is maturing (it builds numpy and scipy now!) I think it would be good for Homebrew to support this, if not in core, but at least permit taps to use PyPy as opposed to Python. The Python dev team do not have speed as a huge priority, if I recall they prioritize readability and ease of understanding of the CPython source code.

This is an opt-in change. Formulae using the recommended `virtualenv_install_with_resources` method of installing a python formula and its dependencies merely need to change `depends_on "python@"` to `depends_on "pypy"`.

There is an API change here, but I've checked `homebrew/core` and the first ten pages of GitHub search and there are no formulae passing an argument to `Language::Python.homebrew_site_packages`. A backwards compatible change could be made but I think it's cleaner this way.

Formulae doing something more advanced need a bit more work, e.g., this diff for `numpy` successfully compiles a PyPy-enabled NumPy installation. I couldn't get a side-by-side install for Python 3, Python 3, and PyPy to work though. There's some compiler confusion going on. But PyPy by itself is a-ok.

```diff
diff --git a/Formula/numpy.rb b/Formula/numpy.rb
index 2b08e8a790..4d4948de23 100644
--- a/Formula/numpy.rb
+++ b/Formula/numpy.rb
@@ -23,8 +23,7 @@ class Numpy < Formula
   option "without-python@2", "Build without python2 support"
 
   depends_on "gcc" => :build # for gfortran
-  depends_on "python@2" => :recommended
-  depends_on "python" => :recommended
+  depends_on "pypy" => :recommended
 
   resource "nose" do
     url "https://files.pythonhosted.org/packages/58/a5/0dc93c3ec33f4e281849523a5a913fa1eea9a3068acfa754d44d88107a44/nose-1.3.7.tar.gz"
@@ -33,17 +32,18 @@ class Numpy < Formula
 
   def install
     Language::Python.each_python(build) do |python, version|
-      dest_path = lib/"python#{version}/site-packages"
+      site_packages = Language::Python.site_packages(python)
+      dest_path = prefix/site_packages
       dest_path.mkpath
 
-      nose_path = libexec/"nose/lib/python#{version}/site-packages"
+      nose_path = libexec/"nose"/site_packages
       resource("nose").stage do
         system python, *Language::Python.setup_install_args(libexec/"nose")
         (dest_path/"homebrew-numpy-nose.pth").write "#{nose_path}\n"
       end
 
       if build.head?
-        ENV.prepend_create_path "PYTHONPATH", buildpath/"tools/lib/python#{version}/site-packages"
+        ENV.prepend_create_path "PYTHONPATH", buildpath/"tools"/site_packages
         resource("Cython").stage do
           system python, *Language::Python.setup_install_args(buildpath/"tools")
         end
@@ -51,7 +51,7 @@ class Numpy < Formula
 
       system python, "setup.py",
         "build", "--fcompiler=gnu95", "--parallel=#{ENV.make_jobs}",
-        "install", "--prefix=#{prefix}",
+        "install", "--prefix=#{prefix}", "--install-scripts=#{bin}",
         "--single-version-externally-managed", "--record=installed.txt"
     end
   end
```

* Currently, only pypy for python2 is supported. pypy3 has some issues that I'm investigating. But once that's figured out it should be simple enough to just add `pypy3` to the changed code.

* Setting the `--install-scripts` path is necessary for PyPy but not for regular Python. Not sure why this is. I dug up this commit https://github.com/Homebrew/homebrew-core/commit/7d376d5fdf5 but there's not a whole lot of context why scripts were installed to `share` instead of `bin`.
    * From what I can tell, regular scripts installed via pip are put in `bin` and I *think* their shebang is modified to point to the `opt_prefix`.
    * This means they survive version upgrades anyway as long as python still is installed. I don't recall if the `opt_prefix` thing was introduced before or after 2012. So if that makes sense I'd like to make a follow up pull request in core to get that removed from the pypy formula.

### DIFF
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -9,13 +9,21 @@ module Language
       Version.create(version.to_s)
     end
 
-    def self.homebrew_site_packages(version = "3.7")
-      HOMEBREW_PREFIX/"lib/python#{version}/site-packages"
+    def self.homebrew_site_packages(python = "python3.7")
+      HOMEBREW_PREFIX/site_packages(python)
+    end
+
+    def self.site_packages(python = "python3.7")
+      if python == "pypy"
+        "site-packages"
+      else
+        "lib/python#{major_minor_version python}/site-packages"
+      end
     end
 
     def self.each_python(build, &block)
       original_pythonpath = ENV["PYTHONPATH"]
-      { "python@3" => "python3", "python@2" => "python2.7" }.each do |python_formula, python|
+      { "python@3" => "python3", "python@2" => "python2.7", "pypy" => "pypy" }.each do |python_formula, python|
         python_formula = Formulary.factory(python_formula)
         next if build.without? python_formula.to_s
 
@@ -23,7 +31,7 @@ module Language
         ENV["PYTHONPATH"] = if python_formula.installed?
           nil
         else
-          homebrew_site_packages(version)
+          homebrew_site_packages(python)
         end
         block&.call python, version
       end
@@ -31,11 +39,10 @@ module Language
     end
 
     def self.reads_brewed_pth_files?(python)
-      version = major_minor_version python
-      return unless homebrew_site_packages(version).directory?
-      return unless homebrew_site_packages(version).writable_real?
+      return unless homebrew_site_packages(python).directory?
+      return unless homebrew_site_packages(python).writable_real?
 
-      probe_file = homebrew_site_packages(version)/"homebrew-pth-probe.pth"
+      probe_file = homebrew_site_packages(python)/"homebrew-pth-probe.pth"
       begin
         probe_file.atomic_write("import site; site.homebrew_was_here = True")
         with_homebrew_path { quiet_system python, "-c", "import site; assert(site.homebrew_was_here)" }
@@ -69,6 +76,7 @@ module Language
         --no-user-cfg
         install
         --prefix=#{prefix}
+        --install-scripts=#{prefix}/bin
         --single-version-externally-managed
         --record=installed.txt
       ]
@@ -101,17 +109,16 @@ module Language
 
         # Find any Python bindings provided by recursive dependencies
         formula_deps = formula.recursive_dependencies
-        xy = Language::Python.major_minor_version python
         pth_contents = formula_deps.map do |d|
           next if d.build?
 
-          dep_site_packages = Formula[d.name].opt_lib/"python#{xy}/site-packages"
+          dep_site_packages = Formula[d.name].opt_prefix/Language::Python.site_packages(python)
           next unless dep_site_packages.exist?
 
           "import site; site.addsitedir('#{dep_site_packages}')\n"
         end.compact
         unless pth_contents.empty?
-          (venv_root/"lib/python#{xy}/site-packages/homebrew_deps.pth").write pth_contents.join
+          (venv_root/Language::Python.site_packages(python)/"homebrew_deps.pth").write pth_contents.join
         end
 
         venv
@@ -140,7 +147,7 @@ module Language
       def virtualenv_install_with_resources(options = {})
         python = options[:using]
         if python.nil?
-          wanted = %w[python python@2 python2 python3 python@3].select { |py| needs_python?(py) }
+          wanted = %w[python python@2 python2 python3 python@3 pypy].select { |py| needs_python?(py) }
           raise FormulaAmbiguousPythonError, self if wanted.size > 1
 
           python = wanted.first || "python2.7"
@@ -177,9 +184,8 @@ module Language
           @formula.resource("homebrew-virtualenv").stage do |stage|
             old_pythonpath = ENV.delete "PYTHONPATH"
             begin
-              xy = Language::Python.major_minor_version(@python)
               staging = Pathname.new(stage.staging.tmpdir)
-              ENV.prepend_create_path "PYTHONPATH", staging/"target/lib/python#{xy}/site-packages"
+              ENV.prepend_create_path "PYTHONPATH", staging/"target"/Language::Python.site_packages(@python)
               @formula.system @python, *Language::Python.setup_install_args(staging/"target")
               @formula.system @python, "-s", staging/"target/bin/virtualenv", "-p", @python, @venv_root
             ensure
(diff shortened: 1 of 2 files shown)


Library/Homebrew/language/python.rb
```diff
@@ -66,6 +73,7 @@ def self.setup_install_args(prefix)
         --no-user-cfg
         install
         --prefix=#{prefix}
+        --install-scripts=#{prefix}/bin
```
DomT4: `#{bin}`?

DomT4:
> there's not a whole lot of context why scripts were installed to share instead of bin.

I assume to prevent things installed using `pypy`'s version of `pip` smashing into executables installed using the more widely used `python` and `python3` versions of `pip`.

> From what I can tell, regular scripts installed via pip 

Outside of Homebrew, or? For me `pip_pypy install virtualenv` results in `virtualenv` living at `/usr/local/share/pypy/virtualenv`, with the shebang `#!/usr/local/Cellar/pypy/6.0.0/bin/pypy`.

Library/Homebrew/language/python.rb
jonchang: That variable doesn't exist in this context. `prefix` is passed from the caller (usually the formula for a more advanced build) and typically goes in the formula equivalent of `libexec` or similar.

jonchang:
> I assume to prevent things installed using pypy's version of pip smashing into executables installed using the more widely used python and python3 versions of pip.

Aren't `pip_pypy` and friends are already renamed?

> Outside of Homebrew, or? For me pip_pypy install virtualenv results in virtualenv living at /usr/local/share/pypy/virtualenv, with the shebang #!/usr/local/Cellar/pypy/6.0.0/bin/pypy.

With `pip`, not `pip_pypy`. With regular pip for python 2 the shebang is at `#!/usr/local/opt/python@2/bin/python2.7`. Not sure what the python@2 formula is doing differently.

Library/Homebrew/language/python.rb
DomT4: I assumed it would exist when this variable would actually be called? We use `buildpath` in this file, for example, which obviously isn't defined at this point. Could be wrong, it's 5am and I definitely should be asleep 😅.

DomT4:
> Aren't pip_pypy and friends are already renamed?

Only the `pip` and `setuptools` we install. Anything the user runs `pip_pypy install` on isn't appended with `_pypy`.

> Not sure what the python@2 formula is doing differently.

Setting `sys.executable`.

Library/Homebrew/language/python.rb
jonchang: Yeah, it does exist while called, but this function doesn't have access to it. `prefix` is passed in by the formula, and it can be anything path-like. Check out e.g. [sphinx-doc.rb](https://github.com/Homebrew/homebrew-core/blob/d9908480053b8ed2db53d44e485bd669baa05970/Formula/sphinx-doc.rb#L142) for what I mean. `libexec` (in the formula's context) turns into `prefix` here.

`buildpath` works because it's part of the `Virtualenv` mixin class, which I believe would have access to its cohabitants variables like `bin` etc.

MikeMcQuaid:
@jonchang As a more general question: what benefits does this DSL provide compared to doing it manually in the formula? Does it matter if you have `pypy` and `python@2` in the same dependency tree (or process)? If so, does this detect or handle that in any way? What are the caveats if any you see from this approach.

FYI when I read the PR description I was 👎 on this but the code change is minimal enough I'm a very tentative 👍.

jonchang:
> As a more general question: what benefits does this DSL provide compared to doing it manually in the formula?  

This provides feature-parity for Homebrew/brew support for python@2 installs using `virtualenv_install_with_resources`. The idea is that an existing formula using that DSL can replace e.g., `depends_on "python@2"` with `depends_on "pypy"` and it should Just Work.

> If so, does this detect or handle that in any way?

It does not detect formulae doing the same thing manually, since I think that would be better handeld by an audit. Many formulae are already doing the "manual approach" (which here I take to mean "not using `virtualenv_install_with_resources`"). This is in theory extremely fragile as any changes to the `python@2` or `python` formulae could break the install for these formulae. Grepping for `site-packages` or `site_packages` show that many formulae hardcode the location of `site-packages`, assuming that it will always live at e.g., `$HOMEBREW_PREFIX/lib/python2.7/site-packages` rather than reading out the value of `Language::Python.homebrew_site_packages`.

If that's something that should be fixed as well or instead of this patch, I think a new audit would be a better approach.

> What are the caveats if any you see from this approach.

The caveat is that if you are doing custom stuff (anything but `virtualenv_install_with_resources`) you'll have to look at the definitions in `language/python.rb` to see how PyPy does things differently. Namely, it installs `site-packages` directly into the prefix, rather than into `lib/python2.7/site-packages`. Formula authors who wish to continue using `python@2` should see no change.

Arguably, if it makes more sense for formula authors to do this manually in all cases, this language-specific support should be removed from Homebrew/brew. I'm agnostic with respect to that option.

> Does it matter if you have pypy and python@2 in the same dependency tree (or process)?

There should be no issue with both pypy and python@2 in the same dependency tree. `virtualenv_install_with_resources` already detects any potential ambiguity and raises an exception asking the formula writer to clarify with `:using => "python@2"` or similar.

I don't see how it would be possible for pypy and python@2 to be in the same process; any scripts in `bin` are sandboxed into their respective virtualenvs and wouldn't have access to modules from another version of python installed. Side-by-side installs of libraries will need special handling, but formulae like `numpy` already do this. Applications should pick the right version of Python for them and stick with it.

> FYI when I read the PR description I was 👎 on this but the code change is minimal enough I'm a very tentative 👍.

👍

MikeMcQuaid:
All sounds good to me @jonchang! Could you try to write some more tests in Library/Homebrew/test/language/python_spec.rb to increase the % of the patch that's hit? After that I'm 👍 on merging this. Thanks!

maxnordlund:
One thing to note about PyPy 3 is that it's 3.5, not 3.7, though they are working on 3.6 and 3.7 was [released](https://www.python.org/downloads/release/python-370/) just a couple of months ago.

jonchang:
@MikeMcQuaid ok, I added tests for some of the functions in `Language::Python`. Will merge when tests are 📗

MikeMcQuaid:
Thanks @jonchang!

Issue #4780, opened 2018-08-31
Add `brew configure-shell` to make installs outside of /usr/local easier


jamescostian:
_Originally this was done for [Linuxbrew](https://linuxbrew.sh/) because it doesn't install to `/usr/local`, but the [leader of Linuxbrew mentioned that this is helpful for non-standard installations of Homebrew for macOS as well](https://github.com/Linuxbrew/brew/pull/829#issuecomment-417446720). For more information and to see some of the deliberation, check out [the issue for Linuxbrew](https://github.com/Linuxbrew/brew/issues/812)._

When installing outside of `/usr/local/` one must update their `PATH`, `MANPATH`, and `INFOPATH` (or at least just their `PATH`). Doing so is tedious and easy to automate. With this PR, [the alternative install guide](https://docs.brew.sh/Installation#untar-anywhere) would have this appended to its oneliner: ` && homebrew/bin/brew configure-shell`.

In addition to simplifying installs outside of `/usr/local` for macOS users, it's important to realize that this command has been created for Linuxbrew users, and after the merge of Linuxbrew and Homebrew, it would be nice if Linuxbrew users could still have this simple way of setting up their `PATH` and friends (that is, `PATH`, `MANPATH`, and `INFOPATH`).

# Code

It's 42 SLoC for the feature, 54 SLoC for tests for said feature, and 8 LoC to document it in `brew.1` and `Manpage.md`.

I'd make the same PR here as I made for Linuxbrew (that is, I'd make [this PR](https://github.com/Linuxbrew/brew/pull/829)). The only difference would be that I would use a regex with `^\s*source #{env_path}` to check if `brew.env` is already being sourced -  [here's the source of that idea](https://github.com/Linuxbrew/brew/pull/829/files#r214203970)

# Motivation

If you're installing outside of `/usr/local` then it's nice for your `PATH` and friends to all be updated. Users can do this manually, but it's tedious and this makes non-`/usr/local` installs feel more like first-class citizens

# Relevancy

This is relevant for 99% of non-`/usr/local` installs, but when you consider the merger with Linuxbrew I think that there will be a lot of non-`/usr/local` installs because Linuxbrew doesn't install to `/usr/local` ([see this](https://github.com/Linuxbrew/brew/wiki/FAQ#why-install-in-homelinuxbrewlinuxbrew)). In addition, there are already users who have non-`/usr/local` installs on their macs.

# Alternatives

[...]

PR #4849, opened 2018-09-06
Add brew shellenv - see #4780

jamescostian:

-----

Adds `brew shellenv` based on conversation in #4780

### DIFF
--- a/Library/Homebrew/test/cmd/shellenv_spec.rb
+++ b/Library/Homebrew/test/cmd/shellenv_spec.rb
@@ -1,0 +1,8 @@
+describe "brew shellenv", :integration_test do
+  it "doesn't fail" do
+    expect { brew "shellenv" }
+      .to output(%r{#{HOMEBREW_PREFIX}/bin}).to_stdout
+      .and not_to_output.to_stderr
+      .and be_a_success
+  end
+end
--- a/Library/Homebrew/cmd/shellenv.rb
+++ b/Library/Homebrew/cmd/shellenv.rb
@@ -1,0 +1,16 @@
+#:  * `shellenv`:
+#:    Prints export statements - run them in a shell and this installation of
+#:    Homebrew will be included into your PATH, MANPATH, and INFOPATH.
+#:    Tip: have your dotfiles eval the output of this command
+
+module Homebrew
+  module_function
+
+  def shellenv
+    puts <<~EOS
+      export PATH="#{HOMEBREW_PREFIX}/bin:#{HOMEBREW_PREFIX}/sbin:$PATH"
+      export MANPATH="#{HOMEBREW_PREFIX}/share/man:$MANPATH"
+      export INFOPATH="#{HOMEBREW_PREFIX}/share/info:$INFOPATH"
+    EOS
+  end
+end
--- a/docs/Manpage.md
+++ b/docs/Manpage.md
@@ -441,6 +441,11 @@ With `--verbose` or `-v`, many commands print extra debugging information. Note 
 
     If `--env=std` is passed, use the standard `PATH` instead of superenv's.
 
+  * `shellenv`:
+    Prints export statements - run them in a shell and this installation of
+    Homebrew will be included into your PATH, MANPATH, and INFOPATH.
+    Tip: have your dotfiles eval the output of this command
+
   * `style` [`--fix`] [`--display-cop-names`] [`--only-cops=``cops`|`--except-cops=``cops`] [`files`|`taps`|`formulae`]:
     Check formulae or files for conformance to Homebrew style guidelines.
 
--- a/manpages/brew.1
+++ b/manpages/brew.1
@@ -404,6 +404,9 @@ If \fB\-\-desc\fR is passed, search formulae with a description matching \fItext
 If \fB\-\-env=std\fR is passed, use the standard \fBPATH\fR instead of superenv\'s\.
 .
 .IP "\(bu" 4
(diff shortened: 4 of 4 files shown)


Library/Homebrew/test/cmd/shellenv_spec.rb
```diff
@@ -0,0 +1,7 @@
+describe "brew shellenv", :integration_test do
+  it "doesn't fail" do
```
MikeMcQuaid: Could you `assert_match` on e.g. the `HOMEBREW_PREFIX/bin` being present?

sjackman:
Looks good to me. Thanks, James! The one minor test change suggested by make would be good, to check that stdout contains `#{HOMEBREW_PREFIX}/bin`.

sjackman:
After this is merge into Homebrew/brew and Linuxbrew/brew, we can update the installation instructions.

jamescostian:
Just added that to the test 👍

sjackman:
Merged. Thanks for your contribution to Homebrew, James! 🏆

sjackman:
There's two ways that `brew shellenv` could be added to ~/.profile.
```sh
~/.linuxbrew/bin/brew shellenv >>~/.profile
```
or
```sh
echo 'eval $(~/.linuxbrew/bin/brew shellenv)' >>~/.profile
```
I have a slight preference for the latter using `eval`. Your opinion?

jamescostian:
Running `time bin/brew shellenv` on my laptop (processor is i7-4870HQ, running at 2.8GHz, mostly idling) reports a real time of 0.40s, which IMO makes it way too slow for me to be willing to `eval` it more than once

MikeMcQuaid:
@sjackman My recommendation would be to advise people put `eval` in their `profile` and those who care about the overhead can hardcode it if necessary.

Another possible extension for this script would be exporting e.g. `HOMEBREW_PREFIX` and `HOMEBREW_CELLAR` so scripts (or the profile itself) can avoid calling `brew --prefix` to further improve performance. 

Nice work @jamescostian!

sjackman:
> My recommendation would be to advise people put eval in their profile and those who care about the overhead can hardcode it if necessary.

Is that so that we can change the output of `brew shellenv`, and have all users up-to-date without needing to modify their shell configuration?

I suspect that implementing this script in shell rather than Ruby would execute quite a bit faster. If it's going in `~/.profile` maybe it's not such a big deal, since it'd only be executed when opening a new terminal (or ssh session). I'd be more concerned if it's going in `~/.bashrc` where it'd be executed for every subshell.

MikeMcQuaid:
> Is that so that we can change the output of brew shellenv, and have all users up-to-date without needing to modify their shell configuration?

Yep 👍 



> I suspect that implementing this script in shell rather than Ruby would execute quite a bit faster

It would indeed. Given it only relies on `HOMEBREW_PREFIX` being defined and even any extensions I see would only need things already defined in the Bash layer it should be possible to rewrite it in Bash, name it `cmd/shellenv.sh` and it'll avoid needing to ever execute the Ruby layer.

sjackman:
@jamescostian Would you be interested in submitting a PR to rewrite `shellenv` in shell so that it takes less time to execute?

jamescostian:
Sure, the bash rewrite is pretty simple:

```bash
echo "export PATH=\"$HOMEBREW_PREFIX/bin:$HOMEBREW_PREFIX/sbin:\$PATH\""
echo "export MANPATH=\"$HOMEBREW_PREFIX/share/man:\$MANPATH\""
echo "export INFOPATH=\"$HOMEBREW_PREFIX/share/info:\$INFOPATH\""
```

But there are two issues:

1. I'm getting an error at the end of the output: "Library/Homebrew/brew.sh: line 405: homebrew-shellenv: command not found" - any tips for that?
2. Even if I comment out the line that makes that error, the execution time is .10s

That execution time seems to be impacted most from [this line](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/brew.sh#L262) and [this line](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/brew.sh#L58). They both talk to `git` and they both look important enough to keep.

One way to decrease the execution time to .01s would be to treat `shellenv` completely differently:

```diff
--- a/bin/brew
+++ b/bin/brew
@@ -65,7 +65,7 @@ do
 done

 # test-bot does environment filtering itself
-if [[ -z "$HOMEBREW_NO_ENV_FILTERING" && "$1" != "test-bot" ]]
+if [[ -z "$HOMEBREW_NO_ENV_FILTERING" && "$1" != "test-bot" && "$1" != "shellenv" ]]
 then
   PATH="/usr/bin:/bin:/usr/sbin:/sbin"

@@ -82,6 +82,11 @@ then
   done

   exec /usr/bin/env -i "${FILTERED_ENV[@]}" /bin/bash "$HOMEBREW_LIBRARY/Homebrew/brew.sh" "$@"
+elif [[ "$1" == "shellenv" ]]
+then
+  echo "export PATH=\"$HOMEBREW_PREFIX/bin:$HOMEBREW_PREFIX/sbin:\$PATH\""
+  echo "export MANPATH=\"$HOMEBREW_PREFIX/share/man:\$MANPATH\""
+  echo "export INFOPATH=\"$HOMEBREW_PREFIX/share/info:\$INFOPATH\""
 else
   source "$HOMEBREW_LIBRARY/Homebrew/brew.sh"
 fi
```

But obviously that's not very elegant...

@MikeMcQuaid thoughts?

PR #4851, opened 2018-09-06
Add Homebrew::DEFAULT_PREFIX for Linux

sjackman:
The default prefix is /usr/local on macOS
and /home/linuxbrew/.linuxbrew on Linux.

-----

### DIFF
--- a/Library/Homebrew/cmd/link.rb
+++ b/Library/Homebrew/cmd/link.rb
@@ -52,7 +52,7 @@ module Homebrew
       end
 
       if keg_only
-        if HOMEBREW_PREFIX.to_s == "/usr/local"
+        if Homebrew.default_prefix?
           f = keg.to_formula
           caveats = Caveats.new(f)
 
--- a/Library/Homebrew/global.rb
+++ b/Library/Homebrew/global.rb
@@ -42,12 +42,20 @@ HOMEBREW_BOTTLE_DOMAIN = ENV["HOMEBREW_BOTTLE_DOMAIN"] ||
                          HOMEBREW_BOTTLE_DEFAULT_DOMAIN
 
 require "fileutils"
+require "os"
+require "os/global"
 
 module Homebrew
   extend FileUtils
 
+  DEFAULT_PREFIX ||= "/usr/local".freeze
+
   class << self
     attr_writer :failed, :raise_deprecation_exceptions, :auditing, :args
+
+    def Homebrew.default_prefix?(prefix = HOMEBREW_PREFIX)
+      prefix.to_s == DEFAULT_PREFIX
+    end
 
     def failed?
       @failed ||= false
@@ -108,7 +116,6 @@ HOMEBREW_INTERNAL_COMMAND_ALIASES = {
 
 require "set"
 
-require "os"
 require "extend/pathname"
 
 require "extend/module"
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -7,6 +7,7 @@ require "dependency_collector"
 require "utils/bottles"
 require "patch"
 require "compilers"
+require "global"
 require "os/mac/version"
 
 class SoftwareSpec
@@ -325,8 +326,8 @@ class Bottle
 end
 
 class BottleSpecification
-  DEFAULT_PREFIX = "/usr/local".freeze
-  DEFAULT_CELLAR = "/usr/local/Cellar".freeze
+  DEFAULT_PREFIX = Homebrew::DEFAULT_PREFIX
+  DEFAULT_CELLAR = "#{DEFAULT_PREFIX}/Cellar".freeze
 
   attr_rw :prefix, :cellar, :rebuild
   attr_accessor :tap
--- a/Library/Homebrew/system_config.rb
+++ b/Library/Homebrew/system_config.rb
@@ -112,9 +112,9 @@ class SystemConfig
         f.puts "Core tap: N/A"
       end
       defaults_hash = {
-        HOMEBREW_PREFIX: "/usr/local",
-        HOMEBREW_REPOSITORY: "/usr/local/Homebrew",
-        HOMEBREW_CELLAR: "/usr/local/Cellar",
+        HOMEBREW_PREFIX: Homebrew::DEFAULT_PREFIX,
+        HOMEBREW_REPOSITORY: "#{Homebrew::DEFAULT_PREFIX}/Homebrew",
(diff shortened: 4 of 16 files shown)


Library/Homebrew/utils/analytics.rb
```diff
@@ -11,7 +11,7 @@ def clear_os_prefix_ci
       def os_prefix_ci
         @os_prefix_ci ||= begin
           os = OS_VERSION
-          prefix = ", non-/usr/local" if HOMEBREW_PREFIX.to_s != "/usr/local"
+          prefix = ", custom-prefix" if HOMEBREW_PREFIX.to_s != HOMEBREW_DEFAULT_PREFIX
```
MikeMcQuaid: This will mean all the previous analytics groupings are broken so probably needs to be tweaked to preserve the existing wording for macOS.

Library/Homebrew/os.rb
```diff
@@ -22,5 +22,7 @@ def self.linux?
     require "os/linux"
     ISSUES_URL = "https://github.com/Linuxbrew/brew/wiki/troubleshooting".freeze
     PATH_OPEN = "xdg-open".freeze
+  else
+    ::HOMEBREW_DEFAULT_PREFIX = "/usr/local".freeze
```
MikeMcQuaid: Feels a bit weird for this to be defined here but not for linux/mac? Does something like `HOMEBREW_DEFAULT_PREFIX ||= "/usr/local".freeze` work in this file and then it could perhaps be omitted on Linux? I dunno the best thing really just this feels odd.

sjackman: Yeah, it felt a bit odd to me to. My best idea was to create an `os/generic` file. I like your idea though.

sjackman:
I've updated references to `/usr/local` excluding comments.
I've not changed `extend/os/mac/` or `test/os/mac/`.
The checks for stray files in `/usr/local` in `diagnostic.rb` are relevant to both Linux and macOS.
I've not changed `extend/ENV/std.rb`, because it's relevant to both Linux and macOS, and Stdenv needs its own PR.

Library/Homebrew/global.rb
```diff
@@ -42,10 +42,15 @@
...
 
 require "fileutils"
+require "os"
+require "os/global"
 
 module Homebrew
   extend FileUtils
 
+  DEFAULT_PREFIX ||= "/usr/local".freeze
+  ANALYTICS_CUSTOM_PREFIX ||= "non-/usr/local".freeze
```
MikeMcQuaid: Can this live in an analytics module?

Library/Homebrew/software_spec.rb
```diff
@@ -325,8 +326,8 @@ def select_download_strategy(specs)
 end
 
 class BottleSpecification
-  DEFAULT_PREFIX = "/usr/local".freeze
-  DEFAULT_CELLAR = "/usr/local/Cellar".freeze
+  DEFAULT_PREFIX = Homebrew::DEFAULT_PREFIX
```
MikeMcQuaid: Maybe scrap this constant?

Library/Homebrew/software_spec.rb
```diff
@@ -325,8 +326,8 @@ def select_download_strategy(specs)
...
+  DEFAULT_CELLAR = "#{DEFAULT_PREFIX}/Cellar".freeze
```
MikeMcQuaid: Tempted to have a `HOMEBREW_DEFAULT_CELLAR`?

Library/Homebrew/system_config.rb
```diff
@@ -112,9 +112,9 @@ def dump_verbose_config(f = $stdout)
         f.puts "Core tap: N/A"
       end
       defaults_hash = {
-        HOMEBREW_PREFIX: "/usr/local",
-        HOMEBREW_REPOSITORY: "/usr/local/Homebrew",
-        HOMEBREW_CELLAR: "/usr/local/Cellar",
+        HOMEBREW_PREFIX: Homebrew::DEFAULT_PREFIX,
+        HOMEBREW_REPOSITORY: "#{Homebrew::DEFAULT_PREFIX}/Homebrew",
```
MikeMcQuaid: Does this need tweaked for Linux?

sjackman: The default repository on Linux is `/home/linuxbrew/.linuxbrew/Homebrew`, so it's good.

Library/Homebrew/software_spec.rb
sjackman: `Linuxbrew/core` is currently using the idiom `HOMEBREW_PREFIX == BottleSpecification. DEFAULT_PREFIX` where Homebrew has been using `HOMEBREW_PREFIX == "/usr/local"`. So for compatibility this constant needs to remain for now. It could possibly be moved to `compat`.

Library/Homebrew/software_spec.rb
sjackman: I see only one use of `DEFAULT_CELLAR` in `software_spec.rb` and one place where it could be used in `system_config.rb`, so to me there doesn't seem to be a need.

Library/Homebrew/cmd/link.rb
```diff
@@ -52,7 +52,7 @@ def link
       end
 
       if keg_only
-        if HOMEBREW_PREFIX.to_s == "/usr/local"
+        if HOMEBREW_PREFIX.to_s == Homebrew::DEFAULT_PREFIX
```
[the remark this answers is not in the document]
sjackman: Good idea. It'll be useful in core too where `HOMEBREW_PREFIX == "/usr/local"` is currently used.

sjackman:
Merged. Thanks for your review, Mike!

lembacon:
Our macOS bottles now all have a `prefix "/usr/local"` set which seems undesired:

- https://github.com/Homebrew/homebrew-core/commit/ebaca957816b4473740ffe1a57b8d6c684c8faf3
- https://github.com/Homebrew/homebrew-core/commit/5de46aaf50a829dc585b5e3164664b1f48913fde
- https://github.com/Homebrew/homebrew-core/commit/6803bce65db0529d2f4852dbdac9671b23608c36
- https://github.com/Homebrew/homebrew-core/commit/0203dd536e88e764544ffec0147d1763e12191b8

This is due to that `brew test-bot --ci-upload` was executed on a Linux machine. 🙃

sjackman:
@BrewTestBot is run on a Linux machine!? Is it a hosted machine? Yes that's certainly undesirable behaviour.

I have a bit of a deadline crunch this week. I suggest reverting this PR, and I'll revisit it next week.

sjackman:
Sorry for breaking BrewTestBot.

PR #4852, opened 2018-09-07
Cask: fixes for quarantining

amyspark:

-----
Apple's Gatekeeper is currently making automated app installation a
nightmare for users. Let's disable this and release all existing
downloads from quarantine. If people still want to use it, let them
know the depth of the rabbit hole.

As part of this, let's also toss in some fixes:
- for non-DMG app installs, where dotfiles or dotfolders are not set as write-able,
- for quarantine users, let's toggle the translocation bit ourselves.

Fixes Homebrew/homebrew-cask#51663, fixes Homebrew/homebrew-cask#51646,
fixes Homebrew/homebrew-cask#51554, fixes Homebrew/homebrew-cask#51691.

### DIFF
--- a/Library/Homebrew/cask/quarantine.rb
+++ b/Library/Homebrew/cask/quarantine.rb
@@ -55,26 +55,58 @@ module Cask
                      print_stderr: false).stdout.rstrip
     end
 
-    def cask(cask: nil, download_path: nil)
+    def disable_translocation!(xattr)
+      fields = xattr.split(";")
+
+      # Fields: status, epoch, download agent, event ID
+      # Let's toggle the app translocation bit, bit 8
+      # http://openradar.me/radar?id=5022734169931776
+
+      fields[0] = (fields[0].to_i(16) | 0x0100).to_s(16).rjust(4, "0")
+
+      fields.join(";")
+    end
+
+    def cask!(cask: nil, download_path: nil, action: true)
       return if cask.nil? || download_path.nil?
 
-      odebug "Quarantining #{download_path}"
+      if action
+        return if detect(download_path)
 
-      quarantiner = system_command(swift,
-                                   args: [
-                                     QUARANTINE_SCRIPT,
-                                     download_path,
-                                     cask.url.to_s,
-                                     cask.homepage.to_s,
-                                   ])
+        odebug "Quarantining #{download_path}"
 
-      return if quarantiner.success?
+        quarantiner = system_command(swift,
+                                    args: [
+                                      QUARANTINE_SCRIPT,
+                                      download_path,
+                                      cask.url.to_s,
+                                      cask.homepage.to_s,
+                                    ])
 
-      case quarantiner.exit_status
-      when 2
-        raise CaskQuarantineError.new(download_path, "Insufficient parameters")
+        return if quarantiner.success?
+
+        case quarantiner.exit_status
+        when 2
+          raise CaskQuarantineError.new(download_path, "Insufficient parameters")
+        else
+          raise CaskQuarantineError.new(download_path, quarantiner.stderr)
+        end
       else
-        raise CaskQuarantineError.new(download_path, quarantiner.stderr)
+        return unless detect(download_path)
+
+        odebug "Releasing #{download_path} from quarantine"
+
+        quarantiner = system_command("/usr/bin/xattr",
+                                    args: [
+                                      "-d",
+                                      QUARANTINE_ATTRIBUTE,
+                                      download_path,
+                                    ],
+                                    print_stderr: false)
+
+        return if quarantiner.success?
+
+        raise CaskQuarantineReleaseError.new(download_path, quarantiner.stderr)
       end
     end
 
@@ -85,9 +117,11 @@ module Cask
 
       odebug "Propagating quarantine from #{from} to #{to}"
 
-      quarantine_status = status(from)
+      quarantine_status = disable_translocation!(status(from))
 
       resolved_paths = Pathname.glob(to/"**/*", File::FNM_DOTMATCH)
+
+      FileUtils.chmod "u+w", resolved_paths
 
       quarantiner = system_command("/usr/bin/xargs",
                                    args: [
--- a/Library/Homebrew/cask/quarantine.rb
+++ b/Library/Homebrew/cask/quarantine.rb
@@ -121,7 +121,7 @@ module Cask
 
       resolved_paths = Pathname.glob(to/"**/*", File::FNM_DOTMATCH)
 
-      FileUtils.chmod "u+w", resolved_paths
+      system_command!("/bin/chmod", args: ["-R", "u+w", to])
 
       quarantiner = system_command("/usr/bin/xargs",
                                    args: [
--- a/Library/Homebrew/cask/quarantine.rb
+++ b/Library/Homebrew/cask/quarantine.rb
@@ -55,7 +55,7 @@ module Cask
                      print_stderr: false).stdout.rstrip
     end
 
-    def disable_translocation!(xattr)
+    def toggle_no_translocation_bit(xattr)
       fields = xattr.split(";")
 
       # Fields: status, epoch, download agent, event ID
@@ -117,7 +117,7 @@ module Cask
 
       odebug "Propagating quarantine from #{from} to #{to}"
 
-      quarantine_status = disable_translocation!(status(from))
+      quarantine_status = toggle_no_translocation_bit(status(from))
 
       resolved_paths = Pathname.glob(to/"**/*", File::FNM_DOTMATCH)
 
--- a/Library/Homebrew/cask/quarantine.rb
+++ b/Library/Homebrew/cask/quarantine.rb
@@ -67,46 +67,46 @@ module Cask
       fields.join(";")
     end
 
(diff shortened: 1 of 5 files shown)


reitermarkus:
So, since using the official APIs is not a valid option, could we actually make this usable by using hard-coded values for the attribute which doesn‘t include the translocation bits?

If not, I don‘t see a point in leaving this in at all.

claui:
> could we actually make this usable by using hard-coded values for the attribute which doesn‘t include the translocation bits?

We can, and I believe that could be the most user-centric solution, and the way for us to proceed.

The [Radar](http://openradar.me/radar?id=5022734169931776) already hints at bit 8 of the xattr value. My cursory tests indicate that this bit may indeed be the _translocation needed_ bit (although my own tests give me `01e1`/`00e1` instead of `0183`/`0083` but that doesn’t affect evidence that bit 8 may be the one).

It’s unlikely that Apple is going to change this private API randomly very often because that would break existing quarantined files.

I’m not a fan of using private APIs but this is one of the rare cases where I feel it’s justified.

MikeMcQuaid:
Personally I'd be in favour of having this as an opt-in security feature. I've had a good experience with using it so far.

MikeMcQuaid:
(to be clear, though, I'm not proposing a change not be made; the Cask maintainers are much better informed than me on making that call)

claui:
> Personally I'd be in favour of having this as an opt-in security feature. I've had a good experience with using it so far.

@MikeMcQuaid You’re referring to the E+D variants? (See table below.)

|  | By default | Optional setting |
|:--|:--|:--|
| **Force quarantine, leave GPR¹ as is ²** | A | B |
| **Force quarantine, force disable GPR ³** | C | D |
| **Do not use Gatekeeper at all** | E | F |


¹ GPR = Gatekeeper Path Randomization

² This is the Apple-endorsed method, and also how browsers do it. Requires user interaction in Finder.

³ This is the private, undocumented technique mentioned above. All we’d need to do is pull up bit 8 in `com.apple.quarantine`’s 16-bit value. If my assumption is correct, we’d basically mimic what Apple’s `DesktopServices` do internally.

___

What I especially like about the **E+D** variants is that there’s minimal impact/risk for us to mess up user experience.

However, I believe it’s more important to offer security out of the box even though this means more bugs/support hassle. That’s why I opt for the **C+F** variants because that’d be probably as secure as our users can get by default (unless we’d force them to drag app icons in Finder at install time, which I feel is a strong no-go).

With that being said, I’m currently not too active in answering Cask-related support incidents so I’ll be cool with whatever our fellow maintainers decide!

amyspark:
@claui Good find! I've coded it in my desktop and can confirm this works. I'll update the pull!

amyspark:
Once Travis reports back, please review. If you are 👍 with this, I can revert the opt-in parts.

claui:
Any relevant cases where you’d say the un`propagate`d artifact would also need the same treatment?

amyspark:
> Any relevant cases where you’d say the unpropagated artifact would also need the same treatment?

@claui, do you mean `Quarantine.cask?` ? That deals only with the downloads, quarantining/releasing them according to what the user is doing. If they don't set `--quarantine`, the download is released (if necessary, to avoid `ditto` and pals propagating quarantine themselves) and `Quarantine.propagate` never acts.

claui:
> do you mean `Quarantine.cask?` ? That deals only with the downloads, quarantining/releasing them according to what the user is doing.

@amyspark Yes, that’s what I meant. LGTM as is!

I’d love feedback from other @Homebrew/cask maintainers though, given I don’t regularly pull my weight when it comes to dealing with support incidents.

Library/Homebrew/cask/installer.rb
```diff
@@ -53,6 +53,27 @@ def self.print_caveats(cask)
...
+          brew cask reinstall #{cask} --no-quarantine
+
+        For more information, please read the following resources:
+
+          https://www.apple.com/business/resources/docs/macOS_Security_Overview.pdf, section "App Security", "Gatekeeper"
+
+          https://www.synack.com/2016/12/16/untranslocating-apps/
+      EOS
+    end
+
```
claui: Is the warning still necessary or helpful to the average user?

Library/Homebrew/cask/cmd/abstract_command.rb
```diff
@@ -12,7 +12,7 @@ class AbstractCommand
       option "--verbose",         :verbose,       false
       option "--outdated",        :outdated_only, false
       option "--require-sha",     :require_sha,   false
-      option "--[no-]quarantine", :quarantine,    true
+      option "--[no-]quarantine", :quarantine,    false
```
claui: Revert all the opt-in changes. (I understand you were planning to do that already.)

Library/Homebrew/cask/quarantine.rb
```diff
@@ -55,26 +55,58 @@ def status(file)
                      print_stderr: false).stdout.rstrip
     end
 
-    def cask(cask: nil, download_path: nil)
+    def disable_translocation(xattr)
+      fields = xattr.split(";")
```
claui: Make that `disable_translocation!`.

Library/Homebrew/cask/quarantine.rb
```diff
@@ -55,26 +55,58 @@ def status(file)
...
+
+      # Fields: status, epoch, download agent, event ID
+      # Let's toggle the app translocation bit, bit 8
+      # http://openradar.me/radar?id=5022734169931776
+
+      fields[0] = (fields[0].to_i(16) | 0x0100).to_s(16).rjust(4, "0")
+
+      fields * ";"
```
claui: Although I love the `*` notation, can you use `fields.join(";")` as that’s much more commonly understood?

Library/Homebrew/cask/quarantine.rb
```diff
@@ -55,26 +55,58 @@ def status(file)
...
+    end
+
+    def cask?(cask: nil, download_path: nil, action: true)
```
claui: Make that `cask!` because the method doesn’t answer a question, and in order to reflect that it makes permanent changes.

Library/Homebrew/cask/installer.rb
amyspark: In light of your translocation bits, no. Will remove.

amyspark:
@claui, I've squashed everything + your comments into a single commit, so as to edit the commit message as well. Feel free to review!

amyspark:
Apologies for the multiple rebases and amends. The `fetch` tests expects files to be untouched, which requires the `--no-quarantine` flag to be 💯  sure.

Library/Homebrew/cask/quarantine.rb
```diff
@@ -55,26 +55,58 @@ def status(file)
                      print_stderr: false).stdout.rstrip
     end
 
-    def cask(cask: nil, download_path: nil)
+    def disable_translocation!(xattr)
```
reitermarkus: Maybe name this `delete_translocation_bit`.

Library/Homebrew/cask/quarantine.rb
```diff
@@ -55,26 +55,58 @@ def status(file)
...
+      fields[0] = (fields[0].to_i(16) | 0x0100).to_s(16).rjust(4, "0")
+
+      fields.join(";")
+    end
+
+    def cask!(cask: nil, download_path: nil, action: true)
       return if cask.nil? || download_path.nil?
 
-      odebug "Quarantining #{download_path}"
+      if action
```
reitermarkus: What does `action` mean? Given that this function is one big if-statement, this should probably be split into two functions with better names.

Library/Homebrew/cask/quarantine.rb
claui: `delete_translocation_bit` would be misleading because it’s actually set, not cleared. `set_translocation_bit` may also be misleading because it’s really a do-not-translocate bit.

@reitermarkus I’d be cool with `set_no_translocation_bit!`, in case you’d prefer that over `disable_translocation!`.

reitermarkus: Ok, `set_no_translocation_bit` then. (Without the `!`, because it doesn't really do anything destructive in the idiomatic Ruby sense.)

amyspark: Rubocop complains:
```
Offenses:

Library/Homebrew/cask/quarantine.rb:58:9: C: Do not prefix writer method names with set_.
    def set_no_translocation_bit(xattr)
        ^^^^^^^^^^^^^^^^^^^^^^^^

690 files inspected, 1 offense detected
```

I've changed it to `toggle_no_translocation_bit`, please let me know if you're fine with this.

claui:
Thanks @amyspark!

Library/Homebrew/cask/quarantine.rb
reitermarkus: `toggle` is also misleading. It's fine to explicitly disable RuboCop for this line because it's not a writer method.

MikeMcQuaid:
> What I especially like about the E+D variants is that there’s minimal impact/risk for us to mess up user experience.

@claui Sorry for late reply but: 👍 

Nice work @amyspark

PR #4862, opened 2018-09-08
make dependency_order_cop a regular audit

commitay:

-----
As we're trying to enforce this in PRs, let's make it a regular audit. Currently ≈ 900 formulae need changes to pass this audit, I'll submit the necessary changes to `core` if people are 👍.

### DIFF
--- a/Library/Homebrew/rubocops/dependency_order_cop.rb
+++ b/Library/Homebrew/rubocops/dependency_order_cop.rb
@@ -2,7 +2,7 @@ require "rubocops/extend/formula_cop"
 
 module RuboCop
   module Cop
-    module NewFormulaAudit
+    module FormulaAudit
       # This cop checks for correct order of `depends_on` in a Formula
       #
       # precedence order:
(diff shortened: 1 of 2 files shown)


colindean:
s/order/whatever I meant I don't remember parliamentary procedure well enough at this hour/g

commitay:
I'm 👎 on making this strict as it isn't enforced by CI and the autocorrect will make a mess of the existing comments and whitespace that we'll then have to deal with in contributor PRs.

DomT4:
Aye, my line of thought was that we ask people to run `audit --strict` on human-filed PRs, but not command-filed PRs, so we'd reduce the level of violations in core over time without having to jump from almost never enforcing it to absolutely always enforcing it.

For better or worse any change we make here is going to whack external taps as well, if they're using Travis/etc to run CI on themselves, which isn't super-uncommon AFAIK. I'm just a bit jumpy about touching 900 formulae at the same time. That's going to be a huge diff, which introduces the risk of missing any erroneous changes.

I'm not necessarily opposed, just... nervous 😄.

commitay:
> That's going to be a huge diff, which introduces the risk of missing any erroneous changes.

> I'm not necessarily opposed, just... nervous 😄.

Lol. I'm not planning on doing all of them at once. Maybe 50 for so per PR to limit the diff, check everything is fixed and then merge this.

commitay:
TBH, I would like to make this strict and let the community submit fixes but as the autocorrect changes aren't always ... correct, we would need to check more things manually and would also make for very nitpicky PR reviews.

commitay:
https://github.com/Homebrew/brew/pull/4863#issuecomment-421503618

PR #4863, opened 2018-09-09
components_order_cop: add all components

commitay:

-----
Follow up to https://github.com/Homebrew/brew/pull/4113

I said in https://github.com/Homebrew/brew/pull/4113#issuecomment-385208377: 
> Adding them individually would be a bit easier to do (and review)

They actually aren't, adding them individually means that some formulae end up needing multiple changes.

Currently ≈ ~~110~~ 130 formulae need changes to pass audit with the components in this order.

This seems to the the "standard" order for these components but I can change it if this isn't the preferred order.

I've probably missed one or two of the less common components as well, shout at me and I'll add them. 

I think we should do this now so it's easier expand this cop to include `stable`/`devel`/`head` blocks later. The same applies if we're going to add more components (IIRC moving `livecheck` to formulae was mentioned.) 

I'll submit the necessary changes to core if people are 👍.
___
https://github.com/Homebrew/homebrew-core/blob/master/Formula/php.rb#L297-L299
```ruby
  def php_version
    version.to_s.split(".")[0..1].join(".")
  end
```

Just as a note, this cop doesn't handle stuff like this, so we probably need an generic `method_definition` sort as well, but that's another PR.


commitay:
Options being removed from formulae is making the changes that would be needed for this and https://github.com/Homebrew/brew/pull/4862 unnecessary for some formulae as they are either fixed or removed.

Leaving this and https://github.com/Homebrew/brew/pull/4862 on hold while options removal is ongoing.

MikeMcQuaid:
@commitay Works for me 👍

MikeMcQuaid:
@commitay For what it's worth I think a mass cleanup PR where you `brew audit --fix`, check the diff yourself and then merge without it going through CI would be reasonable if you felt like that approach. If not: no worries! Thanks for doing this ❤️

PR #4871, opened 2018-09-11
Don’t add error when changing checksum to `:no_check`.

reitermarkus:

---

Fixes the CI error here: https://github.com/Homebrew/homebrew-cask/pull/51795

### DIFF
--- a/Library/Homebrew/cask/audit.rb
+++ b/Library/Homebrew/cask/audit.rb
@@ -133,6 +133,8 @@ module Cask
     end
 
     def check_version_and_checksum
+      return if cask.sha256 == :no_check
+
       return if @cask.sourcefile_path.nil?
 
       tap = @cask.tap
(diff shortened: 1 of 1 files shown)


reitermarkus:
Should changing from `:no_check` still be an error?

claui:
> Should changing from `:no_check` still be an error?

I think it didn’t really matter until https://github.com/Homebrew/homebrew-cask/pull/51795 but we should definitely allow it from now on.

Library/Homebrew/cask/audit.rb
```diff
@@ -133,6 +133,8 @@ def check_required_stanzas
     end
 
     def check_version_and_checksum
+      return if cask.sha256 == :no_check
```
claui: Do we really want to allow `:no_check` for any constellation? Or just for documented cases of known-volatile SHAs, such as https://github.com/Homebrew/homebrew-cask/pull/51795?

reitermarkus: I think we *already* allow it. This is just to skip the check when the checksum changes.

claui: > I think we already allow it.

Indeed. TIL!

Issue #4854, opened 2018-09-07
Regression: formula .pax.gz sources are no longer extracted 


benmwebb:
A number of formulae that I maintain extract their contents from `.pax.gz` files. This used to work fine (although it's not clear to me if it's a supported use case) but recently stopped working. A quick `git bisect` shows that the commit that broke it was 4c80f827a.

Steps to reproduce:

`brew create testformula` and give it the contents
```
require 'formula'
  
class Testformula < Formula
  url 'https://salilab.org/modeller/9.20/modeller-9.20-mac.pax.gz'
  sha256 '161c8d5576a52ead4a90d775549cbc772c33cba71500be33c5e21af75da33729'

  def install
    doc.install "Library/modeller-9.20/ChangeLog"
  end
end
```

`brew update && brew reinstall testformula` then fails with
`Errno::ENOENT: No such file or directory - Library/modeller-9.20/ChangeLog`

(Looking in the build directory, brew downloaded and gunzipped the `.pax.gz` file, but did not extract it, hence future operations on the files from that archive failed.)

`cd /usr/local/Homebrew/ && git checkout f8dc9ef && brew reinstall testformula` successfully installs the formula.

`cd /usr/local/Homebrew/ && git checkout 4c80f827a && brew reinstall testformula` fails in the same way as current master.

If not supporting `.pax.gz` is intentional, please let me know - I can have my formulae explicitly run `pax` in the `install` step (just seems odd since it used to work).

FWIW, the complete formulae that I maintain with this problem are `modeller.rb` and `ifort-runtime.rb` at https://github.com/salilab/homebrew-salilab/

**Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

- [x] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew update` and can still reproduce the problem?
[...]

PR #4873, opened 2018-09-11
Fix extracting `.pax` archives.

reitermarkus:

-----

Add `Pax` unpack strategy.

Fixes https://github.com/Homebrew/brew/issues/4854.


reitermarkus:
> What formulae, new or existing, use it?

None in Homebrew/core, but this was working before all of the refactoring since a `.pax.gz` would be detected with `tar -tf`, now we only use the extension.

colindean:
Thanks for the clarification!

PR #4875, opened 2018-09-11
Elide prefix "/usr/local" in bottle block

sjackman:
brew test-bot --ci-upload is run on a Linux machine.
The macOS bottles have a prefix of `/usr/local`.
The default prefix on Linux is `/home/linuxbrew/.linuxbrew`.
Elide `prefix "/usr/local"` in the bottle block, even it does not match the default prefix. Ditto for cellar.

-----

### DIFF
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -44,12 +44,12 @@ BOTTLE_ERB = <<-EOS.freeze
     <% if !root_url.start_with?(HOMEBREW_BOTTLE_DEFAULT_DOMAIN) %>
     root_url "<%= root_url %>"
     <% end %>
-    <% if prefix != BottleSpecification::DEFAULT_PREFIX %>
+    <% if ![Homebrew::DEFAULT_PREFIX, "/usr/local"].include?(prefix) %>
     prefix "<%= prefix %>"
     <% end %>
     <% if cellar.is_a? Symbol %>
     cellar :<%= cellar %>
-    <% elsif cellar != BottleSpecification::DEFAULT_CELLAR %>
+    <% elsif ![Homebrew::DEFAULT_CELLAR, "/usr/local/Cellar"].include?(cellar) %>
     cellar "<%= cellar %>"
     <% end %>
     <% if rebuild.positive? %>
--- a/Library/Homebrew/global.rb
+++ b/Library/Homebrew/global.rb
@@ -49,6 +49,8 @@ module Homebrew
   extend FileUtils
 
   DEFAULT_PREFIX ||= "/usr/local".freeze
+  DEFAULT_CELLAR = "#{DEFAULT_PREFIX}/Cellar".freeze
+  DEFAULT_REPOSITORY = "#{DEFAULT_PREFIX}/Homebrew".freeze
 
   class << self
     attr_writer :failed, :raise_deprecation_exceptions, :auditing, :args
--- a/Library/Homebrew/software_spec.rb
+++ b/Library/Homebrew/software_spec.rb
@@ -327,7 +327,6 @@ end
 
 class BottleSpecification
   DEFAULT_PREFIX = Homebrew::DEFAULT_PREFIX
-  DEFAULT_CELLAR = "#{DEFAULT_PREFIX}/Cellar".freeze
 
   attr_rw :prefix, :cellar, :rebuild
   attr_accessor :tap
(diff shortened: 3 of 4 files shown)


Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -44,12 +44,12 @@
     <% if !root_url.start_with?(HOMEBREW_BOTTLE_DEFAULT_DOMAIN) %>
     root_url "<%= root_url %>"
     <% end %>
-    <% if prefix != BottleSpecification::DEFAULT_PREFIX %>
+    <% if ![Homebrew::DEFAULT_PREFIX, "/usr/local"].include?(prefix) %>
```
jonchang: What's the rationale for moving away from `BottleSpecification::DEFAULT_PREFIX` to `Homebrew::DEFAULT_PREFIX`?

Library/Homebrew/software_spec.rb
```diff
@@ -327,16 +327,15 @@ def select_download_strategy(specs)
 
 class BottleSpecification
   DEFAULT_PREFIX = Homebrew::DEFAULT_PREFIX
-  DEFAULT_CELLAR = "#{DEFAULT_PREFIX}/Cellar".freeze
```
jonchang: I don't believe that `BottleSpecification::DEFAULT_PREFIX` is used anywhere else, so it may make sense to just drop this.

Library/Homebrew/software_spec.rb
```diff
@@ -327,16 +327,15 @@ def select_download_strategy(specs)
 
 class BottleSpecification
   DEFAULT_PREFIX = Homebrew::DEFAULT_PREFIX
```
JCount: It seems this constant is no longer used? However, it may not be worth sending this through travis again to remove it right now.

Library/Homebrew/global.rb
```diff
@@ -49,6 +49,8 @@ module Homebrew
   extend FileUtils
 
   DEFAULT_PREFIX ||= "/usr/local".freeze
+  DEFAULT_CELLAR = "#{DEFAULT_PREFIX}/Cellar".freeze
+  DEFAULT_REPOSITORY = "#{DEFAULT_PREFIX}/Homebrew".freeze
```
JCount: Just to be certain, the use of `=` over `||=` is deliberate here? (I expect it is.)

Library/Homebrew/software_spec.rb
JCount: The only place that I believe this was used was `Library/Homebrew/dev-cmd/bottle.rb` as `BottleSpecification::DEFAULT_PREFIX`. I believe that was the only other usage outside this file.

Library/Homebrew/software_spec.rb
sjackman: `Linuxbrew/core` is currently using the idiom `HOMEBREW_PREFIX == BottleSpecification::DEFAULT_PREFIX` where Homebrew has been using `HOMEBREW_PREFIX == "/usr/local"`. So for compatibility this constant needs to remain for now. It could possibly be moved to `compat`.

See https://github.com/Homebrew/brew/pull/4851#discussion_r216400554

Library/Homebrew/software_spec.rb
sjackman: `Linuxbrew/core` is currently using the idiom `HOMEBREW_PREFIX == BottleSpecification::DEFAULT_PREFIX` where Homebrew has been using `HOMEBREW_PREFIX == "/usr/local"`. So for compatibility this constant needs to remain for now. It could possibly be moved to `compat`.

See https://github.com/Homebrew/brew/pull/4851#discussion_r216400554

Library/Homebrew/global.rb
sjackman: Yes, it is. The use of `||=` is due to the definition of this same variable at
https://github.com/Homebrew/brew/blob/7e63dfb55656591dfeb5d0b98314090f81ed3304/Library/Homebrew/os/linux/global.rb#L2

Library/Homebrew/dev-cmd/bottle.rb
sjackman: The intention of https://github.com/Homebrew/brew/pull/4851 was to move from `BottleSpecification::DEFAULT_PREFIX` to `Homebrew::DEFAULT_PREFIX`, but I missed this one.

PR #4892, opened 2018-09-12
dev-cmd/bottle: fix --keep-old with --root-url

### LINKED ISSUE #4554: `brew bottle --merge --keep-old` broken if root_url specified

- [ ] ran a `brew` command and reproduced the problem with multiple formulae?
- [ ] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [ ] ran `brew config` and `brew doctor` and included their output with your issue?

Skipped some of these the bug is fairly clear from the source of homebrew itself:

To reproduce, build a bottle with a root url including URI scheme, and with `--json`. Then try to use `brew bottle --merge --json`.

This will fail, as the `:` gets removed from the old root url here: https://github.com/Homebrew/brew/blob/b993c7745ef7dba9e874ca0888f177adc7969fb7/Library/Homebrew/dev-cmd/bottle.rb#L450

As can be demonstrated by:

```
irb(main):003:0> '"https://example.com"'.delete ":\""
=> "https//example.com"
```

... and can be verified by inserting `p` statements here: https://github.com/Homebrew/brew/blob/b993c7745ef7dba9e874ca0888f177adc7969fb7/Library/Homebrew/dev-cmd/bottle.rb#L465

`--keep-old` will fail as `https//example.com` !== `https://example.com`

scpeters:

-----
Don't strip the : from root_url when using --keep-old.
Fixes #4554.

### DIFF
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -448,7 +448,8 @@ module Homebrew
                 valid_key = %w[root_url prefix cellar rebuild sha1 sha256].include? key
                 next unless valid_key
 
-                old_value = old_value_original.to_s.delete ":'\""
+                old_value = old_value_original.to_s.delete "'\""
+                old_value = old_value.to_s.delete ":" if key != "root_url"
                 tag = tag.to_s.delete ":"
 
                 unless tag.empty?


MikeMcQuaid:
Thanks @scpeters! Just for future reference if I ✅ a PR I'll let you merge it yourself.

Issue #4763, opened 2018-08-27
Improve Homebrew/brew code coverage


MikeMcQuaid:
It would be good to increase [our code coverage](https://codecov.io/gh/Homebrew/brew/tree/master/Library/Homebrew). This can be done by adding integration tests (only) for missing `cmd` and `dev-cmd` commands (e.g. https://codecov.io/gh/Homebrew/brew/src/master/Library/Homebrew/cmd/diy.rb), figuring out why some coverage is currently missing (e.g. https://codecov.io/gh/Homebrew/brew/src/master/Library/Homebrew/dev-cmd/ruby.rb despite https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/dev-cmd/ruby_spec.rb) and adding unit tests for files with low coverage (e.g. https://codecov.io/gh/Homebrew/brew/src/master/Library/Homebrew/hardware.rb).

reitermarkus:
@alexbostock, many small PRs are easier to review than a huge one.

MikeMcQuaid:
I think this is now good enough (>70%) that we don't need an open issue.

PR #4915, opened 2018-09-16
Improve code coverage

alexbostock:

-----

I've added tests for a few commands to improve code coverage (issue #4763).

### DIFF
--- a/Library/Homebrew/test/cmd/commands_spec.rb
+++ b/Library/Homebrew/test/cmd/commands_spec.rb
@@ -11,7 +11,6 @@ describe "brew commands", :integration_test do
   end
 
   it "prints a list without headers with the --quiet flag" do
-
     expect { brew "commands", "--quiet" }
       .to be_a_success
       .and not_to_output.to_stderr
--- a/Library/Homebrew/test/cmd/commands_spec.rb
+++ b/Library/Homebrew/test/cmd/commands_spec.rb
@@ -6,6 +6,12 @@ describe "brew commands", :integration_test do
   it "prints a list of all available commands" do
     expect { brew "commands" }
       .to output(/Built-in commands/).to_stdout
+      .and not_to_output.to_stderr
+      .and be_a_success
+
+    expect { brew "commands", "--quiet" }
+      .to output.to_stdout
+      .and not_to_output.to_stderr
       .and be_a_success
   end
 end
--- a/Library/Homebrew/test/cmd/commands_spec.rb
+++ b/Library/Homebrew/test/cmd/commands_spec.rb
@@ -8,11 +8,15 @@ describe "brew commands", :integration_test do
       .to output(/Built-in commands/).to_stdout
       .and not_to_output.to_stderr
       .and be_a_success
+  end
+
+  it "prints a list without headers with the --quiet flag" do
+    expect { brew "commands", "--quiet" }
+      .to_not output(/Built-in commands/).to_stdout
 
     expect { brew "commands", "--quiet" }
-      .to output.to_stdout
+      .to be_a_success
       .and not_to_output.to_stderr
-      .and be_a_success
   end
 end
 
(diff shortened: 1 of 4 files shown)


Library/Homebrew/test/cmd/commands_spec.rb
```diff
@@ -6,6 +6,12 @@
   it "prints a list of all available commands" do
     expect { brew "commands" }
       .to output(/Built-in commands/).to_stdout
+      .and not_to_output.to_stderr
+      .and be_a_success
+
+    expect { brew "commands", "--quiet" }
```
reitermarkus: This should be a separate test and should check that the output doesn't include `Built-in commands`.

Library/Homebrew/test/cmd/info_spec.rb
```diff
@@ -8,6 +8,11 @@
       .to output(/testball: stable 0.1/).to_stdout
       .and not_to_output.to_stderr
       .and be_a_success
+
+    expect { brew "info", "testball", "--json=v1" }
```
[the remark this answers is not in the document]
alexbostock: I can't find an elegant way to parse the output (other than with regex), unless you think it should redirect stdout?

Library/Homebrew/test/cmd/commands_spec.rb
```diff
@@ -6,8 +6,18 @@
...
       .to output(/Built-in commands/).to_stdout
+      .and not_to_output.to_stderr
       .and be_a_success
   end
+
+  it "prints a list without headers with the --quiet flag" do
+    expect { brew "commands", "--quiet" }
+      .to_not output(/Built-in commands/).to_stdout
+
+    expect { brew "commands", "--quiet" }
```
reitermarkus: Can you also connect these with `.and`?

Library/Homebrew/test/cmd/info_spec.rb
```diff
@@ -1,14 +1,23 @@
...
+  it "prints information about a given Formula" do
     expect { brew "info", "testball" }
       .to output(/testball: stable 0.1/).to_stdout
       .and not_to_output.to_stderr
       .and be_a_success
   end
+
+  it "prints as json with the --json=v1 flag" do
+    expect { brew "info", "testball", "--json=v1" }
+      .to output(/\{.+testball.+\}/).to_stdout
```
reitermarkus: You can add a custom `a_json_string` matcher to check if this is parsable.

reitermarkus:
Thanks, @alexbostock! Great work again. 👍

MikeMcQuaid:
Sorry that I'm late to reviewing this. I just noticed this adds a bunch more integration tests which unfortunately isn't desirable as it slows down the test suite. My wording maybe wasn't clear in https://github.com/Homebrew/brew/issues/4763 but:

> This can be done by adding integration tests (only) for missing cmd and dev-cmd commands (e.g. https://codecov.io/gh/Homebrew/brew/src/master/Library/Homebrew/cmd/diy.rb)

What I was trying to say is that commands that already have integration tests (e.g. `brew info`) should prioritise unit tests rather than integration tests. This is because integration tests are pretty dramatically slower than unit tests.

Thoughts (including @reitermarkus)?

alexbostock:
@MikeMcQuaid, sorry about that. Please could you clarify what you mean by integration tests and unit tests?

MikeMcQuaid:
@alexbostock An integration test is one which has `describe "brew ...", :integration_test do` at the top of the file. These are executed by shelling out to a new Ruby process (which is slow). A unit test is any other one that is not an integration test (for the same of this discussion)

alexbostock:
Thank you, @MikeMcQuaid. I'll bear that in mind in future. Should I go back and change what I've done in this PR?

MikeMcQuaid:
@alexbostock That would be great, yes 🙇

PR #4924, opened 2018-09-17
bin/brew: introduce `HOMEBREW_PRESERVED_ENVS` for an ENV whitelist

lembacon:

-----

This provides an alternative approach to make it possible to keep some environment variables unfiltered. Setting the undocumented `HOMEBREW_NO_ENV_FILTERING` may cause more unintended side effects than expected, e.g. when `GEM_PATH` and/or `GEM_HOME` are set.

For example, if someone does not want Homebrew's Python to write `.pyc` files during `brew install --build-from-source` or `brew postinstall`, then `HOMEBREW_PRESERVED_ENVS =PYTHONDONTWRITEBYTECODE` and `PYTHONDONTWRITEBYTECODE=1` could be set. Comparing to setting the undocumented `HOMEBREW_NO_ENV_FILTERING`, this could be a safer approach.

### DIFF
--- a/Library/Homebrew/manpages/brew.1.md.erb
+++ b/Library/Homebrew/manpages/brew.1.md.erb
@@ -244,6 +244,10 @@ Note that environment variables must have a value set to be detected. For exampl
     If set, Homebrew will not use the GitHub API for e.g searches or
     fetching relevant issues on a failed install.
 
+  * `HOMEBREW_PRESERVED_ENVS`:
+    A colon separated list of environment variables denoting not to be
+    filtered by Homebrew.
+
   * `HOMEBREW_PRY`:
     If set, Homebrew will use `pry` for the `brew irb` command.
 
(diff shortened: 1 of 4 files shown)


MikeMcQuaid:
@lembacon The supported way for passing through arbitrary variables is through setting `HOMEBREW_SOME_VAR` which the formula can set to `SOME_VAR` if desired. I'm a very strong 👎 on allowing people to set random variables of their choosing; this defeats the point of the environment filtering feature because people will blindly copy-paste instructions to set this variable which applies to every formula even if only needed by a single one. A better solution is one that can be done on a single formula as desired.

> PYTHONDONTWRITEBYTECODE=1

This could be a dedicated, documented `HOMEBREW_*` variable we could set. If it's a better default we could consider setting this for our bottles, too.

lembacon:
That's fair enough. Closing this for now. 😅

PR #4937, opened 2018-09-19
cleanup: fix spurious detection of stale cache due to naming conflicts

lembacon:

-----

If a formula and a cask share a common name, `brew cleanup` may report false positives.

For example, if I have the formula `flow` installed, `brew cleanup --dry-run` will always report that it would remove `$HOMEBREW_CACHE/flow--0.81.0.mojave.bottle.tar.gz`, as there is also a cask named `flow`.


MikeMcQuaid:
`brew tests` seem to be legitimately failing.

PR #4941, opened 2018-09-19
super/cc: further bump -mmacosx-version-min on Mojave

fxcoudert:
This is a follow-up to #4810, motivated by this comment: 
https://github.com/Homebrew/brew/pull/4810#issuecomment-422944940

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -17,8 +17,8 @@ def mac?
   RUBY_PLATFORM[/darwin/]
 end
 
-def mojave_or_later?
-  mac? && RUBY_PLATFORM[/darwin\d+/][/\d+/].to_i >= 18
+def high_sierra_or_later?
+  mac? && RUBY_PLATFORM[/darwin\d+/][/\d+/].to_i >= 17
 end
 
 def linux?
@@ -187,7 +187,7 @@ class Cmd
       # clang doesn't support these flags
       args << arg unless tool =~ /^clang/
     when /-mmacosx-version-min=10\.(\d)/
-      arg = "-mmacosx-version-min=10.6" if mojave_or_later? && $1.to_i < 6
+      arg = "-mmacosx-version-min=10.9" if high_sierra_or_later? && $1.to_i < 9
       args << arg
     when "--fast-math"
       arg = "-ffast-math" if tool =~ /^clang/


fxcoudert:
@MikeMcQuaid I'm not authorised to merge :)

fxcoudert:
I have just found out that:

> The clang toolchain in Xcode 10 on High Sierra (the so-called CLT) still supports the old features (32 bit binaries, libgcc_s, etc.) but not the SDK. Which explains why some installs may work or not, depending on how they are set up and what combination of Xcode/CLT you have installed.

So it looks like it would be better to make the substitution based on clang/Xcode version. I believe it is safe to make the substitution with anything that is Xcode 10 or more.

fxcoudert:
I have just found out that:

> The clang toolchain in Xcode 10 on High Sierra (the so-called CLT) still supports the old features (32 bit binaries, libgcc_s, etc.) but not the SDK. Which explains why some installs may work or not, depending on how they are set up and what combination of Xcode/CLT you have installed.

So it looks like it would be better to make the substitution based on clang/Xcode version. I believe it is safe to make the substitution with anything that is Xcode 10 or more. I am not sure how to check for Xcode/clang version in that file, though :(

fxcoudert:
Changing unconditionally on High Sierra and later; I don't think it can hurt.

fxcoudert:
So, this is working locally, but not on CI.

Locally, macOS 10.14, Xcode 10:

```
einstein:~ fx$ brew install -s libmodplug
==> Downloading https://downloads.sourceforge.net/modplug-xmms/libmodplug/0.8.9.
Already downloaded: /Users/fx/Library/Caches/Homebrew/downloads/d28b71ceba01597872c349cf1db2ecd0cbf64e567b66ab30ab603d66e357fe50--libmodplug-0.8.9.0.tar.gz
==> ./configure --disable-silent-rules --prefix=/usr/local/Cellar/libmodplug/0.8
==> make install
🍺  /usr/local/Cellar/libmodplug/0.8.9.0: 15 files, 342.2KB, built in 27 seconds
einstein:~ fx$ brew rm libmodplug
Uninstalling /usr/local/Cellar/libmodplug/0.8.9.0... (15 files, 342.2KB)
einstein:~ fx$ brew install -s --build-bottle libmodplug
==> Downloading https://downloads.sourceforge.net/modplug-xmms/libmodplug/0.8.9.
Already downloaded: /Users/fx/Library/Caches/Homebrew/downloads/d28b71ceba01597872c349cf1db2ecd0cbf64e567b66ab30ab603d66e357fe50--libmodplug-0.8.9.0.tar.gz
==> ./configure --disable-silent-rules --prefix=/usr/local/Cellar/libmodplug/0.8
==> make install
==> Not running post_install as we're building a bottle
You can run it manually using `brew postinstall libmodplug`
🍺  /usr/local/Cellar/libmodplug/0.8.9.0: 15 files, 342.3KB, built in 25 seconds
```

On CI, even with this pull request merged:

```
10:24:09 ==> brew prune
10:24:10 Homebrew/brew 1.7.5-44-g2901c3679 (Merge pull request #4941 from fxcoudert/mojave)
10:24:10 homebrew/core origin/master 4910e9426f (fix graphviz head build (#32264))
10:24:10 homebrew/core HEAD 4910e9426f (fix graphviz head build (#32264))
```

it still fails with:

```
10:25:55 libtool: link: clang++ -dynamiclib -Wl,-undefined -Wl,dynamic_lookup -o .libs/libmodplug.1.dylib  .libs/sndmix.o .libs/sndfile.o .libs/snd_fx.o .libs/snd_flt.o .libs/snd_dsp.o .libs/fastmix.o .libs/mmcmp.o .libs/load_xm.o .libs/load_wav.o .libs/load_umx.o .libs/load_ult.o .libs/load_stm.o .libs/load_s3m.o .libs/load_ptm.o .libs/load_okt.o .libs/load_mtm.o .libs/load_mod.o .libs/load_med.o .libs/load_mdl.o .libs/load_it.o .libs/load_far.o .libs/load_dsm.o .libs/load_dmf.o .libs/load_dbm.o .libs/load_ams.o .libs/load_amf.o .libs/load_669.o .libs/load_j2b.o .libs/load_mt2.o .libs/load_psm.o .libs/load_abc.o .libs/load_mid.o .libs/load_pat.o .libs/modplug.o   -lm  -g -O2 -mmacosx-version-min=10.5 -mmacosx-version-min=10.5   -install_name  /usr/local/Cellar/libmodplug/0.8.9.0/lib/libmodplug.1.dylib -compatibility_version 2 -current_version 2.0 -Wl,-single_module
10:25:55 ld: library not found for -lstdc++
```

I have no idea how to dig further.

MikeMcQuaid:
@fxcoudert Is there an Xcode/Clang version difference? Could be an earlier beta in CI?

fxcoudert:
I don't think any of the betas had issues with `-mmacosx-version-min=10.9`. I'm more thinking like some mechanism is preventing the superenv to work on CI? Like it has the wrong RUBY_PLATFORM or something?

Otherwise, is there a way to get the log files out of CI? See `02.make.cc` and figure out what superenv is doing…

fxcoudert:
That's it! On CI machine, `RUBY_PLATFORM` returns `universal.x86_64-darwin9.0`. I have no idea why it thinks it's a Leopard, though…

MikeMcQuaid:
@fxcoudert Ah, interesting; means we can't rely on `RUBY_PLATFORM` as on portable Ruby it's going to be wrong.

fxcoudert:
@MikeMcQuaid apparently RUBY_PLATFORM is defined at build time, based on the target triplet. So definitely not the right thing to use here :(

chdiza:
> How about we just run sw_vers -productVersion?

In my experience `uname -r` is considerably faster than `sw_vers`.  It gives a Darwin version, but it's easy to use that to calculate the macOS version.  (The Darwin minor number is 4 ahead of the macOS minor number).

PR #4995, opened 2018-09-28
Default compiler is gcc on Linux

sjackman:

-----

`DevelopmentTools.default_compiler` is already overridden on macOS to be `clang` (usually).
https://github.com/Homebrew/brew/blob/dae47914ca383e0ed5192436b9e7e11e0bf2640e/Library/Homebrew/extend/os/mac/development_tools.rb#L28

### DIFF
--- a/Library/Homebrew/development_tools.rb
+++ b/Library/Homebrew/development_tools.rb
@@ -99,7 +99,7 @@ class DevelopmentTools
         path = HOMEBREW_PREFIX/"opt/gcc/bin"/cc
         path = locate(cc) unless path.exist?
         version = if path &&
-                     build_version = `#{path} --version`[/gcc(?:-\d(?:\.\d)? \(.+\))? (\d\.\d\.\d)/, 1]
+                     build_version = `#{path} --version`[/gcc(?:(?:-\d(?:\.\d)?)? \(.+\))? (\d\.\d\.\d)/, 1]
           Version.new build_version
         else
           Version::NULL
--- a/Library/Homebrew/extend/ENV/shared.rb
+++ b/Library/Homebrew/extend/ENV/shared.rb
@@ -319,7 +319,11 @@ module SharedEnvExtension
   def compiler_with_cxx11_support?(cc)
     return if compiler_any_clang?(cc)
 
-    version = cc[/^gcc-(\d+(?:\.\d+)?)$/, 1]
+    version = if cc == :gcc
+      non_apple_gcc_version "gcc"
+    else
+      cc[/^gcc-(\d+(?:\.\d+)?)$/, 1]
+    end
     version && Version.create(version) >= Version.create("4.8")
   end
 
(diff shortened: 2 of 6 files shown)


Library/Homebrew/development_tools.rb
```diff
@@ -34,7 +34,7 @@ def default_cc
     end
 
     def default_compiler
-      :clang
+      :gcc
```
MikeMcQuaid: I'd rather this was left as `clang` and changed just for Linux.

Library/Homebrew/extend/ENV/shared.rb
```diff
@@ -319,7 +319,11 @@ def compiler_any_clang?(cc = compiler)
   def compiler_with_cxx11_support?(cc)
     return if compiler_any_clang?(cc)
 
-    version = cc[/^gcc-(\d+(?:\.\d+)?)$/, 1]
+    version = if cc == :gcc
+      non_apple_gcc_version "gcc"
+    else
+      cc[/^gcc-(\d+(?:\.\d+)?)$/, 1]
```
MikeMcQuaid: What breaks with this version?

Library/Homebrew/test/compiler_selector_spec.rb
```diff
@@ -29,7 +29,7 @@
   describe "#compiler" do
     it "raises an error if no matching compiler can be found" do
       software_spec.fails_with(:clang)
-      software_spec.fails_with(:gcc)
+      software_spec.fails_with(:gcc_4_2)
```
MikeMcQuaid: I'm not sure about this; this will change the mean for existing formula that were using `fails_with :gcc`, no?

Library/Homebrew/development_tools.rb
sjackman: Can do. Since `clang` is the default on macOS`, this change affects only `--generic`.

sjackman:
> Unfortunately I'm a bit concerned this is going to result in a change of behaviour on macOS (particularly on older, tap formulae).

`gcc` is not currently a valid value for the compiler. The old compiler option `gcc` was renamed to `gcc-4.2`.

```
❯❯❯ brew install hello --cc=gcc
Error: An exception occured within a child process:
  RuntimeError: Invalid value for --cc: gcc
❯❯❯ brew install hello --cc=gcc-4.2
==> Downloading https://ftp.gnu.org/gnu/hello/hello-2.10.tar.gz
…
```

> I'm not sure about this; this will change the mean for existing formula that were using fails_with :gcc, no?

`fails_with :gcc` is a vestigial DSL, that has no effect, just as `fails_with :foobar` also has no effect.

sjackman:
I've opened PR https://github.com/Homebrew/homebrew-core/pull/32583 to use `fails_with :gcc_4_2` rather than `fails_with :gcc` in Homebrew/core. The latter statement currently has no effect.

sjackman:
Merged. Thanks for the review, Mike.

PR #5000, opened 2018-09-29
Don’t treat `.7z` as part of a version.

reitermarkus:

-----

Fixes https://github.com/Homebrew/homebrew-core/issues/32418.

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -200,7 +200,7 @@ class Pathname
     return archive_ext if archive_ext
 
     # Don't treat version numbers as extname.
-    return "" if basename.match?(/\b\d+\.\d+[^\.]*\Z/)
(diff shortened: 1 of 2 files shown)


Library/Homebrew/extend/pathname.rb
```diff
@@ -200,7 +200,7 @@ def extname(path = to_s)
     return archive_ext if archive_ext
 
     # Don't treat version numbers as extname.
-    return "" if basename.match?(/\b\d+\.\d+[^\.]*\Z/)
+    return "" if basename.match?(/\b\d+\.\d+[^\.]*\Z/) && !basename.end_with?(".7z")
```
MikeMcQuaid: Could this be part of the regex?

reitermarkus: Maybe, but not in a readable way I assume.

vitorgalvao: I [agree with @colindean](https://github.com/Homebrew/brew/pull/5000#pullrequestreview-160122055) regarding readability of this one.

PR #5001, opened 2018-09-29
Make all extracted files writable after extracting.

reitermarkus:

-----

Make everything in the temporary directory writable after extracting so that it can be removed by the exit handler.

### DIFF
--- a/Library/Homebrew/unpack_strategy.rb
+++ b/Library/Homebrew/unpack_strategy.rb
@@ -136,6 +136,8 @@ module UnpackStrategy
       end
 
       Directory.new(tmp_unpack_dir).extract(to: to, verbose: verbose)
+
+      FileUtils.chmod_R "+w", tmp_unpack_dir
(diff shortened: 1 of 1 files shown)


amyspark:
@reitermarkus, with this change, perhaps it's possible to remove these lines as well.

https://github.com/Homebrew/brew/blob/dae47914ca383e0ed5192436b9e7e11e0bf2640e/Library/Homebrew/unpack_strategy/dmg.rb#L96

https://github.com/Homebrew/brew/blob/dae47914ca383e0ed5192436b9e7e11e0bf2640e/Library/Homebrew/cask/quarantine.rb#L131

The latter may require you to use `chmod`, as `FileUtils.chmod_R` doesn't handle dotfiles.

reitermarkus:
@amyspark, this intentionally runs *after* extraction so this only affects the temporary directory, not the unpack directory.

PR #5010, opened 2018-10-01
Skip quitting applications when not logged into GUI.

reitermarkus:

-----
 
Addresses part of https://github.com/Homebrew/homebrew-cask/issues/19076.

### DIFF
--- a/Library/Homebrew/system_command.rb
+++ b/Library/Homebrew/system_command.rb
@@ -8,12 +8,14 @@ require "extend/hash_validator"
 using HashValidator
 require "extend/predicable"
 
-def system_command(*args)
-  SystemCommand.run(*args)
-end
+module Kernel
+  def system_command(*args)
+    SystemCommand.run(*args)
+  end
 
-def system_command!(*args)
-  SystemCommand.run!(*args)
+  def system_command!(*args)
+    SystemCommand.run!(*args)
+  end
 end
 
 class SystemCommand
--- a/Library/Homebrew/utils/user.rb
+++ b/Library/Homebrew/utils/user.rb
@@ -1,0 +1,18 @@
+require "delegate"
+require "etc"
+
+require "system_command"
+
+class User < DelegateClass(String)
+  def gui?
+    out, _, status = system_command "who"
+    return false unless status.success?
+    out.lines
+       .map(&:split)
+       .any? { |user, type,| user == self && type == "console" }
+  end
+
+  def self.current
+    @current ||= new(Etc.getpwuid(Process.euid).name)
+  end
+end
--- a/Library/Homebrew/cask/artifact/abstract_uninstall.rb
+++ b/Library/Homebrew/cask/artifact/abstract_uninstall.rb
@@ -1,5 +1,6 @@
 require "timeout"
 
+require "utils/user"
 require "cask/artifact/abstract_artifact"
 
 module Cask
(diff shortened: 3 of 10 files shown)


Library/Homebrew/utils/user.rb
```diff
@@ -0,0 +1,12 @@
+require "system_command"
+require "etc"
+
+USER = Etc.getlogin
```
MikeMcQuaid: Why all in caps?

Library/Homebrew/utils/user.rb
```diff
@@ -0,0 +1,12 @@
...
+
+def USER.gui?
+  out, _, status = system_command "who"
+  return false unless status.success?
+  out.split("\n")
+     .map { |line| line.split(/\s+/, 3) }
```
MikeMcQuaid: Why 3?

Library/Homebrew/utils/user.rb
reitermarkus: Because it's a constant and not a class.

Library/Homebrew/utils/user.rb
reitermarkus: Name, type, and time (which we don't need).

Library/Homebrew/utils/user.rb
MikeMcQuaid: What is returned, a string? Feels like a class which encapsulates that and has the `gui?` method on it might be more idomatic.

Library/Homebrew/utils/user.rb
MikeMcQuaid: Could that get a comment or variable/constant name?

Library/Homebrew/utils/user.rb
reitermarkus: A class would need to forward every class method to `Etc.getlogin`, which would be much more complex than simply adding a method to this string.

claui: Please do not use `Etc.getlogin`. It’s extremely unreliable.

image

A few seconds later:

image

claui: > would be much more complex than simply adding a method to this string.

Are you thinking about code complexity, or rather complexity at runtime?

Would you consider the following too complex?

```
require 'delegate'
require 'etc'

class User < DelegateClass(String)
  def gui?
    # …
  end
end

USER = User.new(Etc.getpwuid(Process.euid).name)
```

Library/Homebrew/utils/user.rb
```diff
@@ -0,0 +1,18 @@
+require "delegate"
+require "etc"
+
+require "system_command"
+
+class User < DelegateClass(String)
+  def gui?
+    out, _, status = system_command "who"
+    return false unless status.success?
+    out.split("\n")
```
MikeMcQuaid: Can this be `.lines`?

Library/Homebrew/utils/user.rb
```diff
@@ -0,0 +1,18 @@
...
+       .map(&:split)
+       .any? { |user, type, _| user == self && type == "console" }
```
MikeMcQuaid: Can the `_` be omitted?

Library/Homebrew/system_command.rb
```diff
@@ -8,12 +8,14 @@
 using HashValidator
 require "extend/predicable"
 
-def system_command(*args)
-  SystemCommand.run(*args)
-end
+module Kernel
+  def system_command(*args)
```
MikeMcQuaid: Why the move?

MikeMcQuaid: Weird!

Library/Homebrew/utils/user.rb
```diff
@@ -0,0 +1,18 @@
+require "delegate"
+require "etc"
+
+require "system_command"
+
+class User < DelegateClass(String)
+  def gui?
+    out, _, status = system_command "who"
```
szhu: This checks whether the user has a logged in GUI session. It does not check whether the current session is part of that GUI session.

For example, if you are logged in via the GUI and the Homebrew session is via SSH, this will return true, but AppleScript commands will still fail.

Have you tested any cases where this will return false? What are those cases?

szhu: @reitermarkus let me know what you think of the above!

reitermarkus: You're right, it should check the current session, not any session. I'm not sure how to do that, however. Feel free to send a PR to improve this.

PR #5016, opened 2018-10-01
docs/Formula-Cookbook: several content updates

EricFromCanada:
- [N/A] Have you successfully run `brew style` with your changes locally?
- [N/A] Have you successfully run `brew tests` with your changes locally?

-----
Several updates for docs/Formula-Cookbook:
- move all rubydoc.info link URLs to footer to keep the text easier to manage. Also convert each to HTTPS.
- link to specific revisions of formula examples so line highlighting doesn't drift over time.
- monospace formula link text.
- add newer download strategies, and list authenticated strategies in separate table.
- mention that Requirements can specify a minimum version using a String, and that both dependencies and Requirements accept lists.


colindean:
@EricFromCanada Have you validated all of those URLs? Are they all still active and correct?

MikeMcQuaid:
> * move all rubydoc.info link URLs to footer to keep the text easier to manage.

I'm sorry to say I don't agree with this change and think it makes the raw markdown harder to follow.

Otherwise I'm 👍 👏 on everything else.

EricFromCanada:
Very well then. _[wields `rebase` command]_

PR #5018, opened 2018-10-02
relocation_formulae: Poured bottles depend on gcc [Linux]

sjackman:

-----

When pouring a bottle, install the brewed GCC if the major version of the host's GCC is older than the brewed GCC. GCC provides the libraries `libgcc_s.so.1` and `libstdc++.so.6`, needed by bottles.

### DIFF
--- a/Library/Homebrew/extend/os/linux/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/linux/keg_relocate.rb
@@ -83,4 +83,17 @@ class Keg
   def self.relocation_formulae
     ["patchelf"]
   end
+
+  def self.bottle_dependencies
+    @bottle_dependencies ||= begin
+      formulae = relocation_formulae
+      gcc = Formula["gcc"]
+      if !ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] &&
+         DevelopmentTools.non_apple_gcc_version("gcc") < gcc.version.to_i
+        formulae += gcc.recursive_dependencies.map(&:name)
+        formulae << gcc.name
+      end
+      formulae
+    end
+  end
 end
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -489,14 +489,20 @@ class FormulaInstaller
       end
     end
 
-    if pour_bottle && !Keg.relocation_formulae.include?(formula.name)
-      bottle_deps = Keg.relocation_formulae
-                       .map { |formula| Dependency.new(formula) }
-                       .reject do |dep|
+    if pour_bottle && !Keg.bottle_dependencies.empty?
+      bottle_deps = if !Keg.bottle_dependencies.include?(formula.name)
+        Keg.bottle_dependencies
+      elsif !Keg.relocation_formulae.include?(formula.name)
+        Keg.relocation_formulae
+      else
+        []
+      end
+      bottle_deps = bottle_deps.map { |formula| Dependency.new(formula) }
+                               .reject do |dep|
         inherited_options[dep.name] |= inherited_options_for(dep)
         dep.satisfied? inherited_options[dep.name]
       end
-      expanded_deps = Dependency.merge_repeats(bottle_deps + expanded_deps) unless bottle_deps.empty?
+      expanded_deps = Dependency.merge_repeats(bottle_deps + expanded_deps)
     end
 
     expanded_deps.map { |dep| [dep, inherited_options[dep.name]] }
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ branches:
   only:
     - master
 
+env:
+  - HOMEBREW_FORCE_HOMEBREW_ON_LINUX=1
+
 before_install:
   - HOMEBREW_REPOSITORY="$(brew --repo)"
   - sudo chown -R "$USER" "$HOMEBREW_REPOSITORY"
--- a/Library/Homebrew/keg_relocate.rb
+++ b/Library/Homebrew/keg_relocate.rb
@@ -187,6 +187,10 @@ class Keg
   def self.relocation_formulae
     []
   end
+
+  def self.bottle_dependencies
+    relocation_formulae
+  end
 end
 
 require "extend/os/keg_relocate"
--- a/Library/Homebrew/test/formula_installer_bottle_spec.rb
+++ b/Library/Homebrew/test/formula_installer_bottle_spec.rb
@@ -18,6 +18,7 @@ describe FormulaInstaller do
     expect(formula).to pour_bottle
 
     stub_formula_loader formula
+    stub_formula_loader formula("gcc") { url "gcc-1.0" }
     stub_formula_loader formula("patchelf") { url "patchelf-1.0" }
     allow(Formula["patchelf"]).to receive(:installed?).and_return(true)
     described_class.new(formula).install
--- a/Library/Homebrew/test/formulary_spec.rb
+++ b/Library/Homebrew/test/formulary_spec.rb
@@ -138,6 +138,7 @@ describe Formulary do
     context "with installed Formula" do
       before do
         allow(described_class).to receive(:loader_for).and_call_original
+        stub_formula_loader formula("gcc") { url "gcc-1.0" }
         stub_formula_loader formula("patchelf") { url "patchelf-1.0" }
         allow(Formula["patchelf"]).to receive(:installed?).and_return(true)
       end


Library/Homebrew/formula_installer.rb
```diff
@@ -489,14 +489,16 @@ def expand_dependencies(deps)
...
+    if pour_bottle && !Keg.relocation_formulae.empty?
       bottle_deps = Keg.relocation_formulae
                        .map { |formula| Dependency.new(formula) }
                        .reject do |dep|
         inherited_options[dep.name] |= inherited_options_for(dep)
         dep.satisfied? inherited_options[dep.name]
       end
-      expanded_deps = Dependency.merge_repeats(bottle_deps + expanded_deps) unless bottle_deps.empty?
+      i = bottle_deps.find_index { |x| x.name == formula.name }
+      bottle_deps = bottle_deps[0, i] if i
```
sjackman: If the requested formula is one of the dependencies of `gcc`, install all the dependencies of `gcc` up to and including that formula.

Library/Homebrew/formula_installer.rb
```diff
@@ -489,14 +489,16 @@ def expand_dependencies(deps)
       end
     end
 
-    if pour_bottle && !Keg.relocation_formulae.include?(formula.name)
+    if pour_bottle && !Keg.relocation_formulae.empty?
```
sjackman: `!Keg.relocation_formulae.empty?` ensures that this PR has no effect on macOS.

Library/Homebrew/extend/os/linux/keg_relocate.rb
```diff
@@ -81,6 +81,13 @@ def elf_files
   end
 
   def self.relocation_formulae
-    ["patchelf"]
+    return @relocation_formulae if @relocation_formulae
+    f = ["patchelf"]
```
MikeMcQuaid: Can you name this `formulae`? Wasn't obvious on first reading that you would be adding to this.

Library/Homebrew/formula_installer.rb
MikeMcQuaid: I'm not sure I understand the logic here, can you elaborate further?

sjackman:
I've created a new method `Keg.bottle_dependencies`.

sjackman:
Blocked on PR https://github.com/Homebrew/brew/pull/5024

Library/Homebrew/formula_installer.rb
sjackman: > I'm not sure I understand why this isn't done by the existing dependency logic

The only reason is that `patchelf` is not an actual dependency of anything. I have an idea of another way to approach this though.

> I'm nervous about treating the ordering of a flat list rather than dependency tree as something that can be segmented.

Given a sequence of vertices that is a topological ordering of a graph, a prefix of that sequence is also a topological order of that graph. That's not true of an arbitrary substring of a topological ordering, but is true of a prefix.

sjackman: > I'm not sure I understand why this isn't done by the existing dependency logic

The reason is that `patchelf` is an implicit dependency of everything, but not an actual dependency of anything. I have an idea of another way to approach this though.

MikeMcQuaid: > The only reason is that `patchelf` is not an actual dependency of anything. I have an idea of another way to approach this though.

🆒. I think ideally it would be visible as a "proper" dependency e.g. show up in `brew deps` and co. too.

> Given a sequence of vertices that is a topological ordering of a graph, a prefix of that sequence is also a topological order of that graph. That's not true of an arbitrary substring of a topological ordering, but is true of a prefix.

😅. As long as that's covered by more naming and comment: 🆒.

sjackman: > 🆒. I think ideally it would be visible as a "proper" dependency e.g. show up in brew deps and co. too.

I'd agree, but it's a bit tricky, because it's a dependency when pouring something from a bottle, but not otherwise. We don't have that functionality in Homebrew currently.

Library/Homebrew/formula_installer.rb
```diff
@@ -489,14 +489,20 @@ def expand_dependencies(deps)
...
-                       .map { |formula| Dependency.new(formula) }
-                       .reject do |dep|
+    if pour_bottle && !Keg.bottle_dependencies.empty?
+      bottle_deps = if !Keg.bottle_dependencies.include?(formula.name)
+        Keg.bottle_dependencies
+      elsif !Keg.relocation_formulae.include?(formula.name)
+        Keg.relocation_formulae
+      else
+        []
+      end
```
sjackman: @MikeMcQuaid Do you like this code better than what I had before?
To compare see the other version at https://github.com/sjackman/homebrew/commit/64467bff01c93f62bbaed8c37c67f7aedbe1a692#diff-1f60679eb76b419ad46426e2a3b50621

MikeMcQuaid: Yes, much better, thanks 👍

Library/Homebrew/extend/os/linux/keg_relocate.rb
```diff
@@ -83,4 +83,15 @@ def elf_files
   def self.relocation_formulae
     ["patchelf"]
   end
+
+  def self.bottle_dependencies
+    return @bottle_deps if @bottle_deps
```
MikeMcQuaid: Mind doing `@bottle_deps ||= begin`? That's a bit more idiomatic.

sjackman:
```
==> brew install --only-dependencies --include-test testbottest
==> Installing testbottest from homebrew/test-bot
==> Installing dependencies for homebrew/test-bot/testbottest: gmp, isl, mpfr, libmpc, gcc
```
https://travis-ci.org/Homebrew/brew/jobs/437370091#L909
Well this is unexpected behaviour. I'll try to figure out what's going on.

sjackman:
`brew install gcc` fails when using Homebrew/homebrew-core.
```
==> Installing homebrew/test-bot/testbottest dependency: gcc
==> Downloading https://ftp.gnu.org/gnu/gcc/gcc-8.2.0/gcc-8.2.0.tar.xz
==> Patching
patching file gcc/graphite.h
Error: An exception occurred within a child process:
  NameError: uninitialized constant OS::Mac::CLT
==> FAILED
```
https://travis-ci.org/Homebrew/brew/jobs/444411721#L1018

Using Linuxbrew/homebrew-core on Linux would resolve this issue. I've opened PR https://github.com/Homebrew/homebrew-test-bot/pull/212.

MikeMcQuaid:
> `brew install gcc` fails when using Homebrew/homebrew-core.

We definitely don't want to have `testbottest` building (or even pouring) GCC every time.

sjackman:
> We definitely don't want to have testbottest building (or even pouring) GCC every time.

Fair enough. One option is to set `HOMEBREW_FORCE_HOMEBREW_ON_LINUX` on TravisCI. Another option is to use a Docker container that has `/usr/bin/gcc` 8.2.0 installed. Does one of these two solutions work for you, or do you have another suggestion?

MikeMcQuaid:
@sjackman Setting `HOMEBREW_FORCE_HOMEBREW_ON_LINUX` makes sense to me for now 👍

sjackman:
> @sjackman Setting HOMEBREW_FORCE_HOMEBREW_ON_LINUX makes sense to me for now 👍

Done in `.travis.yml`

MikeMcQuaid:
@sjackman I have also been seeing this lots. I'd welcome any help debugging or fixing it.

sjackman:
@MikeMcQuaid Are you okay with merging this PR (after tagging 1.8.0) despite TravisCI timing out on macOS?

MikeMcQuaid:
> Are you okay with merging this PR (after tagging 1.8.0) despite TravisCI timing out on macOS?

@sjackman No, it's a required build so I'd like it to be fixed first.

MikeMcQuaid:
> No, it's a required build so I'd like it to be fixed first.

(or rerun)

MikeMcQuaid:
@sjackman Please rebase this on master which should fix Travis CI issues.

sjackman:
Success! What was the fix?

MikeMcQuaid:
@sjackman https://github.com/Homebrew/brew/commit/284e3b8db47b6460d51b3bf403cde18f52341e48

sjackman:
Merged! Homebrew/brew should largely work on Linux now. I'll review the remaining diff and see if there's anything critical in there.

From https://github.com/Linuxbrew/brew/issues/612#issuecomment-412721962

> Add a Dockerfile to build a Docker image with a Docker Hub automated build

I'd like to add a `Dockerfile` to Homebrew/brew to build a Docker image for Homebrew/brew using Docker Hub automated builds. Since we don't currently have the `homebrew` organization, I'd store this image in the `linuxbrew` organization and name it `linuxbrew/brew`. The current Linuxbrew image is named `linuxbrew/linuxbrew`.

@MikeMcQuaid Are you okay with me adding a `Dockerfile` to the root directory of `Homebrew/brew`? I don't believe a subdirectory would work, because the build context (the Homebrew/brew source code) comes the from the direction that contains the file `Dockerfile`.

sjackman:
The purpose of this image is for Linuxbrew CI to build bottles, and for users.

sjackman:
There's certainly automated testing. `hub.docker.com` builds the image. I don't know how well Docker Hub integrates with either GitHub or Slack. Alternatively, we could use Azure Pipelines to build and deploy the Docker image. Would you prefer that?

MikeMcQuaid:
@sjackman Sounds good. If you can investigate a Slack or GitHub integration that would be great but sounds fine as-is.

PR #5020, opened 2018-10-02
zip: restore extended attributes when under MacOS

amyspark:

-----

Fixes Homebrew/homebrew-cask#52396.

### DIFF
--- a/Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
+++ b/Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
@@ -2,10 +2,21 @@ module UnpackStrategy
   class Zip
     prepend Module.new {
       def extract_to_dir(unpack_dir, basename:, verbose:)
-        volumes = super.stderr.chomp
-                       .split("\n")
-                       .map { |l| l[/\A   skipping: (.+)  volume label\Z/, 1] }
-                       .compact
+        result = super
+
+        volumes = result.stderr.chomp
+                        .split("\n")
+                        .map { |l| l[/\A   skipping: (.+)  volume label\Z/, 1] }
+                        .compact
+
+        if result.stderr.lines.any? { |line| line.start_with?("._") }
+          # Merge ._ files back into extended attributes.
+          # ._ files inside volumes are automatically merged by ditto.
+          system_command!("dot_clean",
+                          args: ["-mv", "--keep=dotbar", unpack_dir],
+                          verbose: verbose,
+                          print_stderr: false)
+        end
 
         return if volumes.empty?
 


Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,12 +2,23 @@ module UnpackStrategy
...
-                       .compact
+        result = super
 
-        return if volumes.empty?
+        volumes = result.stderr.chomp
+                        .split("\n")
+                        .map { |l| l[/\A   skipping: (.+)  volume label\Z/, 1] }
+                        .compact
+
+        if volumes.empty?
```
reitermarkus: This could also be the case if there were skipped volumes.

Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,12 +2,23 @@ module UnpackStrategy
...
+          # Merge ._ files back into native, extended attributes.
+          if result.stdout.lines.grep(/\._/)
```
reitermarkus: This will always be true, because it will be an empty array if nothing matches.

Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,12 +2,23 @@ module UnpackStrategy
...
+            system_command!("dot_clean",
+                            args: ["-mnv", "--keep=dotbar", unpack_dir],
```
reitermarkus:
```
     -m      Always delete dot underbar files.
     -n      Delete dot underbar file if there is no matching native file.
```

Does `-m` not imply `-n`? Also, `-v` should only be passed if `verbose` is true.

Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,10 +2,21 @@ module UnpackStrategy
...
+                        .map { |l| l[/\A   skipping: (.+)  volume label\Z/, 1] }
+                        .compact
+
+        # Merge ._ files back into native, extended attributes.
+        unless result.stdout.lines.grep(/\._/).empty?
+          system_command!("dot_clean",
+                          args: ["-mnv", "--keep=dotbar", unpack_dir],
+                          verbose: verbose,
+                          print_stderr: false)
+          return result
```
reitermarkus: There could still be skipped volumes that need to be handled below.

Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,10 +2,21 @@ module UnpackStrategy
...
-                       .compact
+        result = super
+
+        volumes = result.stderr.chomp
+                        .split("\n")
+                        .map { |l| l[/\A   skipping: (.+)  volume label\Z/, 1] }
+                        .compact
+
+        # Merge ._ files back into native, extended attributes.
+        unless result.stdout.lines.grep(/\._/).empty?
```
reitermarkus: `/\._/` will match anywhere, not just the beginning.

Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,26 +2,39 @@ module UnpackStrategy
...
+              FileUtils.mv tmp_unpack_dir/volume, unpack_dir/volume, verbose: verbose
+            end
           end
         end
+
+        return if result.stderr.lines.grep(/^\._/).empty?
+
+        system_command!("dot_clean",
+                        args: [
+                          "-m#{"v" if verbose}",
```
reitermarkus: Actually, I think this can stay, since we don't print to stdout anyways.

amyspark: Actually, `dot_clean` prints to stderr, which is hidden thanks to `print_stderr: false`.

Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,26 +2,35 @@ module UnpackStrategy
...
-            FileUtils.mv tmp_unpack_dir/volume, unpack_dir/volume, verbose: verbose
+            volumes.each do |volume|
+              FileUtils.mv tmp_unpack_dir/volume, unpack_dir/volume, verbose: verbose
+            end
           end
         end
+
+        return if result.stderr.lines.grep(/^\._/).empty?
+
+        system_command!("dot_clean",
```
reitermarkus: I think I prefer the old ordering, first `dot_clean`, then `ditto`.

Library/Homebrew/extend/os/mac/unpack_strategy/zip.rb
```diff
@@ -2,10 +2,21 @@ module UnpackStrategy
...
-                       .map { |l| l[/\A   skipping: (.+)  volume label\Z/, 1] }
-                       .compact
+        result = super
+
+        volumes = result.stderr.chomp
+                        .split("\n")
+                        .map { |l| l[/\A   skipping: (.+)  volume label\Z/, 1] }
+                        .compact
+
+        unless result.stderr.lines.grep(/^\._/).empty?
```
reitermarkus:
```ruby
if result.stderr.lines.any? { |line| line.start_with?("._") }
```

amyspark:
I've cleaned this up, feel free to merge 🚀

commitay:
Might need to rebase to get Codecov to update.

PR #5024, opened 2018-10-02
CoreTap.default_remote: Use Linuxbrew/core [Linux]

sjackman:

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -24,7 +24,12 @@ git() {
 
 git_init_if_necessary() {
   BREW_OFFICIAL_REMOTE="https://github.com/Homebrew/brew"
-  CORE_OFFICIAL_REMOTE="https://github.com/Homebrew/homebrew-core"
+  if [[ -n "$HOMEBREW_MACOS" ]] || [[ -n "$HOMEBREW_FORCE_HOMEBREW_ORG" ]]
+  then
+    CORE_OFFICIAL_REMOTE="https://github.com/Homebrew/homebrew-core"
+  else
+    CORE_OFFICIAL_REMOTE="https://github.com/Linuxbrew/homebrew-core"
+  fi
 
   safe_cd "$HOMEBREW_REPOSITORY"
   if [[ ! -d ".git" ]]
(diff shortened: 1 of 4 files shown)


Library/Homebrew/cmd/update.sh
```diff
@@ -24,7 +24,12 @@ git() {
 
 git_init_if_necessary() {
   BREW_OFFICIAL_REMOTE="https://github.com/Homebrew/brew"
-  CORE_OFFICIAL_REMOTE="https://github.com/Homebrew/homebrew-core"
+  if [[ -n "$HOMEBREW_MACOS" ]] || [[ -n "$HOMEBREW_FORCE_HOMEBREW_ORG" ]]
```
MikeMcQuaid: This is just used by `brew test-bot`, right?

Library/Homebrew/extend/os/linux/tap.rb
```diff
@@ -0,0 +1,9 @@
+class CoreTap < Tap
+  def default_remote
+    if ENV["HOMEBREW_FORCE_HOMEBREW_ORG"]
+      "https://github.com/Homebrew/homebrew-core".freeze
+    else
+      "https://github.com/Linuxbrew/homebrew-core".freeze
```
MikeMcQuaid: Will this direct people to file Linux formulae issues there (if we don't already do that)?

MikeMcQuaid:
May also wish/need to adjust https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb

Library/Homebrew/cmd/update.sh
sjackman: `HOMEBREW_FORCE_HOMEBREW_ORG` you mean? Yes, I believe so.
See https://github.com/Homebrew/homebrew-test-bot/blob/ac511e47217a49dfbeffcb2f09e384f8aa49e13b/cmd/brew-test-bot.rb#L1325-L1329

Library/Homebrew/extend/os/linux/tap.rb
sjackman: The user is directed to `OS::ISSUES_URL`, which on Linux is https://github.com/Linuxbrew/brew/wiki/Troubleshooting.
As far as I can tell, `CoreTap.instance.issues_url` is not used when `official_tap?` is true.

sjackman:
> May also wish/need to adjust https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb

True. Linuxbrew is using our fork Linuxbrew/test-bot. We'll want to merge that into Homebrew/test-bot. That can come after merging this PR though.

PR #5026, opened 2018-10-03
Cask: attempt to gain permissions before trash

amyspark:

-----

This pull request allows Cask to have permissions on trashable items before issuing the relevant `osascript`.

@commitay: it should work with your use case, but do tell me if something arises.

Fixes Homebrew/homebrew-cask#44993.

### DIFF
--- a/Library/Homebrew/cask/utils.rb
+++ b/Library/Homebrew/cask/utils.rb
@@ -8,13 +8,13 @@ BUG_REPORTS_URL = "https://github.com/Homebrew/homebrew-cask#reporting-bugs".fre
 module Cask
   module Utils
     def self.only_gain_permissions(path, command: SystemCommand)
-      if path.respond_to?(:rmtree) && path.exist?
-        gain_permissions(path, ["-R"], command) do |p|
-          raise CaskError, "The file #{p} cannot be accessed." unless p.writable?
-        end
-      elsif File.symlink?(path)
+      if path.symlink?
         gain_permissions(path, ["-h"], command) do |p|
           raise CaskError, "The symbolic link #{p} cannot be accessed." unless p.writable?
+        end
+      elsif path.exist?
+        gain_permissions(path, ["-R"], command) do |p|
+          raise CaskError, "#{p} cannot be accessed." unless p.writable?
         end
       end
     end


Library/Homebrew/cask/utils.rb
```diff
@@ -7,6 +7,18 @@
 
 module Cask
   module Utils
+    def self.only_gain_permissions(path, command: SystemCommand)
+      if path.respond_to?(:rmtree) && path.exist?
```
MikeMcQuaid: When would it not respond to `rmtree`?

Library/Homebrew/cask/utils.rb
```diff
@@ -7,6 +7,18 @@
...
+        gain_permissions(path, ["-R"], command) do |p|
+          raise CaskError, "The file #{p} cannot be accessed." unless p.writable?
+        end
+      elsif File.symlink?(path)
```
MikeMcQuaid: If this is a Pathname (or could be made a Pathname) how about `path.symlink?`?

Library/Homebrew/cask/utils.rb
amyspark: Judging from the caller, it always should. I'll rewrite this with your comment below.

commitay:
With this PR, `brew cask install virtualbox && brew cask zap virtualbox` still opens a popup for permissions when zapping `/Library/Application Support/VirtualBox`.

amyspark:
@commitay - you're correct. `/Library/Application Support`'s access rights blocks us from trashing the `VirtualBox` directory. Should we `rmdir` this directory manually? It will ask for sudo anyway.

commitay:
The issus was intended to find a solution to:
1. Avoid GUI popup for permissions (which is kind of weird as Homebrew doesn't do this for anything else).
2. Allow us to run `zap` instead of `uninstall` on CI (to avoid spurious warnings for stuff that is still installed but included in the `zap` stanza).

`virtualbox` is just one example, other casks also have this `trash` permission problem as well so we need something that works everywhere, deleting stuff we can't trash isn't really an option as we can't make the assumption that it is safe to delete.

PR #5038, opened 2018-10-04
Add Artifactory download strategy

kralf:

-----
This PR proposes a new download strategy which calls the [Jfrog command-line interface](https://www.jfrog.com/confluence/display/CLI/JFrog+CLI) client to download artefacts. To use it, one should add `:using => :artifactory` in the URL section of a formula/bottle. If the download URL matches the scheme `https://<SERVER_ID>/artifactory/<REPO_NAME>/<REPO_PATH>`, this strategy may adhere to the credentials specified by running `jfrog rt config <SERVER_ID>` for authentication.

The PR intentionally does not add a rule to detect the strategy for Artifactory URL's (which would for instance be possible by matching the path name) such that, for generic and public repositories, the default download strategy remains `curl`.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -1200,6 +1200,68 @@ class FossilDownloadStrategy < VCSDownloadStrategy
   end
 end
 
+# ArtifactoryDownloadStrategy downloads files using the Jfrog command-line
+# interface client. To use it, add `:using => :artifactory` to the URL
+# section of your formula. If the download URL matches the scheme
+# `https://<SERVER_ID>/artifactory/<REPO_NAME>/<REPO_PATH>`, this strategy
+# may use the credentials specified by running `jfrog rt config <SERVER_ID>`
+# for authentication.
+class ArtifactoryDownloadStrategy < AbstractFileDownloadStrategy
+  def initialize(url, name, version, **meta)
+    super
+    parse_url_pattern
+  end
+
+  def parse_url_pattern
+    unless match = url.match(%r{https://([^/]+)/artifactory/([^/]+)/(\S+)})
+      raise "Download failed: Invalid url pattern for Artifactory Repository."
+    end
+
+    _, @server, @repo, @filepath = *match
+  end
+
+  def jfrog_cli
+    Formulary.factory("jfrog-cli-go").installed_prefix/"bin/jfrog"
+  end
+
+  def fetch
+    ohai "Downloading #{url}"
+
+    if cached_location.exist?
+      puts "Already downloaded: #{cached_location}"
+    else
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
What's the use-case for this over something like `HOMEBREW_ARTIFACT_DOMAIN`?

colindean:
@kralf can confirm, but I speculate that it has something to do with using credentials that are only available to the jfrog CLI. 

It would be nice if the jfrog CLI looked at netrc like curl and other tools do.

kralf:
Confirmed. The use case is to transparently handle authentication against Artifactory in cases where this is required. For unauthenticated access, curl download with `HOMEBREW_ARTIFACT_DOMAIN` will do just fine. Hence, this strategy would have to be selected explicitly using the symbol.

MikeMcQuaid:
> The use case is to transparently handle authentication against Artifactory in cases where this is required.

Can you elaborate (a lot) more: when is it required and when is it not required?

kralf:
Sure. Artifactory repositories may either allow [anonymous user access](
https://www.jfrog.com/confluence/display/RTF/Configuring+Security#ConfiguringSecurity-AllowAnonymousAccess) (no authentication required) or have fine-grained access control based on user/group membership. In the latter case, HTTP requests require [basic authentication](
https://www.jfrog.com/confluence/display/RTF/Access+Tokens#AccessTokens-BasicAuthentication) with username and password or access token. Alternatively, access tokens ma be passed in an [HTTP authorization header](
https://www.jfrog.com/confluence/display/RTF/Access+Tokens#AccessTokens-AuthorizationHeaders). Anyways, Jfrog CLI handles authentication transparently, and credentials may be stored securely within configuration objects.

The general concept behind the Artifactory download strategy would be to give the user the possibility to pre-configure access to private repositories (including credentials) using the Jfrog CLI, and to then run the CLI download command from within `brew`. The link between the configuration object and the formula/bottle resource URL would be the server ID which, in the proposed implementation, would have to match the server hostname.

@MikeMcQuaid Does this answer your question?

colindean:
So the value of this is primarily for Homebrew users accessing private Artifactory instances with private formulae.

kralf:
Yes, comparable to using `:github_private_release` or `:s3` instead of `:curl`.

MikeMcQuaid:
What if you're using a private Artifactory with public formulae?

MikeMcQuaid:
I do not want to include any more download strategies that will not be used by Homebrew/homebrew-core. Instead I'm interested in allowing taps/private taps formulae to have ways to specify download strategies within their formulae.

MikeMcQuaid:
I've opened https://github.com/Homebrew/brew/issues/5074 with the proposed new approach.

PR #5042, opened 2018-10-04
Use CurlDownloadStrategy for Apache direct-download links.

brainy:

-----

Sources for most Apache projects are downloaded from URLs like this:

    https://www.apache.org/dyn/closer.cgi?path=package/package-1.0.0.tar.bz2

This URL returns an HTML page where the preferred mirror is shown, and Homebrew parses the mirror URL from there.

But there's a better way to do this; an URL like this:

    https://www.apache.org/dyn/closer.cgi?action=download&filename=package/package-1.0.0.tar.bz2

will redirect to the preferred mirror, so that there's no need to parse the response.

This PR adds support for this format of Apache direct-download URLs.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -1225,8 +1225,9 @@ class DownloadStrategyDetector
     when %r{^https?://.+\.git$},
          %r{^git://}
       GitDownloadStrategy
-    when %r{^https?://www\.apache\.org/dyn/closer\.cgi},
-         %r{^https?://www\.apache\.org/dyn/closer\.lua}
+    when %r{^https?://www\.apache\.org/dyn/closer\.(cgi|lua)\?action=download&filename=}
+      CurlDownloadStrategy
+    when %r{^https?://www\.apache\.org/dyn/closer\.(cgi|lua)}
       CurlApacheMirrorDownloadStrategy
     when %r{^https?://(.+?\.)?googlecode\.com/svn},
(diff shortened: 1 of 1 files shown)


brainy:
BTW, if someone would be kind enough to explain how to write a test for this, I'll be happy to add one, to satisfy the `codecov/patch` check.

colindean:
I think you'd add a test within [this describe block](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/download_strategies_spec.rb#L549).

brainy:
Thanks @colindean, that helped; and here's hoping the tests will work here as well as they did in my local setup.

brainy:
The current `dyn/closer.cgi`tool (or `.lua`, it's the same thing really) will surely be available for a long time, since most Apache projects depend on it to generate their customised download pages.

There's some discussion about creating a download service which would redirect to mirrors and also to `archive.apache.org` when necessary. But it's still early days for that, and even when it's deployed, `dyn/closer.cgi` isn't likely to go away any time soon.

reitermarkus:
> But there's a better way to do this; an URL like this:

Why is that way “better”?

We parse the JSON response to get the preferred mirror *and* backup mirrors in case the preferred one fails.

brainy:
Ah well, I assumed that letting the server redirect would be preferable to parsing some undocumented JSON and hoping its schema doesn't suddenly change. Formulae contain backup mirror URLs, too.

PR #5043, opened 2018-10-04
rubocop: only use config if style is testing specified file(s)

javian:

-----
This fixes the issue in #4991 explained in https://github.com/Homebrew/brew/pull/4991#issuecomment-426875323 .
This will cause `brew style` to recursively read all `.rubocop.yml` files in the `Library/Homebrew` directory.
There is currently a bug somewhere that causes rubocop to barf on *_spec.rb files that are not `rspec` files so to complete this PR that needs to be addressed as well.

I might need to have to update the test as well but I haven't looked at that yet.


javian:
The rubocop-rspec issue has been reported upstream https://github.com/rubocop-hq/rubocop-rspec/issues/696

javian:
This will fail since a few rspec tests has slipped through since its not checking for them at the moment. 

I've gone in the opposite direction since something is broken with the `require` statement loading it like it was intended in the previous PR. The way its set up now it will (before this gets merged) `rubocop-rspec` will get loaded but will apply to the whole source tree of configs so having it the config with the requirement in a subfolder might make it less obvious that it is there. There's also a forced pattern applied on RSPec that will only look for `RSpec` cops in the test directory which I think would suffice. What do you think ?

MikeMcQuaid:
@javian I'm open to whatever gets CI 💚 and the checks working again.

javian:
@MikeMcQuaid I think that's as good as it gets for now

Library/Homebrew/test/.rubocop.yml
```diff
@@ -2,4 +2,3 @@ inherit_from:
   - ../.rubocop.yml
   - .rubocop_todo.yml
 
-require: rubocop-rspec
```
MikeMcQuaid: Can the `RSpec` section above be in this file?

javian: It looks like it, pushing now.

MikeMcQuaid:
@javian To confirm: is there no way to have a `.rubocop.yml` file that's not trying to include the `rubocop-rspec` gem? My concern is that including that gem like we're doing here means that the in-editor `rubocop` integration won't work for anyone who doesn't manually install that gem.

I'm almost tempted to say that we should vendor this gem in `bundle-standalone` for that reason. Thoughts? Also CC @GauthamGoli and @reitermarkus for thoughts.

javian:
I would love to see something like
```
          begin
            require(r)
          rescue LoadError => e
                  puts "cant find requirement #{r}, are you missing a gem or a file?"
                  exit
          end
```
in Rubocop to make it fail more nicely. Right now its just `require(r)`

MikeMcQuaid:
Thanks again @javian! I still reckon it may be worth vendoring `rubocop-rspec`.

MikeMcQuaid:
Thanks again @javian!

PR #5061, opened 2018-10-08
Cask: set write permissions on symlinks

amyspark:

-----

Related to Homebrew/homebrew-cask-versions#6368, turns out some zip files ship with symbolic links set to read-only for the current user, which is not covered by `chmod -R` (it needs the `-h` flag which is mutually exclusive).

This pull request adds an extra pass that only touches symlinks on the staged path, courtesy of `xargs`.

Ping @Homebrew/brew for review, as it blocks the AdoptOpenJDK Cask.

### DIFF
--- a/Library/Homebrew/cask/quarantine.rb
+++ b/Library/Homebrew/cask/quarantine.rb
@@ -136,11 +136,6 @@ module Cask
 
       resolved_paths = Pathname.glob(to/"**/*", File::FNM_DOTMATCH)
 
-      system_command!("/bin/chmod", args: ["-R", "u+w", to])
-
-      # Symlinks cannot be fixed with -R.
-      resolved_symlinks = resolved_paths.select(&:symlink?)
-
       system_command!("/usr/bin/xargs",
                       args: [
                         "-0",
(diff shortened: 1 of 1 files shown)


claui:
@amyspark I may be missing something here but would macOS not [ignore permissions of symlinks](https://superuser.com/questions/1099634/change-permissions-symbolic-link-mac-os#1099808) anyway?

amyspark:
Under normal circumstances yes @claui. However, the propagation step does not follow symlinks (they can be broken e.g. `disk-inventory-x`'s Frameworks folder). It adds the extended attribute on the symlink itself, which is why it needs the `+w` permission.

Library/Homebrew/cask/quarantine.rb
```diff
@@ -138,6 +138,19 @@ def propagate(from: nil, to: nil)
 
       system_command!("/bin/chmod", args: ["-R", "u+w", to])
```
reitermarkus: Might as well remove this and only use the one below for all paths.

amyspark:
Ping @Homebrew/brew for merging.

reitermarkus:
Thanks, @amyspark!

PR #5067, opened 2018-10-08
audit: Permit HEAD for non-official taps

sjackman:

-----

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -642,7 +642,7 @@ module Homebrew
         end
       end
 
-      if formula.head || formula.devel
+      if @official_tap && (formula.head || formula.devel)
         unstable_spec_message = "Formulae should not have a `HEAD` or `devel` spec"
         if @new_formula
           new_formula_problem unstable_spec_message


colindean:
LGTM but I have to ask: What debt does this allow tap maintainers to incur, should an unofficial tap become official? What is the frequency of that? I presume that it's rare enough that the potential work necessary to remediate HEAD taps would be considered as a part of the process of determining if a tap should become official.

sjackman:
I don't think an entire tap is likely to become official, all in one go. Typically formulae would be migrated one-by-one to Homebrew/core, and `head` can be removed in that process.

sjackman:
> Makes sense 👍. At this point I think it's worth replacing all references to @official_tap with tap.core_tap?

I'll do that in a second PR.

sjackman:
>> Makes sense 👍. At this point I think it's worth replacing all references to @official_tap with tap.core_tap?

> I'll do that in a second PR.

On second thought no, it looks like there's an option to override the value of `@official_tap` for testing.

https://github.com/Homebrew/brew/blob/7b13f4a402fb24e4a0a746a16b83f6894d40a220/Library/Homebrew/dev-cmd/audit.rb#L263-L264

MikeMcQuaid:
@sjackman 🆒, thanks for checking.

PR #4383, opened 2018-06-28
cli_parser: man-pages and help text generation using Homebrew::CLI::Parser

GauthamGoli:
- [x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/homebrew/pull/49031).
- [x] Have you successfully run `brew style` with your changes locally?
- [x] Have you successfully run `brew tests` with your changes locally?

-----
https://github.com/Homebrew/brew/pull/4271#issuecomment-394177933
This PR is about generating `--help` text using `OptionParser` and to get feedback on the interface, and the way help text is being shown as its different from what we currently show.

`brew audit --help` would return something like this.

```
Usage: brew audit [options] [<formulae>]

Check <formulae> for Homebrew coding style violations. This should be
run before submitting a new formula.

If no <formulae> are provided, all of them are checked.
        --strict                     Run additional style checks, including Rubocop style checks.
        --online                     Run additional slower style checks that require a
                                     network connection.
        --new-formula                Run various additional style checks to determine if a new formula
                                     is eligible for Homebrew. This should be used when creating
                                     new formula and implies --strict and --online.
        --fix                        Fix style violations automatically using
                                     RuboCop's auto-correct feature.
        --display-cop-names          Include the RuboCop cop name for each violation
                                     in the output.
        --display-filename           Prefix everyline of output with name of the file or
                                     formula being audited, to make output easy to grep.
    -D, --audit-debug                Activates debugging and profiling
        --only                       Passing --only=method will run only the methods named audit_method,
                                     `method` should be a comma-separated list.
        --except                     Passing --except=method will run only the methods named audit_method,
                                     `method` should be a comma-separated list.
        --only-cops                  Passing --only-cops=cops will check for violations of only the listed
                                     RuboCop cops. `cops` should be a comma-separated list of cop names.
        --except-cops                Passing --except-cops=cops will skip checking the listed
                                     RuboCop cops violations. `cops` should be a comma-separated list of cop names.
    -v, --verbose                    Verbose mode.
    -d, --debug                      Display debug info.
    -h, --help                       Show this message
```

PR #5075, opened 2018-10-09
docs: conversion & review of Homebrew::CLI::Parser help in dev-cmd/

EricFromCanada:

-----
Continuing the work started by #4383, plus any useful text or code improvements.

- [x] audit.rb
- [x] bottle.rb
- [x] bump-formula-pr.rb
- [x] create.rb
- [x] edit.rb
- [x] extract.rb
- [x] formula.rb
- [x] irb.rb
- [x] linkage.rb
- [x] man.rb
- [x] mirror.rb
- [x] prof.rb
- [x] pull.rb
- [x] release-notes.rb
- [x] ruby.rb
- [x] tap-new.rb
- [x] test.rb
- [x] tests.rb
- [x] update-test.rb


MikeMcQuaid:
Looking great so far, nice work @EricFromCanada!

Library/Homebrew/dev-cmd/prof.rb
```diff
@@ -1,12 +1,28 @@
 #:  * `prof` [<ruby options>]:
 #:    Run Homebrew with the Ruby profiler.
-#:    For example:
-#:      brew prof readall
+#:
+#:    *Example:* `brew prof readall`
+
+require "cli_parser"
```
GauthamGoli: Cannot use `CLI::Parser` in `prof.rb` as of now. Thats because all the options have to be declared and `prof` takes in ruby options which we cannot declare. Need to add a functionaly in `CLI::Parser` to be able to ignore unknown options.

Library/Homebrew/dev-cmd/prof.rb
```diff
@@ -1,12 +1,28 @@
...
+
+        Run Homebrew with the Ruby profiler.
+
+        *Example:* `brew prof readall`
+      EOS
+    end
+  end
+
   def prof
+    prof_args.parse
```
MikeMcQuaid: `Homebrew.prof_args.parse` or `self-prof_args.parse` I think; not valid syntax as-is unfortunately. CC @GauthamGoli

MikeMcQuaid: Ignore me, I meant to post this below.

Library/Homebrew/dev-cmd/ruby.rb
```diff
@@ -1,13 +1,28 @@
...
-#       brew ruby script.rb
+#:
+#:    *Example:* `brew ruby -e "puts :gcc.f.deps"` or `brew ruby script.rb`
+
+require "cli_parser"
 
 module Homebrew
   module_function
 
+  def ruby_args
```
GauthamGoli: Same reason as mentioned `prof.rb`. Cannot use `CLI::Parser` yet.

Library/Homebrew/dev-cmd/ruby.rb
```diff
@@ -1,13 +1,28 @@
...
+
+        Run a Ruby instance with Homebrew's libraries loaded.
+
+        *Example:* `brew ruby -e "puts :gcc.f.deps"` or `brew ruby script.rb`
+      EOS
+    end
+  end
+
   def ruby
+    prof_args.parse
```
MikeMcQuaid: `ruby_args`

Library/Homebrew/dev-cmd/prof.rb
EricFromCanada: Does the same issue affect `test.rb`?

GauthamGoli: No this issue does not affect `test.rb` as it accepts only specific set of options described in the documentation, so we can use `CLI::Parser` there.

MikeMcQuaid: `brew prof` doesn't have to take arbitrary arguments; we can document the `brew` we want to accept and expand them if necessary?

Library/Homebrew/dev-cmd/ruby.rb
MikeMcQuaid: `brew ruby` doesn't have to take arbitrary arguments; we can document the `ruby` arguments we want to accept (like `ls`)?

Library/Homebrew/dev-cmd/prof.rb
MikeMcQuaid: (like `brew ls`)

Library/Homebrew/dev-cmd/ruby.rb
GauthamGoli: Yeah we can accept a fix set of options. What options are we planning on supporting for `brew ruby`?

MikeMcQuaid: @GauthamGoli I think `-e` and `-v` are fine for now.

GauthamGoli: @EricFromCanada Please declare these options.

EricFromCanada: Have a look.

EricFromCanada:
Regarding colons following each command spec: in the comment-based help text, they're there to prompt `ronn-format` to begin a [definition list](http://rtomayko.github.io/ronn/ronn-format.7.html#DEFINITION-LISTS), and to visually separate the command from its description in the HTML output. In the Parser-based help text they're not necessary for either, so I believe they can be removed.

Library/Homebrew/dev-cmd/ruby.rb
```diff
@@ -1,13 +1,30 @@
...
+      usage_banner <<~EOS
+        `ruby` [<ruby options>]:
+
+        Run a Ruby instance with Homebrew's libraries loaded.
+
+        *Example:* `brew ruby -e "puts :gcc.f.deps"` or `brew ruby script.rb`
+      EOS
+      switch "-e",
+        description: "Execute the provided string argument as a script."
+      switch :verbose
```
GauthamGoli: @EricFromCanada I think we'd have to make it `switch "-v"` instead of `switch :verbose` as it would automatically use `-v` but we want `-v` to print out Ruby version here.

GauthamGoli:
@EricFromCanada Yes, you are right about colons. They can be removed for commands which use `CLI::Parser`

Library/Homebrew/dev-cmd/ruby.rb
EricFromCanada: Seems to work as-is for me.
```
$ ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin15]
$ brew ruby -v
ruby 2.3.7p456 (2018-03-28 revision 63024) [universal.x86_64-darwin9.0]
```

EricFromCanada:
And a style question: currently each command's description wraps at 88 characters ([80 + indent](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/help.rb#L31)); would it be ideal to also wrap each switch's `description:` to that length?

(By the way, I do plan to squash these commits when done.)

MikeMcQuaid:
> And a style question: currently each command's description wraps at 88 characters ([80 + indent](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/help.rb#L31)); would it be ideal to also wrap each switch's `description:` to that length?

@EricFromCanada Yeh, I think so! Not in the manpage which handles wrapping itself but in `--help` for sure.

MikeMcQuaid:
@EricFromCanada Ah, gotcha, We're fine with going up to 118 characters https://github.com/Homebrew/brew/blob/master/Library/Homebrew/.rubocop.yml#L61-L63 but if you can get them down to 80 and still look good that'd be great 👍

EricFromCanada:
That should do it; please let me know if you’ve any further feedback.

EricFromCanada:
Rebased; please let me know if you’ve any further feedback.
@MikeMcQuaid @GauthamGoli

MikeMcQuaid:
@EricFromCanada Getting a test failure: https://travis-ci.org/Homebrew/brew/jobs/443705317#L733

EricFromCanada:
Fixed. Odd that a missing `require` would only affect Linux.

GauthamGoli:
Nice attention to detail and nicely done @EricFromCanada  :D

MikeMcQuaid:
Thanks @EricFromCanada!

PR #5076, opened 2018-10-09
Remove redundant check in formula_files and cask_files

emielwiedijk:
Formula_files consists of every non-recursive child of `formula_dir`, for which `formula_file?` evaluates to true. `formula_file?` checks if the file is a child of `formula_dir`, which it is by definition. It turns out that by removing the check, the time used for `brew search` decreased from 800 ms to 700 ms, noticably faster during tab completion. The same happens with
cask_files and cask_file?

-----

### DIFF
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -382,7 +382,7 @@ class Tap
   # an array of all {Formula} files of this {Tap}.
   def formula_files
     @formula_files ||= if formula_dir.directory?
-      formula_dir.children.select { |file| file.extname == ".rb" }
+      formula_dir.children.select(&method(:ruby_file?))
     else
       []
     end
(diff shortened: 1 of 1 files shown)


Library/Homebrew/tap.rb
```diff
@@ -382,7 +382,7 @@ def contents
   # an array of all {Formula} files of this {Tap}.
   def formula_files
     @formula_files ||= if formula_dir.directory?
-      formula_dir.children.select(&method(:formula_file?))
+      formula_dir.children.select { |file| file.extname == ".rb" }
```
MikeMcQuaid: `formula_file?` is no longer used after this change. It'd be nicer to have a `formula_file?` (or, probably more accurately, `ruby_file?`) method which is called in the same way as previously and shared by both cases.

aimileus: Initially I removed both `formula_file?` and `cask_file?`, but appareantly they are used in `dev-cmd/update-report.rb` for `brew update`. I did add a `ruby_file?` method and used them in `formula_file?` and `cask_file?`.

PR #5085, opened 2018-10-11
bump-formula-pr: gracefully handle unforkable repositories

alyssais:
Hello! 👋 It's been a while.

-----

GitHub seems to be [discouraging](https://help.github.com/articles/allowing-people-to-fork-private-repositories-in-your-organization/) forking private repositories:

> By default, new organizations are configured to disallow the forking of private repositories.

bump-formula-pr tries to create its pull requests from a fork, so it can't be used for private taps set up in this way.

I've added a --no-fork option that will create PRs in the tap repo itself, rather than in a fork, to accommodate this use case.

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-formula-pr.rb
+++ b/Library/Homebrew/dev-cmd/bump-formula-pr.rb
@@ -366,15 +366,12 @@ module Homebrew
             remote_url = response.fetch("clone_url")
           end
           username = response.fetch("owner").fetch("login")
-        rescue *GitHub.api_errors
+        rescue GitHub::AuthenticationFailedError => e
+          raise unless e.github_message =~ /forking is disabled/
           # If the repository is private, forking might be disabled.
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
What about instead of being an option defaulting to not using a fork on private repositories?

alyssais:
Works for me. Didn't want to change behaviour without discussing first. Would mean an extra API call to check the visibility before attempting to create the fork.

An alternative would be to try to fork the repo, and if we get the "Validation Fails" message from GitHub, assume forking is disabled and try to use the base repo instead.

Thoughts?

MikeMcQuaid:
> An alternative would be to try to fork the repo, and if we get the "Validation Fails" message from GitHub, assume forking is disabled and try to use the base repo instead.

That seems good to me. Could do the API call then 👍

MikeMcQuaid:
Nice work, thanks @alyssais!

PR #5087, opened 2018-10-11
Cask: fix quarantine check when xattr is broken 

amyspark:

-----

Fixes Homebrew/homebrew-cask#53219.


reitermarkus:
Thanks, @amyspark!

How can `xattr` be broken? Should this be shown by `brew cask doctor`?

amyspark:
The quarantine check here does just that @reitermarkus 😉 
`xattr` can be broken very easily: `defaults write com.apple.versioner.python Version 2.6`. I don't remember clearly, but I think that my Macbook came with that version by default.

PR #5114, opened 2018-10-15
utils/github: add ValidationFailedError.

MikeMcQuaid:
These have specific errors returned so would be more useful to output.

CI seems to be failing consistently with this error so let's see what it's producing.

-----

### DIFF
--- a/Library/Homebrew/test/utils/github_spec.rb
+++ b/Library/Homebrew/test/utils/github_spec.rb
@@ -31,9 +31,12 @@ describe GitHub do
 
   describe "::search_issues", :needs_network do
     it "queries GitHub issues with the passed parameters" do
-      results = subject.search_issues("brew search", repo: "Homebrew/brew", author: "avetamine", is: "closed")
+      results = subject.search_issues("brew search",
+                                      repo: "Homebrew/legacy-homebrew",
+                                      author: "MikeMcQuaid",
+                                      is: "closed")
       expect(results).not_to be_empty
-      expect(results.last["title"]).to eq("brew search : 422 Unprocessable Entity")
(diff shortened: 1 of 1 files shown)


Library/Homebrew/test/utils/github_spec.rb
```diff
@@ -31,9 +31,11 @@
 
   describe "::search_issues", :needs_network do
     it "queries GitHub issues with the passed parameters" do
-      results = subject.search_issues("brew search", repo: "Homebrew/brew", author: "avetamine", is: "closed")
+      results = subject.search_issues("brew search",
+                                      repo: "Homebrew/brew",
+                                      author: "MikeMcQuaid",
+                                      is: "closed")
       expect(results).not_to be_empty
-      expect(results.last["title"]).to eq("brew search : 422 Unprocessable Entity")
```
reitermarkus: This should still be tested.

Library/Homebrew/test/utils/github_spec.rb
```diff
@@ -31,9 +31,12 @@
 
   describe "::search_issues", :needs_network do
     it "queries GitHub issues with the passed parameters" do
-      results = subject.search_issues("brew search", repo: "Homebrew/brew", author: "avetamine", is: "closed")
+      results = subject.search_issues("brew search",
+                                      repo: "Homebrew/legacy-homebrew",
+                                      author: "MikeMcQuaid",
+                                      is: "closed")
       expect(results).not_to be_empty
```
reitermarkus: Could actually remove this now, because the other test implies that it isn't empty.

MikeMcQuaid: If it is empty it'll be a less pleasant error

Issue #4962, opened 2018-09-21
Cleanup RuboCop todos


MikeMcQuaid:
We have a [`.rubocop_todo.yml`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/.rubocop_todo.yml) and [some TODOs in our `.rubocop.yml`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/.rubocop.yml#L36) that it'd be great to clean up.

These would be a good first issue to fix one of each of these TODO at a time in separate pull requests.

javian:
I get
```
➜  Homebrew git:(master) rubocop
.rubocop_todo.yml: RSpec/FilePath has the wrong namespace - should be Rails
Warning: unrecognized cop RSpec/AnyInstance found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/ContextWording found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/DescribeClass found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/ExampleLength found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/ExpectActual found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/ExpectInHook found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/InstanceVariable found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/MessageSpies found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/MissingExampleGroupArgument found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/MultipleDescribes found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/MultipleExpectations found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/NamedSubject found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/NestedGroups found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/RepeatedDescription found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/RepeatedExample found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/SubjectStub found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/VerifiedDoubles found in .rubocop_todo.yml
Warning: unrecognized cop RSpec/AnyInstance found in .rubocop.yml
Warning: unrecognized cop RSpec/ContextWording found in .rubocop.yml
[...]

PR #5119, opened 2018-10-16
rubocop: fix some rubocop offences

javian:

-----
Fixes a couple of the rubocop offences currently in the tests. First batch of trying to wear down #4962.

### DIFF
--- a/Library/Homebrew/test/.rubocop_todo.yml
+++ b/Library/Homebrew/test/.rubocop_todo.yml
@@ -121,17 +121,6 @@ RSpec/RepeatedDescription:
     - 'rubocops/lines_cop_spec.rb'
     - 'tab_spec.rb'
 
-# Offense count: 10
-RSpec/RepeatedExample:
-  Exclude:
-    - 'compiler_selector_spec.rb'
-    - 'utils/shell_spec.rb'
-
-# Offense count: 2
-RSpec/ScatteredLet:
-  Exclude:
-    - 'cask/artifact/uninstall_zap_shared_examples.rb'
-
 # Offense count: 31
 RSpec/SubjectStub:
   Exclude:
--- a/Library/Homebrew/test/utils/shell_spec.rb
+++ b/Library/Homebrew/test/utils/shell_spec.rb
@@ -7,7 +7,7 @@ describe Utils::Shell do
       expect(subject.profile).to eq("~/.bash_profile")
     end
 
-    it "returns ~/.bash_profile by default" do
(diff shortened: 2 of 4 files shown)


Library/Homebrew/test/.rubocop_todo.yml
```diff
@@ -121,17 +121,6 @@ RSpec/RepeatedDescription:
     - 'rubocops/lines_cop_spec.rb'
     - 'tab_spec.rb'
 
-# Offense count: 10
-RSpec/RepeatedExample:
```
MikeMcQuaid: Are these examples with exactly the same code or just the same `it`/name?

reitermarkus: These are not repeated, they are using a different setup, so I'm not sure what this is flagging.

MikeMcQuaid: Not the same `it` string?

reitermarkus: I don't think so. For example, one has “Sh“ and one has ”Bash“ at the end.

reitermarkus: Oh, is is indeed the same exact code as the one before …

Library/Homebrew/test/utils/shell_spec.rb
```diff
@@ -7,11 +7,6 @@
       expect(subject.profile).to eq("~/.bash_profile")
     end
 
-    it "returns ~/.bash_profile for Sh" do
-      ENV["SHELL"] = "/bin/another_shell"
```
reitermarkus: Change this to `/bin/sh` instead.

Library/Homebrew/test/.rubocop_todo.yml
javian: Yes, unless I've misunderstood something these should all be duplicate tests that I've removed.

Library/Homebrew/test/utils/shell_spec.rb
javian: Will do.

Library/Homebrew/test/.rubocop_todo.yml
javian: The code is (or was) duplicated with different `it` strings.

MikeMcQuaid: You haven't misunderstood, you're correct 👍

MikeMcQuaid:
Thanks again @javian!

PR #5120, opened 2018-10-17
add :desc for a cask to have description

ryenus:
Like a core Formulae, a cask should be allowed to have a description.

-----

### DIFF
--- a/Library/Homebrew/cask/dsl.rb
+++ b/Library/Homebrew/cask/dsl.rb
@@ -66,6 +66,7 @@ module Cask
       :language,
       :languages,
       :name,
+      :desc,
       :sha256,
       :staged_path,
       :url,


commitay:
@ryenus Please see https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md#homebrew-cask-is-not-a-discoverability-service

ryenus:
> @ryenus Please see https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/not_a_discoverability_service.md#homebrew-cask-is-not-a-discoverability-service

Well, this is more about a cask I installed quite a while ago, and I used it a lot back then, but at the moment forgot what it was for, or whether to still keep it, so I was thinking about having `brew cask info FOO` also providing a minimal description about cask `FOO` would get me back to it quickly, probably the most convenient approach, maybe the problem is that the cask name `FOO` doesn't convey its purpose enough, but there's not much we can do about the cask name, but a short description would do the best for that.

I think the no-commitment-on-discoverability is fine by itself, but not the same thing here. Is it possible for a reconsideration? Thank you @commitay

MikeMcQuaid:
> Is it possible for a reconsideration?

Note even if we did consider it we'd probably want to have something similar to what we did for formulae: it is done in a third party tap first and then we merge it when there's descriptions for all formulae (or in this case casks) rather than do it bit-by-bit.

PR #5137, opened 2018-10-19
Fix to deprecated S3DownloadStrategy

amancevice:
Remove references to removed require_aws_sdk method
See commit: [599ecc9b5ad7951b8ddc51490ebe93a976d43b29](https://github.com/Homebrew/brew/commit/599ecc9b5ad7951b8ddc51490ebe93a976d43b29)

I couldn't find any documentation on what exactly users are supposed to do with the warning: 

```
Calling S3DownloadStrategy is deprecated! Use maintaining S3DownloadStrategy in your own formula or tap instead.
```

Is there documentation on this?

-----

### DIFF
--- a/Library/Homebrew/compat/download_strategy.rb
+++ b/Library/Homebrew/compat/download_strategy.rb
@@ -27,6 +27,13 @@ class S3DownloadStrategy < CurlDownloadStrategy
     ENV["AWS_SECRET_ACCESS_KEY"] = ENV["HOMEBREW_AWS_SECRET_ACCESS_KEY"]
 
     begin
+      require "aws-sdk-s3"
+    rescue LoadError
+      Homebrew.install_gem! "aws-sdk-s3", "~> 1.8"
+      require "aws-sdk-s3"
+    end
+
+    begin
       signer = Aws::S3::Presigner.new
       s3url = signer.presigned_url :get_object, bucket: bucket, key: key
     rescue Aws::Sigv4::Errors::MissingCredentialsError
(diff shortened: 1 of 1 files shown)


Library/Homebrew/compat/download_strategy.rb
```diff
@@ -204,12 +211,6 @@ def scp_source
 class DownloadStrategyDetector
   class << self
     module Compat
-      def detect(url, using = nil)
-        strategy = super
-        require_aws_sdk if strategy == S3DownloadStrategy
```
reitermarkus: Was the `require_aws_sdk` method removed?

amancevice: Yeah, it looks like it got taken out during deprecation. I linked the commit in the PR description but here again for reference
https://github.com/Homebrew/brew/commit/599ecc9b5ad7951b8ddc51490ebe93a976d43b29

MikeMcQuaid: @amancevice My bad, could you re-add that method into this `Compat` module instead?

MikeMcQuaid:
> Use maintaining S3DownloadStrategy in your own formula or tap instead.

Whoops, this wording isn't ideal, updating in https://github.com/Homebrew/brew/pull/5138.



> I couldn't find any documentation on what exactly users are supposed to do with the warning:
> Is there documentation on this?

Hopefully the new messaging in https://github.com/Homebrew/brew/pull/5138 is a bit clearer. In short, you're meant to move this code into your formula (or tap if you use `require_relative`) as Homebrew is not going to maintain it any more.

I apologise that this is a relatively recent 180; ideally we'd have done this before accepting these download strategies but I think it's better for everyone if Homebrew isn't maintaining download strategies we don't actually use.

MikeMcQuaid:
Thanks again @amancevice!

Issue #5074, opened 2018-10-09
Download strategies should be able to be embedded in formulae 


MikeMcQuaid:
https://github.com/Homebrew/brew/pull/5070 and https://github.com/Homebrew/brew/pull/5038 are PRs to add new download strategies that will not be used by Homebrew/homebrew-core. Instead of merging PRs like this we should require the addition of a refactoring which will allow download strategies to be created in formulae like in `radare2.rb`. 

After this we should deprecate and then remove all download strategies not used in Homebrew/homebrew-core in favour of allowing taps to maintain them themselves in formulae or in files `require`d by multiple formulae.

MikeMcQuaid:
For example of a Requirement doing this see https://github.com/Homebrew/homebrew-core/blob/7d6810f30f3227ee6ee40b051f5bb6c6a288c366/Formula/radare2.rb#L1-L17

colindean:
I think this is the way to go, too.

I'd like to see the download strategies refactored a good bit in the
process, too. I tried to add a download strategy about 2 years ago but
gave up because I couldn't really understand what was going on super
well, plus my use case may have been a little niche. Granted, that was
two years ago and I was doing it in haste!
Plugable download strategies would be great but a fear that I have is
continue all reimplantation of the same strategy if people can't
discover the ones they want someplace.
Would there be a way to have download strategies installable just
like formula?

MikeMcQuaid:
> Would there be a way to have download strategies installable just like formula?

I don't think so. Basically pretty much all third-party download strategies are niche and I'm unconvinced the new proposed ones would be used by more than a handful for taps/organisations. If they duplicate work: fine.

reitermarkus:
While refactoring the download strategies I had something like this in mind: 

```ruby
url "https://foo.bar/,
    using: lambda do |url, destination, **options|
      # custom download logic
    end
```

This way we don't expose the private internals which makes it easier to refactor.

MikeMcQuaid:
@reitermarkus Yeh, something like that would be good 👍. It'd be nice if `using` could take a custom class, too, but that only requires a very limited number of methods to be defined.

reitermarkus:
Yeah, you could pass anything that provides a `#call` method, like `lambda` and `proc` do.

kralf:
Hi, as the author of #5308, I've initially considered doing something like this in the formula spec:
```
class MyFormula < Formula
  class MyDownloadStrategy < AbstractFileDownloadStrategy
[...]

PR #5141, opened 2018-10-20
Allow taps to require custom lib files

claui:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb). **Deferring tests until there’s some consensus on the feature.**

-----

## Overview

This PR is a simple implementation of what has been proposed in https://github.com/Homebrew/brew/issues/5074#issuecomment-429268642:

> to maintain download strategy specs alongside with a tap's formulae and to then be able to `require` them in the formulae.

## Details

The idea is to allow a tap to contain libraries, which then can be `require`d from formulae (or Casks). The main rationale is to free the core code base from `DownloadStrategies` that are not used in `homebrew-core` but there are numerous other benefits as well.

## Example

I’ve been using this code on private taps for a few months. Here’s a real-life example for a tap that contains a custom library named `lgog_downloader`, which I use in Casks to install games I’ve purchased from [GOG](https://www.gog.com/):

```
.
├── Casks
│   ├── gog-chuchel.rb
│   ├── gog-orwell.rb
│   └── gog-state-of-mind.rb
└── lib
    └── utils
        ├── gog
        │   ├── game.rb
        │   └── package.rb
        ├── gog.rb
        ├── lgog_downloader
        │   ├── constants.rb
        │   ├── game_details_cache.rb
        │   ├── game_details_map.rb
        │   └── lgog_downloader_error.rb
        └── lgog_downloader.rb
```

A snippet from the [gog-chuchel](https://www.gog.com/game/chuchel) cask (not using a `DownloadStrategy` but you get the idea):

```
cask 'gog-chuchel' do
  version '1.0.0'
  sha256 '6fe648eb33db79ce6997dcaca3c867a0fb08b0243fc2eea39b349911c542bca4'

  require 'utils/lgog_downloader'

  url do
    installer = Utils::LGOGDownloader
      .game('chuchel')
      .installer('en2installer0')
    Utils::LGOGDownloader.url(installer)
  end
  name 'CHUCHEL'
  homepage 'https://www.gog.com/game/chuchel'

  depends_on formula: 'lgogdownloader'
  container type: :naked

  app 'CHUCHEL.app'

  preflight do
    # […]
  end
end
```

A `require`d library can provide one or more `DownloadStrategies`, or just about any objects or classes you’d want. The above example shows a Cask but the same `require` works similarly in formulas.

## Security considerations

Security-wise, I can’t think of a threat model where this PR would introduce any vulnerabilities. I’d still invite @Homebrew/maintainers for a thorough review. Thanks!

### DIFF
--- a/Library/Homebrew/global.rb
+++ b/Library/Homebrew/global.rb
@@ -18,6 +18,7 @@ I18n.backend.store_translations :en, support: { array: { last_word_connector: " 
 
 ActiveSupport::Inflector.inflections(:en) do |inflect|
   inflect.irregular "formula", "formulae"
+  inflect.irregular "library", "libraries"
   inflect.irregular "is", "are"
   inflect.irregular "it", "they"
 end
@@ -156,6 +157,8 @@ require "utils"
 require "official_taps"
 require "tap"
 require "tap_constants"
+
+Tap.lib_directories.each { |lib_dir| LoadPath << lib_dir }
 
 if !ARGV.include?("--no-compat") && !ENV["HOMEBREW_NO_COMPAT"]
   require "compat"
--- a/Library/Homebrew/cmd/tap-info.rb
+++ b/Library/Homebrew/cmd/tap-info.rb
@@ -39,12 +39,14 @@ module Homebrew
       tap_count = 0
       formula_count = 0
       command_count = 0
+      library_count = 0
       pinned_count = 0
       private_count = 0
       Tap.each do |tap|
         tap_count += 1
         formula_count += tap.formula_files.size
         command_count += tap.command_files.size
+        library_count += 1 if tap.lib_dir?
         pinned_count += 1 if tap.pinned?
         private_count += 1 if tap.private?
       end
@@ -53,6 +55,7 @@ module Homebrew
       info += ", #{private_count} private"
       info += ", #{formula_count} #{"formula".pluralize(formula_count)}"
       info += ", #{command_count} #{"command".pluralize(command_count)}"
+      info += ", #{library_count} #{"library".pluralize(library_count)}"
       info += ", #{Tap::TAP_DIRECTORY.abv}" if Tap::TAP_DIRECTORY.directory?
       puts info
     else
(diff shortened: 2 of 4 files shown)


claui:
Comments and review especially welcome from @vitorgalvao, @adidalal, @alebcay, @amyspark, and @victorpopkov because this will affect Casks, too.

Library/Homebrew/global.rb
```diff
@@ -157,6 +158,8 @@ def auditing?
 require "tap"
 require "tap_constants"
 
+Tap.lib_directories.each { |lib_dir| LoadPath << lib_dir }
```
MikeMcQuaid: This would allow any tap to open any Homebrew class/module and modify the code within.

claui: That’s correct; is the same thing not already possible today from inside a formula/cask body?

MikeMcQuaid:
@claui Personally I'm 👎 on this for a few reasons.

- Taps may already use `lib` directories which this would start picking up
- This allows taps access to any part of the main `brew` Ruby process which would introduce security problems and the potential for monkeypatching
- This provides a much wider API than is necessary
- `require_relative` allows the functionality of including modules in formulae from within a tap without introducing any of the above. This approach has already been used successfully in existing taps.

Library/Homebrew/global.rb
MikeMcQuaid: @claui No, that is run in a sandboxed, forked process.

MikeMcQuaid:
@claui Shall we pass on this?

claui:
@MikeMcQuaid I’ve realized that my approach is not going to work, mainly in light of @sjackman’s comment:

> There's a few cases where the formula is not loaded out of its typical location on disk.

While that’s a very good point, I’m still determined to convince (not persuade) you folks that some kind of library feature is going to be useful to tap authors, and by extension to lots of users.

**My next idea** is to try a RubyGems-based approach, whose scope will be limited to Casks. I’m specifically thinking about vendoring bundled Gems per installed Cask inside `Caskroom`.

But that’s entirely for another day. ☺️ Closing here.

Issue #5149, opened 2018-10-21
Improve Homebrew::Assertions to provide access to stderr


bval:
I'd like to propose either improving the shell_output/pipe_output assertions to capture stderr as well as stdout, or creating new, similar assertions that provide access to stderr as well.

The motivation for this feature comes from [a request](https://github.com/Homebrew/homebrew-core/pull/32969#discussion_r226865362) from @fxcoudert that I test the error message from a newly installed binary. After looking into doing so, I discovered that shell_output does not capture stderr, and several formulas avoid using the built-in assertions in favor of their own implementations using the Open3 module or other implementations of popen.

Implementing this in Homebrew::Assertions would allow the formulas doing their own thing to be cleaned up to use a standardized method of testing stderr. with fewer dependencies making them potentially less brittle.

The alternative is to leave things as they are.

I'm willing to take a stab at this PR, but figured it merits some discussion on approach from experienced maintainers.

fxcoudert:
We often use `2>&1`, which works well and has the advantage of simplicity :)

bval:
> > We often use `2>&1`, which works well and has the advantage of simplicity :)
> 
> I agree with this 👍
> 
> > creating new, similar assertions that provide access to stderr as well.
> 
> This would need to be the case otherwise it will break a lot of existing tests.

👍both with keeping it simple, and I don't want to introduce anything that requires refactoring a bunch of tests.

Wouldn't just adding `2>&1` to the existing shell_output/pipe_output methods accomplish this simply without changing their signature and breaking anything? My cursory survey of the formulae using Open3.popen shows that they could all just be switched over to using shell_output if it included stderr by default. They're only doing it because the thing they need to check for is in stderr, not doing anything more complicated than that.

I'll try a PR that does that and see what `brew test-bot` has to say about it.

PR #5157, opened 2018-10-23
Capture stderr in Homebrew::Assertions methods

bval:

-----

This PR attempts to address #5149 to capture stderr as well as stdout in the methods provided by `Homebrew::Assertions`. It does so without changing the method signatures of any methods and _should_ not break any existing formulae.

A follow-up to this PR would clean up any formula in Homebrew/homebrew-core that utilize `Open3.popen` or `IO.popen` directly to access stderr.

### DIFF
--- a/Library/Homebrew/formula_assertions.rb
+++ b/Library/Homebrew/formula_assertions.rb
@@ -6,7 +6,7 @@ module Homebrew
     # Returns the output of running cmd, and asserts the exit status
     def shell_output(cmd, result = 0)
       ohai cmd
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> and _should_ not break any existing formulae.

As mentioned in https://github.com/Homebrew/brew/issues/5149 this seems very unlikely to not break existing formulae. You're going to change the output of everything that uses `shell_output` and prints `stderr` as well as `stdout`. This is a breaking change as a result and we'll need another solution.

Issue #4760, opened 2018-08-27
`brew cleanup` should be run automatically


MikeMcQuaid:
One of the most common things we hear on our Twitter account is people saying that they are pleasantly surprised that `brew cleanup` cleans up so much space. We've held off running it automatically in the past as it's not granular enough to be effectively scoped to a package and because we're concerned about linkage to old versions. 

We now have a cached linkage database and the `HOMEBREW_UPGRADE_CLEANUP` variable which cleans up a bunch more things related to a given formula which means we can start to consider having `HOMEBREW_UPGRADE_CLEANUP` run on every upgrade/install by default after ensuring that linkage won't be broken.

chdiza:
Will that shellvar still exist and be able to be turned off?

MikeMcQuaid:
@chdiza If someone can make a pitch as to why it should be toggleable off: I don't see a reason why not.

chdiza:
Because it is a disaster for anyone who wants to keep previous versions of things around.   And keeping previous versions around (at least for really important stuff) is good practice, because breakage is common (here I mainly mean breakage in upstream projects, not HB breakage---though that happens sometimes too).  There's a reason why `brew switch` exists.

To lose this fundmental ability---part of *managing* packages---because doing normal maintenance doesn't occur to some people would be a real shame.

Come to think of it, if such a non-toggleable change were to occur---and were this to mean that I can't carry e.g., `wget` 1.19.4 and 1.19.5 for a while until I'm sure that 1.19.5 fixed a problem with 1.19.4, or for something more mission-critical carry `rsync` 3.1.2 for a while until I'm sure that 3.1.3 won't ruin my life---I think that might mean the end of Homebrew for me.

It's also bad for anyone who tries to keep their HOMEBREW_CACHE in sync across multiple machines, but this is a much smaller concern.

There are ways to keep the aforementioned ability, like creating something akin to `pin` but that simply means "please don't automatically throw this away", but I'm sure they would involve much more code than adding the toggle would.

chdiza:
> Agreed, you've convinced me there should be an opt-out toggle for this.

Wonderful! :)

> Note: the vast majority of package managers don't keep old versions around indefinitely.

I would be surprised if they immediately force removal of old ones with no way to override this.  If they do, they're doing it wrong :)

[...]

PR #5165, opened 2018-10-23
Implement the HOMEBREW_INSTALL_CLEANUP env to trigger cleanup on reinstall/install/upgrade

Wojonatior:

-----

This PR is thethe implementation of the `HOMEBREW_INSTALL_CLEANUP` ENV and associated features as discussed in #4760.


Library/Homebrew/test/cmd/upgrade_spec.rb
```diff
@@ -17,4 +17,17 @@
     expect(HOMEBREW_CELLAR/"testball/0.1").to be_a_directory
     expect(HOMEBREW_CELLAR/"testball/0.0.1").not_to exist
   end
+
+  it "upgrades a Formula and cleans up old versions when `HOMEBREW_INSTALL_CLEANUP` is set" do
+    setup_test_formula "testball"
+    # allow(ENV).to receive(:[]).and_call_original
+    # allow(ENV).to receive(:[]).with("HOMEBREW_INSTALL_CLEANUP").and_return("1")
+    ENV["HOMEBREW_INSTALL_CLEANUP"] = "1"
```
Wojonatior: I've tried both manually setting into `ENV` and mocking it out as shown in the 2 commented lines. I wasn't able to get either working even though I could confirm the functionality manually.

Library/Homebrew/cmd/upgrade.rb
```diff
@@ -3,7 +3,7 @@
 #:
 #:    Options for the `install` command are also valid here.
 #:
-#:    If `--cleanup` is specified or `HOMEBREW_UPGRADE_CLEANUP` is set then remove
+#:    If `--cleanup` is specified, `HOMEBREW_INSTALL_CLEANUP` or `HOMEBREW_UPGRADE_CLEANUP` is set then remove
```
Wojonatior: As a clarification around the original feature, is the intent to keep both ENVs or to enhance what the old one did as well a rename it?

Library/Homebrew/test/cmd/upgrade_spec.rb
MikeMcQuaid: @Wojonatior Ideally we wouldn't add another integration test for this (they are really slow). Either add a unit test or omit one entirely.

Library/Homebrew/cmd/upgrade.rb
MikeMcQuaid: @Wojonatior Good question 👍 Let's remove the documentation for `HOMEBREW_UPGRADE_CLEANUP` in favour of `HOMEBREW_INSTALL_CLEANUP` but keep the `HOMEBREW_UPGRADE_CLEANUP` variable functional for people who are still using it.

Wojonatior:
I've added the formula cleanup to the `install` and `reinstall` commands. I wasn't sure if it was supposed to be the formula cleanup or the full/regular cleanup. The issue seems to discuss the formula cleanup, but I'm not sure what would need to be cleaned up from the formula on an `install`/`reinstall`

Library/Homebrew/cmd/install.rb
```diff
@@ -73,6 +73,9 @@
 #:
 #:    If `--git` (or `-g`) is passed, Homebrew will create a Git repository, useful for
 #:    creating patches to the software.
+#:
+#:    If `HOMEBREW_INSTALL_CLEANUP` is set then remove previously installed versions
+#:    of upgraded <formulae>.
```
MikeMcQuaid: This also will clean up the HOMEBREW_CACHE for that formula

MikeMcQuaid:
@Wojonatior After fixing the docs please run `brew man` and commit the results.

MikeMcQuaid:
> I wasn't sure if it was supposed to be the formula cleanup or the full/regular cleanup. 

Formula cleanup 👍. Could have regular cleanup occurring every e.g. 100 installs or something, though? Thoughts?

> The issue seems to discuss the formula cleanup, but I'm not sure what would need to be cleaned up from the formula on an `install`/`reinstall`

Nothing specifically but there may be leftover cache and cellar entries that it's worth cleaning up at this stage.

Wojonatior:
@MikeMcQuaid I updated the docs and the manpage. I like the idea of the occasional full clean. Would it make sense to count that in the formula clean or specifically for the install? Is there a pattern established somewhere for how I would persist the number of installs?

MikeMcQuaid:
> Would it make sense to count that in the formula clean or specifically for the install? Is there a pattern established somewhere for how I would persist the number of installs?

I was thinking that rather than being by installs it could be by days. `brew cleanup` could write (or `touch` if it already exists) a file like `HOMEBREW_CACHE/.cleaned` which the `mtime` is checked if `HOMEBREW_INSTALL_CLEANUP` is set and a full, non-formula `cleanup` could be run at that time if it hasn't been run for e.g. 30 days (that's how often `git gc` is run). If we want to be safer still that `brew cleanup` could also only remove files/directories that would be removed anyway _and_ are 30 days old (but that may be overkill).

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @Wojonatior! This is useful as-is so figured it was worth merging. Would still love to see the stuff we talked about in a follow-up PR!

Wojonatior:
@MikeMcQuaid Thanks! Sorry about the lack of communication on my part. I definitely want to make the follow-up PR for this, you should hopefully see something around it in the next week or so.

MikeMcQuaid:
@Wojonatior No worries, looking forward to it!

PR #5166, opened 2018-10-23
man: Restore args after generating man pages from respective cmd parsers

GauthamGoli:

-----
Fix: `brew man --fail-if-changed`

Reason why it wasn't working: Initializing `Homebrew::CLI::Parser` in `man.rb` to generate man-pages  for individual commands is resetting `Homebrew.args` , this happens only when `brew man` is run, nowhere else we initialize `Homebrew::CLI::Parser` of other commands except its own.

This is change fixes by restoring `Homebrew.args` after it gets reset.

### DIFF
--- a/Library/Homebrew/dev-cmd/man.rb
+++ b/Library/Homebrew/dev-cmd/man.rb
@@ -189,6 +189,7 @@ module Homebrew
   def generate_cmd_manpages(glob)
     cmd_paths = Pathname.glob(glob).sort
     man_page_lines = []
+    man_args = Homebrew.args
     cmd_paths.each do |cmd_path|
       begin
         cmd_parser = Homebrew.send(cmd_arg_parser(cmd_path))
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/man.rb
```diff
@@ -189,10 +189,12 @@ def target_path_to_format(target)
   def generate_cmd_manpages(glob)
     cmd_paths = Pathname.glob(glob).sort
     man_page_lines = []
+    man_args = Homebrew.args
     cmd_paths.each do |cmd_path|
       begin
         cmd_parser = Homebrew.send(cmd_arg_parser(cmd_path))
         man_page_lines << cmd_manpage_lines(cmd_parser).join
+        Homebrew.args = man_args
```
MikeMcQuaid: Does this need reset for every invocation of the loop?

GauthamGoli: No.

GauthamGoli: I thought within the loop there could be a method which might need `man_args`, but that doesn't happen here. So I'll move this line out of loop.

MikeMcQuaid:
Thanks @GauthamGoli!

PR #5169, opened 2018-10-24
Add Dockerfile

sjackman:


Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+	&& git remote set-url origin https://github.com/Homebrew/brew
+
+USER linuxbrew
+WORKDIR /home/linuxbrew
+ENV PATH=/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:$PATH \
+	SHELL=/bin/bash \
+	USER=linuxbrew
+
+# Install portable-ruby and tap homebrew/core.
+RUN HOMEBREW_NO_ANALYTICS=1 HOMEBREW_NO_AUTO_UPDATE=1 brew tap homebrew/core \
```
scpeters: Do you want to install `git` as well to fix the `brew doctor` complaint?

```suggestion
RUN HOMEBREW_NO_ANALYTICS=1 HOMEBREW_NO_AUTO_UPDATE=1 brew tap homebrew/core \
	&& brew install git \
```

scpeters: I guess that does bring in a few dependencies with it

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:xenial
+LABEL maintainer="Shaun Jackman <sjackman@gmail.com>"
+
+RUN apt-get update \
+	&& apt-get install -y --no-install-recommends bzip2 ca-certificates curl file fonts-dejavu-core g++ git locales make openssh-client patch sudo uuid-runtime \
```
MikeMcQuaid: Can this be split into a few lines? Maybe even one per line?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+	&& rm -rf /var/lib/apt/lists/*
```
MikeMcQuaid: What's this for? If cleaning: is there an `apt` command that can be used instead?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+RUN localedef -i en_US -f UTF-8 en_US.UTF-8 \
+	&& useradd -m -s /bin/bash linuxbrew \
+	&& echo 'linuxbrew ALL=(ALL) NOPASSWD:ALL' >>/etc/sudoers
```
MikeMcQuaid: Is this standard practise for Docker?

Dockerfile
MikeMcQuaid: I agree that ideally this would have no `brew doctor` complaints 👍

Dockerfile
scpeters: This is what they recommend as a [best practice in the docker documentation](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#run). I'm not sure if there's an `apt` command that can be used instead; I've always seen it like this.

Dockerfile
scpeters: I didn't see it recommended on the official docker website, but I see something similar here:

* http://gbraad.nl/blog/non-root-user-inside-a-docker-container.html

Dockerfile
monouser7dig: it might increase the size https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-what-is-better

scpeters: it can still be a single `RUN` statement, but the packages can be written on separate lines with `\`; I think that's what @MikeMcQuaid was suggesting

chdiza:
I hope that this can instead be named `.Dockerfile`, so that users don't have this cluttering up the dir.

Same for the recently added azure thingy; can it be `.azure-piplelines.yml` instead?

The other CI/auxiliary support stuff in the root dir are all dotfiles: `.travis.yml`, `.yardopts`, `.github`, `.editorconfig`.

MikeMcQuaid:
> I hope that this can instead be named `.Dockerfile`, so that users don't have this cluttering up the dir.
> 
> Same for the recently added azure thingy; can it be `.azure-piplelines.yml` instead?

We're using the default names for both because almost all tooling expects them to have those names. Now that this stuff isn't in the `/usr/local` root there's not a strong argument to me for needing to hide things away in this repository compared to a normal one.

Dockerfile
sjackman: I would have expected `apt-get clean` to have an option to remove `/var/lib/apt/lists`, but I was not able to find such an option. See https://linux.die.net/man/8/apt-get

Dockerfile
sjackman: Standard practice for Docker is to leave the user as `root`, but `brew` refuses to run as root.

Minor style point, I can instead put this file in `/etc/sudoers.d/`. Any preference?

Dockerfile
sjackman: Done.

Dockerfile
sjackman: How about changing `check_git_version minimum_version` from 2.14.3 to 2.7.0? Ubuntu 16.04 LTS (Xenial Xerus) ships with git 2.7.4. That change could be for either macOS and Linux or only Linux, at your preference.

sjackman: The only `brew doctor` complain is…
```
Warning: An outdated version (2.7.4) of Git was detected in your PATH.
Git 2.14.3 or newer is required for Homebrew.
```

MikeMcQuaid: I'd rather we kept the Git requirement as it is; we do actually use features that make it require the current version. For the Docker image a PPA would likely suffice.

sjackman: Good idea!

scpeters: https://launchpad.net/~git-core/+archive/ubuntu/ppa seems like a good candidate

sjackman: Done!

Dockerfile
scpeters: > Minor style point, I can instead put this file in `/etc/sudoers.d/`. Any preference?

No preference from me; I defer.

sjackman:
`brew tests` and `brew test-bot` will be clean after PR https://github.com/Homebrew/brew/pull/5180.
@scpeters Care to review this PR for me? I'll merge this PR after that one, and then set up Docker Hub to build the image on each commit.

scpeters:
Sure, I'll review it in between meetings today.

sjackman:
Merged! I'll set up a Docker Hub build now.

scpeters:
nice work! I was just experimenting with it and noticed a problem with the `python@2` formula. Should I report that as an issue at https://github.com/Linuxbrew/homebrew-core/issues ?

sjackman:
Yes, please!

PR #5170, opened 2018-10-24
Keg::MUST_EXIST_DIRECTORIES: Frameworks is for macOS

sjackman:

-----

`Fremworks` need not exist on Linux.


sjackman:
Thanks for the review, Steven. Merged.

Library/Homebrew/extend/os/mac/keg.rb
```diff
@@ -0,0 +1,4 @@
+class Keg
+  GENERIC_KEG_LINK_DIRECTORIES = remove_const :KEG_LINK_DIRECTORIES
```
MikeMcQuaid: Changing constant definitions feels weird to me but I don't have a better suggestion (maybe @reitermarkus does)? 👍 to the PR as a whole, thanks @sjackman ❤️

sjackman: In the past we've simply changed the constant to a function, `keg_link_directories`. I have a mild preference for the constant, but I'm fine with either.

PR #5180, opened 2018-10-24
test/spec_helper: Fix :needs_svn

sjackman:
`which("svn")` detects `shims/scm/svn` even when svn is not installed.

-----

### DIFF
--- a/Library/Homebrew/test/spec_helper.rb
+++ b/Library/Homebrew/test/spec_helper.rb
@@ -98,7 +98,10 @@ RSpec.configure do |config|
   end
 
   config.before(:each, :needs_svn) do
-    skip "subversion not installed." unless which "svn"
+    homebrew_bin = File.dirname HOMEBREW_BREW_FILE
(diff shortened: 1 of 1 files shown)


sjackman:
> Ideally this would handle a Homebrew-installed SVN too.

I'm fine with making this change. On a more philosophical question though, should `brew tests` be affected by which formulae are installed on the host? Perhaps I'm over thinking it.

sjackman:
@MikeMcQuaid How do I get the original `HOMEBREW_PREFIX` in `test/spec_helper.rb`, rather than the one that is set to `Pathname(TEST_TMPDIR)/"prefix"`? I'm using `File.dirname(HOMEBREW_BREW_FILE)`. Good enough?

sjackman:
@MikeMcQuaid Ready for review.

scpeters:
fixes the test in the docker container for me

sjackman:
Thanks for the review, Steven. Merged.
@MikeMcQuaid Feel free to comment if you'd like to see changes.

PR #5181, opened 2018-10-24
cask: Don’t remove login items on upgrade.

reitermarkus:

---

Don't remove login items when upgrading an application.

### DIFF
--- a/Library/Homebrew/cask/artifact/abstract_uninstall.rb
+++ b/Library/Homebrew/cask/artifact/abstract_uninstall.rb
@@ -169,7 +169,9 @@ module Cask
         end
       end
 
-      def uninstall_login_item(*login_items, command: nil, **_)
+      def uninstall_login_item(*login_items, command: nil, upgrade: false, **_)
+        return if upgrade
+
         login_items.each do |name|
           ohai "Removing login item #{name}"
           command.run!(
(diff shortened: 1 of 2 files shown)


vitorgalvao:
> Don't remove login items when upgrading an application.

Why not? What if the current login items are outdated/no longer needed? What’s the harm in removing them and letting the application add them again?

commitay:
> What if the current login items are outdated/no longer needed?

Good point.

> Don't remove login items when upgrading an application.

Is this intended to help with the `login_item` problems on mojave?

reitermarkus:
The reason for this is that every time I run `upgrade --greedy`, my Dropbox login item gets removed. You could argue that `login_item` should be in `zap` because it isn't added as part of the installation process.

vitorgalvao:
> You could argue that `login_item` should be in `zap` because it isn't added as part of the installation process.

If I uninstall an app, I may want to keep their settings, but I’ll never want to keep a broken login item. That justifies having `login_item` in `uninstall`.

I don‘t particularly agree with the change, but since you have a real need for it, I also don’t particularly oppose it.

Out of curiosity, when you upgrade Dropbox what do you need to do to readd the login item?

reitermarkus:
I prefer a broken login item to a broken setup. A login item changing is an exception and not the norm.

> Out of curiosity, when you upgrade Dropbox what do you need to do to readd the login item?

I have to start Dropbox manually, which I shouldn't have to. And of course I only realise this once I actually *need* documents which haven't synced yet.

PR #5182, opened 2018-10-25
Prepend shims path to PATH used by `system` and `safe_system`

EricFromCanada:

-----
While writing [this forum post](https://discourse.brew.sh/t/help-needed-to-install-under-osx-10-6-8/3253/4), I found that Git commands run by `system` and `safe_system` always use the system-provided `git`, as shown by running `brew ruby -e 'safe_system "git", "--version"'` on an OS X 10.11 system (which should always be using a brewed Git):
```
$ brew config
HOMEBREW_VERSION: 1.8.0-39-g3939908
HEAD: 39399085b113fb309507d51d0d7801490359b8da
Last commit: 44 minutes ago
Core tap HEAD: 55ecb060faead81c7ea5f075bc4939c1e6a1b39a
Core tap last commit: 8 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_DEV_CMD_RUN: 1
CPU: octa-core 64-bit haswell
Homebrew Ruby: 2.3.7 => /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.7/bin/ruby
Clang: 8.0 build 800
Git: 2.19.0 => /usr/local/opt/git/bin/git
Curl: 7.43.0 => /usr/bin/curl
Java: 1.8.0_111, 1.6.0_65-b14-468
macOS: 10.11.6-x86_64
CLT: 8.2.0.0.1.1480973914
Xcode: 8.2.1
XQuartz: 2.7.11 => /opt/X11

$ brew ruby -e 'safe_system "git", "--version"'
git version 2.10.1 (Apple Git-78)
```
Although this isn't as much of an issue on newer macOS versions, it does mean that a custom `git` path set by `HOMEBREW_GIT_PATH` won't be used.

This change prepends Homebrew's `git` shim to the `PATH` used by `system` and `safe_system`, instead of appending.

I also updated the manual to more accurately reflect how Homebrew decides which `git` (and `curl`) to use, and hid `HOMEBREW_GIT` from `brew config`'s output, since it isn't directly settable 
from the environment and is already printed beside the Git version number further down.


Library/Homebrew/manpages/brew.1.md.erb
```diff
@@ -181,21 +189,22 @@ Note that environment variables must have a value set to be detected. For exampl
...
+    system version. Automatically set if the system version of `curl` is too old.
 
   * `HOMEBREW_FORCE_VENDOR_RUBY`:
     If set, Homebrew will always use its vendored, relocatable Ruby version
     even if the system version of Ruby is new enough.
 
-  * `HOMEBREW_GIT`:
-    When using Git, Homebrew will use `GIT` if set,
-    a Homebrew-built Git if installed, or the system-provided binary.
+  * `HOMEBREW_GIT_PATH`:
```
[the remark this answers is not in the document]
EricFromCanada: Got it, so instead I fixed `brew.sh` to allow setting `HOMEBREW_GIT` via the environment.

Library/Homebrew/manpages/brew.1.md.erb
```diff
@@ -151,6 +152,13 @@ Note that environment variables must have a value set to be detected. For exampl
     If set, Homebrew will not pass `-q` when invoking `curl`(1) (which disables
     the use of `curlrc`).
 
+  * `HOMEBREW_CURL_PATH`:
```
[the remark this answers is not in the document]
EricFromCanada: Got it, so instead I fixed `brew.sh` to allow setting `HOMEBREW_CURL` via the environment.

sjackman:
If you had the time and inclination, it'd be great to test that this PR also fixes picking up the correct brewed `curl` and `git` on Linux systems where the brewed `curl` and `git` are too old. No worries if not.

Library/Homebrew/brew.sh
```diff
@@ -136,6 +136,9 @@ then
 elif [[ -n "$HOMEBREW_DEVELOPER" && -x "$HOMEBREW_CURL_PATH" ]]
 then
   HOMEBREW_CURL="$HOMEBREW_CURL_PATH"
+elif [[ -x "$HOMEBREW_CURL" ]]
+then
+  HOMEBREW_CURL="$HOMEBREW_CURL"
```
MikeMcQuaid: Were these variables not working at all without these changes? If so, I'm wondering if we remove the docs entirely and just keep the `FORCE` variables around. Thoughts @EricFromCanada?

EricFromCanada: Correct; originally, `brew.sh` would overwrite whatever values `HOMEBREW_CURL` and `HOMEBREW_GIT` had in the original environment with its own. I've no opinion either way.

MikeMcQuaid: @EricFromCanada Yeh, let's just remove these variables I reckon as no-one complained 👍 Thanks and sorry for the back and forth!

Library/Homebrew/brew.rb
```diff
@@ -37,7 +37,7 @@
   homebrew_path = PATH.new(ENV["HOMEBREW_PATH"])
 
   # Add SCM wrappers.
-  path.append(HOMEBREW_SHIMS_PATH/"scm")
+  path.prepend(HOMEBREW_SHIMS_PATH/"scm")
   homebrew_path.append(HOMEBREW_SHIMS_PATH/"scm")
```
EricFromCanada: Should this also be changed to `prepend`?

Library/Homebrew/brew.sh
```diff
@@ -114,6 +114,25 @@ else
   : "${HOMEBREW_OS_VERSION:=$(uname -r)}"
   HOMEBREW_OS_USER_AGENT_VERSION="$HOMEBREW_OS_VERSION"
 
+  # Ensure the system Curl is a version that supports modern HTTPS certificates.
+  HOMEBREW_SYSTEM_CURL_VERSION="$($(type -P curl) --version 2>/dev/null | head -n1 | awk '{print $2}')"
```
sjackman: I generally prefer `command -v` over `type -P` because the former is a standard and the latter is a bashism.

Library/Homebrew/brew.sh
```diff
@@ -114,6 +114,25 @@ else
...
+  # shellcheck disable=SC2086,SC2183
+  printf -v HOMEBREW_SYSTEM_CURL_VERSION_NUMERIC "%02d%02d%02d" ${HOMEBREW_SYSTEM_CURL_VERSION//./ }
+  if [[ "10#$HOMEBREW_SYSTEM_CURL_VERSION_NUMERIC" -lt "10#074100" ]]
+  then
+    HOMEBREW_SYSTEM_CURL_TOO_OLD="1"
+    HOMEBREW_FORCE_BREWED_CURL="1"
+  fi
+
+  # Ensure the system Git is at or newer than the version set in diagnostic.rb.
+  HOMEBREW_SYSTEM_GIT_VERSION="$($(type -P git) --version 2>/dev/null | awk '{print $3}')"
```
sjackman: Ditto.

sjackman:
Thanks, Eric!

CI says `brew man` is unhappy.

Library/Homebrew/brew.rb
EricFromCanada: Or to put it another way, is it likely that an external command would want to use the `git` specified by `HOMEBREW_GIT` (e.g. if set with `HOMEBREW_GIT_PATH`) rather than whichever is first in the user's `PATH`? I'd say yes.

Library/Homebrew/brew.sh
EricFromCanada: Good to know. Eventually migrating the other instances of `type -P` in the codebase might be ideal.

EricFromCanada:
Not sure why the tests on Linux aren't passing.

Library/Homebrew/brew.sh
MikeMcQuaid: @sjackman We're specifically using a lot of Bashisms to simplify the code so I think they are generally a pro rather than a con. Everywhere we `[[`: that's a Bashism.

MikeMcQuaid:
> Tested on Ubuntu 12 that if either are outdated or missing, brewed versions will be installed on the next `brew update`.

Question: what happens when we "force" but it's not available (i.e. `brew update` hasn't been run yet?) and `git` is used?

Library/Homebrew/brew.sh
```diff
@@ -114,6 +114,25 @@ else
...
+  then
+    HOMEBREW_SYSTEM_CURL_TOO_OLD="1"
+    HOMEBREW_FORCE_BREWED_CURL="1"
+  fi
+
+  # Ensure the system Git is at or newer than the version set in diagnostic.rb.
+  HOMEBREW_SYSTEM_GIT_VERSION="$($(command -v git) --version 2>/dev/null | awk '{print $3}')"
+  # shellcheck disable=SC2086,SC2183
+  printf -v HOMEBREW_SYSTEM_GIT_VERSION_NUMERIC "%02d%02d%02d%02d" ${HOMEBREW_SYSTEM_GIT_VERSION//./ }
+  if [[ "10#$HOMEBREW_SYSTEM_GIT_VERSION_NUMERIC" -lt "10#02070000" ]]
```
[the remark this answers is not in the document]
EricFromCanada: Ensures the numbers are compared using base 10. Otherwise the leading zero makes them read as octal.

MikeMcQuaid:
> Not sure why the tests on Linux aren't passing.

Check the output for `FAILED`: `brew style` is failing due to the use of `-lt` for a non-integer comparison.

Library/Homebrew/brew.sh
```diff
@@ -114,6 +114,25 @@ else
   : "${HOMEBREW_OS_VERSION:=$(uname -r)}"
   HOMEBREW_OS_USER_AGENT_VERSION="$HOMEBREW_OS_VERSION"
 
+  # Ensure the system Curl is a version that supports modern HTTPS certificates.
+  HOMEBREW_SYSTEM_CURL_VERSION="$($(command -v curl) --version 2>/dev/null | head -n1 | awk '{print $2}')"
```
MikeMcQuaid: Would be preferable to use Bash string operators (like the line below) rather than `awk` here (and for Git below): https://stackoverflow.com/a/3162500

Library/Homebrew/brew.sh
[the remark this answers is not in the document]
MikeMcQuaid: Can the leading zeros be stripped/avoided instead?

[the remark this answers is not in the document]
MikeMcQuaid: Another thought: rather than manually keeping this in sync ideally the logic in `diagnostic.rb` would use something from here instead of having versions hardcoded in both places.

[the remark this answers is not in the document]
MikeMcQuaid: `HOMEBREW_SYSTEM_GIT_TOO_OLD` or similar, perhaps.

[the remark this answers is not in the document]
EricFromCanada: Easily. Note to self: revisit this when `git` hits version 10.

Library/Homebrew/brew.sh
```diff
@@ -114,6 +114,25 @@ else
...
+  # shellcheck disable=SC2086,SC2183
```
MikeMcQuaid: Can you provide a similar comment to the previous usage above this explaining why it's needed?

Library/Homebrew/brew.sh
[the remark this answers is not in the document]
MikeMcQuaid: I think it's a very reasonable assumption that Git and Curl are both always >=1.0.0

EricFromCanada:
> Question: what happens when we "force" but it's not available (i.e. `brew update` hasn't been run yet?) and `git` is used?
```
$ brew config
HOMEBREW_VERSION: >=1.7.1 (shallow or no git repository)
ORIGIN: (none)
HEAD: (none)
Last commit: never
Core tap ORIGIN: (none)
Core tap HEAD: (none)
Core tap last commit: never
HOMEBREW_PREFIX: /home/linuxbrew/.linuxbrew
HOMEBREW_CACHE: /home/lassosoft/.cache/Homebrew
HOMEBREW_NO_ANALYTICS_THIS_RUN: 1
CPU: single-core 64-bit merom
Homebrew Ruby: 2.3.7 => /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.7/bin/ruby
Clang: N/A
Git: N/A
Curl: 7.22.0 => /usr/bin/curl
Kernel: Linux 3.13.0-117-generic x86_64 GNU/Linux
OS: Ubuntu 12.04.5 LTS (precise)
Host glibc: 2.15
/usr/bin/gcc: 4.6.3
glibc: N/A
gcc: N/A
xorg: N/A
$ brew update --force
==> Installing dependencies for git: patchelf, zlib, binutils, linux-headers, glibc, m4, gmp, mpfr, libmpc, isl@0.18, gcc, ncurses, gettext, bzip2, pcre2, openssl, curl, libbsd and expat
...
```
Seems to work.

MikeMcQuaid:
> Seems to work.

🆒

Library/Homebrew/brew.sh
EricFromCanada: Like one of these?
```
HOMEBREW_SYSTEM_CURL_VERSION="$($(command -v curl) --version 2>/dev/null | head -n1 | cut -d ' ' -f 2)"
HOMEBREW_SYSTEM_CURL_VERSION="$($(command -v curl) --version 2>/dev/null | head -c 12 | { read x; echo ${x#curl }; })"
```

EricFromCanada:
In brew.sh I've functioned out the conversion of dotted version numbers to integers for comparison and made it the authoritative source of minimum `curl` and `git` version numbers. I've confirmed that `brew doctor` correctly reports outdated Git on macOS and Linux, and that `brew config` on Linux now uses a brewed curl if `/usr/bin/curl` is outdated.

Library/Homebrew/brew.sh
MikeMcQuaid:
```bash
$ git_version_output=$(git --version)
$ git_version=${git_version##* }
$ echo $git_version
2.19.1
```

```bash
$ curl_version_output=$(curl --version)
$ curl_name_and_version=${curl_version_output%% \(*}
$ curl_version=${curl_name_and_version##* }
$ echo $curl_version
7.54.0
```

MikeMcQuaid: (Could use this format for `HOMEBREW_CURL_VERSION` too)

MikeMcQuaid: at https://github.com/Homebrew/brew/blob/d9918109f42cbca688d6432eb56603b92c9c9dc1/Library/Homebrew/brew.sh#L181

MikeMcQuaid: In fact they could share the `curl` invocation, too.

EricFromCanada: Not really, since the first `curl` in the `PATH` might differ from the one `HOMEBREW_CURL` is set to.

PR #5191, opened 2018-10-27
audit: make `throttled` a regular audit

commitay:

-----
This was originally made a strict audit to avoid failures when intentionally updating the formulae out of cycle. 

Currently it doesn't discourage contributors from submitting PRs as `bump-formula-pr` doesn't check strict by default and still needs to be enforced manually in PRs as CI also doesn't check strict.

If an out of cycle update is necessary I think the audit failure is acceptable.

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -705,7 +705,7 @@ module Homebrew
         next if formula.stable.nil?
 
         version = formula.stable.version.to_s.split(".").last.to_i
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Makes sense, thanks again @commitay!

commitay:
> `bump-formula-pr` doesn't check strict by default

This is incorrect, by default it doesn't run audit at all.

Thoughts on making it opt-out rather than opt-in?

MikeMcQuaid:
> Thoughts on making it opt-out rather than opt-in?

This seems like a good idea 👍. In general the defaults should be optimised to save the most work for Homebrew/core maintainers

PR #5209, opened 2018-10-29
bump-formula-pr: run audit by default

commitay:

-----
Have `bump-formula-pr` run `brew audit` by default with an option to skip the audit.


MikeMcQuaid:
@commitay Do you think we need the option at all or can it be 💀?

commitay:
It's probably not needed for us with `core` PRs but without it we might break `bump-formula-pr` for people using it in external taps.

MikeMcQuaid:
@commitay What do you think about removing and adding on request? Would rather only keep around options we know we or others are definitely using (and their workflow will need to change with this new option anyway).

commitay:
Their workflow will need to change but I think people should be able to use this without having to make their formulae audit compliant.

PR #5217, opened 2018-10-30
Check if cached downloads are up-to-date.

reitermarkus:

-----

Check the remote time against the cached download's time, and re-download it if it is outdated.

Closes https://github.com/Homebrew/homebrew-cask/pull/54162.

### DIFF
--- a/Library/Homebrew/download_strategy.rb
+++ b/Library/Homebrew/download_strategy.rb
@@ -287,12 +287,18 @@ class CurlDownloadStrategy < AbstractFileDownloadStrategy
 
       ohai "Downloading #{url}"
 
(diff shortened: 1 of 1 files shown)


Library/Homebrew/download_strategy.rb
```diff
@@ -421,9 +433,12 @@ def mirrors
 
   def resolve_url_and_basename(url)
```
MikeMcQuaid: Would be good to adjust the name to clarify this also returns a time. I think ideally this would now return a hash to improve readability; `resolve_url_and_basename(url)[0..1]` is a bit opaque.

PR #5238, opened 2018-11-01
docs: update instructions for OS X < 10.9

EricFromCanada:
- [N/A] Have you successfully run `brew style` with your changes locally?

-----
New instructions for anyone who wants to install on older versions of OS X that don't support TLS 1.2 out of the box. Tested on 10.5 - 10.9, although the current version of OpenSSL 1.0.2 fails to install on 10.5 and 10.6.

Paired with Homebrew/install#154.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -19,9 +19,10 @@ it does it too. And you have to confirm everything it will do before it starts.
 
 ## Alternative Installs
 
-### OS X Lion 10.7 and below
+### OS X Mountain Lion (10.8) and below
+Because GitHub now only allows clients that support TLS 1.2 to access repositories over HTTPS, the Homebrew installer will use the GIT protocol when run on systems older than OS X Mavericks (10.9). This requires the availability of a `git` binary, which can be provided by pre-installing the [Command Line Tools or Xcode](https://developer.apple.com/download/more/) on Lion or Mountain Lion, or a [prepackaged installer](https://code.google.com/archive/p/git-osx-installer/downloads) on Leopard or Snow Leopard. Homebrew will also require the Command Line Tools or Xcode in order to automatically compile and install a newer `curl` and `git` with support for TLS 1.2.
 
-Using the instructions on https://brew.sh or below whenever you call `curl` you must pass `--insecure` as an argument. This is because your system `curl` is too old to speak to GitHub using HTTPS. Don't worry, on the first `brew update` Homebrew will install a newer, more secure `curl` for your machine.
+Also note that when installing on OS X Leopard (10.5), you need to bypass its outdated built-in certificates by adding `--insecure` to the [installation command](https://brew.sh/#install)'s list of `curl` flags.
 
 ### Untar anywhere
 Just extract (or `git clone`) Homebrew wherever you want. Just avoid:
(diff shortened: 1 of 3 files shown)


docs/Installation.md
```diff
@@ -42,6 +38,25 @@ mkdir homebrew && curl -L https://github.com/Homebrew/brew/tarball/master | tar
 ### Multiple installations
 Create a Homebrew installation wherever you extract the tarball. Whichever `brew` command is called is where the packages will be installed. You can use this as you see fit, e.g. a system set of libs in `/usr/local` and tweaked formulae for development in `~/homebrew`.
 
+### OS X Mountain Lion (10.8) and below
```
MikeMcQuaid: I reckon it's worth keeping this at the top to avoid issues being filed. Thoughts?

docs/Installation.md
```diff
@@ -42,6 +38,25 @@ mkdir homebrew && curl -L https://github.com/Homebrew/brew/tarball/master | tar
...
+Because GitHub now only allows clients that support TLS 1.2 to access repositories, systems running anything older than OS X Mavericks (10.9) will need their Homebrew installation to be completed manually.
+
+1. Using a newer web browser or a separate machine, download the files at the following URLs (do not use Safari if its *Open "safe" files after downloading* option is enabled) and move each to `~/Downloads`:
```
MikeMcQuaid: Great work compiling these instructions. They kinda make me weep though so I'm wondering about modifying the installer to make life a bit easier. Could these be downloaded over an HTTP link, I wonder, and that be an optional parameter that's used? Alternatively on some versions we could use the system Git and the `git://` protocol perhaps?

docs/Installation.md
EricFromCanada: Makes sense.

docs/Installation.md
EricFromCanada: Each of the below GitHub download links always performs a 301 redirect to HTTPS if the request is made with HTTP, preventing < 10.9 from downloading with any built-in tools. But it turns out that checking out via SSH still works, which can be used on any system whose Xcode/CLT ships with Git (i.e. Snow Leopard / Xcode 4 and up), as long as its public key auth is already set up. It also turns out that Bintray still allows downloads over HTTP. I've revised these steps and they seem to test fine so far, although full testing will be delayed since my ESXi Mac's SSD just went bad. 🙈

Alternatively, if there was an official HTTP mirror for snapshots of `brew` and `homebrew-core`, the installer could be modified to grab those instead on < 10.9.

MikeMcQuaid: `git://` is preferable to `ssh` because of the lack of setup that's required. I reckon I (or you) could adjust the installer accordingly to automate some of what's mentioned here.

EricFromCanada: Take a look at the current state of Homebrew/install#154, in which installation is automated on < 10.9 as long as Git is preinstalled. If that's the approach you're after, this PR can be revised to eliminate the manual instructions.

MikeMcQuaid: @EricFromCanada Yeh, I definitely reckon something like that is a step in the right direction. I don't think it's unreasonable to expect a bit more manual intervention to be required to make the official installer work (i.e. preinstalling Xcode/CLT and/or Git) if it let's us simplify assumptions elsewhere 👍

MikeMcQuaid:
Thanks again @EricFromCanada!

PR #5243, opened 2018-11-02
Check if umask is 000 in brew doctor

vangorden:

-----
Users on Window Subsystem for Linux have a umask of 000 when logging into their shell causing a [sticky bit error](https://github.com/Linuxbrew/brew/issues/852) when updating and installing packages in Linuxbrew.  This is a [known issue](https://github.com/Microsoft/WSL/issues/352) in WSL.
 
This PR adds a check to `brew doctor` warning users if their umask is 000, and how to resolve it.


vangorden:
@sjackman

Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -48,6 +48,18 @@ def check_xdg_data_dirs
...
         EOS
       end
+
+      def check_umask_not_zero
+        return unless File.umask.zero?
+
+        <<~EOS
+          umask is currently set to 000. Directories created by Homebrew cannot
+          be world-writable. This issue can be resolved by adding umask 002 to
+          your #{shell_profile},
```
sjackman:
```suggestion
          your #{shell_profile}
```

Library/Homebrew/extend/os/linux/diagnostic.rb
```diff
@@ -48,6 +48,18 @@ def check_xdg_data_dirs
...
+          For example:
```
sjackman:
```suggestion

```

sjackman:
The error message is…
```
Error: An exception occured within a child process:
  ArgumentError: parent directory is world writable but not sticky
```
It's caused by having a umask of 0. The default umask on WSL is 0.

sjackman:
Thank you very much for your first contribution to Homebrew, Noah! Two minor niggles, and I'm ✅

Library/Homebrew/extend/os/linux/diagnostic.rb
vangorden: fixed

sjackman:
You can if you like squash these two commits down to a single commit.

sjackman:
I use `git rebase -i origin/master` to squash commits.

vangorden:
Squashed. Thanks for the tip!

PR #5247, opened 2018-11-05
Improve application quitting.

reitermarkus:

-----

Fixes the failure in https://github.com/Homebrew/homebrew-cask-drivers/pull/621, which is caused by the application still starting while we are already trying to quit it.

### DIFF
--- a/Library/Homebrew/cask/artifact/abstract_uninstall.rb
+++ b/Library/Homebrew/cask/artifact/abstract_uninstall.rb
@@ -107,25 +107,26 @@ module Cask
             next unless Pathname(service).exist?
 
             command.run!("/bin/launchctl", args: ["unload", "-w", "--", service], sudo: with_sudo)
-            command.run!("/bin/rm",        args: ["-f", "--", service], sudo: with_sudo)
+            command.run!("/bin/rm", args: ["-f", "--", service], sudo: with_sudo)
             sleep 1
           end
         end
       end
 
-      def running_processes(bundle_id, command: nil)
-        command.run!("/bin/launchctl", args: ["list"]).stdout.lines
-               .map { |line| line.chomp.split("\t") }
-               .map { |pid, state, id| [pid.to_i, state.to_i, id] }
-               .select do |(pid, _, id)|
-                 pid.nonzero? && id.match?(/^#{Regexp.escape(bundle_id)}($|\.\d+)/)
-               end
+      def running_processes(bundle_id)
+        system_command!("/bin/launchctl", args: ["list"])
+          .stdout.lines
+          .map { |line| line.chomp.split("\t") }
+          .map { |pid, state, id| [pid.to_i, state.to_i, id] }
+          .select do |(pid, _, id)|
+            pid.nonzero? && id.match?(/^#{Regexp.escape(bundle_id)}($|\.\d+)/)
+          end
       end
 
       # :quit/:signal must come before :kext so the kext will not be in use by a running process
       def uninstall_quit(*bundle_ids, command: nil, **_)
         bundle_ids.each do |bundle_id|
-          next if running_processes(bundle_id, command: command).empty?
+          next if running_processes(bundle_id).empty?
 
           unless User.current.gui?
             ohai "Not logged into a GUI; skipping quitting application ID '#{bundle_id}'."
@@ -138,7 +139,7 @@ module Cask
             Timeout.timeout(10) do
               Kernel.loop do
                 next unless quit(bundle_id).success?
-                if running_processes(bundle_id, command: command).empty?
+                if running_processes(bundle_id).empty?
                   puts "Application '#{bundle_id}' quit successfully."
                   break
                 end
@@ -187,7 +188,7 @@ module Cask
 
           signal, bundle_id = pair
           ohai "Signalling '#{signal}' to application ID '#{bundle_id}'"
-          pids = running_processes(bundle_id, command: command).map(&:first)
+          pids = running_processes(bundle_id).map(&:first)
           next unless pids.any?
 
           # Note that unlike :quit, signals are sent from the current user (not
@@ -207,13 +208,12 @@ module Cask
 
         login_items.each do |name|
           ohai "Removing login item #{name}"
-          command.run!(
+          system_command!(
             "osascript",
             args: [
               "-e",
               %Q(tell application "System Events" to delete every login item whose name is "#{name}"),
             ],
-            sudo: false,
           )
           sleep 1
         end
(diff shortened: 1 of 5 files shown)


Library/Homebrew/cask/artifact/abstract_uninstall.rb
```diff
@@ -130,21 +130,52 @@ def uninstall_quit(*bundle_ids, command: nil, **_)
...
               end
             end
           rescue Timeout::Error
+            opoo "Application '#{bundle_id}' did not quit."
             next
           end
         end
       end
 
+      def quit(bundle_id)
```
claui:
```suggestion
      def quit(bundle_id, command: nil)
```

Library/Homebrew/cask/artifact/abstract_uninstall.rb
```diff
@@ -130,21 +130,52 @@ def uninstall_quit(*bundle_ids, command: nil, **_)
...
-          ohai "Quitting application ID '#{bundle_id}'."
-          command.run!("/usr/bin/osascript", args: ["-e", %Q(tell application id "#{bundle_id}" to quit)], sudo: true)
+          ohai "Quitting application '#{bundle_id}'..."
 
           begin
-            Timeout.timeout(3) do
+            Timeout.timeout(10) do
               Kernel.loop do
-                break if running_processes(bundle_id, command: command).empty?
+                next unless quit(bundle_id).success?
```
claui:
```suggestion
                next unless quit(bundle_id, command: command).success?
```

Library/Homebrew/cask/artifact/abstract_uninstall.rb
```diff
@@ -130,21 +130,52 @@ def uninstall_quit(*bundle_ids, command: nil, **_)
...
+              if (app.running()) {
+                $.exit(1)
+              }
+            }
+
+            $.exit(0)
+          }
+        JAVASCRIPT
+
+        system_command "osascript", args:         ["-l", "JavaScript", "-e", script, bundle_id],
```
claui:
```suggestion
        command.run "osascript", args:         ["-l", "JavaScript", "-e", script, bundle_id],
```

Library/Homebrew/cask/artifact/abstract_uninstall.rb
```diff
@@ -130,21 +130,52 @@ def uninstall_quit(*bundle_ids, command: nil, **_)
...
+                                    print_stderr: false,
```
claui:
```suggestion
                                 print_stderr: false,
```

Library/Homebrew/cask/artifact/abstract_uninstall.rb
```diff
@@ -130,21 +130,52 @@ def uninstall_quit(*bundle_ids, command: nil, **_)
...
+                                    sudo:         true
```
claui:
```suggestion
                                 sudo:         true
```

Library/Homebrew/cask/artifact/abstract_uninstall.rb
reitermarkus: I'm actually trying to get rid of everything other than `system_command`, i.e. unify all the `system`, `quiet_system` and `command.run` calls.

claui: How are we going to deal with the numerous tests that rely on `fake_system_command`?

reitermarkus: Refactor them as we go. The ones using the `FakeSystemCommand` class are not idiomatic RSpec anyways.

claui: You’re right; your solution is indeed more idiomatic. It also makes both the tests and the code under test simpler and easier to read! 👍

Library/Homebrew/test/cask/artifact/shared_examples/uninstall_zap.rb
```diff
@@ -89,43 +91,65 @@
...
+      allow(subject).to receive(:quit).with(bundle_id)
+                                      .and_return(instance_double("SystemCommand::Result", success?: false))
+
+      time = Benchmark.measure do
+        expect {
+          subject.public_send(:"#{artifact_dsl_key}_phase", command: fake_system_command)
+        }.to output(/Application 'my.fancy.package.app' did not quit\./).to_stderr
+      end
+
+      expect(time.real).to be_within(0.5).of(10)
```
claui: Make the upper bound less strict; `timeout` may take longer if the OS preempts (or blocks on) our process.

```suggestion
      expected_time = 10
      expect(time.real).to be_between(expected_time - 0.5, expected_time + 5)
```

Library/Homebrew/test/cask/artifact/shared_examples/uninstall_zap.rb
```diff
@@ -89,43 +91,65 @@
...
-      subject
+      expect {
+        subject.public_send(:"#{artifact_dsl_key}_phase", command: fake_system_command)
+      }.to output(/Application 'my.fancy.package.app' quit successfully\./).to_stdout
+    end
+
+    it "tries to quit the application for 10 seconds" do
+      allow(User.current).to receive(:gui?).and_return true
+
+      allow(subject).to receive(:running_processes).with(bundle_id).and_return([[0, "", bundle_id]])
```
claui: Your test is a much better approach than the end-to-end test code I had in mind!

For reference, the following script (which I’m going to abandon now) blocks the AppleScript command `tell application id "homebrew.cask.test.AppThatQuitsSlowly" to quit` for 15 seconds:

```
#!/usr/bin/python

import time
from AppKit import NSApplication, NSObject
from Foundation import NSBundle
from PyObjCTools import AppHelper

class AppThatQuitsSlowly:
    def start(self):
        info = NSBundle.mainBundle().infoDictionary()
        info['CFBundleIdentifier'] = 'homebrew.cask.test.AppThatQuitsSlowly'
        NSApplication.sharedApplication().setDelegate_(self._create_app_delegate().alloc().init())
        AppHelper.runEventLoop()

    def _create_app_delegate(self):
        class AppDelegate(NSObject):
            def applicationShouldTerminate_(self, notification):
                time.sleep(15)
                return True

        return AppDelegate

AppThatQuitsSlowly().start()
```

Just burying the script here in case it comes in handy for some future ad-hoc test 😄

Library/Homebrew/test/cask/artifact/shared_examples/uninstall_zap.rb
```diff
@@ -173,18 +196,21 @@
...
-      before(:each) do
+      before do
+        # rubocop:disable RSpec/AnyInstance
         allow_any_instance_of(Cask::Artifact::AbstractUninstall).to receive(:trash_paths)
           .and_wrap_original do |method, *args|
-            result = method.call(*args)
-            FileUtils.rm_rf result.stdout.split("\0")
+            method.call(*args).tap do |result|
+              FileUtils.rm_rf result.stdout.split("\0")
+            end
```
claui: Nice catch!

Library/Homebrew/test/cask/artifact/shared_examples/uninstall_zap.rb
reitermarkus: I'd rather keep this more readable for now and change it if it actually fails.

claui:
```suggestion
      expect(time.real).to be_within(3).of(10)
```

If readability is your concern, then make it 3 seconds, preserving symmetry.

PR #5262, opened 2018-11-06
home: Use CLI::Parser to parse args

GauthamGoli:

-----
#1860

### DIFF
--- a/Library/Homebrew/cmd/home.rb
+++ b/Library/Homebrew/cmd/home.rb
@@ -4,11 +4,27 @@
 #:  * `home` <formula>:
 #:    Open <formula>'s homepage in a browser.
 
+require "cli_parser"
+
 module Homebrew
   module_function
 
+  def home_args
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/home.rb
```diff
@@ -4,11 +4,28 @@
...
 
+  def home_args
+    Homebrew::CLI::Parser.new do
+      usage_banner <<~EOS
+        `home` [<formula>]
+
+        Open <formula>'s homepage in a browser. If no formula is provided,
+        open Homebrew's own homepage in a browser.
+      EOS
+      switch :verbose
```
MikeMcQuaid: @GauthamGoli What do you think about only adding the `verbose` switch where it results in behaviour or output change?

GauthamGoli: That'd be better and intuitive. Will remove where its redundant.

PR #5277, opened 2018-11-08
List updated packages below each other instead of as a comma separated list

DerKobe:

-----


MikeMcQuaid:
Could you explain your motivation for this? Thanks!

MikeMcQuaid:
@DerKobe Cool, thanks. We've not had feedback like this before and it would introduce some inconsistency with the output of other commands so passing on this now but we may reconsider in future. Sorry and thanks again!

PR #5310, opened 2018-11-11
DependencyCollector: Add java_dep_if_needed [Linux]

sjackman:
Use the `openjdk` formula if it's installed.
Use the host's Java if it's sufficient.
Otherwise install the `openjdk` formula.

### DIFF
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -79,6 +79,10 @@ class DependencyCollector
     Dependency.new("bzip2", tags) unless which("bzip2")
   end
 
+  def java_dep_if_needed(tags)
+    JavaRequirement.new(tags)
+  end
+
   def ld64_dep_if_needed(*); end
 
   def self.tar_needs_xz_dependency?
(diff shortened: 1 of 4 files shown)


sjackman:
This change is hopefully the last fix needed to transition `Brewsci/bio` to using Homebrew/brew rather than Linuxbrew/brew.

MikeMcQuaid:
This looks a lot like `default_formula` written another way. Are you sure it's sensible to allow either `openjdk` or any satisfaction of the Java requirement?

sjackman:
The `openjdk` keg is 170 MB, and many systems already have Java installed, so I was aiming to avoid installing it unnecessarily. `JavaRequirement` tests that the system-provided Java meets the required minimum version, so it seems sensible to me.

jonchang:
Anecdotally, I haven't had any issues relying on system Java on Linux on even a CentOS install, which is usually a few versions behind.

sjackman:
FWIW we've been using this patch in Linuxbrew/brew since 1.5.5 (2018-03-02).

MikeMcQuaid:
I’m away at a conference but will get back to you further ASAP when given more review and thoughts. Thanks!

MikeMcQuaid:
My objections still stand that I think having two paths for resolving Java is a bad idea but given it's Linux-only I don't care enough to block this being merged. :shipit:

sjackman:
Objection noted. Thanks for merging, Mike. I'm going to try migrating Brewsci/bio to Homebrew/brew now.

PR #5312, opened 2018-11-12
Fix RuboCop for Casks.

reitermarkus:

---

Split out RuboCop config used for Casks.

### DIFF
--- a/Library/.rubocop_cask.yml
+++ b/Library/.rubocop_cask.yml
@@ -1,0 +1,96 @@
+inherit_from: ./.rubocop_shared.yml
+
+# Cask Cops
+
+Cask/HomepageMatchesUrl:
+  Description: 'Ensure that the homepage and url match, otherwise add a comment. More info at https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/url.md#when-url-and-homepage-hostnames-differ-add-a-comment'
+  Enabled: true
+
+Cask/HomepageUrlTrailingSlash:
+  Description: 'Ensure that the homepage url has a slash after the domain name.'
+  Enabled: true
+
+Cask/NoDslVersion:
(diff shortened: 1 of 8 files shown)


Library/.rubocop_shared.yml
```diff
@@ -0,0 +1,7 @@
+AllCops:
```
MikeMcQuaid: What’s the point of this file? Given the cask file already duplicates much it seems strange to inherit to save the few lines this does when running ‘rubocop’ outside of ‘brew cask style’ in an editor will return wrong results again.

Library/.rubocop_cask.yml
```diff
@@ -0,0 +1,96 @@
...
+  Description: 'Ensure that cask stanzas are grouped correctly. More info at https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#stanza-order'
+  Enabled: true
+
+Cask/StanzaOrder:
+  Description: 'Ensure that cask stanzas are sorted correctly. More info at https://github.com/Homebrew/homebrew-cask/blob/master/CONTRIBUTING.md#stanza-order'
+  Enabled: true
+
+# Built-In Cops
+
+Layout/AlignHash:
```
MikeMcQuaid: A bunch of these values are identical to rubocop.yml or rubocop defaults. I spent a bunch of time checking both.

Library/.rubocop_shared.yml
reitermarkus: > when running ‘rubocop’ outside of ‘brew cask style’ in an editor will return wrong results again.

It doesn't return wrong results, each tap has a config file inheriting from `.rubocop_cask.yml`.

Library/.rubocop_cask.yml
reitermarkus: I will take another look and move the duplicated rules into `.rubocop_shared.yml`.

PR #5325, opened 2018-11-15
[missing_formula] direct heroku to heroku/brew

JBallin:
**yes but there are 30 unrelated 'offenses'**
- [x] Have you written new tests for your changes?

-----

`heroku` was removed in #5145 but users who haven't tapped `heroku/brew` will be confused when they try to `brew install heroku`:

```
$ brew install heroku
Error: No available formula with the name "heroku"
==> Searching for a previously deleted formula (in the last month)...
Warning: homebrew/core is shallow clone. To get complete history run:
 git -C "$(brew --repo homebrew/core)" fetch --unshallow

heroku was deleted from homebrew/core in commit f3c308ab:
 heroku: remove formula.

To show the formula before removal run:
 git -C "$(brew --repo homebrew/core)" show f3c308ab^:Formula/heroku.rb

If you still use this formula consider creating your own tap:
 https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap
```

I've added a message that directs them to install it via the `heroku/brew` tap:
```
$ brew install heroku
Error: No available formula with the name "heroku" 
We no longer package heroku but you can install it from heroku/brew's tap:
  brew install heroku/brew/heroku
```

### DIFF
--- a/Library/Homebrew/missing_formula.rb
+++ b/Library/Homebrew/missing_formula.rb
@@ -79,6 +79,10 @@ module Homebrew
           If you wish to use the 2.x release you can install with Homebrew Cask:
             brew cask install ngrok
         EOS
+        when "heroku" then <<~EOS
+          We no longer package heroku but it can be installed using heroku/brew's tap:
+            brew install heroku/brew/heroku
+        EOS
         end
       end
       alias generic_blacklisted_reason blacklisted_reason
--- a/Library/Homebrew/test/missing_formula_spec.rb
+++ b/Library/Homebrew/test/missing_formula_spec.rb
@@ -77,6 +77,10 @@ describe Homebrew::MissingFormula do
     specify "Xcode is blacklisted", :needs_macos do
       expect(%w[xcode Xcode]).to all be_blacklisted
     end
+
+    specify "heroku is blacklisted" do
+      expect("heroku").to be_blacklisted
+    end
   end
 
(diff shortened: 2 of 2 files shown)


MikeMcQuaid:
`tap_migrations.json` in Homebrew/homebrew-core would be the right fix for this but in this case we're intentionally not migrating people automatically. Thanks and sorry!

JBallin:
@MikeMcQuaid I agree they shouldn't be migrated automatically (which this doesn't do).

I was talking to a beginner yesterday who tried `brew install heroku` and got really confused at the output and ended up installing heroku using npm...I showed her that `brew install heroku/brew/heroku` is the officially recommended way to install the CLI.

I was hoping this would help other people like her. Is this just not worth special casing? Or is the issue that makes it seem like brew stands behind the tap, in which case we can change the language slightly?

JBallin:
@MikeMcQuaid, I'm aware of the issues with Heroku (I was the one that opened the [issue](https://github.com/heroku/homebrew-brew/issues/5), which has since been deleted by Heroku).

I still don't see why we wouldn't provide a helpful message saying "brew doesn't support `heroku` but heroku's official install is X".

MikeMcQuaid:
I don’t want to point to a tap that’s been actively adversarial towards this project and our maintainers. It suggests we support or at least encourage the use of it which is not the case.

PR #5333, opened 2018-11-20
Fix Language::Java::java_home_cmd for Linux

sjackman:
`/usr/libexec/java_home` is specific to macOS.

### DIFF
--- a/Library/Homebrew/language/java.rb
+++ b/Library/Homebrew/language/java.rb
@@ -1,16 +1,29 @@
 module Language
   module Java
-    def self.java_home_cmd(version = nil)
-      version_flag = " --version #{version}" if version
-      "/usr/libexec/java_home#{version_flag}"
+    def self.java_home_cmd(_ = nil)
+      # macOS provides /usr/libexec/java_home, but Linux does not.
+      raise NotImplementedError
+    end
+
+    def self.java_home(version = nil)
+      req = JavaRequirement.new [*version]
+      raise UnsatisfiedRequirements, req.message unless req.satisfied?
+      req.java_home
+    end
+
+    # @private
+    def self.java_home_shell(version = nil)
+      java_home(version).to_s
     end
 
     def self.java_home_env(version = nil)
-      { JAVA_HOME: "$(#{java_home_cmd(version)})" }
+      { JAVA_HOME: java_home_shell(version) }
     end
 
     def self.overridable_java_home_env(version = nil)
-      { JAVA_HOME: "${JAVA_HOME:-$(#{java_home_cmd(version)})}" }
+      { JAVA_HOME: "${JAVA_HOME:-#{java_home_shell(version)}}" }
     end
   end
 end
+
+require "extend/os/language/java"
--- a/Library/Homebrew/test/language/java_spec.rb
+++ b/Library/Homebrew/test/language/java_spec.rb
@@ -1,25 +1,37 @@
 require "language/java"
 
 describe Language::Java do
+  describe "::java_home" do
+    it "returns valid JAVA_HOME if version is specified", :needs_java do
+      java_home = described_class.java_home("1.8+")
+      expect(java_home/"bin/java").to be_an_executable
+    end
+
+    it "returns valid JAVA_HOME if version is not specified", :needs_java do
+      java_home = described_class.java_home
+      expect(java_home/"bin/java").to be_an_executable
+    end
+  end
+
   describe "::java_home_env" do
-    it "returns java_home path with version if version specified" do
+    it "returns java_home path with version if version specified", :needs_macos do
       java_home = described_class.java_home_env("blah")
       expect(java_home[:JAVA_HOME]).to include("--version blah")
     end
 
-    it "returns java_home path without version if version is not specified" do
+    it "returns java_home path without version if version is not specified", :needs_java do
       java_home = described_class.java_home_env
       expect(java_home[:JAVA_HOME]).not_to include("--version")
     end
   end
 
   describe "::overridable_java_home_env" do
-    it "returns java_home path with version if version specified" do
+    it "returns java_home path with version if version specified", :needs_macos do
       java_home = described_class.overridable_java_home_env("blah")
       expect(java_home[:JAVA_HOME]).to include("--version blah")
     end
 
-    it "returns java_home path without version if version is not specified" do
+    it "returns java_home path without version if version is not specified", :needs_java do
       java_home = described_class.overridable_java_home_env
       expect(java_home[:JAVA_HOME]).not_to include("--version")
     end
--- a/Library/Homebrew/test/spec_helper.rb
+++ b/Library/Homebrew/test/spec_helper.rb
@@ -85,6 +85,16 @@ RSpec.configure do |config|
     skip "Not on macOS." unless OS.mac?
   end
 
+  config.before(:each, :needs_java) do
+    java_installed = if OS.mac?
+      Utils.popen_read("/usr/libexec/java_home", "--failfast")
+      $CHILD_STATUS.success?
+    else
+      which("java")
+    end
+    skip "Java not installed." unless java_installed
+  end
+
   config.before(:each, :needs_python) do
     skip "Python not installed." unless which("python")
   end
--- a/Library/Homebrew/extend/os/language/java.rb
+++ b/Library/Homebrew/extend/os/language/java.rb
@@ -1,0 +1,1 @@
+require "extend/os/mac/language/java" if OS.mac?
--- a/Library/Homebrew/extend/os/mac/language/java.rb
+++ b/Library/Homebrew/extend/os/mac/language/java.rb
@@ -1,0 +1,18 @@
+module Language
+  module Java
+    def self.java_home_cmd(version = nil)
+      version_flag = " --version #{version}" if version
+      "/usr/libexec/java_home#{version_flag}"
+    end
+
+    def self.java_home(version = nil)
+      cmd = Language::Java.java_home_cmd(version)
+      Pathname.new Utils.popen_read(cmd).chomp
+    end
+
+    # @private
+    def self.java_home_shell(version = nil)
+      "$(#{java_home_cmd(version)})"
+    end
+  end
+end
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -356,7 +356,7 @@ class Pathname
   def write_jar_script(target_jar, script_name, java_opts = "", java_version: nil)
     mkpath
     java_home = if java_version
-      "JAVA_HOME=\"$(#{Language::Java.java_home_cmd(java_version)})\" "
+      "JAVA_HOME=\"#{Language::Java.java_home_shell(java_version)}\" "
     end
     join(script_name).write <<~SH
       #!/bin/bash
--- a/Library/Homebrew/requirements/java_requirement.rb
+++ b/Library/Homebrew/requirements/java_requirement.rb
@@ -1,6 +1,8 @@
 require "language/java"
 
 class JavaRequirement < Requirement
+  attr_reader :java_home
+
   fatal true
   download "https://www.oracle.com/technetwork/java/javase/downloads/index.html"
 


sjackman:
Linuxbrew/brew has been using this fix since 2015-03-07.

Library/Homebrew/language/java.rb
```diff
@@ -1,16 +1,15 @@
 module Language
   module Java
-    def self.java_home_cmd(version = nil)
-      version_flag = " --version #{version}" if version
-      "/usr/libexec/java_home#{version_flag}"
-    end
+    def self.java_home_cmd(_version = nil); end
```
MikeMcQuaid: I don't think this is a correct implementation. See the uses of this in Homebrew/homebrew-core; they won't work after this. It would be good to use something here that makes sense there.

Library/Homebrew/language/java.rb
```diff
@@ -1,16 +1,15 @@
...
 
-    def self.java_home_env(version = nil)
-      { JAVA_HOME: "$(#{java_home_cmd(version)})" }
+    def self.java_home_env(_version = nil)
```
MikeMcQuaid: I don't think this is a correct implementation. See the uses of this in Homebrew/homebrew-core; they won't work after this. It would be good to use something here that makes sense there.

Library/Homebrew/language/java.rb
```diff
@@ -1,16 +1,15 @@
...
+      {}
     end
 
-    def self.overridable_java_home_env(version = nil)
-      { JAVA_HOME: "${JAVA_HOME:-$(#{java_home_cmd(version)})}" }
+    def self.overridable_java_home_env(_version = nil)
```
MikeMcQuaid: I don't think this is a correct implementation. See the uses of this in Homebrew/homebrew-core; they won't work after this. It would be good to use something here that makes sense there.

Library/Homebrew/language/java.rb
```diff
@@ -2,7 +2,7 @@ module Language
   module Java
     def self.java_home_cmd(version = nil)
       version_flag = " --version #{version}" if version
-      "/usr/libexec/java_home#{version_flag}"
+      "#{HOMEBREW_BREW_FILE} java-home#{version_flag}"
```
MikeMcQuaid: A new documented command seems pretty overkill? Could perhaps do something like `raise NotImplementedError` on the generic/non-Mac implementation? Is this used in Linuxbrew formulae and needed to work?

sjackman: `write_jar_script` uses `Language::Java.java_home_cmd`, so for that reason, it does need to work. Making it an option of an existing command doesn't seem as clean to me as a new command.

MikeMcQuaid: What was the existing Linuxbrew implementation of this? We make `write_jar_script` use a more generically named function which used `java_home_cmd` only on macOS.

sjackman: The existing implementation of `java_home_cmd` returns `nil`, and the existing implementations of `java_home_env` and `overridable_java_home_env` return `{}`, which was the first version of this PR that I submitted. When `JAVA_HOME` is not defined, most tools will use the `java` interpreter found in `PATH`, which, when there is only a single Java interpreter installed, is good enough.

One place when I've seen this not work is when building a native JNI library, and the build script expects `JAVA_HOME` to be defined to find `$JAVA_HOME/include`.

sjackman:
The shell script created by `write_jar_script` looks like this:
```sh
#!/bin/bash
JAVA_HOME="$(/usr/libexec/java_home --version 1.8)" exec "/usr/local/Cellar/igv/2.4.14/libexec/igv.sh" "$@"
```
rather than being hard coded like for example this:
```sh
JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_73.jdk/Contents/Home exec "/usr/local/Cellar/igv/2.4.14/libexec/igv.sh" "$@"
```
The former is better, because the latter would break when the Java interpreter is upgraded.

I feel that providing a script that is functionally equivalent to macOS's `/usr/libexec/java_home` is the most straight forward approach to achieving this same functionality on Linux.

Library/Homebrew/language/java.rb
MikeMcQuaid: > The existing implementation of `java_home_cmd` returns `nil`, and the existing implementations of `java_home_env` and `overridable_java_home_env` return `{}`, which was the first version of this PR that I submitted.

I don't think those versions should be empty no-ops but I think a new command is a bit overkill in the opposite direction. The original implementation with `raise NotImplementedError` scattered around where appropriate feels like it may be a good call.

MikeMcQuaid:
> I feel that providing a script that is functionally equivalent to macOS's `/usr/libexec/java_home` is the most straight forward approach to achieving this same functionality on Linux.

I don't like the idea of having a new Homebrew command that's only relevant to make Linux formulae behave a bit more like Homebrew's. There must be a way that e.g. Debian and Fedora already handle something similar.

Library/Homebrew/language/java.rb
sjackman: This PR implements equivalent functionality to macOS, which I think is the best possible scenario. If your objection is that `brew java-home` is user-facing and documented, we can find a way to hide this tool so that it's internal to Homebrew.

sjackman:
This PR implements equivalent functionality to macOS, which I think is the best possible scenario. If your objection is that brew java-home is user-facing and documented, we can find a way to hide this tool so that it's internal to Homebrew.

sjackman:
I've pushed take three of this PR. It uses `/usr/libexec/java_home` on macOS. 
`Language::Java::java_home_cmd` is not implemented on Linux and raises `NotImplementedError`.
`Language::Java::java_home` uses `java_home_cmd` on macOS and `JavaRequirement.java_home` on Linux.
`Language::Java::java_home_env` uses `Language::Java::java_home`.

Library/Homebrew/language/java.rb
```diff
@@ -1,16 +1,23 @@
 module Language
   module Java
-    def self.java_home_cmd(version = nil)
-      version_flag = " --version #{version}" if version
-      "/usr/libexec/java_home#{version_flag}"
+    def self.java_home_cmd(_version = nil)
+      raise NotImplementedError
+    end
+
+    def self.java_home(version = nil)
```
MikeMcQuaid: Do you see this being a new method that's called by formulae? If not, could move it somewhere else less public or add a `@private` at least.

Library/Homebrew/test/language/java_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "language/java"
 
 describe Language::Java do
-  describe "::java_home_env" do
+  describe "::java_home_env", :needs_macos do
```
MikeMcQuaid: This should work on Linux, no?

Library/Homebrew/test/language/java_spec.rb
```diff
@@ -13,7 +13,7 @@
     end
   end
 
-  describe "::overridable_java_home_env" do
+  describe "::overridable_java_home_env", :needs_macos do
```
MikeMcQuaid: This should work on Linux, no?

Library/Homebrew/language/java.rb
sjackman: 9/14 uses of `java_home_cmd` look like…
```ruby
    cmd = Language::Java.java_home_cmd("1.8")
    ENV["JAVA_HOME"] = Utils.popen_read(cmd).chomp
```
A function that immediately evaluates to `java_home` (like the Linux implementation) seems useful.
```ruby
    ENV["JAVA_HOME"] = Language::Java.java_home("1.8")
```
That doesn't however work with the current macOS implementation of `java_home`, which evaluates to the string `"$(/usr/libexec/java_home …)"`, and not the actual home.

I'll take a stab at making a useful `java_home` function and hiding the private one that possibly evaluates to `$(…)` on macOS.

Library/Homebrew/test/language/java_spec.rb
sjackman: The current test which searches for the `version` string (e.g. `"1.8+"`) in `java_home` does not work. I can modify the test.

Library/Homebrew/test/language/java_spec.rb
sjackman: Ditto.

Library/Homebrew/language/java.rb
sjackman: On second though, it's not worth creating a new function for just 9 uses in Homebrew/core. I'll make `java_home` private.

Library/Homebrew/test/language/java_spec.rb
sjackman: Hmm. No, `:needs_macos` is more correct than `:needs_java`.
```ruby
java_home = described_class.java_home_env("blah")
```
On macOS this test is fine, but on Linux it fails unless Java version `blah` is installed, because on Linux it uses `JavaRequirement` to check that that particular version of Java is installed.

Library/Homebrew/language/java.rb
sjackman: I'm game for implementing it, but the current implementation is not sufficient. On macOS,
```ruby
ENV["JAVA_HOME"] = Language::Java.java_home("1.8")
```
would evaluate to 
```ruby
ENV["JAVA_HOME"] = "$(/usr/libexec/java_home --version 1.8"
```
which is incorrect. It should evaluate to
```ruby
ENV["JAVA_HOME"] = "/Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home"
```

Library/Homebrew/test/language/java_spec.rb
MikeMcQuaid: @sjackman Fine with me as long as we have new tests for each of the generic cases that are implemented.

Library/Homebrew/language/java.rb
sjackman: > I'm game for it being public if it'd be used for that. I think 9 uses is sufficient 👍

`Language::Java.java_home` is now implemented, public, and tested.

Library/Homebrew/test/language/java_spec.rb
sjackman: I've added a test for the new `Language::Java.java_home` method.

sjackman:
@MikeMcQuaid Thanks for the feedback. Good to merge?

Library/Homebrew/language/java.rb
```diff
@@ -1,16 +1,28 @@
 module Language
   module Java
-    def self.java_home_cmd(version = nil)
-      version_flag = " --version #{version}" if version
-      "/usr/libexec/java_home#{version_flag}"
+    def self.java_home_cmd(_version = nil)
+      raise NotImplementedError
```
MikeMcQuaid: A comment might be useful to explain why this is the case.

Library/Homebrew/language/java.rb
```diff
@@ -1,16 +1,28 @@
 module Language
   module Java
-    def self.java_home_cmd(version = nil)
-      version_flag = " --version #{version}" if version
-      "/usr/libexec/java_home#{version_flag}"
+    def self.java_home_cmd(_version = nil)
```
MikeMcQuaid: `_ = nil`

Library/Homebrew/test/spec_helper.rb
```diff
@@ -85,6 +85,10 @@
     skip "Not on macOS." unless OS.mac?
   end
 
+  config.before(:each, :needs_java) do
+    skip "Java not installed." unless which("java")
```
MikeMcQuaid: This will behave oddly on macOS with the `java` shim but Java not being installed. Might want to use similar logic to the `JavaRequirement`.

sjackman: Good catch. Fixed.

Library/Homebrew/test/spec_helper.rb
```diff
@@ -85,6 +85,16 @@
     skip "Not on macOS." unless OS.mac?
   end
 
+  config.before(:each, :needs_java) do
+    java_installed = if OS.mac?
+      Utils.popen_read("/usr/libexec/java_home", "--failfast")
+      $CHILD_STATUS.success?
```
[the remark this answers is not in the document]
sjackman: I used `Utils.popen_read` so that it has the standard behaviour of silencing `stderr` by default and showing it when `HOMEBREW_STDERR` is enabled.

sjackman:
> One final comment: feel free to address or merge as-is, don't feel strongly. Thanks for bearing with me here and sorry for all the back and forth. We're getting close 😍

Thanks for the review, Mike. Merged!

PR #5340, opened 2018-11-23
Use install(1) instead of mkdir(1) + chown(8)

rjc:
Two-step action can be reduced to a single step - fewer things to type, fewer mistakes to make :^)

-----

### DIFF
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -311,8 +311,7 @@ module Homebrew
           #{not_exist_dirs.join("\n")}
 
           You should create these directories and change their ownership to your account.
-            sudo mkdir -p #{not_exist_dirs.join(" ")}
-            sudo chown -R $(whoami) #{not_exist_dirs.join(" ")}
+            sudo install -d -o $(whoami) #{not_exist_dirs.join(" ")}
         EOS
       end
 
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -311,7 +311,7 @@ module Homebrew
           #{not_exist_dirs.join("\n")}
 
           You should create these directories and change their ownership to your account.
-            sudo install -d -o $(whoami) #{not_exist_dirs.join(" ")}
+            sudo install -d -o $(id -un) #{not_exist_dirs.join(" ")}
         EOS
       end
 
@@ -326,7 +326,7 @@ module Homebrew
           #{not_writable_dirs.join("\n")}
 
           You should change the ownership of these directories to your user.
-            sudo chown -R $(whoami) #{not_writable_dirs.join(" ")}
+            sudo chown -R $(id -un) #{not_writable_dirs.join(" ")}
         EOS
       end
 


rjc:
The 2nd commit replaces usage of the obsoleted whoami(1) utility with the id(1) one.

MikeMcQuaid:
Would rather leave this as-is, sorry. It's easier to understand any individual failures when it's two steps. We've also not had any problems with this approach.

rjc:
Understood. What about the 2nd commit - using utilities marked as *obsolete*?

vitorgalvao:
@MikeMcQuaid I agree, but maybe changing the `whoami` to `id -un` makes sense. `man whoami` says in the description:

> The whoami utility has been obsoleted by the id(1) utility, and is equivalent to ``id -un''.

MikeMcQuaid:
If it's changed consistently everywhere: we'll consider it. I don't personally see the need to change something that is working fine, though. I'd be interested whether that message still applies on e.g. 10.5 and in general whether it behaves identically in literally every situation on every OS. Regardless, I'm still not sure it's worth the change until it's actually fixing user issues.

As a personal preference I also find `whoami` infinitely more readable than `id -un` which doesn't even have the ability to do e.g. `id --effective-user-id --name` for readability.

rjc:
id(1) is a [`POSIX`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/id.html) utility - whoami(1) is **not**

Short, single-dash options, as specified by `POSIX`, are available on any `POSIX`-compliant OS - double-dash options are `GNU` extensions specific *mostly* to `Linux`.

And yes - `whoami` and `id -un` **are** equivalent - the latter construct is specifically for `whoami` backward-compatibility.

MikeMcQuaid:
> id(1) is a [`POSIX`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/id.html) utility - whoami(1) is **not**

This makes literally no difference to Homebrew which only targets macOS and Linux both of which provide `whoami`.

> Short, single-dash options, as specified by `POSIX`, are available on any `POSIX`-compliant OS - double-dash options are `GNU` extensions specific _mostly_ to `Linux`.

Many, many non-GNU tools provide double-dash options. They are significantly more readable which is why they are preferable. If no POSIX utilities ever provide them: that's a pretty strong argument _against_ us using them.

MikeMcQuaid:
> So are you interested in `id`'s behaviour on every OS or just `macOS` and `Linux`? :^)

I should have elaborated "every OS we attempt to support" which is just macOS and Linux.

vitorgalvao:
> apart from the fact that `whoami(1)` utility, which comes from `2BSD` BTW, has been deprecated for [_at least_ 25 years](https://www.freebsd.org/cgi/man.cgi?query=whoami)?

To me, that’s an argument to not stop using `whoami`. If it’s been deprecated for 25 years but it’s still here and working, will it ever be removed, really?

> > Many, many non-GNU tools provide double-dash options.
> 
> True... unfortunately.

You lost me there. Very few tools provide long-form options without a short-form equivalent, so what’s the problem?

> > They are significantly more readable which is why they are preferable.
> 
> Debatable. I prefer short, i.e. single-dash and single letter, options

Single letter options are fine when you’re writing commands interactively. When you’re typing a script, long-form options are better *every time* because they are clearer.

> quicker (less typing)

Which is irrelevant on a script. Comprehension (readability) later on is way more important.

> (long options are just asking for a typo!).

If you mistype a long-form option, you get an error. If you mistype a short-form option, you may activate a different option and screw up you work.

> At the end of the day, what are manual pages for, then!?

They’re definitely not for you to need to keep them open and reference multiple times when reading a single line of a script. With long-form options, you at least have a chance of understanding what is happening without needing to read the `man` pages.

Are you advocating that it’s fine for interfaces to be confusing as long as they have a manual? That literally gets people killed.

> > If no POSIX utilities ever provide them: that's a pretty strong argument _against_ us using them.
> 
> No sure if I understand correctly - strong argument **against** using `POSIX` utilities?

Yes. If being POSIX doesn’t offer any benefit over the alternative (in our use cases) but it does offer disadvantages (again, for our desired usage), there’s no point to using them.

rjc:
@MikeMcQuaid 

> I should have elaborated "every OS we attempt to support" which is just macOS and Linux.

Then there's no issue here - they behave **exactly** the same way. That's one of the arguments **for** `POSIX` - portability!

rjc:
@vitorgalvao 

> > No sure if I understand correctly - strong argument against using POSIX utilities?
>
> Yes. If being POSIX doesn’t offer any benefit over the alternative (in our use cases) but it does offer disadvantages (again, for our desired usage), there’s no point to using them.

You do know *what* `POSIX` is and *why* it is there, right?

vitorgalvao:
> You do know _what_ `POSIX` is and _why_ it is there, right?

I also know what the statue of liberty is and why it’s there. It doesn’t mean it’s useful for the goals of the project.

> Then there's no issue here - they behave **exactly** the same way. That's one of the arguments **for** `POSIX` - portability!

Again, portability for portability’s sake means nothing. We don’t need that portability for the systems we act on. We are arguing that *in this case*, readability trumps portability (because we don’t need that portability and care about readability).

rjc:
@vitorgalvao

> > apart from the fact that `whoami(1)` utility, which comes from `2BSD` BTW, has been deprecated for [_at least_ 25 years](https://www.freebsd.org/cgi/man.cgi?query=whoami)?
> 
> To me, that’s an argument to not stop using `whoami`. If it’s been deprecated for 25 years but it’s still here and working, will it ever be removed, really?

Utility has been deprecated - it even says so in the manual page -
`id(1)` is the replacement and is also portable (`POSIX`).

*- Nah, we're good - not changing*.

Do you see where I am going with this? Yes, that kind of approach
contributes to the above problem.

> > At the end of the day, what are manual pages for, then!?
> 
> They’re definitely not for you to need to keep them open and reference multiple times when reading a single line of a script. With long-form options, you at least have a chance of understanding what is happening without needing to read the `man` pages.
> 
> Are you advocating that it’s fine for interfaces to be confusing as long as they have a manual? That literally gets people killed.

I'm advocating for reading manual pages, lack of which I consider a bug.

rjc:
On Fri, Nov 23, 2018 at 06:00:01PM GMT, Vítor Galvão wrote:
> > You do know _what_ `POSIX` is and _why_ it is there, right?
> 
> I also know what the statue of liberty is and why it’s there. It doesn’t mean it’s useful for the goals of the project.
> 
> > Then there's no issue here - they behave **exactly** the same way. That's one of the arguments **for** `POSIX` - portability!
> 
> Again, portability for portability’s sake means nothing. We don’t need that portability for the systems we act on. We are arguing that *in this case*, readability trumps portability (because we don’t need that portability and care about readability).

Well, you support two distinct Unix families - the *BSDs and Linux - so
portability *is* important to you, at least at some level.
Ironically, `POSIX` made the job much easier! ;^)

Anyway, this was the part that has clearly confused me.

Also, I should have spotted `set +o posix` earlier.

Thanks for clarifying.

/me shrugs

rjc:
@vitorgalvao

> > apart from the fact that `whoami(1)` utility, which comes from `2BSD` BTW, has been deprecated for [_at least_ 25 years](https://www.freebsd.org/cgi/man.cgi?query=whoami)?
> 
> To me, that’s an argument to not stop using `whoami`. If it’s been deprecated for 25 years but it’s still here and working, will it ever be removed, really?

Utility has been deprecated - it even says so in the manual page -
`id(1)` is the replacement and is also portable (`POSIX`).

*- Nah, we're good - not changing*.

Do you see where I am going with this? Yes, that kind of approach
contributes to the above problem.

> > At the end of the day, what are manual pages for, then!?
> 
> They’re definitely not for you to need to keep them open and reference multiple times when reading a single line of a script. With long-form options, you at least have a chance of understanding what is happening without needing to read the `man` pages.
> 
> Are you advocating that it’s fine for interfaces to be confusing as long as they have a manual? That literally gets people killed.

I'm advocating for reading manual pages, lack of which I consider a bug.

rjc:
> > You do know _what_ `POSIX` is and _why_ it is there, right?
> 
> I also know what the statue of liberty is and why it’s there. It doesn’t mean it’s useful for the goals of the project.
> 
> > Then there's no issue here - they behave **exactly** the same way. That's one of the arguments **for** `POSIX` - portability!
> 
> Again, portability for portability’s sake means nothing. We don’t need that portability for the systems we act on. We are arguing that *in this case*, readability trumps portability (because we don’t need that portability and care about readability).

Well, you support two distinct Unix families - the *BSDs and Linux - so
portability *is* important to you, at least at some level.
Ironically, `POSIX` made the job much easier! ;^)

Anyway, this was the part that has clearly confused me.

Also, I should have spotted `set +o posix` earlier.

Thanks for clarifying.

/me shrugs

MikeMcQuaid:
Thanks for the conversation but let’s leave this one now.

PR #5343, opened 2018-11-24
Added Utils::Shell.subshell for embedding capturing subshells

kevinastone:
This should improve diagnostic messages on the fish-shell which doesn't
use `$()` for capturing the subshell output.

-----

### DIFF
--- a/Library/Homebrew/cmd/log.rb
+++ b/Library/Homebrew/cmd/log.rb
@@ -3,6 +3,7 @@
 #:    recognizes can be passed before the formula list.
 
 require "formula"
+require "utils/shell"
 
 module Homebrew
   module_function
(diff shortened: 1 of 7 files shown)


MikeMcQuaid:
I'm unconvinced it's worth altering the readability of all of these strings to support `fish` which is used by a very small minority of Homebrew users. @Homebrew/maintainers thoughts?

zbeekman:
In general I'm in favor of increasing portability and generality across shells; however, this case strikes me as borderline. The changes do not reduce legibility by much IMO, but I do think it will be hard to remember to *consistently* use `Utils::shell.subshell()` over `$()`. Ruby is great for its expressiveness, and syntactic sugar: If there were a more natural feeling or euphemistic way of accomplishing this without undue increase in complexity it could be worth it, IMO. I'll defer to others here, happy either way.

MikeMcQuaid:
Passing on this then, sorry @kevinastone!

reitermarkus:
> @Homebrew/maintainers thoughts?

I'm using `fish`, so I'd be in favour of this.

kevinastone:
Let me know if you think #5348 is more acceptable.

PR #5348, opened 2018-11-25
Added Shell Translation

kevinastone:
String#for_shell provides translation for capturing subshells to allow
    fish-shell compatible diagnostic messages.

Alternative to #5343

-----

### DIFF
--- a/Library/Homebrew/cmd/log.rb
+++ b/Library/Homebrew/cmd/log.rb
@@ -3,6 +3,9 @@
 #:    recognizes can be passed before the formula list.
 
 require "formula"
+require "utils/shell"
+
+using Utils::Shell
 
 module Homebrew
   module_function
@@ -31,7 +34,7 @@ module Homebrew
     end
 
     if File.exist? "#{repo}/.git/shallow"
-      opoo <<~EOS
+      opoo <<~EOS.for_shell
         #{name} is a shallow clone so only partial output will be shown.
         To get a full clone run:
           git -C "#{git_cd}" fetch --unshallow
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -5,6 +5,8 @@ require "version"
 require "development_tools"
 require "utils/shell"
 
+using Utils::Shell
+
 module Homebrew
   module Diagnostic
     def self.missing_deps(ff, hide = nil)
@@ -306,7 +308,7 @@ module Homebrew
         not_exist_dirs = Keg::MUST_EXIST_DIRECTORIES.reject(&:exist?)
         return if not_exist_dirs.empty?
 
-        <<~EOS
+        <<~EOS.for_shell
           The following directories do not exist:
           #{not_exist_dirs.join("\n")}
 
@@ -322,7 +324,7 @@ module Homebrew
                                            .reject(&:writable_real?)
         return if not_writable_dirs.empty?
 
-        <<~EOS
+        <<~EOS.for_shell
           The following directories are not writable by your user:
           #{not_writable_dirs.join("\n")}
 
@@ -595,7 +597,7 @@ module Homebrew
         branch = coretap_path.git_branch
         return if branch.nil? || branch =~ /master/
 
-        <<~EOS
+        <<~EOS.for_shell
           Homebrew/homebrew-core is not on the master branch
 
           Check out the master branch by running:
--- a/Library/Homebrew/update_migrator.rb
+++ b/Library/Homebrew/update_migrator.rb
@@ -1,5 +1,8 @@
 require "cask/cask_loader"
 require "cask/download"
+require "utils/shell"
+
+using Utils::Shell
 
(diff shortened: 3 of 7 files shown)


Library/Homebrew/update_migrator.rb
```diff
@@ -270,8 +271,9 @@ def migrate_legacy_repository_if_necessary
           You should change the ownership and permissions of #{HOMEBREW_PREFIX}
           temporarily back to your user account so we can complete the Homebrew
           repository migration:
-            sudo chown -R $(whoami) #{HOMEBREW_PREFIX}
+            sudo chown -R $('whoami') #{HOMEBREW_PREFIX}
```
MikeMcQuaid: Why does this one need a quote?

kevinastone: leftover typo...

zbeekman:
FWIW (not much) I like this implementation. Again I defer to the people who are actually spending most of their time working on `brew` and am happy to merge or pass.

Library/Homebrew/cmd/log.rb
```diff
@@ -22,10 +23,10 @@ def git_log(cd_dir, path = nil, tap = nil)
     repo = Utils.popen_read("git rev-parse --show-toplevel").chomp
     if tap
       name = tap.to_s
-      git_cd = "$(brew --repo #{tap})"
+      git_cd = "$(brew --repo #{tap})".for_shell
     elsif cd_dir == HOMEBREW_REPOSITORY
       name = "Homebrew/brew"
-      git_cd = "$(brew --repo)"
+      git_cd = "$(brew --repo)".for_shell
```
reitermarkus: There's a problem here. This is used below with surrounding quotes, however `"(brew --repo)"` in `fish` will just be a string, not a subshell.

Library/Homebrew/diagnostic.rb
```diff
@@ -601,6 +603,7 @@ def check_coretap_git_origin
           Check out the master branch by running:
             git -C "$(brew --repo homebrew/core)" checkout master
```
reitermarkus: Also using surrounding quotes here.

Library/Homebrew/cmd/log.rb
kevinastone: in `fish` it gets translated to `(brew --repo)` which should still be correct for the fish-shell.

reitermarkus: Look at the usage below, which is `"#{git_cd}"`.

kevinastone: ah, I see this issue.  Let me investigate....

kevinastone: I addressed the issue by removing the enclosing quotes for fish-shell (they're not needed in fish).

Library/Homebrew/diagnostic.rb
kevinastone: also fixed

kevinastone:
I addressed these comments (thanks!).  Keep in mind these shell translations are fairly hacky and fragile.

reitermarkus:
> Keep in mind these shell translations are fairly hacky and fragile.

Yes, then let's use the previous approach, but with a refinement.

```ruby
<<~EOS
  echo #{"whoami".subshell}
EOS
```

```ruby
refine String
  def subshell
    case Utils::Shell.preferred
    when :fish
      "(#{self})"
    else
      "\"$(#{self})\""
    end
  end
end
```

kevinastone:
> Yes, then let's use the previous approach, but with a refinement.

Nobody seemed to like that explicit substitution in #5343.

reitermarkus:
> Nobody seemed to like that explicit substitution in #5343.

I didn't actually comment before the original PR was closed. I don't think we can get around the fact that 

> it will be hard to remember to *consistently* use

I think, however, that explicit substitution is preferable to a “fairly hacky and fragile” regex. You still have to remember to actually use it regardless of where the substitution is done.

MikeMcQuaid:
> I think, however, that explicit substitution is preferable to a “fairly hacky and fragile” regex. You still have to remember to actually use it regardless of where the substitution is done.

I think in that case it's worth passing on not just the approach but the feature/change in general.

reitermarkus:
I'd still like to see this implemented, because it's just annoying having to correct a command after copy-pasting it. “remembering to actually use it” is not enough of a reason to pass on this. That reasoning could be applied to virtually anything.

MikeMcQuaid:
A stronger reason against this:

- our install instructions (https://brew.sh) use `$(...)`
- our docs consistently use `$(...)`
- our output instructions (which this PR seeks to change) consistently use `$(..)`

Combine this with:

- The above means `fish` users using Homebrew are already adjusting to the existing commands with little complaint
- All Homebrew's shell code assumes the use of Bash
- macOS does not ship with `fish`
- `fish` is not used by the (vast) majority of Homebrew users (https://formulae.brew.sh/analytics/install/90d/)

PR #5365, opened 2018-12-01
bottle: Mismatch in cellar is a warning

iMichka:

-----

When there is no previous linux bottle, and the bottle block looks like this:
```
bottle do
  sha256 "5cbcce4556b41ece91f01778068f481514bc3a0d5447ddbff048797236cc9b29" => :mojave
  sha256 "15748ef1ffefa088f7c958e6935c6eec794781858e6190196e18e94ab768adc2" => :high_sierra
  sha256 "49eb0e5d27b8868b2dba20efbde3ef75becc0cbf5ea230c00e5745e2df697cb6" => :sierra
end
```

then the bottling fails with:
```
Error: --keep-old was passed but there are changes in:
cellar: old: "/home/linuxbrew/.linuxbrew/Cellar", new: :any_skip_relocation
==> FAILED
```

This change allows to run the bottling step for these cases.

### DIFF
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -390,7 +390,10 @@ module Homebrew
       mismatches = [:root_url, :prefix, :cellar, :rebuild].reject do |key|
         old_spec.send(key) == bottle.send(key)
       end
-      mismatches.delete(:cellar) if old_spec.cellar == :any && bottle.cellar == :any_skip_relocation
+      if old_spec.cellar == :any && bottle.cellar == :any_skip_relocation
+        mismatches.delete(:cellar)
+        bottle.cellar :any
+      end
       unless mismatches.empty?
         bottle_path.unlink if bottle_path.exist?
 


iMichka:
@sjackman I am not sure I understood 100% of the code in there. I just regrouped code around the `mismatches.delete(:cellar)` line in a single `if/end` block.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -390,20 +390,32 @@ def bottle_formula(f)
       mismatches = [:root_url, :prefix, :cellar, :rebuild].reject do |key|
         old_spec.send(key) == bottle.send(key)
       end
-      mismatches.delete(:cellar) if old_spec.cellar == :any && bottle.cellar == :any_skip_relocation
+      if old_spec.cellar == :any && bottle.cellar == :any_skip_relocation
+        mismatches.delete(:cellar)
+        bottle.cellar :any if OS.linux?
```
MikeMcQuaid: We shouldn't scatter `if OS.linux?` around the codebase but instead use the `extend/` pattern. There's also no explanation on why this would be the case or why `:any_skip_relocation` should be used on Linux.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -390,20 +390,32 @@ def bottle_formula(f)
...
+      end
       unless mismatches.empty?
-        bottle_path.unlink if bottle_path.exist?
+        if bottle.prefix == "/home/linuxbrew/.linuxbrew"
```
MikeMcQuaid: `DEFAULT_PREFIX` should be used instead.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -390,20 +390,32 @@ def bottle_formula(f)
...
+          bottle.cellar old_spec.cellar
+          bottle.prefix old_spec.prefix
+          opoo <<~EOS
```
MikeMcQuaid: You're turning an error into a warning for no apparent reason.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -505,16 +517,31 @@ def merge
                 next if key == "cellar" && old_value == "any" && value == "any_skip_relocation"
                 next unless old_value.empty? || value != old_value
 
+                if key == "cellar" &&
+                   ["any", "any_skip_relocation"].include?(old_value) &&
+                   value == "/home/linuxbrew/.linuxbrew/Cellar"
```
MikeMcQuaid: `DEFAULT_PREFIX` should be used instead.

MikeMcQuaid:
> This change allows to run the bottling step for these cases.

Can you explain these changes in a lot more detail? This doesn't happen when there's a missing macOS bottle: why?

In general I really don't like the approach taken here of essentially turning off all the `cellar` bottling checks on Linux and turning failures into warnings. There's a reason that `cellar :any` and `cellar :any_no_relocation` are different. Bottles should be generated with the correct `cellar` values instead of hacking them after the fact to not through errors.

As the code already handles: it's acceptable to treat a `cellar :any_no_relocation` bottle as `cellar :any` but you can't do this with any other cases as it's incorrect and will break things.

sjackman:
When the macOS bottle is cellar :any, and the Linux bottles is not detected as cellar :any, in all but one case (curl), the bottle has been relocatable.

I see two ways of going forward.

1. Use an environment variable to downgrade this error to a warning when that environment variable is enabled, which will be enabled for bottling in Linuxbrew/core.

2. Fixing `brew bottle` to address the false positives that cause some Linux bottles to be identified as non-relocatable when they are in fact `cellar :any`.

The latter is clearly more work. I'd like to look at a handful of cases. Perhaps there's one or two causes, which we could address, that account for 99% of the cases. What was the last case that failed? I'll take a look at it.

Library/Homebrew/dev-cmd/bottle.rb
MikeMcQuaid: I'd need to understand why this fix is necessary to agree.

Library/Homebrew/dev-cmd/bottle.rb
MikeMcQuaid: There are various macOS bottles that we also consider to be manually relocatable but aren't detected as such. It's not acceptable to me to simply override this. Additionally, you've provided a counterpoint which implies there may well be more: how will these ever be detected? If Linuxbrew wishes to claim bottles are relocatable which are not detected as such this logic should live in (in descending order of preference) CI scripts, a new external command or `brew test-bot`. A particular Linuxbrew bottling decision like this should not result in incorrect/inconsistent bottling on other Linux systems. I'm also concerned about behaviour changes like this affecting Homebrew's bottling some of which is run on Linux.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -505,16 +517,31 @@ def merge
...
                   --keep-old was passed but there are changes in:
                   #{mismatches.join("\n")}
                 EOS
+                if ARGV.include? "--keep-going"
+                  opoo message
+                  update_or_add = nil
+                  break
+                else
+                  odie message
+                end
```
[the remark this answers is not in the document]
MikeMcQuaid: I would like to see it removed for the same reasons as above: it turns a legitimate error into an warning.

Library/Homebrew/dev-cmd/bottle.rb
sjackman: When the old bottle is `cellar :any`, and the new bottle is detected as `:any_skip_relocation`, then it's safe to mark the bottle as `cellar :any`.

Library/Homebrew/dev-cmd/bottle.rb
sjackman: The example I gave, `curl`, is not relocatable on Linux, because it's built using `openssl` on Linux, which causes it not to be relocatable. `curl-openssl` is also similarly not relocatable on macOS. That's the only exception I have run into. A tool that is relocatable on macOS is also relocatable on Linux, empirically true 99.92% of the time (1289/1290).

MikeMcQuaid: We've talked about this before in the context of relocatable bottles and I'm not convinced that things that simply pass `test do` when relocated have 100% of their functionality working perfectly when relocated. Again, that's a debate about improving the behaviour of relocation rather than simply overriding it in this case.

MikeMcQuaid:
> I see two ways of going forward.
> 
> 1. Use an environment variable to downgrade this error to a warning when that environment variable is enabled, which will be enabled for bottling in Linuxbrew/core.

As mentioned above I consider this to be simply ignoring a valid error case.

> 2. Fixing `brew bottle` to address the false positives that cause some Linux bottles to be identified as non-relocatable when they are in fact `cellar :any`.

This is what I see as the best of the two solutions.

Another option, as mentioned above, is that Linuxbrew/homebrew-core instead has a script that overrides the bottle JSON file to be `cellar :any` in the desired cases. This will not require any code changes and allows Linuxbrew/homebrew-core to adjust the behaviour until `brew bottle` is addressed.

To quote myself from above:

> I'm not convinced that things that simply pass test do when relocated have 100% of their functionality working perfectly when relocated.

If there's a programmatic solution that e.g. strips some ELF data or similar for 2) I can be convinced it's the correct solution.

I think this is a question of approach: I'd rather have bottles not flagged `cellar :any` when they could be rather than vice-versa as false negatives will not cause user errors but false positives will.

sjackman:
Michka pointed me to `topgrade`.
`topgrade` is detected as non-relocatable on macOS.
`topgrade` is detected as `cellar :any_skip_relocation` on Linux.
`brew bottle --keep-old topgrade` fails in Linuxbrew/core due to this mismatch.

```console
$ uname
Linux
$ brew bottle --keep-old -v topgrade
==> Detecting if topgrade--1.1.0.x86_64_linux.bottle.tar.gz is relocatable...
/home/linuxbrew/.linuxbrew/bin/patchelf --force-rpath --set-rpath /home/linuxbrew/.linuxbrew/Cellar/topgrade/1.1.0/lib:/home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/Cellar/topgrade/1.1.0/bin/topgrade
Error: --keep-old was passed but there are changes in:
cellar: old: "/home/linuxbrew/.linuxbrew/Cellar", new: :any_skip_relocation
```

```console
$ uname
Darwin
$ brew install topgrade
…
$ brew bottle -v topgrade
…
Warning: String '/usr/local/Cellar' still exists in these files:
/usr/local/Cellar/topgrade/1.1.0/bin/topgrade
 --> match '/usr/local/Cellar/rust/1.30.1/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-09705dba1dc68cd9.rlib(compiler_builtins-09705dba1dc68cd9.compiler_builtins.10ocm1pa-cgu.15.rcgu.o)' at offset 0x247a2c
…
$ strings /usr/local/bin/topgrade
…
__ZN74_$LT$topgrade..ctrlc..unix..RUNNING$u20$as$u20$core..ops..deref..Deref$GT$5deref11__stability4LAZY17h912228b0dc3ec236E
/private/tmp/rust-20181109-86056-qruqlr/rustc-1.30.1-src/src/rustc/compiler_builtins_shim/../../libcompiler_builtins/src/lib.rs/@/
compiler_builtins.10ocm1pa-cgu.15
/usr/local/Cellar/rust/1.30.1/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-09705dba1dc68cd9.rlib(compiler_builtins-09705dba1dc68cd9.compiler_builtins.10ocm1pa-cgu.15.rcgu.o)
___rust_probestack
rustc/compiler_builtins_shim/../../libcompiler_builtins/src/probestack.rs
__ZN197_$LT$serde..de..impls..$LT$impl$u20$serde..de..Deserialize$LT$$u27$de$GT$$u20$for$u20$alloc..vec..Vec$LT$T$GT$$GT$..deserialize..VecVisitor$LT$T$GT$$u20$as$u20$serde..de..Visitor$LT$$u27$de$GT$$GT$9visit_seq17h80f819d228ec7690E
…
$ du -h /usr/local/Cellar/rust/1.30.1/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-09705dba1dc68cd9.rlib
708K	/usr/local/Cellar/rust/1.30.1/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-09705dba1dc68cd9.rlib
$ topgrade -n
…
App Store: OK
$ brew remove rust
$ du -h /usr/local/Cellar/rust/1.30.1/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-09705dba1dc68cd9.rlib
du: /usr/local/Cellar/rust/1.30.1/lib/rustlib/x86_64-apple-darwin/lib/libcompiler_builtins-09705dba1dc68cd9.rlib: No such file or directory
$ topgrade -n
…
App Store: OK
```

Although the Rust executable `topgrade` contains an absolute path to a file in the `rust` package on macOS, that file is not needed for execution.

Library/Homebrew/dev-cmd/bottle.rb
sjackman: @iMichka Could you please pare this PR down to keeping just this one chunk?

MikeMcQuaid: Why is the bottle.cellar :any needed rather than deleting the mismatch?

sjackman:
```
      if old_spec.cellar == :any && bottle.cellar == :any_skip_relocation
        mismatches.delete(:cellar)
        bottle.cellar :any
      end
```
This section is for `--keep-old`. The previous value is `:any`. The current value is `:any_skip_relocation`. If the previous bottles are in fact `:any`, they will not work with `:any_skip_relocation. The current value must be changed to `:any`.

MikeMcQuaid: Gotcha 👍🏻

iMichka:
I updated the code, sorry for the delay. I tried to follow the discussion and I hope I got it right. I think that the effort needs to be put in solution 2:
> Fixing brew bottle to address the false positives that cause some Linux bottles to be identified as non-relocatable when they are in fact cellar :any.

Looks like @sjackman already found some things to improve here. We will need to give it another try with some bottles to see if this covers most of the cases.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -390,20 +390,32 @@ def bottle_formula(f)
       mismatches = [:root_url, :prefix, :cellar, :rebuild].reject do |key|
         old_spec.send(key) == bottle.send(key)
       end
-      mismatches.delete(:cellar) if old_spec.cellar == :any && bottle.cellar == :any_skip_relocation
+      if old_spec.cellar == :any && bottle.cellar == :any_skip_relocation
+        mismatches.delete(:cellar)
+        bottle.cellar :any
+      end
```
sjackman: @iMichka Could you please remove remove all diff chunks from this PR except for this one?

sjackman:
Merged. Thanks, Michka!

PR #5366, opened 2018-12-02
bottle: Skip matches to files in build deps

sjackman:
Files in build dependencies are not required at run time.

-----

This PR addresses the issue that `topgrade` is detected as being non-relocatable on macOS, when it ought to be `cellar :any_skip_relocation`.
For details see https://github.com/Homebrew/brew/pull/5365#issuecomment-443525087
cc @iMichka

### DIFF
--- a/Library/Homebrew/dev-cmd/bottle.rb
+++ b/Library/Homebrew/dev-cmd/bottle.rb
@@ -134,7 +134,7 @@ module Homebrew
     end
   end
 
-  def keg_contain?(string, keg, ignores)
+  def keg_contain?(string, keg, ignores, formula_and_runtime_deps_names = nil)
     @put_string_exists_header, @put_filenames = nil
 
     print_filename = lambda do |str, filename|
@@ -177,6 +177,19 @@ module Homebrew
 
           offset, match = str.split(" ", 2)
           next if linked_libraries.include? match # Don't bother reporting a string if it was found by otool
+
+          # Do not report matches to files that do not exist.
+          next unless File.exist? match
+
+          # Do not report matches to build dependencies.
+          if formula_and_runtime_deps_names.present?
+            begin
+              keg_name = Keg.for(Pathname.new(match)).name
+              next unless formula_and_runtime_deps_names.include? keg_name
+            rescue NotAKegError
+              nil
+            end
+          end
 
           result = true
           text_matches << [match, offset]
@@ -272,6 +285,7 @@ module Homebrew
 
     ohai "Bottling #{filename}..."
 
+    formula_and_runtime_deps_names = [f.name] + f.runtime_dependencies.map(&:name)
     keg = Keg.new(f.prefix)
     relocatable = false
     skip_relocation = false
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -178,6 +178,18 @@ def keg_contain?(string, keg, ignores)
           offset, match = str.split(" ", 2)
           next if linked_libraries.include? match # Don't bother reporting a string if it was found by otool
 
+          # Do not report matches to directories that do not exist.
```
MikeMcQuaid: Could this just check for paths that don't exist? Seems like it might be more specific and therefore more likely to find `cellar :any` relocations.

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -178,6 +178,18 @@ def keg_contain?(string, keg, ignores)
...
+          dir = Pathname.new(File.dirname(match))
+          next unless dir.directory?
+
+          # Do not report matches to build dependencies.
+          begin
+            keg_name = Keg.for(dir).name
```
MikeMcQuaid: Does this match cases where it's a random path inside a `Keg` rather than the actual keg?

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -178,6 +178,18 @@ def keg_contain?(string, keg, ignores)
...
+            next unless @runtime_deps.include? keg_name
```
MikeMcQuaid: Could this be passed through as an argument (to `keg_contain?`) instead? More obvious to me than relying on a instance variable declared elsewhere.

Library/Homebrew/dev-cmd/bottle.rb
sjackman: Yes

Library/Homebrew/dev-cmd/bottle.rb
sjackman: Sure

Library/Homebrew/dev-cmd/bottle.rb
sjackman: Done.

sjackman:
> Seems like a good approach, nice work 👍

Thanks, Mike!

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -178,6 +178,19 @@ def keg_contain?(string, keg, ignores)
           offset, match = str.split(" ", 2)
           next if linked_libraries.include? match # Don't bother reporting a string if it was found by otool
 
+          # Do not report matches to files that do not exist.
+          next unless File.exist? match
+
+          # Do not report matches to build dependencies.
+          if runtime_deps && !runtime_deps.empty?
```
MikeMcQuaid: Make `runtime_deps = []` by default or use `if runtime_deps.present?`

Library/Homebrew/dev-cmd/bottle.rb
```diff
@@ -272,6 +285,7 @@ def bottle_formula(f)
 
     ohai "Bottling #{filename}..."
 
+    runtime_deps = [f.name] + f.runtime_dependencies.map(&:name)
```
MikeMcQuaid: `formula_and_runtime_deps_names`?

sjackman:
@MikeMcQuaid Good to merge?

PR #5367, opened 2018-12-02
Add shims/super/gcc-8 and shims/super/g++-8

sjackman:

-----

These shims are conspicuously missing.


sjackman:
This missing `gcc-8` shim caused Superenv not to be used.
No superenv meant no `refurbish_args` to remove `cc -g`, which meant debugging symbols remained.
Debugging symbols caused the bottles of anything built with `gcc-8` to be detected as non-relocatable, when some may be `cellar :any`.

sjackman:
Note that `refurbish_args` is only used by default with the `make` build system. Formulae built with build systems other than `make` may still have debugging symbols, and may be incorrectly detected as not being relocatable.

sjackman:
Note that `shims/linux/super/gcc-8` from abac6dfaf8409695c367eef4c4722780c47362c5 `Add shims/linux/super`

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
> Would be good to perhaps use the GCC regex or similar to produce a simple test for the existence of these files so it doesn’t bite us in future?

Merged without this change, but yes very good idea.

PR #5371, opened 2018-12-03
Cask: Remove check on sha-only change

vitorgalvao:

-----

Removes the check for `sha256`-only (i.e. `version` stayed the same) on the Cask side.

Refs https://github.com/Homebrew/homebrew-cask/issues/48862. In short, this check is a well-intentioned wrong solution to the problem. We’re currently tackling the situation in better ways, namely quarantining `app` bundles and not allowing unsigned `pkg`s to run (that was already in place).

We install other types of software, true, but in those this situation almost never arises, and they’re less dangerous (e.g. font files).

### DIFF
--- a/Library/Homebrew/cask/audit.rb
+++ b/Library/Homebrew/cask/audit.rb
@@ -24,7 +24,6 @@ module Cask
 
     def run!
       check_required_stanzas
-      check_version_and_checksum
       check_version
       check_sha256
       check_url
@@ -134,31 +133,6 @@ module Cask
       # TODO: specific DSL knowledge should not be spread around in various files like this
       installable_artifacts = cask.artifacts.reject { |k| [:uninstall, :zap].include?(k) }
       add_error "at least one activatable artifact stanza is required" if installable_artifacts.empty?
-    end
-
-    def check_version_and_checksum
-      return if cask.sha256 == :no_check
-
-      return if @cask.sourcefile_path.nil?
-
-      tap = @cask.tap
-      return if tap.nil?
-
-      return if commit_range.nil?
-
-      previous_cask_contents = Git.last_revision_of_file(tap.path, @cask.sourcefile_path, before_commit: commit_range)
-      return if previous_cask_contents.empty?
-
-      begin
-        previous_cask = CaskLoader.load(previous_cask_contents)
-
-        return unless previous_cask.version == cask.version
-        return if previous_cask.sha256 == cask.sha256
-
-        add_error "only sha256 changed (see: https://github.com/Homebrew/homebrew-cask/blob/master/doc/cask_language_reference/stanzas/sha256.md)"
-      rescue CaskError => e
-        add_warning "Skipped version and checksum comparison. Reading previous version failed: #{e}"
-      end
     end
 
     def check_version


MikeMcQuaid:
I appreciate much discussion has already happened here but as this is a PR I feel it warrants some vague discussion here too. Feel free to link to previous comments rather than repeating yourself as desired.

This policy was added because of https://github.com/Homebrew/homebrew-cask/issues/19504 where Homebrew Cask was installing a version of Transmission which contain malware because there was no verification of a SHA256 change.

The main question I have (and the elephant in the room): what has changed since this is the case so this could no longer happen?

claui:
> The main question I have (and the elephant in the room): what has changed since this is the case so this could no longer happen?

I feel @vitorgalvao has made a very good case in https://github.com/Homebrew/homebrew-cask/issues/48862#issue-334976494:

> [The current policy] places a burden on maintainers and contributors that beared no fruit. Everyone is annoyed by it and no one wants to do it.

And further down in the same comment:

> It’s also easily circumventable, as an attacker that replaces a binary needs only change its version on the download page.

And the following tl;dr, if you ask me, pretty much hits the spot:

> It’s impossible for us to know when an app was tampered with and we need to set expectations accordingly. Just like we have --require-sha for security-minded users, we also need to make it their decision the risk of accepting or refusing non-codesingned apps.

That discussion more or less kindled the quarantine feature, which @amyspark went on to implement. I feel it would be a good thing if the burden of deciding whether or not to trust a piece of software were with the user. Now that Cask fully supports quarantining the downloaded artifacts, users can finally make use of Apple’s app trust model if they want. We have realized that with upstream binaries, we don’t have a snowball’s chance in hell to detect a server-side compromise. We need to act accordingly, and that includes not doing anything that could possibly blur that message.

MikeMcQuaid:
> [The current policy] places a burden on maintainers and contributors that beared no fruit. Everyone is annoyed by it and no one wants to do it.

The whole point of security policy is that the "fruit" is a lack of similar vulnerabilities.

> It’s also easily circumventable, as an attacker that replaces a binary needs only change its version on the download page.

I disagree strongly with this. The difference in attack between "replace a single file on a server without being noticed" and "replace a file on a server and add a new version onto a human-readable download page" is different. In many cases these require breaching multiple systems rather than a single one.

> It’s impossible for us to know when an app was tampered with and we need to set expectations accordingly. Just like we have --require-sha for security-minded users, we also need to make it their decision the risk of accepting or refusing non-codesingned apps.

Well, a SHA changing indicates an application has been changed. This alone may not indicate tampering but it does indicate something. Why bother to have SHA256 checksums at all if they are going to be changed blindly again? Use something like MD5 and make clear they are only to avoid download corruption or just evaluate the size of the expected download.

> Now that Cask fully supports quarantining the downloaded artifacts, users can finally make use of Apple’s app trust model if they want. We have realized that with upstream binaries, we don’t have a snowball’s chance in hell to detect a server-side compromise. We need to act accordingly, and that includes not doing anything that could possibly blur that message.

Again I have to ask: 

>  what has changed since this is the case so this could no longer happen?

To be explicit: is this "by default: nothing"? 

If you don't pass `--require-sha` are quarantines done for all packages by default?

---

As an aside, Homebrew Cask is of course able to run the project as they see fit but under the wider Homebrew umbrella and now that Homebrew Cask is an official part of Homebrew it's worth remembering that security incidents on Homebrew Cask affect the reputation of Homebrew (and vice versa). Part of this could be moving things to a `homebrew-cask-insecure` tap or similar to indicate where there's a difference here for cases where you feel upstream are not interested in working with you on a secure download mechanism.

claui:
These are the threat models I currently have in mind:

1. A malicious actor gains access to the upstream **development** infrastructure, either directly (source code write access, commit rights) or indirectly (compromising dependencies). No SHA check can possibly guard against that.

2. A malicious actor gains access to the upstream **release** infrastructure, for example by stealing an access token like in the recent es-lint incident. No SHA check can possibly guard against that.

3. A malicious actor gain access to an upstream **web server**. The attacker is then likely to go for the low-hanging fruit and just replace the binary package. That actor will probably not bother with version bumps, release pipelines or HTML.

@MikeMcQuaid Does that (roughly) match what you had in mind? Against which threats would you suggest that Homebrew Cask take action?

___

> Why bother to have SHA256 checksums at all if they are going to be changed blindly again? Use something like MD5

It’s not at all unusual to use SHA for checking integrity (and integrity only). One example is AUR’s [PKGBUILD system](https://wiki.archlinux.org/index.php/PKGBUILD#Integrity).

___

> and make clear they are only to avoid download corruption or just evaluate the size of the expected download.

That’s exactly what I hoped we would achieve by supporting quarantine, and by dropping what I’d call “authenticity theatre”.

___

> If you don't pass `--require-sha` are quarantines done for all packages by default?

Yes, thanks to @amyspark’s and @reitermarkus’s work, Cask now quarantines everything by default.

___

> To be explicit: is this "by default: nothing"?

If I recall correctly, it was the [es-lint incident](https://github.com/eslint/eslint-scope/issues/39) and the [discussions](https://github.com/Homebrew/brew/pull/4120) that [followed](https://github.com/Homebrew/homebrew-cask/issues/48862) that caused us to revisit the policy. I may be mistaken though.

___

> now that Homebrew Cask is an official part of Homebrew it's worth remembering that security incidents on Homebrew Cask affect the reputation of Homebrew (and vice versa).

(I wouldn’t even object against a common governance.) I feel it’s important that we acknowledge that Homebrew Cask will always have more security incidents than Homebrew proper. So reputation-wise, Cask is always going to be a liability for the Homebrew project as a whole, and it’s our responsibility to manage that. That’s why I appreciate this debate a lot! We can’t have enough eyes on that.

MikeMcQuaid:
I wrote a bunch more discussion but I deleted it to focus on what's important here (as it's not important whether or not we actually agree on this).

> Yes, thanks to @amyspark’s and @reitermarkus’s work, Cask now quarantines everything by default.

In the Transmission case: would this have resulted in a different outcome? Does the quarantine mean that once Apple updates their malware list that quarantined items are blocked from being run but non-quarantined items are not.

ran-dall:
@MikeMcQuaid I wasn't here for the Transmission problem, BUT if we're talking about the **KeRanger** or **Keydnap** incidents or something similar incident; then I don't think if we quarantined it; would've have made a difference because the app was "correctly" signed. 🤷‍♂️

>[It’s (the malware infected version of Transmission) different from the legitimate Transmission certificate, but is still signed by Apple and bypasses Gatekeeper protection.](https://www.welivesecurity.com/2016/08/30/osxkeydnap-spreads-via-signed-transmission-application/)

How now that **Notarized apps** have landed in **Mojave**, maybe things will be different? [Here's a good piece with a good stand point by **eclecticlight.com.**](https://eclecticlight.co/2018/09/24/has-that-app-been-notarized-and-what-are-the-benefits-to-the-user/)

> [When you install Mac apps, plug-ins, and installer packages from outside the App Store, macOS checks the Developer ID signature and notarization status to verify that the software is from an identified developer and that it has not been altered. With macOS Mojave, developers can also have their app notarised by Apple—an indication that the app was uploaded to Apple and passed a security check before it was distributed. ](https://support.apple.com/en-us/HT202491)

vitorgalvao:
> then I don't think if we quarantined it; would've have made a difference because the app was "correctly" signed.

That is exactly why quarantining *would* have made a difference. The app was signed, but not by the official developer, meaning Apple only needed to revoke the malicious key and the app no longer ran. Without quarantining, affected users could have been left in the dark indefinitely.

vitorgalvao:
@MikeMcQuaid Yes, *and* it would have done so without our interference (macOS would have done it).

Now, if Transmission wasn’t a signed app, there wouldn’t be much we could’ve done either way, *but* on first run users would have gotten a serious warning from macOS preventing the app from running. It’s easy to bypass that (i.e. say “yes, I don’t care, I’ll take the risk”) but it needs to be done deliberately ([our docs explain the process succinctly](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/the_app_cant_be_opened.md)). This is, I feel, an important step towards setting accurate expectations to users that while we react fast to problems, they’re the last line of responsibility regarding what runs on their machine. We are leveraging the system Apple built to be more effective at that\*.

I’ll add that one of the reasons nobody likes the old verification process is that legitimate `sha256`-only changes happen *all the time*. We get a crapton of version bumping PRs daily, and there’s always some of those. They take effort to verify, which leads to stalled PRs and several repeated issues (depending on the popularity of the cask).

The interesting part here is that our verification process consists of uploading the apps to VirusTotal to check if they’re correctly signed. We do it that way so the submitter can do the process themselves and send us the link (we can verify from the `sha256` on VirusTotal that the uploaded file is the same). So in essence, our new codesigning method is like doing the verification automatically on every user’s machine, where they get asked “do you want to take this risk?” and let them decide for themselves.

---

\* Not going to lie: they could definitely help us there by making quarantining an easier process from the CLI. I’ve opened a rdar a while ago (way before we implemented it ourselves) but it seems to have been largely ignored.

ran-dall:
@vitorgalvao Thank you for stepping in! 👌 Please correct my thinking...but, in the Transmission situation, the code **was signed** by a proper Apple developer. So it was passing quarantine and the gatekeeper checks, because it was only verifying the origin of the file and it didn't detect obvious malware. If Apple did revoked the certificate or updated their malware definitions, then that would only have prevent future installs; the people affected would have already passed gatekeeper. It's worth considering that it would take time for Apple to update their definitions, which just leaves many people 'in the wind' until those updates happen.

However to be being sensitive to @MikeMcQuaid comments about responsiblity, (also in my experience) I think that HBC **can** react much faster than Apple can being an community driven effort. If we find a problem, we can issue a PR fairly easily thru the 'glorious yet simple' `cask-repair`by you (@vitorgalvao ). Not to mention, that your tool, makes maintaining the `sha256` very easy. **IMO it this is a last line of defense**.
<hr>

#### Suggestion
I think it was suggested before, but could we possibly make it a requirement to updates thru `cask-repair`? Would that address most of the `sha256` issues and a another fair amount of the 'clapton' oversights that are sent in?

claui:
> I think it was suggested before, but could we possibly make it a requirement to updates thru `cask-repair`?

Please don’t. I can’t use `cask-repair`. My Mac is one of those affected by [this](https://github.com/vitorgalvao/tiny-scripts/issues/111#issuecomment-352260407) somewhat [rare issue](https://github.com/vitorgalvao/tiny-scripts/issues/125#issuecomment-379247754) and I’ve been unable to fix it ever since.

ran-dall:
> Please don’t. I can’t use `cask-repair`. My Mac is one of those affected by [this](https://github.com/vitorgalvao/tiny-scripts/issues/111#issuecomment-352260407) somewhat [rare issue](https://github.com/vitorgalvao/tiny-scripts/issues/125#issuecomment-379247754) and I’ve been unable to fix it ever since.

@claui AH! I wasn't aware of this. **I rescind my suggestion earlier.**

MikeMcQuaid:
> @MikeMcQuaid Yes

> It’s easy to bypass that (i.e. say “yes, I don’t care, I’ll take the risk”) but it needs to be done deliberately ([our docs explain the process succinctly](https://github.com/Homebrew/homebrew-cask/blob/master/doc/faq/the_app_cant_be_opened.md)).

Given this I'm happy to have this merged.

I'd perhaps go further and say that it's worth the removal of the checksums from anything where quarantines are effective and upstream are in a habit of changing them but: that's up to you.

MikeMcQuaid:
Thanks for all the explanations, all, and hope I've not been too much of a bore here.

claui:
> Thanks for all the explanations, all, and hope I've not been too much of a bore here.

Thanks @MikeMcQuaid for your thoughts. I still stand by what I wrote earlier:

> I appreciate this debate a lot! We can’t have enough eyes on that.

vitorgalvao:
@ran-dall 

> If Apple did revoke the certificate or updated their malware definitions, then that would only have prevented future installs

If I’m recalling the situation properly, that is incorrect. macOS has a “block list” Apple can push updates to silently. The user is never even aware that it happened, but any key can be revoked, precisely due to these situations.

> It's worth considering the time that it would take for Apple to update their definitions

Again, if I’m recalling correctly, in this case they were surprisingly expedite. The news of the malware occurring happened almost at the same time of the news of the key having been revoked.

---

@claui 

> I can’t use `cask-repair`.

I’m always more motivated to (and have a track record of) adding specific features or fixing specific bugs in `cask-repair` if they affect other Homebrew maintainers or other heavy users of the script.

That specific one is just another in a long list of `hub` bugs that I’m tired of debugging. To fix them all, I’ll eventually just start using the API directly. But when I think of that, I also think that remaking it in ruby makes much more sense. And then I realise how monstrous that task is, and that I presently can’t justify the time commitment when the script works (with a lot of duct tape) for most people. For them it’s typically smooth sailing, but I can hear the machinery clunking underneath.

---

@MikeMcQuaid Thank you for the discussion and the merge.

ran-dall:
@vitorgalvao Thanks for the context. 👌
There's **is** a bit of an 'optimist FLOSS mentality' in my comment; a FLOSS community-driven effort vs a corporation like Apple; but there's also some professional experience in there too.😇 But again, I wasn't a part of the Transmission problem 🤷‍♂️

PR #5384, opened 2018-12-06
Update bash completion docs

schneems:
When I try to run the proposed script in my `~/.bashrc` file I get this output:

```
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: have: command not found
-bash: [: =: unary operator expected
# ...
```

It goes on for quite some time and does not give me bash completion.

When I execute `source "$(brew --prefix)/etc/bash_completion"` it works and I get bash completion.

This script also emits a helpful error message letting people know if bash completion was not installed already.

I'll also add that the bash completion script takes a fairly long amount of time. I've seen between one quarter and a half of a second to execute it:

```
$ time source "$(brew --prefix)/etc/bash_completion"

real	0m0.254s
```

Though that's unrelated to this PR, making a note of it.

-----

### DIFF
--- a/docs/Shell-Completion.md
+++ b/docs/Shell-Completion.md
@@ -7,13 +7,14 @@ Homebrew comes with completion definitions for the `brew` command. Some packages
 You must configure your shell to enable the completion support. This is because the Homebrew-managed completions are stored under `HOMEBREW_PREFIX`, which your system shell may not be aware of, and because it is difficult to automatically configure `bash` and `zsh` completions in a robust manner, so the Homebrew installer cannot do it for you.
 
 ## Configuring Completions in `bash`
+
 To make Homebrew's completions available in `bash`, you must source the definitions as part of your shell startup. Add the following to your `~/.bashrc` file:
 
 ```sh
 if type brew 2&>/dev/null; then
-  for completion_file in $(brew --prefix)/etc/bash_completion.d/*; do
-    source "$completion_file"
-  done
+  source "$(brew --prefix)/etc/bash_completion"
+else
+  echo "run: brew install git bash-completion"
 fi
 ```
 


olivierlacan:
TIL: 

```
$ wc -l /usr/local/etc/bash_completion
    1690 /usr/local/etc/bash_completion
```

colindean:
Question on this:

I have for the longest time had this in my bash profile:

```sh
if [ -f $(brew --prefix)/etc/bash_completion ]; then
  . $(brew --prefix)/etc/bash_completion
fi
```

Is there a notable performance or ergonomics difference here, other than checking if `brew` is available first?

olivierlacan:
@colindean It would fail with a confusing error message if `brew` wasn't available: 

```bash
$ if [ -f $(notreal --prefix)/etc/bash_completion ]; then
> echo "yay"
> else
> echo "nay"
> fi
-bash: notreal: command not found
nay
```

Whereas this new suggestion would echo: 
> run: brew install git bash-completion

That said if `brew` wasn't available then I'm not sure suggesting a `brew install` command makes sense either since that would suggest the `brew` command isn't available at all.

schneems:
@MikeMcQuaid thanks for the merge and for all your hard work! Homebrew (and you) rock!!

@colindean On the performance note `$(brew --prefix)` is not free. I actually changed it to a hardcoded path on my local machine.

```
$ time echo $(brew --prefix)
/usr/local

real	0m0.035s
user	0m0.014s
sys	0m0.019s
```

If you call `$(brew --prefix)` 8 times in your `.bashrc` that's a quarter of a second.

The `type brew` check that is in the current doc is much faster than checking for the file since it does not execute a `brew --prefix`.

The perf of `brew --prefix` MUCH worse if you pass it an argument:

```
$ time echo $(brew --prefix qt@5.5)
/usr/local/opt/qt@5.5

real	0m1.285s
user	0m0.614s
sys	0m0.531s
```

🙀🙀🙀

Also I see your a member of the org too, so thanks! Would it make sense to split any of this perf info into an issue?

MikeMcQuaid:
> For the things that use `brew --prefix` in my bashrc it was easy to replace with the literals they generate, however, I wasn't able to replace the bash completion script.

I also do something similar: https://github.com/MikeMcQuaid/dotfiles/blob/74c1c0c16b4df25a1e7c88b41cffb1ad3de754b6/shrc.sh#L97

PR #5388, opened 2018-12-08
_brew | allow 'brew commands' to be cached by zsh

bosr:

-----

Hi,

Here is an update to `_brew` where the most expensive command (`__brew_all_commands`) has its results cached. This results in faster (immediate?) completion. This should help people with slow machine (my mac at work is hugely slowed down by the security measures).

I tried my best and used the standard mechanism; the code follows along the many available examples. 

The user will have its results cached if his/her shell is configured to do so: meaning that the standard zstyle (`use-cache`) has to be set to `'yes'` and that the cache will be saved by default in `~/.zcompcache/brew_all_commands` (also configurable using the zstyle `cache-path`). The delay for cache invalidation is set to 24h [completions/zsh/_brew#135](https://github.com/Homebrew/brew/pull/5388/files#r240019986). I expect no kitten to be harmed with this PR.

Regarding tests, I only conducted manual checks:
- with no existing cache file at `~/.zcompcache/brew_all_commands`: the cache gets created
- with an invalid (>24h) cache file: the cache gets recreated
- with a valid cache file (<=24h): the cache is simply used and thus will eventually get older than 24h.

I hope you'll review it and it eventually gets accepted.
Have a great day.

### DIFF
--- a/completions/zsh/_brew
+++ b/completions/zsh/_brew
@@ -129,9 +129,26 @@ __brew_common_commands() {
   _describe -t common-commands 'common commands' commands
 }
 
+# completions are cached for 24 hour
+__brew_commands_caching_policy() {
+  local -a oldp
+  oldp=( "$1"(Nmh+24) )
+  (( $#oldp ))
+}
+
 __brew_all_commands() {
+  local cache_policy
+  zstyle -s ":completion:${curcontext}:" cache-policy cache_policy
+  if [[ -z "$cache_policy" ]]; then
+    zstyle ":completion:${curcontext}:" cache-policy __brew_commands_caching_policy
+  fi
   local -a commands
-  commands=($(_call_program brew brew commands --quiet --include-aliases))
+  local comp_cachename=brew_all_commands
+
+  if  _cache_invalid $comp_cachename  || ! _retrieve_cache $comp_cachename; then
+    commands=($(_call_program brew brew commands --quiet --include-aliases))
+    _store_cache $comp_cachename commands
+  fi
   _describe -t all-commands 'all commands' commands
 }
 


completions/zsh/_brew
```diff
@@ -129,9 +129,26 @@ __brew_common_commands() {
   _describe -t common-commands 'common commands' commands
 }
 
+# completions are cached for 24 hour
+__brew_commands_caching_policy() {
+  local -a oldp
+  oldp=( "$1"(Nmh+24) )
```
bosr: cache invalidation deadline = 24h

MikeMcQuaid:
Nice work, this seems pretty neat. Would it be worth similarly caching the names of formulae or is that pretty fast?

CC @apjanke who often has ZSH opinions.

> I tried my best and used the standard mechanism; the code follows along the many available examples.

Could you post some links here for our learning?

Thanks and great work!

bosr:
Hi Mike, thanks for the feedback :) Here are some references and impl details. Cheers!

## Some references
Here is a good example of zsh completion involving the caching mechanism, I used it as my master example: [zsh-completions/src/_pygmentize](https://github.com/zsh-users/zsh-completions/blob/d3a273cedf788181db9a05726dcd4d930da9e248/src/_pygmentize), along with other completions from [zchee/zsh-completions](https://github.com/zchee/zsh-completions).

For going further, the Zsh reference manual for completion is [chapter 20](http://zsh.sourceforge.net/Doc/Release/Completion-System.html). In particular the functions used here are all described: `_retrieve_cache`, `_cache_invalid`, `_store_cache`, and the zstyles `use-cache`, `cache-path`, `cache-policy`. It is famously difficult to digest, I think it's because of the many cross-dependencies between the concepts. I spent some time recently and finally start to get more familiar with it... until I forget it all again!

## Implementation detail
The main difference with my code is in the `if` statements, for instance this one at [line 86](https://github.com/zsh-users/zsh-completions/blob/d3a273cedf788181db9a05726dcd4d930da9e248/src/_pygmentize#L86):

```zsh
_get_formatters() {
    # ...
    if ( [[ ${+_pygmentize_formatter} -eq 0 ]] || _cache_invalid pygmentize_formatter ) \
          && ! _retrieve_cache pygmentize_formatter; then
        # recreate cache
    fi
    # ...
}
```

When `_get_formatters` is invoked, the global variable (`_pygmentize_formatter`) may be not null, so their tests make sense and the "semantics" are correct.

In my code for the function `__brew_all_commands`, because I did not want to change the way it is currently factored, the array `commands` is local to the function and thus gets created on each invocation. So I had to drop the first test `${+...} -eq 0` in order to let a chance to the `_cache_invalid` mechanism a chance to execute, and maintain the same "semantics".

bosr:
Regarding formulae name caching, good catch, let's try it!

apjanke:
This sounds like a good idea, and as far as I can tell, looks like idiomatic ZSH code.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @bosr!

PR #5393, opened 2018-12-10
shims/super/cc: Add HOMEBREW_CACHE to white list

sjackman:
Compiling rust projects requires -I$HOMEBREW_CACHE/cargo_cache/...

-----

`brew install -s exa` fails on Linux because Superenv removes `-I/home/linuxbrew/.cache/Homebrew/cargo_cache/registry/src/github.com-1ecc6299db9ec823/libgit2-sys-0.6.14/libgit2/src`

See PR https://github.com/Linuxbrew/homebrew-core/pull/10162

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -26,7 +26,7 @@ def linux?
 end
 
 class Cmd
-  attr_reader :config, :prefix, :cellar, :opt, :tmpdir, :sysroot, :deps
+  attr_reader :config, :prefix, :cellar, :opt, :cachedir, :tmpdir, :sysroot, :deps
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Why is this not needed on macOS out of interest? Any ideas?

MikeMcQuaid:
Works for me, thanks for explaining @sjackman! I was thinking we could maybe even in future restrict this to the macOS SDK path.

sjackman:
Thanks for merging, Mike.

> Works for me, thanks for explaining @sjackman! I was thinking we could maybe even in future restrict this to the macOS SDK path.

I think that's a great idea. We'd need to add the other allowable 3rd party tools that we make use of, like Java, XQuartz, et c. It should be pretty easy to find them all by looking through `requirements/`.

PR #5406, opened 2018-12-14
Clarify the meaning of open source for core formula

jasontedor:
This commit clarifies the meaning of "open-source" when explaining that only open source formulae are acceptable in core. In particular, the open source license must be an OSI-approved license.


jasontedor:
Heya @MikeMcQuaid, following up on our discussion, I am suggesting this clarification to the docs. Would you be able to review and let me know your thoughts?

PR #5413, opened 2018-12-17
mac/keg_relocate: Rewrite rpaths in install names

woodruffw:

-----

This removes `@rpath` as a special case for not performing relocations, leaving only `@loader_path` and `@executable_path`. I still need to do some local testing (and write some unit tests) for this change, so WIP until then.

Note to self: Remove the remaining `LC_RPATH`s from the files as well, since they'll be vestigial once we rewrite each install name.

cc @tschoonj @sjackman @MikeMcQuaid

### DIFF
--- a/Library/Homebrew/extend/os/mac/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/mac/keg_relocate.rb
@@ -79,10 +79,10 @@ class Keg
     suffix = bad_name.sub(/^@rpath/, "")
 
     # short circuit: we expect lib to be usually correct, so we try it first
-    return (lib + suffix) if (lib + suffix).exist?
+    return (lib / suffix) if (lib / suffix).exist?
 
     file.rpaths.each do |rpath|
-      return (rpath + suffix) if (rpath + suffix).exist?
+      return (rpath / suffix) if (rpath / suffix).exist?
     end
 
     opoo "Could not expand an RPATH in #{file}"
@@ -101,7 +101,7 @@ class Keg
       "@loader_path/#{bad_name}"
     elsif file.mach_o_executable? && (lib + bad_name).exist?
       "#{lib}/#{bad_name}"
-    elsif bad_name.start_with? "@rpath"
+    elsif bad_name.start_with? "@rpath" && ENV["HOMEBREW_RELOCATE_RPATHS"]
       expand_rpath bad_name
     elsif (abs_name = find_dylib(bad_name)) && abs_name.exist?
       abs_name.to_s
--- a/Library/Homebrew/extend/os/mac/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/mac/keg_relocate.rb
@@ -95,14 +95,6 @@ class Keg
     bad_name
   end
 
-  def expand_loader_path(file, bad_name)
-    suffix = bad_name.sub(/^@loader_path/, "")
-
-    # Note: Weak dylibs are allowed to not exist on disk, so
-    # we don't check for existence & complain here.
-    file.parent/suffix
-  end
-
   # If file is a dylib or bundle itself, look for the dylib named by
   # bad_name relative to the lib directory, so that we can skip the more
   # expensive recursive search if possible.
@@ -112,13 +104,11 @@ class Keg
     elsif bad_name.start_with? CELLAR_PLACEHOLDER
       bad_name.sub(CELLAR_PLACEHOLDER, HOMEBREW_CELLAR)
     elsif (file.dylib? || file.mach_o_bundle?) && (file.parent + bad_name).exist?
-      file.parent/bad_name
+      "@loader_path/#{bad_name}"
     elsif file.mach_o_executable? && (lib + bad_name).exist?
       "#{lib}/#{bad_name}"
     elsif bad_name.start_with?("@rpath") && ENV["HOMEBREW_RELOCATE_METAVARS"]
       expand_rpath file, bad_name
-    elsif bad_name.start_with?("@loader_path") && ENV["HOMEBREW_RELOCATE_METAVARS"]
-      expand_loader_path file, bad_name
     elsif (abs_name = find_dylib(bad_name)) && abs_name.exist?
       abs_name.to_s
     else
@@ -129,7 +119,7 @@ class Keg
 
   def each_install_name_for(file, &block)
     dylibs = file.dynamically_linked_libraries
-    dylibs.reject! { |fn| fn =~ /^@executable_path/ }
+    dylibs.reject! { |fn| fn =~ /^@(loader|executable)_path/ }
     dylibs.each(&block)
   end
 
--- a/Library/Homebrew/extend/os/mac/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/mac/keg_relocate.rb
@@ -95,6 +95,14 @@ class Keg
     bad_name
   end
 
+  def expand_loader_path(file, bad_name)
+    suffix = bad_name.sub(/^@loader_path/, "")
+
+    # Note: Weak dylibs are allowed to not exist on disk, so
+    # we don't check for existence & complain here.
+    file.parent/suffix
+  end
+
   # If file is a dylib or bundle itself, look for the dylib named by
   # bad_name relative to the lib directory, so that we can skip the more
   # expensive recursive search if possible.
@@ -104,11 +112,13 @@ class Keg
     elsif bad_name.start_with? CELLAR_PLACEHOLDER
       bad_name.sub(CELLAR_PLACEHOLDER, HOMEBREW_CELLAR)
     elsif (file.dylib? || file.mach_o_bundle?) && (file.parent + bad_name).exist?
-      "@loader_path/#{bad_name}"
+      file.parent/bad_name
     elsif file.mach_o_executable? && (lib + bad_name).exist?
       "#{lib}/#{bad_name}"
-    elsif bad_name.start_with?("@rpath") && ENV["HOMEBREW_RELOCATE_RPATHS"]
-      expand_rpath bad_name
+    elsif bad_name.start_with?("@rpath") && ENV["HOMEBREW_RELOCATE_METAVARS"]
+      expand_rpath file, bad_name
+    elsif bad_name.start_with?("@loader_path") && ENV["HOMEBREW_RELOCATE_METAVARS"]
+      expand_loader_path file, bad_name
     elsif (abs_name = find_dylib(bad_name)) && abs_name.exist?
       abs_name.to_s
     else
@@ -119,7 +129,7 @@ class Keg
 
   def each_install_name_for(file, &block)
     dylibs = file.dynamically_linked_libraries
-    dylibs.reject! { |fn| fn =~ /^@(loader_|executable_)path/ }
+    dylibs.reject! { |fn| fn =~ /^@executable_path/ }
     dylibs.each(&block)
   end
 
--- a/Library/Homebrew/extend/os/mac/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/mac/keg_relocate.rb
@@ -78,14 +78,11 @@ class Keg
   def expand_rpath(file, bad_name)
     suffix = bad_name.sub(/^@rpath/, "")
 
-    # short circuit: we expect lib to be usually correct, so we try it first
-    return (lib / suffix) if (lib / suffix).exist?
-
     file.rpaths.each do |rpath|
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Nice, makes sense.

> I still need to do some local testing (and write some unit tests) for this change, so WIP until then.

A suggestion: make this enabled only when an environment variable is set. That way you can do easier local testing and we can roll it out to `brew test-bot` before it goes out to all other users.

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -87,6 +87,8 @@ def fixed_name(file, bad_name)
       "@loader_path/#{bad_name}"
     elsif file.mach_o_executable? && (lib + bad_name).exist?
       "#{lib}/#{bad_name}"
+    elsif bad_name.start_with? "@rpath"
+      bad_name.sub("@rpath", lib)
```
sjackman: The `LC_RPATH` may be set to `#{lib}`, but it could be set to something else. You'll need to emulate the logic of the dynamic linker and search each directory specified by `LC_RPATH` to locate the given library, to determine the correct absolute path.

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -75,6 +75,20 @@ def fix_dynamic_linkage
     generic_fix_dynamic_linkage
   end
 
+  def expand_rpath(file, bad_name)
+    suffix = bad_name.sub(/^@rpath/, "")
+
+    # short circuit: we expect lib to be usually correct, so we try it first
+    return (lib + suffix) if (lib + suffix).exist?
```
sjackman: Prefer `lib/suffix` over `lib + suffix` for style.

sjackman: I'm inclined though to remove this short circuit. I don't think it'll make a significant performance difference, and is otherwise unnecessary code.

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -75,6 +75,20 @@ def fix_dynamic_linkage
...
+
+    file.rpaths.each do |rpath|
+      return (rpath + suffix) if (rpath + suffix).exist?
```
sjackman:
```suggestion
      return rpath/suffix if (rpath/suffix).exist?
```

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -75,6 +75,20 @@ def fix_dynamic_linkage
...
+    end
+
+    opoo "Could not expand an RPATH in #{file}"
```
sjackman:
```suggestion
    opoo "Could not find library #{bad_name} for #{file}"
```

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -78,14 +78,11 @@ def fix_dynamic_linkage
   def expand_rpath(file, bad_name)
     suffix = bad_name.sub(/^@rpath/, "")
 
-    # short circuit: we expect lib to be usually correct, so we try it first
-    return (lib / suffix) if (lib / suffix).exist?
-
     file.rpaths.each do |rpath|
-      return (rpath / suffix) if (rpath / suffix).exist?
+      return (rpath/suffix) if (rpath/suffix).exist?
```
sjackman: I prefer no parentheses around `return (rpath/suffix)` myself. They're not necessary.

woodruffw:
> Can you please remove the `RPATH` from the MachO file, so that the bottle can be `cellar :any`?

Yeah, I was thinking about how best to do that: we could make a second pass over `mach_o_files` after all install name rewrites are finished, or just keep a temporary list of the changed files and remove the RPATHs from those (at the risk of leaving dangling RPATHs in other files). Thoughts?

Library/Homebrew/os/mac/mach.rb
```diff
@@ -60,6 +60,14 @@ def dylib_id
     macho.dylib_id
   end
 
+  def rpaths
+    macho.rpaths
+  end
+
+  def delete_rpath(rpath)
+    macho.delete_rpath(rpath)
+  end
```
sjackman: I believe you can use `delegate` like so…
```ruby
delegate [:delete_rpath, :rpaths] => :macho
```
We use it elsewhere in the Homebrew code base.

MikeMcQuaid:
@woodruffw Any news on this? Would be 🆒 if possible to have it in the next few weeks as a nice big 2.0.0 feature 🎉

woodruffw:
> Any news on this?

Yup, I should have some free time to work on it soon (an evening this week or this weekend). Sorry for the delay, things have been a little hectic 😅

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -84,9 +112,13 @@ def fixed_name(file, bad_name)
...
       bad_name.sub(CELLAR_PLACEHOLDER, HOMEBREW_CELLAR)
     elsif (file.dylib? || file.mach_o_bundle?) && (file.parent + bad_name).exist?
-      "@loader_path/#{bad_name}"
+      file.parent/bad_name
     elsif file.mach_o_executable? && (lib + bad_name).exist?
       "#{lib}/#{bad_name}"
+    elsif bad_name.start_with?("@rpath") && ENV["HOMEBREW_RELOCATE_METAVARS"]
+      expand_rpath file, bad_name
+    elsif bad_name.start_with?("@loader_path") && ENV["HOMEBREW_RELOCATE_METAVARS"]
+      expand_loader_path file, bad_name
```
sjackman: Is there an advantage to rewriting `@loader_path`, or is it simpler to leave it as is?

woodruffw: Oh, that was as a follow up to https://github.com/Homebrew/homebrew-core/issues/36685 -- I don't think there's an advantage per se (we're just doing AOT what `dyld` would do at load-time). Perhaps I misunderstood what you were asking for in that issue?

sjackman: That's exactly what I said, but I spoke too quickly. After re-reading https://github.com/Homebrew/homebrew-core/issues/36685, it's not clear to me what exactly the root-problem is. Sorry for sending you astray.

This change would make some kegs that were previously `cellar :any_skip_relocatable` because they use `@loader_path` to find their libraries would require relocation and become `cellar :any` after this change. For this reason, I don't think this change is desirable.

Library/Homebrew/extend/os/mac/keg_relocate.rb
```diff
@@ -84,9 +112,13 @@ def fixed_name(file, bad_name)
     elsif bad_name.start_with? CELLAR_PLACEHOLDER
       bad_name.sub(CELLAR_PLACEHOLDER, HOMEBREW_CELLAR)
     elsif (file.dylib? || file.mach_o_bundle?) && (file.parent + bad_name).exist?
-      "@loader_path/#{bad_name}"
+      file.parent/bad_name
```
sjackman: Ditto this change.

Library/Homebrew/extend/os/mac/keg_relocate.rb
sjackman: What does AOT stand for?

woodruffw: No problem! And sorry -- AOT is "ahead of time." But you're right -- it just adds a relocation case where none is necessary.

sjackman:
This PR gave me the idea that if we were to use paths relative to `@loader_path` everywhere rather than `@HOMEBREW_PREFIX@` and `@HOMEBREW_CELLAR@` then MachO executables would not need to be relocated (plain text files would still need to be relocated). The Linux dynamic loader has this same feature, but calls it `$ORIGIN` rather than `@loader_path`. This assumes that `$HOMEBREW_CELLAR` is `$HOMEBREW_PREFIX/Cellar`. If it's not, they would still need to be relocated, but it's a nice optimization for the typical case. Thoughts?

woodruffw:
> Thoughts?

I really like that idea! I wonder why we didn't do it originally.  I want to push this PR forwards (I feel terrible for letting it stagnate for so long), but replacing our current placeholders with `@loader_path` is what I'll look into next.

sjackman:
I definitely didn't mean to suggest for this PR. It's great. Just an idea that popped into my head for a future PR.
For a typical binary in `$HOMEBREW_PREFIX/Cellar/foo/1.2.3/bin`
`@HOMEBREW_CELLAR@` would be replaced with `@loader_path/../../..`
`@HOMEBREW_PREFIX@` would be replaced with `@loader_path/../../../..`
and the rare use of
`@HOMEBREW_REPOSITORY@` would be replaced with `@loader_path/../../../../Homebrew`
Because each is a prefix of the next, their replacement needs to be done in the correct order.

woodruffw:
Cool! I'll merge this tonight (and open a PR to enable it on test-bot), unless any other feedback comes up.

sjackman:
You likely already know this… [Pathname.relative_path_from](https://ruby-doc.org/stdlib-2.3.7/libdoc/pathname/rdoc/Pathname.html#method-i-relative_path_from) can be used to construct the relative path from the MachO file to its needed dependency.

scpeters:
I'm not sure if it was this or https://github.com/Homebrew/homebrew-test-bot/pull/242 but there are now lots of linkage errors reported in the ongoing boost build:

* https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/39335/version=mojave/

for reference, there were no linkage errors reported in the previous build:

* https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/39208/

woodruffw:
Yeah, I'll go ahead and revert this for now. I'll root cause tomorrow to see if it is indeed a weak dylib issue (which I suspect).

AuroraWright:
This seems to break building `wine`, reverting brew to 2.0.4 fixes it:
```
dyld: Library not loaded: /usr/local/Cellar/wine/4.4/lib/libwine.1.dylib
  Referenced from: /usr/local/Cellar/wine/4.4/lib/../bin/wineserver
  Reason: no suitable image found.  Did find:
	/usr/local/Cellar/wine/4.4/lib/libwine.1.dylib: mach-o, but wrong architecture
	/usr/local/Cellar/wine/4.4/lib/libwine.1.dylib: stat() failed with errno=1
	/usr/local/lib/libwine.1.dylib: mach-o, but wrong architecture
	/usr/local/Cellar/wine/4.4/lib/libwine.1.0.dylib: mach-o, but wrong architecture
	/usr/local/Cellar/wine/4.4/lib/libwine.1.0.dylib: mach-o, but wrong architecture
```
```
dyld: Library not loaded: /usr/local/Cellar/wine/4.4/lib/libwine.1.dylib
  Referenced from: /usr/local/bin/wine
  Reason: no suitable image found.  Did find:
	/usr/local/Cellar/wine/4.4/lib/libwine.1.dylib: mach-o, but wrong architecture
	/usr/local/Cellar/wine/4.4/lib/libwine.1.dylib: stat() failed with errno=1
	/usr/local/lib/libwine.1.dylib: mach-o, but wrong architecture
	/usr/local/Cellar/wine/4.4/lib/libwine.1.0.dylib: mach-o, but wrong architecture
	/usr/local/Cellar/wine/4.4/lib/libwine.1.0.dylib: mach-o, but wrong architecture
```

woodruffw:
> It seems the issue is that 64-bit executables wrongly try to load dylibs from the "lib" folder, rather than the "lib64" folder.

Interesting, thanks for reporting. I'll make a PR to revision the wine formula and fix the linkage.

sjackman:
Since we don't support 32-bit Intel, one possible fix is to remove `lib` and rename `lib64` to `lib` (assuming `lib` is 32-bit libraries and `lib64` is 64-bit libraries).

AuroraWright:
Unfortunately wine doesn't work without the 32 bit side of things as of yet (as it needs the 32 bit executables to load 32 bit Windows executables), which is probably why 32 bit building wasn't removed from the formula

sjackman:
Thanks for that info, @AuroraWright !

fwal:
This broke swiftgen too 😢 https://github.com/Homebrew/homebrew-core/issues/38044

sjackman:
This PR should hopefully prevent more breakage: https://github.com/Homebrew/brew/pull/5894

PR #5421, opened 2018-12-18
Allow incomplete downloads to be resumed even when server rejects HEAD requests

smammy:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb). _In order to write a test for this I'd have to refactor `curl_download` or (I guess) mock `curl_output` and `curl` and I don't know how to do that. Old behavior wasn't unit-tested either so I don't think it's a huge deal._

-----

Use GET rather than HEAD when checking for range support in curl_download.
   * Some HTTP servers apparently support ranges but don't support HEAD.
   * This is a more realistic check anyway since the actual download request
     will use GET (not HEAD).
   * This fixes Homebrew/brew#5420.

### DIFF
--- a/Library/Homebrew/utils/curl.rb
+++ b/Library/Homebrew/utils/curl.rb
@@ -52,7 +52,7 @@ def curl_download(*args, to: nil, **options)
   destination.dirname.mkpath
 
   continue_at = if destination.exist? &&
-                   curl_output("--location", "--head", "--range", "0-1",
+                   curl_output("--location", "--range", "0-1",
                                "--write-out", "%{http_code}",
                                "--output", "/dev/null", *args, **options).stdout.to_i == 206 # Partial Content
     "-"


MikeMcQuaid:
CC @reitermarkus for thoughts.

smammy:
Although, honestly I don't understand what was wrong with the behavior before 0721271 where curl_download always tried `--continue-at -` and recovered from the error if the server didn't support partial content. (Most servers seem to these days, so the old pre-0721271 behavior would save us an HTTP request in the common case.) #4847 doesn't justify adding the upfront check, but I assume there was some reason for it.

reitermarkus:
I think the reason for the upfront check was that some servers which don't support ranges wouldn't actually fail with the appropriate error code.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @smammy!

PR #5426, opened 2018-12-20
Fix missing shared symlinks for versioned Postgres

pirj:
When creating a formula that depends on a versioned Postgres, e.g. `postgres@10`, with an intention to install to shared Postgres directory, the symlinks were not created.

`postgres` in `SHARE_PATHS` part of the `case` statement was only matching when passed `postgres`, but not `postgres@10`.

Example formula:

```ruby
    class FooExtension < Formula
      desc "Foo extension for PostgreSQL"
      homepage "https://www.example.com/"
      url "https://github.com/zombodb/zombodb/archive/v10-1.0.0.tar.gz"
      sha256 "2023180933a9513045f9a2e7b67868a8118e8b5c80038e155bba907195596307"
      depends_on "postgresql@10"

      def install
        mkdir "stage"
        system "mkdir -p stage/FAKE/share/postgresql@10/extension/*"
        system "echo fake > stage/FAKE/share/postgresql@10/extension/fake.control"
        (share/"postgresql@10/extension").install Dir["stage/**/share/postgresql@10/extension/*"]
      end
    end
```

When installed the formula was not creating `fake.control` in `/usr/local/share/postgresql@10/extension`.

With this fix it does.

I intend to submit a [ZomboDB](https://github.com/zombodb/zombodb) PostgreSQL formula but at the moment ZomboDB only supports PostgreSQL 10.

I feel that I need help to write a good test for it. I've tried to add the following to `keg_spec.rb` but got lost in the semantics, and the spec passes on `master`:
```
      specify "shared symlinks are linked with a versioned formula" do
        link = HOMEBREW_PREFIX/"share"/"postgres@10"/"extension"/"helloworld"
        link.parent.mkpath
        link.make_symlink dst
        keg.link

        expect(link.resolved_path).to be_a_symlink
        expect(link.lstat).to be_a_symlink
      end
```

-----

### DIFF
--- a/Library/Homebrew/keg.rb
+++ b/Library/Homebrew/keg.rb
@@ -108,7 +108,7 @@ class Keg
     man/cat5 man/cat6 man/cat7 man/cat8
     applications gnome gnome/help icons
     mime-info pixmaps sounds postgresql
-  ].freeze
+  ].map { |dir| Regexp.new(dir) }.freeze
 
   # Given an array of kegs, this method will try to find some other kegs
   # that depend on them. If it does, it returns:


MikeMcQuaid:
Sorry but this is intentional behaviour. We don't want to symlink a `/usr/local/share/postgresql@10` directory. Thanks for the great write-up though ❤️

pirj:
Thanks for the quick response, Mike.
Any hints on how to build a Postgres extension formula that targets Postgres 10 (10 and 11 in the near future)?

MikeMcQuaid:
@pirj because it's keg-only i.e. not linked by default into `/usr/local`. This is intentional so multiple versions can be installed side-by-side.

Issue #5427, opened 2018-12-20
"undefined method `shelljoin' for nil:NilClass" error when patches failed


apjanke:
- [X] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [ ] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap. *I can only reproduce it with the one formula.*
- [X] ran `brew update` and can still reproduce the problem?
- [X] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [X] ran `brew config` and `brew doctor` and included their output with your issue?

# What you were trying to do (and why)

Building `octave-head`, my custom variant of the `octave` formula.

(I've also added an `octave-head-test` formula which reproduces the same issue, but is much quicker to build so it'll be easier for people to test this.)

# What happened (include command output)

Some patches failed, and instead of getting a nice error message, `brew` threw an exception and printed a stack trace.

I don't know if this is _because_ of the patch failures, but that's when it appeared to happen.

<details>
  <summary>Command output</summary>
  <pre>
$ brew install -s octave-head
==> Installing octave-head from octave-app/octave-app-bases
==> Cloning https://hg.savannah.gnu.org/hgweb/octave
Updating /Users/janke/Library/Caches/Homebrew/octave-head--hg
pulling from https://hg.savannah.gnu.org/hgweb/octave
searching for changes
no changes found
==> Checking out branch default
0 files updated, 0 files merged, 0 files removed, 0 files unresolved
(leaving bookmark @)
==> Downloading https://savannah.gnu.org/support/download.php?file_id=44041
Already downloaded: /Users/janke/Library/Caches/Homebrew/downloads/a1e44d4b6fcd4c0ca6d151fc8927da5900a8a26214f9da6cc7fdcf1798f7be78--high_res_update5.patch
==> Downloading https://gist.githubusercontent.com/apjanke/da92f70978aa8db01f484c782aed89a9/raw/37e1
[...]

PR #5440, opened 2018-12-23
build: Propagate entire ErrorDuringExecution

woodruffw:

-----

Fixes https://github.com/Homebrew/brew/issues/5427.

The underlying issue here is that we propagate child errors twice -- once in `Utils.safe_fork`, and again within the specific scripts that get executed within the forked child (e.g., `build.rb`). However, the propagation in `build.rb` wasn't special-casing `ErrorDuringExecution`, so the first exception received over the error pipe was missing the expected `cmd` field.

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -300,7 +300,7 @@ end
 def safe_system(cmd, *args, **options)
   return if Homebrew.system(cmd, *args, **options)
 
-  raise(ErrorDuringExecution.new([cmd, *args], status: $CHILD_STATUS))
+  raise ErrorDuringExecution.new([cmd, *args], status: $CHILD_STATUS)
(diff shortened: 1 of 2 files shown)


Library/Homebrew/utils.rb
```diff
@@ -300,7 +300,7 @@ def with_custom_locale(locale)
 def safe_system(cmd, *args, **options)
   return if Homebrew.system(cmd, *args, **options)
 
-  raise(ErrorDuringExecution.new([cmd, *args], status: $CHILD_STATUS))
+  raise ErrorDuringExecution.new([cmd, *args], status: $CHILD_STATUS)
```
MikeMcQuaid: Merging anyway but in general I'd avoid refactoring like this in the same PR.

Issue #4527, opened 2018-07-21
"Cannot read termcap database" for TERM not supported by default ncurses version


mqudsi:
- [x] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew update` and can still reproduce the problem?
- [ ] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [ ] ran `brew config` and `brew doctor` and included their output with your issue?

When running under tmux with `TERM=tmux-256color`, brew is unable to determine the capabilities of the terminal and emits warnings:

```
mqudsi@QBook ~/P/messagebridge> brew install libiconv                                                       
Cannot read termcap database;                                                                                
using dumb terminal settings.
```

I'm not sure what termcap database brew is trying to use, because it seems like everything is OK:
```
mqudsi@MacBook-Pro ~> echo $TERM
tmux-256color
mqudsi@MacBook-Pro ~> infocmp
#       Reconstructed via infocmp from file: /usr/local/Cellar/ncurses/6.1/share/terminfo/74/tmux-256color
tmux-256color|tmux with 256 colors,
        am, hs, km, mir, msgr, xenl,
        colors#0x100, cols#80, it#8, lines#24, pairs#0x10000,
        acsc=++\,\,--..00``aaffgghhiijjkkllmmnnooppqqrrssttuuvvwwxxyyzz{{||}}~~,
        bel=^G, blink=\E[5m, bold=\E[1m, cbt=\E[Z, civis=\E[?25l,
        clear=\E[H\E[J, cnorm=\E[34h\E[?25h, cr=\r,
        csr=\E[%i%p1%d;%p2%dr, cub=\E[%p1%dD, cub1=^H,
        cud=\E[%p1%dB, cud1=\n, cuf=\E[%p1%dC, cuf1=\E[C,
        cup=\E[%i%p1%d;%p2%dH, cuu=\E[%p1%dA, cuu1=\EM,
[...]

PR #5445, opened 2018-12-26
Whitelist TERMINFO environment variable

jacwah:

-----

The `TERMINFO` environment variable is used by ncurses et al to find files describing current terminal's capabilities (see e.g. `man ncurses`). When the variable is filtered out, code using `ncurses` down the line gets confused if the terminal is not included in the default termcap database. An example of this is the "Cannot read termcap database; using dumb terminal settings" message described in #4527. This is written by libedit, loaded from irb, loaded from debrew. With this patch the message is not printed any more.

An alternative solution is to place termcap files under `~/.terminfo` instead. Some terminals do not do this by default, for instance Kitty.

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -72,7 +72,7 @@ then
 
   FILTERED_ENV=()
   # Filter all but the specific variables.
-  for VAR in HOME SHELL PATH TERM COLUMNS LOGNAME USER CI TRAVIS SSH_AUTH_SOCK SUDO_ASKPASS \
+  for VAR in HOME SHELL PATH TERM TERMINFO COLUMNS LOGNAME USER CI TRAVIS SSH_AUTH_SOCK SUDO_ASKPASS \
              http_proxy https_proxy ftp_proxy no_proxy all_proxy HTTPS_PROXY FTP_PROXY ALL_PROXY \
              "${!HOMEBREW_@}" "${!TRAVIS_@}"
   do


MikeMcQuaid:
Can you provide the exact reproduction case for this?

jacwah:
To reproduce:
1. Use a terminal that does not have its termcap database distributed with ncurses, but has a termcap database located at `$TERMINFO`. In my case, I am using [Kitty](https://sw.kovidgoyal.net/kitty/) with `TERMINFO=/Applications/kitty.app/Contents/Frameworks/kitty/terminfo`. This is the default setup for this particular terminal emulator.
2. Run a Homebrew install command, for instance `brew install` (without arguments).
3. The message "Cannot read termcap database; using dumb terminal settings" is written to stderr.

If readline was actually being used, this would probably lead to weird behavior like backspace not working etc, as ncurses would not know how to interact with the terminal.

When `TERMINFO` is whitelisted, this message does not appear.

I traced this message to the initialization of libedit ([terminal.c](http://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/lib/libedit/terminal.c)). As you can see, it is written when ncurses does not find a termcap for the current `$TERM`. On my machine, the ruby `readline` module uses libedit under the hood:

```
> Readline::VERSION
=> "EditLine wrapper"
```

As mentioned in the OP, readline is transitively required by debrew.rb.

MikeMcQuaid:
Which if any of those variables are customised from the application default values?

jacwah:
No variables are customized. As I said, I am using the default setup for the application.

MikeMcQuaid:
Cool, seems reasonable. Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @jacwah!

smancill:
@jacwah `TERMINFO_DIRS` should also be whitelisted. Otherwise the error mentioned in #4527  doesn't go away.

maxim-belkin:
@smancill, if you can provide a reproducible scenario when this is necessary -- submit a PR! 😉

PR #5446, opened 2018-12-27
brew.sh: set correct log path [Linux]

jonchang:

-----

On Linux using the Homebrew/brew remote, log files are stored in `~/Library/Logs/Homebrew`. This is the macOS path and is incorrect for Linux. This pull request moves the log path to `~/.cache/Homebrew/Logs`, respecting XDG environment variables if specified.

Tested on macOS 10.14.2 and on a Linux HPC cluster with a user install. `brew install hello` confirms that logs are stored in their appropriate places.

Potential issues for followup:

* Logs and Cache are in completely separate folders on macOS; on Linux the log path is a subdirectory of the cache path.
* Should all OS-specific environment variables be defined in `brew.sh` vs `config.rb`?
* `brew config` will always print the cache path on Linux, since the macOS default path is hard coded into `system_config.rb` 

c.f. https://github.com/Homebrew/brew/issues/4758

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -113,6 +113,7 @@ then
   fi
 
   HOMEBREW_CACHE="${HOMEBREW_CACHE:-${HOME}/Library/Caches/Homebrew}"
+  HOMEBREW_LOGS="${HOMEBREW_LOGS:-${HOME}/Library/Logs/Homebrew}"
   HOMEBREW_SYSTEM_TEMP="/private/tmp"
 else
   HOMEBREW_PROCESSOR="$(uname -m)"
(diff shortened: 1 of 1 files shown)


Library/Homebrew/config.rb
```diff
@@ -36,7 +36,7 @@
 HOMEBREW_CACHE_FORMULA = HOMEBREW_CACHE/"Formula"
 
 # Where build, postinstall, and test logs of formulae are written to
-HOMEBREW_LOGS = Pathname.new(ENV["HOMEBREW_LOGS"] || "~/Library/Logs/Homebrew/").expand_path
+HOMEBREW_LOGS = Pathname.new(ENV["HOMEBREW_LOGS"])
```
MikeMcQuaid: This will need to keep the `||` or it'll break on `brew update`. It doesn't need to be platform specific here, though.

jonchang: I couldn't reproduce an update error using `brew update-test --commit=HEAD^`. I'm also not sure I understand what you're asking for; are you suggesting that I leave this line unchanged instead?

MikeMcQuaid: Yes leave it unchanged. It may not be reproducible by update test. When an autoupdate occurs the Bash environment variables won’t have been updated but the Ruby ones will. This may justify a separate issue (and a PR would be great) to resource all env var initalisation on a successful auto update.

jonchang:
Thanks for the reviews @MikeMcQuaid @sjackman.

PR #5448, opened 2018-12-27
diagnostic: share prefix check

jonchang:

-----

requested by https://github.com/Homebrew/brew/pull/5447#discussion_r244200557

Shares the prefix check logic between macOS and Linux.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -35,7 +35,11 @@ jobs:
     vmImage: ubuntu-16.04
   steps:
     - bash: |
-        "$PWD/bin/brew" test-bot
+        HOMEBREW_REPOSITORY=/home/linuxbrew/.linuxbrew
+        sudo mkdir -p /home/linuxbrew
+        sudo mv "$PWD" "$HOMEBREW_REPOSITORY"
+        sudo ln -s "$HOMEBREW_REPOSITORY" "$PWD"
+        "$HOMEBREW_REPOSITORY/bin/brew" test-bot
       displayName: Run brew test-bot
       env:
         HOMEBREW_GITHUB_API_TOKEN: $(github.publicApiToken)
--- a/Library/Homebrew/diagnostic.rb
+++ b/Library/Homebrew/diagnostic.rb
@@ -791,6 +791,17 @@ module Homebrew
         end.join("\n")
       end
 
+      def check_homebrew_prefix
+        return if HOMEBREW_PREFIX.to_s == Homebrew::DEFAULT_PREFIX
+
+        <<~EOS
+          Your Homebrew's prefix is not #{Homebrew::DEFAULT_PREFIX}.
+          You can install Homebrew anywhere you want but some bottles (binary packages)
+          can only be used with a standard prefix and some formulae (packages)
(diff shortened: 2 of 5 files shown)


sjackman:
Looks good to me. Only need to address the failing `brew tests` by copying/symlinking Homebrew into the default prefix.

sjackman:
Thanks, Jon!

jonchang:
Ok, finally got CI to cooperate. Had to move the repository to `/home/linuxbrew/.linuxbrew`

azure-pipelines.yml
```diff
@@ -35,7 +35,13 @@ jobs:
...
   steps:
     - bash: |
-        "$PWD/bin/brew" test-bot
+        OLDPWD="$PWD"
+        HOMEBREW_REPOSITORY=/home/linuxbrew/.linuxbrew
+        sudo mkdir -p /home/linuxbrew
+        sudo mv "$PWD" "$HOMEBREW_REPOSITORY"
+        "$HOMEBREW_REPOSITORY/bin/brew" test-bot
+        sudo mkdir -p "$OLDPWD"
+        cd "$OLDPWD"
```
sjackman: I suggest…
```sh
	        HOMEBREW_REPOSITORY=/home/linuxbrew/.linuxbrew
	        sudo mkdir -p /home/linuxbrew
	        sudo mv "$PWD" "$HOMEBREW_REPOSITORY"
	        mkdir -p "$PWD"
	        "$HOMEBREW_REPOSITORY/bin/brew" test-bot
```

MikeMcQuaid: A symlink would be preferable.

MikeMcQuaid: So instead of the mv do an ln -s and mkdir the parent dir if necessary. You’ll also need to update the brew-test-bot repo similarly.

jonchang: I don't think this will work since [`bin/brew` resolves symlinks](https://github.com/Homebrew/brew/blob/fdffb395146d054b24947d127b6a1d25573473ce/bin/brew#L22-L24). I tried for a few hours yesterday to get it to work using the linking method that you suggest and gave up.

Steps to reproduce on macOS:

```
cd
ln -s /usr/local/ prefix
cd prefix
bin/brew config
```

The output will say `HOMEBREW_PREFIX: /usr/local` but should say `~/prefix/`

MikeMcQuaid: Thanks for trying. In that case you could move the directory and symlink it back to the original location.

jonchang:
Passed CI with the move + symlink method. Thanks @sjackman @MikeMcQuaid

MikeMcQuaid:
> You’ll also need to update the brew-test-bot repo similarly.

MikeMcQuaid:
> You’ll also need to update the brew-test-bot repo similarly.

This should have been done before this was merged as now CI has been broken there.

jonchang:
My apologies, I'll back out this change until I / we can fix the test-bot repo.

sjackman:
@jonchang Could you please reopen this PR to revert the reversion of PR https://github.com/Homebrew/brew/pull/5453?

PR #5469, opened 2019-01-02
shims/super/cc: Do not remove -Xpreprocessor argument

fxcoudert:

`brew tests` is failing because `Ignoring rainbow-2.2.2 because its extensions are not built.  Try: gem pristine rainbow --version 2.2.2`. If I run the recommended command, it's complaining that `You don't have write permissions for the /Library/Ruby/Gems/2.3.0 directory`. 😩 

-----

`-Xpreprocessor` is used to pass the next flag to the `clang` preprocessor. It is a rather obscure flag, and currently only used in the idiom `-Xpreprocessor -fopenmp` to force Apple's `clang` to compile OpenMP source files (which are later linked against our `libomp`).

However, currently superenv removes `-fopenmp` when it sees it, even though we want to keep it when it's used after `-Xpreprocessor`. This patch does that by passing any argument after `-Xpreprocessor` straight.

Even though I am not aware of any other use of `-Xpreprocessor` out there, I would argue that it's a better behaviour anyway (we never want `-Xpreprocessor foo bar` to become `-Xpreprocessor bar`).

### DIFF
--- a/Library/Homebrew/shims/super/cc
+++ b/Library/Homebrew/shims/super/cc
@@ -187,6 +187,9 @@ class Cmd
       "-fuse-linker-plugin", "-frounding-math"
       # clang doesn't support these flags
       args << arg unless tool =~ /^clang/
+    when "-Xpreprocessor"
+      # used for -Xpreprocessor -fopenmp
+      args << arg << enum.next
     when /-mmacosx-version-min=10\.(\d)/
       arg = "-mmacosx-version-min=10.9" if high_sierra_or_later? && $1.to_i < 9
       args << arg


fxcoudert:
With this patch, the pull request at https://github.com/Homebrew/homebrew-core/pull/35637 allows to build imagemagick with OpenMP support.

MikeMcQuaid:
> If I run the recommended command, it's complaining

What is that command?

> even though we want to keep it when it's used after `-Xpreprocessor`

What would be the consequences of this flag being passed in error e.g. without a desired `libomp` dependency?

fxcoudert:
Re. the tests failing locally for me, even on master:
- `brew tests` says `Ignoring rainbow-2.2.2 because its extensions are not built. Try: gem pristine rainbow --version 2.2.2` in some tests, which makes it fail because it's not the output it is expecting.
- Running `gem pristine rainbow --version 2.2.2` fails with `You don't have write permissions for the /Library/Ruby/Gems/2.3.0 directory`
But they're fine on CI, and it's unrelated to this PR.

---

> What would be the consequences of this flag being passed in error

I'm not sure how it would be passed in error. `-Xpreprocessor` is a really rarely used flag, and always paired with a further option for the preprocessor. If you want to make the code more specific, we could detect the pair `-Xpreprocessor -fopenmp` and check whether there is a `libomp` dependency. I don't think it's worth special-casing it further (and I don't know how to look up the next item in the `enum` without shifting it).

MikeMcQuaid:
> * `Try: gem pristine rainbow --version 2.2.2` in some tests, which makes it fail because it's not the output it is expecting.

@fxcoudert We don't control that message. The fix is to run `bundle pristine` in `/usr/local/Homebrew/Library/Homebrew/test`.

> I'm not sure how it would be passed in error

Ok. I'll defer to your judgement and we can address any other issues that may occur later.

PR #5482, opened 2019-01-04
info: Fix style: Use a guard clause

sjackman:
Fix the style error:
```
Library/Homebrew/cmd/info.rb:88:7: C:
Use a guard clause instead of wrapping the code inside a conditional expression.
```

### DIFF
--- a/Library/Homebrew/cmd/info.rb
+++ b/Library/Homebrew/cmd/info.rb
@@ -85,11 +85,8 @@ module Homebrew
   def info
     info_args.parse
     if args.json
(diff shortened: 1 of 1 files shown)


sjackman:
> Any ideas why @sjackman; what's different on Docker Hub?

No idea, sorry. My best guess would be a different version of `rubocop` somehow?

> I'd also argue if our CI is green then Docker Hub's should be too.

Yes, I agree. I just don't have the time right now to dig into it. Thanks for merging though.

PR #5500, opened 2019-01-08
Disable cop for #set_cpu_flags, #set_cpu_cflags

claui:
- [ ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb). **No tests, just disabling a cop**
- [ ] Have you successfully run `brew tests` with your changes locally? **No tests**

-----

In `extend/ENV/std.rb`, the methods `set_cpu_flags` and `set_cpu_cflags` have been violating the `Naming/AccessorMethodName` cop.

One of the changes in #5491 appears to have awakened the cop. That means we’re likely dealing with legacy code here, predating the introduction of RuboCop into Homebrew.

This PR disables the cop for either method. I believe refactoring wouldn’t currently be worth the effort, especially given that `#set_cpu_cflags` is now mostly called without arguments.

### DIFF
--- a/Library/Homebrew/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/ENV/std.rb
@@ -198,7 +198,7 @@ module Stdenv
   # Sets architecture-specific flags for every environment variable
   # given in the list `flags`.
   # @private
-  def set_cpu_flags(flags, map = Hardware::CPU.optimization_flags)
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Reproduced the issue in https://github.com/Homebrew/brew/pull/5497 so merging this. Thanks @claui!

MikeMcQuaid:
(after rerunning build for 💚)

sjackman:
Docker Hub is back to 💚. Thanks, Claudia!

PR #5501, opened 2019-01-08
Treat future SDKs as "latest"

indirect:
Some SDKs return a number higher than the hard-coded `latest_version`
in this file. With the current `==` check, those higher version numbers
are treated as "not the latest". This prevents machines with
higher-versioned SDKs from being able to use Homebrew.

To resolve that problem, this PR changes the check to `>=`, which allows
machines with higher-versioned SDKs to also use Homebrew to install
packages.

### DIFF
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -44,7 +44,7 @@ module OS
       end
 
       def latest_sdk_version?
-        OS::Mac.version == OS::Mac.latest_sdk_version
+        OS::Mac.version >= OS::Mac.latest_sdk_version
       end
 
       def needs_clt_installed?


MikeMcQuaid:
Can you be more specific about what OS and Xcode version you're encountering this issue on?

MikeMcQuaid:
> Completely hypothetically

We don't fix hypothetical bugs, sorry.


> if Homebrew reads your macOS version as "10.15" and your Xcode version as "11.0".

This would be a bug and not one fixed by this PR.

> It will also happen with any macOS 10.13+ as long as the Xcode version is higher than "10.1". Which includes "10.2", "11.0", "11.1", etc.

I'll ask again:

Can you be more specific about what OS and Xcode version you're encountering this issue on?

Please provide the details of the specific issue you're experiencing, the specific Xcode version you're running and macOS version you're running. If you're unable or unwilling to do that: please close this PR.

Thanks!

indirect:
Okay then, this specific patch definitely fixes a non-hypothetical problem on machines that Homebrew reads running as macOS 10.15 and Xcode 11.0.

Can you explain more about how that would be a bug and not one fixed by this PR?

PR #5505, opened 2019-01-08
Use -march=native for non-bottle source builds [Linux]

sjackman:

-----

Use `-march=native` for source builds on Linux. Brew currently uses `-march=core2` by default for both bottle and source builds. Previous behaviour was `-march=core2` for bottle builds and `-march=native` for source builds.


MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!

PR #5516, opened 2019-01-09
Change default compile option to O2

DesWurstes:
Definition from GCC docs: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

It's because Os harms performance a little, and O2 stays neutral.

Meanwhile: 1) Should I add `-mtune=native`, which doesn't change the instruction set 2) Should I apply the same change for CXXFLAGS?

This change is subjective, so comments are (especially) welcome.


MikeMcQuaid:
Sorry, we're not going to change the global CFLAGS without benchmarks of a significant proportion of the formulae that are affected.

Issue #5456, opened 2018-12-30
Allow "brew info ..." to show info for cask.


rhwood:
# A detailed description of the proposed feature

When using the command `brew info name`, if a formula `name` does not exist, but a cask `name` does exist, provide the user with the info for the cask, as if `brew cask info name` had been the issued command.

# The motivation for the feature

I frequently find that the first command issued after a `brew [cask] search ...` is `brew [cask] info ...` on one of the results returned from the search.

Since the consolidation of `brew search ...` and `brew cask search ...` commands and removal of `brew cask search ...`, I often find that I fail to use `brew cask info ...` and that `brew info ...` merely gives me an error getting the info for a result returned in the prior search.

It would be beneficial if the behaviors of search and info were the same in regard to handling of casks since the two commands are likely used together.

# How the feature would be relevant to at least 90% of Homebrew users

It would be beneficial if the behaviors of search and info were the same in regard to handling of casks since the two commands are likely used together by brew users (for example, searching for a youtube downloaders using `brew search youtube` returns 7 results (2 formulas and 5 casks). Being able to run a single command to get the infos for all results would be preferable to having to think about which command to use for which result.

# What alternatives to the feature have been considered

- Do nothing and maintain the current behavior.

- Have `brew info name` (and other commands that act on formulas suggest the `brew cask ...` equivalent command if both an equivalent cask command and a cask matching `name` exists).

MikeMcQuaid:
We would consider a pull request for this but have no plans for further integration of commands between `brew` and `brew cask` at this time, thanks!

MikeMcQuaid:
Actually, will leave this open and `help wanted` for now.

zachauten:
I’d like to take a crack at this if that’s alright!

I’ve currently got the feature working and I’d appreciate if someone could answer a quick question:

 I’d like to add tests that create dummy casks in the test setup/before, similar to how `info_spec.rb` calls `setup_test_formula "testball”` on line 5. What’s the standard way to generate that test data?

MikeMcQuaid:
> I’d like to take a crack at this if that’s alright!

Feel free.

> What’s the standard way to generate that test data?

I'm not sure. Check out other cask tests or just use one of the existing cask fixtures. Open a PR; it's easier to discuss code when able to look at it.

MikeMcQuaid:
Fixed by @zachauten's great PR in https://github.com/Homebrew/brew/pull/5538. Will be in the next tagged version of Homebrew/brew.

PR #5538, opened 2019-01-15
Make brew info output cask if no formula exists

zachauten:
…t exist

- [ x] Have you followed the guidelines in our [Contributing](https://github.com/Homebrew/brew/blob/master/CONTRIBUTING.md) document?
- [ x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x ] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb).
- [x ] Have you successfully run `brew style` with your changes locally?
- [x ] Have you successfully run `brew tests` with your changes locally?

-----
https://github.com/Homebrew/brew/issues/5456#issue-394876970


Library/Homebrew/cmd/info.rb
```diff
@@ -126,6 +130,11 @@ def print_info
           if (reason = MissingFormula.reason(f))
```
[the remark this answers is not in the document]
zachauten: You mean create a "reason" method on `MissingFormula` that searches for a cask and returns it's info printout? I think it would be a little cleaner. 

But there's a similar call to `MissingFormula.reason()` in `Homebrew/cmd/install.rb`, this could cause an attempting to call `brew install` on a missing formula to search for a cask with the same name and print it's info. Is this desirable?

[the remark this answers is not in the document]
MikeMcQuaid: @zachauten I think the reason method could suggest a Cask when a formula isn't available. In this case if you want it to be handled automatically for `brew info` to show casks I think it's still worth adding to `MissingFormula` but it would be good to search for a cask before you search for `MissingFormula` (which will take longer and print output about e.g. searching which seems unnecessary). Additionally, the output here for searching for casks is only necessary if it takes >1s on your machine.

Library/Homebrew/missing_formula.rb
```diff
@@ -1,11 +1,17 @@
...
+require "cask/cask_loader"
+require "cask/installer"
 
 module Homebrew
   module MissingFormula
     class << self
       def reason(name, silent: false)
-        blacklisted_reason(name) || tap_migration_reason(name) ||
-          deleted_reason(name, silent: silent)
+        search_for_cask(name)
```
MikeMcQuaid: The API for this is e.g. `MissingFormula.reason(query, silent: true)` so it will need to return a string with the Cask info or `nil` if the cask is not found. If there's any output at all it should be silenced if `silent: true` is passed.

Library/Homebrew/missing_formula.rb
```diff
@@ -167,6 +173,11 @@ def deleted_reason(name, silent: false)
         end
       end
 
+      def search_for_cask(name)
```
MikeMcQuaid: Add the `rescue` to this method instead of using it above and only rescue the specific exception(s) you need rather than all. Additionally, return the info result as a string rather than printing it.

Library/Homebrew/test/cmd/info_spec.rb
```diff
@@ -18,6 +18,14 @@
       .and not_to_output.to_stderr
       .and be_a_success
   end
+
+  it "looks for a Cask with the same name if a given Formula does not exist", :needs_network do
+    setup_remote_tap "homebrew/cask"
+
+    expect { brew "info", "firefox" }
+      .to output(/No available formula with the name "firefox"/).to_stderr
+      .and output(/firefox/).to_stdout
```
MikeMcQuaid: Can you make this check more specific to the cask output?

zachauten: Should I use a different cask besides firefox? I used it because it was used in similar integration tests in `search_spec.rb`, but if I make the test match the exact output I would need to use a regex or something to match the version number, which isn’t something that I’ve seen done in other tests.

Library/Homebrew/missing_formula.rb
zachauten: Would this mean refactoring the `Cask::Cmd::Info.info(cask)` method in `info.rb` to return the info instead of printing it?

MikeMcQuaid: Yes, perhaps.

Library/Homebrew/test/cmd/info_spec.rb
MikeMcQuaid: even matching e.g. `Mozilla Firefox` or `firefox: ` or the URL would be preferable.

Library/Homebrew/cask/cmd/info.rb
```diff
@@ -25,52 +26,62 @@ def self.help
         "displays information about the given Cask"
       end
 
+      def self.get_info(cask)
+        "#{title_info(cask)}\n"\
```
MikeMcQuaid: Use `<~EOS` which allows you to embed the newlines.

zachauten: Done! Also, do you know what I might need to do to get the CI to pass? The only test failure I can replicate locally is `brew tests --generic --online`, but `brew tests --online` passes. I'm not sure why the test would be different with the `--generic` flag.

Library/Homebrew/cask/cmd/info.rb
```diff
@@ -25,52 +26,64 @@ def self.help
         "displays information about the given Cask"
       end
 
+      def self.get_info(cask)
+        <<~EOS.chomp
+          #{title_info(cask)}
+          #{Formatter.url(cask.homepage) if cask.homepage}
```
MikeMcQuaid: This will make a newline if there is no Cask homepage. I think the previous version was probably better but just use it to keep adding to a string.

Library/Homebrew/cask/cmd/info.rb
```diff
@@ -25,52 +26,64 @@ def self.help
...
+                            .concat(" (")
+                            .concat(
+                              if versioned_staged_path.exist?
+                              then versioned_staged_path.abv
+                              else Formatter.error("does not exist")
+                              end,
+                            ).concat(")")
+            install_info << "\n"
           end
+          return install_info.strip
```
MikeMcQuaid: `return` is unnecessary here as it's the last statement.

Library/Homebrew/cask/cmd/info.rb
```diff
@@ -82,16 +95,18 @@ def self.repo_info(cask)
...
       end
 
       def self.artifact_info(cask)
-        ohai "Artifacts"
         cask.artifacts.each do |artifact|
           next unless artifact.respond_to?(:install_phase)
           next unless DSL::ORDINARY_ARTIFACT_CLASSES.include?(artifact.class)
 
-          puts artifact.to_s
+          return <<~EOS
```
MikeMcQuaid: `return` is unnecessary here as it's the last statement.

Library/Homebrew/test/cmd/info_spec.rb
```diff
@@ -18,6 +18,23 @@
       .and not_to_output.to_stderr
       .and be_a_success
   end
+
+  it "looks for a Cask with the same name if a given Formula does not exist", :needs_network do
+    setup_remote_tap "homebrew/cask"
+
+    expect { brew "info", "firefox" }
+      .to output(%r{Error:\sNo\savailable\sformula\swith\sthe\sname\s"firefox"\s\n
```
MikeMcQuaid: Replace `\s` with a space. Can include the line below but don't need to include every other line. Also, this output should probably note that this is the output for a cask.

MikeMcQuaid:
@zachauten Try to run `brew style` locally; it'll point out a bunch of the things I have. `brew style --fix` will fix some of them automatically.

Library/Homebrew/cask/cmd/info.rb
zachauten: Okay, you mean instead of using a multiline string with interpolation, just append the return values of each function to a value that gets returned by get_info?

Library/Homebrew/test/cmd/info_spec.rb
zachauten: I don't think I can replace `\s` with literal whitespace when I'm using free-spacing mode on the regular expression. I'm using free spacing mode to break the regex up across multiple lines for readability. But I think escaping the whitespace like `\ ` works if that's preferable to the `\s` pattern.

Do you have a preference for how to indicate the output is for a cask? Maybe just something like `Found the following cask:` to stdout?

zachauten:
Thanks for the tip about `brew style --fix`! That's really convenient. 

I actually remembered to run `brew style` locally for this commit (although I forgot it on the last one) and it passed, and I believe it passed on the Azure Pipeline as well. But I've seen it complain about unnecessary return statements before, so I'm not sure why it didn't didn't pick up on these.

Library/Homebrew/test/cmd/info_spec.rb
```diff
@@ -18,6 +18,13 @@
       .and not_to_output.to_stderr
       .and be_a_success
   end
+
+  it "looks for a Cask with the same name if a given Formula does not exist", :needs_network do
```
jonchang: This is failing `brew tests --generic --online`. Cask is macOS-only so the test will `:needs_macos` since it shouldn't be run on a generic OS.

Library/Homebrew/utils.rb
```diff
@@ -24,9 +24,13 @@ def require?(path)
   raise unless e.message.include?(path)
 end
 
-def ohai(title, *sput)
+def ohai_title(title)
   title = Tty.truncate(title) if $stdout.tty? && !ARGV.verbose?
-  puts Formatter.headline(title, color: :blue)
+  Formatter.headline(title, color: :blue)
+end
```
jonchang: Not sure I understand this change. What's wrong with `oh1` for this case?

Library/Homebrew/cask/cmd/fetch.rb
```diff
@@ -12,7 +12,7 @@ def initialize(*)
 
       def run
         casks.each do |cask|
-          Installer.print_caveats(cask)
+          puts Installer.print_caveats(cask)
```
jonchang: Total nitpick but it feels odd to `puts` a method that is named `print_caveats`

Library/Homebrew/cask/installer.rb
```diff
@@ -86,7 +88,7 @@ def install
 
       check_conflicts
 
-      print_caveats
+      print print_caveats
```
jonchang: Again this feels very odd. Maybe the function should be renamed?

Library/Homebrew/cask/cmd/info.rb
```diff
@@ -25,52 +26,70 @@ def self.help
...
+                              then versioned_staged_path.abv
+                              else Formatter.error("does not exist")
+                              end,
+                            ).concat(")")
+            install_info << "\n"
           end
+          install_info
         else
-          puts "Not installed"
+          "Not installed\n"
```
jonchang: Another bikeshed but it feels strange to have a function returning strings with newlines in them.

Library/Homebrew/test/cmd/info_spec.rb
zachauten: Thanks so much for this!

zachauten:
@jonchang Thanks for the review! I think you've brought up good points.

Library/Homebrew/utils.rb
zachauten: I needed a function that would return the formatted string that `ohai` prints, instead of printing it immediately with `puts`.

Library/Homebrew/cask/cmd/info.rb
zachauten: I'll try to move newlines out of the helper methods and into `get_info` where I can; In this case if I removed this newline I would have to strip the newline from `install_info` when it gets returned on line 73, because it gets appended in a loop on line 71. Stripping a newline in a callee method only to re-append it in the caller also feels funky to me.

Library/Homebrew/cask/cmd/info.rb
```diff
@@ -25,52 +26,69 @@ def self.help
         "displays information about the given Cask"
       end
 
+      def self.get_info(cask)
+        output = title_info(cask) + "\n"
+        output << Formatter.url(cask.homepage) + "\n" if cask.homepage
+        output << installation_info(cask)
+        repo = repo_info(cask)
+        output << repo + "\n" unless repo.nil?
```
MikeMcQuaid: `if repo`

Library/Homebrew/cask/cmd/info.rb
```diff
@@ -25,52 +26,69 @@ def self.help
...
+        output << name_info(cask)
+        language = language_info(cask)
+        output << language unless language.nil?
```
MikeMcQuaid: `if language`

Library/Homebrew/missing_formula.rb
```diff
@@ -187,6 +190,16 @@ def deleted_reason(name, silent: false)
         end
       end
 
+      def cask_reason(name, silent: false)
+        return if silent
+
+        cask = Cask::CaskLoader.load(name)
+        reason = "Found the following cask named \"#{name}\" instead:\n"
+        reason << Cask::Cmd::Info.get_info(cask) unless silent
```
MikeMcQuaid: Can remove `unless silent`; you've already checked above.

Library/Homebrew/missing_formula.rb
```diff
@@ -187,6 +190,16 @@ def deleted_reason(name, silent: false)
         end
       end
 
+      def cask_reason(name, silent: false)
+        return if silent
+
+        cask = Cask::CaskLoader.load(name)
+        reason = "Found the following cask named \"#{name}\" instead:\n"
+        reason << Cask::Cmd::Info.get_info(cask)
```
MikeMcQuaid: Testing this out: I'm not sure what I think about outputting the `brew cask info` output if you `brew install firefox` and it fails; I'm interested what other @Homebrew/maintainers think. It might be enough in that case to just output the `Found the following cask...` and skip the `brew cask info` bit (but still output that at `brew info` time).

Library/Homebrew/test/cmd/info_spec.rb
```diff
@@ -20,6 +20,13 @@
   end
 end
 
+describe "brew info", :integration_test, :needs_macos do
```
MikeMcQuaid: Could this be in `Library/Homebrew/test/missing_formula_spec.rb` instead? Integration tests are dramatically slower so we want to avoid them where we can.

Library/Homebrew/missing_formula.rb
vitorgalvao: Agreed. If I tried an `install` command and got an `info` output, I’d be confused and start wondering what I typed wrong. Yes, the reason is there, but it could get lost in the rest of the (larger, familiar, and ignorable) `info` text.

Since this equates to a failed command, it should output only the minimum amount of relevant information to “fix”.

zachauten: I pushed a solution that has MissingFormula.cask_reason being called only during the `brew info` command, let me know what you think.

MikeMcQuaid: @zachauten I think it would still be good to print out that a Cask was found but just not the `brew info` output. You could add an argument to `def reason` which is passed to `cask_reason` to conditionally output the `brew cask info` output.

MikeMcQuaid:
Incredible work here @zachauten and thanks for all your patience; it's definitely been worth it. I hope to see more contributions from you and thanks again for all your hard work: you rock! ❤️ 🎉

zachauten:
@MikeMcQuaid Thank you! This was my first time opening a PR on GitHub and I just want to say I really appreciate you being so helpful and patient with me. It was a lot of fun and I’m looking forward to hopefully contributing more to homebrew in the future!

MikeMcQuaid:
@zachauten Wow, that's impressive work for your first GitHub PR, I wish my first one was this good 😂. You're very welcome: look forward to more work from you and feel free to contact me directly if you want any ideas or pointers!

PR #5560, opened 2019-01-18
cat: Use CLI::Parser to parse args

GauthamGoli:

-----
#1860

### DIFF
--- a/Library/Homebrew/cmd/cat.rb
+++ b/Library/Homebrew/cmd/cat.rb
@@ -1,16 +1,29 @@
 #:  * `cat` <formula>:
 #:    Display the source to <formula>.
 
+require "cli_parser"
+
 module Homebrew
   module_function
 
+  def cat_args
+    Homebrew::CLI::Parser.new do
+      usage_banner <<~EOS
+        `cat` <formula>
+
+        Display the source to <formula>.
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cmd/cat.rb
```diff
@@ -1,16 +1,29 @@
...
     # do not "fix" this to support multiple arguments, the output would be
     # unparsable, if the user wants to cat multiple formula they can call
     # brew cat multiple times.
     formulae = ARGV.formulae
     raise FormulaUnspecifiedError if formulae.empty?
-    raise "`brew cat` doesn't support multiple arguments" if formulae.size > 1
+    raise "`brew cat` doesn't support multiple arguments" if args.remaining.size > 1
 
     cd HOMEBREW_REPOSITORY
     exec "cat", formulae.first.path, *ARGV.options_only
```
GauthamGoli: @MikeMcQuaid  seems like `brew cat` accepts other CLI args of `cat` command. What options do we plan on supporting ?

MikeMcQuaid: @GauthamGoli I think we can avoid supporting any arguments really; I don't think they are widely used.

PR #5595, opened 2019-01-23
CaskSha256MismatchError: link to fix

vitorgalvao:

-----

Haven’t run the tests because I’m on an old version of macOS with a semi-broken ruby install (I really need to do a clean upgrade to this machine) and it fails to even install the proper gems. That said, the change is a message modification, so I don’t expect it to fail, hence why I submitted anyway.

This is part of https://github.com/Homebrew/homebrew-cask/pull/57844. People keep submitting issues for this problem without following the documentation, even if it’s right in front of them in the issue template. So let’s put it right in their face at the time of the error.

### DIFF
--- a/Library/Homebrew/cask/exceptions.rb
+++ b/Library/Homebrew/cask/exceptions.rb
@@ -136,7 +136,9 @@ module Cask
         Actual:   #{Formatter.error(actual.to_s)}
         File:     #{path}
 
-        To retry an incomplete download, remove the file above.
+        To retry an incomplete download, remove the file above. If the issue persists, visit:
+
+          https://github.com/Homebrew/homebrew-cask/blob/master/doc/reporting_bugs/checksum_does_not_match_error.md
       EOS
     end
   end


Library/Homebrew/cask/exceptions.rb
```diff
@@ -136,7 +136,9 @@ def to_s
         Actual:   #{Formatter.error(actual.to_s)}
         File:     #{path}
 
-        To retry an incomplete download, remove the file above.
+        To retry an incomplete download, remove the file above. If the issue persists, visit this page:
```
claui:
```suggestion
        To retry an incomplete download, remove the file above. If the issue persists, visit:
```

Library/Homebrew/cask/exceptions.rb
```diff
@@ -136,7 +136,9 @@ def to_s
         Actual:   #{Formatter.error(actual.to_s)}
         File:     #{path}
 
-        To retry an incomplete download, remove the file above.
+        To retry an incomplete download, remove the file above. If the issue persists, visit:
+
```
MikeMcQuaid: Maybe kill the empty line?

vitorgalvao: I’ve added the line for consistency with our `caveats`. When we add an indented line with a single command / web address, we tend to have the empty line as well.

But if not having the line would make it consistent with other core error messages, I’m fine with it.

PR #5597, opened 2019-01-23
Added missing method to_a to Cask::DSL:ConflictsWith class.

cmbernard333:

-----
#5596

### DIFF
--- a/Library/Homebrew/cask/cask.rb
+++ b/Library/Homebrew/cask/cask.rb
@@ -133,7 +133,7 @@ module Cask
         end,
         "caveats"        => caveats,
         "depends_on"     => depends_on,
-        "conflicts_with" => conflicts_with.to_a,
+        "conflicts_with" => conflicts_with.to_h,
         "container"      => container,
         "auto_updates"   => auto_updates,
       }
--- a/Library/Homebrew/cask/dsl/conflicts_with.rb
+++ b/Library/Homebrew/cask/dsl/conflicts_with.rb
@@ -26,10 +26,8 @@ module Cask
         end
       end
 
-      def to_a
-        conflicts={}
-        @pairs.each { |k, v| conflicts[k]=v }
-        conflicts
+      def to_h
+        Hash[VALID_KEYS.map { |key| [key, instance_variable_get("@#{key}").to_a] }]
       end
     end
   end
--- a/Library/Homebrew/cask/dsl/conflicts_with.rb
+++ b/Library/Homebrew/cask/dsl/conflicts_with.rb
@@ -25,8 +25,9 @@ module Cask
           instance_variable_set("@#{key}", instance_variable_get("@#{key}").merge([*value]))
         end
       end
+
       def to_a
-        (@pairs.values)
+        @pairs.values
       end
     end
   end
--- a/Library/Homebrew/cask/dsl/conflicts_with.rb
+++ b/Library/Homebrew/cask/dsl/conflicts_with.rb
@@ -25,6 +25,9 @@ module Cask
           instance_variable_set("@#{key}", instance_variable_get("@#{key}").merge([*value]))
         end
       end
+      def to_a
+        (@pairs.values)
+      end
     end
   end
 end
--- a/Library/Homebrew/cask/dsl/conflicts_with.rb
+++ b/Library/Homebrew/cask/dsl/conflicts_with.rb
@@ -27,7 +27,9 @@ module Cask
       end
 
       def to_a
-        @pairs.values
+        conflicts={}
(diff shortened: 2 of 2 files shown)


Library/Homebrew/cask/dsl/conflicts_with.rb
```diff
@@ -25,6 +25,10 @@ def initialize(pairs = {})
           instance_variable_set("@#{key}", instance_variable_get("@#{key}").merge([*value]))
         end
       end
+
+      def to_a
```
reitermarkus: I think this should rather be a hash.

cmbernard333: conflicts_with is initialized with a hash that becomes @pairs in the class context. Should it just return the @pairs instance variable?

reitermarkus: It should merge the pairs into a hash, i.e.

```json
{ "cask": ["firefox-beta", "firefox-esr"] }
```

Library/Homebrew/cask/dsl/conflicts_with.rb
```diff
@@ -25,6 +25,12 @@ def initialize(pairs = {})
...
+        conflicts={}
+        @pairs.each { |k, v| conflicts[k]=v }
```
reitermarkus: This will override keys if there are multiple of the same. This should work:

```ruby
Hash[VALID_KEYS.map { |key| [key, instance_variable_get("@#{key}")] }]
```

Library/Homebrew/cask/dsl/conflicts_with.rb
```diff
@@ -25,6 +25,12 @@ def initialize(pairs = {})
           instance_variable_set("@#{key}", instance_variable_get("@#{key}").merge([*value]))
         end
       end
+
+      def to_a
```
reitermarkus: Also this has to be renamed to `to_h`.

Library/Homebrew/cask/dsl/conflicts_with.rb
reitermarkus: The test failures don't seem to be related to this.

Library/Homebrew/cask/dsl/conflicts_with.rb
cmbernard333: Thanks for the feedback. Didn't realize there would be overlapping values. I'll update it.

Library/Homebrew/cask/dsl/conflicts_with.rb
cmbernard333: Will have to update `cask.rb` in that case - it uses `to_a` or atleast expects it to be converted to an array instead of a hash.

Library/Homebrew/cask/dsl/conflicts_with.rb
cmbernard333: I'm not sure we want to iterate through the whole list - I think we just want to pull in the `cask` field from the instance vars so it resembles the `conflicts_with` field in traditional formula

`brew info --json=v1 moreutils`
```
...
"conflicts_with":["parallel","pwntools","task-spooler"]
```

Library/Homebrew/cask/cask.rb
```diff
@@ -133,7 +133,7 @@ def to_h
         end,
         "caveats"        => caveats,
         "depends_on"     => depends_on,
-        "conflicts_with" => conflicts_with.to_a,
+        "conflicts_with" => conflicts_with.to_h,
```
reitermarkus: Does it work if you leave off `.to_h` altogether? Neither of the other values call this explicitly.

cmbernard333: It tries to produce the object string representation like this ```"conflicts_with":"#<Cask::DSL::ConflictsWith:0x00007f94700138d8>"```

Library/Homebrew/cask/dsl/conflicts_with.rb
reitermarkus: Casks can conflict with both formulae and casks, however.

cmbernard333: I saw this during some local testing. I've updated the code to do the full iteration and produce, hopefully, the correct output.

Library/Homebrew/cask/cask.rb
reitermarkus: Ah, `depends_on` works without it because it subclasses `DelegateClass(Hash)`.

reitermarkus:
Thanks, @cmbernard333!

PR #5619, opened 2019-01-26
install: Use CLI::Parser to parse args

GauthamGoli:

-----
#1860

### DIFF
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -132,7 +132,7 @@ module Homebrew
         description: "If <formula> defines it, install the HEAD version, aka master, trunk, unstable."
       switch "--fetch-HEAD"
       switch "--keep-tmp",
-        description: "Dont delele the temporary files created during installation."
+        description: "Dont delete the temporary files created during installation."
       switch "--build-bottle",
         description: "Prepare the formula for eventual bottling during installation."
       switch :force,
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -129,7 +129,10 @@ module Homebrew
         description: "If <formula> defines it, install the development version."
       switch "--HEAD",
         description: "If <formula> defines it, install the HEAD version, aka master, trunk, unstable."
-      switch "--fetch-HEAD"
+      switch "--fetch-HEAD",
+        description: "Fetch the upstream repository to detect if the HEAD installation of the "\
+                     "formula is outdated. Otherwise, the repository's HEAD will be checked for "\
+                     "updates when a new stable or devel version has been released."
       switch "--keep-tmp",
         description: "Dont delete the temporary files created during installation."
       switch "--build-bottle",
--- a/Library/Homebrew/cmd/install.rb
+++ b/Library/Homebrew/cmd/install.rb
@@ -180,209 +180,207 @@ module Homebrew
       end
     end
 
-    begin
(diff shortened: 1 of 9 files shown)


Library/Homebrew/cmd/install.rb
```diff
@@ -79,20 +79,100 @@
...
+        description: "Print the verification and postinstall steps."
+      switch "--display-times",
+        description: "Print install times for each formula at the end of the run."
+      switch "-i", "--interactive",
+        description: "Download and patch <formula>, then open a shell. This allows the user to "\
+                     "run `./configure --help` and otherwise determine how to turn the software "\
+                     "package into a Homebrew package."
+      switch "-g", "--git",
+        description: "Create a Git repository, useful for creating patches to the software."
+      unless ARGV.formulae&.empty?
```
MikeMcQuaid: `if ARGV.formulae.present?`

MikeMcQuaid: I'm not 100% sure this is needed though; are you sure `ARGV.formulae.each` fails with no formulae?

Library/Homebrew/cmd/install.rb
```diff
@@ -317,11 +397,11 @@ def install_formula(f)
     fi = FormulaInstaller.new(f)
     fi.options              = build_options.used_options
     fi.invalid_option_names = build_options.invalid_option_names
```
MikeMcQuaid: Might be a good idea to use the `build_options.invalid_option_names` above instead of the formulae loop?

MikeMcQuaid:
Would be good to use/replace this logic: https://github.com/Homebrew/brew/blob/06702829080aaf1e8b741ff34f7a2f4118b7e411/Library/Homebrew/formula_installer.rb#L270-L272

Library/Homebrew/cmd/install.rb
GauthamGoli: Yeah, you are right! It won't fail. ( I was just trying to avoid NPE, now realised `ARGV.formulae` would be an empty array instead of `nil` )

Library/Homebrew/cmd/install.rb
GauthamGoli: Sorry, I'm not sure I understand. Can you explain this more?

MikeMcQuaid: @GauthamGoli If we're only looking at externally passed options I think it's fine/better to error out rather than ignoring and messaging 👍

Library/Homebrew/cmd/install.rb
```diff
@@ -79,20 +79,95 @@
...
+      switch "--display-times",
+        description: "Print install times for each formula at the end of the run."
+      switch "-i", "--interactive",
+        description: "Download and patch <formula>, then open a shell. This allows the user to "\
+                     "run `./configure --help` and otherwise determine how to turn the software "\
+                     "package into a Homebrew package."
+      switch "-g", "--git",
+        description: "Create a Git repository, useful for creating patches to the software."
+
+      ARGV.formulae.each do |f|
```
GauthamGoli: @MikeMcQuaid If a non existent formula name is passed, there would be an exception due to the way `ARGV.formulae` is handled. (Hence some tests in `install_spec` would fail)

MikeMcQuaid: @GauthamGoli Yup, these exceptions are handled by the `rescue` at the bottom of the `install` method.

PR #5620, opened 2019-01-26
brew: define CI env variable for travis and circleci too

iMichka:

-----

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -65,8 +65,8 @@ do
   export "$VAR_NEW"="${!VAR}"
 done
 
-# Set CI variable for Azure Pipelines, Jenkins.
-if [[ -n "$TF_BUILD" || -n "$JENKINS_HOME" ]]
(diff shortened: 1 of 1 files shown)


iMichka:
We do not use that variable for the moment in the travis/circleci context, but there are probably some places where we could.

@sjackman also started using the CI env variable here: https://github.com/Linuxbrew/brew/pull/925.
The code is still being run in linuxbrew-core when a commit is pushed to origin. That should not happen, I need to see why it is running with Linuxbrew/brew and not Homebrew/brew, but that is a different topic.

sjackman:
CircleCI sets `CI` as well.
See https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

iMichka:
Right. I got confused by another bug and thought it was due to the changes made around that env variable. I did not know that it was already set for Travis and circle.

PR #5633, opened 2019-01-28
cleaner: only skip non-.la symlinks

RandomDSdevel:
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?  (Yes, see my commit message.)  
- [ ] Have you written new tests for your changes?  [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb).  (No, as I believe the existing tests already cover this sufficiently well.  This can be changed on request.)  
- [ ] Have you successfully run `brew tests` with your changes locally?  (All tests pass locally except for:  
   - '`test/os/mac_spec.rb`,' which fails with an unrelated error stating that, "`'Locale::parse' failed to parse 'OS::Mac::language'`."
   - Some formula package installation tests due only to unhandled OS version warnings.)  

-----

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I decided only to fix the logic used for cleaning the results of builds from source to handle symbolic links to static libraries and have punted on tracking down where the logic for cleaning up after bottle builds is and unifying that with the former for now.  

Reopens and closes Homebrew/homebrew-core#35269.

### DIFF
--- a/Library/Homebrew/cleaner.rb
+++ b/Library/Homebrew/cleaner.rb
@@ -88,10 +88,12 @@ class Cleaner
 
       Find.prune if @f.skip_clean? path
 
-      next if path.symlink? || path.directory?
+      next if path.directory?
 
       if path.extname == ".la"
         path.unlink
+      elsif path.symlink?
+        # Skip it.
       elsif path.basename.to_s == "perllocal.pod"
         # Both this file & the .packlist one below are completely unnecessary
         # to package & causes pointless conflict with other formulae. They are


Library/Homebrew/cleaner.rb
```diff
@@ -88,10 +88,19 @@ def clean_dir(d)
 
       Find.prune if @f.skip_clean? path
 
-      next if path.symlink? || path.directory?
+      if path.symlink?
+        if path.extname != ".la" || path.readlink.extname != ".la"
```
MikeMcQuaid: This is a duplication of the logic below. Instead consider moving the `next if path.symlink?` below the `.la` check below and having just `next if path.directory?` here.

Library/Homebrew/cleaner.rb
```diff
@@ -88,10 +88,19 @@ def clean_dir(d)
...
+          next
+        end
+      end
+      next if path.directory?
 
       if path.extname == ".la"
         path.unlink
+      elsif path.symlink?
+        if path.extname == ".la" || path.readlink.extname == ".la"
```
MikeMcQuaid: The first part will already be handled by `path.extname == ".la"` above. For the latter part please add a check to make sure it's `path.symlink? && path.exist?` as otherwise `path.readlink` will fail and throw an exception.

sjackman:
@MikeMcQuaid Why do we delete libtool's `.la` files? `libtool` can make use them.

MikeMcQuaid:
@sjackman 🤷‍♂️ check `git blame` I guess. If no-one has noticed/found them necessary at this stage I'm hesitant to remove this logic for no reason.

sjackman:
Libtool uses `.la` files to make the linking of both dynamic and static libraries consistent across platforms. For example, linking to a dynamic library `libfoo.so` automatically pulls in its dependencies like `libbar.so`, so that `gcc -o hello hello.c -lfoo` is sufficient and `-lbar` is not needed. That's not true of static libraries though, where you need to specify all recursive dependencies, `gcc -o hello hello.c -lfoo -lbar`. Libtools provides a consistent interface for linking against both dynamic and static libraries. The needed information is stored in its `.la` files. Anyway, that's why they exist. We've managed without them, like you say, so we can keep on keeping on, but I see no reason to delete them either. They're very small ASCII files. Debian distributes them.
https://packages.ubuntu.com/search?searchon=contents&keywords=.la&mode=&suite=cosmic&arch=amd64

MikeMcQuaid:
> Anyway, that's why they exist. We've managed without them, like you say, so we can keep on keeping on

👍

RandomDSdevel:
@MikeMcQuaid:  

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Comments addressed locally, but I'm still running '`brew tests`.'

Library/Homebrew/cleaner.rb
```diff
@@ -88,10 +88,14 @@ def clean_dir(d)
...
 
-      next if path.symlink? || path.directory?
+      next if path.directory?
 
       if path.extname == ".la"
         path.unlink
+      elsif path.symlink? && path.exist?
+        if path.readlink.extname == ".la"
+          path.unlink
+        end
```
sjackman: I suggest…
```ruby
      elsif path.symlink?
        path.unlink if path.exist? && path.realpath.extname == ".la"
```
This code handles the case where the symlink does not have a `.la` extension, and that symlink links to a file that does have a `.la` extension. Do you have an example of that case? If not, I suggest the simpler…
```ruby
      elsif path.symlink?
        # Skip symlinks
```
The symlinks with `.la` extension will be removed by line 93 `if path.extname == ".la"`.

RandomDSdevel: @sjackman:  True, this is likely a case of 'YAGNI.'

RandomDSdevel:
@sjackman:  Comment addressed.

Library/Homebrew/cleaner.rb
```diff
@@ -88,10 +88,17 @@ def clean_dir(d)
...
 
       if path.extname == ".la"
         path.unlink
+      elsif path.symlink?
+        # Skip it.
+        #
+        # (Note that the case where the symbolic link has no extension or a
+        # different one entirely but its _target's_ `path.extname == ".la"`
+        # _could_ get handled here, but such a case is unlikely enough that this
+        # logic ignores it for now.)
```
sjackman: Thanks for the detailed comment. It probably also is not needed, since this hypothetical case likely will not arise. I suggest replacing lines 96 to 101 with `# Skip symlinks.`

MikeMcQuaid: Skip the comment altogether; even `# Skip symlinks` is self-explanatory from the code.

sjackman: When the body of an if statement is empty, I tend to include a comment to indicate that it's left empty intentionally. Personal preference though. I'm fine with either.

MikeMcQuaid: Yes, fair. Either is fine.

RandomDSdevel:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I don't think the test failure here is related to my changes, as I haven't touched the '`man`' pages at all.  Is anybody else experiencing transient issues?  (In any case, let's try running CI again.  @BrewTestBot test this please.)

RandomDSdevel:
@BrewTestBot test this please.

MikeMcQuaid:
Thanks @RandomDSdevel!

RandomDSdevel:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;You're welcome, @MikeMcQuaid (and @sjackman, too!)

PR #5653, opened 2019-02-01
Cask handle already downloaded file

FuegoFro:
The current logic for `curl_download` tries to restart any existing partial download, but this fails for some servers when the file is already fully downloaded, returning a status code 416. This updates the function to check for the file being already downloaded and if so bails early.

The following command used to fail (as did the added test, without the fix) when run more than once but now works as expected:
```
brew cask info https://raw.githubusercontent.com/homebrew/homebrew-cask/master/Casks/docker.rb
```

-----

### DIFF
--- a/Library/Homebrew/utils/curl.rb
+++ b/Library/Homebrew/utils/curl.rb
@@ -51,10 +51,20 @@ def curl_download(*args, to: nil, **options)
   destination = Pathname(to)
   destination.dirname.mkpath
 
-  continue_at = if destination.exist? &&
-                   curl_output("--location", "--range", "0-1",
-                               "--write-out", "%{http_code}",
-                               "--output", "/dev/null", *args, **options).stdout.to_i == 206 # Partial Content
+  range_stdout = curl_output("--location", "--range", "0-1",
+                             "--dump-header", "-",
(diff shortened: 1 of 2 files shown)


Library/Homebrew/utils/curl.rb
```diff
@@ -51,10 +51,20 @@ def curl_download(*args, to: nil, **options)
...
+  range_stdout = curl_output("--location", "--range", "0-1",
+                             "--dump-header", "-",
+                             "--write-out", "%{http_code}",
+                             "--output", "/dev/null", *args, **options).stdout
+  headers, _, http_status = range_stdout.partition("\r\n\r\n")
+
+  supports_partial_download = http_status.to_i == 206 # Partial Content
+  if supports_partial_download &&
+     destination.exist? &&
+     destination.size == %r{^.*Content-Range: bytes \d+-\d+/(\d+)\r\n.*$}m.match(headers)[1].to_i
```
MikeMcQuaid: Can this handle if the match fails?

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @FuegoFro!

PR #5661, opened 2019-02-03
With symlinked Cellar, don't try to find /usr/local prefix

sjackman:

-----

This PR permits Cellar to be a symlink. A Cellar may be a symlink to a common shared Cellar. For example, `~/work/project1/Cellar` may be a symlink to `/usr/local/Cellar`. This change allows using a private prefix of `~/work/project1/bin` with a shared Cellar of `/usr/local/Cellar`. This change enables a `brew env` feature that @jonchang is working on.

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -39,7 +39,7 @@ then
 fi
 
 # Try to find a /usr/local HOMEBREW_PREFIX where possible (for bottles)
-if [[ -L "/usr/local/bin/brew" ]]
+if [[ -L "/usr/local/bin/brew" && ! -L "$HOMEBREW_PREFIX/Cellar" ]]
 then
   USR_LOCAL_BREW_FILE_DIRECTORY="$(symlink_target_directory "/usr/local/bin/brew" "/usr/local/bin")"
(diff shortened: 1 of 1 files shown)


Library/Homebrew/brew.sh
```diff
@@ -13,6 +13,12 @@ else
   HOMEBREW_CELLAR="$HOMEBREW_PREFIX/Cellar"
 fi
 
+# If Cellar is a symlink, resolve it.
+if [[ -L "$HOMEBREW_CELLAR" ]]
+then
+  HOMEBREW_CELLAR=$(cd "$HOMEBREW_CELLAR"; pwd -P)
```
MikeMcQuaid: Can this use `safe_cd`?

sjackman: Yes. Thanks for the tip. Done.

sjackman:
I've update this PR. The previous PR did three related but separable things:

1. Disable magic to find /usr/local prefix when Cellar is a symlink
2. Disable brew doctor Cellar symlink check
3. Resolve Cellar symlink to its real path

I've reduced this PR to just the first, which is needed by the `brew env` feature that @jonchang is working on. 2 and 3 can wait until I have more time for testing.

sjackman:
Merged the slimmed down version of this PR.

PR #5678, opened 2019-02-05
missing_formula.rb: add help message for cargo

waldyrious:
...modeled after the existing message for pip.

Also sync format of the gfortran and lzma messages,
and use typographic apostrophe in the *tex message.

Fixes https://github.com/Homebrew/homebrew-core/issues/36560.

### DIFF
--- a/Library/Homebrew/missing_formula.rb
+++ b/Library/Homebrew/missing_formula.rb
@@ -15,7 +15,7 @@ module Homebrew
         EOS
         when "tex", "tex-live", "texlive", "latex" then <<~EOS
           Installing TeX from source is weird and gross, requires a lot of patches,
-          and only builds 32-bit (and thus can't use Homebrew dependencies)
+          and only builds 32-bit (and thus can’t use Homebrew dependencies)
 
           We recommend using a MacTeX distribution: https://www.tug.org/mactex/
 
@@ -36,8 +36,10 @@ module Homebrew
           You can read more about it at:
             #{Formatter.url("https://github.com/MacRuby/MacRuby")}
         EOS
-        when /(lib)?lzma/
-          "lzma is now part of the xz formula."
+        when /(lib)?lzma/ then <<~EOS
+          lzma is now part of the xz formula, and can be installed with:
+            brew install xz.
+        EOS
         when "gtest", "googletest", "google-test" then <<~EOS
           Installing gtest system-wide is not recommended; it should be vendored
           in your projects that use it.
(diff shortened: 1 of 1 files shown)


Library/Homebrew/missing_formula.rb
```diff
@@ -15,7 +15,7 @@ def blacklisted_reason(name)
         EOS
         when "tex", "tex-live", "texlive", "latex" then <<~EOS
           Installing TeX from source is weird and gross, requires a lot of patches,
-          and only builds 32-bit (and thus can't use Homebrew dependencies)
+          and only builds 32-bit (and thus can’t use Homebrew dependencies)
```
MikeMcQuaid: Please change this back, thanks!

Library/Homebrew/missing_formula.rb
```diff
@@ -36,8 +36,10 @@ def blacklisted_reason(name)
           You can read more about it at:
             #{Formatter.url("https://github.com/MacRuby/MacRuby")}
         EOS
-        when /(lib)?lzma/
-          "lzma is now part of the xz formula."
+        when /(lib)?lzma/ then <<~EOS
+          lzma is now part of the xz formula, and can be installed with:
+            brew install xz.
```
MikeMcQuaid: Good idea but remove the trailing `.`

Library/Homebrew/missing_formula.rb
```diff
@@ -79,6 +81,9 @@ def blacklisted_reason(name)
           If you wish to use the 2.x release you can install with Homebrew Cask:
             brew cask install ngrok
         EOS
+        when "cargo" then <<~EOS
+          Homebrew provides cargo via: `brew install rust`.
```
MikeMcQuaid: How about:

```
cargo is part of the rust formula, and can be installed with:
  brew install rust
```

Library/Homebrew/missing_formula.rb
waldyrious: Sure. I only included this because github's syntax highlighter seems to choke on the single quote.

Library/Homebrew/missing_formula.rb
waldyrious: I thought about that too, but I wasn't sure about using something so close to the "is _now_ part of..." message, and decided to model the message after the one for pip, which is a nearly identical use case.

That said, if we change this, the pip one should also be adjusted in the same way, right?

Library/Homebrew/missing_formula.rb
waldyrious: Oh! not sure how that slipped in unnoticed. Will change 👍

Library/Homebrew/missing_formula.rb
MikeMcQuaid: @waldyrious Yeh, changing `pip` too sounds good! Consistency in general is great 👍

waldyrious: Cool, will do.

waldyrious:
@MikeMcQuaid updated with the changes you suggested. Thanks for the prompt review!

Issue #5671, opened 2019-02-04
"java-beta" cask breaks brew's java dependencies


gcbartlett:
- [X] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [X] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [X] ran `brew update` and can still reproduce the problem?
- [X] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [X] ran `brew config` and `brew doctor` and included their output with your issue?

With the `java-beta` cask installed, dependencies on `java` fail even when the correct java installations are present (e.g. installing/upgrading `tomcat` fails).  This appears similar to issue #2405, except that in this case the output to `java -version` has no `'.'` character, so I suspect this causes problems for brew's regexp (see below):
```
$ brew cask install java-beta
==> Satisfying dependencies
==> Downloading https://download.java.net/java/early_access/jdk12/30/GPL/openjdk-12-ea+30_osx-x64_bin.tar.gz
Already downloaded: /Users/garybartlett/Library/Caches/Homebrew/downloads/1baf871f2fd7e0eb7836531d05029647d13ec5afebf1cc43c09d98833f1e26c2--openjdk-12-ea 30_osx-x64_bin.tar.gz
==> Verifying SHA-256 checksum for Cask 'java-beta'.
==> Installing Cask java-beta
🍺  java-beta was successfully installed!

$ brew config
HOMEBREW_VERSION: 2.0.0-16-g81b6623
ORIGIN: https://github.com/Homebrew/brew
HEAD: 81b6623774089c5eaff9eb52066d46717e56f286
Last commit: 74 minutes ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 385c946a4ad879cde422d33b84fdbcccb2b7d5bd
Core tap last commit: 9 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_DEV_CMD_RUN: 1
HOMEBREW_GITHUB_API_TOKEN: set
[...]

PR #5682, opened 2019-02-06
Add support for java version strings without periods, e.g. "12-ea"

gcbartlett:
Make the minor version number optional when extracting the version number from `java -version` output.  This should allow version strings such as "12-ea" (present in the latest `java-beta` cask) to be recognized.

Fixes https://github.com/Homebrew/brew/issues/5671.

### DIFF
--- a/Library/Homebrew/requirements/java_requirement.rb
+++ b/Library/Homebrew/requirements/java_requirement.rb
@@ -129,7 +129,7 @@ class JavaRequirement < Requirement
   end
 
   def satisfies_version(java)
-    java_version_s = system_command(java, args: ["-version"], print_stderr: false).stderr[/\d+(\.\d)?/]
+    java_version_s = system_command(java, args: ["-version"], print_stderr: false).stderr[/\d+(\.\d+)?/]
     return false unless java_version_s
 
     java_version = Version.create(java_version_s)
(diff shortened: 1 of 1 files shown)


Library/Homebrew/requirements/java_requirement.rb
```diff
@@ -129,7 +129,7 @@ def oracle_java_os
   end
 
   def satisfies_version(java)
-    java_version_s = system_command(java, args: ["-version"], print_stderr: false).stderr[/\d+\.\d/]
+    java_version_s = system_command(java, args: ["-version"], print_stderr: false).stderr[/\d+(\.\d)?/]
```
reitermarkus: The second number could be > 9, so let's also use `\d+` for it.

gcbartlett: OK, done.  I've never seen a java minor version > 9 (only ..., 1.6, 1.7, 1.8, 9.0, 10.0, 11.0, 12) but that's easy to account for.  My only question in this hypothetical scenario would be whether the existing code handles any "+" comparison correctly, e.g. would it treat "1.11" as > "1.8"?

reitermarkus: As long as the existing code uses the `Version` class, comparisons are handled appropriately.

PR #5683, opened 2019-02-06
Bugfix: Calculate actual delta in usage

tueksta:

-----
I read the guidelines three times, I do not understand it. Most of it seems to apply to formula changes only? Maybe needs better headline/section structure.

Changes:
This fixes a bug, that when you do a cleanup and the cleanup fails for a path (e.g. no permissions to remove path), the size of that path would still be added to the cleanup estimate. By calculating the delta of disk usage before and after the removal, we get a clear picture of how much disk space was saved.

### DIFF
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -321,7 +321,7 @@ module Homebrew
         yield
       end
 
-      @disk_cleanup_size += disk_usage
+      @disk_cleanup_size += disk_usage - path.disk_usage
     end
 
     def cleanup_lockfiles(*lockfiles)
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -316,12 +316,13 @@ module Homebrew
 
       if dry_run?
         puts "Would remove: #{path} (#{path.abv})"
+        @disk_cleanup_size += disk_usage
       else
         puts "Removing: #{path}... (#{path.abv})"
         yield
+        @disk_cleanup_size += disk_usage - path.disk_usage
       end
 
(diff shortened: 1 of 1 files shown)


reitermarkus:
> the cleanup fails for a path (e.g. no permissions to remove path)

Wouldn't it abort completely in that case?

tueksta:
No, that's how I discovered this bug. Kept saying 15MB were freed, and that it couldn't delete python 3.7.2 and that python has 15MB in the path. So I wondered if it means the 15MB were freed by something else, but all successive rerans it still kept saying 15MB were cleaned and that it couldn't delete python 3.7.2.

Library/Homebrew/cleanup.rb
```diff
@@ -321,7 +321,7 @@ def cleanup_path(path)
         yield
       end
 
-      @disk_cleanup_size += disk_usage
+      @disk_cleanup_size += disk_usage - path.disk_usage
```
reitermarkus: The correction should only be applied when actually removing the files, not when `dry_run?` is `true`.

tueksta: Should it really? Would be weird if you ran a dry run and it says 1.5GB will be freed, and then you run it and it doesn't free 1.5GB. So maybe we should show both values, minimum and maximum that can be freed?

tueksta: No wait, you're totally right, that wouldn't work at all :)

MikeMcQuaid:
Closed and reopened to re-trigger CI.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @tueksta!

PR #5690, opened 2019-02-07
Only create target directory if it doesn’t exist.

reitermarkus:

-----

Fixes https://github.com/Homebrew/homebrew-cask/issues/58564.

### DIFF
--- a/Library/Homebrew/cask/artifact/moved.rb
+++ b/Library/Homebrew/cask/artifact/moved.rb
@@ -40,13 +40,13 @@ module Cask
         end
 
         ohai "Moving #{self.class.english_name} '#{source.basename}' to '#{target}'."
-        if target.dirname.parent.writable?
+        if target.dirname.ascend.find(&:directory?).writable?
           target.dirname.mkpath
         else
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Was just about to submit a PR for this.

Won’t it work if we just change line 43 from `if target.dirname.parent.writable?` to `if target.dirname.writable?` (i.e. remove `parent.`)? The `dirname` of `target` is the directory we’re moving it to, so why check its parent?

I’ve tried `if target.dirname.writable?` both with `app` and with `java`, and they both worked.

reitermarkus:
I have updated it with a different solution.

Library/Homebrew/cask/artifact/moved.rb
```diff
@@ -40,13 +40,13 @@ def move(force: false, command: nil, **options)
         end
 
         ohai "Moving #{self.class.english_name} '#{source.basename}' to '#{target}'."
-        if target.dirname.parent.writable?
+        if target.dirname.ascend.detect(&:directory?).writable?
```
vitorgalvao: `style` suggests `find` over `detect`.

robamaton:
Awesome, thanks! Any idea when this might be released?

PR #5695, opened 2019-02-08
Enable Formula#plist stanza to accept a {hash}

### LINKED ISSUE #5691: Proposal: Formula#plist stanza update to enable ability to specify a Ruby {hash} instead of raw <xml>
# A detailed description of the proposed feature

The Formula#`plist` stanza could be refactored to enable the ability to specify a Ruby `{hash}` instead of raw `<xml>`.

So this could enable something like:
~~~ rb
def plist
  {
    Label: plist_name,
    ProgramArguments: ["#{opt_bin}/example", "--do-this"],
    RunAtLoad: true,
    KeepAlive: true,
    StandardErrorPath: "/dev/null",
    StandardOutPath: "/dev/null"
  }
end
~~~
which would populate the `<dict>` (with the surrounding `<DOCTYPE plist>` etc. boilerplate taken care of behind the scenes), instead of:
~~~ rb
def plist; <<~EOS
    <?xml version="1.0" encoding="UTF-8"?>
    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
    <plist version="1.0">
      <dict>
        <key>Label</key>
          <string>#{plist_name}</string>
        <key>ProgramArguments</key>
        <array>
          <string>#{opt_bin}/example</string>
          <string>--do-this</string>
        </array>
        <key>RunAtLoad</key>
        <true/>
        <key>KeepAlive</key>
        <true/>
        <key>StandardErrorPath</key>
        <string>/dev/null</string>
        <key>StandardOutPath</key>
        <string>/dev/null</string>
      </dict>
    </plist>
  EOS
end
~~~
🤢

# The motivation for the feature

This would be much better, and cut down on a lot of the duplicate boilerplate across many formulae.

# How the feature would be relevant to at least 90% of Homebrew users

Maybe 90% is a stretch… but the easier a formula is to create (and importantly maintain), the better for everyone…

# What alternatives to the feature have been considered

I couldn't find any related issues.

danielbayley:
Closes #5691

-----

### DIFF
--- a/Library/Homebrew/formula.rb
+++ b/Library/Homebrew/formula.rb
@@ -872,39 +872,32 @@ class Formula
     @active_log_type = old_log_type
   end
 
-  # This method can be overridden to provide a plist.
+  # This method can be used to provide a plist.
   # @see https://www.unix.com/man-page/all/5/plist/ Apple's plist(5) man page
-  # <pre>def plist; <<~EOS
-  #  <?xml version="1.0" encoding="UTF-8"?>
-  #  <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-  #  <plist version="1.0">
-  #  <dict>
-  #    <key>Label</key>
-  #      <string>#{plist_name}</string>
-  #    <key>ProgramArguments</key>
-  #    <array>
-  #      <string>#{opt_bin}/example</string>
-  #      <string>--do-this</string>
-  #    </array>
-  #    <key>RunAtLoad</key>
-  #    <true/>
-  #    <key>KeepAlive</key>
-  #    <true/>
-  #    <key>StandardErrorPath</key>
-  #    <string>/dev/null</string>
-  #    <key>StandardOutPath</key>
-  #    <string>/dev/null</string>
-  #  </dict>
-  #  </plist>
-  #  EOS
+  # <pre>plist do
+  #  {
+  #    Label: "#{plist_name}.override",
+  #    ProgramArguments: ["#{opt_bin}/example", "--do-this"],
+  #    RunAtLoad: true,
+  #    KeepAlive: true,
+  #    StandardErrorPath: "/dev/null",
+  #    StandardOutPath: "/dev/null",
+  #  }
   # end</pre>
   def plist
-    nil
+    return if self.class.plist.empty?
+    @plists ||= {}
+    self.class.plist.each do |block|
+      plist_hash = instance_eval(&block)
+      plist_hash[:Label] ||= plist_name
+      @plists[plist_hash[:Label]] = plist_hash.to_plist
+    end
+    @plists
   end
 
   # The generated launchd {.plist} service name.
   def plist_name
-    "homebrew.mxcl." + name
+    self.class.plist_name("homebrew.mxcl." + name)
   end
 
   # The generated launchd {.plist} file path.
@@ -2443,6 +2436,18 @@ class Formula
     def plist_options(options)
       @plist_startup = options[:startup]
       @plist_manual = options[:manual]
+    end
+
+    # @private
+    def plist_name(label)
+      label
+    end
+
+    # @private
+    def plist(&block)
+      @plist_hashes ||= []
+      return @plist_hashes unless block_given?
+      @plist_hashes << block
     end
 
     # @private
--- a/Library/Homebrew/formula_installer.rb
+++ b/Library/Homebrew/formula_installer.rb
@@ -15,6 +15,7 @@ require "cache_store"
 require "linkage_checker"
 require "install"
 require "messages"
+require "plist"
 
 class FormulaInstaller
   include FormulaCellarChecks
@@ -863,12 +864,18 @@ class FormulaInstaller
   end
 
   def install_plist
-    return unless formula.plist
+    plist = formula.plist
+    return unless plist
 
-    formula.plist_path.atomic_write(formula.plist)
-    formula.plist_path.chmod 0644
-    log = formula.var/"log"
-    log.mkpath if formula.plist.include? log.to_s
+    plist = { formula.plist_name => plist } if plist.is_a? String
+
+    plist.each do |label, xml|
+      plist_path = formula.plist_path.sub(formula.plist_path.stem, label)
+      plist_path.atomic_write(xml)
+      plist_path.chmod 0644
+      log = formula.var/"log"
+      log.mkpath if formula.plist.include? log.to_s
+    end
   rescue Exception => e # rubocop:disable Lint/RescueException
     onoe "Failed to install plist file"
     ohai e, e.backtrace if debug?


Library/Homebrew/formula_installer.rb
```diff
@@ -863,9 +864,13 @@ def link(keg)
   end
 
   def install_plist
-    return unless formula.plist
+    return unless plist = formula.plist
```
MikeMcQuaid: Not sure I understand this bit; why use `formula.plist` vs. `plist` below?

Library/Homebrew/formula.rb
```diff
@@ -874,29 +874,15 @@ def with_logging(log_type)
...
-  #  </plist>
-  #  EOS
+  # <pre>def plist
+  #  {
+  #    Label: plist_name,
+  #    ProgramArguments: ["#{opt_bin}/example", "--do-this"],
+  #    RunAtLoad: true,
+  #    KeepAlive: true,
+  #    StandardErrorPath: "/dev/null",
+  #    StandardOutPath: "/dev/null"
```
MikeMcQuaid: Add a trailing comma on this line.

Library/Homebrew/formula_installer.rb
danielbayley: Maybe there is a more proper way I am missing, but for some reason trying to reassign `formula.plist =` just throws a "`plist=` undefined method" error… doing it this way gets around that.

danielbayley:
> Could you open a PR on homebrew-core too to change a few formulae so we can take a look at that side of the API too?

@MikeMcQuaid Sure, just any random couple of formulas that use `def plist` will do?

Library/Homebrew/formula_installer.rb
```diff
@@ -863,9 +864,13 @@ def link(keg)
...
 
-    formula.plist_path.atomic_write(formula.plist)
+    if formula.plist.is_a? Hash
+      plist = `/usr/bin/plutil -convert xml1 -o - - <<< '#{plist.to_json}'`
```
colindean: 👍 nice

MikeMcQuaid:
> @MikeMcQuaid Sure, just any random couple of formulas that use `def plist` will do?

@danielbayley Yes, please. Ideally a little variation between them so we can get a feel of what it looks like.

reitermarkus:
We already have the `Plist` gem as a vendored dependency, use that instead of `plutil`.

danielbayley:
> We already have the `Plist` gem as a vendored dependency, use that instead of `plutil`.

@reitermarkus I did wonder about that actually… even better!

Library/Homebrew/formula_installer.rb
```diff
@@ -863,9 +864,11 @@ def link(keg)
   end
 
   def install_plist
-    return unless formula.plist
+    return unless plist = formula.plist
 
-    formula.plist_path.atomic_write(formula.plist)
+    plist = plist.to_plist if plist.is_a? Hash
```
reitermarkus: Check for `plist.respond_to?(:to_plist)` instead.

danielbayley:
> @danielbayley Yes, please. Ideally a little variation between them so we can get a feel of what it looks like.

@MikeMcQuaid See #36831

danielbayley:
It's kind of a bummer that Rubocop insists we have to use `:key => value` syntax for this… but still beats having to use `<xml>` 🤢

reitermarkus:
We could change the RuboCop rule to `no_mixed_keys`, which would allow both styles.

MikeMcQuaid:
> We could change the RuboCop rule to `no_mixed_keys`, which would allow both styles.

I'm game for this 👍

zbeekman:
I'm eager to see this merged, it may be helpful for #5743. Is there anything I can do to help? What are we waiting on?

MikeMcQuaid:
@zbeekman See discussion in https://github.com/Homebrew/homebrew-core/pull/36831

danielbayley:
> What are we waiting on?

Me 😅 Having the time to have another run at this, along with my current lack of familiarity with the core codebase (I usually try and stay away from `class` heirarchy/OOP stuff… I much prefer functional programming, and even in Ruby tend to stick to lambdas etc…)

zbeekman:
> Me 😅 Having the time to have another run at this, along with my current lack of familiarity with the core codebase

No worries, I'm more or less in the same boat.

MikeMcQuaid:
@danielbayley Feel free to ask for any questions/clarification on your pull requests and I'll do my best to help in a timely fashion.

danielbayley:
@MikeMcQuaid @reitermarkus So the discussion on this code was happening over at https://github.com/Homebrew/homebrew-core/pull/36831#discussion_r257304430, but it makes more sense to move it here where the actual implementation is…

I just pushed what I have [kind of] working so far…

> Yes, the `plist do` I suggested is `def self.plist`, while the one used by the formula is just `def plist`. So `plist` will have to forward the call to `self.plist`.

Right, so the `plist do` in the actual formula passes the block to the _instance_ method, which forwards it to the _class_ method:
~~~ rb
# tap/Formula/dnsmasq.rb
plist do
  {
    Label: "sh.brew.dnsmasq",
    ProgramArguments: %w[#{opt_sbin}/dnsmasq --keep-in-foreground -C #{etc}/dnsmasq.conf],
    RunAtLoad:        true,
    KeepAlive:        true,
  }
end

# Library/Homebrew/formula.rb
def self.plist
  @plists ||= {}

  if block_given?
    plist = yield
    plist[:Label] ||= self.plist_name
    @label = plist[:Label]

    def self.plist_name
      @label
    end

    @plists[@label] = plist.to_plist
  else
    @plists[@label]
  end
end
~~~
But then the problem became how do I access the `plist_name` instance method from within this class method (because `plist_name` [and by extension `plist_path`] and the `Label` need to match)… and the only way that seemed possible was to do the same thing, which is to say [re]`def`ine it twice, both as an instance and a class method. Doing it this way works, but throws up _another_ problem; which is that now referencing `name` inside the replacement class method returns `homebrew.mxcl.Formulary//FormulaNamespace{shasum}//Name` instead of just the formula name string as it did before:
~~~ rb
def self.plist_name
  "homebrew.mxcl." + name # is now homebrew.mxcl.Formulary//FormulaNamespace{sha}//Name…
end

def plist_name
  # whereas `name` here was just `Name`
  self.class.plist_name
end
~~~
How do I get the plain `name` string back?

> and get it with
```ruby
plist 'sh.brew.dnsmasq'
```
@reitermarkus What did you mean specifically by "_get_ it"; I know the method you sketched out returns the `@plists[label]` content given a label as opposed to a block which instead assigns it, but from where do I need to call it exactly, and how in that context would I even know what label to use as a reference, given that `plist_name` and/or `name` are much harder to access, and that a given formula can override these with a non default/generated label?

Also, one last question; is the point of implementing it this way to be able to call `plist do` multiple times in a given formula, resulting in each plist file being written? I mean, is the `install_plist` method (inside `formula_installer.rb`) supposed to run each time `do plist` is called, or would it need to [somehow access] and `map` over the `@plists` instance variable we assign to in the `def self.plist` class method? Because as it stands, this semi-working implementation I just pushed only writes out the last `plist do` definition in the formula…

> Yeh, in general it's worth looking at the existing plist and other DSL code to see how it works and it requires both an instance and class method to be defined.

I did look at the others initially, but none of them really work in exactly the same way as this DSL method needs to.

I have to say this is a nightmare to implement even a simple feature compared to functional programming 🤦‍😅 But progress has been made… so just those 2 major issues to hash out I think. I hope that all makes sense?

reitermarkus:
> How do I get the plain `name` string back?

How is `def name` implemented?

> What did you mean specifically by "_get_ it"

What I meant by this is that you don't define a new method for every plist but have a single one which takes the label of the plist you want it to return.

So in the default case, it would be `plist plist_name do` to define it and `plist plist_name` to _get_ it.

danielbayley:
> How is `def name` implemented?

There's no such method (there is in `Cask::DSL`, but that is unrelated). `name` is defined on initialisation:
~~~ rb
# @private
def initialize(name…
  @name = name
~~~
So is there a way to access this from inside the `self.plist_name` _class_ method?

> What I meant by this is that you don't define a new method for every plist but have a single one which takes the label of the plist you want it to return.

Sure, but isn't the goal here to be able to define multiple plists in a single formula… is that even possible with this approach? Also, for these cases would an array of hashes be a better API anyway; something like the in original suggestion?

> So in the default case, it would be `plist plist_name do` to define it and `plist plist_name` to _get_ it.

Right, but I'm not sure how that's supposed to work when a custom label string is supplied (either as an argument to the `self.plist` class method, or from the `Label:` property (as in the current implimentation)… from which other place would we _get_ it, and at that point how do we even know which label string was supplied as an override? Wouldn't it have to first deal set the returned value of `plist_name` anyway? Does that make sense?

reitermarkus:
> be able to define multiple plists in a single formula… is that even possible with this approach?

Yes, by passing any label other than `plist_name`.

> I'm not sure how that's supposed to work when a custom label string is supplied

How are custom labels specified currently?

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,41 @@ def plist_startup
...
+  #  {
+  #    Label: plist_name,
+  #    ProgramArguments: ["#{opt_bin}/example", "--do-this"],
+  #    RunAtLoad: true,
+  #    KeepAlive: true,
+  #    StandardErrorPath: "/dev/null",
+  #    StandardOutPath: "/dev/null",
+  #  }
+  # end</pre>
+  def self.plist
```
reitermarkus: This needs to take a label as argument.

Library/Homebrew/formula.rb
```diff
@@ -872,39 +872,14 @@ def with_logging(log_type)
...
+  # The generated launchd {.plist} service name.
+  def self.plist_name
+    "homebrew.mxcl." + name
   end
 
-  # The generated launchd {.plist} service name.
+  # @private
   def plist_name
-    "homebrew.mxcl." + name
+    self.class.plist_name
```
reitermarkus: This should actually stay as is.

Library/Homebrew/formula.rb
reitermarkus: However I cannot find a good way to get the formula `name` in here, so we might need to use a placeholder for `plist_name` here and then insert it in `def plist`.

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,41 @@ def plist_startup
     self.class.plist_startup
   end
 
+  # This method can be overridden to provide a plist.
```
reitermarkus: Use instead of overridden in this case.

Library/Homebrew/formula.rb
danielbayley: So then we just have `"homebrew.mxcl." + name` duplicated in both `def plist_name` instance, and `def self.plist_name` class methods? (`name` of course being still inacessible from the later… and I don't think it's possible to do the reverse, which is to say forward the `def self.plist_name` class method to this instance method 🤦‍♂️)

Library/Homebrew/formula.rb
danielbayley: > This needs to take a label as argument. And if this argument is `nil`, use `plist_name` instead.

That's basically what the code is already doing as it stands, except this is being passed as the `Label` within the hash instead of as a direct argument to the method, as that seems like a nicer API to me, but this could be switched… are you suggesting that the latter approach will make things run any differently in terms of being able to define multiple `plist do` calls in a single formula?

> However I cannot find a good way to get the formula `name` in here,

This part is the fundamental problem… I'd rather get this working first, before we decide exactly how it looks.

> so we might need to use a placeholder for `plist_name` here and then insert it in `def plist`.

A placeholder? Do you mean some kind of `FORMULA_NAME` string that we `sub` inside of `def plist` just because we have access to `name` there… or something less hacky?

This feels like trying to code inside of a cage 🙈😅

danielbayley:
> How are custom labels specified currently?

Just by re`def`ining the `plist_name` method… and this only works for the first plist file. Formulas that need multiple just call `prefix/"custom.label".write` directly… not ideal.

Library/Homebrew/formula.rb
reitermarkus: > Do you mean some kind of `FORMULA_NAME` string that we `sub` inside of `def plist` just because we have access to `name` there… or something less hacky?

Yes, that's exactly what I mean. The alternative would be to refactor `Formulary` to load the formula class with the name.

danielbayley:
So now the remaining problem here is that no matter how I seem to implement this class method, the `formula_installer.rb#install_plist` method only writes out a single defined `plist do` block in the test formula.

This is the other way of passing in the `Label` as just another hash entry… which as mentioned previously seems like a nicer API to me:
~~~ rb
def self.plist
  @plists ||= {}

  if block_given?
    plist = yield
    plist[:Label] ||= self.plist_name
    @label = plist[:Label]

    def self.plist_name
      @label
    end

    @plists[@label] = plist.to_plist
  else
    @plists[@label]
  end
end
~~~
If no `Label` is defined, then the default is the same as doing it the other way @reitermarkus suggested… but either way makes no difference in terms of being able to run `install_plist` for each defined `plist do` block in the test formula… @MikeMcQuaid Do you have any ideas on how to get that working?

MikeMcQuaid:
> Do you have any ideas on how to get that working?

If I understand correctly: I think it's an acceptable limitation that the label needs defined in order for more than one plist to be used. Ideally the label would only need defined on the 2nd, 3rd, etc. Thoughts?

danielbayley:
> If I understand correctly: I think it's an acceptable limitation that the label needs defined in order for more than one plist to be used. Ideally the label would only need defined on the 2nd, 3rd, etc. Thoughts?

@MikeMcQuaid Sure, it would be… but from testing both approaches, I don't think whether the label is passed directly as a argument to the class method, or as the `Label` property of the {hash} passed as the block makes any difference… that seems like just a stylistic choice. The problem is that the `install_plist` method only ever writes one file. Does that make sense?

MikeMcQuaid:
> The problem is that the `install_plist` method only ever writes one file. Does that make sense?

Yup. That method will likely need modified. It's not an external/formula API so that should't be problematic.

MikeMcQuaid:
> I don't think whether the label is passed directly as a argument to the class method, or as the `Label` property of the {hash} passed as the block makes any difference… that seems like just a stylistic choice.

Agreed it's a stylistic choice but if we can infer that information that seems preferable to requiring it be stated explicitly.

danielbayley:
> Agreed it's a stylistic choice but if we can infer that information that seems preferable to requiring it be stated explicitly.

Right, which is why I prefer [optionally] passing in the `Label` along with the rest of the hash.

> Yup. That method will likely need modified. It's not an external/formula API so that should't be problematic.

@MikeMcQuaid But I don't even think modifying the `install_plist` method solves this problem because it only ever _runs_ once; no matter how many `plist do`s are defined in the formula… so what to do?

MikeMcQuaid:
> @MikeMcQuaid But I don't even think modifying the `install_plist` method solves this problem because it only ever _runs_ once; no matter how many `plist do`s are defined in the formula… so what to do?

@danielbayley Modify it to be `install_plists` and install multiple plists?

danielbayley:
I think I finally cracked it! The solution was to have `def self.plist` return the whole `@plists` instance variable ({hash}) instead of trying to somehow access each item in the hash later by `label` from somewhere else, as @reitermarkus suggested (the OO codebase feels way too restrictive for that to be even practical… as demonstrated trying to access `name`) But now I can loop over that from the `install_plist` method and it works.

So I have now implemented it in a way that shouldn't interfere with any existing formulae containing `def plist`, but enables the new `plist do` block (which can be specified multiple times, and should eventually replace the old `def plist <<~EOF` way).

@MikeMcQuaid @reitermarkus @colindean @zbeekman Please test and review…

danielbayley:
Looks like I spoke too soon 🙈

> I have now implemented it in a way that shouldn't interfere with any existing formulae

…that is, if access to variables like `opt_sbin` can be maintaned inside the `plist do` blocks somehow? See https://github.com/Homebrew/homebrew-core/pull/36831#issuecomment-470865297

Library/Homebrew/formula.rb
```diff
@@ -922,6 +898,41 @@ def plist_startup
...
+  # end</pre>
+  def self.plist
+    @plists ||= {}
+
+    if block_given?
+      plist = yield
+      plist[:Label] ||= self.plist_name
+      @label = plist[:Label]
+
+      def self.plist_name
```
MikeMcQuaid: Feels weird to define this inside the method rather than its own method.

Library/Homebrew/formula.rb
```diff
@@ -922,6 +898,41 @@ def plist_startup
...
+        @label
+      end
+
+      @plists[@label] = plist.to_plist
+    else
+      return @plists
```
MikeMcQuaid: This `return` is unnecessary in Ruby.

MikeMcQuaid: (`brew style` will tell you that, `brew style --fix` should fix that).

MikeMcQuaid:
Nice work @danielbayley 👏. Getting closer.

> …that is, if access to variables like `opt_sbin` can be maintaned inside the `plist do` blocks somehow? See [Homebrew/homebrew-core#36831 (comment)](https://github.com/Homebrew/homebrew-core/pull/36831#issuecomment-470865297)

Those are instance methods but you're trying to invoke them inside a `yield` in a class method. You'll need to `yield` them in an instance method instead.

Library/Homebrew/formula.rb
danielbayley: Yeah I am aware Ruby has implicit returns… it just seemed more readable. No problem

Library/Homebrew/formula.rb
danielbayley: Yes, but how else could this work?

MikeMcQuaid: Refactor the existing `plist_name` method.

danielbayley:
> You'll need to yield them in an instance method instead.

@MikeMcQuaid How would I do that? The block (or anything else for that matter) doesn't get passed to the `def plist` instance method, which is why we had to make it a class method in the first place…

MikeMcQuaid:
> @MikeMcQuaid How would I do that? The block (or anything else for that matter) doesn't get passed to the `def plist` instance method, which is why we had to make it a class method in the first place…

Looking at `test do` may help. I'm afraid I don't have the resources to help you in more depth than that but @reitermarkus might be able to.

Library/Homebrew/formula.rb
danielbayley: Fixed

reitermarkus:
`test do` doesn't do anything special, it's basically just a proxy for `def test`.

Library/Homebrew/formula.rb
```diff
@@ -872,39 +872,15 @@ def with_logging(log_type)
...
-  #    <string>/dev/null</string>
-  #  </dict>
-  #  </plist>
-  #  EOS
-  # end</pre>
-  def plist
-    nil
+  # The generated launchd {.plist} service name.
+  def self.plist_name(label = nil)
+    plist_name = name.split("::").last.gsub(/([a-z])([A-Z])/, '\1-\2').downcase
```
reitermarkus: This won't work well. The class name cannot be easily reverted, since the original name can contain either `-` or `_`, both of which are stripped, `+` is converted to `x` and `@` is converted to `AT`.

reitermarkus:
I think the only way to do this right now is to make the class method populate a hash which maps the label to the given block. Then the instance method fetches the block and evaluates it in the instance.

Library/Homebrew/formula.rb
danielbayley: > This won't work well.

It feels too hacky anyway…

> The alternative would be to refactor Formulary to load the formula class with the name.

This sounds like the more _correct_ approach… but how exactly do I expose it? From inside `self.load_formula`?

danielbayley:
> Looking at test do may help.

> test do doesn't do anything special, it's basically just a proxy for def test.

~~~ rb
def plist(&block)
  define_method(:plist, &block)
end
~~~
@MikeMcQuaid Yeah it isn't particularly helpful. Although copying that approach does restore access to instance methods like `opt_sbin` and so passes the correct hash to `install_plist`… it also reverts the ability to have multiple `plist do` blocks work. This would be the way to go if we didn't need multiple plists in some formulae.

> I think the only way to do this right now is to make the class method populate a hash which maps the label to the given block.

Or we can just store the `block`s in an array so that we can still keep the `Label` inside the `&block`… I had already tried this, and although I can access this array from both the `def plist` instance method, and `install_plist` method, `.call`ing one of these blocks throws up the same problem in that it still throws `undefined` errors on instance methods like `opt_sbin`.

> Then the instance method fetches the block and evaluates it in the instance.

@reitermarkus Does this differ from the approach I just explained? How do I evaluate the block in the correct context?

danielbayley:
Ok, just came back to this and pretty sure I definitely have it this time! 🎉 

I needed `instance_eval` in the right place and was missing the `&` (`to_proc`) in `&block` when passing it in…

@MikeMcQuaid @reitermarkus @colindean @zbeekman All rewritten formulas work (see https://github.com/Homebrew/homebrew-core/pull/36831) and `brew style` passes on all modified files… should be finally ready to merge?🤞

Library/Homebrew/formula.rb
danielbayley: Never mind… think I found a way to avoid this altogether!

MikeMcQuaid:
`brew style` is still failing. I'll review this again when `brew style` and `brew tests` are both passing. Just a heads up that I may still need to request changes after that. Thanks for all your hard work on this @danielbayley ❤️

danielbayley:
> `brew style` is still failing.
@MikeMcQuaid Fixed. I don't think any of the test failures have anything to do with this PR… the same ones fail on master.

MikeMcQuaid:
@danielbayley These tests are not failing on `master`.

Library/Homebrew/formula.rb
```diff
@@ -872,39 +872,14 @@ def with_logging(log_type)
...
-  end
-
   # The generated launchd {.plist} service name.
   def plist_name
-    "homebrew.mxcl." + name
+    self.class.plist_name("homebrew.mxcl." + name)
+  end
+
+  # @private
+  def self.plist_name(label)
```
reitermarkus: How is this function useful?

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,32 @@ def plist_startup
...
+  #  {
+  #    Label: "#{plist_name}.override",
+  #    ProgramArguments: ["#{opt_bin}/example", "--do-this"],
+  #    RunAtLoad: true,
+  #    KeepAlive: true,
+  #    StandardErrorPath: "/dev/null",
+  #    StandardOutPath: "/dev/null",
+  #  }
+  # end</pre>
+  def plist
```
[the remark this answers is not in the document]
reitermarkus: It should still return a string when no argument is given.

Library/Homebrew/formula.rb
danielbayley: @reitermarkus Because without it I get an `undefined method `plist_name'` error… and doing it this way (essentially inverting the behaviour of the previous implementation) I have access to `name` which is now passed from the instance method to the class method and everything works, without having to resort to messy hacks like we had before.

Library/Homebrew/formula.rb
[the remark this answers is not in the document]
danielbayley: @reitermarkus The [previous](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb#L895-L897) version of this method was literally just
~~~ rb
def plist
  nil
end
~~~
so it only returned `nil` and didn't take any arguments… only there to be overriden if specified in a formula (which do currently return a raw xml string - the point of this PR being to avoid having to do that eventually) in which case `formula.plist` when called from `formula_installer.rb` is not `nil` and so passes the `return` and gets written. I don't see how this breaks that because `def plist` in a formula still just overrides this method and works as before… Have you tested it? Both [multiple] `plist do` blocks and `def plist` (overriden mehtods) in a test formula work as expected for me. Am I missing something here?

danielbayley:
> @danielbayley These tests are not failing on `master`.

@MikeMcQuaid Ah I had tests failing on `master` on my machine… but they are actually different tests looking at that situation again. Looks like the (seemingly unrelated) `caveats_spec` expects the `def plist` instance method to return `nil`… although I think maybe the test should be changed to instead check for `.empty?` (because now we use an array (or hash with `.to_h`) This would remove the need for the [extra LOC](https://github.com/Homebrew/brew/pull/5695/files#diff-b4a6fcb9de3a9163fd8cdac432ad8270R918) I just added to fix it. What do you think?

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
     self.class.plist_startup
   end
 
+  # This method can be used to provide a plist.
```
MikeMcQuaid: Can you move this method to the same location as the old one? It makes seeing the diff easier.

Library/Homebrew/formula.rb
MikeMcQuaid: Please move all `self.` methods to under `class << self` in `formula.rb` to be consistent.

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
...
+  #    RunAtLoad: true,
+  #    KeepAlive: true,
+  #    StandardErrorPath: "/dev/null",
+  #    StandardOutPath: "/dev/null",
+  #  }
+  # end</pre>
+  def plist
+    plists = self.class.plist.map do |block|
+      plist = instance_eval(&block)
+      label = plist[:Label] ||= plist_name
```
MikeMcQuaid: Skip the `label =` here

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
...
+  #    ProgramArguments: ["#{opt_bin}/example", "--do-this"],
+  #    RunAtLoad: true,
+  #    KeepAlive: true,
+  #    StandardErrorPath: "/dev/null",
+  #    StandardOutPath: "/dev/null",
+  #  }
+  # end</pre>
+  def plist
+    plists = self.class.plist.map do |block|
+      plist = instance_eval(&block)
```
MikeMcQuaid: `plist_hash = ` means you don't end up with the confusing `plist.to_plist` below.  I'm also wondering if there's a new method or other variables that could be used somewhere called `plist_hashes` to clarify that the only stored block plists are hashes.

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
...
+      label = plist[:Label] ||= plist_name
+      [label, plist.to_plist]
```
MikeMcQuaid: Consider building a hash here instead of turning into a hash later. `plists[plist[:Label]] = plist.to_plist`.

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
...
+    end
+    plists.empty? ? nil : plists.to_h
```
MikeMcQuaid: Before the `map` you can do `return if self.class.plist.empty?`

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
...
+  end
+
+  # @private
+  def self.plist(&block)
```
MikeMcQuaid: Maybe `plist_hashes` would be a better name here.

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
...
+    @plists ||= []
```
MikeMcQuaid: Maybe `@plist_hashes` would be a better name here.

Library/Homebrew/formula.rb
```diff
@@ -922,6 +897,33 @@ def plist_startup
...
+    block_given? ? @plists.push(block) : @plists
```
MikeMcQuaid:
```ruby
return @plists unless block_given?
@plists << block
```

Library/Homebrew/formula_installer.rb
```diff
@@ -863,12 +864,17 @@ def link(keg)
   end
 
   def install_plist
-    return unless formula.plist
+    return unless plist = formula.plist
```
MikeMcQuaid: Can you make this:
```ruby
plist = formula.plist
return unless plist
```

as the assignment in the `unless` reads a little weird.

Library/Homebrew/formula_installer.rb
```diff
@@ -863,12 +864,17 @@ def link(keg)
...
 
-    formula.plist_path.atomic_write(formula.plist)
-    formula.plist_path.chmod 0644
-    log = formula.var/"log"
-    log.mkpath if formula.plist.include? log.to_s
+    plist = { formula.plist_name => plist } if plist.is_a? String
```
MikeMcQuaid: This indicates the Formula API is broken, I'm afraid. `brew services` (and perhaps other tools) assume the output is always a string. Need to figure out a way to make `formula.plist` not return a non-string here; perhaps with a `plists` method instead which can return multiple values through a Hash.

Library/Homebrew/formula_installer.rb
```diff
@@ -863,12 +864,17 @@ def link(keg)
...
+
+    plist.each do |label, xml|
+      plist_path = formula.plist_path.sub(formula.plist_path.stem, label)
```
MikeMcQuaid: Can you explain this a bit? Not totally clear what it's doing and why. A code comment might be good if it stays the same.

Library/Homebrew/formula.rb
danielbayley: That breaks everything… unless we then have _another_ method, just to be able to rename this method… is that desirable?
~~~ rb
# @private
def plist(&block)
  self.plist_hashes(&block)
end
~~~

Library/Homebrew/formula_installer.rb
danielbayley: Because the `plist_path`(`.stem`) needs to match the `label`, and the unique `label` is passed in here… so it needs to be substituted here. Directly reassigning `formula.plist_path =` here throws an error, and I think the only way to modify `formula.plist_path` before this point is to redefine `def plist_path` inside the `def plist` instance method (something which was frowned on before) Same goes for `plist_name` also…

If any of this begs the question "why?"… it's likely because I had to impliment some obscure workaround just to get it to work within the limiting constraints of this OOP architecture. It feels more like trying to play whack-a-mole wearing a straightjacket in the dark than writing a simple feature! 😅

Library/Homebrew/formula_installer.rb
danielbayley: Well if `formula.plist` _must_ be a string before this point then I don't see how we can even have multiple `plist do` blocks defined in the first place… because the `install_plist` method only runs once; forcing the need to loop over an array/hash of strings inside this method. Are you saying that we would have to specify `plists do` blocks inside formulae instead?

There are no other references to `formula.plist` outside of this `install_plist` method… Does `brew services` not just operate on the files this method writes?

MikeMcQuaid: > Well if formula.plist must be a string before this point then I don't see how we can even have multiple plist do blocks defined in the first place

You can have multiple defined and introduce a new API to return them `formula.plists` and just have `formula.plist` always return the first one instead for external consumers. `plist do` can remain as-is but the consumption `formula.plist` API needs to not change.

> Does `brew services` not just operate on the files this method writes… and if not, why not?

No, it doesn't, you can see/test the code to see why. Given it does these I'm not totally confident other tools won't either and maintaining some degree of backwards compatibility is important. I could be convinced that nothing else except `brew services` uses this but then it would need to have a PR that lands pretty much simultaneously to this stuff which handles both the old code (until this goes out in a tagged stable release) and the new code (for those running Homebrew/brew's master branch).

Library/Homebrew/formula_installer.rb
MikeMcQuaid: > Because the `plist_path`(`.stem`) needs to match the `label`, and the unique `label` is passed in here

So the `plist_path` is basically returning the wrong information because it needs to match the label?

Library/Homebrew/formula.rb
MikeMcQuaid: As discussed below there's already API breakage so this may make sense to rename as part of that change.

zbeekman:
I think this is still a WIP. I can't edit the tags, since I'm not part of homebrew/brew.

MikeMcQuaid:
@zbeekman It's a WIP but it's right for it to get automatically closed out if there's no activity; that's kinda the point :wink:

zbeekman:
The reason I'm interested is that I was hoping to build off this for the plists that get utilized by unattended upgrade/unattended prefetch... but I should probably just forge ahead and converge later, if this ever gets sorted.

It would be a shame for all this hard work to go to waste though. Maybe I'll see if I can't hack at it a bit myself.

MikeMcQuaid:
> Maybe I'll see if I can't hack at it a bit myself.

Yup please do 👍

danielbayley:
@zbeekman Have you done anything with this? I've been meaning to come back around to it… but been busy with other things, and it went from one simple PR into 3 more complicated ones 🙄 I will get back to it, but if you've made any progress let me know 👍

zbeekman:
Not yet, been busy with the day job.

torarnv:
Wouldn't this whole thing be easier to solve by some kind of sub-formula concept? Why does a Formula require a url, etc? If we allowed this
```rb

class ServiceA < Formula
  def plist; <<~EOS
     ...
  EOS
  end
end

class ServiceB < Formula
  def plist; <<~EOS
     ...
  EOS
  end
end

class Foo < Formula
   url ...
end

```

The "main" formula could install the services separately without having to download or install anything.

torarnv:
Lack of meta-formulas (without url) also means you can't add a formula that depends on another and then just does a post-install/install step.

MikeMcQuaid:
> Wouldn't this whole thing be easier to solve by some kind of sub-formula concept?

No.

torarnv:
> > Wouldn't this whole thing be easier to solve by some kind of sub-formula concept?
> 
> No.

Ok 😃

MikeMcQuaid:
Closing this out as there's been no commits for several months.

PR #5696, opened 2019-02-08
elf.rb: avoid corrupted elf files

iMichka:
Some elf files like unittest files or memory dumps may not be completely
readable by readelf.

Readelf will fail after the following message:
readelf: Warning: possibly corrupt ELF header - it has a non-zero program header offset, but no program headers

This patches avoid these files when there is a non zero offset but no
program headers

-----

### DIFF
--- a/Library/Homebrew/os/linux/elf.rb
+++ b/Library/Homebrew/os/linux/elf.rb
@@ -133,8 +133,10 @@ module ELFShim
       soname = nil
       needed = []
       command = ["readelf", "-d", path.expand_path.to_s]
-      lines = Utils.safe_popen_read(*command).split("\n")
+      lines = Utils.popen_read(*command, err: :out).split("\n")
       lines.each do |s|
+        next if s.start_with?("readelf: Warning: possibly corrupt ELF header")
+
         filename = s[/\[(.*)\]/, 1]
         next if filename.nil?
 


iMichka:
- I do not know if parsing the output of `readelf -h` is the best solution. What if that output changes too much (though I would say that readelf is pretty stable now).
- I did not add tests. If I am not mistaken there are no tests for this particular elf.rb file, but we could add some tests. The corrupted file is quite small so it could be added for testing.

Library/Homebrew/os/linux/elf.rb
```diff
@@ -78,6 +79,24 @@ def dynamic_elf?
...
+    nbr_program_headers = 0
+    Utils.popen_read("readelf", "-h", to_path).each_line do |line|
+      if line["Start of program headers:"]
+        offset = line.scan(/\d+/)[0].to_i
+      end
+      if line["Number of program headers:"]
+        nbr_program_headers = line.scan(/\d+/)[0].to_i
+      end
+    end
+    nbr_program_headers.zero? && !offset.zero?
```
jonchang: Would it be better to just check stderr for `Warning: possibly corrupt ELF header - it has a non-zero program header offset, but no program headers` since it seems like this code is duplicating the check `readelf` is already doing?

sjackman: I like this solution. That's how I handled `patchelf` failing in some cases.
https://github.com/Homebrew/brew/blob/6ae600b5e67008e81ef2e3d595ebfb986912e0a7/Library/Homebrew/extend/os/linux/keg_relocate.rb#L21-L25

iMichka:
The output is on stderr, so I had to use popen3. This can be reviewed again.

Library/Homebrew/os/linux/elf.rb
```diff
@@ -78,6 +79,18 @@ def dynamic_elf?
...
 
+  def corrupted_elf?
+    # elf files that throw the following message when parsed with readelf:
+    # Warning: possibly corrupt ELF header
+    # it has a non-zero program header offset, but no program headers
+    # Example: the early-section-headers.so unittest file from the rust package
+    message = ""
+    Open3.popen3("readelf", "-d", to_path) do |_, _, stderr|
+      message = stderr.read
+    end
```
sjackman: I suggest instead…
```ruby
    message = Utils.popen_read("readelf", "-d", to_path, err: :out).strip
```

Library/Homebrew/os/linux/elf.rb
```diff
@@ -78,6 +79,18 @@ def dynamic_elf?
...
+    message.include?("readelf: Warning: possibly corrupt ELF header")
```
sjackman:
```suggestion
    message.start_with?("readelf: Warning: possibly corrupt ELF header")
```
or `==` if that's the complete message.

Library/Homebrew/os/linux/elf.rb
jonchang: Neat trick!

Library/Homebrew/os/linux/elf.rb
iMichka: The message is a little bit longer. I will go for start_with.

Library/Homebrew/os/linux/elf.rb
```diff
@@ -78,6 +79,16 @@ def dynamic_elf?
     end
   end
 
+  def corrupted_elf?
+    # elf files that throw the following message when parsed with readelf:
+    # Warning: possibly corrupt ELF header
+    # it has a non-zero program header offset, but no program headers
+    # Example: the early-section-headers.so unittest file from the rust package
+    Utils.popen_read("readelf", "-d", to_path, err: :out)
+         .strip
```
sjackman: The `.strip` here is probably not needed, unless there's white space to the left of the string.

iMichka:
I removed the .strip call. Good to be reviewed again.

sjackman:
Which `readelf` call was failing that lead to this PR? To avoid executing `readelf` twice, can we move this test to the original call of `readelf`? Or was it `patchelf` that was failing?

iMichka:
There is more explanation for this on top of this PR and in the commit message. I can get you the exact error message if you want.

You suggest we should rescue the failing readelf call and look at the error message for speed reasons? I had not the feeling that the pouring was way slower for the `rust` formula, though that may be subjective.

sjackman:
> There is more explanation for this on top of this PR and in the commit message. I can get you the exact error message if you want.

> You suggest we should rescue the failing readelf call and look at the error message for speed reasons? I had not the feeling that the pouring was way slower for the `rust` formula, though that may be subjective.

If there's one existing call to `readelf` that was failing, I'd prefer to rescue that failing `readelf` call. It'd be tidier, and as a bonus, perhaps a bit faster. If more than one `readelf` call needs to be fixed up, then this PR is tidier as is.

iMichka:
```
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/postinstall.rb (Formulary::FromPathLoader): loading /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/rust.rb
Error: Failure while executing; `readelf -d /home/linuxbrew/.linuxbrew/Cellar/rust/1.32.0/share/rust/rust_src/tools/lldb/unittests/ObjectFile/ELF/Inputs/early-section-headers.so` exited with 1. Here's the output:
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/utils/popen.rb:10:in `safe_popen_read'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/os/linux/elf.rb:136:in `needed_libraries_using_readelf'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/os/linux/elf.rb:111:in `needed_libraries'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/os/linux/elf.rb:87:in `initialize'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/os/linux/elf.rb:152:in `new'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/os/linux/elf.rb:152:in `metadata'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/os/linux/elf.rb:160:in `dynamically_linked_libraries'
/home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/linkage_checker.rb:98:in `block in check_dylibs'
```

iMichka:
I need to test if rescuing the error there will help, or if there are other calls that will need a rescue.

iMichka:
I'll have a look if I can add a test in a second PR.

Issue #5692, opened 2019-02-08
brew fails to open web pages on Linux, please whitelist DISPLAY variable


mtmiller:
**Please note that we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

- [x] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew update` and can still reproduce the problem?
- [x] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [x] ran `brew config` and `brew doctor` and included their output with your issue?

# What you were trying to do (and why)

Any `brew` command that passes a URL to open a tab in the web browser fails on Linux. For example, `brew home` or `brew info --github`.

# What happened (include command output)

The correct web browser command runs, but fails with an error that the display cannot be opened and the tab is not opened.

<details>
  <summary>Command output</summary>
  <pre>
  
  mike@host:~$ brew home
  
  (google-chrome:32228): Gtk-WARNING **: 20:57:36.156: cannot open display: 
  mike@host:~$ brew info --github gcc
  
  (google-chrome:32447): Gtk-WARNING **: 20:57:40.127: cannot open display: 
  
  </pre>
</details>

# What you expected to happen

The web browser should open a new tab with the given URL.

The reason is clearly that the `DISPLAY` environment variable is removed in the sanitized environment that `brew` runs itself in. If I locally modify `brew` to include `DISPLAY` in the list of whitelisted environment variables, then the web page is opened as expected.

# Step-by-step reproduction instructions (by running `brew` commands)

1. Install Linuxbrew on a GNU/Linux system, accept the default installation options.
2. Run any command that tries to open a web page
```
brew home
brew home gcc
brew info --github gcc
brew search --debian gcc
```

## Output of `brew config` and `brew doctor`

```
mike@host:~$ brew config
[...]

PR #5706, opened 2019-02-11
Set DISPLAY environment variable for browser on Linux

mtmiller:

-----

This change stashes the user's original `DISPLAY` into `HOMEBREW_DISPLAY` and restores it when calling out to the browser. See #5692.

I suspect this should also be documented in docs and/or manpages, right? Should I also add a test case, something like `HOMEBREW_BROWSER='echo $DISPLAY'? Any other ways I can make this better?

Cc @sjackman for review/advice

### DIFF
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -388,9 +388,7 @@ def exec_browser(*args)
   browser ||= OS::PATH_OPEN if defined?(OS::PATH_OPEN)
   return unless browser
 
-  if ENV["HOMEBREW_DISPLAY"]
-    ENV["DISPLAY"] = ENV["HOMEBREW_DISPLAY"]
-  end
+  ENV["DISPLAY"] = ENV["HOMEBREW_DISPLAY"]
 
   safe_exec(browser, *args)
 end
--- a/Library/Homebrew/utils.rb
+++ b/Library/Homebrew/utils.rb
@@ -388,6 +388,10 @@ def exec_browser(*args)
   browser ||= OS::PATH_OPEN if defined?(OS::PATH_OPEN)
   return unless browser
 
+  if ENV["HOMEBREW_DISPLAY"]
+    ENV["DISPLAY"] = ENV["HOMEBREW_DISPLAY"]
+  end
+
   safe_exec(browser, *args)
 end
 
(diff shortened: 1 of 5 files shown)


Library/Homebrew/utils.rb
```diff
@@ -388,6 +388,10 @@ def exec_browser(*args)
   browser ||= OS::PATH_OPEN if defined?(OS::PATH_OPEN)
   return unless browser
 
+  if ENV["HOMEBREW_DISPLAY"]
```
MikeMcQuaid: Can skip the `if`; won't do any harm if it's not set.

MikeMcQuaid:
> I suspect this should also be documented in docs and/or manpages, right? 

Yes, please! Under "environment variables" would be ideal.

> Should I also add a test case, something like `HOMEBREW_BROWSER='echo $DISPLAY'? Any other ways I can make this better?

Don't need a test for this, thanks!

Library/Homebrew/utils.rb
sjackman: My personal preference would be…
```ruby
ENV["DISPLAY"] = ENV["HOMEBREW_DISPLAY"] if ENV["HOMEBREW_DISPLAY"]
```
The error message from a browser may be different if the `DISPLAY` environment variable is not set versus set to the empty string.
@mtmiller Could you please test…
```sh
env -u DISPLAY firefox
env DISPLAY= firefox
```
and see what the error message is?

mtmiller: Thanks for the feedback. I eliminated the `if` entirely. I'm not much of a rubyist, but I found that `ENV` DTRT, assigning `nil` deletes from the environment.

mtmiller: > Could you please test…
> and see what the error message is?

I did, the error message does differ by browser, but the effect is the same, no tab opened.

Firefox does differentiate:

```
$ env -u DISPLAY firefox
Error: no DISPLAY environment variable specified
$ env DISPLAY= firefox
Unable to init server: Broadway display type not supported: 
Error: cannot open display: 
```

Chrome doesn't care whether it's unset or empty:

```
$ env -u DISPLAY google-chrome

(google-chrome:32416): Gtk-WARNING **: 12:48:05.569: cannot open display: 
$ env DISPLAY= google-chrome

(google-chrome:32480): Gtk-WARNING **: 12:48:09.083: cannot open display: 
```

mtmiller:
> Yes, please! Under "environment variables" would be ideal.

Done.

mtmiller:
Thanks for the guidance and feedback to get this done.

This is my first contribution to Homebrew, and I'm not a rubyist, so it took a little while to get past some bundle/gem setup problems to generate the man pages, but got it working in the end.

Happy to squash commits, rewrite commit messages, or add anything else you want to see in this PR.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @mtmiller!

PR #5718, opened 2019-02-13
Add `brew update -f` option and improve help string format

mtmiller:

-----

This change adds `brew update -f` as an alias for `brew update --force` and improves the help string format to include both long and short options. This format improves readability by making the help output consistent with ruby-based commands that use `OptionParser`.

Also documents existing common options that weren't listed in the `brew update` help string but should have been (`--verbose`, `--debug`, and `--help`).

I did not document the options `--preinstall` and `--simulate-from-current-branch` because they seem internal or developer-use only and I'm honestly not sure how to describe them.

### DIFF
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -2,8 +2,11 @@
 #:
 #:  Fetch the newest version of Homebrew and all formulae from GitHub using `git`(1) and perform any necessary migrations.
 #:
-#:       --merge                         `git merge` is used to include updates (rather than `git rebase`).
-#:       --force                         Always do a slower, full update check (even if unnecessary).
+#:          --merge                      `git merge` is used to include updates (rather than `git rebase`).
+#:      -f, --force                      Always do a slower, full update check (even if unnecessary).
+#:      -v, --verbose                    Print the directories checked and `git` operations performed.
+#:      -d, --debug                      Display a trace of all shell commands as they are executed.
+#:      -h, --help                       Show this message.
 
 # Don't need shellcheck to follow this `source`.
 # shellcheck disable=SC1090
--- a/Library/Homebrew/cmd/update.sh
+++ b/Library/Homebrew/cmd/update.sh
@@ -306,6 +306,7 @@ homebrew-update() {
       -*)
         [[ "$option" = *v* ]] && HOMEBREW_VERBOSE=1
         [[ "$option" = *d* ]] && HOMEBREW_DEBUG=1
+        [[ "$option" = *f* ]] && HOMEBREW_UPDATE_FORCE=1
         ;;
       *)
         odie <<EOS
--- a/Library/Homebrew/dev-cmd/man.rb
+++ b/Library/Homebrew/dev-cmd/man.rb
@@ -195,7 +195,9 @@ module Homebrew
       line = line.slice(4..-1)
       next unless line
 
-      lines << line.gsub(/^ +(-+[a-z-]+) */, "* `\\1`:\n  ")
+      # Format one option or a comma-separated pair of short and long options.
+      lines << line.gsub(/^ +(-+[a-z-]+), (-+[a-z-]+) +/, "* `\\1`, `\\2`:\n  ")
+                   .gsub(/^ +(-+[a-z-]+) +/, "* `\\1`:\n  ")
     end
     lines
   end
--- a/Library/Homebrew/dev-cmd/man.rb
+++ b/Library/Homebrew/dev-cmd/man.rb
@@ -195,6 +195,9 @@ module Homebrew
       line = line.slice(4..-1)
       next unless line
 
+      # Omit the common global_options documented separately in the man page.
+      next if line =~ /--(debug|force|help|quiet|verbose) /
+
       # Format one option or a comma-separated pair of short and long options.
       lines << line.gsub(/^ +(-+[a-z-]+), (-+[a-z-]+) +/, "* `\\1`, `\\2`:\n  ")
                    .gsub(/^ +(-+[a-z-]+) +/, "* `\\1`:\n  ")
--- a/docs/Manpage.md
+++ b/docs/Manpage.md
@@ -540,9 +540,6 @@ Fetch the newest version of Homebrew and all formulae from GitHub using `git`(1)
 * `--merge`:
   `git merge` is used to include updates (rather than `git rebase`).
 
-* `--force`:
-  Always do a slower, full update check (even if unnecessary).
-
 ### `update-reset` [*`repository`*]
 Fetches and resets Homebrew and all tap repositories (or any specified `repository`) using `git`(1) to their latest `origin/master`. Note this will destroy all your uncommitted or committed changes.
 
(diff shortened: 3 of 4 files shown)


Library/Homebrew/cmd/update.sh
```diff
@@ -2,8 +2,11 @@
 #:
 #:  Fetch the newest version of Homebrew and all formulae from GitHub using `git`(1) and perform any necessary migrations.
 #:
-#:       --merge                         `git merge` is used to include updates (rather than `git rebase`).
-#:       --force                         Always do a slower, full update check (even if unnecessary).
+#:          --merge                      `git merge` is used to include updates (rather than `git rebase`).
+#:      -f, --force                      Always do a slower, full update check (even if unnecessary).
+#:      -v, --verbose                    Print the directories checked and `git` operations performed.
+#:      -d, --debug                      Display any debugging information.
+#:      -h, --help                       Show this message.
```
MikeMcQuaid: This can be skipped as we're not documenting in on other commands.

Library/Homebrew/cmd/update.sh
```diff
@@ -2,8 +2,11 @@
 #:
 #:  Fetch the newest version of Homebrew and all formulae from GitHub using `git`(1) and perform any necessary migrations.
 #:
-#:       --merge                         `git merge` is used to include updates (rather than `git rebase`).
-#:       --force                         Always do a slower, full update check (even if unnecessary).
+#:          --merge                      `git merge` is used to include updates (rather than `git rebase`).
```
MikeMcQuaid: Inline this with the `-f`.

Library/Homebrew/cmd/update.sh
```diff
@@ -2,8 +2,11 @@
...
+#:      -f, --force                      Always do a slower, full update check (even if unnecessary).
+#:      -v, --verbose                    Print the directories checked and `git` operations performed.
+#:      -d, --debug                      Display any debugging information.
```
MikeMcQuaid: Would be good to elaborate on exactly what this is doing. I believe it's going to output every single command executed in Bash.

Library/Homebrew/dev-cmd/man.rb
```diff
@@ -192,7 +192,8 @@ def cmd_comment_manpage_lines(cmd_path)
                  .each do |line|
       line = line.slice(4..-1)
       next unless line
-      lines << line.gsub(/^ +(-+[a-z-]+) */, "* `\\1`:\n  ")
+      lines << line.gsub(/^ +(-+[a-z-]+), (-+[a-z-]+) +/, "* `\\1`, `\\2`:\n  ")
```
MikeMcQuaid: What's this change for?

Library/Homebrew/cmd/update.sh
mtmiller: This is documented in the `Homebrew::CLI::Parser.global_options`, so it does get listed in the `brew help` output for every command. You're right though it does not get pulled in when generating the man page markdown. Is it more important to exclude this for consistency in the man page or include it for consistency at the shell command line?

Library/Homebrew/cmd/update.sh
mtmiller: Really? This is an example of what I am trying to match:

```
$ brew help info | tail -n4
        --installed                  Get information on all installed formulae.
    -v, --verbose                    See more verbose analytics data.
    -d, --debug                      Display any debugging information.
    -h, --help                       Show this message.
```

Library/Homebrew/cmd/update.sh
mtmiller: Sure, can elaborate. I just borrowed that description from `Homebrew::CLI::Parser.global_options`.

Library/Homebrew/dev-cmd/man.rb
mtmiller: This function is only used in parsing the help string comment in shell commands. The current regexp grabs the first word as the option and the rest of the line is description. Without this change, a line like this

    #:    -v, --verbose                           Be verbose

would end up formatted like this

    * `-v`:
      --verbose                           Be verbose

instead of what we want

    * `-v`, `--verbose`:
      Be verbose

So this change is necessary if we want to allow the help string to list both short and long options, makes the raw help as well as the markdown consistent with ruby `OptionParser`-based commands.

MikeMcQuaid: Gotcha, thanks 👍

Library/Homebrew/cmd/update.sh
MikeMcQuaid: Thanks!

Library/Homebrew/cmd/update.sh
MikeMcQuaid: Ok, 🆒, I get this now from below!

Library/Homebrew/cmd/update.sh
MikeMcQuaid: Ooh, nice spot. Ideally the manpage and help output would be consistent regardless for all commands (i.e. display `--help` in both places or neither). I'm easy on either approach. Could you add that in this PR or a separate one? I don't mind which. Thanks!

MikeMcQuaid:
@mtmiller Any news here? Thanks!

mtmiller:
Apologies for the delay, getting back to this now!

Library/Homebrew/cmd/update.sh
mtmiller: @MikeMcQuaid It looks like man.rb was designed with the intent of documenting [global options](https://docs.brew.sh/Manpage#global-options) separately from "command-specific options", so that's why `--help` does not appear in each command usage. I don't really want to mess with that organization. But all options do and should appear at the command line.

So how about a change that just filters out `--help` with a regex from shell command when building the man page?

MikeMcQuaid: > So how about a change that just filters out `--help` with a regex from shell command when building the man page?

Sounds good if that's the only inconsistency 👍

Library/Homebrew/cmd/update.sh
mtmiller: This is now better described in 8b85625.

Library/Homebrew/cmd/update.sh
mtmiller: In 88dea00 I added a regex to filter out all of the "global options" from shell command usage blocks. This is now consistent with the behavior of ruby cmds. All options are printed on the command line with `brew help update`, but the man page and html only lists the `--merge` option since all other options are in the "global options" set, including `--force`. Hope this makes sense.

mtmiller:
I see there's a merge conflict now, I'm happy to rebase and rewrite these commits to make this read better anyway, just let me know.

MikeMcQuaid:
> I see there's a merge conflict now, I'm happy to rebase and rewrite these commits to make this read better anyway, just let me know.

@mtmiller Yep, rebase would be great and code's looking good enough I'm happy to merge when this is ✅.

mtmiller:
@MikeMcQuaid This should be good now, thanks for all your patience and feedback.

MikeMcQuaid:
Great work here @mtmiller! Thanks again for your contribution!

PR #5737, opened 2019-02-15
Add info on env variables for Formula Cookbook

davidjb:
This explains the levels of environment variable filtering present in Homebrew, explaining how and why variables without a `HOMEBREW_` prefix or those which contain `TOKEN` etc are made unavailable to a Formula. 

Previously, this behaviour was unclear (eg `HOMEBREW_FOO` works but `HOMEBREW_TOKEN` didn't) so I've added this documentation with references explaining the process.   There may be other env filtering at work, but these are the two levels that were influencing my attempts to figure out why a third party formula written some time ago was no longer able to see its environment variable.

- [ ] ~Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb).~
- [ ] ~Have you successfully run `brew style` with your changes locally?~
- [ ] ~Have you successfully run `brew tests` with your changes locally?~

-----

### DIFF
--- a/docs/Formula-Cookbook.md
+++ b/docs/Formula-Cookbook.md
@@ -696,6 +696,16 @@ Homebrew provides two formula DSL methods for launchd plist files:
 * [`plist_name`](https://www.rubydoc.info/github/Homebrew/brew/master/Formula#plist_name-instance_method) will return e.g. `homebrew.mxcl.<formula>`
 * [`plist_path`](https://www.rubydoc.info/github/Homebrew/brew/master/Formula#plist_path-instance_method) will return e.g. `/usr/local/Cellar/foo/0.1/homebrew.mxcl.foo.plist`
 
+### Using environment variables
+
+Homebrew has multiple levels of environment variable filtering which affects variables available to formulae.
+
+Firstly, the overall environment in which Homebrew runs is filtered to avoid environment contamination breaking from-source builds ([ref](https://github.com/Homebrew/brew/issues/932)).  In particular, this process filters all but the given whitelisted variables, but allows environment variables prefixed with `HOMEBREW_`. The specific implementation can be seen in the [`brew`](https://github.com/Homebrew/brew/blob/master/bin/brew) script.
+
+The second level of filtering removes sensitive environment variables (such as credentials like keys, passwords or tokens) to avoid malicious subprocesses obtaining them ([ref](https://github.com/Homebrew/brew/pull/2524)).  This has the effect of preventing any such variables from reaching a formula's Ruby code as they are filtered before it is called.  The specific implementation can be seen in the [`clear_sensitive_environment` method](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/ENV.rb).
+
+In summary, environment variables used by a formula need to conform to these filtering rules in order to be available.
+
 ## Updating formulae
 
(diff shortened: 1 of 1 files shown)


docs/Formula-Cookbook.md
```diff
@@ -696,6 +696,16 @@ Homebrew provides two formula DSL methods for launchd plist files:
...
+
+Firstly, the overall environment in which Homebrew runs is [filtered](https://github.com/Homebrew/brew/blob/master/bin/brew#L54-L96) to avoid environment contamination breaking from-source builds ([ref](https://github.com/Homebrew/brew/issues/932)).  In particular, this process filters all but the given whitelisted variables relating to the shell, terminal, OS, proxies and so on, but allows anything prefixed with `HOMEBREW_` or related to CI and Travis through.
+
+The second level of filtering [removes "sensitive" environment variables](https://github.com/Homebrew/brew/blob/4897ec02d0f00c37a21ddaa0bebff6b83b27d027/Library/Homebrew/extend/ENV.rb#L32) to avoid malicious subprocesses obtaining them ([ref](https://github.com/Homebrew/brew/pull/2524)).  This has the effect of preventing any such variables from reaching a formula's Ruby code as they are filtered before it is reached.  This filtering remove variables that match the given case-insensitive pattern, which is to say that environment variables containing the names `COOKIE`, `KEY`, `TOKEN` or `PASSWORD` will be removed.
```
MikeMcQuaid: Again, would you mind removing:
```
`COOKIE`, `KEY`, `TOKEN` or `PASSWORD`
```
and pointing to the code instead?

Similarly, you've linked to a specific commit here which means the line number will remain correct but unfortunately it'll not be up to date. Might want to reference a function name or similar instead?

docs/Formula-Cookbook.md
```diff
@@ -696,6 +696,16 @@ Homebrew provides two formula DSL methods for launchd plist files:
 * [`plist_name`](https://www.rubydoc.info/github/Homebrew/brew/master/Formula#plist_name-instance_method) will return e.g. `homebrew.mxcl.<formula>`
 * [`plist_path`](https://www.rubydoc.info/github/Homebrew/brew/master/Formula#plist_path-instance_method) will return e.g. `/usr/local/Cellar/foo/0.1/homebrew.mxcl.foo.plist`
 
+### Using environment variables
+
+Homebrew has multiple levels of environment variable filtering which impact which variables are available to formulae.
+
+Firstly, the overall environment in which Homebrew runs is [filtered](https://github.com/Homebrew/brew/blob/master/bin/brew#L54-L96) to avoid environment contamination breaking from-source builds ([ref](https://github.com/Homebrew/brew/issues/932)).  In particular, this process filters all but the given whitelisted variables relating to the shell, terminal, OS, proxies and so on, but allows anything prefixed with `HOMEBREW_` or related to CI and Travis through.
```
MikeMcQuaid: Can you remove `relating to the shell, terminal, OS, proxies and so on,` and `or related to CI and Travis through`; this is implied by looking at the code and if we update the code we'll likely forget to update these docs. 

Additionally, it may be worth noting the `#L54-L96` will break if the number of lines in that file changes. It may be worth omitting the lines count as we want to link to something up-to-date.

docs/Formula-Cookbook.md
davidjb: Yep, for sure.  I'll link to the file itself as a whole and reference the function name.

docs/Formula-Cookbook.md
davidjb: For sure, I'll speak more generally and link to the general file itself so people can always get the right info.

davidjb:
@MikeMcQuaid PR updated with the changes, plus some improvements in readability for the rest.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @davidjb!

Issue #5736, opened 2019-02-14
--quiet conflicts with HOMEBREW_VERBOSE for brew-outdated


zmwangx:
I have `HOMEBREW_VERBOSE` set for many years, and it never conflicted with the `--quiet` flag on the command line. When `--quiet` is specified, it simply overrides `HOMEBREW_VERBOSE`, as is the convention for handling option vs environment variable in most *nix tools.

However, a few weeks ago, 867b155479be4a50e02580e05df770aec84438be (CC author @MikeMcQuaid) broke this. Now, using `--quiet` with `brew outdated` when `HOMEBREW_VERBOSE` is set simply results in an error saying `--quiet` and `--verbose` are mutually exclusive. (This broke my auto update script.) The problem apparently stems from making no distinction between `HOMEBREW_VERBOSE` and the `--verbose` option on the command line.

The point is: *command line option should override a conflicting env var when necessary, instead of being treated as a conflicting equal.*

# What you were trying to do (and why)

```console
$ HOMEBREW_VERBOSE=1 brew outdated --quiet
```

Why? Because I have `HOMEBREW_VERBOSE` set globally, and only override it on the command line as I see fit.

# What happened

```console
$ HOMEBREW_VERBOSE=1 brew outdated --quiet
Error: Options --quiet and --verbose are mutually exclusive.
```

# What you expected to happen

No error.

P.S. The help text doesn't help.

```console
$ brew outdated -h
Usage: brew outdated [options]

Show formulae that have an updated version available.

By default, version information is displayed in interactive shells, and
suppressed otherwise.

    -q, --quiet                      List only the names of outdated brews
                                     (takes precedence over --verbose).
```

Note *takes precedence over*.

RandomDSdevel:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;This was fixed in #5738.

zmwangx:
Pretty sure it's not.

MikeMcQuaid:
@zmwangx Will reopen when you confirm/paste current error, thanks.

zmwangx:
Well, just the same?

Unsetting all `HOMEBREW_*` env vars to make sure:

```console
$ brew config
HOMEBREW_VERSION: 2.0.2-78-g59aac33-dirty
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 59aac3347acf83e2b7950b8cb20df2b92986f4ec
Last commit: 24 minutes ago
[...]

PR #5738, opened 2019-02-15
Prioritize CLI arguments over env vars when they conflict

BenMusch:
- [x] Have you successfully run `brew style` with your changes locally? **(Note: Brew style fails, but the failure is there to keep it consistent with other tests in the file. Happy to change it if y'all prefer)**

-----

Addresses an issue mentioned here: https://github.com/Homebrew/brew/issues/5736

(Caveat: The issue linked also points out a disconnect between the help text of the command and its actual behavior, which should probably be fixed separately)

I agree with that issue that CLI flags should take priority over environment variables when they conflict, and it's behavior I'm used to in other applications. This PR does that by selecting the CLI-passed argument when there are conflicting switches and all but one of the switches are specified via an env var.

Not 100% tied to this being the correct behavior -- just opening this in case y'all agree that it is

Also, I think there's also an opportunity for a richer API to resolve conflicts here, something along the lines of:

```ruby
def switch(*names, description: nil, env: nil, required_for: nil, depends_on: nil, env_priority: 1, cli_priority: 2)
```

Then, you could specify the priority of an argument so that it would always take priority over another when they conflict, etc. But that seems like it may be more code than is worth maintaining

### DIFF
--- a/Library/Homebrew/cli_parser.rb
+++ b/Library/Homebrew/cli_parser.rb
@@ -27,6 +27,7 @@ module Homebrew
         Homebrew.args.instance_eval { undef tap }
         @constraints = []
         @conflicts = []
+        @switch_sources = {}
         @processed_options = []
         @desc_line_length = 43
         @hide_from_man_page = false
@@ -58,7 +59,7 @@ module Homebrew
           set_constraints(name, required_for: required_for, depends_on: depends_on)
         end
 
-        enable_switch(*names) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
+        enable_switch(*names, source: :env_var) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
       end
       alias switch_option switch
 
@@ -172,9 +173,16 @@ module Homebrew
 
       private
 
-      def enable_switch(*names)
+      def enable_switch(*names, source: :cli_arg)
         names.each do |name|
+          @switch_sources[option_to_name(name)] = source
           Homebrew.args["#{option_to_name(name)}?"] = true
+        end
+      end
+
+      def disable_switch(*names)
+        names.each do |name|
+          Homebrew.args.delete_field("#{option_to_name(name)}?")
         end
       end
 
@@ -225,7 +233,17 @@ module Homebrew
 
           next if violations.count < 2
 
-          raise OptionConflictError, violations.map(&method(:name_to_option))
+          env_var_options = violations.select do |option|
+            @switch_sources[option_to_name(option)] == :env_var
+          end
+
+          if violations.count - env_var_options.count == 1
+            env_var_options.each do |option|
+              disable_switch option
+            end
+          else
+            raise OptionConflictError, violations.map(&method(:name_to_option))
+          end
         end
       end
 
--- a/Library/Homebrew/cli_parser.rb
+++ b/Library/Homebrew/cli_parser.rb
@@ -59,7 +59,7 @@ module Homebrew
           set_constraints(name, required_for: required_for, depends_on: depends_on)
         end
 
-        enable_switch(*names, source: :env_var) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
+        enable_switch(*names, from: :env) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
       end
       alias switch_option switch
 
@@ -173,9 +173,9 @@ module Homebrew
 
       private
 
-      def enable_switch(*names, source: :cli_arg)
+      def enable_switch(*names, from: :arg)
         names.each do |name|
-          @switch_sources[option_to_name(name)] = source
+          @switch_sources[option_to_name(name)] = from
           Homebrew.args["#{option_to_name(name)}?"] = true
         end
       end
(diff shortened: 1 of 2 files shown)


Library/Homebrew/cli_parser.rb
```diff
@@ -58,7 +59,7 @@ def switch(*names, description: nil, env: nil, required_for: nil, depends_on: ni
           set_constraints(name, required_for: required_for, depends_on: depends_on)
         end
 
-        enable_switch(*names) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
+        enable_switch(*names, source: :env_var) if !env.nil? && !ENV["HOMEBREW_#{env.to_s.upcase}"].nil?
```
BenMusch: would prefer to extract these env/cli values into constants somewhere, but not sure what the style for that is, or even if a simple boolean would be better here

MikeMcQuaid:
> But that seems like it may be more code than is worth maintaining

Agreed.



> I agree with that issue that CLI flags should take priority over environment variables when they conflict, and it's behavior I'm used to in other applications.

Also agreed 👍

Library/Homebrew/cli_parser.rb
BenMusch: I think this phrasing is clearer:

Would it might be better to extract `:env` and `:args` to constants, to avoid having magic values floating around, or even just to change it to a boolean `is_env_var?` flag?

MikeMcQuaid: Sorry, still being stupid here! Can you give some examples of what values you think should maybe be constants?

Library/Homebrew/cli_parser.rb
```diff
@@ -172,12 +173,19 @@ def hide_from_man_page!
 
       private
 
-      def enable_switch(*names)
+      def enable_switch(*names, source: :cli_arg)
```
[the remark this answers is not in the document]
MikeMcQuaid: @BenMusch Thoughts on making this a non-default parameter?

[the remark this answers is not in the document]
BenMusch: Fine with me, will push a change in an hour or two

Library/Homebrew/cli_parser.rb
BenMusch: something like:

```
class Parser
  SWITCH_SOURCE_ENV_VAR = :env 
  SWITCH_SOURCE_CLI_ARG = :arg 

  ...

  def switch(...)
    ...
    enable_switch(*names, source: SWITCH_SOURCE_ENV_VAR)
  end
  
  ...

  def check_conflicts
    ...
    env_var_options = violations.select do |option|
      @switch_sources[option_to_name(option)] == SWITCH_SOURCE_ENV_VAR
    end
  end
end
```

Just to avoid having `:env` and `:arg` scattered around the code as magic values

MikeMcQuaid: Feels a bit overkill. I think the symbols are fine.

BenMusch: Sounds good. This is all set from my perspective, last thing would just be making sure that y'all are okay with the `brew style` failure, or if it would be better to update https://github.com/Homebrew/brew/pull/5738/files#diff-40e44d1232a302a388e451cc3a0a22f9R186 to `expect(...).to be_switch_a`, etc.

MikeMcQuaid: @BenMusch Need to have `brew style` be passing I'm afraid, thanks. `brew style --fix` may be able to do that for you automatically.

Library/Homebrew/cli_parser.rb
```diff
@@ -51,14 +52,14 @@ def switch(*names, description: nil, env: nil, required_for: nil, depends_on: ni
         end
         process_option(*names, description)
         @parser.on(*names, *wrap_option_desc(description)) do
-          enable_switch(*names)
+          enable_switch(*names, from: :args)
```
MikeMcQuaid: Yeh, much prefer this, thanks again!

MikeMcQuaid:
Thanks again @BenMusch, great work!

Library/Homebrew/cli_parser.rb
```diff
@@ -225,7 +233,13 @@ def check_conflicts
 
           next if violations.count < 2
 
-          raise OptionConflictError, violations.map(&method(:name_to_option))
+          env_var_options = violations.select do |option|
+            @switch_sources[option_to_name(option)] == :env_var
```
zmwangx: `:env_var` should be `:env`?

BenMusch: Sounds right. Has this been fixed yet? I should also probably re-write the tests to catch this bug in the future

MikeMcQuaid: Not yet!

scpeters: Yeah, the tests should ideally catch this failure

BenMusch: Sounds good, will take care of this today

BenMusch: So, looks like the issue with the test is that stubbing env vars isn't working at all. My tests unfortunately were structured in a way that they passed regardless of that. For example, this fails:

```ruby
allow(ENV).to receive(:[]).with("HOMEBREW_SWITCH_A").and_return("1")
expect(ENV["HOMEBREW_SWITCH_A"]).to eq("1") # actually is nil
```

Any idea how to fix this?

BenMusch: Found the issue, I had my stubs in the wrong order. Pushing a change now, sorry for the bugs!

PR #5740, opened 2019-02-15
Add `brew cask automerge` command.

reitermarkus:

-----

A command to automatically merge “simple” version bump PRs.

### DIFF
--- a/Library/Homebrew/cask/cmd/automerge.rb
+++ b/Library/Homebrew/cask/cmd/automerge.rb
@@ -1,0 +1,115 @@
+# frozen_string_literal: true
+
+require "cask/cmd/abstract_internal_command"
+require "tap"
+require "utils/formatter"
+require "utils/github"
+
+module Cask
+  class Cmd
+    class Automerge < AbstractInternalCommand
+      OFFICIAL_CASK_TAPS = [
+        "homebrew/cask",
+        "homebrew/cask-drivers",
+        "homebrew/cask-eid",
+        "homebrew/cask-fonts",
+        "homebrew/cask-versions",
+      ].freeze
+
+      def run
+        raise "This command may only be run by Homebrew maintainers." unless ENV["HOMEBREW_DEVELOPER"]
+
+        Homebrew.install_gem! "git_diff"
+        require "git_diff"
+
+        failed = []
+
+        OFFICIAL_CASK_TAPS.each do |tap_name|
+          tap = Tap.fetch(tap_name)
+
+          open_pull_requests = GitHub.pull_requests(tap.full_name, state: :open, base: "master")
+
+          open_pull_requests.each do |pr|
+            next unless passed_ci(pr)
+            next unless check_diff(pr)
+
+            number = pr["number"]
+            sha = pr.dig("head", "sha")
+
+            print "#{Formatter.url(pr["html_url"])} "
+
+            begin
+              GitHub.merge_pull_request(
+                tap.full_name,
+                number: number, sha: sha,
+                merge_method: :squash,
+                commit_message: "Squashed and auto-merged via `brew cask automerge`."
+              )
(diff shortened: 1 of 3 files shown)


reitermarkus:
Yeah, I'd also like to eventually have a bot do this. Anyways, this is the first step so we have the logic which the bot can then absorb.

MikeMcQuaid:
@reitermarkus Yeh, good call. Now we have GitHub Actions that could be worth playing with for any bot.

vitorgalvao:
I’m still ambivalent about this regarding users that submit “wrong” versions (unstable software to stable repos).

I’d feel safer with a hardcoded list of know good contributors. There’s less than a handful that fit that description and submit the bulk of the PRs, and if we ever implement our own system for automatic submission, it would be trivial to only allow that.

Library/Homebrew/cask/cmd/automerge.rb
```diff
@@ -0,0 +1,115 @@
...
+      OFFICIAL_CASK_TAPS = [
+        "homebrew/cask",
+        "homebrew/cask-drivers",
+        "homebrew/cask-eid",
+        "homebrew/cask-fonts",
+        "homebrew/cask-versions",
+      ].freeze
+
+      def run
+        raise "This command should only be run by Homebrew maintainers." unless ENV["HOMEBREW_DEVELOPER"]
```
vitorgalvao:
```suggestion
        raise "This command may only be run by Homebrew maintainers." unless ENV["HOMEBREW_DEVELOPER"]
```

I just know non-maintainers will try to run this, and be puzzled. “may” instead of “can” because non-maintainers can still run technically it, it just won’t work.

Library/Homebrew/cask/cmd/automerge.rb
```diff
@@ -0,0 +1,115 @@
...
+      OFFICIAL_CASK_TAPS = [
+        "homebrew/cask",
+        "homebrew/cask-drivers",
+        "homebrew/cask-eid",
+        "homebrew/cask-fonts",
+        "homebrew/cask-versions",
+      ].freeze
+
+      def run
+        raise "This command may only be run by Homebrew maintainers." unless ENV["HOMEBREW_DEVELOPER"]
```
vitorgalvao: It shouldn’t stop merging this now, but we should consider for the future (as an alternative to this line) adding a `has_write_permission` method to `utils/github` to check if the caller can write to the repo, and if not nip this in the bud early (“You do not have permissions to merge pull requests in this repo”) instead of checking for `HOMEBREW_DEVELOPER` (which users can set as well) and then try to merge and fail.

reitermarkus:
> users that submit “wrong” versions

I don't see this as enough of a reason to have a whitelist of “good” contributors.

The way I see this scenario play out:

Worst Case: We get an issue regarding the unstable version not working.
Likely Case: The unstable version is still stable enough and nobody notices.
Best Case: We get a PR updating it to the correct version, which can get auto-merged as well.

Library/Homebrew/cask/cmd/automerge.rb
reitermarkus: Done.

vitorgalvao:
@reitermarkus Alright, I’m convinced with the scenarios.

core-code:
> Best Case: We get a PR updating it to the correct version, which can get auto-merged as well.

not happy there. finding invalid updates that have already been merged is a LOT more work than finding valid updates and submitting PRs.

> Likely Case: The unstable version is still stable enough and nobody notices.

i don't like the idea of being OK with unofficial versions in HBC just because they seem 'stable enough so that no-one complains'. the rule that we only want official, stable versions is a good one and in my opinion we should stick to it, except there is enough evidence and consent that we should make an exception.

core-code:
i'm seeing too many bad merged by the bot.

but i don't want to be someone who just complains, so when i'm back in the office i'll be working on implementing something to detect beta versions,

PR #5746, opened 2019-02-15
Dockerfile: Remove environment variable USER

sjackman:
The default user is now `root`.


sjackman:
How long does `coverage/coveralls` usually take?

PR #5752, opened 2019-02-16
Dockerfile: do not chown linuxbrew folder

iMichka:
The linuxbrew user has been removed from the Dockerfile

-----

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,6 @@ ADD . /home/linuxbrew/.linuxbrew/Homebrew
 RUN cd /home/linuxbrew/.linuxbrew \
 	&& mkdir -p bin etc include lib opt sbin share var/homebrew/linked Cellar \
 	&& ln -s ../Homebrew/bin/brew /home/linuxbrew/.linuxbrew/bin/ \
-	&& chown -R linuxbrew: /home/linuxbrew/.linuxbrew \
 	&& cd /home/linuxbrew/.linuxbrew/Homebrew \
 	&& git remote set-url origin https://github.com/Homebrew/brew
 


iMichka:
This should fix the error in the latest docker hub build:
```
Step 9/9 : RUN HOMEBREW_NO_ANALYTICS=1 HOMEBREW_NO_AUTO_UPDATE=1 brew tap homebrew/core && rm -rf ~/.cache
---> Using cache
---> 3ffb2a60da6d
Successfully built 3ffb2a60da6d
Successfully tagged b5ykqxe3xtpfwkomvyo4xw6_sut:latest
Creating b5ykqxe3xtpfwkomvyo4xw6_sut_1 ...
Creating b5ykqxe3xtpfwkomvyo4xw6_sut_1 ... done
Error: Can't create update lock in /home/linuxbrew/.linuxbrew/var/homebrew/locks!
Fix permissions by running:
sudo chown -R $(whoami) /home/linuxbrew/.linuxbrew/var/homebrew
```

MikeMcQuaid:
Thanks @iMichka! What commands do Dockerhub run? I wonder if it's worth running them in our CI (and possibly even running `brew test-bot` in the Docker image)?

iMichka:
I think it just runs `docker build` on the dockerfile to build the image.

We could probably add this to the brew build on pull requests, but the builds are quite (this build has to even started right now, docker hub seems to be really slow). So the only way to know something is broken at the moment is to monitor https://hub.docker.com/r/linuxbrew/brew/builds. Failed builds are not displayed, you need to be logged in to see the logs ...

sjackman:
The `linuxbrew` user still exists, though the default user is now `root`. I had intentionally left `/home/linuxbrew` owned by the `linuxbrew` user, so that downstream Docker images can add `USER linuxbrew` to run as the `linuxbrew` user, if their workflow depends on that. That's still possible, but they'll need to add…
```Dockerfile
RUN chown -R linuxbrew: /home/linuxbrew/.linuxbrew
USER linuxbrew
```
(mostly documenting this for myself)

The `root` user should always have write access to `/home/linuxbrew/.linuxbrew` regardless of its ownership.

PR #5755, opened 2019-02-16
cleanup: disable auto-cleanup in CI

mistydemeo:

-----

A user pointed out to me that Travis builds which use Homebrew are now always, 100% of the time starting with a `cleanup` run. The reason for this is that Travis's images include a preinstalled Homebrew which is more than 30 days old, and so these images are always eligible for an auto-cleanup. In a CI environment where the Homebrew installation is clean, we don't need to worry about cleaning up; running it adds time to every build.

I chose `CI` as the variable to use since [it's one of the ones Travis defines](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) and it is likely to be used by several CI systems.

### DIFF
--- a/Library/Homebrew/cleanup.rb
+++ b/Library/Homebrew/cleanup.rb
@@ -160,6 +160,7 @@ module Homebrew
 
     def self.install_formula_clean!(f)
       return if ENV["HOMEBREW_NO_INSTALL_CLEANUP"]
+      return if ENV["CI"]
 
       cleanup = Cleanup.new
       if cleanup.periodic_clean_due?
@@ -171,6 +172,7 @@ module Homebrew
 
     def periodic_clean_due?
       return false if ENV["HOMEBREW_NO_INSTALL_CLEANUP"]
+      return false if ENV["CI"]
       return true unless PERIODIC_CLEAN_FILE.exist?
 
       PERIODIC_CLEAN_FILE.mtime < CLEANUP_DEFAULT_DAYS.days.ago


mistydemeo:
I don't think it should be up to the CI system to opt out. I see the auto-cleanup as an intrusive but desirable feature for regular usage, but something where it's on us to limit the environments where it's applied.

> and will break the configuration on our Jenkins, too

The same default should also apply to Jenkins. Perhaps we should offer an explicit opt-in that overrides CI detection.

MikeMcQuaid:
> I don't think it should be up to the CI system to opt out. I see the auto-cleanup as an intrusive but desirable feature for regular usage

I'm not sure I understand why users should need to opt-out but CI systems should not.

> The same default should also apply to Jenkins. Perhaps we should offer an explicit opt-in that overrides CI detection.

My point wasn't so much pointing out that would negatively affect our Jenkins builds (although it would as-is) but that the assumption that Homebrew in CI environment is always run in an ephemeral environment that doesn't ever want cleanup. If people run CI systems, read the announcements and our release notes and expect Homebrew to cleanup automatically it will be surprising for them for us to reverse this behaviour.

To bring back the auto-update example: this definitely seems something that individual CI providers should configure in their environments accordingly. For example even:

> The reason for this is that Travis's images include a preinstalled Homebrew which is more than 30 days old

is a Travis CI issue that they could (and I believe should) resolve themselves rather than us changing the global CI defaults for their benefit.

PR #5758, opened 2019-02-16
pull: allow to pull from merge commits

iMichka:

-----

### DIFF
--- a/Library/Homebrew/bottle_publisher.rb
+++ b/Library/Homebrew/bottle_publisher.rb
@@ -1,0 +1,162 @@
+# frozen_string_literal: true
+
+require "utils"
+require "formula_info"
+
+class BottlePublisher
+  def initialize(tap, changed_formulae_names, bintray_org, no_publish, warn_on_publish_failure)
+    @tap = tap
+    @changed_formulae_names = changed_formulae_names
+    @no_publish = no_publish
+    @bintray_org = bintray_org
+    @warn_on_publish_failure = warn_on_publish_failure
+  end
+
+  def publish_and_check_bottles
+    # Formulae with affected bottles that were published
+    bintray_published_formulae = []
+
+    # Publish bottles on Bintray
+    unless @no_publish
+      published = publish_changed_formula_bottles
+      bintray_published_formulae.concat(published)
+    end
+
+    # Verify bintray publishing after all patches have been applied
+    bintray_published_formulae.uniq!
+    verify_bintray_published(bintray_published_formulae)
+  end
+
+  def publish_changed_formula_bottles
+    raise "Need to load formulae to publish them!" if ENV["HOMEBREW_DISABLE_LOAD_FORMULA"]
+
+    published = []
+    bintray_creds = { user: ENV["HOMEBREW_BINTRAY_USER"], key: ENV["HOMEBREW_BINTRAY_KEY"] }
+    if bintray_creds[:user] && bintray_creds[:key]
+      @changed_formulae_names.each do |name|
+        f = Formula[name]
+        next if f.bottle_unneeded? || f.bottle_disabled?
+
+        bintray_org = @bintray_org || @tap.user.downcase
+        next unless publish_bottle_file_on_bintray(f, bintray_org, bintray_creds)
+
+        published << f.full_name
+      end
+    else
+      opoo "You must set HOMEBREW_BINTRAY_USER and HOMEBREW_BINTRAY_KEY to add or update bottles on Bintray!"
+    end
+    published
+  end
+
+  # Publishes the current bottle files for a given formula to Bintray
+  def publish_bottle_file_on_bintray(f, bintray_org, creds)
+    repo = Utils::Bottles::Bintray.repository(f.tap)
+    package = Utils::Bottles::Bintray.package(f.name)
+    info = FormulaInfo.lookup(f.full_name)
+    raise "Failed publishing bottle: failed reading formula info for #{f.full_name}" if info.nil?
+
+    unless info.bottle_info_any
+      opoo "No bottle defined in formula #{package}"
+      return false
+    end
+    version = info.pkg_version
+    ohai "Publishing on Bintray: #{package} #{version}"
+    curl "--write-out", '\n', "--silent", "--fail",
+         "--user", "#{creds[:user]}:#{creds[:key]}", "--request", "POST",
+         "--header", "Content-Type: application/json",
+         "--data", '{"publish_wait_for_secs": 0}',
+         "https://api.bintray.com/content/#{bintray_org}/#{repo}/#{package}/#{version}/publish"
+    true
+  rescue => e
+    raise unless @warn_on_publish_failure
+
+    onoe e
+    false
+  end
+
+  # Verifies that formulae have been published on Bintray by downloading a bottle file
+  # for each one. Blocks until the published files are available.
+  # Raises an error if the verification fails.
+  # This does not currently work for `brew pull`, because it may have cached the old
+  # version of a formula.
+  def verify_bintray_published(formulae_names)
+    return if formulae_names.empty?
+
+    raise "Need to load formulae to verify their publication!" if ENV["HOMEBREW_DISABLE_LOAD_FORMULA"]
+
+    ohai "Verifying bottles published on Bintray"
+    formulae = formulae_names.map { |n| Formula[n] }
+    max_retries = 300 # shared among all bottles
+    poll_retry_delay_seconds = 2
+
+    HOMEBREW_CACHE.cd do
+      formulae.each do |f|
+        retry_count = 0
+        wrote_dots = false
+        # Choose arbitrary bottle just to get the host/port for Bintray right
+        jinfo = FormulaInfo.lookup(f.full_name)
+        unless jinfo
+          opoo "Cannot publish bottle: Failed reading info for formula #{f.full_name}"
+          next
+        end
+        bottle_info = jinfo.bottle_info_any
+        unless bottle_info
+          opoo "No bottle defined in formula #{f.full_name}"
+          next
+        end
+
+        # Poll for publication completion using a quick partial HEAD, to avoid spurious error messages
+        # 401 error is normal while file is still in async publishing process
+        url = URI(bottle_info.url)
+        puts "Verifying bottle: #{File.basename(url.path)}"
+        http = Net::HTTP.new(url.host, url.port)
+        http.use_ssl = true
+        retry_count = 0
+        http.start do
+          loop do
+            req = Net::HTTP::Head.new bottle_info.url
+            req.initialize_http_header "User-Agent" => HOMEBREW_USER_AGENT_RUBY
+            res = http.request req
+            break if res.is_a?(Net::HTTPSuccess) || res.code == "302"
+
+            unless res.is_a?(Net::HTTPClientError)
+              raise "Failed to find published #{f} bottle at #{url} (#{res.code} #{res.message})!"
+            end
+
+            raise "Failed to find published #{f} bottle at #{url}!" if retry_count >= max_retries
+
+            print(wrote_dots ? "." : "Waiting on Bintray.")
+            wrote_dots = true
+            sleep poll_retry_delay_seconds
+            retry_count += 1
+          end
+        end
+
+        # Actual download and verification
+        # We do a retry on this, too, because sometimes the external curl will fail even
+        # when the prior HEAD has succeeded.
+        puts "\n" if wrote_dots
+        filename = File.basename(url.path)
+        curl_retry_delay_seconds = 4
+        max_curl_retries = 1
+        retry_count = 0
+        # We're in the cache; make sure to force re-download
+        loop do
+          begin
+            curl_download url, to: filename
+            break
+          rescue
+            raise "Failed to download #{f} bottle from #{url}!" if retry_count >= max_curl_retries
+
+            puts "curl download failed; retrying in #{curl_retry_delay_seconds} sec"
+            sleep curl_retry_delay_seconds
+            curl_retry_delay_seconds *= 2
+            retry_count += 1
+          end
+        end
+        checksum = Checksum.new(:sha256, bottle_info.sha256)
+        Pathname.new(filename).verify_checksum(checksum)
+      end
+    end
+  end
+end
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -90,8 +90,6 @@ module Homebrew
 
     do_bump = args.bump? && !args.clean?
 
-    # Formulae with affected bottles that were published
-    bintray_published_formulae = []
     tap = nil
 
     ARGV.named.each do |arg|
@@ -132,6 +130,7 @@ module Homebrew
       HOMEBREW_CACHE.mkpath
 
       # Store current revision and branch
+      merge_commit = merge_commit?(url)
       orig_revision = `git rev-parse --short HEAD`.strip
       branch = `git symbolic-ref --short HEAD`.strip
 
@@ -139,27 +138,25 @@ module Homebrew
         opoo "Current branch is #{branch}: do you need to pull inside master?"
       end
 
-      patch_puller = PatchPuller.new(url, args)
-      patch_puller.fetch_patch
-      patch_changes = files_changed_in_patch(patch_puller.patchpath, tap)
+      unless merge_commit
+        patch_puller = PatchPuller.new(url, args)
+        patch_puller.fetch_patch
+        patch_changes = files_changed_in_patch(patch_puller.patchpath, tap)
 
-      is_bumpable = patch_changes[:formulae].length == 1 && patch_changes[:others].empty?
-      if do_bump
-        odie "No changed formulae found to bump" if patch_changes[:formulae].empty?
-        if patch_changes[:formulae].length > 1
-          odie "Can only bump one changed formula; bumped #{patch_changes[:formulae]}"
-        end
-        odie "Cannot bump if non-formula files are changed" unless patch_changes[:others].empty?
+        is_bumpable = patch_changes[:formulae].length == 1 && patch_changes[:others].empty?
+        check_bumps(patch_changes) if do_bump
+        old_versions = current_versions_from_info_external(patch_changes[:formulae].first) if is_bumpable
+        patch_puller.apply_patch
       end
-      old_versions = current_versions_from_info_external(patch_changes[:formulae].first) if is_bumpable
-      patch_puller.apply_patch
+
+      end_revision = end_revision?(url, merge_commit)
 
       changed_formulae_names = []
 
       if tap
         Utils.popen_read(
           "git", "diff-tree", "-r", "--name-only",
-          "--diff-filter=AM", orig_revision, "HEAD", "--", tap.formula_dir.to_s
+          "--diff-filter=AM", orig_revision, end_revision, "--", tap.formula_dir.to_s
         ).each_line do |line|
           next unless line.end_with? ".rb\n"
 
@@ -239,7 +236,6 @@ module Homebrew
         safe_system "git", "commit", "--amend", "--signoff", "--allow-empty", "-q", "-m", message
       end
 
-      # Bottles: Pull bottle block commit and publish bottle files on Bintray
       if fetch_bottles
         bottle_commit_url = if testing_job
           bottle_branch = "testing-bottle-#{testing_job}"
@@ -252,22 +248,67 @@ module Homebrew
 
         curl "--silent", "--fail", "--output", "/dev/null", "--head", bottle_commit_url
 
-        safe_system "git", "checkout", "--quiet", "-B", bottle_branch, orig_revision
-        pull_patch bottle_commit_url, "bottle commit"
-        safe_system "git", "rebase", "--quiet", branch
-        safe_system "git", "checkout", "--quiet", branch
-        safe_system "git", "merge", "--quiet", "--ff-only", "--no-edit", bottle_branch
-        safe_system "git", "branch", "--quiet", "-D", bottle_branch
-
-        # Publish bottles on Bintray
-        unless args.no_publish?
-          published = publish_changed_formula_bottles(tap, changed_formulae_names)
-          bintray_published_formulae.concat(published)
+        if merge_commit
+          fetch_merge_patch(bottle_commit_url, args, issue)
+        else
+          fetch_bottles_patch(bottle_commit_url, args, bottle_branch, branch, orig_revision)
         end
+        publish_and_check_bottles(tap, args, changed_formulae_names)
+      elsif merge_commit
+        fetch_merge_patch(url, args, issue)
       end
 
       ohai "Patch changed:"
-      safe_system "git", "diff-tree", "-r", "--stat", orig_revision, "HEAD"
+      safe_system "git", "diff-tree", "-r", "--stat", orig_revision, end_revision
+    end
+  end
+
+  def check_bumps(patch_changes)
+    if patch_changes[:formulae].empty?
+      odie "No changed formulae found to bump"
+    elsif patch_changes[:formulae].length > 1
+      odie "Can only bump one changed formula; bumped #{patch_changes[:formulae]}"
+    elsif !patch_changes[:others].empty?
+      odie "Can not bump if non-formula files are changed"
+    end
+  end
+
+  def fetch_issue(url)
+    issue = url[%r{/pull\/([0-9]+)}, 1]
+    safe_system "git", "fetch", "--quiet", "origin", "pull/#{issue}/head"
+  end
+
+  def merge_commit?(url)
+    fetch_issue(url)
+    Utils.popen_read("git", "rev-list", "--parents", "-n1", "FETCH_HEAD").count(" ") > 1
+  end
+
+  def end_revision?(url, merge_commit)
+    fetch_issue(url)
+    Utils.popen_read("git", "rev-parse", merge_commit ? "FETCH_HEAD" : "HEAD").strip
+  end
+
+  def fetch_bottles_patch(bottle_commit_url, args, bottle_branch, branch, orig_revision)
+    safe_system "git", "checkout", "--quiet", "-B", bottle_branch, orig_revision
+    PatchPuller.new(bottle_commit_url, args, "bottle commit").pull_patch
+    safe_system "git", "rebase", "--quiet", branch
+    safe_system "git", "checkout", "--quiet", branch
+    safe_system "git", "merge", "--quiet", "--ff-only", "--no-edit", bottle_branch
+    safe_system "git", "branch", "--quiet", "-D", bottle_branch
+  end
+
+  def fetch_merge_patch(url, args, issue)
+    PatchPuller.new(url, args, "merge commit").pull_merge_commit(issue)
+  end
+
+  def publish_and_check_bottles(tap, args, changed_formulae_names)
+    # Formulae with affected bottles that were published
+    bintray_published_formulae = []
+
+    # Publish bottles on Bintray
+    unless args.no_publish?
+      published = publish_changed_formula_bottles(tap, changed_formulae_names)
+      bintray_published_formulae.concat(published)
     end
 
     # Verify bintray publishing after all patches have been applied
@@ -300,10 +341,6 @@ module Homebrew
     published
   end
 
-  def pull_patch(url, description = nil)
-    PatchPuller.new(url, args, description).pull_patch
-  end
-
   class PatchPuller
     attr_reader :base_url
     attr_reader :patch_url
@@ -321,6 +358,21 @@ module Homebrew
     def pull_patch
       fetch_patch
       apply_patch
+    end
+
+    def pull_merge_commit(issue)
+      # Used by forks of homebrew-core that use merge-commits (for example linuxbrew)
+      ohai "Fast-forwarding to the merge commit"
+      test_bot_origin = patch_url[%r{(https://github\.com/[\w-]+/[\w-]+)/compare/}, 1]
+      safe_system "git", "fetch", "--quiet", test_bot_origin, "pr-#{issue}" if test_bot_origin
+      safe_system "git", "merge", "--quiet", "--ff-only", "--no-edit", "FETCH_HEAD"
+      return if $CHILD_STATUS.success?
+
+      safe_system "git", "reset", "--hard", "FETCH_HEAD"
+      odie <<~EOS
+        Not possible to fast-forward.
+        Maybe somebody pushed commits to origin/master between the merge commit creation and now.
+      EOS
     end
 
     def fetch_patch
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -8,6 +8,8 @@ require "formula"
 require "formulary"
 require "version"
 require "pkg_version"
+require "bottle_publisher"
+require "formula_info"
 
 module GitHub
   module_function
@@ -253,7 +255,13 @@ module Homebrew
         else
           fetch_bottles_patch(bottle_commit_url, args, bottle_branch, branch, orig_revision)
         end
-        publish_and_check_bottles(tap, args, changed_formulae_names)
+        BottlePublisher.new(
+          tap,
+          changed_formulae_names,
+          args.bintray_org,
+          args.no_publish?,
+          args.warn_on_publish_failure?,
+        ).publish_and_check_bottles
       elsif merge_commit
         fetch_merge_patch(url, args, issue)
       end
@@ -299,46 +307,6 @@ module Homebrew
 
   def fetch_merge_patch(url, args, issue)
     PatchPuller.new(url, args, "merge commit").pull_merge_commit(issue)
-  end
-
-  def publish_and_check_bottles(tap, args, changed_formulae_names)
-    # Formulae with affected bottles that were published
-    bintray_published_formulae = []
-
-    # Publish bottles on Bintray
-    unless args.no_publish?
-      published = publish_changed_formula_bottles(tap, changed_formulae_names)
-      bintray_published_formulae.concat(published)
-    end
-
-    # Verify bintray publishing after all patches have been applied
-    bintray_published_formulae.uniq!
-    verify_bintray_published(bintray_published_formulae)
-  end
-
-  def force_utf8!(str)
-    str.force_encoding("UTF-8") if str.respond_to?(:force_encoding)
-  end
-
-  def publish_changed_formula_bottles(tap, changed_formulae_names)
-    raise "Need to load formulae to publish them!" if ENV["HOMEBREW_DISABLE_LOAD_FORMULA"]
-
-    published = []
-    bintray_creds = { user: ENV["HOMEBREW_BINTRAY_USER"], key: ENV["HOMEBREW_BINTRAY_KEY"] }
-    if bintray_creds[:user] && bintray_creds[:key]
-      changed_formulae_names.each do |name|
-        f = Formula[name]
-        next if f.bottle_unneeded? || f.bottle_disabled?
-
-        bintray_org = args.bintray_org || tap.user.downcase
-        next unless publish_bottle_file_on_bintray(f, bintray_org, bintray_creds)
-
-        published << f.full_name
-      end
-    else
-      opoo "You must set HOMEBREW_BINTRAY_USER and HOMEBREW_BINTRAY_KEY to add or update bottles on Bintray!"
-    end
-    published
   end
 
   class PatchPuller
@@ -440,7 +408,7 @@ module Homebrew
   # Returns info as a hash (type => version), for pull.rb's internal use.
   # Uses special key `:nonexistent => true` for nonexistent formulae.
   def current_versions_from_info_external(formula_name)
-    info = FormulaInfoFromJson.lookup(formula_name)
+    info = FormulaInfo.lookup(formula_name)
     versions = {}
     if info
       [:stable, :devel, :head].each do |spec_type|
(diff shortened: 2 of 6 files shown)


iMichka:
@sjackman I adapted this from the code we had in Linuxbrew (I moved the change to a better place though).

@MikeMcQuaid As we use often use merge commits, I need to be able to pull them. Until now I was using an older copy of Linuxbrew to get things done, but as we will have soon migrated everything I need to be able to pull these pull requests.

I admit that there may be a need to refactor this. Maybe I could also add a test that checks if I am able to pull a merge commit. Though I may need to use a linuxbrew-core URL for that test.

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -145,22 +145,37 @@ def pull
...
-
-      is_bumpable = patch_changes[:formulae].length == 1 && patch_changes[:others].empty?
-      if do_bump
-        odie "No changed formulae found to bump" if patch_changes[:formulae].empty?
-        if patch_changes[:formulae].length > 1
-          odie "Can only bump one changed formula; bumped #{patch_changes[:formulae]}"
+      patch_url = url + ".patch"
+      issue = patch_url[/([0-9]+)\.patch$/, 1]
+      safe_system "git", "fetch", "--quiet", "origin", "pull/#{issue}/head"
+      if Utils.popen_read("git", "rev-list", "--parents", "-n1", "FETCH_HEAD").count(" ") > 1
```
sjackman: These two lines 150–151 could be factored out to a function `merge_commit?`.

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -145,22 +145,37 @@ def pull
...
+        # Used by forks of homebrew-core that use merge-commits
+        ohai "Fast-forwarding to the merge commit"
+        test_bot_origin = patch_url[%r{(https://github\.com/[\w-]+/[\w-]+)/compare/}, 1]
+        safe_system "git", "fetch", "--quiet", test_bot_origin, "pr-#{issue}" if test_bot_origin
+        system "git", "merge", "--quiet", "--ff-only", "--no-edit", "FETCH_HEAD"
+        unless $CHILD_STATUS.success?
+          safe_system "git", "reset", "--hard", "FETCH_HEAD"
+          odie "Not possible to fast-forward"
         end
```
sjackman: These lines 152–160 could be factored to a function named something like `git_checkout_pr`.
Then…
```ruby
if merge_commit? "pull/#{issue}/head"
  git_checkout_pr patch_url
else
…
```

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -145,22 +145,37 @@ def pull
...
-        odie "Can not bump if non-formula files are changed" unless patch_changes[:others].empty?
-      end
-      if is_bumpable
-        old_versions = current_versions_from_info_external(patch_changes[:formulae].first)
+      else
+        patch_puller = PatchPuller.new(url, args)
```
MikeMcQuaid: Would like to have some of the above logic live in `PatchPuller` and generally share more code with that.

iMichka:
I changed the code a little bit more, because I forgot to test the case where I was pulling bottles with the `--bottle` flag.

Here is the output I get after this change:
```
brew pull --clean --bottle --warn-on-publish-failure --bintray-org=linuxbrew https://github.com/Linuxbrew/homebrew-core/pull/11715
==> jenkins: skipping unneeded bottle.
==> Fast-forwarding to the merge commit
==> Publishing on Bintray: mage 1.8.0
{"files":0}
==> Publishing on Bintray: mariadb:10.1 10.1.38
curl: (22) The requested URL returned error: 404 Not Found

==> No bottle found on Bintray for: mariadb:10.1 10.1.38 - skipping
==> Publishing on Bintray: pixman 0.38.0
{"files":0}
==> Publishing on Bintray: ship 0.33.1
curl: (22) The requested URL returned error: 404 Not Found

==> No bottle found on Bintray for: ship 0.33.1 - skipping
==> Patch changed:
 Formula/jenkins.rb      |  4 ++--
 Formula/mage.rb         | 34 ++++++++++++++++++++++++++++++++++
 Formula/mariadb@10.1.rb | 10 +++++-----
 Formula/pixman.rb       | 13 +++++++------
 Formula/ship.rb         | 10 +++++-----
 5 files changed, 53 insertions(+), 18 deletions(-)
Verifying bottle: mage-1.8.0.x86_64_linux.bottle.tar.gz
Verifying bottle: pixman-0.38.0.x86_64_linux.bottle.tar.gz
==> Published bottles for:
mage pixman
```

@sjackman I have been using `--warn-on-publish-failure`. In linuxbrew, you had a different patch that would recalculate the list of formulae that actually needed a bottle pull. To avoid porting that part of the code, I will use `--warn-on-publish-failure` from now on.

I am just a little bit annoyed that I get the `curl` errors in the output, not sure how I can avoid that.

@MikeMcQuaid it is expected that not all bottles are built in a merge commit (because our CI sucks, or because there is a new bug somewhere). In that case I pull the bottles that have been built, and open separate pull requests to fix/build the missing ones.

I hope the strategy used for the PatchPuller is fine. We could go for a second round of refactoring and hide more logic inside the PatchPuller. But maybe we should then also rename that class, as it does not only pull patches, but fetches merges commits. Maybe it could be called `ChangesPuller`?

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -145,29 +145,34 @@ def pull
...
+        if is_bumpable
+          old_versions = current_versions_from_info_external(patch_changes[:formulae].first)
+        end
+        patch_puller.apply_patch
       end
-      patch_puller.apply_patch
 
       changed_formulae_names = []
 
+      end_revision = merge_commit ? "FETCH_HEAD" : "HEAD"
```
sjackman: I'm not crazy on relying on using `FETCH_HEAD` head. I consider it to be transient, since the next `git fetch` would blow it away. It could create a difficult to troubleshoot bug if some function called some function that called `git fetch`. I suggest checking out the result of `FETCH_HEAD` to a named branch to update `HEAD`, and then this logic can go away.

MikeMcQuaid: I think it's fine to use `FETCH_HEAD` as long as we're doing the `git fetch` immediately before getting its value. Creating a new branch should be unnecessary but `FETCH_HEAD` should be stored in a variable right after the fetch.

sjackman: How about we store the SHA1 in that variable rather than literal `FETCH_HEAD`?

iMichka: I changed the code to do that.

iMichka:
I forgot another use-case ... merge commits without bottles, like https://github.com/Linuxbrew/homebrew-core/pull/11725

The code is currently quite ugly, I need to refactor this a little bit more. And this needs a little bit more of testing too. I also saw that I broke the test-bot with this change, so I need to address that issue too.

I'll need a little more time for this. You can have a look at the code, but it is definitely not the final version.

Library/Homebrew/dev-cmd/pull.rb
MikeMcQuaid: Works for me 👍

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -237,40 +243,72 @@ def pull
         safe_system "git", "commit", "--amend", "--signoff", "--allow-empty", "-q", "-m", message
       end
 
+      if testing_job
```
MikeMcQuaid: `bottle_branch = if testing_job`

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -461,9 +507,12 @@ def publish_bottle_file_on_bintray(f, bintray_org, creds)
          "https://api.bintray.com/content/#{bintray_org}/#{repo}/#{package}/#{version}/publish"
     true
   rescue => e
-    raise unless @args.warn_on_publish_failure?
-
-    onoe e
+    if @args.warn_on_publish_failure?
+      ohai "No bottle found on Bintray for: #{package} #{version} - skipping"
```
MikeMcQuaid: How is this change related?

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -321,6 +355,18 @@ def pull_patch
...
 
+    def pull_merge_commit(issue)
+      # Used by forks of homebrew-core that use merge-commits (for example linuxbrew)
+      ohai "Fast-forwarding to the merge commit"
+      test_bot_origin = patch_url[%r{(https://github\.com/[\w-]+/[\w-]+)/compare/}, 1]
+      safe_system "git", "fetch", "--quiet", test_bot_origin, "pr-#{issue}" if test_bot_origin
+      system "git", "merge", "--quiet", "--ff-only", "--no-edit", "FETCH_HEAD"
+      unless $CHILD_STATUS.success?
+        safe_system "git", "reset", "--hard", "FETCH_HEAD"
+        odie "Not possible to fast-forward"
```
MikeMcQuaid: Would be good to improve this message to include why/what a workaround is.

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -321,6 +355,18 @@ def pull_patch
       apply_patch
     end
 
+    def pull_merge_commit(issue)
+      # Used by forks of homebrew-core that use merge-commits (for example linuxbrew)
+      ohai "Fast-forwarding to the merge commit"
+      test_bot_origin = patch_url[%r{(https://github\.com/[\w-]+/[\w-]+)/compare/}, 1]
+      safe_system "git", "fetch", "--quiet", test_bot_origin, "pr-#{issue}" if test_bot_origin
+      system "git", "merge", "--quiet", "--ff-only", "--no-edit", "FETCH_HEAD"
+      unless $CHILD_STATUS.success?
```
MikeMcQuaid: `return if $CHILD_STATUS.success?`

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -237,40 +243,72 @@ def pull
...
+
+        ohai "Published bottles for:"
+        puts bintray_published_formulae.join " "
+      end
+
+      if merge_commit && !fetch_bottles
+        ohai "Fast-forwarding to the merge commit"
+        safe_system "git", "checkout", "--quiet", "-B", bottle_branch, orig_revision
+        fetch_issue(url)
+        system "git", "merge", "--quiet", "--ff-only", "--no-edit", "FETCH_HEAD"
```
MikeMcQuaid: Why is this one not `safe_system`?

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -237,40 +243,72 @@ def pull
...
         unless args.no_publish?
           published = publish_changed_formula_bottles(tap, changed_formulae_names)
           bintray_published_formulae.concat(published)
         end
+
+        # Verify bintray publishing after all patches have been applied
+        bintray_published_formulae.uniq!
+        verify_bintray_published(bintray_published_formulae)
+
+        ohai "Published bottles for:"
```
MikeMcQuaid: Why this change?

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -137,27 +137,33 @@ def pull
...
+      merge_commit = merge_commit?(url)
+
+      unless merge_commit
+        patch_puller = PatchPuller.new(url, args)
+        patch_puller.fetch_patch
+        patch_changes = files_changed_in_patch(patch_puller.patchpath, tap)
+
+        is_bumpable = patch_changes[:formulae].length == 1 && patch_changes[:others].empty?
+        if do_bump
+          odie "No changed formulae found to bump" if patch_changes[:formulae].empty?
```
MikeMcQuaid: Make this and the two conditions below all `if`/`elsif`/else` as they all return anyway.

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -137,27 +137,33 @@ def pull
...
-      patch_changes = files_changed_in_patch(patch_puller.patchpath, tap)
-
-      is_bumpable = patch_changes[:formulae].length == 1 && patch_changes[:others].empty?
-      if do_bump
-        odie "No changed formulae found to bump" if patch_changes[:formulae].empty?
-        if patch_changes[:formulae].length > 1
-          odie "Can only bump one changed formula; bumped #{patch_changes[:formulae]}"
+      merge_commit = merge_commit?(url)
+
+      unless merge_commit
```
MikeMcQuaid: Maybe this could all be a separate function?

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -237,40 +243,72 @@ def pull
...
 
         curl "--silent", "--fail", "--output", "/dev/null", "--head", bottle_commit_url
 
-        safe_system "git", "checkout", "--quiet", "-B", bottle_branch, orig_revision
-        pull_patch bottle_commit_url, "bottle commit"
-        safe_system "git", "rebase", "--quiet", branch
-        safe_system "git", "checkout", "--quiet", branch
-        safe_system "git", "merge", "--quiet", "--ff-only", "--no-edit", bottle_branch
-        safe_system "git", "branch", "--quiet", "-D", bottle_branch
+        if merge_commit
```
MikeMcQuaid: Maybe this could all be a separate function?

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -237,40 +243,72 @@ def pull
...
+
+        # Verify bintray publishing after all patches have been applied
+        bintray_published_formulae.uniq!
+        verify_bintray_published(bintray_published_formulae)
+
+        ohai "Published bottles for:"
+        puts bintray_published_formulae.join " "
+      end
+
+      if merge_commit && !fetch_bottles
```
MikeMcQuaid: Maybe this could all be a separate function?

Library/Homebrew/dev-cmd/pull.rb
iMichka: That's a copy/paste typo. I'll change that.

Library/Homebrew/dev-cmd/pull.rb
iMichka: It's just cosmetics. But I am still unhappy with the result. I will go back to the initial code, this change is unrelated. I'll open a separate PR for this later.

Library/Homebrew/dev-cmd/pull.rb
iMichka: I'll add an explanation of the most common issue which triggers this.

Library/Homebrew/dev-cmd/pull.rb
iMichka: To have a feedback when the merge is done. Else the output is quite empty and you are not sure if it worked. I will remove this for now, and make a different PR for this one later, so we can bike-shed a little bit the output. I need to check how it looks on mac too.

Library/Homebrew/dev-cmd/pull.rb
iMichka: Will change that; I do not have the ruby reflex to write code like that :) Thanks.

iMichka:
FYI, I tried to do something like:
`end_revision = Utils.popen_read("git", "rev-parse", merge_commit ? "FETCH_HEAD" : "HEAD")`

In the case of FETCH_HEAD, it gives me a commit I can access (using gitk for example); but as soon as you have a fetch inbetween (or some other git calls, nut sure which ones), that commit is gone.

So when you need "FETCH_HEAD", you need to use it as soon as possible. I will try to encapsulate these calls in standalone functions.

sjackman:
> In the case of FETCH_HEAD, it gives me a commit I can access (using gitk for example); but as soon as you have a fetch inbetween (or some other git calls, nut sure which ones), that commit is gone.
So when you need "FETCH_HEAD", you need to use it as soon as possible. I will try to encapsulate these calls in standalone functions.

If the ref doesn't have a name, it can be garbage collected. That's shouldn't happen until the SHA1 is more than two weeks old with no name. You can use `git fetch origin pull/1234/head:pr-1234` to ensure that it has a name, even if you don't use that name.

Library/Homebrew/dev-cmd/pull.rb
iMichka: Can I do that in a separate PR? The change is already getting quite big, and changing this part is a different part of refactoring. Or should I squash my commits, and make a second separate commit for this part?

Library/Homebrew/dev-cmd/pull.rb
iMichka: Same comment as above. Can I do that in a separate PR?

iMichka:
Ok, I had a bug and `end_revisions` needed a `.strip` to have the right sha1 ... now it works.
I refactored the code so that we have more smaller functions.

One change I may still make: `pull_merge_commit(issue, end_revision)` has 2 arguments. I can pass these through the `PatchPuller` constructor. At this point, with all these arguments, I am even think of introducing an intermediate object like a `PatchPullerConfig` class, to pass the different arguments around. Any opinion on this? We can also keep that for a second PR.

iMichka:
```
  1) brew pull fetches a patch from a GitHub commit or pull request and applies it
     Failure/Error:
       expect { brew "pull", "1" }
         .to output(/Fetching patch/).to_stdout
         .and output(/Current branch is new\-branch/).to_stderr
         .and be_a_failure

       expected block to output /Fetching patch/ to stdout, but output nothing
     # ./test/dev-cmd/pull_spec.rb:13:in `block (2 levels) in <top (required)>'
```

Library/Homebrew/dev-cmd/pull.rb
MikeMcQuaid: If you feel strongly you could but I think given the added branching this PR adds it's making `pull` harder to read and splitting into functions would mitigate (and perhaps improve) this. Up to you 👍

iMichka:
This was working, until we changed the repo to `linuxbrew-core`. Looks like it will ned some more love :)

iMichka:
There is one mac-failure that is unrelated, I will rebase this and re-run the CI:

```
Failures:

  1) Cask::Pkg#uninstall removes files and dirs referenced by the pkg
     Failure/Error: pkg.uninstall

     ErrorDuringExecution:
       Failure while executing; `/usr/bin/xargs -0 -- /bin/rm --` exited with 1. Here's the output:
       rm: /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/special_file20190302-18852-1hfwnlr: No such file or directory
     # ./system_command.rb:107:in `assert_success'
     # ./system_command.rb:50:in `run!'
     # ./system_command.rb:27:in `run'
     # ./test/support/helper/cask/never_sudo_system_command.rb:5:in `run'
     # ./system_command.rb:31:in `run!'
     # ./cask/pkg.rb:31:in `uninstall'
     # ./test/cask/pkg_spec.rb:22:in `block (3 levels) in <top (required)>'
     # ./test/support/helper/spec/shared_context/homebrew_cask.rb:47:in `block (2 levels) in <top (required)>'
     # ./test/spec_helper.rb:160:in `block (2 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:123:in `block in run'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `loop'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `run'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:37:in `block (2 levels) in setup'

Finished in 3 minutes 41.8 seconds (files took 2.7 seconds to load)
544 examples, 1 failure

Failed examples:

rspec ./test/cask/pkg_spec.rb:7 # Cask::Pkg#uninstall removes files and dirs referenced by the pkg
```

iMichka:
```
     1.1) Failure/Error:
            expect { brew "pull", "--bump", "https://api.github.com/repos/Homebrew/homebrew-core/pulls/122" }
              .to output(/Fetching patch/).to_stdout
              .and output(/Can only bump one changed formula/).to_stderr
              .and be_a_failure

            expected block to output /Can only bump one changed formula/ to stderr, but output "fatal: Needed a single revision\nWarning: Current branch is new-branch: do you need to pull inside master?\nError: No changed formulae found to bump\n"
          # ./test/dev-cmd/pull_spec.rb:22:in `block (2 levels) in <top (required)>'
Tests Failed
          # ./test/support/helper/spec/shared_context/integration_test.rb:44:in `block (2 levels) in <top (required)>'
          # ./test/spec_helper.rb:160:in `block (2 levels) in <top (required)>'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:123:in `block in run'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `loop'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `run'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
          # ./test/spec_helper.rb:122:in `block (2 levels) in <top (required)>'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:123:in `block in run'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `loop'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:110:in `run'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
          # ./vendor/bundle/ruby/2.3.0/gems/rspec-retry-0.6.1/lib/rspec/retry.rb:37:in `block (2 levels) in setup'
```

The interesting part of the output is the following: `No changed formulae found to bump`.
The test is expecting: `Can only bump one changed formula`

I do get the wrong error message too locally. But if I try again with the master branch, without my changes, I also get the wrong error message. So for me the initial test is already broken.

We have this on top of the test code:
```
    CoreTap.instance.path.cd do
      system "git", "init"
      system "git", "checkout", "-b", "new-branch"
    end
```
How is `brew pull --bump https://api.github.com/repos/Homebrew/homebrew-core/pulls/122` supposed to return `Can only bump one changed formula` in an empty core git repo?

sjackman:
> How is `brew pull --bump https://api.github.com/repos/Homebrew/homebrew-core/pulls/122` supposed to return Can only bump one changed formula in an empty core git repo?

It fetches https://github.com/Homebrew/homebrew-core/pull/122.patch using `curl` and see that is has modified five formulae. It doesn't need a git repo for that.

```console
$ curl -Ls https://github.com/Homebrew/homebrew-core/pull/122.patch | diffstat
 cfssl.rb       |    5 +++--
 fontforge.rb   |   12 ++++--------
 godep.rb       |    4 ++--
 letsencrypt.rb |   45 +++++++++++++++++++++++++--------------------
 libuv.rb       |    4 ++--
 5 files changed, 36 insertions(+), 34 deletions(-)
```

iMichka:
Ok Thanks. I made some progress. I tried the following:
```
cd /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/homebrew/
mv homebrew-core homebrew-core2
mkdir homebrew-core
cd homebrew-core
git init
git checkout -b new-branch
git remote add origin https://github.com/Homebrew/homebrew-core
pull --bump https://api.github.com/repos/Homebrew/homebrew-core/pulls/122 
```

Works fine on mac and on Linux with these changes. Is there a way to connect to the azure build with ssh to debug this there? I can not reproduce the failure locally.

MikeMcQuaid:
> Is there a way to connect to the azure build with ssh to debug this there?

No (but it's passing now).

iMichka:
Yes, because I commented out the following line: `merge_commit = merge_commit?(url)` to see what part of my changes are breaking this. This allowed the tests to pass.

I moved `merge_commit = merge_commit?(url)` after the `orig_revision`  and `branch` variable creation now, to see if this was maybe the problem. I am blindly poking around right now until I get it to work.

iMichka:
```
expect { brew "pull", "https://github.com/Homebrew/brew/pull/1249" }
      .to output(/Fetching patch/).to_stdout
      .and output(/Patch failed to apply/).to_stderr
      .and be_a_failure
```

I do not get the logic of that last test. If you run this on mac, I get:
```
==> Fetching patch 
Patch: https://github.com/Homebrew/brew/pull/1249.patch
==> Applying patch
Applying: Revert "Migrate a Formula to Multiple Versions: add docs."
Using index info to reconstruct a base tree...
A	docs/Migrate-A-Formula-To-Multiple-Versions.md
Falling back to patching base and 3-way merge...
No changes -- Patch already applied.
==> Patch closes issue #1249
==> Patch changed:
```

This is not what is expected by the test.
Can somebody run `brew pull https://github.com/Homebrew/brew/pull/1249` to see what is the output on mac?

Are the tests run in a specific order? Is there any setup of the repositories done somewhere else? In what state is the main brew repo during the tests?.

sjackman:
> Can somebody run brew pull https://github.com/Homebrew/brew/pull/1249 to see what is the output on mac?

```
❯❯❯ brew pull https://github.com/Homebrew/brew/pull/1249
Warning: Current branch is install_symlink_p: do you need to pull inside master?
==> Fetching patch 
Patch: https://github.com/Homebrew/brew/pull/1249.patch
==> Applying patch
Applying: Revert "Migrate a Formula to Multiple Versions: add docs."
Using index info to reconstruct a base tree...
A	docs/Migrate-A-Formula-To-Multiple-Versions.md
Falling back to patching base and 3-way merge...
No changes -- Patch already applied.
==> Patch closes issue #1249
==> Patch changed:
```

> Are the tests run in a specific order?

They're run in an arbitrary order, and since they're multithreaded, it can be different each time. Occasionally this difference in order has caused intermittent test failures for me. One intermittent failure in particular involves the style of a temporary shell file failing to lint.

> Is there any setup of the repositories done somewhere else? In what state is the main brew repo during the tests?

I believe the tests are all run on fake or repos that are created for the purpose of testing, though I'm not entirely sure where that setup is done.

jonchang:
Integration tests are set up in https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/support/helper/spec/shared_context/integration_test.rb

sjackman:
```
expected block to output /Patch failed to apply/ to stderr, but output "fatal: not a git repository (or any of the parent directories): .git\nfatal: not a git repository (or any of the parent directories): .git\nfatal: not a git repository (or any of the parent directories): .git\nError: Failure while executing; `git fetch --quiet origin pull/1249/head:merge-branch` exited with 128.\n"
```
`fatal: not a git repository` It's confusing that it's not being run in a git repo. Tweaking the git commands will have no effect as long as it's not being run in a git repo.

iMichka:
I went for a completely different strategy. The tests will setup fake repositories in /tmp. With the previous logic, the brew pull test will fetch the upstream repo, which is slow (at least on my connection at home).

Instead of fetching the upstream branch to check if it is a merge commit, I am using the github API to check if the “merge” label is set. We automatically set that label for merge commits in linuxbrew-core.

I could also check for the “Merge” word in the pull request title. And maybe make sure the PR is open.

At this point I also could just add a “—merge-commit” flag to brew pull. That would also be a valid strategy, but I would like to avoid the multiplication of flags here ...

Thoughts?

MikeMcQuaid:
> Instead of fetching the upstream branch to check if it is a merge commit, I am using the github API to check if the “merge” label is set. We automatically set that label for merge commits in linuxbrew-core.

This seems a bit error-prone. You can probably use the GitHub API to query if it's a merge commit instead.

iMichka:
> This seems a bit error-prone. You can probably use the GitHub API to query if it's a merge commit instead.

Agreed. Now that I have understood how the API works, I changed the code to count the number of parents of the last commit of the pull request (commits are ordered with the last one at the bottom).

iMichka:
One weird thing, if I have more than 30 commits it will not work, for example for https://github.com/Homebrew/linuxbrew-core/pull/12070

The GitHub documentation tells me that it should work for 250 commits: https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -250,22 +248,60 @@ def pull
...
+      odie "No changed formulae found to bump"
+    elsif patch_changes[:formulae].length > 1
+      odie "Can only bump one changed formula; bumped #{patch_changes[:formulae]}"
+    elsif !patch_changes[:others].empty?
+      odie "Can not bump if non-formula files are changed"
+    end
+  end
+
+  def merge_commit?(user, repo, issue)
+    GitHub.commits_for_pr(user, repo, issue)[-1]["parents"].length == 2
```
sjackman:
```suggestion
    GitHub.commits_for_pr(user, repo, issue)[-1]["parents"].length > 1
```

sjackman:
> One weird thing, if I have more than 30 commits it will not work, for example for Homebrew/linuxbrew-core#12070

Perhaps it's paginated? I'm not quite sure how that works. Mike?

MikeMcQuaid:
Yes, it'll be the pagination. 

Honestly this is starting to sound all a bit hacky/unreliable. Could we consider making this workflow just be in Git and splitting the bottle publishing out of this command and into something standalone?

iMichka:
> Honestly this is starting to sound all a bit hacky/unreliable. Could we consider making this workflow just be in Git 

I can go back to a pure git version. Still need to get the tests going though.

> splitting the bottle publishing out of this command and into something standalone?

Should this be done in this PR?

MikeMcQuaid:
> I can go back to a pure git version. Still need to get the tests going though.

At this stage even the pure Git version feels like it might be better placed in another command.



> Should this be done in this PR?

Could be or another PR, up to you.

iMichka:
We had a pull-Linux command: https://github.com/Linuxbrew/homebrew-developer/blob/master/cmd/brew-pull-linux.rb

I never used it, it is probably outdated. We could rework it. It will need a little bit of time on my side, I’ll start working on this this week.

sjackman:
The `pull-linux` command was used interactively. `brew pull` in this situation is being used by `brew test-bot`. If there is a separate command, perhaps it could behave similarly to `brew update` and `brew update-reset`. This new command could be `brew pull-reset`, which simply resets to the identified commit, rather than trying to pull the `.patch` with `curl` and apply it with `git am`. As a bonus, it works with merge commits.

iMichka:
> brew pull in this situation is being used by brew test-bot

Right, forgot about that.

I rebased this on master. I also went back to the initial solution, which relies only on git.
@MikeMcQuaid I have seen the changes you made in the tests. Do you think this may help here?

cc @jonchang who wanted to test this too.

sjackman:
The failure is…
```
expected block to output /Patch failed to apply/ to stderr, but output "fatal: not a git repository (or any of the parent directories): .git\nError: Failure while executing; `git fetch --quiet origin pull/1249/head` exited with 128.\n"
```
https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=6477&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb

The test `dev-cmd/pull` is being run outside of a `git` repo, and it now needs to run inside a repo. In the function `merge_commit?`, I don't see where it specifies the location of the `git` repo. It would help clarify to modify this function (and other new calls to `git`) to specify the location of the `git` repo with `git -C …`. The other existing calls to `git` don't `git -C` either though.
https://github.com/Homebrew/brew/pull/5758/files#diff-de3bb69738cf9f6002692059922a16fbR280

MikeMcQuaid:
> @MikeMcQuaid I have seen the changes you made in the tests. Do you think this may help here?

It may do. There's less to go wrong, at least. You may need to rebase on `master`.

---

Just to be clear so there's no more unused work in here: I do increasingly think that a separate command is going to be a good idea for this that can be used by `test-bot`. `hub` could probably be used for this to simplify a lot of the logic if it was a separate command. Additionally, separating the `brew pull` logic to get the commits/bottle commits and the actual Bintray publish could also be useful, simplify this code and allow stuff to be reused.

iMichka:
Setting `git -C` should not be necessary because we `cd` to the right repository: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/pull.rb#L122-L127

Looks like the current path is something like `/private/tmp/homebrew-tests-20190402-52878-1fn5kvs/prefix` after that.  I am trying to find out how this folder is setup now.

iMichka:
Ok I understood what is going on. I was able to setup the tests on my computer and were able to reproduce.

The current test has this code:
```
CoreTap.instance.path.cd do
  system "git", "init"
  system "git", "checkout", "-b", "new-branch"
end
```

But the command that is tested is `brew pull https://github.com/Homebrew/brew/pull/1249`. We are pulling `brew` here, not `homebrew-core`.

The output was:

To stderr:
```
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
Warning: Current branch is : do you need to pull inside master?
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
Error: Patch failed to apply: aborted.
```

To stdout:
```
Fetching patch \nPatch: https://github.com/Homebrew/brew/pull/1249.patch\n==> Applying patch\n
```

So, the initial test had already two issues: the setup was wrong, and the output is full of issues. We are not really testing if `brew pull` is working correctly, we are barely trying to just run it and looks how it fails when the git context is wrong.

My change in the test: setup the right repository, and change the test check.
The new output will be:

To stderr:
```
fatal: Needed a single revision
fatal: ambiguous argument 'HEAD^..': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
fatal: You have nothing to amend.
Error: Failure while executing; `git commit --amend --signoff --allow-empty -q -m '\n'Closes\\ \\#1249.` exited with 128.
```

To stdout:
```
Fetching patch
Patch: https://github.com/Homebrew/brew/pull/1249.patch
==> Applying patch\nApplying: Revert \"Migrate a Formula to Multiple Versions: add docs.\"
Using index info to reconstruct a base tree...
A\tdocs/Migrate-A-Formula-To-Multiple-Versions.md
Falling back to patching base and 3-way merge...
No changes -- Patch already applied.
==> Patch closes issue #1249
```

Not better but at least the test passes now.

@MikeMcQuaid I would like to find a compromise with you: I sort of "need" that patch so I can run merge commits in linuxbrew-core again. This would speed up my linuxbrew-core merging again. Once this is merged, it should free up some time for me to work more on this.

Now that I have digged through this part of the code, I feel more confident about making changes in it. We definitively need better tests here. This can only be done if the code is split and simplified.

I think the first step after this will be to extract my `publish_and_check_bottles` function to a separate brew command. This would already remove a lot of code from this file, and allow to write separate tests for that part.

MikeMcQuaid:
> Now that I have digged through this part of the code, I feel more confident about making changes in it. We definitively need better tests here. This can only be done if the code is split and simplified.

If we definitely need better tests: they should be added _before_ rather than _after_ making a large change/refactoring like this. Tests do not achieve much by themselves but they do protect you against making changes to code which inadvertently breaks things.

> I think the first step after this will be to extract my `publish_and_check_bottles` function to a separate brew command. This would already remove a lot of code from this file, and allow to write separate tests for that part.

I'd suggest that be done as part of this PR instead.

> I would like to find a compromise with you: I sort of "need" that patch so I can run merge commits in linuxbrew-core again. This would speed up my linuxbrew-core merging again. Once this is merged, it should free up some time for me to work more on this.

I appreciate and am sorry from your pain but from my perspective this is a change that's performing a refactoring, adding a feature/functionality and requires modifying the existing, minimal integration test in order to pass. That makes me really uneasy about this not causing additional breakages and this command is extremely critical to all Homebrew formula CI and maintainer workflow.

iMichka:
Ok, I'll work on the refactoring and the addition of more tests. I hope it won't take me as much time as this first set of changes :)

iMichka:
I wrote some tests, and `brew style` complais about:

```
.linuxbrew/Homebrew/Library/Homebrew/test/bottle_publisher_spec.rb:5:1: C: Spec path should end with homebrew/bottle_publisher*_spec.rb.
describe Homebrew::BottlePublisher do
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

What does it want from me?

sjackman:
Based on the message, my best guess is that
`test/bottle_publisher_spec.rb`
should be
`test/homebrew/bottle_publisher_spec.rb`

iMichka:
> Based on the message, my best guess is that
> `test/bottle_publisher_spec.rb`
> should be
> `test/homebrew/bottle_publisher_spec.rb`

There are other tests doing exactly the same as I am doing. For example `search.spec` or `missing_formula_spec.rb`. There is not homebrew folder in the test folder, so I am not sure this is the right thing to do.

Library/Homebrew/bottle_publisher.rb
```diff
@@ -0,0 +1,164 @@
+# frozen_string_literal: true
+
+require "utils"
+require "formula_info"
+
+module Homebrew
```
MikeMcQuaid: Remove this `module` and it'll fix the tests.

iMichka:
Audit error is unrelated, see #6133

iMichka:
Ok, all green.

I'll squash the latest commits, to boil this down to:
> pull: allow to pull from merge commits
> pull_spec: fix test
> pull: extract bottle publishing code

Code coverage has increased. I am aware that the added tests are not really testing anything interesting, but at least the coverage increased in parts which were not tested before.

@MikeMcQuaid is this going in the right direction? I can add some more tests if you want, need to have a look what can still be done.

I can also extract more code from the main `pull` code, do we want to do this in the current PR? Just to know where we set the coverage threshold to get this merged.

Library/Homebrew/formula_info.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+# Formula info drawn from an external `brew info --json` call
+
+class FormulaInfoFromJson
```
MikeMcQuaid: Name the file matching the class (or vice-versa)

Library/Homebrew/formula_info.rb
```diff
@@ -0,0 +1,75 @@
...
+    PkgVersion.new(version(spec_type), revision)
+  end
+
+  def revision
+    info["revision"]
+  end
+end
+
+# Bottle info as used internally by pull, with alternate platform support.
+class BottleInfo
```
MikeMcQuaid: This could live in `brew pull` I think; there's not enough here to be useful standalone.

Library/Homebrew/test/bottle_publisher_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require "bottle_publisher"
+
+describe BottlePublisher do
+  let(:tap) { CoreTap.new }
+
+  changed_formulae_names = ["#{CoreTap.instance.name}/hello.rb"]
+  bottle_publisher = described_class.new(
```
MikeMcQuaid: Make this the `subject`: https://relishapp.com/rspec/rspec-core/v/2-8/docs/subject/explicit-subject

Library/Homebrew/test/bottle_publisher_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    CoreTap.instance, changed_formulae_names, "homebrew", false, false
+  )
+
+  describe "publish_and_check_bottles" do
+    it "fails if ENV[\"HOMEBREW_DISABLE_LOAD_FORMULA\"] is set to 1" do
```
MikeMcQuaid: Use `'` instead of `"` rather than escaping `"`

MikeMcQuaid:
@iMichka Yes, this seems like a big step in the right direction, nice work!

> I can add some more tests if you want, need to have a look what can still be done.

Yes, I think if we're going to be moving code around out of the command (which makes it more generic) that warrants at least a handful of unit (not integration) tests being added for the relevant classes. In this case I think `FormulaInfoFromJson` would be all that's needed.

> I can also extract more code from the main `pull` code, do we want to do this in the current PR? Just to know where we set the coverage threshold to get this merged.

I'm not so much looking just at a coverage number here but seeing a few unit tests for each added class. Once `FormulaInfoFromJson` has a few tests I think this is good to go 👍. I'd hold off making any other big changes for a new PR.

Library/Homebrew/test/bottle_publisher_spec.rb
iMichka: I will change this to: `it "fails if HOMEBREW_DISABLE_LOAD_FORMULA is set to 1" do`

Library/Homebrew/formula_info.rb
iMichka: The BottleInfo class is only called by the formula_info class once. The comment is misleading (and I am not really sure it is helpful):

> # Bottle info as used internally by pull, with alternate platform support.

I removed that comment

Library/Homebrew/formula_info.rb
```diff
@@ -0,0 +1,75 @@
...
```
MikeMcQuaid: This could be made a simple hash instead. If not, please put this class inside the `FormulaInfo` class so there's one class per file.

iMichka: I changed this to a simple hash.

iMichka:
> expect(described_class.lookup("hello").bottle_info["url"]).to match(HOMEBREW_BOTTLE_DOMAIN)

This supposes that there is alway a bottle for the `hello` formula. Is this an issue? If yes, we will need to stub that part in a way or another.

sjackman:
`brew tests` use a dummy loader that has no formulae, and it has no access to Homebrew/core. The formulae used by `brew tests` need to be stubbed out, using `stub_formula_loader`.
See for example https://github.com/Homebrew/brew/blob/e098c37dc3734dbe097ed9804a3a88737ff71add/Library/Homebrew/test/tab_spec.rb#L145

iMichka:
>  stub_formula_loader formula("foo") { url "foo-1.0" } 

This does not work, because I need a json file somewhere.

I ran `brew info --json=v1 hello` and put that in a file.
I also added a new constructor to the FormulaInfo class, which allows me to test all the methods of that class, as the json is now provided by the json file saved in fixtures.

The initial constructor is also tested, using the local `hello` formula.

I have issues with the style though: how do I disable the style check for the lines in formula_info_spec? The content are strings from the json file, so there is no need to run a style-check on these.

MikeMcQuaid:
> I ran `brew info --json=v1 hello` and put that in a file.

You definitely don't need a file on disk. You can stub the relevant behaviour instead. See the use of `expect` in the `Library/Homebrew/test/*_spec.rb` files.



> I have issues with the style though: how do I disable the style check for the lines in formula_info_spec? The content are strings from the json file, so there is no need to run a style-check on these.

You can do this but I think ideally you'd just follow the appropriate style.

iMichka:
I can write something like:
```
it "checks the bottle_info hash" do
    f = formula "foo" do
      url "foo-1.0"

      bottle do
        cellar(:any)
        sha256(TEST_SHA256 => Utils::Bottles.tag)
      end
    end
    stub_formula_loader f
    expect(
      described_class.lookup("foo"),
    ).to eq(nil)
  end
```

But `lookup("foo")` fails to find the formula (actually `brew info --json=v1` fails to find the formula). We are calling brew from the outside API here, so I guess that the foo formula is not loaded for that call. Ideally we should not call `brew info --json=v1`, but call the equivalent brew class/methods?

iMichka:
I have the feeling that I need to use `install_test_formula "testball"`. I simplified the formula_info_spec test to the bare minimum.

`described_class.lookup("testball")´ returns nil, which means that `Utils.popen_read(HOMEBREW_BREW_FILE, "info", "--json=v1", name)` in the lookup method failed.

Not sure why this is not working. I am really struggling to get a test setup here :(

MikeMcQuaid:
@iMichka I can take a look at this if that would be useful. Let me know if so (and hold off on making any more changes) and I'll give it a go.

MikeMcQuaid:
(i.e. I will try and fix this locally and commit a fix to this PR)

iMichka:
Yes please. I would just need an test example for `formula_info_spec`. Once the `look-up` method works, I can add the tests for all the other methods.

MikeMcQuaid:
@iMichka Fixed for you.

Library/Homebrew/test/formula_info_spec.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require "formula_info"
+require "global"
+
+describe FormulaInfo, :integration_test do
```
MikeMcQuaid: Can these not be done without an `integration_test`? `integration_test`s are really slow because they reinitialise Homebrew every time. If not: please strip this down to a single test (that could have multiple assertions on the `lookup` return value instead)

iMichka:
`install_test_formula` is part of `integration_test`. I can not get rid of that.

I regrouped all the tests in a single test, which brought that specific spec file run from 40 to 12 seconds on my computer.

Thana11:
null

ในวันที่ พฤ. 4 ก.ค. 2019 02:20 น. Michka Popoff <notifications@github.com>
เขียนว่า:

MikeMcQuaid:
@iMichka This finally looks good to go! Good perseverance with this. Merge it at a time you're comfortable and will be around for a while to help deal with e.g. any CI or maintainer hiccups that result. Thanks again for the great work ❤️

iMichka:
I'll merge this next week, as I won't have any time this weekend to follow up if this breaks anything.

sjackman:
Congratulations on getting this PR merged, Michka! 🏆

PR #5761, opened 2019-02-17
azure-pipelines.yml: Build and test Docker image

sjackman:

### DIFF
--- a/Dockerfile.test.yml
+++ b/Dockerfile.test.yml
@@ -1,7 +1,15 @@
 sut:
   build: .
-  command: "sudo -i -u linuxbrew /home/linuxbrew/.linuxbrew/bin/brew test-bot"
+  command:
+    - sh
+    - -xc
+    - |
+      sudo -E -u linuxbrew /home/linuxbrew/.linuxbrew/bin/brew test-bot
+      status=$$?
+      cp brew-test-bot.xml /tmp/test-bot/
+      exit $$status
   environment:
+    - HOME=/home/linuxbrew
     - BUILD_REASON
     - BUILD_REPOSITORY_URI
     - BUILD_SOURCEVERSION
--- a/Dockerfile.test.yml
+++ b/Dockerfile.test.yml
@@ -1,3 +1,11 @@
 sut:
   build: .
   command: "sudo -i -u linuxbrew /home/linuxbrew/.linuxbrew/bin/brew test-bot"
+  environment:
+    - BUILD_REASON
+    - BUILD_REPOSITORY_URI
+    - BUILD_SOURCEVERSION
+    - HOMEBREW_GITHUB_API_TOKEN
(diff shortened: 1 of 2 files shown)


sjackman:
Moving `HOMEBREW_REPOSITORY` to `…/Homebrew` was need to prevent this `brew doctor` error:
```
Warning: You have multiple Cellars.
You should delete /home/linuxbrew/.linuxbrew/Homebrew/Cellar:
  rm -rf /home/linuxbrew/.linuxbrew/Homebrew/Cellar
```

MikeMcQuaid:
@sjackman 

> Successful in 41m — Linux succeeded
> Successful in 11m — macOS succeeded

I'm a bit concerned about the massive increase in build time (23m to 41m). Anything we can do to help speed this up a bit? If not let's maybe block this on making `brew test-bot` run more tests on macOS rather than Linux.

sjackman:
Yep. Running `brew test-bot` both outside the container and inside the container is redundant. We can run it only within the container.

```
Run brew test-bo 22m 37s
Build and test Docker image 18m 22s
```

sjackman:
Removed the duplication. Ready for re-review.

MikeMcQuaid:
@sjackman Oh, much nicer. A few things:

> ##[warning]No test result files matching brew-test-bot.xml were found.

Probably need to copy that out the docker container. Just for future-proofing: may also want to copy the coverage files too (see "Publish brew tests code coverage" on the macOS build and https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L1030-L1031 e.g. the `coverage` directory in the build root).

Also need to ensure that the `TF_BUILD` variable is passed through: https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L1575 along with other Azure Pipelines environment variables e.g. https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L407-L408 to ensure the correct code is being tested.

Also, in an ideal world if it's technically possible it'd be nice to split the "build the docker image" and "run test-bot inside the docker image" into two separate steps so their output can be differentiated.

sjackman:
Good grief that was surprisingly difficult to get the environment variables exported through Docker and sudo to test-bot, and get the `brew-test-bot.xml` out of Docker back to Azure.

@MikeMcQuaid Ready for re-review.

MikeMcQuaid:
Nice work here @sjackman! Pretty cool to see us actually building and testing the docker image here. I think we should consider making the Docker image "official" in a future version; I suspect it'll get widely used.

sjackman:
For installing packages required by CI to test a package, when the versions provided by the host package manager (likely `apt-get`) are not new enough, using Linuxbrew is a good alternative to using Ubuntu PPAs (Personal Package Archives).

Issue #5730, opened 2019-02-13
Cleanup ARGV usage


MikeMcQuaid:
Now we have [a proper argument parser](https://github.com/Homebrew/brew/issues/1860) (thanks @GauthamGoli!) we should replace all uses of ARGV in the Homebrew/brew codebase with calls to `Homebrew.args` instead.

[extend/ARGV.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/ARGV.rb) should be moved into https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cli/args.rb where needed or removed where not. This may warrant changes to formulae and `odeprecated`.

Note `formula_installer.rb` and `build.rb` should be handled last as they read from and modify ARGV: https://github.com/Homebrew/brew/blob/b4f73e61649fcfc5aaa779c311e2514619ce01e7/Library/Homebrew/formula_installer.rb#L678-L751

No need to say "I'm planning on working on this" or similar in this issue but feel free to open a PR that doesn't work (yet!) and ask for help.

BenMusch:
This seems like something I'd be happy to help out with, though it seems like an incremental approach to this is best (rather than getting rid of it all in one PR)

Would this be a reasonable way to do that?

1. Add a nested `Homebrew::CLI::Parser::ArgvExtension` module
2. Add `ARGV.extend(Homebrew::CLI::Parser::ArgvExtension)` to `global.rb` (but do not remove the other `ARGV.extend` call)
3. From this point, extract logic to the new `ArgvExtension` module when needed, and replace `ARGV.*` methods with `Homebrew.args.*` when possible

And then after repeating step 3 on all the various `ARGV` calls, it would eventually be possible to remove the old `HomebrewArgvExtension`

MikeMcQuaid:
@BenMusch Sounds great!

> For example, we could pre-compute the named arguments and add it as a field to `Homebrew.args`, and then delete `ARGV.named`. But doing this for _every_ ARGV method would make the `Homebrew.args` struct somewhat bloated (which might still be worth it!)

I think we can avoid precomputing but we could cache it intelligently.

Part of the motivation for this is that I don't think we will need to do it for every ARGV method; some of them will be able to be removed or replaced with better APIs. Any that aren't used in Homebrew/brew or Homebrew/homebrew-core can be considered internal APIs and removed. Even among those it may be they can be refactored so they aren't needed.

[...]

PR #5762, opened 2019-02-17
ARGV cleanup: Deprecate ARGV methods that are never used internally

BenMusch:

-----

Addresses part of https://github.com/Homebrew/brew/issues/5730

This PR deletes `ARGV` methods which are never used internally in `ARGV` (for example, it keeps `build_head?` because it is used in `build_stable?`, and refactoring methods which have internal usage is a bit more complex of a question in terms of how it is refactored. Will definitely look into cleaning that up later, but wanted to open this PR for all the work which could be done without too much thought


Library/Homebrew/test/formula_installer_spec.rb
```diff
@@ -87,10 +87,8 @@ def temporary_install(formula)
   end
 
   specify "Formula is not poured from bottle when compiler specified" do
-    expect(ARGV.cc).to be nil
-
-    cc_arg = "--cc=clang"
-    ARGV << cc_arg
+    expect(Homebrew.args.cc).to be nil
+    allow(Homebrew.args).to receive(:cc).and_return("clang")
```
BenMusch: Just wanna make sure: This is a valid replacement in tests for adding values to `ARGV`, right?

Library/Homebrew/test/sandbox_spec.rb
```diff
@@ -47,7 +48,7 @@
     end
 
     it "ignores bogus Python error" do
-      ENV["HOMEBREW_VERBOSE"] = "1"
```
BenMusch: Same here: Wanna make sure it's okay to replace env var values with Homebrew.args stubs

BenMusch:
Not really sure I follow why tests are failing (`brew tests` passes locally), but it definitely seems like it may be related to my changes. Any idea what's going on?

Library/Homebrew/test/formula_installer_spec.rb
MikeMcQuaid: Yep 👍

Library/Homebrew/test/sandbox_spec.rb
MikeMcQuaid: Yep 👍

MikeMcQuaid:
> Looks like brew test bot might have to be updated? https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb#L274 If so, are there any other packages this might affect?

Good catch! The following will likely need updates:
https://github.com/Homebrew/homebrew-test-bot
https://github.com/Homebrew/homebrew-services
https://github.com/Homebrew/homebrew-bundle

scpeters:
Also https://github.com/Homebrew/homebrew-livecheck

Perhaps we could deprecate these old methods to give downstream taps a chance to notice the changes (see https://github.com/Homebrew/brew/pull/5477 for an example). Then we could merge this without being blocked or breaking homebrew commands in external taps.

MikeMcQuaid:
> Perhaps we could deprecate these old methods to give downstream taps a chance to notice the changes (see #5477 for an example).

This seems like a good call 👍 . For those that we notice have been used anywhere outside Homebrew/brew an `odeprecated` and `odisabled` make sense before their removal. For consistency they could also call back to `Homebrew.args.foo` rather than querying `ARGV.foo` directly.

BenMusch:
Looks like the deprecations are still failing the build -- should I be doing this differently? https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=4180

jonchang:
> Looks like the deprecations are still failing the build -- should I be doing this differently? https://dev.azure.com/Homebrew/Homebrew/_build/results?buildId=4180

The issue is in https://github.com/Homebrew/homebrew-test-bot/, not your pull request.

I believe you'll have to do the pull request in several parts: first to add the new API, then a pull request to test-bot to use the new API, then the deprecations in a final pull request to deprecate the old API. Probably a pull request to homebrew-core fixing all uses of the old API should be in there somewhere as well.

Longer term we should probably look at moving the homebrew/test-bot code into homebrew/brew.

jonchang:
Actually it looks like you're not adding any new APIs here, just deprecating old ones. So a pull request to homebrew-test-bot fixing its calls should be sufficient.

BenMusch:
Ah thanks @jonchang! Didn't see the line about throwing an error when `homebrew_developer?` is true. Opened a PR to change the test-bot: https://github.com/Homebrew/homebrew-test-bot/pull/238

MikeMcQuaid:
@BenMusch Some failing tests but not `brew test-bot` any more 🎉

MikeMcQuaid:
Cross-posting from https://github.com/Homebrew/homebrew-test-bot/pull/240:

> I would suggest making use of `ARGV.freeze` to catch these cases (both here and in Homebrew/brew) and fix/refactor them before any more `ARGV` to `Homebrew.args` conversion is done.

BenMusch:
@MikeMcQuaid should we also find a way to freeze `ENV`, since Homebrew.args values are also controlled by that?

MikeMcQuaid:
> @MikeMcQuaid should we also find a way to prevent modification of `ENV` (hopefully without interfering with anything non-argv-related values), since Homebrew.args values are also controlled by that?

@BenMusch Good question. I don't believe there are (m)any cases where `ENV` is overridden and the expectation is that `ARGV.whatever?` will change so we're fine from that perspective. Also, we relatively often need to modify `ENV` for subprocesses so I think we're left with what we have for now (without introducing a new API for ENV which also updates CLI::Parser which I think would be overkill for now).

BenMusch:
Sounds good. Just posting this as a heads up: probably won't be able to make progress on the PR for a week or so due to some schoolwork that's come up. But definitely intend on getting back to it once I'm free

MikeMcQuaid:
No worries, thanks @BenMusch!

BenMusch:
Just a heads up: Busier with schoolwork than I thought so it might not be a while until I pick this up again. Will check-in periodically to keep the PR up-to-date with `master`, and anyone who wants to take over the work on this should feel welcome to if my timeliness becomes an issue!

MikeMcQuaid:
@BenMusch Would you be able to extract the minimal version that still passes tests (perhaps into a new PR)? Thanks!

BenMusch:
@MikeMcQuaid Definitely, would a minimal version of this be: not deprecating existing methods, and only changing the `ARGV.*` to `Homebrew.args.*` when it doesn't result in test failures?

scpeters:
> would a minimal version of this be: not deprecating existing methods, and only changing the ARGV.* to Homebrew.args.* when it doesn't result in test failures?

that sounds good to me

Issue #5743, opened 2019-02-15
More responsive outdated & upgrade commands through scheduled tasks and caching


zbeekman:
# A detailed description of the proposed feature

I’m interested in adding the ability for Homebrew users to enable:

1. Background/scheduled updates (makes running brew commands faster if `brew update` happens intermittently in the background)
2. Background pre-fetching of bottles for outdated formulae
3. Background unattended upgrades for outdated formulae
4. Automated caching of outdated formula on disk. (so that you can have login messages like those present in many linux distros telling you how many/which packages are out of date)

These would all be utilizing `launchctl` plists, possibly through `brew services`. I have `launchctl` plist files that I use on my machine that implement 1, 2, and 4 at https://github.com/zbeekman/launchctl-tasks. Should we decide to add these capabilities, I would like feedback on some implementation details:

1. The service label. Currently I think `sh.brew.<name>` is a good label, but I don’t know if this is appropriately canonical and if we’ve used a different label naming scheme for services in the past
2. Whether these should be implemented as
   - a) Formula that just configure and install the plist files, that could be added to `homebrew-core`,
   - b) as functionality within `Homebrew/brew` that users would then opt in to enable, or
   - c) as a separate Tap which can define additional commands
3. Where the scheduled tasks should write logs & data about what is upgraded, outdated and fetched, and if running brew commands can and should update & cleanup this data. E.g., on my system I print the list of outdated packages upon launching a shell, however, this list is only managed by the the scheduled tasks I have running and `brew alias`es I have that will update and clean it up when I run e.g, `brew upgrade <pkg>`
4. How the UI should work. Do users set an environment variable to enable this? Do they run `brew services start brew-unattended-upgrades`? Do they run a special new command, like `brew unattended-upgrades-on`?

# The motivation for the feature

More responsive and seamless user experience. More information available to users and sys admins.

# How the feature would be relevant to at least 90% of Homebrew users

Prefetching outdated bottles will make `brew upgrade` faster. Background/scheduled `brew update` will prevent `brew update` from running when the user invokes an arbitrary command, or at least reduce the frequency of this happening. This in turn will make the command run significantly faster, especially if there is a slow internet connection, etc. Unattended upgrades will allow users who want the latest and greatest version of a package to always have it, without manual effort. Logging and caching outdated packages (and updating this when they are upgraded or uninstalled) will allow the user to know which installed packages are out of date without running `brew outdated` which can be slow. (This could be done for `mas` app-store software and casks as well.)

[...]

PR #5766, opened 2019-02-18
tap: skip auto-update if called with no args

scpeters:

-----

Any invocation of `brew tap` may currently lead to an auto-update, even though `brew tap` with no arguments simply lists the current taps on the local machine. It was noted recently (https://github.com/Homebrew/brew/issues/5743#issuecomment-464645856):

> "In the "no argument `brew tap` case": we could handle that more intelligently and not run `brew tap`."

So that's what I've done here. It makes it run faster, and also avoids the auto-update text being printed by this command that isn't part of the list of taps.

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -414,7 +414,8 @@ update-preinstall() {
   [[ -z "$HOMEBREW_AUTO_UPDATE_CHECKED" ]] || return
   [[ -z "$HOMEBREW_UPDATE_PREINSTALL" ]] || return
 
-  if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || "$HOMEBREW_COMMAND" = "tap" ||
+  if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" ||
+        "$HOMEBREW_COMMAND" = "tap" && $HOMEBREW_ARG_COUNT -gt 1 ||
         "$HOMEBREW_CASK_COMMAND" = "install" || "$HOMEBREW_CASK_COMMAND" = "upgrade" ]]
   then
     export HOMEBREW_AUTO_UPDATING="1"


MikeMcQuaid:
Nice work thanks @scpeters!

MikeMcQuaid:
`shellcheck` is complaining on Linux.

scpeters:
> `shellcheck` is complaining on Linux.

I've tested 013dfda on macOS and linux and it looks ok

scpeters:
no, I think the logic is backwards

scpeters:
how do I test auto-updating?

scpeters:
I'm testing with `HOMBREW_VERBOSE=1` and that seems to help

scpeters:
I fixed the logic in ec9b7d5

`HOMEBREW_ARG_COUNT` is never `0`, so it was just always skipping the auto-update for any tap command, with or without arguments.

`brew tap`: `HOMEBREW_ARG_COUNT=1`

`brew tap fake`: `HOMEBREW_ARG_COUNT=2`

So I changed the logic to only auto-update if `HOMEBREW_ARG_COUNT -gt 1`

Library/Homebrew/brew.sh
```diff
@@ -414,7 +414,8 @@ update-preinstall() {
   [[ -z "$HOMEBREW_AUTO_UPDATE_CHECKED" ]] || return
   [[ -z "$HOMEBREW_UPDATE_PREINSTALL" ]] || return
 
-  if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" || "$HOMEBREW_COMMAND" = "tap" ||
+  if [[ "$HOMEBREW_COMMAND" = "install" || "$HOMEBREW_COMMAND" = "upgrade" ||
+        ( "$HOMEBREW_COMMAND" = "tap" && $HOMEBREW_ARG_COUNT -gt 1 ) ||
```
vitorgalvao: Doesn’t the parenthesis create a new subshell, or is the behaviour different inside `test`? I think this should be `[[ "$HOMEBREW_COMMAND" = "tap" && $HOMEBREW_ARG_COUNT -gt 1 ]]`.

vitorgalvao:
```suggestion
        { "$HOMEBREW_COMMAND" = "tap" && $HOMEBREW_ARG_COUNT -gt 1 } ||
```

For [grouping without invoking a subshell](https://stackoverflow.com/a/29261377/1661012).

vitorgalvao:
```suggestion
        { "$HOMEBREW_COMMAND" = "tap" && $HOMEBREW_ARG_COUNT -gt 1 } ||
```

[Grouping without invoking a subshell](https://stackoverflow.com/a/29261377/1661012).

scpeters: Thanks for the suggestion, but it gave me a syntax error when I tried it. I've instead just removed the `()` in 45b576c313c6f71b5be54cb86349b3a87f4805dc after confirming that `&&` has higher precedence than `||`, so I think it has the same effect

* http://tldp.org/LDP/abs/html/opprecedence.html

zbeekman: I think you need a semicolon in there with the command grouping.

PR #5776, opened 2019-02-20
tests: Set environment variable USER if not set

sjackman:
These two tests fail if the enivornment variable USER is not set:

```
./test/utils/user_spec.rb:6 # User should eq nil
./test/utils/user_spec.rb:22 # User#gui? when the current user is in a console session gui? should equal true
```

-----

The environment variable `USER` is not set in Docker by default.

### DIFF
--- a/Library/Homebrew/dev-cmd/tests.rb
+++ b/Library/Homebrew/dev-cmd/tests.rb
@@ -49,6 +49,8 @@ module Homebrew
       ENV["HOMEBREW_TEST_GENERIC_OS"] = "1" if args.generic?
       ENV["HOMEBREW_TEST_ONLINE"] = "1" if args.online?
 
+      ENV["USER"] ||= Utils.popen_read("id", "-nu").chomp
+
       # Avoid local configuration messing with tests e.g. git being configured
       # to use GPG to sign by default
       ENV["HOME"] = "#{HOMEBREW_LIBRARY_PATH}/test"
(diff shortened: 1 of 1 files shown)


Library/Homebrew/dev-cmd/tests.rb
```diff
@@ -49,6 +49,8 @@ def tests
       ENV["HOMEBREW_TEST_GENERIC_OS"] = "1" if args.generic?
       ENV["HOMEBREW_TEST_ONLINE"] = "1" if args.online?
 
+      ENV["USER"] ||= Utils.popen_read("id", "-nu").chomp
```
reitermarkus: Use `system_command!("id", args: ["-nu"]).stdout.chomp` instead.

sjackman: `system_command` doesn't observe `HOMEBREW_STDERR`, so I prefer to use `Utils.popen_read`.
https://github.com/Homebrew/brew/blob/4d68bee1be06625251afcf340030caffef1ec4cd/Library/Homebrew/utils/popen.rb#L31
What is the advantage of `system_command` over `Utils.popen_read`?

reitermarkus: > What is the advantage of `system_command` over `Utils.popen_read`?

`system_command!` raises an exception when the command fails, which also includes the content of `stderr`.

The goal would be to replace all the `popen*` commands with `system_command`.

sjackman: > `system_command!` raises an exception when the command fails, which also includes the content of `stderr`.

Ah, that's a nice feature!

`system_command!` is more heavyweight (using `popen3` and `select` to poll multiple file descriptors) than `Utils.popen_read`. Perhaps something to keep in mind for larger I/O jobs.

MikeMcQuaid: > The goal would be to replace all the `popen*` commands with `system_command`.

@reitermarkus Would be worth opening a `help wanted` issue for that?

MikeMcQuaid:
Makes sense, thanks @sjackman!

PR #5785, opened 2019-02-22
fish 2.x support for `brew shellenv`

ntkme:
This PR makes a small amendment to #5773.

It replaces fish 3.x '||' operator with fish 2.x '; or' syntax so that it can work for both fish 2.x and 3.x.


reitermarkus:
Thanks, @ntkme!

PR #5790, opened 2019-02-22
Set HOMEBREW_PROCESSOR to arm64v8 for ARM64 [Linux]

sjackman:

-----

The value of `HOMEBREW_PROCESSOR` is used for the bottle tag. Options are…

1. `aarch64` (like GNU)
2. `arm64` (like Debian)
3. `armv8` (like Homebrew ARM32 which uses `armv6`)
4. `arm64v8` (like Docker)

### DIFF
--- a/Library/Homebrew/brew.sh
+++ b/Library/Homebrew/brew.sh
@@ -129,6 +129,10 @@ then
   HOMEBREW_SYSTEM_TEMP="/private/tmp"
 else
   HOMEBREW_PROCESSOR="$(uname -m)"
+  if [[ "$HOMEBREW_PROCESSOR" == "aarch64" ]]
+  then
+    HOMEBREW_PROCESSOR="arm64v8"
+  fi
   HOMEBREW_PRODUCT="${HOMEBREW_SYSTEM}brew"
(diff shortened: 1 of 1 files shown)


zbeekman:
> The value of `HOMEBREW_PROCESSOR` is used for the bottle tag. Options are…
> 
> 1. `aarch64` (like GNU)
> 2. `arm64` (like Debian)
> 3. `armv8` (like Homebrew ARM32 which uses `armv6`)
> 4. `arm64v8` (like Docker)

4, `arm64v8` is my favorite, since it is unambiguous, and somewhat future proof, if ARM decide to make another 64 bit version of their ISA. My second favorite is 1, `aarch64`, just because that's what `uname` returns, so its consistent. I think 1 is also probably more common.

zbeekman:
Actually, scratch that. If we're just pulling the value from `uname -m` then I think we should keep it as is, `aarch64`. But I'm open to persuasion.

MikeMcQuaid:
> The value of `HOMEBREW_PROCESSOR` is used for the bottle tag.

If we do tweak this: could it be tweaked at the bottle tag level in Ruby?

sjackman:
Let's just stick with `aarch64` then. It was my least favourite option, but it's growing on me. It's certainly the easiest.

PR #5793, opened 2019-02-22
install_symlink_p: Fix when dest includes a symlink

sjackman:
install_symlink_p does not work as intended when dst includes a symlink in its path.
relative_path_from requires that both src and dst be real paths without symlinks.

From https://ruby-doc.org/stdlib-2.3.7/libdoc/pathname/rdoc/Pathname.html#method-i-relative_path_from

> This method doesn't access the filesystem. It assumes no symlinks.

-----

### DIFF
--- a/Library/Homebrew/extend/pathname.rb
+++ b/Library/Homebrew/extend/pathname.rb
@@ -135,10 +135,11 @@ class Pathname
   end
 
   def install_symlink_p(src, new_basename)
-    src = Pathname(src).expand_path(self)
-    dst = join(new_basename)
     mkpath
-    FileUtils.ln_sf(src.relative_path_from(dst.parent), dst)
+    dstdir = realpath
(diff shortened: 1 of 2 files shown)


sjackman:
I've marked this as `do not merge` because I'd like to test it some more first.

MikeMcQuaid:
Suggestion: write a test to reproduce the issue you saw.

sjackman:
I've been using this PR myself for a while, and I'm happy with it. I'll write a test.

sjackman:
I've added a test.

MikeMcQuaid:
Thanks again @sjackman!

sjackman:
Thanks for merging, Mike!
This PR fixed an issue for me when `/usr/local/Cellar` was a symlink to a directory found in my home directory. A non-standard unsupported setup to be sure, but revealed this bug.

PR #5801, opened 2019-02-25
vendor-install: Add portable-ruby for aarch64 (ARM64)

sjackman:

### DIFF
--- a/Library/Homebrew/cmd/vendor-install.sh
+++ b/Library/Homebrew/cmd/vendor-install.sh
@@ -29,6 +29,11 @@ then
       ruby_URL2="https://github.com/Homebrew/homebrew-portable-ruby/releases/download/2.3.7/portable-ruby-2.3.7.x86_64_linux.bottle.tar.gz"
       ruby_SHA="9df214085a0e566a580eea3dd9eab14a2a94930ff74fbf97fb1284e905c8921d"
       ;;
+    aarch64)
+      ruby_URL="$HOMEBREW_BOTTLE_DOMAIN/bottles-portable-ruby/portable-ruby-2.3.7.aarch64_linux.bottle.tar.gz"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @sjackman!

zbeekman:
Just, catching up here, but did you resolve the pouring of bottles issue? Or is aarch64 not affected, because there are no bottles for it yet?

sjackman:
Thanks for merging, Mike!

sjackman:
> Just, catching up here, but did you resolve the pouring of bottles issue?

Is that the UID/GID issue with `tar`? If so, it only affects pouring bottles while root, which should only affect Docker and CI.

zbeekman:
> Is that the UID/GID issue with `tar`? If so, it only affects pouring bottles while root, which should only affect Docker and CI.

No I was referring to the fact that my adventures with portable ruby on x86_64 Linux last Friday were unsuccessful because `brew portable-package portable-ruby` was pouring bottled glibc, gcc, etc.: https://github.com/Homebrew/homebrew-portable-ruby/issues/74

I was curious if that problem hit you at all, however, I think I answered my own question: on aarch64 there are no bottles (until you added portable ruby, etc.) so when you built portable-ruby there would have been no way for homebrew to have poured bottles for you.

sjackman:
You're right. That issue of pulling in `glibc` and `gcc` did not occur on AArch64, where there are no bottles. To troubleshoot the issue on `x86_64_linux`, run `brew deps -n --include-build portable-ruby` and install the dependencies one-by-one with `brew install -s foo`. None of them should install `glibc` or `gcc`. When you find one that does, you've found the bug.

It may be a bug in `brew install` itself, that `brew install -s a b c d e` is not equivalent to `for i in a b c d e; do brew install -s $i; done`.

zbeekman:
> When you find one that does, you've found the bug.

:+1:



> It may be a bug in `brew install` itself, that `brew install -s a b c d e` is not equivalent to `for i in a b c d e; do brew install -s $i; done`.

I tried both methods. I'll go through one at a time and see who the culprit is. I know it's not `portable-zlib`.

PR #5805, opened 2019-02-25
Fix bug in CLI arg/env var prioritization + corresponding tests

BenMusch:

-----

Fixes bug in https://github.com/Homebrew/brew/pull/5738
Fixes https://github.com/Homebrew/brew/issues/5736

Explanation of what happened:

 - When I changed the names of the :env_var value to :env, I left one unchanged
 - The tests were passingly misleadingly because the stubbing order made it so that all env vars were actually ignored (and one of my tests was testing the non-env-var behavior due to my mistake in the test)


MikeMcQuaid:
Thanks again @BenMusch!

PR #5808, opened 2019-02-25
pull_spec: pull #2 so it works on linux too

scpeters:

-----

This is a follow-up to this conversation thread: https://github.com/Homebrew/brew/pull/5726#discussion_r259942109

I think the test expectation was passing on macOS but failing on Linux because Homebrew/homebrew-core#1  is a pull request but Homebrew/linuxbrew-core#1 is an issue. I haven't tested it, but I think changing it to `2` will work because Homebrew/homebrew-core#2 and Homebrew/linuxbrew-core#2 are both pull requests.


scpeters:
Ok, it didn't work; I'll stop being lazy and test it on my own machine instead of submitting a PR.

PR #5809, opened 2019-02-25
pull: fix tap name for linuxbrew

iMichka:

-----

### DIFF
--- a/Library/Homebrew/dev-cmd/pull.rb
+++ b/Library/Homebrew/dev-cmd/pull.rb
@@ -109,10 +109,10 @@ module Homebrew
       elsif (api_match = arg.match HOMEBREW_PULL_API_REGEX)
         _, user, repo, issue = *api_match
         url = "https://github.com/#{user}/#{repo}/pull/#{issue}"
-        tap = Tap.fetch(user, repo) if repo.start_with?("homebrew-")
+        tap = Tap.fetch(user, repo) if repo.match?(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX)
       elsif (url_match = arg.match HOMEBREW_PULL_OR_COMMIT_URL_REGEX)
         url, user, repo, issue = *url_match
-        tap = Tap.fetch(user, repo) if repo.start_with?("homebrew-")
+        tap = Tap.fetch(user, repo) if repo.match?(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX)
       else
         odie "Not a GitHub pull request or commit: #{arg}"
       end
--- a/Library/Homebrew/tap.rb
+++ b/Library/Homebrew/tap.rb
@@ -26,7 +26,7 @@ class Tap
 
     # We special case homebrew and linuxbrew so that users don't have to shift in a terminal.
     user = user.capitalize if ["homebrew", "linuxbrew"].include? user
-    repo = repo.delete_prefix "homebrew-"
+    repo = repo.sub(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX, "")
 
     return CoreTap.instance if ["Homebrew", "Linuxbrew"].include?(user) && ["core", "homebrew"].include?(repo)
 
--- a/Library/Homebrew/tap_constants.rb
+++ b/Library/Homebrew/tap_constants.rb
@@ -8,3 +8,4 @@ HOMEBREW_TAP_DIR_REGEX = %r{#{Regexp.escape(HOMEBREW_LIBRARY)}/Taps/(?<user>[\w-
 HOMEBREW_TAP_PATH_REGEX = Regexp.new(HOMEBREW_TAP_DIR_REGEX.source + %r{(?:/.*)?$}.source)
 # match official taps' casks, e.g. homebrew/cask/somecask or homebrew/cask-versions/somecask
 HOMEBREW_CASK_TAP_CASK_REGEX = %r{^(?:([Cc]askroom)/(cask|versions)|(homebrew)/(cask|cask-[\w-]+))/([\w+-.]+)$}.freeze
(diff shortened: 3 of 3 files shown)


iMichka:
I think this is just the beginning ... :)

I am a little bit unhappy about this part of the code now: the user has a different value in the `url` compared to the one in the `user` variable. Maybe the old `user` from the url should not be called `user` anymore.

I did not integrate these changes into #5758, so that we keep the matters separated. But of course #5758 will only work with these changes.

Library/Homebrew/dev-cmd/pull.rb
```diff
@@ -109,10 +109,10 @@ def pull
       elsif (api_match = arg.match HOMEBREW_PULL_API_REGEX)
         _, user, repo, issue = *api_match
         url = "https://github.com/#{user}/#{repo}/pull/#{issue}"
-        tap = Tap.fetch(user, repo) if repo.start_with?("homebrew-")
+        tap = Tap.fetch(user, repo) if repo.start_with?("homebrew-", "linuxbrew-")
```
colindean: These could probably be constants now, perhaps

```
OFFICIAL_REPO_PREFIXES = ["homebrew","linuxbrew"]
OFFICIAL_REPO_PREFIXES_DASHED = OFFICIAL_REPO_PREFIXES.map {|r| r + '-'}
```

or something like that.

It makes both this line and the line below read a little better and single-sources the list.

Library/Homebrew/tap.rb
```diff
@@ -27,6 +27,7 @@ def self.fetch(*args)
     # We special case homebrew and linuxbrew so that users don't have to shift in a terminal.
     user = user.capitalize if ["homebrew", "linuxbrew"].include? user
     repo = repo.delete_prefix "homebrew-"
+    repo = repo.delete_prefix "linuxbrew-"
```
colindean: These two lines and the line above it become consumers of that constant.

sjackman:
I agree with @colindean about extracting these two uses into a constant (local to this module). Other than that, 👍. Thanks, Michka.

sjackman:
A regex seems better suited here.
```ruby
HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX = /^homebrew-|^linuxbrew-/;
repo.match?(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX)
repo.sub(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX, "")
```

colindean:
> A regex seems better suited here.

Yes, but it'd be ideal to omit the dashes somehow in order to reuse the constant [here](https://github.com/Homebrew/brew/pull/5809/files#diff-5f336a727e42d98a4d5125c8972b6c07R28) and a few other places I've seen that now have `["homebrew","linuxbrew"]`.

sjackman:
The simplest fix is…
```
HOMEBREW_OFFICIAL_REPO_PREFIXES = ["homebrew", "linuxbrew"].freeze
HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX = /^homebrew-|^linuxbrew-/;
repo.match?(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX)
repo.sub(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX, "")
```
disregarding the obvious duplication. I'm sure `HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX` could be constructed programatically from `HOMEBREW_OFFICIAL_REPO_PREFIXES` if we wanted to.

iMichka:
I liked the initial version better. Not sure why it has to be a regex. The file is called `tap_constants`, not `tap_regexes`. I find also the "normal" code is more readable than a regex, so I am not sure we should artificially introduce one when not needed just to be consistent with the bunch of other constants in that file.

sjackman:
In fact all five of the existing constants in that file are regexes.

My motivation was that I find
```ruby
HOMEBREW_OFFICIAL_REPO_PREFIXES_DASHED.map { |prefix| repo = repo.delete_prefix prefix }
```
less readable than
```ruby
repo.sub(HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX, "")
```

iMichka:
> In fact all five of the existing constants in that file are regexes.

I can add 10 more other constants which are not regexes  :)

> My motivation was that I find ... less readable than ...

Agreed. I'll change the second constant then. I'm fine with the duplication.

iMichka:
Actually we do not need `HOMEBREW_OFFICIAL_REPO_PREFIXES = ["homebrew", "linuxbrew"].freeze` so I will just omit it. It can be added later if needed.

Library/Homebrew/tap_constants.rb
```diff
@@ -8,3 +8,4 @@
 HOMEBREW_TAP_PATH_REGEX = Regexp.new(HOMEBREW_TAP_DIR_REGEX.source + %r{(?:/.*)?$}.source)
 # match official taps' casks, e.g. homebrew/cask/somecask or homebrew/cask-versions/somecask
 HOMEBREW_CASK_TAP_CASK_REGEX = %r{^(?:([Cc]askroom)/(cask|versions)|(homebrew)/(cask|cask-[\w-]+))/([\w+-.]+)$}.freeze
+HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX = /^homebrew-|^linuxbrew-/.freeze
```
MikeMcQuaid: Needs a newline on this file.

MikeMcQuaid:
```suggestion
HOMEBREW_OFFICIAL_REPO_PREFIXES_REGEX = /^(home|linux)brew-/.freeze
```

sjackman:
Thanks, Michka!

PR #5827, opened 2019-03-02
Allow head in new formulas

fxcoudert:
I would like us to allow `head` in new formulas. The current policy does not make much sense, since people are allowed to add `head` once the formula has been accepted into Homebrew core. We spend a lot of time asking people to remove `head` from the new formula PRs, for actually little effect.

(I can't run `brew style` and `brew test` locally, because of the Mojave Ruby issues.)

### DIFF
--- a/Library/Homebrew/dev-cmd/audit.rb
+++ b/Library/Homebrew/dev-cmd/audit.rb
@@ -640,18 +640,14 @@ module Homebrew
 
       problem "Formulae should not have a `devel` spec" if formula.devel
 
-      if formula.head
+      if formula.head && @versioned_formula
         head_spec_message = "Formulae should not have a `HEAD` spec"
-        if @new_formula
-          new_formula_problem head_spec_message
-        elsif @versioned_formula
-          versioned_head_spec = %w[
-            bash-completion@2
-            imagemagick@6
-            python@2
-          ]
-          problem head_spec_message unless versioned_head_spec.include?(formula.name)
-        end
+        versioned_head_spec = %w[
+          bash-completion@2
+          imagemagick@6
+          python@2
+        ]
+        problem head_spec_message unless versioned_head_spec.include?(formula.name)
       end
 
       throttled = %w[


chdiza:
> The current policy does not make much sense

It makes the same amount of sense as the Great Options Removal.  HEAD is not tested by CI, people open issues, can just do it in your own tap, etc. etc.

fxcoudert:
There is one major upside of HEAD:  The presence of HEAD allows upstream projects (or avid users) to test newer code and its integration into Homebrew, meaning that bugs can be discovered and fixed before a new release.

And we don't support HEAD issues, i.e. we will simply close issues related to HEAD builds: if it works, fine, otherwise you're on your own.

vitorgalvao:
> It makes the same amount of sense as the Great Options Removal.

But people are not allowed to submit options after a cask is accepted. With `head`, [on the other hand](https://github.com/Homebrew/brew/pull/5827#issue-257655299):

> people are allowed to add `head` once the formula has been accepted into Homebrew core.

For both cases to be the same, `head` would need to *never* be accepted. Since it is, removing this rule makes sense to me.

sjackman:
I like having `head` for the reasons that @fxcoudert cites, primarily the ability for upstream to test a patch with users before tagging a new release by asking them to `brew install --HEAD foo`.

If it doesn't already `brew doctor` should complain loudly about any `HEAD` formulae that are installed. If a `brew install` fails and any `HEAD` formulae are installed, rather than giving instructions to submit an issue, it should give instructions to reinstall those formulae with a stable release.

javian:
I would agree with this policy in general but why would we not allow it in versioned formulae ?

fxcoudert:
I don't have an opinion on versioned formulas. They're supposed to be relatively stable ones, no, so head makes less sense?

zbeekman:
> I would agree with this policy in general but why would we not allow it in versioned formulae ?

Well aren't versioned formulae typically duplicates of the latest stable release formula which is in core (and which can have HEAD)? So including HEAD in a versioned formula would, effectively, be including a development/unstable build from a historical/old version, no? I think, in general, not having HEAD in versioned formulae makes sense, but I'm open to being convinced otherwise and concede that I may be overlooking something.

iMichka:
I would not allow it for versioned formulae. These are pinned to a specific version. If you want to use HEAD you can use the main formula.

Allowing it for versioned formulae implies that you need to add HEAD-specific hacks to each versioned formula ...

javian:
That might be true in most cases but there are exceptions hence the whitelist. My thoughts went to mature projects such as `php` and `node` which have multiple supported releases but we can always handle them as an exception and add them to the list. I would agree with you that the default should stay as it is.

iMichka:
I guess that for php and node, the devs continue to push changes to their different older versions? I agree we can make an exception for bigger projects, case by case.

MikeMcQuaid:
> I don't have an opinion on versioned formulas. They're supposed to be relatively stable ones, no, so head makes less sense?

The only situation in which `head` makes sense in versioned formulae is if it points to a versioned e.g. release branch that will be used to make the next patch version in that release series. I'd agree that if we allow it a whitelist is the only sensible way.

Issue #5839, opened 2019-03-04
zsh-completion error : showing `usage: date [-jnu] ...`


zijuexiansheng:
**Please note that we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

- [x] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew update` and can still reproduce the problem?
- [x] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [ ] ran `brew config` and `brew doctor` and included their output with your issue?

# What you were trying to do (and why)

I'm trying to use the autocompletion of homebrew and got some error output.

# What happened (include command output)

When I hit the tab to auto complete,  I constantly get the following output.

<details>
  <summary>Command output</summary>
  <pre>
  
usage: date [-jnu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... 
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
usage: date [-jnu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... 
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
usage: date [-jnu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... 
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
  
  </pre>
</details>

# Issue description and proposed solution

This error will not appear in Linux system (or maybe newer Mac OS). My system is Mac OS 10.10.5 and the `date` command is different than the one in the Linux system.

[...]

PR #5842, opened 2019-03-05
completions/zsh: Improve caching behaviour

okdana:

-----

This addresses #5839 and a few related issues:

* Eliminate reliance on external utility calls for cache invalidation
* Update cache policy to also invalidate if cache file is >=2 weeks old or if tap indexes are non-existent
* Do not override any cache policy the user might already have set
* Handle default cache policy determination as early as possible to ensure that a default is set appropriately for all helpers (this will mostly be useful if other helpers need caching in the future)

It's notoriously hard to test zsh completion functions, so if anyone feels like double-checking my work that would be pretty cool.

I noticed some other issues with this function too; i'll try to get around to making separate PRs for them

### DIFF
--- a/completions/zsh/_brew
+++ b/completions/zsh/_brew
@@ -41,18 +41,19 @@ __brew_formulae_or_ruby_files() {
 
 # completions remain in cache until any tap has new commits
 __brew_completion_caching_policy() {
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @okdana!

Issue #5640, opened 2019-01-28
Can't reinstall/upgrade existing Casks that had 'depends_on macos'  for older MacOS versions when originally installed.


chadcatlett:
**Please note that we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.**

- [X] are reporting a bug others will be able to reproduce and not asking a question. If you're not sure or want to ask a question do so on our Discourse: https://discourse.brew.sh
- [X] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [X] ran `brew update` and can still reproduce the problem?
- [X] ran `brew doctor`, fixed all issues and can still reproduce the problem?
- [X] ran `brew config` and `brew doctor` and included their output with your issue?

# What you were trying to do (and why)

Upgrade already installed Casks that were installed prior to deprecation of certain MacOS versions.

# What happened (include command output)

I get an error indicating that the depends_on macos symbol that was present when the cask was installed is not valid. I also tried reinstalling an installed Cask, but I didn't think about it would probably trigger some of the same logic as upgrade.

<details>
  <summary>Command output</summary>
  <pre>
$ brew doctor
Your system is ready to brew.

$ brew config
HOMEBREW_VERSION: 1.9.3-50-gcbb62a3
ORIGIN: https://github.com/Homebrew/brew
HEAD: cbb62a34f54adcb0db69c0bd978800cf158004d7
Last commit: 40 minutes ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: d5b669e1eaa131c1d0b00c72fb37ca7a03e6d33d
Core tap last commit: 5 hours ago
HOMEBREW_PREFIX: /usr/local
HOMEBREW_DEV_CMD_RUN: 1
HOMEBREW_GITHUB_API_TOKEN: set
HOMEBREW_LOGS: /Users/ccatlett/Library/Logs/Homebrew
HOMEBREW_NO_AUTO_UPDATE: 1
[...]

PR #5845, opened 2019-03-06
Keep deprecated OS version definitions

metayan:
.metadata files might refer to old OS versions, so keep
the old version names/numbers defined.
This solves issues https://github.com/Homebrew/homebrew-cask/issues/58046 https://github.com/Homebrew/homebrew-cask/issues/58046 #5640
The .metadata can stay as they are.

- [] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb).

-----

### DIFF
--- a/Library/Homebrew/os/mac/version.rb
+++ b/Library/Homebrew/os/mac/version.rb
@@ -4,12 +4,18 @@ module OS
   module Mac
     class Version < ::Version
       SYMBOLS = {
-        mojave:      "10.14",
-        high_sierra: "10.13",
-        sierra:      "10.12",
-        el_capitan:  "10.11",
-        yosemite:    "10.10",
-        mavericks:   "10.9",
+        mojave:        "10.14",
+        high_sierra:   "10.13",
+        sierra:        "10.12",
+        el_capitan:    "10.11",
+        yosemite:      "10.10",
+        mavericks:     "10.9",
+        mountain_lion: "10.8",
+        lion:          "10.7",
+        snow_leopard:  "10.6",
+        leopard_64:    "10.5",
+        leopard:       "10.5",
+        tiger:         "10.4",
       }.freeze
 
       def self.from_symbol(sym)


metayan:
Not to leave old code, just the **definitions** for the old systems.

vitorgalvao:
I also disagree with this. Fundamentally, this arcs back to the way we deal with the `.metadata` files being inadequate. It’s this problem this time, but there were others in the past. Since we basically need those files for uninstall instructions, one way to fix this would be to, at install time, create a bash file with the uninstall commands that would be saved. That way, it’d be future-proof and independent of any core changes.

metayan:
Please give it another consideration, because putting the definitions back is really a very simple solution to the upgrade/uninstall issues that have popped up recently – this way eliminating the need for users having to search for the cause and then being forced to run the bash script from https://github.com/Homebrew/homebrew-cask/issues/58046 .
Not removing old OS version definitions from `Library/Homebrew/os/mac/version.rb` is a good way to keep already installed casks happy, and also the way deprecation is "normally" handled. Just removing the definitions completely was (obviously) not tested, and a bit of a shot from the hip, but it's easily remediated. (Anyway, thanks for all the effort you put into Homebrew. It's enormously appreciated.)

MikeMcQuaid:
> putting the definitions back is really a very simple solution to the upgrade/uninstall issues that have popped up recently

It seems simple but in reality it's "leave around dead code indefinitely" which further discourages this being fixed properly.

> this way eliminating the need for users having to search for the cause and then being forced to run the bash script from [Homebrew/homebrew-cask#58046](https://github.com/Homebrew/homebrew-cask/issues/58046) 

I agree this is not a good solution and something better should be done.

> Not removing old OS version definitions from `Library/Homebrew/os/mac/version.rb` is a good way to keep already installed casks happy, and also the way deprecation is "normally" handled

No, sorry, it's not. There's also other deprecated and removed methods that will continue to blow up in this way even if these versions have been removed.

> Just removing the definitions completely was (obviously) not tested, and a bit of a shot from the hip, but it's easily remediated.

They were tested but I decided to do so anyway for Homebrew 2.0 to try and force a proper solution for this to be implemented at some stage. Never removing code from parts of Homebrew to keep old casks alive is not a good solution; we have ways we handle this same thing for formulae and have talked about these before in https://github.com/Homebrew/brew/pull/4730. I'd suggest reading that thread for possible inspiration on other approaches.

> (Anyway, thanks for all the effort you put into Homebrew. It's enormously appreciated.)

Thanks for the kind words; thanks to you too for created a PR! ❤️

metayan:
> > putting the definitions back is really a very simple solution to the upgrade/uninstall issues that have popped up recently
> 
> It seems simple but in reality it's "leave around dead code indefinitely" which further discourages this being fixed properly.

Keeping **only the definitions** is the point here.
This is definitely not the same as "leave around dead code indefinitely". (Pun intended.)
Now, by removing the definitions, a situation has been created where active code (the `.metadata` files) ceases to work, whereby brew crashes.
If the definitions are put back, the old code works, and users can be given some information about how to proceed if brew deems itself unable to handle the situation that has appeared.


> > this way eliminating the need for users having to search for the cause and then being forced to run the bash script from [Homebrew/homebrew-cask#58046](https://github.com/Homebrew/homebrew-cask/issues/58046)
> 
> I agree this is not a good solution and something better should be done.

That sounds good.
For me the issue appeared with wireshark, and at first I let the upgrade fail for some weeks, hoping that it was a temporary glitch in the brewing apparatus, but it kept on, so that's why I investigated what was happening. I'm quite certain that there are less proficient (and maybe not as stubborn) Homebrew users who would benefit from getting a proper error message or instructions instead of a crash.


> > Not removing old OS version definitions from `Library/Homebrew/os/mac/version.rb` is a good way to keep already installed casks happy, and also the way deprecation is "normally" handled
> 
> No, sorry, it's not. There's also other deprecated and removed methods that will continue to blow up in this way even if these versions have been removed.

Removing the **definitions** is the cause for the crash here. Deprecated and old methods should be removed, but not the definitions that still active code depends on. (And the .metadata files are active code.)


> > Just removing the definitions completely was (obviously) not tested, and a bit of a shot from the hip, but it's easily remediated.
> 
> They were tested but I decided to do so anyway for Homebrew 2.0 to try and force a proper solution for this to be implemented at some stage.

This, to me, is a bit worrying, because I firmly believe that a program should never "crash". In this case, a semi-proper way would be to exit gracefully and present something that is useful to the "normal" user, instead of
```
Error: Cask 'wireshark' definition is invalid: invalid 'depends_on macos' value: ":mountain_lion"
/usr/local/Homebrew/Library/Homebrew/cask/dsl.rb:211:in `rescue in depends_on'
/usr/local/Homebrew/Library/Homebrew/cask/dsl.rb:208:in `depends_on'
/usr/local/Caskroom/wireshark/.metadata/2.6.6/20190109033447.442/Casks/wireshark.rb:11:in `block in load'
/usr/local/Homebrew/Library/Homebrew/cask/cask.rb:48:in `instance_eval'
/usr/local/Homebrew/Library/Homebrew/cask/cask.rb:48:in `config='
/usr/local/Homebrew/Library/Homebrew/cask/cask.rb:39:in `initialize'
/usr/local/Homebrew/Library/Homebrew/cask/cask_loader.rb:33:in `new'
/usr/local/Homebrew/Library/Homebrew/cask/cask_loader.rb:33:in `cask'
/usr/local/Homebrew/Library/Homebrew/cask/cask_loader.rb:73:in `cask'
/usr/local/Caskroom/wireshark/.metadata/2.6.6/20190109033447.442/Casks/wireshark.rb:1:in `load'
/usr/local/Homebrew/Library/Homebrew/cask/cask_loader.rb:60:in `instance_eval'
/usr/local/Homebrew/Library/Homebrew/cask/cask_loader.rb:60:in `load'
/usr/local/Homebrew/Library/Homebrew/cask/cask_loader.rb:176:in `load'
/usr/local/Homebrew/Library/Homebrew/cask/cmd/list.rb:26:in `block in list'
/usr/local/Homebrew/Library/Homebrew/cask/cmd/list.rb:18:in `each'
/usr/local/Homebrew/Library/Homebrew/cask/cmd/list.rb:18:in `list'
/usr/local/Homebrew/Library/Homebrew/cask/cmd/list.rb:14:in `run'
/usr/local/Homebrew/Library/Homebrew/cask/cmd/abstract_command.rb:34:in `run'
/usr/local/Homebrew/Library/Homebrew/cask/cmd.rb:90:in `run_command'
/usr/local/Homebrew/Library/Homebrew/cask/cmd.rb:156:in `run'
/usr/local/Homebrew/Library/Homebrew/cask/cmd.rb:121:in `run'
/usr/local/Homebrew/Library/Homebrew/cmd/cask.rb:7:in `cask'
/usr/local/Homebrew/Library/Homebrew/brew.rb:102:in `<main>'
Error: Kernel.exit
/usr/local/Homebrew/Library/Homebrew/cask/cmd.rb:161:in `exit'
/usr/local/Homebrew/Library/Homebrew/cask/cmd.rb:161:in `rescue in run'
/usr/local/Homebrew/Library/Homebrew/cask/cmd.rb:144:in `run'
/usr/local/Homebrew/Library/Homebrew/cask/cmd.rb:121:in `run'
/usr/local/Homebrew/Library/Homebrew/cmd/cask.rb:7:in `cask'
/usr/local/Homebrew/Library/Homebrew/brew.rb:102:in `<main>'
```
(Apologies for that long piece, but it *is* meant as an example of something annoying. ;) )


> Never removing code from parts of Homebrew to keep old casks alive is not a good solution; we have ways we handle this same thing for formulae and have talked about these before in #4730. I'd suggest reading that thread for possible inspiration on other approaches.

Thanks. I read that discussion, and some other solution than the `.metadata` files might be needed in the future, but this PR is really about just keeping the **OS version definitions**, so that brew doesn't crash when there are references to them in active code.


> Thanks for the kind words; thanks to you too for created a PR! ❤️

Having (happily) used MacPorts for a long time before, then both that and HomeBrew together, I now completely rely on HomeBrew, which is why I'm very eager to have it working properly, so that many people are happy with it and it can thrive and evolve. This is why I'm so persistent in "pushing" for this solution.

Summing up: We don't want "dead code" lying around, but by keeping the **OS version definitions**, we don't unnecessarily kill live code. Yeah!

MikeMcQuaid:
> This is definitely not the same as "leave around dead code indefinitely". (Pun intended.)

This code is considered dead and should not be used. It is not used in any active `master` code.

> Now, by removing the definitions, a situation has been created where active code (the `.metadata` files) ceases to work, whereby brew crashes.

The handling of unreadable/unparseable `.metadata` files is what should be fixed. `cask_loader.rb` should be updated to handle **all** exceptions that come from `.metadata` and instead throw a warning for `HOMEBREW_DEVELOPER`s.

See https://github.com/Homebrew/brew/blob/01bd14be616904f6ebbb8e01492dea54740942f1/Library/Homebrew/formulary.rb#L23-L27 and https://github.com/Homebrew/brew/blob/01bd14be616904f6ebbb8e01492dea54740942f1/Library/Homebrew/formulary.rb#L155-L159 and https://github.com/Homebrew/brew/blob/01bd14be616904f6ebbb8e01492dea54740942f1/Library/Homebrew/utils.rb#L91-L96 for how we're handling these changes similarly elsewhere for formulae/bottles that have similar problems.

vitorgalvao:
@metayan Simply put (and I don’t really want to drag the issue further), I feel you have the right intention but the wrong fix. Your solution is a band-aid that will indeed leave old code code lingering, *and you don’t even need it*. I appreciate your concern for the other users, but most seem perfectly content with the (non-ideal but workable and easy) one-liner fix.

I’d rather have this small bump and force a correction of the `.metadata` situation (which in it’s current implementation has caused other issues). The alternative hides part of the underlying problem, making it less likely to get fixed.

MikeMcQuaid:
I agree with @vitorgalvao. To be clear: I'm grateful for this PR and sorry I can't agree for it to be merged but I'd love if it turns into something that can be merged.

metayan:
> The handling of unreadable/unparseable `.metadata` files is what should be fixed. `cask_loader.rb` should be updated to handle **all** exceptions that come from `.metadata` and instead throw a warning for `HOMEBREW_DEVELOPER`s.

Indeed, the way to go. Agreed.
Thanks.

PR #5851, opened 2019-03-07
completions/zsh: Hide *instal from all commands

okdana:

-----

After my other completion PR was merged, someone contacted me to ask if i could help with [this Discourse post](https://discourse.brew.sh/t/zsh-brew-completion-feature-bug-or-external-conflict/4228). The user is bothered by the fact that the zsh completion function offers one-letter-short aliases like `instal` and `uninstal`. I agree that they aren't helpful, as they produce confusing-looking output and force unnecessary disambiguation when the user enters (e.g.) `unin`<kbd>tab</kbd>.

These aliases were already hidden in the bash completion function in PR #3052. This change just does the equivalent for the zsh function.

Note for anyone testing this: If you have caching enabled, you may need to delete the cache files (under `~/.zcompcache` by default) to see the change right away.


jonchang:
Thanks for the help @okdana!

PR #5853, opened 2019-03-08
cpu: tests permit new families and architectures

jonchang:

-----

Update tests to match new architechtures and CPU families in `extend/os/{mac,linux}/hardware/cpu.rb`. Reported via Discourse https://discourse.brew.sh/t/brew-tests-fails/4330

### DIFF
--- a/Library/Homebrew/test/hardware/cpu_spec.rb
+++ b/Library/Homebrew/test/hardware/cpu_spec.rb
@@ -4,7 +4,9 @@ describe Hardware::CPU do
   describe "::type" do
     let(:cpu_types) {
       [
+        :arm,
         :intel,
+        :ppc,
         :dunno,
       ]
     }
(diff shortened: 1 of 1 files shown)


Library/Homebrew/test/hardware/cpu_spec.rb
```diff
@@ -4,7 +4,9 @@
   describe "::type" do
     let(:cpu_types) {
       [
+        :arm,
```
sjackman:
```suggestion
        :aarch64,
        :arm,
```

jonchang:
Thanks for the review @sjackman

sjackman:
Thanks for the fix, Jon!

MikeMcQuaid:
We'll need to keep adding to this list when new CPUs are released. Could we consider stubbing instead so it's not reliant on the system CPU?

sjackman:
I suggest stubbing out `RUBY_PLATFORM` to include `aarch64` so that `CPU.type` returns `:arm`, which will work for both Linux and macOS (unlike Intel).

jonchang:
I’ll look into a follow up.

PR #5855, opened 2019-03-09
help: rewrap output to 80 columns

EricFromCanada:
- [n/a] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb).

-----
Adds a call to `Formatter.wrap` to CLI::Parser so that descriptions of commands are pre-wrapped when running `brew help <command>`. Also added to `help.rb` for for commands still using comment-based descriptions. 

`Formatter.wrap` now does the following:
- converts any hard-wrapped paragraphs to a single line
- wraps any option descriptions with an indent that matches what's automatically generated by CLI::Parser (only found in comment-based descriptions) and then rewraps any wrapped overflowing indented lines to the given width
- wraps each non-indented line to the given width

Also fixes help text in `--cellar`, `analytics`, `install`, `list`, `prof`, `search`, `unlink`, and `vendor-install`.

### DIFF
--- a/Library/Homebrew/utils/formatter.rb
+++ b/Library/Homebrew/utils/formatter.rb
@@ -31,8 +31,21 @@ module Formatter
     label(label, string, :red)
   end
 
+  # Wraps text to fit within a given number of columns using regular expressions that:
+  #
+  # 1. convert hard-wrapped paragraphs to a single line
+  # 2. find any option descriptions longer than a pre-set length and wrap between words
+  #    with a hanging indent, without breaking any words that overflow
+  # 3. wrap any remaining description lines that need wrapping with the same indent
+  # 4. wrap all lines to the given width.
+  # @see https://macromates.com/blog/2006/wrapping-text-with-regular-expressions/
   def wrap(s, width = 172)
-    s.gsub(/(.{1,#{width}})(\s+|\Z)/, "\\1\n")
+    desc = OPTION_DESC_WIDTH
+    indent = width - desc
+    s.gsub(/(?<=\S) *\n(?=\S)/, " ")
+     .gsub(/^( +-.+  +(?=\S.{#{desc}}))(.{1,#{desc}})( +|$)\n?/, "\\1\\2\n" + " " * indent)
+     .gsub(/^( {#{indent}}(?=\S.{#{desc}}))(.{1,#{desc}})( +|$)\n?/, "\\1\\2\n" + " " * indent)
+     .gsub(/(.{1,#{width}})( +|$)\n?/, "\\1\n")
   end
 
   def url(string)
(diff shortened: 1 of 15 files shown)


Library/Homebrew/utils/formatter.rb
```diff
@@ -32,7 +32,12 @@ def error(string, label: nil)
   end
 
   def wrap(s, width = 172)
-    s.gsub(/(.{1,#{width}})(\s+|\Z)/, "\\1\n")
+    col = 43
```
MikeMcQuaid: why 43?

Library/Homebrew/utils/formatter.rb
```diff
@@ -32,7 +32,12 @@ def error(string, label: nil)
...
+    indent = width - col
+    s.gsub(/(?<=\S) *\n(?=\S)/, " ")
+     .gsub(/^( +-.+  +(?=\S.{#{col}}))(.{1,#{col}})( +|$)\n?/, "\\1\\2\n" + " " * indent)
+     .gsub(/^( {#{indent}}(?=\S.{#{col}}))(.{1,#{col}})( +|$)\n?/, "\\1\\2\n" + " " * indent)
+     .gsub(/(.{1,#{width}})( +|$)\n?/, "\\1\n")
```
MikeMcQuaid: Could you add a bunch more comments here? At a glance it's very hard to figure out what's going on here. Also: does this wrap at whitespace or in the middle of words?

Library/Homebrew/utils/formatter.rb
EricFromCanada: This matches the value of `@desc_line_length` in `cli_parser.rb`, used when it wraps option descriptions for parser-based commands.

Library/Homebrew/utils/formatter.rb
EricFromCanada: Yes, regular expressions can be like that. 🙃

The first line turns any hard-wrapped paragraphs into a single line. The other three are different takes on the same concept: find a run of characters up to length `n`, then backtrack  (unless the line is already shorter than `n`) until a space is found and replace it with a line break. The algorithm is lifted from this [blog post](https://macromates.com/blog/2006/wrapping-text-with-regular-expressions/), except it doesn't attempt to also line-break words longer than `n`, e.g. URLs.

Library/Homebrew/utils/formatter.rb
MikeMcQuaid: Can you pull both into a constant? Thanks!

Library/Homebrew/utils/formatter.rb
MikeMcQuaid: Great, thanks. Would be good to document your comments here inline in Ruby and link the blog post too?

MikeMcQuaid:
Great, thanks again @EricFromCanada!

PR #5856, opened 2019-03-10
[WIP] ARGV cleanup: replace uses of ARGV methods with Homebrew.args

BenMusch:

-----

This is a simplified version of https://github.com/Homebrew/brew/pull/5762, designed to replace ARGV with Homebrew.args only when doing so does not break the tests, and without deprecating the ARGV methods.

This PR may need some work to pass tests -- I'm opening it right now so that CI is run on my commits. Let me know if it's easier for me to close it until it's done!

Once tests pass I'll rebase the commits into 1 so it's not based on the original branch for the first PR


BenMusch:
I'm a bit confused about this failure. It seems to be caused by https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cli_parser.rb#L138 (not the line above it -- I've tested locally)

But nothing in this PR modifies `ARGV`, right? Any idea what's going on here?

jonchang:
This looks like a test bot failure. You might need to update https://github.com/Homebrew/homebrew-test-bot/blob/master/cmd/brew-test-bot.rb before this will pass tests.

BenMusch:
@jonchang I don't think it's test-bot since I also get the error when running things like `brew install` locally. I'll try some bisecting to see if I can figure it out

BenMusch:
I think this will pass tests now, but the git history is a bit messy. Gonna clean things up before removing the [WIP] tag

scpeters:
there is also a linux test failure

    FormulaInstaller Formula is not poured from bottle when compiler specified
    Failure/Error: allow(Homebrew.args).to receive(:cc).and_return("clang")

Library/Homebrew/test/formula_installer_spec.rb
```diff
@@ -14,6 +14,10 @@
     match(&:poured_from_bottle)
   end
 
+  before do
```
BenMusch: Any preference about this approach vs. something like `allow(Homebrew).to receive(:args).and_return(some_mock_object)`?

PR #5859, opened 2019-03-11
python: support pypy3

jonchang:

-----

A follow up to https://github.com/Homebrew/brew/pull/4833, now that Python 2 is soon-to-be end-of-lifed, we can look at supporting PyPy3 a bit better as well. I tested this with a third-party formula that had a gnarly scipy/numpy dependency stack and it worked perfectly.


MikeMcQuaid:
Thanks again @jonchang!

PR #5873, opened 2019-03-12
docs/Homebrew-on-Linux: Centre the logo

sjackman:


jonchang:
Maybe we can fix the sponsor badges as well? It might be worthwhile to add some or an `_include` to https://github.com/Homebrew/brew.sh for prettier images.

sjackman:
> Maybe we can fix the sponsor badges as well?

Done.

docs/Homebrew-on-Linux.md
```diff
@@ -5,7 +5,10 @@ redirect_from:
   - /Linux
   - /Linuxbrew
 ---
+
+<p style="text-align:center">
```
MikeMcQuaid: I think at the point we're adding this to the Markdown it means we should instead add/modify CSS on brew.sh as @jonchang suggests. Shout if you need help with that.

sjackman: I've opened PR https://github.com/Homebrew/brew.sh/pull/342 to modify the style sheet.

MikeMcQuaid:
Closing in favour of
https://github.com/Homebrew/brew.sh/pull/342

sjackman:
The sizes of the images should be fixed by PR https://github.com/Homebrew/brew.sh/pull/342, but it doesn't centre the images, which was the original purpose of this PR.

sjackman:
For some unknown reason I can't reopen this PR, so I've opened PR https://github.com/Homebrew/brew/pull/5886

PR #5882, opened 2019-03-15
Fix class name for brew extract

kiendang:

-----
Fix an oversight in #5868. When run
```sh
brew extract --version=1.10.5 kubernetes-cli custom/formulae
```
class name for `kubernetes-cli@1.10.5` is `KubernetescliAT1105` instead of `KubernetesCliAT1105` as expected.

So sorry for not testing this out properly earlier. Wanted to add a test case for this but looks like can't `setup_test_formula` a formula name with hyphen.


MikeMcQuaid:
No worries, thanks for the quick fix and thanks again for contributing to Homebrew!

PR #5884, opened 2019-03-16
Fix repo name consistency

eplanet:
It seems `homebrew-repo` refers to `user/repo`. It probably should be `user/homebrew-repo`.

-----


scpeters:
Actually, you don't need to include `homebrew-` in the tap command; it's a bit more concise this way. For https://github.com/Homebrew/homebrew-test-bot you use:

~~~
brew tap homebrew/test-bot
~~~

eplanet:
I see! My bad for not checking that, it seems obvious...

PR #5886, opened 2019-03-16
docs/Homebrew-on-Linux: Centre the logos

sjackman:
See PR https://github.com/Homebrew/brew/pull/5873

### DIFF
--- a/docs/Homebrew-on-Linux.md
+++ b/docs/Homebrew-on-Linux.md
@@ -5,7 +5,10 @@ redirect_from:
   - /Linux
   - /Linuxbrew
 ---
-<img src="https://brew.sh/assets/img/linuxbrew.png" alt="Linuxbrew logo" style="min-width:128px" height="128" width="128">
+
+<p style="text-align:center">
+<img src="https://brew.sh/assets/img/linuxbrew.png" alt="Linuxbrew logo" height="128" width="128">
+</p>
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'd actually rather we went the opposite direction and moved all these back to being raw markdown again.

Stylistically I don't think it looks good to have everything else on every page bar these logos on this page be left-aligned and the logos be centre-aligned.

sjackman:
At https://docs.brew.sh/Homebrew-on-Linux I think it looks strange that the Homebrew logo is centred and the Linuxbrew logo is left aligned. I'm fine with the sponsor logos being left-aligned.

sjackman:
I've reverted the change to the sponsor logos.

docs/Homebrew-on-Linux.md
```diff
@@ -5,7 +5,10 @@ redirect_from:
   - /Linux
   - /Linuxbrew
 ---
-<img src="https://brew.sh/assets/img/linuxbrew.png" alt="Linuxbrew logo" style="min-width:128px" height="128" width="128">
+
+<p style="text-align:center">
```
MikeMcQuaid: Have you tested this locally? This won't do a centre align of the image.

sjackman: Yes I have tested it locally, and yes it does centre the image. I guess `text-align:center` is a misnomer, since it centre aligns all inline elements in the paragraph, including images, not just text.

MikeMcQuaid: My apologies, yes, this works as expected. It still feels a bit gross having the raw HTML here. What would you think about a YAML frontmatter option that replaced the image at the top of the page instead? It does feel a little weird right now to have a Homebrew logo and then an identically sized Linuxbrew logo below it. Could also be nice for e.g. the @BrewTestBot page too.

sjackman: Yep, I like the idea. I don't have the time for it right at this moment, but definitely a good wishlist item. Since this PR is hardly any worse than what we have currently, can we merge this PR as is?

MikeMcQuaid: Done in https://docs.brew.sh/Homebrew-on-Linux

sjackman: Thanks, Mike!

PR #5887, opened 2019-03-17
brew extract from all taps

### LINKED ISSUE #5857: brew extract for formula in taps other than core
# A detailed description of the proposed feature

Currently `brew extract` only works for formulae in core tap. Would be cool if this works for those in other taps as well. Tap where the formula is extracted from may be specified via a `--from` argument. Another thing to consider is the tap where the formula is extracted to should be one different from the `--from` tap and also different from core. I would be happy to submit a pull request.

# The motivation for the feature

I have been a fan of the very useful but quite obscure `brew extract`. Today I wanted to checkout an older version of [`coursier`](https://github.com/coursier/homebrew-formulas) and realized it is not possible.

# How the feature would be relevant to at least 90% of Homebrew users

I would argue that it is as relevant as how the current `brew extract` is relevant. The ability to install an older version of software should not apply only to core. This would be useful for taps that provide formulae for science-related software such as [OSGeo/homebrew-osgeo4mac](https://github.com/OSGeo/homebrew-osgeo4mac) and [brewsci/homebrew-bio](https://github.com/brewsci/homebrew-bio.git) when people need to install specific versions of the tools in order to reproduce results in a paper (I actually ran into this problem before when I needed older versions of `bamtools` and `samtools` when they were in [homebrew-science](https://github.com/homebrew/homebrew-science)).

# What alternatives to the feature have been considered

Edit the formula/git checkout an old commit myself I guess.

kiendang:

-----
Close #5857

Enable `brew extract` from a tap other than core with this syntax:
```sh
brew extract [options] [user/tap/]formula destination-tap
```
e.g.
```sh
brew extract --version=1.1.0-M10 coursier/formulas/coursier custom/formulae
```
Formula cannot be extracted to core tap or the same tap where it comes from.

This turned out to be a bit more involved than I thought. Mainly because originally the formula file path is assumed to be `<tap_path>/Formula/<formula>.rb` when technically if the tap is not core it can be anywhere under `<tap_path>`, i.e. `<tap_path>/{,**/}<formula>.rb`. Thus we have to look in git history for the revision that includes any file that matches this pattern and get the exact file path.

To do this, I added a helper function `last_revision_commit_of_files`, which is similar to `last_revision_commit_of_file`, but it returns both the `rev` and the file paths `[rev, files]`, using the `--name-only` option of `git log`.

I know this is not a small change and I'm also not proficient in ruby so would really appreciate feedback.

### DIFF
--- a/Library/Homebrew/dev-cmd/extract.rb
+++ b/Library/Homebrew/dev-cmd/extract.rb
@@ -97,47 +97,64 @@ module Homebrew
     # Expect exactly two named arguments: formula and tap
     raise UsageError if ARGV.named.length != 2
 
+    if ARGV.named.first !~ HOMEBREW_TAP_FORMULA_REGEX
+      name = ARGV.named.first.downcase
+      source_tap = CoreTap.instance
+    else
+      name = Regexp.last_match(3).downcase
+      source_tap = Tap.fetch(Regexp.last_match(1), Regexp.last_match(2))
+      raise TapFormulaUnavailableError.new(source_tap, name) unless source_tap.installed?
+    end
+
     destination_tap = Tap.fetch(ARGV.named.second)
     odie "Cannot extract formula to homebrew/core!" if destination_tap.core_tap?
+    odie "Cannot extract formula to the same tap!" if destination_tap == source_tap
     destination_tap.install unless destination_tap.installed?
 
-    name = ARGV.named.first.downcase
-    repo = CoreTap.instance.path
-    # Formulae can technically live in "<repo>/<formula>.rb" or
-    # "<repo>/Formula/<formula>.rb", but explicitly use the latter for now
-    # since that is how the core tap is structured.
-    file = repo/"Formula/#{name}.rb"
+    repo = source_tap.path
+    pattern = source_tap.core_tap? ? repo/"Formula/#{name}.rb" : repo/"{,**/}#{name}.rb"
 
     if args.version
       ohai "Searching repository history"
       version = args.version
       rev = "HEAD"
       test_formula = nil
+      result = ""
       loop do
-        loop do
-          rev = Git.last_revision_commit_of_file(repo, file, before_commit: "#{rev}~1")
-          break if rev.empty?
-          break unless Git.last_revision_of_file(repo, file, before_commit: rev).empty?
+        rev, (path,) = Git.last_revision_commit_of_files(repo, pattern, before_commit: "#{rev}~1")
+        odie "Could not find #{name}! The formula or version may not have existed." if rev.nil?
 
+        file = repo/path
+        result = Git.last_revision_of_file(repo, file, before_commit: rev)
+        if result.empty?
           ohai "Skipping revision #{rev} - file is empty at this revision" if ARGV.debug?
+          next
         end
+
         test_formula = formula_at_revision(repo, name, file, rev)
         break if test_formula.nil? || test_formula.version == version
 
         ohai "Trying #{test_formula.version} from revision #{rev} against desired #{version}" if ARGV.debug?
       end
       odie "Could not find #{name}! The formula or version may not have existed." if test_formula.nil?
-      result = Git.last_revision_of_file(repo, file, before_commit: rev)
-    elsif File.exist?(file)
-      rev = "HEAD"
-      version = Formulary.factory(file).version
-      result = File.read(file)
     else
-      ohai "Searching repository history"
-      rev = Git.last_revision_commit_of_file(repo, file)
-      version = formula_at_revision(repo, name, file, rev).version
-      odie "Could not find #{name}! The formula or version may not have existed." if rev.empty?
-      result = Git.last_revision_of_file(repo, file)
+      files = Dir[repo/"{,**/}"].map do |dir|
+        Pathname.glob(["#{dir}/#{name}.rb"]).find(&:file?)
+      end.compact
+
+      if files.empty?
+        ohai "Searching repository history"
+        rev, (path,) = Git.last_revision_commit_of_files(repo, pattern)
+        odie "Could not find #{name}! The formula or version may not have existed." if rev.nil?
+        file = repo/path
+        version = formula_at_revision(repo, name, file, rev).version
+        result = Git.last_revision_of_file(repo, file)
+      else
+        file = files.first.realpath
+        rev = "HEAD"
+        version = Formulary.factory(file).version
+        result = File.read(file)
+      end
     end
 
     # The class name has to be renamed to match the new filename,
--- a/Library/Homebrew/dev-cmd/extract.rb
+++ b/Library/Homebrew/dev-cmd/extract.rb
@@ -81,7 +81,8 @@ module Homebrew
 
         Look through repository history to find the most recent version of <formula> and
         create a copy in <tap>`/Formula/`<formula>`@`<version>`.rb`. If the tap is not
-        installed yet, attempt to install/clone the tap before continuing.
+        installed yet, attempt to install/clone the tap before continuing. To extract
+        a <formula> from a tap that is not homebrew/core use <user>/<repo>/<formula>.
       EOS
 
       flag "--version=",
@@ -112,7 +113,12 @@ module Homebrew
     destination_tap.install unless destination_tap.installed?
 
     repo = source_tap.path
-    pattern = source_tap.core_tap? ? repo/"Formula/#{name}.rb" : repo/"{,**/}#{name}.rb"
+    pattern = if source_tap.core_tap?
+      [repo/"Formula/#{name}.rb"]
+    else
+      # A formula can technically live in the root directory of a tap or in any of its subdirectories
+      [repo/"#{name}.rb", repo/"**/#{name}.rb"]
+    end
 
     if args.version
       ohai "Searching repository history"
@@ -138,6 +144,7 @@ module Homebrew
       end
       odie "Could not find #{name}! The formula or version may not have existed." if test_formula.nil?
     else
+      # Search in the root directory of <repo> as well as recursively in all of its subdirectories
       files = Dir[repo/"{,**/}"].map do |dir|
         Pathname.glob(["#{dir}/#{name}.rb"]).find(&:file?)
       end.compact
--- a/Library/Homebrew/test/dev-cmd/extract_spec.rb
+++ b/Library/Homebrew/test/dev-cmd/extract_spec.rb
@@ -1,5 +1,5 @@
 describe "brew extract", :integration_test do
-  it "retrieves the specified version of formula, defaulting to most recent" do
+  it "retrieves the specified version of formula from core tap, defaulting to most recent" do
     path = Tap::TAP_DIRECTORY/"homebrew/homebrew-foo"
     (path/"Formula").mkpath
     target = Tap.from_path(path)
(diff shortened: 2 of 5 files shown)


Library/Homebrew/dev-cmd/extract.rb
```diff
@@ -97,47 +97,64 @@ def extract
     # Expect exactly two named arguments: formula and tap
     raise UsageError if ARGV.named.length != 2
 
+    if ARGV.named.first !~ HOMEBREW_TAP_FORMULA_REGEX
```
MikeMcQuaid: This warrants a documentation change.

Library/Homebrew/dev-cmd/extract.rb
```diff
@@ -97,47 +97,64 @@ def extract
...
     destination_tap.install unless destination_tap.installed?
 
-    name = ARGV.named.first.downcase
-    repo = CoreTap.instance.path
-    # Formulae can technically live in "<repo>/<formula>.rb" or
-    # "<repo>/Formula/<formula>.rb", but explicitly use the latter for now
-    # since that is how the core tap is structured.
-    file = repo/"Formula/#{name}.rb"
+    repo = source_tap.path
+    pattern = source_tap.core_tap? ? repo/"Formula/#{name}.rb" : repo/"{,**/}#{name}.rb"
```
MikeMcQuaid: The `**` usage warrants a comment.

Library/Homebrew/dev-cmd/extract.rb
```diff
@@ -97,47 +97,64 @@ def extract
...
       version = args.version
       rev = "HEAD"
       test_formula = nil
+      result = ""
       loop do
-        loop do
-          rev = Git.last_revision_commit_of_file(repo, file, before_commit: "#{rev}~1")
-          break if rev.empty?
-          break unless Git.last_revision_of_file(repo, file, before_commit: rev).empty?
+        rev, (path,) = Git.last_revision_commit_of_files(repo, pattern, before_commit: "#{rev}~1")
```
MikeMcQuaid: Why the `()`?

Library/Homebrew/dev-cmd/extract.rb
```diff
@@ -97,47 +97,64 @@ def extract
...
-      rev = "HEAD"
-      version = Formulary.factory(file).version
-      result = File.read(file)
     else
-      ohai "Searching repository history"
-      rev = Git.last_revision_commit_of_file(repo, file)
-      version = formula_at_revision(repo, name, file, rev).version
-      odie "Could not find #{name}! The formula or version may not have existed." if rev.empty?
-      result = Git.last_revision_of_file(repo, file)
+      files = Dir[repo/"{,**/}"].map do |dir|
```
MikeMcQuaid: The `**` warrants a comment (and perhaps a constant as it's being also used above)

Library/Homebrew/dev-cmd/extract.rb
```diff
@@ -97,47 +97,64 @@ def extract
...
+        Pathname.glob(["#{dir}/#{name}.rb"]).find(&:file?)
+      end.compact
+
+      if files.empty?
+        ohai "Searching repository history"
+        rev, (path,) = Git.last_revision_commit_of_files(repo, pattern)
+        odie "Could not find #{name}! The formula or version may not have existed." if rev.nil?
```
MikeMcQuaid: These two lines are duplicated above, would be nice to avoid that if possible.

Library/Homebrew/test/dev-cmd/extract_spec.rb
```diff
@@ -29,4 +29,69 @@
 
     expect(Formulary.factory(path/"Formula/testball@0.1.rb").version).to be == "0.1"
   end
+
+  it "retrieves the specified version of formula from a tap other than core, defaulting to most recent" do
```
MikeMcQuaid: Please don't add more integration tests for this, they are super slow. Either make this a unit test, alter the existing test or leave it untested, thanks.

Library/Homebrew/utils/git.rb
```diff
@@ -14,6 +14,19 @@ def last_revision_commit_of_file(repo, file, before_commit: nil)
     out.chomp
   end
 
+  def last_revision_commit_of_files(repo, file, before_commit: nil)
+    args = [before_commit.nil? ? "--skip=1" : before_commit.split("..").first]
```
MikeMcQuaid: Use an `if` here; this is pretty hard to follow.

Library/Homebrew/utils/git.rb
```diff
@@ -14,6 +14,19 @@ def last_revision_commit_of_file(repo, file, before_commit: nil)
...
+
+    cmd = [
+      HOMEBREW_SHIMS_PATH/"scm/git", "-C", repo,
+      "log", "--format=%h", "--abbrev=7", "--max-count=1", "--name-only",
+      *args, "--", file
+    ]
+    out, = Open3.capture3("/bin/bash", "-c", cmd.join(" "))
```
MikeMcQuaid: It seems like a very bad idea to be using `bash` here with arguments that are user supplied; it would be very easy to trick users into running commands they didn't want to.

kiendang: I see. I use `bash` there because without it the globstar `**` wouldn't work. I will try another way to search for the exact file path then.

MikeMcQuaid: Great, thanks ❤️

Library/Homebrew/dev-cmd/extract.rb
kiendang: `last_revision_commit_of_files` return a list of files, i.e. `[rev, [file1, file2,...]]`. Here Just need to get the 1 file.

Library/Homebrew/dev-cmd/extract.rb
kiendang: `last_revision_commit_of_files` return a list of files, i.e. `[rev, [file1, file2,...]]`. Here just need to get the 1 file.

MikeMcQuaid: Cool. Never seen that syntax before.

Library/Homebrew/utils/git.rb
```diff
@@ -14,6 +20,31 @@ def last_revision_commit_of_file(repo, file, before_commit: nil)
     out.chomp
   end
 
+  def last_revision_commit_of_files(repo, files, before_commit: nil)
+    args = [
+      if before_commit.nil?
```
MikeMcQuaid: This is repeated above; could it be moved into a shared function?

Library/Homebrew/utils/git.rb
```diff
@@ -14,6 +20,31 @@ def last_revision_commit_of_file(repo, file, before_commit: nil)
...
+    #   <file_path1>
+    #   <file_path2>
+    #   ...
+    # return [<commit_hash>, [file_path1, file_path2, ...]]
+    cmd = [
+      HOMEBREW_SHIMS_PATH/"scm/git", "-C", repo, "log",
+      "--pretty=format:%h", "--abbrev=7", "--max-count=1",
+      "--diff-filter=d", "--name-only", *args, "--", *files
+    ]
+    out, = Open3.capture3(cmd.join(" "))
```
MikeMcQuaid: Why join these?

Library/Homebrew/utils/git.rb
```diff
@@ -14,6 +20,31 @@ def last_revision_commit_of_file(repo, file, before_commit: nil)
     out.chomp
   end
 
+  def last_revision_commit_of_files(repo, files, before_commit: nil)
```
MikeMcQuaid: This function feels like it has a lot of overlap. Could it share some code with the previous function or be one function that takes different parameters? Also, it may be worth just living in `extract.rb` until it has more than one caller.

Library/Homebrew/dev-cmd/extract.rb
```diff
@@ -97,47 +98,70 @@ def extract
...
-      odie "Could not find #{name}! The formula or version may not have existed." if rev.empty?
-      result = Git.last_revision_of_file(repo, file)
+      # Search in the root directory of <repo> as well as recursively in all of its subdirectories
+      files = Dir[repo/"{,**/}"].map do |dir|
+        Pathname.glob(["#{dir}/#{name}.rb"]).find(&:file?)
+      end.compact
+
+      if files.empty?
+        ohai "Searching repository history"
+        rev, (path,) = Git.last_revision_commit_of_files(repo, pattern)
```
MikeMcQuaid: Could be nice to share this similar code in a function.

Library/Homebrew/utils/git.rb
kiendang: I tried not joining as well as putting them as arguments directly but the command didn't work. Will investigate further.

Library/Homebrew/utils/git.rb
kiendang: I think it's better to leave them as separate functions. They both call `git log` with different arguments but these arguments are set to make sure `git log` output is of a certain format. The two functions are functionally different too. They expect different things, look for different things and return different things. They do look kinda similar now but I've gone through versions of `last_revision_commit_of_files` where it looks very different from `last_revision_commit_of_file`. In the future if there's a need to update `last_revision_commit_of_files` it would be easier because it's not constrained by the other one and vice versa.

Library/Homebrew/dev-cmd/extract.rb
kiendang: I thought of putting in a function but haven't come up with anything that wouldn't make it look more complicated.

Library/Homebrew/utils/git.rb
kiendang: Works now. I must have forgotten to unpack earlier.

Library/Homebrew/utils/git.rb
kiendang: Also I think should leave it here with the other functions coz they're related.

Library/Homebrew/dev-cmd/extract.rb
kiendang: I knew this from python. Just tried it in ruby and it works :)

MikeMcQuaid: Heh nice 👏

Library/Homebrew/utils/git.rb
MikeMcQuaid: I'm unsure about this being a "utils" function when it's used only once. I could see the argument for keeping them alongside each other if they shared code but they don't. Your thoughts?

Library/Homebrew/dev-cmd/extract.rb
MikeMcQuaid: Ok.

Library/Homebrew/utils/git.rb
```diff
@@ -4,7 +4,13 @@ module Git
   module_function
 
   def last_revision_commit_of_file(repo, file, before_commit: nil)
-    args = [before_commit.nil? ? "--skip=1" : before_commit.split("..").first]
+    args = [
+      if before_commit.nil?
```
MikeMcQuaid: Sorry to be a pain: can the `if` live outside the array; this it not typical Ruby usage. Instead consider something like
```ruby
args = if before_commit.nil?
  ["skip=1"]
```

Library/Homebrew/utils/git.rb
kiendang: My argument for keeping it in the same place with the other two functions is that they are kinda close in terms of what they do. But I'm happy either way.
I notice that the other 2 functions in `utils/git` are also only used in `extract` 🤔

Library/Homebrew/utils/git.rb
kiendang: No worries. It's definitely better style.

Library/Homebrew/utils/git.rb
MikeMcQuaid: 🆒 👍

MikeMcQuaid:
Thanks again so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @kiendang!

kiendang:
Thanks a lot @MikeMcQuaid for patiently guiding me through this!

MikeMcQuaid:
@kiendang You're very welcome ❤️

PR #5888, opened 2019-03-17
cask automerge: restrict allowed chars in version bump

vitorgalvao:

-----

I still have a broken ruby install (hope to upgrade the OS in the next few days), hence why no `style` or `tests` run.

This changes the `automerge` rule in Cask to *only* merge if the `version` conforms to a subset of characters, namely:

* Numbers (common in versions).
* `.` (common in versions).
* `-` (common in versions).
* `_` (common in versions).
* `,` (we use it break version sections).
* `:` (we use it break version sections).

The reasoning is that we rely on interplation in `url` *a lot*. That has proven itself a huge win (and the only reason `automerge` can exist). But if we’re not seeing what gets merged and only if two lines changed (and CI passed), that introduces an attack vector: a malicious actor could find the right cask hosted on the right place and introduce a version change with characters such as `/` that when expanded would direct the download somewhere else.

By requiring manual verification (i.e. no automerging) in PRs that don’t conform to the subset of safe characters, we make such an attack unfeasible (rather than improbable). My proposal is that we start with the collection of allowed characters as small as possible (current PR), and expand only if necessary.

As a bonus, not allowing letters should curb some of the automerging of non-stable casks in stable repos when they are obvious (cc @core-code).

### DIFF
--- a/Library/Homebrew/cask/cmd/automerge.rb
+++ b/Library/Homebrew/cask/cmd/automerge.rb
@@ -98,6 +98,10 @@ module Cask
         file.a_path.match?(%r{\ACasks/[^/]+\.rb\Z})
       end
 
+      def check_allowed_chars(line)
+        line.sub(/.*'(.+)'.*/, '\1').match?(/^[0-9\.\-_,:]+$/)
+      end
+
       def diff_only_version_or_checksum_changed(diff)
         lines = diff.files.flat_map(&:hunks).flat_map(&:lines)
 
@@ -108,7 +112,11 @@ module Cask
         return false if additions.count != deletions.count
         return false if additions.count > 2
 
-        changed_lines.all? { |line| diff_line_is_version(line.to_s) || diff_line_is_sha256(line.to_s) }
+        changed_lines.all? do |line|
+          next true if diff_line_is_version(line.to_s)
+          next true if check_allowed_chars(line.to_s)
+          next true if diff_line_is_sha256(line.to_s)
+        end
       end
 
       def diff_line_is_sha256(line)


core-code:
sounds good, all for it. i am also seeing a lot of '-' and '_' used in version stanzas and i doubt these could be used for malicious purposes.

vitorgalvao:
> i am also seeing a lot of '-' and '_' used in version stanzas and i doubt these could be used for malicious purposes.

Agreed. Expanded the list.

vitorgalvao:
Not being able to run tests locally is really biting me on this one. I really have to upgrade my system.

core-code:
if anyone can do it by running a few terminal commands, just send them along and i'll post the output ;)

vitorgalvao:
> if anyone can do it by running a few terminal commands, just send them along and i'll post the output ;)

Thank you. Seems like this last time did it. Finally passing.

Library/Homebrew/cask/cmd/automerge.rb
```diff
@@ -108,7 +112,10 @@ def diff_only_version_or_checksum_changed(diff)
         return false if additions.count != deletions.count
         return false if additions.count > 2
 
-        changed_lines.all? { |line| diff_line_is_version(line.to_s) || diff_line_is_sha256(line.to_s) }
+        changed_lines.all? do |line|
+          (diff_line_is_version(line.to_s) && check_allowed_chars(line.to_s)) ||
```
MikeMcQuaid: Might want to use a `next if` and `next unless` here as it's not super easy to parse when this will pass/fail using binary precedence.

MikeMcQuaid:
> By requiring manual verification (i.e. no automerging) in PRs that don’t conform to the subset of safe characters, we make such an attack unfeasible (rather than improbable). My proposal is that we start with the collection of allowed characters as small as possible (current PR), and expand only if necessary.

👍

reitermarkus:
Letters are also allowed.

With letters added, could you check if there are actually cask versions which don't match? If there aren't any casks with unsafe characters we could actually make this an audit.

reitermarkus:
Also, this is the wrong place for this change. This command should actually be deleted now that this is done by GitHub Actions: https://github.com/Homebrew/homebrew-cask/tree/master/.github/actions/automerge

core-code:
i also see many version numbers that are prefixed with a single 'v', e.g. 'v1.2.3'

core-code:
whats the rationale for only automerging version increases? 

version decreases occur to fix wrong beta submissions and are arguably much more urgent.

scpeters:
I'm not a cask maintainer, but here's my guesses:

> whats the rationale for only automerging version increases?

I believe it's for security, to prevent the attack whereby an old version with known vulnerabilities is submitted and auto-merged.

> version decreases occur to fix wrong beta submissions and are arguably much more urgent.

They are urgent, but human review may be necessary in these cases.

core-code:
sounds reasonable. i guess the right angle-of-attack is to prevent beta-merges from occurring in the first place. i have a system in place that detects some of these, but its run on the current casks, i.e. can only detect things after the fact. i'll have a look if that can be hooked up to the 'automerge' tool mentioned above.

reitermarkus:
@core-code, I have [this](https://github.com/Homebrew/brew/pull/5893) in the works which will prevent at least obvious beta versions.

core-code:
@reitermarkus thanks for the info. this definitely looks like an improvement. it wouldn't have prevented cases like #60692 today though. obviously no solution will ever be perfect, but i'm seeing good success with the method i've implemented which checks whether the version is contained in the appcast. we are running it daily now for all casks that have non-sparkle (there are other checks for those), non-github appcasts - but its only done after the merge took place and only once a day. i guess it should ideally be implemented either in the Travis CI or in the Automerge. which one do you think would be better? another question is the necessary configuration. while many appcasts contain the whole version (well, if you cut at ',' '-' and '_' and take the left part), some appcasts contain the version number in some different notation. we have a configuration file with transformations for 35 casks. where would that live? i guess ideally it would be an addition to the appcast stanza like the referer.

reitermarkus:
> i guess it should ideally be implemented either in the Travis CI or in the Automerge. which one do you think would be better?

I think the CI script would be best, since `automerge` would merely ignore it.

reitermarkus:
> i guess ideally it would be an addition to the appcast stanza like the referer.

Yes, another parameter for `appcast` would probably be the best place for this.

vitorgalvao:
I disagree with adding more options to `appcast`. In my view, we should even be considering to *remove* the `appcast`, and move their workings to an alternative repo/system.

There’s no point to allowing/automerging PRs from everyone but then introducing checks to notice when those PRs are stable or beta and merge or not. If we add the capability to detect the correct versions, then let’s just update it!

We don’t need to take updates from everyone. I think I mentioned this in the automerge PR, but I’d be fine with only automerging specific users, such as @core-code and @brianmorton, which we know submit often and correctly.

If we implement a system to make the checks, we could even go so far as to *block* submissions that don’t come from it. If the system has a problem, users can submit a fix to it.

core-code:
> I disagree with adding more options to appcast. In my view, we should even be considering to remove the appcast, and move their workings to an alternative repo/system.


i am not a HBC maintainer so i don't really care how things are implemented (as long as the objectives are met ;), thats for you guys to decide. 
some points to consider
• iff we have a system where PRs are checked for validity through the appcast, and it requires configuration in some cases (which i don't think we can do without), then it makes sense to keep that configuration 'near' the appcast, or at least in some place that can be updated by users like us. of course moving appcasts to another repo could allow for that just as placing it next to the appcast stanza does.
• i don't see the long-term plan of removing appcasts as a reason not to implement this in the proposed way. what i am proposing is a short term solution to a pressing problem (beta releases automerged). we had the situation previously where we didn't implement useful changes because of some imagined long-term changes that didn't come do be anyway. the problem is, proposed enhancements often don't happen, or don't happen as soon as hoped for - so preventing short term improvements because of them doesn't make much sense to me.

> introducing checks to notice when those PRs are stable or beta and merge or not. 

hm, its better with this check than without, which is our current status ;)

> If we add the capability to detect the correct versions, then let’s just update it!

well, checking if a proposed new version in a PR is contained in the appcast and therefore legal, is something different than finding out the newest version of an app automatically. while we would be able do that for apps with sparkle-feed with some effort and probably even for apps with an github-feed, i don't see it happening anytime soon for apps with custom-appcast or html-"feed".

> but I’d be fine with only automerging specific users

> If we implement a system to make the checks

i am all for it and willing to put effort to help in there but it sounds like a long-term improvement to me. until this is here the proposed solutions of restricting automerges via appcast-checking black/whitelisting of submitting users could help.

brianmorton:
I feel like a fish out of water on this, as I am not familiar with the automerge abilities and haven't done automated scanning/PRs in a while, although I appreciate being tagged. Without educating myself on this matter I don't think I have anything intelligent to add, but I have full confidence in those of you involved in this thread!

vitorgalvao:
> iff we have a system where PRs are checked for validity through the appcast

To clarify, my suggestion is to skip checking PRs at all; just the upgrade strategies. We (a bot) would create the PRs.

> and it requires configuration in some cases (which i don't think we can do without)

Agreed.

> then it makes sense to keep that configuration 'near' the appcast, or at least in some place that can be updated by users like us.

That would still happen. We’re just talking a separate repo with all the upgrade strategies laid out, but they’d even follow the token rules of naming.

I also agree with your second point. I’m just not sure the complexity of the interim system is worth it.

> while we would be able do that for apps with sparkle-feed with some effort and probably even for apps with an github-feed, i don't see it happening anytime soon for apps with custom-appcast or html-"feed".

App by app it’s pretty doable.

> proposed enhancements often don't happen (…) i am all for it and willing to put effort to help in there but it sounds like a long-term improvement to me.

Making this system is one of the coding projects I’m most excited about and have several ideas. I don’t want to make it HBC-specific, but tool-agnostic so everyone can benefit from it. My biggest impediment right now is justifying the time commitment. Maybe there’s something like GSoC where I could apply with the idea and finally get it off the ground. I’ve been toying with it for far too long.

---

@brianmorton Nothing needs to change in your process, so no need to worry about it.

---

Closing this PR in favour of https://github.com/Homebrew/brew/pull/5893.

core-code:
@vitorgalvao thanks for your reply, sounds sensible to me. the only thing where i disagree is the demand for an "interim" solution. but there is nothing to worry about - if i can whip something up (i don't think it needs to be that complex to improve upon what we have), i'll send a PR, and if not, so be it. all positive on the long-term plans - sadly you've missed the GSoC 2019 deadline ;/

vitorgalvao:
> sadly you've missed the GSoC 2019 deadline ;/

I couldn’t apply anyway; it’s only for students.

core-code:
ok, i've worked on this a bit. i've implemented a new flag --appcast for 'brew cask audit' that does what we've talked about, i.e. it checks if the version is contained in the downloaded appcast. the idea here would be that Travis CI passes this flag (which is off by default), to make the Travis check fail when the version submitted in the PR is not contained in the appcast, therefore fulfilling our goal of preventing auto-merge of beta versions.

i think its quite useful as-is. i've run 'brew cask audit --appcast' which has audited 3072 casks. the new check has failed for 197 of them, which is around 6%. considering that some of these failures are due to casks that are actually not up-to-date (just submitted two updates), the check would currently prevent auto-merging on less than 5% of the casks. if we want to perfect this system even further, we could just add the configuration we've talked about. i think there are two parts here: 1.) allow configuration if the appcast does not contain the version number in exactly the format as specified in the cask and we would need some transformation, e.g. dots to underscores. [note: configuration could also increase the accuracy of the check. for example HTML appcasts are often nearly a megabyte and contain a lot of stuff. the (invalid) version number could also be contained in the appcast by accident. configuration could provide a pattern that includes the version number, so that the check only succeeds if the whole string including the version number is found, e.g. we search for "Newest Version: #{version}" instead of just for "#{version}"]  2.) a blacklist for appcasts that actually do not contain the version number and where this check is therefore not useful. note that some appcasts that do not contain the version number should just be removed or replaced, some are still useful, because although they don't contain the version number they can be monitored for changes (e.g. a link to the SHA of the current release). i've already done finished the configuration and blacklist files for a subset of all casks (those without sparkle appcast), covering around 40 configured and 40 blacklisted casks. 

i've attached 
appcastcheck.patch: which contains the propsed source code changes
appcastcheck_blacklist.txt: a blacklist for some appcasts that do not contain version numbers. this is not yet used or useful with the current version of the patch i've provided. 
appcastcheck_configuration.json: which contains the talked about configuration. this is not yet useful with the current version patch i've provided but could be used if we add the mentioned configuration capability. i've not yet ported all of them to ruby, some use our internal transformation system

since i am neither familiar with Ruby nor with the HBC internals, i think i've done as much as i can and would probably need help if we want to implement the "configuration", whether e.g. if it is to be used as an optional additional parameter for the appcast stanze.

please let me know how to proceed.
[appcastcheck_blacklist.txt](https://github.com/Homebrew/brew/files/3026956/appcastcheck_blacklist.txt)

[appcastcheck_configuration.txt](https://github.com/Homebrew/brew/files/3026957/appcastcheck_configuration.txt)
[appcastcheck_patch.txt](https://github.com/Homebrew/brew/files/3026958/appcastcheck_patch.txt)

vitorgalvao:
5% doesn’t mean it *will* happen 5% of the time, but that it *might happen at most* 5% of the time, so it’s good enough. The work to shave that off would be disproportional to the gains.

> please let me know how to proceed.

Either a PR or a new issue with the attachments and the most relevant information, so we get a clean discussion, please.

core-code:
@vitorgalvao thanks for your feedback!

i agree that it is useful (i.e. better than what we have now) as-is, but i am certainly willing to put a little more effort into it, with a little guidance ;)

i've opened a PR here:
https://github.com/Homebrew/brew/pull/5972

PR #5894, opened 2019-03-19
fix_dynamic_linkage: Check HOMEBREW_RELOCATE_METAVARS

sjackman:
Only remove RPATH when HOMEBREW_RELOCATE_METAVARS is set.

See PR https://github.com/Homebrew/brew/pull/5413

### DIFF
--- a/Library/Homebrew/extend/os/mac/keg_relocate.rb
+++ b/Library/Homebrew/extend/os/mac/keg_relocate.rb
@@ -72,7 +72,8 @@ class Keg
 
         # If none of the install names reference RPATH(s), then we can safely
         # remove all RPATHs from the file.
-        if file.dynamically_linked_libraries.none? { |lib| lib.start_with?("@rpath") }
+        if ENV["HOMEBREW_RELOCATE_METAVARS"] &&
+           file.dynamically_linked_libraries.none? { |lib| lib.start_with?("@rpath") }
           # Note: This could probably be made more efficient by reverse-sorting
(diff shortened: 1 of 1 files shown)


woodruffw:
I think the problematic line is this one:

https://github.com/Homebrew/brew/blob/f67fad1e6e4f10d1c02cf85d3069c283648564b9/Library/Homebrew/extend/os/mac/keg_relocate.rb#L123

Since we're no longer filtering `@rpath`-prefixed install names, they end up getting passed through `fixed_name` and falling through to the `bad_name` case. We should either revert that regexp back to `/^@(loader_|executable_|r)path/` or add an additional environment check there like:

```ruby
dylibs.reject! { |fn| fn =~ /^@rpath/ } unless ENV["HOMEBREW_RELOCATE_METAVARS"]
```

MikeMcQuaid:
> We should either revert that regexp back to `/^@(loader_|executable_|r)path/` or add an additional environment check there like:
> 
> ```ruby
> dylibs.reject! { |fn| fn =~ /^@rpath/ } unless ENV["HOMEBREW_RELOCATE_METAVARS"]
> ```

👍

sjackman:
I've added…
```ruby
dylibs.reject! { |fn| fn =~ /^@rpath/ } unless ENV["HOMEBREW_RELOCATE_METAVARS"]
```

jonchang:
https://github.com/Homebrew/brew/pull/5894 might have fixed it. Can you open a pull request to bump the revision of `swiftgen`?

PR #5919, opened 2019-03-24
formula-cookbook: devel specs not allowed in core

fxcoudert:
- State that `devel` specs are not allowed in homebrew-core
- Since `head` specs are more often used that `devel`, list them first


MikeMcQuaid:
Thanks @fxcoudert!

PR #5926, opened 2019-03-25
Fixed checking absolute path of brew

dracorp:
Here is fix for **Cowardly refusing to continue at this prefix: /usr**
Functions borrowed from [basher](https://github.com/basherpm/basher.git).

-----

### DIFF
--- a/bin/brew
+++ b/bin/brew
@@ -7,6 +7,24 @@ if ! [[ -d "$PWD" ]]; then
   exit 1
 fi
 
+resolve_link() {
+  $(type -p greadlink readlink | head -1) "$1"
+}
+
+abs_dirname() {
+  local cwd="$(pwd)"
+  local path="$1"
+
+  while [ -n "$path" ]; do
+    cd "${path%/*}"
+    local name="${path##*/}"
+    path="$(resolve_link "$name" || true)"
+  done
+
+  pwd
+  cd "$cwd"
+}
+
 quiet_cd() {
   cd "$@" >/dev/null || return
 }
@@ -19,7 +37,7 @@ symlink_target_directory() {
   quiet_cd "$directory" && quiet_cd "$target_dirname" && pwd -P
 }
 
-BREW_FILE_DIRECTORY="$(quiet_cd "${0%/*}/" && pwd -P)"
+BREW_FILE_DIRECTORY="$(abs_dirname "$0")"
 HOMEBREW_BREW_FILE="${BREW_FILE_DIRECTORY%/}/${0##*/}"
 HOMEBREW_PREFIX="${HOMEBREW_BREW_FILE%/*/*}"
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
What are you trying to fix with this? We need more context to review or consider this PR. Thanks!

dracorp:
Ok, I've installed brew on Linux, Archlinux. By default some packages go to /opt/ directory and they have a symlink in /usr/bin to the main program. For example for brew:
```
$ which brew | xargs ls -l
lrwxrwxrwx 1 root root 23 Mar 26 19:52 /usr/bin/brew -> /opt/linuxbrew/bin/brew
```
Unfortunately this does not work:
```
$ brew
Error: Cowardly refusing to continue at this prefix: /usr
```
With my fix now it works:
```
$ which brew | xargs ls -l
lrwxrwxrwx 1 root root 23 Mar 26 20:53 /usr/bin/brew -> /opt/linuxbrew/bin/brew
$ brew
Example usage:
  brew search [TEXT|/REGEX/]
  brew info [FORMULA...]
  brew install FORMULA...
  brew update
  brew upgrade [FORMULA...]
  brew uninstall FORMULA...
  brew list [FORMULA...]

Troubleshooting:
  brew config
  brew doctor
  brew install --verbose --debug FORMULA

Contributing:
  brew create [URL [--no-fetch]]
  brew edit [FORMULA...]

Further help:
  brew commands
  brew help [COMMAND]
  man brew
  https://docs.brew.sh
```

MikeMcQuaid:
> lrwxrwxrwx 1 root root 23 Mar 26 19:52 /usr/bin/brew -> /opt/linuxbrew/bin/brew

This is not a use-case we wish to support. Adding 20 lines of code for this edge-case is not worth the pay-off, I'm afraid. Sorry!

iMichka:
I think you should just put brew in your PATH, to be able to run it, instead of symlinking it into /usr/bin. Or is that not OK on Archlinux?

dracorp:
[Yes and no](https://wiki.archlinux.org/index.php/Arch_package_guidelines#Directories). If I put the brew in /usr/bin, it will not work because of this:
```
if [[ "$HOMEBREW_PREFIX" = "/" || "$HOMEBREW_PREFIX" = "/usr" ]]
then
  # it may work, but I only see pain this route and don't want to support it
  odie "Cowardly refusing to continue at this prefix: $HOMEBREW_PREFIX"
fi
```
and this
```
BREW_FILE_DIRECTORY="$(quiet_cd "${0%/*}/" && pwd -P)"
```
This is the system installation and not just for a single user.

MikeMcQuaid:
@dracorp We intentionally do not support a `/usr` installation (as noted by the comment).

dracorp:
Ok.

PR #5940, opened 2019-03-27
xcode: upgrade for Xcode 10.2 release

lembacon:

-----

From Xcode 10.2 Release Notes,
> Xcode 10.2 requires a Mac running macOS 10.14.3 or later.

### DIFF
--- a/Library/Homebrew/os/mac.rb
+++ b/Library/Homebrew/os/mac.rb
@@ -200,6 +200,7 @@ module OS
       "9.4"   => { clang: "9.1", clang_build: 902 },
       "10.0"  => { clang: "10.0", clang_build: 1000 },
       "10.1"  => { clang: "10.0", clang_build: 1000 },
+      "10.2"  => { clang: "10.0", clang_build: 1001 },
     }.freeze
 
     def compilers_standard?
--- a/Library/Homebrew/os/mac/xcode.rb
+++ b/Library/Homebrew/os/mac/xcode.rb
@@ -14,12 +14,12 @@ module OS
         when "10.11" then "8.2.1"
         when "10.12" then "9.2"
         when "10.13" then "10.1"
-        when "10.14" then "10.1"
+        when "10.14" then "10.2"
         else
           raise "macOS '#{MacOS.version}' is invalid" unless OS::Mac.prerelease?
 
           # Default to newest known version of Xcode for unreleased macOS versions.
-          "10.1"
+          "10.2"
         end
       end
 
@@ -171,8 +171,8 @@ module OS
         when 81      then "8.3"
         when 90      then "9.2"
         when 91      then "9.4"
-        when 100     then "10.1"
-        else              "10.1"
+        when 100     then "10.2"
+        else              "10.2"
         end
       end
 
@@ -240,7 +240,7 @@ module OS
         # on the older supported platform for that Xcode release, i.e there's no
         # CLT package for 10.11 that contains the Clang version from Xcode 8.
         case MacOS.version
-        when "10.14" then "1000.10.44.2"
+        when "10.14" then "1001.0.46.3"
         when "10.13" then "1000.10.44.2"
         when "10.12" then "900.0.39.2"
         when "10.11" then "800.0.42.1"


MikeMcQuaid:
👍 once we've updated the base images.

jdhealy:
Hey, to bring awareness of an issue with the Swift aspect of Xcode 10.2, [command line applications built with `swiftlang-1001.0.69.5` (shipped with Xcode 10.2)](https://forums.swift.org/t/command-line-applications-crashes-with-xcode-10-2/22192/7) will need [an additional package (installed to `/usr/lib/swift`)](https://forums.swift.org/t/command-line-applications-crashes-with-xcode-10-2/22192/12) or else [pre-macOS-10.14.4 versions of Mojave](https://forums.swift.org/t/command-line-applications-crashes-with-xcode-10-2/22192) will face crashes when they run.

Sorry, to have such a small sliver of CLI apps have an impact on the process here, but this is going to cause issues for the [Carthage](https://github.com/Homebrew/homebrew-core/blob/5d00dd42f3e13c5b68438adbb3f6173c066f5ede/Formula/carthage.rb) and [SwiftLint](https://github.com/Homebrew/homebrew-core/blob/5d00dd42f3e13c5b68438adbb3f6173c066f5ede/Formula/swiftlint.rb) among others.

Open to all solutions, but one pathway for bottling these suitably for all Mojave releases might be to allow them to install Xcode 10.1 for their Mojave bottling process? Just a high-level thought, happy to work with y'all on other solutions or implementation details.

MikeMcQuaid:
@jdhealy Thanks for the heads up! Is this only an issue if they are built without that package installed? What about if they are build with a full Xcode installed?

jdhealy:
My understanding is — on a build like [this one](https://app.bitrise.io/build/a35b64ee2340f819) — a reliance (via `-rpath -Xlinker /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx`) on an Xcode 10.2+ installation being at that baked-in path is the only thing allowing it to run on pre-macOS-10.14.4 versions of Mojave.

MikeMcQuaid:
So if folks are on 10.14.4 OR we build with < Xcode 10.2 there's no problem/changes needed? If so we could give a grace period of a few weeks before we upgrade and we could make the Xcode build dependency require the specific version.

lembacon:
- Binaries will not be linked against system shipped Swift libraries unless the deployment target is explicitly set to macOS 10.14.4 or later. IIRC we're not currently passing `-target` while invoking `swift`, so this might cause some trouble for users who are still using macOS 10.14~10.14.3 after we upgrade our CI machines to macOS 10.14.4 and Xcode 10.2.
- However, we are forcing formulae using Swift to be linked against Swift libraries statically via `-Xswiftc -static-stdlib`.

lembacon:
FYI, [Swift 5 Release Notes for Xcode 10.2](https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_release_notes/swift_5_release_notes_for_xcode_10_2).

jdhealy:
sure @lembacon is aware of this — but to bring everyone same-page: [`swiftlang-1001.0.69.5`’s `swiftc` “no longer supports statically linking the Swift libraries”](https://forums.swift.org/t/command-line-applications-crashes-with-xcode-10-2/22192/12)

lembacon:
Oops, didn't realize that support for `-static-stdlib` has been dropped. Thanks @jdhealy for pointing this out. 😅

jdhealy:
For Carthage — we're already seeing users non-bottle-installing on sub-10.14.4-Mojave and asking for help in our GitHub issues because they've renamed their 10.2 Xcode bundle.

We'd really love to safeguard them against that brittleness by steering them toward a bottle or a local-Xcode-10.1 build:

```ruby
  depends_on :xcode => ["9.4", :build]

  throw [
    "∆placeholder∆ See <https://github.com/Homebrew/brew/pull/5940#issuecomment-477583315>.",
    "Perhaps, try `--force-bottle` with «DEVELOPER_DIR='/var/empty' HOMEBREW_NO_BOTTLE_SOURCE_FALLBACK=true brew»?",
    "Or, perhaps, if Xcode 10.1 is available, «xcode-select» that?"
  ].join("\n• ") if (
    MacOS::Xcode.version >= "10.2" &&
    (0..3).any? { |x| "10.14.#{x}" == String(MacOS.full_version) }
  )

  def install
```

Would something like that be amenable as part of the solution? We have a fair number of users regularly `brew install --HEAD`.

MikeMcQuaid:
> For Carthage — we're already seeing users non-bottle-installing on sub-10.14.4-Mojave and asking for help in our GitHub issues because they've renamed their 10.2 Xcode bundle.

@jdhealy Can you elaborate on this, I'm not sure I fully understand?

> Would something like that be amenable as part of the solution? We have a fair number of users regularly `brew install --HEAD`.

I don't think so, I'm afraid. We need to work out a solution that doesn't break in that case. It may be that the formula is configured to hard fail if their Xcode is in a location that doesn't allow it to be found at runtime?

MikeMcQuaid:
@lembacon What do you think is the best solution here? Always set `-target` on 10.14.4 or later?

jdhealy:
> Can you elaborate on this, I'm not sure I fully understand?

If a user on sub-10.14.4-Mojave using a Swift 5 CLI app happens to remove or relocate Xcode.app, say to make room for Xcode-beta.app, without [the additional package installed to `/usr/lib/swift`](https://forums.swift.org/t/command-line-applications-crashes-with-xcode-10-2/22192/12), all they will see is [the very unhelpful `dyld: Library not loaded: @rpath/libswiftCore.dylib → Abort trap: 6`](https://forums.swift.org/t/command-line-applications-crashes-with-xcode-10-2/22192).

We'd like to steer them in the direction of getting a 10.1-build bottle, `xcode-select`ing 10.1 for a local build, or upgrading (if corporate allows) to macOS 10.14.4.

MikeMcQuaid:
@jdhealy Cool, makes sense. I'd suggest a PR to add an `odie` to `def install` that fails to build from source if they are on macOS >=10.14, macOS <10.14.4 and with Xcode >=10.2.

In terms of the solution for when we upgrade to Xcode 10.2: I'm not sure. Any suggestions, @jdhealy?

jdhealy:
For Carthage — we'd definitely like to get one more 10.1-on-Mojave-bottled release out within the timeframe of a week (prob. closer within a day or two). Under foreseen circumstances, we'd be supporting Carthage building on both Xcodes 10.1 and 10.2 for a while.

That gives us leeway before we devote time to engineering our future-focused homebrew solution.

A somewhat-easy, widely-supporting solution — we'd be very in favor if y'all would want to just copy-and-rename the !0.1-built-High-Sierra bottling (for Carthage or select others) to have that be the Mojave bottle. Maybe that works as a slight side-route within the current bottling business logic?

MikeMcQuaid:
> A somewhat-easy, widely-supporting solution — we'd be very in favor if y'all would want to just copy-and-rename the !0.1-built-High-Sierra bottling (for Carthage or select others) to have that be the Mojave bottle. Maybe that works as a slight side-route within the current bottling business logic?

The Mojave bottle can be removed and the High Sierra bottle will be used by default. That won't fix the "build from source" issue though, obviously.

jdhealy:
> The Mojave bottle can be removed and the High Sierra bottle will be used by default.

We'd love to do that, what's the communication process for that?

> That won't fix the "build from source" issue though, obviously.

For Carthage, we were planning on supporting source builds on both Xcodes 10.1 and 10.2 for an extended period anyways (caveat on unforeseen circumstances).

MikeMcQuaid:
> We'd love to do that, what's the communication process for that?

@jdhealy Submit a pull request 😄.

MikeMcQuaid:
Have opened https://github.com/Homebrew/brew/pull/5985 for a check to ensure the Xcode and macOS versions will be compatible for Swift packages. 

What we still need to figure out (and I need everyone's help with) is how we're going to adjust our compilation on 10.14.4 to ensure we ship Swift packages that work.

jdhealy:
Just downloaded https://bintray.com/homebrew/bottles/download_file?file_path=carthage-0.33.0.mojave.bottle.tar.gz — it was definitely built with Swift 5 (`strings carthage/0.33.0/bin/carthage | grep --only-matching 'swiftlang-[^s]*'` → `swiftlang-1001.0.69.5/`).

Could that bottle be removed (as discussed in https://github.com/Homebrew/brew/pull/5940#issuecomment-479008608)? Do I send another pull request for that, @MikeMcQuaid?

MikeMcQuaid:
Yes, it seems we have Xcode 10.2 installed on the build machines through either autoupdate or a maintainer update. May as well merge this as-is, then.

PR #5958, opened 2019-03-31
CompilerSelector: prefer the gcc version offered by the gcc formula

### LINKED ISSUE #5953: [Linux] Wrong GCC is picked as default compiler
- [x] ran a `brew` command and reproduced the problem with multiple formulae? If it's a problem with a single, official formula (not cask) please file this issue at Homebrew/homebrew-core: https://github.com/Homebrew/homebrew-core/issues/new/choose. If it's a `brew cask` problem please file this issue at https://github.com/Homebrew/homebrew-cask/issues/new/choose. If it's a tap (e.g. Homebrew/homebrew-php) problem please file this issue at the tap.
- [x] ran `brew config` and `brew doctor` and included their output with your issue?

If any gcc newer than gcc-5 is installed (such as `gcc@8`), the newer gcc will be picked as the default compiler. This would cause other formulae (if being built from source) to be linked to the libraries from `gcc@*`. As a result, if `gcc@*` were later uninstalled, these formulae may not work probably. We should always use `gcc` formula as the default compiler.

This issue is due to the downstream code https://github.com/Linuxbrew/brew/pull/676 is not merged back into Homebrew/brew.

<details>
  <summary>brew config</summary>
  <pre><code>$ brew config
HOMEBREW_VERSION: 2.0.6-61-g372a8d4
ORIGIN: https://github.com/Homebrew/brew
HEAD: 372a8d45325f53355c36c0f1b264b72cc186bcb0
Last commit: 5 hours ago
Core tap ORIGIN: https://github.com/Linuxbrew/homebrew-core
Core tap HEAD: 20b90d90b178b950c36ad1be71310eb53e198e71
Core tap last commit: 5 hours ago
HOMEBREW_PREFIX: /home/comp/chengxu/usr
HOMEBREW_REPOSITORY: /home/comp/chengxu/usr
HOMEBREW_CELLAR: /home/comp/chengxu/usr/Cellar
HOMEBREW_CACHE: /tmp/chengxu/cache/homebrew
HOMEBREW_DEVELOPER: true
HOMEBREW_FORCE_BREWED_GIT: true
HOMEBREW_FORCE_VENDOR_RUBY: true
HOMEBREW_LOGS: /tmp/chengxu/cache/homebrew/logs
HOMEBREW_NO_ANALYTICS: true
HOMEBREW_NO_AUTO_UPDATE: true
HOMEBREW_SANDBOX: true
HOMEBREW_SYSTEM_CURL_TOO_OLD: 1
CPU: 24-core 64-bit westmere
Homebrew Ruby: 2.3.7 => /home/comp/chengxu/usr/Library/Homebrew/vendor/portable-ruby/2.3.7/bin/ruby
Clang: 8.0 build 800
Git: 2.21.0 => /home/comp/chengxu/usr/opt/git/bin/git
Curl: 7.64.0 => /home/comp/chengxu/usr/opt/curl/bin/curl
Java: 1.8.0_201
Kernel: Linux 2.6.32-754.10.1.el6.x86_64 x86_64 GNU/Linux
OS: CentOS release 6.10 (Final) (Final)
Host glibc: 2.14
/usr/bin/gcc: 4.4.7
glibc: 2.23
gcc: 5.5.0_4
xorg: N/A
</code></pre>
</details>

xu-cheng:

-----

When GCC is used (default for Linux), we should prefer the gcc offered
by the gcc formula. As such even if users install a gcc with higher
version from `gcc@*` formula, it will not be picked up to build other
formulae. This would also allow users to safely delete `gcc@*` formula.

Closes #5953.

### DIFF
--- a/Library/Homebrew/compilers.rb
+++ b/Library/Homebrew/compilers.rb
@@ -105,11 +105,19 @@ class CompilerSelector
 
   private
 
+  def gnu_gcc_versions
+    # prioritize gcc version provided by gcc formula.
+    v = Formulary.factory("gcc").version.to_s.slice(/\d/)
+    GNU_GCC_VERSIONS - [v] + [v] # move the version to the end of the list
+  rescue FormulaUnavailableError
+    GNU_GCC_VERSIONS
+  end
+
   def find_compiler
     compilers.each do |compiler|
       case compiler
       when :gnu
-        GNU_GCC_VERSIONS.reverse_each do |v|
+        gnu_gcc_versions.reverse_each do |v|
           name = "gcc-#{v}"
           version = compiler_version(name)
           yield Compiler.new(name, version) unless version.null?
--- a/Library/Homebrew/test/compiler_selector_spec.rb
+++ b/Library/Homebrew/test/compiler_selector_spec.rb
@@ -39,6 +39,12 @@ describe CompilerSelector do
       expect(subject.compiler).to eq("gcc-7")
     end
 
+    it "returns gcc-6 if gcc formula offers gcc-6" do
+      software_spec.fails_with(:clang)
+      allow(Formulary).to receive(:factory).with("gcc").and_return(double(version: "6.0"))
+      expect(subject.compiler).to eq("gcc-6")
+    end
+
     it "raises an error when gcc or llvm is missing" do
       software_spec.fails_with(:clang)
       software_spec.fails_with(gcc: "7")


Library/Homebrew/compilers.rb
```diff
@@ -105,11 +105,19 @@ def compiler
 
   private
 
+  def gnu_gcc_versions
+    # prioritize gcc version provided by gcc formula.
+    v = Formulary.factory("gcc").version.to_s.slice(/\d/)
+    GNU_GCC_VERSIONS - [v] + [v]
```
MikeMcQuaid: It's not obvious here why the same thing is being added and removed; could this perhaps have a comment saying that it's being deliberately added to the end/beginning of the list specifically? Thanks!

MikeMcQuaid:
Thanks for the PR @xu-cheng!

MikeMcQuaid:
Thanks again @xu-cheng!

PR #5961, opened 2019-03-31
Add bump-revision command

Moisan:

-----

`brew bump-revision` will increase by one the revision of the formula or add "revision 1" if it doesn't exist.

I took the liberty of not opening an issue before the pull request since this is a feature I quickly discussed during the February 2019 meeting.

Points I am unsure:
- It currently only allows one formula as argument, should it handle many?
- The test is based on `bump-formula-pr`, should it extend to more cases?

### DIFF
--- a/Library/Homebrew/dev-cmd/bump-revision.rb
+++ b/Library/Homebrew/dev-cmd/bump-revision.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+require "formula"
+require "cli/parser"
+
+module Homebrew
+  module_function
+
+  def bump_revision_args
+    Homebrew::CLI::Parser.new do
+      usage_banner <<~EOS
+        `bump-revision` [<options>] [<formula>]
+
+        Create a commit to increment the revision of the formula. If no revision is
+         present, "revision 1" will be added.
+      EOS
+      switch "-n", "--dry-run",
+        description: "Print what would be done rather than doing it."
+      flag "--message=",
+        description: "Append the provided <message> to the default commit message."
+
+      switch :force
+      switch :quiet
+      switch :verbose
+      switch :debug
+    end
+  end
+
+  def bump_revision
+    bump_revision_args.parse
+
+    # As this command is simplifying user run commands then let's just use a
+    # user path, too.
+    ENV["PATH"] = ENV["HOMEBREW_PATH"]
+
+    raise "Multiple formulae given, only one is allowed." if ARGV.formulae.length > 1
+
+    formula = ARGV.formulae.first
+    current_revision = formula.revision
+
+    if current_revision.zero?
+      formula_spec = formula.stable
+      hash_type, old_hash = if (checksum = formula_spec.checksum)
+        [checksum.hash_type, checksum.hexdigest]
+      end
+
+      if hash_type
+        # insert replacement revision after hash
+        old = <<~EOS
+          #{hash_type} "#{old_hash}"
+        EOS
+      else
+        # insert replacement revision after :revision
+        old = <<~EOS
+          :revision => "#{formula_spec.specs[:revision]}"
+        EOS
+      end
+      replacement = old + "  revision 1\n"
+
+    else
+      old = "revision #{current_revision}"
+      replacement = "revision #{current_revision+1}"
+    end
+
+    if args.dry_run?
+      ohai "replace #{old.inspect} with #{replacement.inspect}" unless Homebrew.args.quiet?
+    else
+      Utils::Inreplace.inreplace(formula.path) do |s|
+        s.gsub!(old, replacement)
+      end
+    end
+
+    message = "#{formula.name}: revision bump #{args.message}"
+    if args.dry_run?
+      ohai "git commit --no-edit --verbose --message=#{message} -- #{formula.path}"
+    else
+      formula.path.parent.cd do
+        safe_system "git", "commit", "--no-edit", "--verbose",
+          "--message=#{message}", "--", formula.path
+      end
+    end
+  end
+end
--- a/Library/Homebrew/test/dev-cmd/bump-revision_spec.rb
+++ b/Library/Homebrew/test/dev-cmd/bump-revision_spec.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+require "cmd/shared_examples/args_parse"
+
+describe "Homebrew.bump_revision_args" do
+  it_behaves_like "parseable arguments"
+end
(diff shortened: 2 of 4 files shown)


Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,86 @@
...
+    else
+      old = "revision #{current_revision}"
+      new = "revision #{current_revision+1}"
+      if args.dry_run? && !arg.write?
+        ohai "replace #{old.inspect} with #{new.inspect}" unless Homebrew.args.quiet?
+      else
+        Utils::Inreplace.inreplace(formula.path) do |s|
+          s.gsub!(old, new)
+        end
+      end
```
scpeters: some of this code is duplicated across the if/else; I'll see if it can be consolidated

scpeters: for your consideration: https://github.com/scpeters/brew/commit/304da75fa5bf77b5d9065091fa253997cd9be68c

scpeters:
when you run with `--dry-run` and without `--write`, it prints the old and new test on the same line, which is very likely truncated:

~~~
brew bump-revision -n boost
==> replace "sha256 \"8f32d4617390d1c2d16f26a27ab60d97807b35440d45891fa340fc2648
==> git commit --no-edit --verbose --message=boost: revision bump  -- /usr/local
~~~

it would be nicer if it printed them on separate lines with proper alignment

MikeMcQuaid:
> * It currently only allows one formula as argument, should it handle many?

No, that seems fine.

> * The test is based on `bump-formula-pr`, should it extend to more cases?

You can write additional unit tests but no more than one integration test, please.

Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,86 @@
...
+  def bump_revision_args
+    Homebrew::CLI::Parser.new do
+      usage_banner <<~EOS
+        `bump-revision` [<options>] [<formula>]
+
+        Create a commit to increment the revision of the formula. If no revision is
+         present, "revision 1" will be added.
+      EOS
+      switch "-n", "--dry-run",
+        description: "Print what would be done rather than doing it."
```
MikeMcQuaid: Is this a useful flag to have? I'm wondering if we need it for a v1 and it makes the code more complex.

Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,86 @@
...
+      switch "--write",
+        depends_on:  "--dry-run",
+        description: "When passed along with `--dry-run`, perform a not-so-dry run by making the expected "\
+                     "file modifications but not taking any Git actions."
+      flag   "--message=",
+        description: "Append the provided <message> to the default PR message."
```
MikeMcQuaid: Do you mean commit message?

Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,86 @@
...
+    Homebrew::CLI::Parser.new do
+      usage_banner <<~EOS
+        `bump-revision` [<options>] [<formula>]
+
+        Create a commit to increment the revision of the formula. If no revision is
+         present, "revision 1" will be added.
+      EOS
+      switch "-n", "--dry-run",
+        description: "Print what would be done rather than doing it."
+      switch "--write",
```
MikeMcQuaid: Is this a useful flag to have? I'm wondering if we need it for a v1.

Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def bump_revision
+    bump_revision_args.parse
+
+    # As this command is simplifying user run commands then let's just use a
+    # user path, too.
+    ENV["PATH"] = ENV["HOMEBREW_PATH"]
+
+    formula = ARGV.formulae.first
```
MikeMcQuaid: May want to error if there's more than one.

Library/Homebrew/dev-cmd/bump-revision.rb
Moisan: I think so, it's useful at least for debugging. Combined with the changes from @scpeters it is only two additional `if` block.

Library/Homebrew/dev-cmd/bump-revision.rb
Moisan: Thanks, I included your commit directly.

Moisan:
@scpeters regarding the dry-run truncated line: this is also the case for `bump-formula-pr` unless using `--verbose`. I tried to output the whole line by adding `\n` in the string but it still gets truncated at the same character. I suspect this is due to the lack of `--verbose` that limits the length of the strings outputted.

MikeMcQuaid:
@Moisan Any news on this?

Moisan:
I believe I've either commented or fixed the changes requested so I thought the ball was in the reviewers court.

Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    if current_revision.zero?
+      formula_spec = formula.stable
+      hash_type, old_hash = if (checksum = formula_spec.checksum)
+        [checksum.hash_type, checksum.hexdigest]
+      end
+
+      if hash_type
+        # insert new revision after hash
+        old = "#{hash_type} \"#{old_hash}\"\n"
```
MikeMcQuaid: Use `<<~EOS` to add the newline and avoid the `\"`

Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,77 @@
...
+      else
+        # insert new revision after :revision
+        old = ":revision => \"#{formula_spec.specs[:revision]}\"\n"
```
MikeMcQuaid: Use `<<~EOS` to add the newline and avoid the `\"`

Library/Homebrew/dev-cmd/bump-revision.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+      new = old + "  revision 1\n"
```
MikeMcQuaid: Don't use `new` as it's a special Ruby keyword.

Moisan:
@MikeMcQuaid : gentle poke.

MikeMcQuaid:
Thanks @Moisan, nice work! FYI you need to post a message on a thread after you force push as people watching don't get any notification otherwise.

PR #5972, opened 2019-04-02
Add checking whether an appcast contains the version number

core-code:

-----

This pull request a new flag --appcast for 'brew cask audit' that checks if the version (from the version stanza) is contained in the downloaded appcast (from the appcast stanza). The new flag is off by default. The idea here is that the Travis CI check can pass this optional flag so that beta versions submitted to Homebrew Cask will fail the Travis check (assuming they are not contained in the appcast) and will therefore not be automatically merged by @BrewTestBot 
Please see here for the discussion that lead up to this PR:
https://github.com/Homebrew/brew/pull/5888

### DIFF
--- a/Library/Homebrew/cask/audit.rb
+++ b/Library/Homebrew/cask/audit.rb
@@ -9,11 +9,16 @@ require "utils/git"
 module Cask
   class Audit
     include Checkable
+    extend Predicable
 
     attr_reader :cask, :commit_range, :download
 
-    def initialize(cask, download: false, check_token_conflicts: false, commit_range: nil, command: SystemCommand)
+    attr_predicate :check_appcast?
+
+    def initialize(cask, check_appcast: false, download: false, check_token_conflicts: false,
+                   commit_range: nil, command: SystemCommand)
       @cask = cask
+      @check_appcast = check_appcast
       @download = download
       @commit_range = commit_range
       @check_token_conflicts = check_token_conflicts
@@ -40,6 +45,7 @@ module Cask
       check_latest_with_appcast
       check_latest_with_auto_updates
       check_stanza_requires_uninstall
+      check_appcast_contains_version
       self
     rescue => e
       odebug "#{e.message}\n#{e.backtrace.join("\n")}"
@@ -290,6 +296,22 @@ module Cask
       Verify.all(cask, downloaded_path)
     rescue => e
       add_error "download not possible: #{e.message}"
+    end
+
+    def check_appcast_contains_version
+      return unless check_appcast?
+      return if cask.appcast.to_s.empty?
+
+      appcast_stanza = cask.appcast.to_s
+      appcast_contents, = curl_output("--max-time", "5", appcast_stanza)
+      version_stanza = cask.version.to_s
+      adjusted_version_stanza = version_stanza.split(",")[0].split("-")[0].split("_")[0]
+      return if appcast_contents.include? adjusted_version_stanza
+
+      add_warning "appcast at URL '#{appcast_stanza}' does not contain"\
+                  " the version number: '#{adjusted_version_stanza}'"
+    rescue
+      add_error "appcast at URL '#{appcast_stanza}' offline or looping"
     end
 
     def check_https_availability
--- a/Library/Homebrew/cask/auditor.rb
+++ b/Library/Homebrew/cask/auditor.rb
@@ -4,15 +4,24 @@ require "cask/download"
 
 module Cask
   class Auditor
-    def self.audit(cask, audit_download: false, check_token_conflicts: false, quarantine: true, commit_range: nil)
-      new(cask, audit_download, check_token_conflicts, quarantine, commit_range).audit
+    include Checkable
+    extend Predicable
+
+    def self.audit(cask, audit_download: false, audit_appcast: false,
+                   check_token_conflicts: false, quarantine: true, commit_range: nil)
+      new(cask, audit_download: audit_download,
+                audit_appcast: audit_appcast,
+                check_token_conflicts: check_token_conflicts,
+                quarantine: quarantine, commit_range: commit_range).audit
     end
 
     attr_reader :cask, :commit_range
 
-    def initialize(cask, audit_download, check_token_conflicts, quarantine, commit_range)
+    def initialize(cask, audit_download: false, audit_appcast: false,
+                   check_token_conflicts: false, quarantine: true, commit_range: nil)
       @cask = cask
       @audit_download = audit_download
+      @audit_appcast = audit_appcast
       @quarantine = quarantine
       @commit_range = commit_range
       @check_token_conflicts = check_token_conflicts
@@ -22,9 +31,7 @@ module Cask
       @audit_download
     end
 
-    def quarantine?
-      @quarantine
-    end
+    attr_predicate :audit_appcast?, :quarantine?
 
     def check_token_conflicts?
       @check_token_conflicts
@@ -57,7 +64,8 @@ module Cask
 
     def audit_cask_instance(cask)
       download = audit_download? && Download.new(cask, quarantine: quarantine?)
-      audit = Audit.new(cask, download:              download,
+      audit = Audit.new(cask, check_appcast:         audit_appcast?,
+                              download:              download,
                               check_token_conflicts: check_token_conflicts?,
                               commit_range:          commit_range)
       audit.run!
--- a/Library/Homebrew/cask/cmd/audit.rb
+++ b/Library/Homebrew/cask/cmd/audit.rb
@@ -4,6 +4,7 @@ module Cask
   class Cmd
     class Audit < AbstractCommand
       option "--download",        :download,        false
+      option "--appcast",         :appcast,         false
       option "--token-conflicts", :token_conflicts, false
 
       def self.help
@@ -22,6 +23,7 @@ module Cask
       def audit(cask)
         odebug "Auditing Cask #{cask}"
         Auditor.audit(cask, audit_download:        download?,
+                            audit_appcast:         appcast?,
                             check_token_conflicts: token_conflicts?,
                             quarantine:            quarantine?)
       end
--- a/Library/Homebrew/test/cask/cmd/audit_spec.rb
+++ b/Library/Homebrew/test/cask/cmd/audit_spec.rb
@@ -21,7 +21,7 @@ describe Cask::Cmd::Audit, :cask do
       expect(Cask::CaskLoader).to receive(:load).with(cask_token).and_return(cask)
 
       expect(Cask::Auditor).to receive(:audit)
-        .with(cask, audit_download: false, check_token_conflicts: false, quarantine: true)
+        .with(cask, audit_download: false, audit_appcast: false, check_token_conflicts: false, quarantine: true)
         .and_return(true)
 
       described_class.run(cask_token)
@@ -32,7 +32,7 @@ describe Cask::Cmd::Audit, :cask do
     it "does not download the Cask per default" do
       allow(Cask::CaskLoader).to receive(:load).and_return(cask)
       expect(Cask::Auditor).to receive(:audit)
-        .with(cask, audit_download: false, check_token_conflicts: false, quarantine: true)
+        .with(cask, audit_download: false, audit_appcast: false, check_token_conflicts: false, quarantine: true)
         .and_return(true)
 
       described_class.run("casktoken")
@@ -41,7 +41,7 @@ describe Cask::Cmd::Audit, :cask do
     it "download a Cask if --download flag is set" do
       allow(Cask::CaskLoader).to receive(:load).and_return(cask)
       expect(Cask::Auditor).to receive(:audit)
-        .with(cask, audit_download: true, check_token_conflicts: false, quarantine: true)
+        .with(cask, audit_download: true, audit_appcast: false, check_token_conflicts: false, quarantine: true)
         .and_return(true)
 
       described_class.run("casktoken", "--download")
@@ -52,7 +52,7 @@ describe Cask::Cmd::Audit, :cask do
     it "does not check for token conflicts per default" do
       allow(Cask::CaskLoader).to receive(:load).and_return(cask)
       expect(Cask::Auditor).to receive(:audit)
-        .with(cask, audit_download: false, check_token_conflicts: false, quarantine: true)
+        .with(cask, audit_download: false, audit_appcast: false, check_token_conflicts: false, quarantine: true)
         .and_return(true)
 
       described_class.run("casktoken")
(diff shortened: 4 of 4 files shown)


amyspark:
@vitorgalvao, what's the rationale for this change? (Please, tl;dr version if possible.)

Library/Homebrew/cask/auditor.rb
```diff
@@ -1,16 +1,20 @@
 require "cask/download"
+require "net/http"
 
 module Cask
+  # rubocop:disable Metrics/ParameterLists
   class Auditor
-    def self.audit(cask, audit_download: false, check_token_conflicts: false, quarantine: true, commit_range: nil)
-      new(cask, audit_download, check_token_conflicts, quarantine, commit_range).audit
+    def self.audit(cask, audit_download: false, audit_appcast: false,
```
amyspark: Just FTR, this signature is becoming quite bloated.

vitorgalvao:
@amyspark 

> The idea here is that the Travis CI check can pass this optional flag so that beta versions submitted to Homebrew Cask will fail the Travis check (assuming they are not contained in the appcast) and will therefore not be automatically merged by @BrewTestBot

Library/Homebrew/cask/auditor.rb
core-code: any ideas how to avoid that without significantly complicating the implementation?

core-code:
is this going anywhere? @reitermarkus any opinions on the PR? thanks!

amyspark:
Pinging @Homebrew/maintainers for finalizing review.

Library/Homebrew/cask/auditor.rb
reitermarkus: Convert the `initialize` below to use named parameters.

reitermarkus: Convert the `initialize` below to use named parameters.

reitermarkus:
The main problem I'm seeing with this right now is that this probably won't pass for every cask. Have you checked that it does?

core-code:
i think i've talked about that already in the mentioned thread https://github.com/Homebrew/brew/pull/5888

around 5% of the casks fail (because the appcast does not contain the version number as-is). i'll try to get more accurate numbers now. 

getting that down to probably zero failures would be possible by annotating the appcast stanza with a modifier for the version number (e.g. if the appcast only contains the version with dots replaced by underscores) as we've already discussed. those appcasts that do not contain the version number at all (but are still useful because they can be monitored for changes (think checksums), could be annotated to ignore them for this check.

@vitorgalvao mentioned that further work would be overkill because preventing auto-merging on just 5% of casks is not a big problem

Library/Homebrew/cask/auditor.rb
core-code: thanks i'll ready up on ruby and try to do that.

core-code:
real numbers:

`brew cask audit --appcast `

seemed to check 5549 casks. of those 6 were failing with offline/looping appcasts and 377 were failing with 'does not contain version'. thats a 6.9% failure rate. if we take those into account that are not failing because the cask is not actually up-to-date, the 5% that would need further configuration to work seems like a good guess.

Library/Homebrew/cask/auditor.rb
```diff
@@ -60,11 +68,41 @@ def audit_cask_instance(cask)
...
+      audit.success? && audit_cask_appcast
+    end
+
+    def audit_cask_appcast
+      if audit_appcast?
+        loaded_cask = CaskLoader.load(cask.sourcefile_path)
+        appcast_stanza = loaded_cask.appcast.to_s
+        if appcast_stanza.length.positive?
+          begin
+            appcast_contents = Net::HTTP.get(URI(appcast_stanza))
```
reitermarkus: Use `curl_output` for this instead.

Library/Homebrew/cask/auditor.rb
```diff
@@ -60,11 +68,41 @@ def audit_cask_instance(cask)
                               commit_range:          commit_range)
       audit.run!
       puts audit.summary
-      audit.success?
+      audit.success? && audit_cask_appcast
```
reitermarkus: Can you move this check into `Audit`?

core-code:
thanks for the review. i'm currently moving so i don't have time right now but will try to make the requested changes in a few days.

core-code:
i think all raised points are addressed now:
• use curl_output
• use named parameter
• move the check to the 'Audit' class

core-code:
hm having some troubles with git, will try to reapply...

core-code:
i think it looks good now

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,24 @@ def check_download
       add_error "download not possible: #{e.message}"
     end
 
+    def check_appcast_stanza
```
reitermarkus: Maybe call this `check_appcast_contains_version`.

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,24 @@ def check_download
...
+      return unless check_appcast?
+      return unless cask.appcast.to_s.length.positive?
```
reitermarkus:
```suggestion
      return if cask.appcast.to_s.empty?
```

Library/Homebrew/cask/auditor.rb
```diff
@@ -1,16 +1,23 @@
 require "cask/download"
 
 module Cask
+  # rubocop:disable Metrics/ParameterLists
```
reitermarkus: This shouldn't be needed now.

core-code:
thanks for the review, i've applied the suggested improvements

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,24 @@ def check_download
...
 
+    def check_appcast_contains_version
+      return unless check_appcast?
+      return if cask.appcast.to_s.empty?
+
+      appcast_stanza = cask.appcast.to_s
+      appcast_contents, = curl_output("--max-time", "5", appcast_stanza)
+      version_stanza = cask.version.to_s
+      adjusted_version_stanza = version_stanza.split(",")[0].split("-")[0].split("_")[0]
+      if appcast_contents.include? adjusted_version_stanza
```
reitermarkus:
```suggestion
      return if appcast_contents.include? adjusted_version_stanza
```

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,24 @@ def check_download
...
+        true
+      else
+        w2 = "does not contain the version number:"
+        add_error "Warning: appcast at URL '#{appcast_stanza}' #{w2} '#{adjusted_version_stanza}'\n"
+      end
+    rescue
+      add_error "Error: appcast at URL '#{appcast_stanza}' offline or looping\n"
```
reitermarkus: Remove `Error: ` here.

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,24 @@ def check_download
...
+
+      appcast_stanza = cask.appcast.to_s
+      appcast_contents, = curl_output("--max-time", "5", appcast_stanza)
+      version_stanza = cask.version.to_s
+      adjusted_version_stanza = version_stanza.split(",")[0].split("-")[0].split("_")[0]
+      if appcast_contents.include? adjusted_version_stanza
+        true
+      else
+        w2 = "does not contain the version number:"
+        add_error "Warning: appcast at URL '#{appcast_stanza}' #{w2} '#{adjusted_version_stanza}'\n"
```
reitermarkus: Remove `Warning: ` here.

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,24 @@ def check_download
...
+      return if cask.appcast.to_s.empty?
+
+      appcast_stanza = cask.appcast.to_s
+      appcast_contents, = curl_output("--max-time", "5", appcast_stanza)
+      version_stanza = cask.version.to_s
+      adjusted_version_stanza = version_stanza.split(",")[0].split("-")[0].split("_")[0]
+      if appcast_contents.include? adjusted_version_stanza
+        true
+      else
+        w2 = "does not contain the version number:"
```
reitermarkus: Remove this variable, you can split the string on multiple lines like this:

```ruby
"single" \
"line" \
"string"

```

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,24 @@ def check_download
...
 
+    def check_appcast_contains_version
+      return unless check_appcast?
+      return if cask.appcast.to_s.empty?
+
+      appcast_stanza = cask.appcast.to_s
+      appcast_contents, = curl_output("--max-time", "5", appcast_stanza)
+      version_stanza = cask.version.to_s
+      adjusted_version_stanza = version_stanza.split(",")[0].split("-")[0].split("_")[0]
+      return if appcast_contents.include? adjusted_version_stanza
```
reitermarkus: You can remove the `else` now.

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +297,23 @@ def check_download
...
     end
 
+    def check_appcast_contains_version
+      return unless check_appcast?
+      return if cask.appcast.to_s.empty?
+
+      appcast_stanza = cask.appcast.to_s
+      appcast_contents, = curl_output("--max-time", "5", appcast_stanza)
+      version_stanza = cask.version.to_s
+      adjusted_version_stanza = version_stanza.split(",")[0].split("-")[0].split("_")[0]
```
reitermarkus: Use `.split(/[,-_]/).first`.

Library/Homebrew/cask/audit.rb
```diff
@@ -22,6 +24,10 @@ def check_token_conflicts?
       @check_token_conflicts
     end
 
+    def check_appcast?
```
reitermarkus: Replace this with `attr_predicate :check_appcast?`.

Library/Homebrew/cask/auditor.rb
```diff
@@ -20,6 +26,10 @@ def audit_download?
       @audit_download
     end
 
+    def audit_appcast?
```
reitermarkus: Replace this with `attr_predicate :audit_appcast?, :quarantine?`.

Library/Homebrew/cask/audit.rb
reitermarkus: You'll need to add `extend Predicable` to the class (below `include Checkable`).

Library/Homebrew/cask/audit.rb
core-code: thanks

Library/Homebrew/cask/audit.rb
```diff
@@ -22,6 +25,8 @@ def check_token_conflicts?
       @check_token_conflicts
     end
 
+    attr_predicate :check_appcast?
```
reitermarkus: Move this below `attr_reader`.

Library/Homebrew/cask/auditor.rb
```diff
@@ -20,9 +28,7 @@ def audit_download?
       @audit_download
     end
 
-    def quarantine?
-      @quarantine
-    end
+    attr_predicate :audit_appcast?, :quarantine?
```
reitermarkus: Move this below `attr_reader`.

core-code: not sure if there should be a free line between attr_reader and attr_predicate, i found both styles being in use

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +296,23 @@ def check_download
...
+
+      appcast_stanza = cask.appcast.to_s
+      appcast_contents, = curl_output("--max-time", "5", appcast_stanza)
+      version_stanza = cask.version.to_s
+      adjusted_version_stanza = version_stanza.split(/[,-_]/).first
+
+      return if appcast_contents.include? adjusted_version_stanza
+
+      add_warning "appcast at URL '#{appcast_stanza}' does not contain" \
+                  "the version number: '#{adjusted_version_stanza}'\n"
```
reitermarkus: Remove the trailing newline.

Library/Homebrew/cask/audit.rb
```diff
@@ -290,6 +296,23 @@ def check_download
...
+    rescue
+      add_error "appcast at URL '#{appcast_stanza}' offline or looping\n"
```
reitermarkus: Remove the trailing newline.

core-code:
thanks for your review ;) i'll have to read up on that squashy thing

core-code:
tried to 'squash', not sure if it worked ;/

reitermarkus:
@core-code, doesn't seem like it. Look up how to do an “interactive rebase”, that's how I usually do it.

vitorgalvao:
@core-code Try `git rebase --interactive HEAD~<N>`, where `<N>` is the number of commits you want to process. In this case it should be `27`: `git rebase --interactive HEAD~27`.

The text file that opens has instructions on the bottom. You’ll likely want to use `fixup` more than `squash`.

Only change the first word in each line. If you want to modify a message, you’ll have to pick `reword` but don’t change it then; it’ll ask you after you end that step.

core-code:
thanks for your help. however i don't think that worked either. i'll try to send a new, clean, PR

core-code:
Git beat me, all the changes are lost.

vitorgalvao:
@core-code Copy the directory somewhere so you have a backup and use [`git reflog`](http://gitready.com/advanced/2009/01/17/restoring-lost-commits.html). It’s not too late to recover everything.

core-code:
@vitorgalvao many thanks, as always ;)

core-code:
re-opened as a single commit. note that i've omitted the suggested changes to 'split' on purpose, they don't work (cause an exception which forces the 'offline or looping' path)

core-code:
anything still blocking this?

reitermarkus:
Thanks, @core-code!

core-code:
thanks too ;)

core-code:
glad this is merged. 
so, what are the next steps:
• the reason for these changes were to prevent @BrewTestBot from auto-merging beta versions. if someone points me into the right direction i can send a PR to enable that
• if the common consent is that this prevents too many auto-merges, i can look into adding the aforementioned configuration so that we could specify what part of the version stanza should actually be searched for in the appcast. however, i'd need some help there, e.g. some pointers how to extend to  be able to extend the appcast stanta with optional additional arguments (akin to the referer or useragent stuff)

reitermarkus:
You will have to enable the option in `brew cask ci`, which is in the `homebrew-cask` repo.

Issue #5203, opened 2018-10-28
Don't abort entire brew cask upgrade if one cask fails


Amorymeltzer:
Right now, if you run `brew cask upgrade` and one of your casks fails (e.g., the `depends_on macos` stanza), two things happen:
1. You get the correct message (e.g. `Error: Cask cask_name depends on macOS release...`, `Error: Checksum for Cask 'superduper' does not match`, `Error: Download failed on Cask cask_name...`, etc.)
2. `brew cask upgrade` stops

`brew cask` thus won't ever upgrade any casks coming alphabetically after the failing cask.  While most issues should be fixable, if there is a macOS dependency issue, there's nothing to "fix" outside of updating to the latest macOS: a cask depending on `mojave` will obstruct `high_sierra` users, despite homebrew-cask officially supporting the two latest macOS point releases.

One workaround is to manually parse out each cask by piping `brew cask outdated` through `grep -v` and then into `brew cask upgrade`; this should work but is repetitive and obviously not ideal or easily scalable.  Another is to remove the relevant files from the caskroom and stop using `brew-cask` for the application(s) in question; likewise, this isn't ideal.

Expected behavior would be for `installer` to raise the error as appropriate but for `upgrade` to `rescue` and continue with any remaining casks.  This would match what `brew` does after #3644 and I believe is essentially what @ilovezfs suggested should be the default behavior in https://github.com/Homebrew/brew/issues/3634#issuecomment-356304225.

reitermarkus:
No objections from me.

MikeMcQuaid:
@ascheman Please 👍 in future instead, thanks.

issmirnov:
Stale bot, keep this open!

tls11:
keep it open!

ascheman:
keep it open!

jonchang:
If you are interested in adding this feature, feel free to submit it as a pull request and a maintainer will review it. Even a partial implementation is fine since we can help you work through the rest of it. "Upvoting" an issue doesn't help since no maintainer is currently working on this feature.

issmirnov:
I did a bit of digging, but don't have time right now to finish a fix. Adding code pointers to any others who are open to picking this up.

- cask upgrade code: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/cmd/upgrade.rb
- Interception point: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cask/cmd/upgrade.rb#L94

[...]

PR #5983, opened 2019-04-03
'brew cask upgrade' will continue upgrading casks after a failure

zachauten:
'cask upgrade' command collects all exceptions thrown
from individual casks during the upgrade process. If
there were more than one cask that raised exceptions
during the upgrade process, a MultipleCaskErrors
exception will be thrown.

- [x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/brew/pulls) for the same change?
- [x ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [ x] Have you written new tests for your changes? [Here's an example](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/test/PATH_spec.rb).
- [ x] Have you successfully run `brew style` with your changes locally?
- [x ] Have you successfully run `brew tests` with your changes locally?

-----
Issue #5203

### DIFF
--- a/Library/Homebrew/cask/cmd/upgrade.rb
+++ b/Library/Homebrew/cask/cmd/upgrade.rb
@@ -31,72 +31,74 @@ module Cask
 
         ohai "Casks with `auto_updates` or `version :latest` will not be upgraded" if args.empty? && !greedy?
         oh1 "Upgrading #{outdated_casks.count} #{"outdated package".pluralize(outdated_casks.count)}:"
-        cask_upgrades = outdated_casks.map do |cask|
-          if cask.installed_caskfile.nil?
-            "#{cask.full_name} #{cask.version}"
-          else
-            "#{cask.full_name} #{CaskLoader.load(cask.installed_caskfile).version} -> #{cask.version}"
+        caught_exceptions = []
+        outdated_casks.each do |cask|
+          begin
+            old_cask = CaskLoader.load(cask.installed_caskfile)
+            puts "#{cask.full_name} #{old_cask.version} -> #{cask.version}"
+            upgrade_cask(old_cask)
+          rescue CaskError => e
+            caught_exceptions << e
+            next
           end
         end
-        puts cask_upgrades.join(", ")
+        return if caught_exceptions.empty?
+        raise MultipleCaskErrors, caught_exceptions if caught_exceptions.count > 1
+        raise caught_exceptions.first if caught_exceptions.count == 1
+      end
 
-        outdated_casks.each do |old_cask|
-          odebug "Started upgrade process for Cask #{old_cask}"
-          raise CaskUnavailableError.new(old_cask, "The Caskfile is missing!") if old_cask.installed_caskfile.nil?
+      def upgrade_cask(old_cask)
+        odebug "Started upgrade process for Cask #{old_cask}"
+        old_config = old_cask.config
 
-          old_cask = CaskLoader.load(old_cask.installed_caskfile)
+        old_cask_installer =
+          Installer.new(old_cask, binaries: binaries?,
+                                  verbose:  verbose?,
+                                  force:    force?,
+                                  upgrade:  true)
 
-          old_config = old_cask.config
+        new_cask = CaskLoader.load(old_cask.token)
 
-          old_cask_installer =
-            Installer.new(old_cask, binaries: binaries?,
-                                    verbose:  verbose?,
-                                    force:    force?,
-                                    upgrade:  true)
+        new_cask.config = Config.global.merge(old_config)
 
-          new_cask = CaskLoader.load(old_cask.token)
+        new_cask_installer =
+          Installer.new(new_cask, binaries:       binaries?,
+                                  verbose:        verbose?,
+                                  force:          force?,
+                                  skip_cask_deps: skip_cask_deps?,
+                                  require_sha:    require_sha?,
+                                  upgrade:        true,
+                                  quarantine:     quarantine?)
 
-          new_cask.config = Config.global.merge(old_config)
+        started_upgrade = false
+        new_artifacts_installed = false
 
-          new_cask_installer =
-            Installer.new(new_cask, binaries:       binaries?,
-                                    verbose:        verbose?,
-                                    force:          force?,
-                                    skip_cask_deps: skip_cask_deps?,
-                                    require_sha:    require_sha?,
-                                    upgrade:        true,
-                                    quarantine:     quarantine?)
+        begin
+          # Start new Cask's installation steps
+          new_cask_installer.check_conflicts
 
-          started_upgrade = false
-          new_artifacts_installed = false
+          puts new_cask_installer.caveats
 
-          begin
-            # Start new Cask's installation steps
-            new_cask_installer.check_conflicts
+          new_cask_installer.fetch
 
-            puts new_cask_installer.caveats
+          # Move the old Cask's artifacts back to staging
+          old_cask_installer.start_upgrade
+          # And flag it so in case of error
+          started_upgrade = true
 
-            new_cask_installer.fetch
+          # Install the new Cask
+          new_cask_installer.stage
 
-            # Move the old Cask's artifacts back to staging
-            old_cask_installer.start_upgrade
-            # And flag it so in case of error
-            started_upgrade = true
+          new_cask_installer.install_artifacts
+          new_artifacts_installed = true
 
-            # Install the new Cask
-            new_cask_installer.stage
-
-            new_cask_installer.install_artifacts
-            new_artifacts_installed = true
-
-            # If successful, wipe the old Cask from staging
-            old_cask_installer.finalize_upgrade
-          rescue CaskError => e
-            new_cask_installer.uninstall_artifacts if new_artifacts_installed
-            new_cask_installer.purge_versioned_files
-            old_cask_installer.revert_upgrade if started_upgrade
-            raise e
-          end
+          # If successful, wipe the old Cask from staging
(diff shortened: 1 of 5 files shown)


Library/Homebrew/cask/cmd/upgrade.rb
```diff
@@ -40,68 +40,81 @@ def run
...
+            next
+          end
+        end
+        return if caught_exceptions.empty?
+        raise caught_exceptions[0] if caught_exceptions.length == 1
+        raise MultipleCaskErrors, caught_exceptions if caught_exceptions.length > 1
+      end
 
-          old_cask = CaskLoader.load(old_cask.installed_caskfile)
+      def self.help
```
MikeMcQuaid: Could this move back to the bottom of the file?

Library/Homebrew/cask/cmd/upgrade.rb
```diff
@@ -40,68 +40,81 @@ def run
...
-          odebug "Started upgrade process for Cask #{old_cask}"
-          raise CaskUnavailableError.new(old_cask, "The Caskfile is missing!") if old_cask.installed_caskfile.nil?
+          begin
+            upgrade_cask(old_cask)
+          rescue CaskError => e
+            caught_exceptions << e
+            next
+          end
+        end
+        return if caught_exceptions.empty?
```
MikeMcQuaid:
```ruby
if caught_exceptions.length > 1
  ...
elsif caught_exceptions.length == 1
  ...
end
``` 

might be a bit more readable?

Library/Homebrew/cask/exceptions.rb
```diff
@@ -1,6 +1,20 @@
 module Cask
   class CaskError < RuntimeError; end
 
+  class MultipleCaskErrors < CaskError
+    def initialize(errors)
+      @errors = errors
+    end
+
+    def to_s
+      message = "Problems with multiple casks:\n"
```
MikeMcQuaid:
```ruby
<<~EOS
  Problems with multiple casks:
    #{error.map(&:to_s).join("\n  ")}
EOS
```

might be more readable?

Library/Homebrew/test/cask/cmd/upgrade_spec.rb
```diff
@@ -222,4 +222,61 @@
...
+      ]
+    }
+
+    before do
+      installed.each { |cask| Cask::Cmd::Install.run(cask) }
+
+      allow_any_instance_of(described_class).to receive(:verbose?).and_return(true)
+    end
+
+    it "will not end the upgrade process" do
```
MikeMcQuaid: Nice work on testing 👏

Library/Homebrew/cask/cmd/upgrade.rb
zachauten: I agree. I originally had something like:
```
return if caught_exceptions.empty?
if caught_exceptions.length > 1
  raise MultipleCaskErrors, caught_exceptions
elsif caught_exceptions.length == 1
  raise caught_exceptions[0]
end
```
But `brew style` complained with the message `Use a guard clause instead of wrapping the code inside a conditional expression.`

It seems like the linter wants all single line if statements to be guard clauses?

Library/Homebrew/cask/cmd/upgrade.rb
```diff
@@ -40,68 +40,81 @@ def run
...
         puts cask_upgrades.join(", ")
 
+        caught_exceptions = []
         outdated_casks.each do |old_cask|
-          odebug "Started upgrade process for Cask #{old_cask}"
-          raise CaskUnavailableError.new(old_cask, "The Caskfile is missing!") if old_cask.installed_caskfile.nil?
+          begin
+            upgrade_cask(old_cask)
+          rescue CaskError => e
+            caught_exceptions << e
```
reitermarkus: Maybe collect the cask and the corresponding exception instead of just the exception.

Library/Homebrew/cask/cmd/upgrade.rb
```diff
@@ -40,68 +40,81 @@ def run
...
+            next
+          end
+        end
+        return if caught_exceptions.empty?
+        raise caught_exceptions[0] if caught_exceptions.length == 1
```
reitermarkus: Use `.first` and `.count`, respectively.

Library/Homebrew/cask/cmd/upgrade.rb
```diff
@@ -40,68 +40,81 @@ def run
...
+        raise MultipleCaskErrors, caught_exceptions if caught_exceptions.length > 1
```
reitermarkus: Also use `.count` here.

Library/Homebrew/test/cask/cmd/upgrade_spec.rb
```diff
@@ -222,4 +222,61 @@
...
+        "outdated/bad-checksum",
+        "outdated/local-transmission",
+        "outdated/bad-checksum2",
+      ]
+    }
+
+    before do
+      installed.each { |cask| Cask::Cmd::Install.run(cask) }
+
+      allow_any_instance_of(described_class).to receive(:verbose?).and_return(true)
```
reitermarkus: Is this just for debugging the tests?

Library/Homebrew/cask/cmd/upgrade.rb
```diff
@@ -40,68 +40,81 @@ def run
...
 
-          old_cask = CaskLoader.load(old_cask.installed_caskfile)
+      def self.help
+        "upgrades all outdated casks"
+      end
 
-          old_config = old_cask.config
+      def upgrade_cask(old_cask)
+        odebug "Started upgrade process for Cask #{old_cask}"
+        raise CaskUnavailableError.new(old_cask, "The Caskfile is missing!") if old_cask.installed_caskfile.nil?
```
reitermarkus: I think `CaskLoader` will raise the same exception, so this is redundant.

Library/Homebrew/cask/cmd/upgrade.rb
```diff
@@ -40,68 +40,81 @@ def run
...
 
-          old_cask_installer =
-            Installer.new(old_cask, binaries: binaries?,
-                                    verbose:  verbose?,
-                                    force:    force?,
-                                    upgrade:  true)
+        old_cask = CaskLoader.load(old_cask.installed_caskfile)
```
reitermarkus: Why reload it, if we already get it as an argument?

Library/Homebrew/cask/cmd/upgrade.rb
zachauten: What should be done with the collected casks?

Library/Homebrew/test/cask/cmd/upgrade_spec.rb
zachauten: I think so. I just added it because it was in the existing tests in the spec file.

Library/Homebrew/cask/cmd/upgrade.rb
zachauten: Not sure. I didn't actually change any of the code in `upgrade_cask`, just moved it out of a loop and into it's own method. Should I remove it?

Library/Homebrew/cask/cmd/upgrade.rb
reitermarkus: I was just thinking about outputting the name and the corresponding exception.

zachauten: I think each exception is initialized with a reference to the cask that caused the exception, and `to_s` on the exception gives the cask name.

Library/Homebrew/cask/cmd/upgrade.rb
zachauten: So it looks like the cask object loaded here is actually different from the cask object passed in as a parameter, but this cask is actually loaded earlier in the parent method for the purpose of printing out "old version -> new version". 

I'll try to change it so it doesn't get called twice.

Library/Homebrew/cask/cmd/upgrade.rb
amyspark: Add `# rubocop:disable Style/GuardClause` to this bit and it should go away.

zachauten: Thank you!

reitermarkus: Don't disable RuboCop for this. Just move the `if` after the `raise`.

zachauten: If I change the `if` back into a guard clause, then I'll have to change the `elsif` into a guard clause as well, and then the code will be the same as it was before @MikeMcQuaid requested the change.

MikeMcQuaid: @zachauten That's fine; I defer to RuboCop 😁

MikeMcQuaid: We don't get notified on force pushes so give me a shout when you've done that!

zachauten: Thanks, I didn't know that about force pushing.

zachauten:
@MikeMcQuaid @reitermarkus @amyspark I fixed that style issue!

MikeMcQuaid:
Thanks again for another great contribution @zachauten! I look forward to the next one! Feel free to email me again if you want any more brainstorming of what to work on next.

zachauten:
@MikeMcQuaid Thanks! Shoutout to @issmirnov for digging into the issue before I started on it.

issmirnov:
@zachauten thanks for picking this up and landing it!

PR #6007, opened 2019-04-10
atomic_write: decouple permissions and ownership

### LINKED ISSUE #5916: inreplace is changing file permissions
Let's consider the following minimal formula:

```
$ cat libspatialite.rb
class Libspatialite < Formula
  url "https://www.gaia-gis.it/gaia-sins/libspatialite-sources/libspatialite-4.3.0a.tar.gz"
  sha256 "88900030a4762904a7880273f292e5e8ca6b15b7c6c3fb88ffa9e67ee8a5a499"

  patch do
   url "https://raw.githubusercontent.com/Homebrew/formula-patches/27a0e51936e01829d0a6f3c75a7fbcaf92bb133f/libspatialite/sqlite310.patch"
   sha256 "459434f5e6658d6f63d403a7795aa5b198b87fc9f55944c714180e7de662fce2"
  end

  def install
    system "stat", "configure"
    inreplace "configure",
              "shrext_cmds='`test .$module = .yes && echo .so || echo .dylib`'",
              "shrext_cmds='.dylib'"
    system "stat", "configure"
    system "false"
  end
end
```

The `inreplace` statement is changing the permissions of the `configure` file, as a verbose install reveals:

```
==> stat configure
16777220 29422446 -rwxr-xr-x 1 fx wheel 0 651938 "Mar 23 22:10:35 2019" "Mar 23 22:10:35 2019" "Mar 23 22:10:35 2019" "Mar 23 22:10:35 2019" 4096 1280 0 configure
==> stat configure
16777220 29422455 -rw------- 1 fx admin 0 651852 "Mar 23 22:10:35 2019" "Mar 23 22:10:35 2019" "Mar 23 22:10:35 2019" "Mar 23 22:10:35 2019" 4096 1280 0 configure
==> false
```

In the first `stat`, `configure` was `-rwxr-xr-x`, in the second one it's `-rw-------`.

This issue is making `libspatialite` in Homebrew core fail to build, while it works working a while ago… so it is actually a `brew` regression.

amyspark:

In #5916, the demo formula fails to build because `atomic_write` (as called by `inreplace`) cannot copy group ownership between the original `./configure` and the temporary file. The original file's group owner is `wheel`, which is valid for Linux, but not for MacOS; here the correct group is `admin`.

This pull request fixes #5916 by copying the ownership/permissions in three separate steps, and adding the relevant `rescue`s. I think I should upstream this too, in case someone else in Rails runs into this.

### DIFF
--- a/Library/Homebrew/vendor/bundle/ruby/2.3.0/gems/activesupport-5.2.3/lib/active_support/core_ext/file/atomic.rb
+++ b/Library/Homebrew/vendor/bundle/ruby/2.3.0/gems/activesupport-5.2.3/lib/active_support/core_ext/file/atomic.rb
@@ -37,12 +37,20 @@ class File
 
       if old_stat
         # Set correct permissions on new file
+        # Decouple the operations to at least copy the permissions
         begin
-          chown(old_stat.uid, old_stat.gid, temp_file.path)
+          # Copy the ownership in two steps in case gid is invalid
+          chown(old_stat.uid, nil, temp_file.path)
+          chown(nil, old_stat.gid, temp_file.path)
+        rescue Errno::EPERM, Errno::EACCES => e
+          # Changing file ownership failed, moving on.
+        end
+
+        begin
           # This operation will affect filesystem ACL's
           chmod(old_stat.mode, temp_file.path)
         rescue Errno::EPERM, Errno::EACCES
-          # Changing file ownership failed, moving on.
+          # Changing file permissions failed, moving on.
         end
       end
 


amyspark:
(for clarity: `wheel` is not the original group owner of the file in #5916, it's `admin`. For some reason, `patch` changes it, and this is what triggers this bug.)

MikeMcQuaid:
@amyspark Thanks for the PR but this should be made to the upstream Rails project in https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/file/atomic.rb. Any chances we make to `vendor` will be removed when we update the gems.

